PR #7, opened 2026-06-01
feat: add per-bullet feedback panel

### LINKED ISSUE #5: Per-bullet feedback: surface which bullets to focus on changing
## Context

Surfaced by Sriya during early dogfooding — twice, in two different ways:

> *"What was confusing was which parts of the resume led to lower scores. There is some feedback under each section like carrying a metric/missing a summary but this doesn't help much for figuring out which bullet points you should focus on changing."*

The current score readout gives a *dimension-level* verdict ("you're losing points on Specificity") but doesn't say *which bullet* is the problem — e.g., "bullet #3 has no metric — add one" or "bullet #1 is two words — expand it." That gap is the difference between "I have a low score and don't know what to do" and "I have three concrete edits to make."

The decisions per-bullet are **already being computed inside the scorer and immediately discarded**. This issue is about surfacing them, not adding new heuristics.

## What's already computed per bullet

In `src/lib/score/score.ts`, three independent classifications run on every bullet inside `scoreBulletPool` (lines 131–172):

| Check | Function | Source rule |
|---|---|---|
| Has a metric? | `bulletHasMetric` (`score.ts:88`) | `STRONG_METRIC_PATTERNS` (%, $, K/M/B, Nx) + any-digit fallback with year-token strip |
| Starts with action verb? | `startsWithActionVerb` (`score.ts:106`) | hardcoded 44-word `ACTION_VERBS` set |
| Well-formed length? | inline in `scoreBulletPool` (`score.ts:155`) | `BULLET_LENGTH_MIN_WORDS` (8) ≤ word count ≤ `BULLET_LENGTH_MAX_WORDS` (30) |

These three booleans get pooled into `specificity`, `structure`, and dropped. The current `AnonymousAtsScore` shape only exposes aggregate counts (`metricBullets` / `totalBullets`, `goodBullets` / `totalBullets`) — not per-bullet rows.

The anonymous bullet extractor (`extractBulletsFromText` at `score.ts:448`) is already the source of truth for "what counts as a bullet" — preserving per-bullet identity through `scoreBulletPool` is the missing link.

### Step 1 — Add per-bullet observation shape to types

In `src/lib/score/score.ts`, add a new exported interface near `AnonymousAtsScoreDimension` (around line 347):

```ts
export interface BulletObservation {
  /** Original text of the bullet, as extracted from rawText. */
  text: string;
  /** Index in the order bullets were extracted from rawText. Stable for UI lists. */
  index: number;
  hasMetric: boolean;
  startsWithActionVerb: boolean;
  wellFormedLength: boolean;
  /** Word count after marker strip — useful for the "two words — expand it" copy. */
  wordCount: number;
}
```

### Step 2 — Have `scoreBulletPool` return per-bullet rows

Extend the function's return shape (or sit a parallel `analyzeBullets(bullets: string[]): BulletObservation[]` helper next to it — implementer's call, but the parallel-helper path keeps the math signature stable for any downstream caller).

### Step 3 — Surface bullets on `AnonymousAtsScore`

Add an optional `bullets: BulletObservation[]` field to `AnonymousAtsScore`. Optional rather than required so a future caller persisting the shape (e.g., session storage) can omit it.

Populate it inside `computeAnonymousAtsScore` (`score.ts:464`) using the existing extracted-bullets pool. If no bullets were detected at all, return an empty array — don't synthesize observations.

### Step 4 — Render the observations in the UI

In `src/components/Result.tsx` (in `ParsedCard`, currently rendering `AtsScoreReadout`), add a section that lists each bullet with its observations. Suggested shape — three pills per bullet:

```
Bullet 3: "Designed and shipped the new onboarding flow"
  [verb ✓]  [length ✓]  [metric ✗ — no number or percentage]
```

Render bullets where every check fails (or the bullet has no metric AND poor structure) more prominently so the "which ones to focus on changing" question is answered at a glance.

### Step 5 — No analytics surface changes

The PostHog payload in `src/lib/analytics.ts` already emits `score_specificity`, `score_structure`, `score_completeness` — adding aggregated per-bullet counts would be redundant with the existing `metricBullets`/`goodBullets`/`totalBullets`. **Do not** send bullet text to PostHog — that's resume content and violates the no-PII guarantee in `analytics.ts` (the "Your PDF stays in this browser tab" claim).

## UI design notes

- This is the first surface to do per-bullet feedback — which fits resumelint's "linter for resume PDFs" positioning (diagnostic, line-by-line).
- Keep the copy diagnostic, not prescriptive — name *what's missing* ("no metric"), don't tell the user what to write ("add a metric like '$2M revenue'"). See the copy discipline in `CONTRIBUTING.md`.
- Don't say "this bullet is bad" — say "this bullet has no metric." The reader infers the action.

## Acceptance criteria

- [ ] `BulletObservation` interface exported from `src/lib/score/score.ts`.
- [ ] `AnonymousAtsScore.bullets: BulletObservation[]` populated by `computeAnonymousAtsScore` from the same bullet pool the dimension scoring uses.
- [ ] A new section in `Result.tsx` (inside `ParsedCard`) renders each bullet with its three observations and word count.
- [ ] Bullets with no metric AND poor structure are visually distinguishable from clean ones.
- [ ] When `extractBulletsFromText` returns zero bullets, the new section either hides or renders an explicit "No bullet-shaped lines detected" message (no empty header).
- [ ] PostHog event shape in `analytics.ts` is unchanged — no bullet text in payloads.
- [ ] Existing 155 tests still pass; new unit tests cover the per-bullet shape (at minimum: `analyzeBullets` returns one observation per input bullet with correct flag values across the three checks).
- [ ] Algorithm-version stamp `ATS_SCORE_ALGO_VERSION` is **not** bumped — this is a presentation change. The pooled scores per dimension remain identical for the same input.

## Out of Scope

- Per-bullet rewrites or suggestion generation — that's the WebLLM bullet-rewrite pilot (separate issue), which is gated on a separate explicit-click flow. This issue is diagnostic-only.
- Expanding the action-verb set or tuning the length window. Use the current rules verbatim — tuning is its own conversation.

## Risks

- **UI density.** Three pills per bullet across a 20-bullet resume is a lot of pills. Mitigate with subdued styling for "everything-passes" bullets so the eye lands on the ones with failures.
- **False negatives on action verbs.** The 44-word set is conservative; "Authored," "Coded," "Engineered," "Owned" etc. don't match. Per-bullet display will make this visible in a way the aggregate score didn't — surface, don't fix here.

## Why this matters now

Early dogfooding is generating the first independent feedback on the surface, and this same gap came up from two different angles. It's a well-scoped first issue that's neither plumbing nor a green field — a good entry point for a new contributor.

Vaishnavi1709:
## Summary

Closes #5.

Adds a per-bullet feedback panel under the ATS score so a low Specificity / Structure number turns into "here are the bullets to focus on," not just "you lost points." Also fixes a latent bullet-detection bug that was zeroing out the score for every Microsoft Word-exported resume.

### What changed

**1. Bullet recognition: U+F0B7 (`src/lib/score/score.ts`)**

`BULLET_MARKER_RE` now includes **U+F0B7** — the Symbol-font bullet glyph Microsoft Word emits in the private-use area for every default `•`. pdfjs passes this codepoint through unchanged, so without it every Word-exported resume registered as having zero bullets. Parallel to the existing U+FFFD handling.

**2. `PerBulletFeedback` component (`src/components/Result.tsx`)**

Renders one row per extracted bullet with three check pills (verb / length / metric), highlights "needs attention" bullets in amber, and shows a header summary ("16 of 28 bullets need attention — missing a metric and at least one structure check"). Renders an empty state when no bullets are detected.

**3. Tests (`src/lib/score/score.test.ts`)**

- New U+F0B7 test mirroring the existing U+FFFD case.
- Per-bullet observation tests covering: stable index, hasMetric / startsWithActionVerb / wellFormedLength flags, empty array on no bullets, parity between the bullets array and the dimension `totalBullets`/`metricBullets` counts.

### Before / after on a Word-exported resume

Same input PDF (28 bullets, all U+F0B7-prefixed):

| | Before | After |
|---|---|---|
| Overall score | **30 / 100** | **61 / 100** |
| Specificity | — (0/0 bullets) | 10 / 40 (4/28 bullets carry a metric) |
| Structure | — (0/0 bullets) | 21 / 30 (20/28 bullets in 8–30 word window) |
| Per-bullet panel | "No bullet-shaped lines detected." | 28 rows with pills + attention summary |

### Scope notes

- No algo-version bump. The rules `scoreBulletPool` applies are unchanged; this PR only widens what the extractor *sees* as a bullet (previously-invisible bullets now get scored) and surfaces the per-bullet view that was already being computed but not rendered. The displayed score for any non-Word PDF is unchanged.
- The per-bullet panel reads `AnonymousAtsScore.bullets` (already populated by `computeAnonymousAtsScore`); no new shape needed.
- The first commit (`chore: sync package-lock.json`) is a one-line catch-up — the previous license-declaration commit on `package.json` didn't regenerate the lockfile.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run test` — 161 / 161 tests pass (includes new U+F0B7 + per-bullet cases)
- [ ] Drop a Word-exported resume (e.g. one using `•` bullets) → bullets show in the panel, Specificity/Structure report real numbers instead of `—`
- [ ] Drop a LaTeX/Markdown resume using `-` or `*` bullets → still works (regression check)
- [ ] Drop a `fonts_unmappable` PDF → falls through to `LimitedParsingCard` as before (no per-bullet panel rendered)
- [ ] Drop a single-column text PDF with no bullets at all → panel renders the "No bullet-shaped lines detected." empty state

---

This PR was originally opened against the now-retired `s-annam/resumelint` repo and is being re-opened here per maintainer request. Commits cherry-picked cleanly onto `resumelint-org/resumelint:main` (no conflicts; same pre-feature file state).

### DIFF
--- a/src/components/Result.tsx
+++ b/src/components/Result.tsx
@@ -2,7 +2,10 @@
 // Copyright 2026 The resumelint Authors
 
 import type { CascadeResult, LayoutTrigger } from "../lib/heuristics/types.ts";
-import type { AnonymousAtsScore } from "../lib/score/score.ts";
+import type {
+  AnonymousAtsScore,
+  BulletObservation,
+} from "../lib/score/score.ts";
 import { PdfPreview } from "./PdfPreview";
 
 interface ResultProps {
@@ -101,6 +104,7 @@ function ParsedCard({
 
       <LayoutFlagsList triggers={result.triggers} />
       <AtsScoreReadout score={score} />
+      <PerBulletFeedback bullets={score.bullets} />
     </section>
   );
 }
@@ -227,6 +231,127 @@ function Dimension({
         {hint}
       </p>
     </div>
+  );
+}
+
+function PerBulletFeedback({
+  bullets,
+}: {
+  bullets: BulletObservation[] | undefined;
+}) {
+  if (!bullets || bullets.length === 0) {
+    return (
+      <section className="flex flex-col gap-2">
+        <h2 className="text-xs font-semibold uppercase tracking-wider text-neutral-500">
+          Per-bullet feedback
+        </h2>
+        <p className="text-sm text-neutral-600 dark:text-neutral-400">
+          No bullet-shaped lines detected.
+        </p>
+      </section>
+    );
+  }
+
+  const attentionCount = bullets.filter(needsAttention).length;
+  const summary =
+    attentionCount === 0
+      ? `All ${bullets.length} bullets pass every check.`
+      : `${attentionCount} of ${bullets.length} bullet${
+          bullets.length === 1 ? "" : "s"
+        } need attention — missing a metric and at least one structure check.`;
+
+  return (
+    <section className="flex flex-col gap-2">
+      <h2 className="text-xs font-semibold uppercase tracking-wider text-neutral-500">
+        Per-bullet feedback
+      </h2>
+      <p className="max-w-prose text-xs text-neutral-600 dark:text-neutral-400">
+        Each bullet checked against three rules: an action verb, the 8–30-word
+        length window, and a metric. {summary}
+      </p>
+      <ul className="flex flex-col gap-1.5">
+        {bullets.map((b) => (
+          <BulletRow key={b.index} bullet={b} />
+        ))}
+      </ul>
+    </section>
+  );
+}
+
+function needsAttention(b: BulletObservation): boolean {
+  return !b.hasMetric && (!b.startsWithActionVerb || !b.wellFormedLength);
+}
+
+function BulletRow({ bullet }: { bullet: BulletObservation }) {
+  const attention = needsAttention(bullet);
+  const allPass =
+    bullet.hasMetric && bullet.startsWithActionVerb && bullet.wellFormedLength;
+
+  const containerCls = attention
+    ? "border-amber-300 bg-amber-50/40 dark:border-amber-700/60 dark:bg-amber-950/20"
+    : allPass
+      ? "border-neutral-200 dark:border-neutral-800"
+      : "border-neutral-300 dark:border-neutral-700";
+  const textCls = allPass
+    ? "text-neutral-500 dark:text-neutral-400"
+    : "text-neutral-800 dark:text-neutral-100";
+
+  const lengthLabel = bullet.wellFormedLength
+    ? `${bullet.wordCount} words`
+    : bullet.wordCount < 8
+      ? `${bullet.wordCount} words (too short)`
+      : `${bullet.wordCount} words (too long)`;
+
+  return (
+    <li
+      className={`flex flex-col gap-1.5 rounded border p-2 ${containerCls}`}
+    >
+      <p className={`text-xs leading-snug ${textCls}`}>
+        <span className="mr-1.5 font-mono text-[10px] text-neutral-400">
+          #{bullet.index + 1}
+        </span>
+        {bullet.text}
+      </p>
+      <div className="flex flex-wrap gap-1.5">
+        <CheckPill
+          ok={bullet.startsWithActionVerb}
+          okLabel="verb"
+          failLabel="no action verb"
+        />
+        <CheckPill
+          ok={bullet.wellFormedLength}
+          okLabel={lengthLabel}
+          failLabel={lengthLabel}
+        />
+        <CheckPill
+          ok={bullet.hasMetric}
+          okLabel="metric"
+          failLabel="no metric"
+        />
+      </div>
+    </li>
+  );
+}
+
+function CheckPill({
+  ok,
+  okLabel,
+  failLabel,
+}: {
+  ok: boolean;
+  okLabel: string;
+  failLabel: string;
+}) {
+  const cls = ok
+    ? "bg-emerald-50 text-emerald-800 dark:bg-emerald-900/30 dark:text-emerald-200"
+    : "bg-amber-100 text-amber-900 dark:bg-amber-900/40 dark:text-amber-200";
+  return (
+    <span
+      className={`inline-flex items-center gap-1 rounded-full px-2 py-0.5 text-[10px] font-medium ${cls}`}
+    >
+      <span>{ok ? "✓" : "✗"}</span>
+      <span>{ok ? okLabel : failLabel}</span>
+    </span>
   );
 }
 
--- a/src/lib/score/score.ts
+++ b/src/lib/score/score.ts
@@ -51,8 +51,15 @@ export const ATS_SCORE_ALGO_VERSION = "1.0";
 // splitBullets, or extractBulletsFromText sees the change.
 
 /** Bullet markers we recognize at the start of a line. Wider than typical to
- *  catch en-dash / mid-dot resumes alongside the common dash/asterisk/bullet. */
-const BULLET_MARKER_RE = /^[\s ]*[-*•●–▪◦‣▶►·]\s+/;
+ *  catch en-dash / mid-dot resumes alongside the common dash/asterisk/bullet.
+ *  U+FFFD (replacement char) handles PDFs whose font has the bullet glyph
+ *  but ships a ToUnicode map that doesn't decode it — pdfjs surfaces these
+ *  as `□`. U+F0B7 is the Symbol-font bullet glyph Microsoft Word emits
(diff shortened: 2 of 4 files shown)


src/components/Result.tsx
```diff
@@ -230,6 +234,127 @@ function Dimension({
...
+    );
+  }
+
+  const attentionCount = bullets.filter(needsAttention).length;
+  const summary =
+    attentionCount === 0
+      ? `All ${bullets.length} bullets pass every check.`
+      : `${attentionCount} of ${bullets.length} bullet${
+          bullets.length === 1 ? "" : "s"
+        } need attention — missing a metric and at least one structure check.`;
```
s-annam, line 261: [Nit]: With a single bullet this reads "1 of 1 bullet need attention" — the pluralization toggles the noun (`bullet`/`bullets`) but not the verb (`need`/`needs`). Edge case, purely cosmetic.

src/lib/score/score.ts
```diff
@@ -469,6 +520,7 @@ export function computeAnonymousAtsScore(
   // surfaces apply identical per-bullet rules.
   const bullets = extractBulletsFromText(input.rawText);
   const pool = scoreBulletPool(bullets);
+  const observations = analyzeBullets(bullets);
```
s-annam, line 523: [Suggestion]: `observations` is built regardless of the `gradable` gate just below, so a 1–2 bullet resume shows real per-bullet rows while Specificity/Structure read 0 (not gradable). Probably fine — the panel is diagnostic and the rows are still accurate — but worth a conscious call on whether the panel should also respect `ANON_MIN_BULLETS_TO_GRADE`.

s-annam (approved):
## PR Review: feat: add per-bullet feedback panel

### Summary
Clean, well-scoped implementation of #5. Surfaces the already-computed per-bullet checks as a UI panel and widens bullet detection to catch Word-exported resumes. Faithful to the spec, no scope creep, no algo-version bump. **161/161 tests pass, typecheck clean** (verified locally on the PR branch).

### Spec Alignment (#5)
| Acceptance criterion | Status | Notes |
|---|---|---|
| `BulletObservation` exported from `score.ts` | ✅ | Matches the spec's suggested shape exactly |
| `AnonymousAtsScore.bullets` populated from same pool | ✅ | `analyzeBullets(bullets)` runs on the *same* extracted array `scoreBulletPool` gets — parity is structural, not coincidental |
| New section in `ParsedCard` with 3 observations + word count | ✅ | `PerBulletFeedback` + `BulletRow` + `CheckPill` |
| No-metric-AND-poor-structure bullets visually distinct | ✅ | `needsAttention` → amber; all-pass → subdued |
| Zero-bullets → explicit message, no empty header | ✅ | "No bullet-shaped lines detected." |
| PostHog payload unchanged, no bullet text | ✅ | `analytics.ts` untouched |
| 155 existing tests pass + new per-bullet tests | ✅ | 161 total now; new U+F0B7, U+FFFD, parity, empty-array cases |
| `ATS_SCORE_ALGO_VERSION` not bumped | ✅ | Still `"1.0"` |

### Highlights
- **Verified the headline fix at the codepoint level**: `main` had *neither* U+FFFD nor U+F0B7; this PR adds both to `BULLET_MARKER_RE`. The Word-resume `30 → 61` improvement is real. (The diff renders both new chars as a single `�`, which is why it looks like one addition.)
- `analyzeBullets` reuses `bulletHasMetric` / `startsWithActionVerb` and the same word-count split as `scoreBulletPool` — so the per-bullet view can't silently drift from the dimension math. The parity test locks that in.
- Pills are diagnostic, not prescriptive ("no metric", "too short") — exactly the copy discipline #5 asked for.

### Key Findings (all non-blocking)
1. **[Nit]** The description calls U+FFFD handling "existing" — it's actually net-new in this same diff (both codepoints were added here). Code is correct; just the writeup.
2. **[Nit]** Singular grammar in the summary (see inline).
3. **[Suggestion]** Per-bullet rows render even when the pool isn't gradable (see inline).

### Verdict
**APPROVE** — 0 blocking, meets every acceptance criterion, tests green. The nits and suggestion are optional fold-ins; nothing gates the merge. Nice work, Vaishnavi.

PR #13, opened 2026-06-03
test(corpus): add snapshot harness + 6 synthetic PII-free PDFs

Vaishnavi1709:
## Summary

Closes part of #1. Replacement for PR #11 (which had real third-party PII in 4 of the 6 seed PDFs). Fresh branch from `main` so the real-PII blobs never enter `main`'s history.

Adds the snapshot-driven regression harness for the parser cascade + scorer, and seeds it with 6 PDFs across 3 categories. **All personas are synthetic** — verified at body text, PDF Info dict, XMP packet, full object tree, and raw byte level.

### Framework (unchanged from #11 — was reviewed clean)

- `src/lib/heuristics/corpus.test.ts` — auto-discovers PDFs under `tests/fixtures/pdfs/<category>/`, runs `runCascade` + `computeAnonymousAtsScore`, diffs each against a co-located `<name>.expected.json` snapshot. Schema-versioned (`SNAPSHOT_SCHEMA_VERSION`) so stale snapshots fail until re-baked.
- `package.json` — `npm run bake-fixtures` (sets `UPDATE_FIXTURES=1`).
- `vite.config.ts` — test-only alias pins `pdfjs-dist` to its legacy build so Node test env doesn't trip on `Promise.withResolvers()`. Production bundle unchanged.

### Snapshot shape — lossy by design (zero PII in JSON)

Counts, dimension scores, triggered tiers, `fieldsPopulated` as keys only — never raw text or field values.

### Seed corpus (6 PDFs / 3 categories)

| Category | Files | Persona | Generator |
|---|---|---|---|
| `latex/` | `awesome-cv-resume.pdf`, `awesome-cv-cv.pdf` | Jane Smith / `jane.smith@example.com` / `(+1) 555-010-0123` | LuaTeX (Awesome-CV) — re-exported with synthetic data |
| `latex/` | `deedy-resume-macfonts.pdf`, `deedy-resume-openfonts.pdf` | Jane Smith (same) — incl. scrubbed `publications.bib` + scrubbed collaborator names | XeLaTeX (Deedy-Resume) — re-exported |
| `word/` | `openresume-laverne-word-quartz.pdf` | Leo Leopard (genuinely synthetic upstream) | Word → macOS Quartz |
| `unknown/` | `openresume-react-pdf.pdf` | John Doe (genuinely synthetic upstream) | react-pdf |

Empty category dirs (`google-docs/`, `mac-pages/`, `mac-preview/`, `indesign/`) tracked via `.gitkeep`.

### Docs

[...]

Issue #2, opened 2026-05-29
JD matching (v1, deterministic) — paste-a-JD textarea + skill / keyword coverage check


s-annam:
## Context

The next headline feature for resumelint is JD (job description) matching. The motivating user flow: drop a resume PDF as today, then paste a JD into a textarea, and the UI surfaces *what the JD asks for that the resume doesn't claim*. This is the single biggest lever we have for repeat usage — a parser-only audit is a one-shot; a JD-match check is something a user will run on every application.

**Frame this as a lint check, not as AI tailoring.** resumelint sits in the "linter for resume PDFs" lane. "Tailoring" / "matching" framing belongs to other products in the space; resumelint stays diagnostic. The user-visible deliverable here is "your resume mentions X of the skills the JD asks for, here's what's missing" — diagnostic, not prescriptive.

V1 is deterministic — token + skill-phrase overlap, no LLM call. An LLM-assisted "suggest a rewrite that covers this missing skill" is a follow-up, gated behind the WebLLM bullet-rewrite pilot (separate issue).

## Goal

After this issue closes:

- The single-page app has a JD input alongside the existing resume dropzone.
- After both are present, the result panel shows a JD-match score, a list of skills/keywords from the JD that the resume covers, and a list of those it doesn't.
- All matching is deterministic, in-browser, byte-stays-on-device — same privacy story as the parser.
- Every claim in the UI ("we found N skills") maps to a function or constant in code — no copy that can't be traced back to logic.

### Part 1 — JD ingestion

Add a `<textarea>` in `App.tsx` that accepts a pasted JD. Strip whitespace, normalize, but otherwise leave the text intact. No file upload for v1; pasted text only.

### Part 2 — skill / keyword extraction from the JD

New file: `src/lib/jd-match/extract-jd-terms.ts`

Approach: deterministic. Two passes:

1. **Skill phrase pass.** Maintain a curated skill dictionary (`src/lib/jd-match/skills.ts`) of common tech / role skills with their canonical form and aliases. E.g. `{"react": ["react", "reactjs", "react.js"], "kubernetes": ["kubernetes", "k8s"]}`. Walk the JD text, find phrase matches (case-insensitive, word-boundary). Output a `Set<string>` of canonical skill IDs.

[...]

PR #14, opened 2026-06-04
fix(heuristics): reject resume-doc-title boilerplate as candidate name

### LINKED ISSUE #10: Name detection misreads a document header as the candidate name (and misses the real name nearby)
**Reported by @sriyau64** during product dogfooding, while testing how the parser handles redacted resumes.

### Observed
On a resume whose top line was the template header **"Functional Resume Sample"**, with the actual name a bit lower next to the rest of the contact info:
- The parser reported **"Functional Resume Sample" as the name** — consistently, across both surfaces and across three redaction variants.
- Separately, on another resume the **real name was detected only with low confidence** — likely because it was set apart from the contact block.

### Hypothesis
Name detection appears to favor the **top-most / largest text line** rather than the line **proximate to the contact cluster** (email, phone, LinkedIn). That produces both failure modes: a false positive (a header above the name wins) and a false negative / low confidence (a real name spatially separated from the contact block loses).

### Why it matters
Name presence is a **Completeness** signal, so a wrong name silently distorts the score, and "your name couldn't be detected" is exactly the false alarm that erodes trust on a privacy-first tool. It's also a clean, demonstrable parser bug — good visible fix.

### Repro
A public "Functional Resume Sample" template triggers it; so does any resume with a title/header above the name, or a name spatially separated from the contact block.

### Acceptance criteria
- [ ] Reproduce both modes (header-as-name false positive; separated-name low confidence) on fixtures
- [ ] Bias name detection toward proximity to the contact cluster rather than position/size alone
- [ ] Add snapshot tests for both fixtures
- [ ] Confirm no regression on standard single-column resumes where the name *is* the top line

Vaishnavi1709:
## Summary

Closes #10 (mode 1). Partial coverage for mode 2.

The name selector at [`extract-fields.ts:75`](src/lib/heuristics/extract-fields.ts#L75) scored "first profile line + largest font + title-case" up to 1.0, which lets a Microsoft-style `Functional Resume Sample` doc-title header win over the real candidate name on the next line. Reported by @sriyau64 during dogfooding.

The data needed for a proper fix was already on `PdfLine` (y-coords of each profile line and the contact tokens) — the selector just never consumed it. Fix is fully local to `extractName()`.

## What changed

| Layer | Behaviour |
|---|---|
| **Hard rejection** | Drops candidates where ≥60 % of tokens are doc-title boilerplate (`resume / cv / curriculum vitae / sample / template / chronological / functional / example / draft / combination / profile / biography`). `"Jane Smith Resume"` (1/3 boilerplate) **still passes**. `"Functional Resume Sample"` / `"Curriculum Vitae"` / `"Resume Sample"` all reject. |
| **First-eligible reindex** | When the literal first line is rejected, the next surviving candidate inherits the `+0.4` first-line bonus — it's effectively in the header position. Without this, the runner-up's confidence would drop to 0.45 (under `ANON_CONTACT_CONFIDENCE_FLOOR = 0.5` in [`score.ts:470`](src/lib/score/score.ts#L470)), and the completeness scorer would mark the (correctly-detected) name as *missing* — mode 2 of #10 manifesting inside the mode-1 fix. |
| **Soft contact-cluster proximity** | `+0.15` when candidate y is within ~80 pt of the email / phone / LinkedIn line, via `findContactClusterY()`. Backup signal for ambiguous header layouts; intentionally too small to override `+0.4` first-line. |

## Verification

**Programmatic**
- `npm run typecheck` — clean
- `npm run test` — **174 / 174** (167 baseline + 6 new `extractName` unit tests + 1 new corpus fixture)
- 6 existing corpus snapshots: byte-identical, zero regression
- `pdftotext` + `pdfinfo` on the new fixture: zero real-author tokens, no `Author` / `Title` / XMP metadata leakage

**Real-world PDF (the exact failure pattern in the issue)**

Verified locally against a real "Functional Resume Sample" PDF (John W. Smith / Colorado State / public template). Before / after on the same file:

| | Pre-fix (`main`) | Post-fix |
|---|---|---|
| `full_name` | `"Functional Resume Sample"` ❌ | `"John W. Smith"` ✅ |
| `given_name` | `"Functional"` | `"John"` |
| `family_name` | `"Resume Sample"` | `"W. Smith"` |
| `fieldConfidence.full_name` | `1.00` | `0.95` |
| `completeness.missing` | omits `name` (wrong name still passes the 0.5 floor) | omits `name` (right name passes) |

The score numerics don't move here because the bug is "wrong name picked confidently" rather than "no name picked" — both candidates clear the confidence floor, so completeness scores name as present either way. The user-visible win is the **displayed name** flipping from doc title to the real candidate. (The `extractName` confidence bump from 0.45 → 0.85 from the first-eligible reindex is what keeps the score from *regressing* on the fixed branch.)

## Fixture

`tests/fixtures/pdfs/latex/header-as-name-functional-resume.pdf` — synthetic single-page XeLaTeX export structured to reproduce mode 1:

```
Functional Resume Sample         (Huge, line 0 — the boilerplate)
Jane Smith                       (Large, line 1 — the real name)
123 Example Way, ...             (line 2 — contact starts)
jane.smith@example.com · ...     (line 3)
SUMMARY / EXPERIENCE / EDUCATION (3 sections, 5 bullets total)
```

LaTeX source captured in the commit body so it's regenerable. PII-clean: synthetic persona only.

## Acceptance criteria from #10

- ✅ Reproduce mode 1 (header-as-name) on a fixture
- ✅ Bias toward proximity to contact cluster (soft `+0.15` via `findContactClusterY`)
- ✅ Snapshot test pinning correct behaviour on the fixture
- ✅ Unit tests covering header-rejection + no-regression on standard top-line names (6 new tests in `extract-fields.test.ts`)
- ✅ No regression on standard single-column resumes (6/6 existing corpus snapshots byte-identical)
- ⚠️ Mode 2 ("name set apart from contact, low confidence") — the in-miniature case (boilerplate rejection dropping a candidate below the floor) is fixed via the reindex. The broader case the reporter described (Tier 1 fails entirely → Tier 1.5 falls back at 0.5) needs its own repro fixture to investigate properly. Documented in commit body as needing follow-up.

## Reviewer note — UI verification gap

The Result UI doesn't currently render `parsed.full_name` as a visible field — only the completeness "Missing" list, which is a *negative* signal. That made the fix invisible end-to-end during local verification. I worked around it with a temporary dev banner under `AtsScoreReadout` (reverted before commit). Worth a separate UX discussion about whether to surface detected fields for parser-audit verification — not filing yet, deferring to @s-annam's call on whether resumelint's "parser-audit lane" framing wants that surface.

Refs #10


src/lib/heuristics/extract-fields.ts
```diff
@@ -94,14 +165,20 @@ export function extractName(
     const letterRatio =
       text.replace(/[^A-Za-z]/g, "").length / Math.max(text.length, 1);
     if (letterRatio < 0.7) continue;
+    if (looksLikeDocTitleBoilerplate(words)) continue;
+
+    if (firstEligibleIdx === null) firstEligibleIdx = i;
 
     let score = 0;
-    if (i === 0) score += 0.4;
+    if (i === firstEligibleIdx) score += 0.4;
```
s-annam, line 173: [Suggestion]: The reindex is a great catch for the boilerplate case, but its blast radius is broader than the commit title implies. `firstEligibleIdx` is the first line surviving *all* the `continue` filters above (digits, `@`, length>60, word-count, letterRatio < 0.7), not just boilerplate rejection. So if line 0 is filtered for any of those reasons, the next line now inherits the `+0.4` first-line bonus where previously it got nothing. Likely an improvement, but worth either narrowing the comment to reflect that or confirming the intent — the 6 byte-identical snapshots are reassuring but only cover the existing corpus.

src/lib/heuristics/extract-fields.ts
```diff
@@ -64,13 +64,75 @@ function allMatches(re: RegExp, text: string): string[] {
...
 
+/**
+ * Words that signal "this is a resume document title, not the candidate's name"
+ * — e.g. "Functional Resume Sample", "Chronological CV Template". Conservative:
+ * "Jane Smith Resume" still passes because only one of three words is boilerplate.
+ * See `looksLikeDocTitleBoilerplate` below for the rule.
+ */
+const NAME_BOILERPLATE_WORDS = new Set([
+  "resume",
+  "résumé",
```
s-annam, line 75: [Nit]: This `"résumé"` entry is dead — `looksLikeDocTitleBoilerplate` lowers each token with `.replace(/[^a-z]/g, "")`, so `"Résumé"` normalizes to `"rsum"` and never matches this set member. The unaccented `"resume"` still catches the common case, so impact is small, but consider dropping it or adding the normalized form so the intent isn't misleading.

src/lib/heuristics/extract-fields.ts
```diff
@@ -64,13 +64,75 @@ function allMatches(re: RegExp, text: string): string[] {
...
+ *  or undefined if no contact-bearing line is found. Used as a soft signal —
+ *  candidate names close to this y get a small bonus. */
+function findContactClusterY(lines: PdfLine[]): number | undefined {
+  for (const line of lines) {
+    if (
+      EMAIL_RE.test(line.text) ||
+      PHONE_RE.test(line.text) ||
+      LINKEDIN_RE.test(line.text)
+    ) {
+      // Reset lastIndex defensively; the constants are recompiled per call
```
s-annam, line 114: [Nit]: Minor comment inaccuracy — `EMAIL_RE`/`PHONE_RE`/`LINKEDIN_RE` are module-level constants, not "recompiled per call." The defensive `lastIndex = 0` reset is correct and harmless (global regexes auto-reset `lastIndex` on a failed `.test()`, and the success path resets before returning), so no behavior change needed — just the comment.

s-annam (approved):
## PR Review: reject resume-doc-title boilerplate as candidate name

### Summary
Clean, well-scoped, well-tested fix for mode 1 of #10. The hard-rejection approach is sound and the author is admirably transparent about what's *not* covered (mode 2) and about the verification gaps. **174/174 green, typecheck clean, fixture PII-verified locally** (`Jane Smith` / `@example.com` / `555` phone, no `Author`/`Title`/XMP leakage). No blocking issues.

### Spec Alignment (issue #10)
|---|---|---|
| Reproduce mode 1 (header-as-name) on a fixture | ✅ | `header-as-name-functional-resume.pdf` |
| Reproduce mode 2 (separated-name, low confidence) | ⚠️ Partial | In-miniature case fixed via reindex; broader case deferred with rationale |
| Bias toward contact-cluster proximity | ✅ (soft) | `+0.15` via `findContactClusterY`; intentionally too weak to overturn `+0.4` first-line — the actual mode-1 fix is the hard rejection, not proximity |
| Snapshot test on fixture | ✅ | mode 1 |
| No regression on standard single-column resumes | ✅ | 6 corpus snapshots byte-identical + new unit test |

### Highlights
- Fix is fully local to `extractName()` and reuses data already on `PdfLine` — no new plumbing.
- The `firstEligibleIdx` reindex is a genuinely sharp catch: it stops the mode-1 fix from regressing into mode-2 (dropping the real name below the 0.5 completeness floor). The unit test guarding that threshold boundary is excellent.
- Conservative `≥60%`-boilerplate rule, with a test pinning that `"Jane Smith Resume"` still passes.

### Key Findings (all non-blocking)
1. **[Suggestion] `Closes #10` will auto-close the issue on merge**, but mode 2's acceptance criteria aren't met. Consider `Refs #10` + a follow-up issue for mode 2, so the remaining work isn't silently lost.
2. **[Suggestion] Merge ordering** — this PR is stacked on #13 (it carries `974761f … (#13)`, the corpus harness). The new fixture's snapshot only runs via `corpus.test.ts` from #13. Merge #13 first, or #14 drags the whole harness in with it.
3. **[Suggestion] `firstEligibleIdx` blast radius** is wider than "boilerplate" — see inline.
4. **[Nit] Dead set entry** `"résumé"` — see inline.
5. **[Nit] Comment inaccuracy** in `findContactClusterY` — see inline.

### Verdict
**APPROVE** — 0 blocking items; findings are suggestions/nits that needn't gate merge. Recommend landing #13 first.

*Verified locally: full suite (incl. corpus snapshot) 174/174, `tsc -b --noEmit` clean, fixture binary inspected with `pdftotext`/`pdfinfo`.*

s-annam:
**Correction to my review above** — finding #2 ("merge ordering / this PR is stacked on #13") is **obsolete**: #13 was already merged into `main` (commit `974761f`, 2026-06-04). My local checkout was a commit behind, so I mis-read the merge-base. GitHub's 4-file diff is correct, and this PR applies directly on current `main` with the corpus harness already present.

The verification still stands — the full suite I ran (174/174, incl. the new corpus snapshot) reflects the real post-merge state. The remaining findings (1, 3, 4, 5) are unchanged.

Issue #1, opened 2026-05-29
Parser corpus expansion — find more Tier 0 / Tier 1 edge cases with diverse resume PDFs


s-annam:
## Context

An early parsing bug (found by Vaishnavi running the v1 slice on her own resume) exposed a real failure mode: Tier 0's `hasEOL`-driven text assembly collapsed bullets into a single paragraph, and Specificity + Structure clamped to 0. The fix replaced the assembly with positional-line reassembly and added unit tests in `src/lib/heuristics/pdf-extract.test.ts`.

That fix covered *one* failure mode that happened to show up on the first PDF dropped on the dropzone. We don't know how many similar latent issues exist in the heuristics — the test corpus is small, and we've only ever exercised the parser against PDFs from a handful of generators.

This issue tracks expanding the test-PDF corpus and the regression-test surface so we can find the next 5–10 bugs like that one *before* a user does.

## Goal

After this issue closes:
- We have at least 15 diverse resume PDFs under a tracked directory, each producing a known-good or known-broken cascade result.
- Each new failure mode found in the corpus has a corresponding unit test in `src/lib/heuristics/` or `src/lib/score/`.
- The repo has a documented process for adding a new PDF to the corpus (e.g. drop the file, run a fixture-update script, commit both).

### Part 1 — set up the corpus

Create `tests/fixtures/pdfs/` with subfolders for source diversity:

- `latex/` — Jake's Resume, Awesome-CV, Deedy-Resume, moderncv variants (at least 4)
- `word/` — Word-to-PDF exports from at least 2 different Word versions
- `google-docs/` — Google Docs "Download as PDF" exports (at least 2)
- `mac-pages/` and `mac-preview/` — macOS Pages and Preview re-saves (at least 2)
- `indesign/` — if accessible, 1 InDesign export
- `unknown/` — 2–3 PDFs from real users (your own + ask friends)

For each PDF, also commit a tiny `*.expected.json` next to it with the expected cascade outputs (which fields populate, whether bullets parse, whether the file triggers `fonts_unmappable`). The expected file is the snapshot — when the cascade output changes, the test fails and you decide whether the snapshot needs updating or the cascade has a bug.

### Part 2 — fixture-driven tests

Add `src/lib/heuristics/corpus.test.ts`:

- Walk `tests/fixtures/pdfs/**/*.pdf`
- For each PDF, run `runCascade(buffer)` and `computeAnonymousAtsScore(...)`
[...]

PR #18, opened 2026-06-08
fix: bullet count under-reports on short bullets and two-column layouts

### LINKED ISSUE #9: Bullet count under-reports — parser shows 18 vs. 19 extracted lines vs. 20+ in source PDF
**Reported by @sriyau64** during product dogfooding, comparing the same resume across the production and OSS surfaces.

### Observed
On a multi-section resume that uses bullets, three counts disagree down a chain:
- **20+** bullet lines visible in the source PDF
- **19** lines in the plain-text extraction box
- **18** reported as the bullet count in the score breakdown

So we lose ~1 bullet between extraction and counting, and a few more between source and extraction.

### Hypothesis
Some bullets sit in a **column layout**, which likely explains the source→extraction loss (column reads can drop/merge lines). The **19→18** gap is the more interesting one — it's *after* extraction, so it points at the bullet-detection/counting step itself (a wrap-continuation line being merged, or a leading bullet glyph not matching the detector).

### Why it matters
Per-bullet feedback (#5) is built on accurately identifying *which* lines are bullets. If the count is off by one before we start, per-bullet guidance will mis-map. Worth nailing the counter before that feature leans on it.

### Repro
The exact PDF is held by the reporter (columns + 20+ bullets) — or construct a fixture with a known bullet count and a two-column experience section. Fold it into the corpus snapshot tests (#1) once isolated.

### Acceptance criteria
- [ ] Reproduce the 19→18 discrepancy on a fixture with a known bullet count
- [ ] Root-cause the post-extraction loss (counter vs. detector)
- [ ] Add a snapshot test pinning the expected bullet count for the fixture
- [ ] Document any *intentional* exclusions (e.g. sub-bullets) so the count is explainable

Vaishnavi1709:
## Summary

Closes #9. Two distinct root causes both manifest as bullet under-reporting; both fixed here, in two cleanly-separated commits.

| Commit | Root cause | Files | Affected fixtures |
|---|---|---|---|
| `126ad13` | `ANON_BULLET_MIN_WORDS = 4` in score.ts silently dropped legitimate short bullets like `"• Everything that matters."` | `src/lib/score/score.ts`, `score.test.ts` | Awesome-CV cv / resume (off-by-1, the issue's reported pattern) |
| `aabae8a` | `groupIntoLines` clusters items purely by y-proximity; on two-column layouts the right-column bullet glyph ends up after the left-column text and never reaches line-start | `src/lib/heuristics/sections.ts`, `pdf-extract.test.ts` | Deedy macfonts / openfonts (catastrophic — `bulletCount: 1`/`2` vs 8 visible) |

## Root cause #1 (Awesome-CV off-by-1)

`extractBulletsFromText` was applying a 4-word floor before any line could count as a bullet, conflating *bullet detection* with *bullet grading*. The reporter's pattern (`20+ visible · 19 extracted · 18 reported`) is one short bullet getting filtered post-extraction.

Localised by tracing every marker-prefixed line through the pipeline. The dropped bullet on both Awesome-CV fixtures was the same 3-word line: `"• Everything that matters."`.

**Fix:** lower `ANON_BULLET_MIN_WORDS` from 4 to 1. Empty marker-only lines (`"• "`) are still skipped because `split(/\s+/).filter(Boolean)` returns length 0. Quality grading is now left entirely to the existing well-formed length window (8–30 words) in `analyzeBullets` / `scoreBulletPool`, which naturally penalises short bullets via per-bullet feedback chips rather than hiding them.

## Root cause #2 (Deedy two-column catastrophic loss)

`groupIntoLines` clusters items by y-proximity (`LINE_Y_EPS = 3.5`) with no concept of column structure. On Deedy's asymmetric 0.33/0.66 layout, the left-column education text and right-column experience bullets land at the same y, so the bullet glyph gets concatenated *after* the education text. `extractBulletsFromText` correctly skips it (it requires `^\s*[bullet glyph]\s+`) and the cascade reports `bulletCount: 1` against 8 visible bullets.

The asymmetric layout doesn't trigger `isTwoColumn` in `pdf-layout.ts` (which requires roughly equal columns + 60% density), so gating on that flag wasn't an option.

**Fix:** in `groupIntoLines`'s `flush()`, scan the (x-sorted) same-y cluster for any gap ≥ `COLUMN_GAP_THRESHOLD = 50pt` between consecutive items and emit each side as its own `PdfLine`.

**Why 50pt:** measured the in-line gap distributions across all 7 corpus fixtures. Awesome-CV's `\hfill` lines produce 0pt gaps because LuaTeX includes trailing whitespace in item widths. Deedy's column jumps in at 70–130pt. 50pt is safely between the two regimes.

## Empirical impact

`visible bullets in pdftotext` vs `reported bulletCount` per fixture, before vs after:

| Fixture | Pre-fix delta | Post-fix delta |
|---|---|---|
| `awesome-cv-cv` | **−1** | **0 ✅** |
| `awesome-cv-resume` | **−1** | **0 ✅** |
| `deedy-resume-macfonts` | **−7** | **0 ✅** |
| `deedy-resume-openfonts` | **−6** | **0 ✅** |
| `header-as-name-functional-resume` | 0 | 0 |
| `openresume-react-pdf` | 0 | 0 |
| `openresume-laverne-word-quartz` | 0 | 0 |

## Snapshot deltas (all correct, none regressions)

**Awesome-CV** — overall score −1 on each. The previously-hidden short bullet has no metric and is outside the 8–30 word window, so it correctly adds 1 to both Specificity and Structure denominators without contributing to either numerator. Pre-fix scores were inflated by silently dropping a bad bullet.

**Deedy** — line-grouping fix unblocks every downstream parser, not just the bullet counter:

| | macfonts | openfonts |
|---|---|---|
| `bulletCount` | 1 → **8** | 2 → **8** |
| `skillsCount` | 0 → **25** | 0 → **24** |
| `experienceCount` | 0 → **6** | 0 → **6** |
| `metricBullets` | 0 → 1 | 0 → 1 |
| `goodBullets` | 1 → 6 | 1 → 6 |
| `overall` | 17 → **47** | 17 → **47** |

The Deedy PDFs were essentially unparseable before — the parser wasn't dropping individual fields, it was misreading the line structure end-to-end. Score moving from 17 to 47 reflects the parser actually seeing the resume's content.

**openresume-react-pdf** — `skillsCount` 16 → 20. The skills line `HTML  CSS  Python  TypeScript  React  C++` had ~160–175pt gaps; splitting these into per-token lines exposed 4 additional skills the parser had been missing. Bonus.

**Awesome-CV / header-as-name / laverne** — byte-identical, zero negative impact.

## Tests added

- `score.test.ts` (1 test) — pins that short bullets count toward the displayed total while still being graded low, guarding against re-introducing the off-by-1 regression.
- `pdf-extract.test.ts` (2 tests) — pins the two-column same-y split behaviour AND a single-column regression guard (small gaps must NOT fragment a line).

## Verification

- `npm run typecheck` — clean
- `npm run test` — **177 / 177** (175 baseline + 3 new)
- `npm run bake-fixtures` → `git diff` after — empty (snapshots round-trip)
- `pdftotext`-vs-reported bulletCount: delta = 0 on every fixture (was −1, −1, −7, −6 pre-fix)

## Reviewer note on blast radius

The Deedy fix touches `groupIntoLines`, which is called by every PDF parse path. Risk is mitigated by:
- the single-column regression test (small gaps stay merged)
- 7-fixture corpus all green with no negative metric changes
- the 50pt threshold being measured empirically against the actual gap distributions in the corpus

But it's worth a closer read than the Awesome-CV constant change — flagging up-front in case you want to scrutinise the `flush()` split logic specifically.

Closes #9

### DIFF
--- a/src/lib/score/score.ts
+++ b/src/lib/score/score.ts
@@ -469,11 +469,16 @@ export interface AnonymousAtsScoreInput {
 
 const ANON_CONTACT_CONFIDENCE_FLOOR = 0.5;
 const ANON_MIN_BULLETS_TO_GRADE = 3;
-/** Word-count floor for raw-text bullet extraction. The authed splitBullets
- *  uses a char floor instead because its input is already a curated
- *  description string; raw-text extraction needs to skip headers / one-line
- *  section labels that share a leading marker. */
-const ANON_BULLET_MIN_WORDS = 4;
+/** Word-count floor for raw-text bullet extraction. Set to 1 so the displayed
+ *  bullet count matches what the user can see in the PDF — every line that
+ *  begins with a recognised marker AND has at least one non-empty word is a
+ *  bullet. Quality grading (well-formed length window, action verb, metric)
+ *  is handled downstream by `scoreBulletPool` and `analyzeBullets`, which
+ *  naturally penalise short bullets without hiding them from the count.
+ *  Issue #9 — previously set to 4, which silently dropped legitimate short
+ *  bullets like "• Everything that matters." and caused bullet count to
+ *  under-report vs. the visible PDF. */
+const ANON_BULLET_MIN_WORDS = 1;
 
 const ANON_CONTACT_FIELDS: readonly {
   key: "full_name" | "email" | "phone" | "location" | "linkedin_url";
@@ -489,7 +494,11 @@ const ANON_CONTACT_FIELDS: readonly {
 /**
  * Pull bullet-like lines out of raw resume text. A line counts as a bullet
  * when it starts with a recognized bullet marker (`-`, `•`, etc.) or a
- * numbered list prefix and contains 4+ words after the marker is stripped.
+ * numbered list prefix and the stripped line contains at least one word.
+ * The displayed count is meant to match what a reader sees in the PDF —
+ * short or low-quality bullets are still bullets, they just get flagged
+ * by the downstream length / verb / metric checks in `scoreBulletPool`
+ * and `analyzeBullets`.
  *
  * We deliberately do NOT try to grade unmarked indented lines — most modern
  * resumes use markers, and grading paragraphs would either over-count
--- a/tests/fixtures/pdfs/unknown/openresume-react-pdf.expected.json
+++ b/tests/fixtures/pdfs/unknown/openresume-react-pdf.expected.json
@@ -23,7 +23,7 @@
       "skills",
       "website_url"
     ],
-    "skillsCount": 16,
+    "skillsCount": 20,
     "experienceCount": 1,
     "educationCount": 1,
     "rawTextCharCount": 1923,
--- a/src/lib/heuristics/pdf-extract.test.ts
+++ b/src/lib/heuristics/pdf-extract.test.ts
@@ -68,4 +68,38 @@ describe("assembleTextFromLines", () => {
   it("returns empty string for empty input (scanned PDFs)", () => {
     expect(assembleTextFromLines([])).toBe("");
   });
+
+  it("splits two-column same-y items so right-column bullets keep line-start position (issue #9)", () => {
+    // Deedy-style two-column layout: left column has education text, right
+    // column has experience bullets, both at the same baseline y. Pre-fix
+    // groupIntoLines merged them into a single PdfLine like
+    //   "BS in Computer Science                            • Led migration"
+    // which dropped the bullet glyph out of line-start position and the
+    // score-side bullet counter missed it. The 50pt column-gap split keeps
+    // each column on its own line.
+    const items = [
+      // Left column: education text starting at x=35
+      item("BS in Computer Science", 35, 200, 1, 120),
+      // Right column: experience bullet starting at x=300 — 145pt gap
+      item("•", 300, 200, 1, 6),
+      item("Led migration of legacy auth system", 315, 200, 1, 200),
+    ];
+    const text = assembleTextFromLines(items);
+    const lines = text.split("\n").filter((l) => l.length > 0);
+    expect(lines).toHaveLength(2);
+    expect(lines[0]).toBe("BS in Computer Science");
+    expect(lines[1]).toMatch(/^•\s+Led migration/);
+  });
+
+  it("does NOT split a normal single-column line at small gaps", () => {
+    // Regression guard: a tightly-laid-out single-column line with normal
+    // word/run spacing (gaps well under 50pt) must stay as one PdfLine,
+    // otherwise we'd fragment every Awesome-CV / OpenResume-style PDF.
+    const items = [
+      item("Jane", 72, 100, 1, 30),
+      item("Smith", 110, 100, 1, 35),       // ~8pt gap
+      item("Senior Engineer", 160, 100, 1, 100),  // ~15pt gap
+    ];
+    expect(assembleTextFromLines(items)).toBe("Jane Smith Senior Engineer");
+  });
 });
--- a/src/lib/heuristics/sections.ts
+++ b/src/lib/heuristics/sections.ts
@@ -42,6 +42,19 @@ export interface PdfSection {
 /** Items within this vertical distance (PDF points) are treated as same line. */
 const LINE_Y_EPS = 3.5;
 
+/**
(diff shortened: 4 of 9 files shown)


src/lib/score/score.ts
```diff
@@ -469,11 +469,16 @@ export interface AnonymousAtsScoreInput {
...
+/** Word-count floor for raw-text bullet extraction. Set to 1 so the displayed
+ *  bullet count matches what the user can see in the PDF — every line that
+ *  begins with a recognised marker AND has at least one non-empty word is a
+ *  bullet. Quality grading (well-formed length window, action verb, metric)
+ *  is handled downstream by `scoreBulletPool` and `analyzeBullets`, which
+ *  naturally penalise short bullets without hiding them from the count.
+ *  Issue #9 — previously set to 4, which silently dropped legitimate short
+ *  bullets like "• Everything that matters." and caused bullet count to
+ *  under-report vs. the visible PDF. */
+const ANON_BULLET_MIN_WORDS = 1;
```
s-annam, line 481: [Suggestion]: The original comment on this constant explicitly noted: "raw-text extraction needs to skip headers / one-line section labels that share a leading marker." Lowering to 1 means `"• Summary"` or `"• Education"` (1 word after marker strip) now counts as a bullet. The corpus shows no regression across 7 fixtures, so this is not blocking — but the guard is removed without a test that pins the boundary. Consider a unit test in `score.test.ts` that feeds a resume text with a single-word bullet-prefixed section label and asserts it does **not** inflate the bullet count (or asserts the expected behavior explicitly), so future changes to this threshold can't silently regress.

tests/fixtures/pdfs/unknown/openresume-react-pdf.expected.json
```diff
@@ -23,7 +23,7 @@
       "skills",
       "website_url"
     ],
-    "skillsCount": 16,
+    "skillsCount": 20,
```
s-annam, line 26: [Nit]: The PR body explains this well (the skills line `HTML  CSS  Python  TypeScript  React  C++` has ~160–175pt inter-token gaps, so the column-split logic emits each token as its own `PdfLine`, surfacing 4 extra skills). But this is a side-effect of the column-gap fix, not the primary intent, and `.expected.json` has no comment slot. Consider adding a note in a fixture-level README or the snapshot test group explaining that this fixture's skills line has unusually wide inter-token gaps — otherwise a future reader sees the count jump with no in-repo explanation.

s-annam (approved):
## PR Review: fix: bullet count under-reports on short bullets and two-column layouts

### Summary
Two independent bugs, two clean commits, both fixed correctly. Root cause analysis is thorough, the 50pt threshold is empirically calibrated against the real corpus, and all 7 fixtures are verified with zero regressions.

### Highlights
- PR description quality sets a high bar — blast-radius self-assessment, measured gap distributions, per-fixture before/after table. Reference example for how to document a parser change.
- Two-commit structure cleanly isolates each bug, making bisect easy if either fix needs reverting.
- `buildLine` extraction from `flush()` is a clean refactor — makes the new split loop readable.
- `assembleTextFromLines` → `groupIntoLines` chain is correctly exercised by the new tests (confirmed: `pdf-extract.ts:35`).
- Single-column regression guard is the right safety net for a change to a widely-called function.
- Deedy: 17 → 47 is not incremental improvement — the parser now actually reads the resume.

### Key Findings

1. **[Suggestion]** `ANON_BULLET_MIN_WORDS = 1` — single-word section header edge case — see inline comment on `score.ts`.

2. **[Nit]** `openresume-react-pdf` `skillsCount` 16 → 20 side-effect has no in-repo explanation — see inline comment on the fixture.

### Verdict
**Action:** APPROVE
**Rationale:** No blocking issues. Implementation is correct, empirically grounded, and the single-column regression guard is specifically designed to catch the blast-radius risk. The `ANON_BULLET_MIN_WORDS` suggestion is future-hardening, not a merge blocker.

---
*0 blocking · 1 suggestion · 1 nit*

🤖 Reviewed with [Claude Code](https://claude.com/claude-code)

PR #39, opened 2026-06-11
feat(webllm): per-bullet "Suggest a rewrite" pilot — WebGPU-only, Qwen2-1.5B

### LINKED ISSUE #3: WebLLM bullet-rewrite pilot (v1.0 core) — per-bullet "Suggest a rewrite", Qwen2-1.5B, WebGPU-only
## Context

resumelint's "trial-and-error LLM bullet rewrite" is a designed-but-not-yet-shipped capability:

- A browser-side (WebLLM) rewrite provider. Cloud / desktop providers are out of scope for this repo.
- Model: **Qwen2-1.5B-Instruct-q4f16_1-MLC** (~1.2GB, Apache-2.0). Pinned. Do not switch to Phi-3 or Llama-3 without measurement justification.
- WebGPU-only. No WASM fallback. Browsers without WebGPU **hide the CTA** rather than silent-degrade.
- First capability: per-bullet "Suggest a rewrite" affordance on the existing result panel. Short input, short output, model-suited task.
- Lazy load only: weights downloaded **on click**, never on page render. Cold-start UX is the conversion killer, not inference cost.

Rohit expressed interest in running a local model on his Mac — this is a good fit. Samhit can take it instead if Rohit prefers JD-match.

## Goal

After this issue closes:

- On the resume result panel, each parsed bullet has a "Suggest a rewrite" button.
- On browsers without WebGPU, the button is hidden — not greyed out, not a fallback, hidden.
- On supported browsers, clicking the button shows a progress UI while the model downloads (first time only), then a rewritten bullet appears inline.
- Bytes never leave the browser. No network call after the model weight download.
- Telemetry (env-gated, see existing `src/lib/analytics.ts`) captures: `webllm_capability_detected`, `webllm_download_started`, `webllm_loaded`, `webllm_first_rewrite`. Same `snake_case` naming and anonymization rules as existing events.

### Part 1 — capability detection

New file: `src/lib/webllm/capability.ts`

```ts
export type WebGpuCapability = "available" | "no-webgpu" | "unsupported-os";

export async function detectWebGpu(): Promise<WebGpuCapability> { ... }
```

Detect WebGPU via `navigator.gpu` + `requestAdapter()`. Return `"available"` only if an adapter is granted. Cache the result for the page lifetime.

### Part 2 — provider scaffolding

New folder: `src/lib/webllm/`

- `web-llm.ts` — pinned `MODEL_ID` constant, `loadEngine()` lazy loader (dynamic import of `@mlc-ai/web-llm`), progress callback for UI
- `rewrite-bullet.ts` — pure function `rewriteBulletWithLlm(bullet: string, engine: WebLlmEngine): Promise<string>` with a tight prompt (see notes)
- `types.ts` — `WebLlmEngine` interface + capability type re-export

Pin the WebLLM npm version in `package.json`. Lock the model ID in one constant — `MODEL_ID = "Qwen2-1.5B-Instruct-q4f16_1-MLC"`.

### Part 3 — UI

New component: `src/components/RewriteButton.tsx`

- Render a small button next to each parsed bullet
- On mount: call `detectWebGpu()`. If not `"available"`, render `null`.
- On click: load the engine (with download progress shown inline) → call `rewriteBulletWithLlm` → display the result inline below the original bullet, with a "use this" copy-to-clipboard affordance
- First click on the page triggers the model download (~1.2GB); subsequent clicks are instant

Update `Result.tsx` to render `<RewriteButton bullet={b} />` per parsed bullet from the cascade output.

### Part 4 — progress + fallback UI

When the model is downloading for the first time:

- Inline progress bar with percent + downloaded MB
- Affirmative copy: "Loading the bullet-rewrite model (~1.2GB, one-time download)"
- A "What's happening?" disclosure that explains: model runs entirely in the browser, never leaves the device, takes about a minute on a typical connection

When `detectWebGpu()` returns `"no-webgpu"` or `"unsupported-os"`:

- Hide the button entirely. Don't show a "your browser isn't supported" message — that's friction with no payoff.
- Document this stance in the component file's header comment so future contributors don't add a fallback.

### Part 5 — telemetry

Add to `src/lib/analytics.ts`:

- `webllm_capability_detected` — fires once per session, payload: `{ capability: "available" | "no-webgpu" | "unsupported-os" }`
- `webllm_download_started` — fires when the user clicks the first rewrite button
- `webllm_loaded` — fires when the engine is ready
- `webllm_first_rewrite` — fires on first successful rewrite

Same env-gating as existing events (`VITE_POSTHOG_KEY` unset → no-op + dead-code-eliminated at build time).

### Part 6 — prompt

The prompt template lives in `rewrite-bullet.ts`. Keep it short and concrete:

```
You are rewriting a single resume bullet to be more specific and outcome-oriented.
Rules:
- Keep it one line.
- Lead with a strong verb.
- Preserve concrete numbers from the original.
- If the original is already strong, return it unchanged.

Original: {bullet}
Rewritten:
```

Tune iteratively. The prompt belongs in code as a string constant so it round-trips to UI copy claims.

## Where to share progress

Drop progress, screenshots of working rewrites, and questions as comments on this issue as you go. The first rewrite working end-to-end on your machine is a good moment for a screenshot or short Loom. Anything blocking you, reach out — that's what I'm here for.

## Acceptance criteria

- [ ] WebLLM npm dep pinned in `package.json`
- [ ] Model ID pinned via a single constant in `src/lib/webllm/web-llm.ts`
- [ ] `detectWebGpu()` correctly returns `"no-webgpu"` on Firefox without flags and `"available"` on Chrome 121+
- [ ] Button hidden entirely on `"no-webgpu"` / `"unsupported-os"` (no fallback message, no greyed button)
- [ ] On supported browsers, the button appears next to each parsed bullet
- [ ] First click triggers download with inline progress UI
- [ ] Subsequent clicks are instant (engine cached for page lifetime)
- [ ] Rewritten bullet displayed inline with a copy-to-clipboard affordance
- [ ] No network calls after model download — verified by DevTools Network tab
- [ ] All four telemetry events fire correctly when `VITE_POSTHOG_KEY` is set; verify in PostHog
- [ ] Telemetry compiles to no-op when `VITE_POSTHOG_KEY` is unset; verify via `npm run build` + grep
- [ ] Build size impact: the WebLLM lib is dynamic-imported, so the main entry chunk should not grow by more than ~5KB; verify with `npm run build` output
- [ ] Existing tests still pass

## Out of scope

- Anything beyond bullet rewrite (cover letters, full-resume rewrite, JD-tailored rewrite) — follow-up issues
- WASM fallback for non-WebGPU browsers — explicitly out of scope per the design
- Authenticated / cloud rewrite paths — out of scope for this repo
- Model switching at runtime — pinned single model for v1
- Telemetry beyond the four events listed — focus on the funnel

## Notes for the implementer

- WebLLM docs: https://github.com/mlc-ai/web-llm. The README has the standard "load + generate" pattern. The model IDs are listed in their model registry.
- The cold-start UX rule is **non-negotiable** — never trigger the download on page render, scroll-into-view, or hover. Click only.
- The "hide on unsupported" rule is also non-negotiable — silent degradation is worse than hiding the feature.
- The prompt is going to need 5–10 iterations to produce non-garbage output on a 1.5B model. That's normal. The "rules" in the prompt above are a starting point, not the answer.
- If the model returns hallucinated numbers (made-up percentages, etc.), tighten the "preserve concrete numbers" rule into a more emphatic constraint.
- Browser support today: Chrome desktop, Edge desktop, Safari 18+ desktop. iOS Safari and Firefox-without-flag don't qualify.

Vaishnavi1709:
## Summary

First in-browser AI capability: a per-bullet "Suggest a rewrite" CTA that runs Qwen2-1.5B-Instruct via WebLLM on the user's GPU. Bytes never leave the browser. Resolves #3.

**Library** — `src/lib/webllm/`
- `types.ts` — narrow `WebLlmEngine` contract (only the surface `rewrite-bullet` consumes) so tests stub without importing the real lib.
- `capability.ts` — `detectWebGpu()` returns `"available" | "no-webgpu" | "unsupported-os"`, cached as a module-level promise so the discriminator fires exactly once per page.
- `web-llm.ts` — pinned `MODEL_ID = "Qwen2-1.5B-Instruct-q4f16_1-MLC"`. `loadEngine()` dynamic-imports `@mlc-ai/web-llm` and caches the engine for the page lifetime — concurrent clicks across bullets share one download.
- `rewrite-bullet.ts` — pure `rewriteBulletWithLlm(bullet, engine)`. Single-source-of-truth `BULLET_REWRITE_SYSTEM_PROMPT`; post-processing strips a `"Rewritten:"` prefix, wrapping quotes, and trims to the first non-empty line.

**UI** — `src/components/RewriteButton.tsx`
- On mount: `detectWebGpu()`; if not `"available"`, the component returns `null`. **Hidden, not greyed.** Header comment documents the stance so a future contributor doesn't "improve" it with a fallback banner.
- On click: progress panel (percent + status text + green progress bar + "What's happening?" disclosure that mirrors the prompt rules).
- On done: inline green box with rewritten text + "Use this — copy to clipboard".
- Wired into `Result.tsx::BulletRow` next to the existing check pills via `<RewriteButton bullet={bullet.text} />`.

**Telemetry** — `src/lib/analytics.ts`
- Four new `snake_case` events through the existing env-gated `track()` helper: `webllm_capability_detected` (payload `{capability}`), `webllm_download_started`, `webllm_loaded`, `webllm_first_rewrite`. Each fires at most once per page.
- Compiles to no-op when `VITE_POSTHOG_KEY` is unset — verified no `posthog` string in built assets.

**Bundle** — the `@mlc-ai/web-llm` dynamic import keeps the library in its own ~5.7 MB chunk (`dist/assets/index-CereMTgS.js`); the entry chunk holds only the tiny dynamic-import reference. Per AC, main chunk does not grow >5 KB.

## Out of scope (explicit per issue)
- WASM fallback for non-WebGPU browsers
- Cover letters, full-resume rewrite, JD-tailored rewrite
- Cloud rewrite paths, runtime model switching
- Telemetry beyond the four events above

## Test plan
- [x] `npm run test` — 194 tests pass (13 new in `src/lib/webllm/`: 6 capability + 7 rewrite-bullet)
- [x] `npm run build` clean; `@mlc-ai/web-llm` lives in its own chunk; `grep -r posthog dist/assets/` returns nothing with `VITE_POSTHOG_KEY` unset
- [x] Chrome (WebGPU available) — dropped a synthetic PDF, all 6 parsed bullets show "Suggest a rewrite" below their check pills
- [ ] Chrome — clicked rewrite → confirm progress UI, model download in DevTools Network, rewritten bullet inline, "copy to clipboard" works, second click on a different bullet skips the download
- [ ] Firefox — confirm the button is **absent** (not greyed) under every bullet
- [ ] PostHog Live Events show the four events firing in the documented order (only with `VITE_POSTHOG_KEY` set)

## Notes for review
- The narrow `WebLlmEngine` interface intentionally duplicates the shape of `MLCEngineInterface` rather than re-exporting it, so the type graph for non-WebLLM call-sites stays light and tests don't pull the library.
- Component placement is flat (`src/components/RewriteButton.tsx`) matching the existing `DropZone.tsx` / `PdfPreview.tsx` / `Result.tsx` — the `ui/` + `features/` tiers from CLAUDE.md don't exist yet on `main`. Happy to move once a primitive folder lands.

### DIFF
--- a/src/components/Result.tsx
+++ b/src/components/Result.tsx
@@ -18,7 +18,7 @@ import {
   scoreBandTextClass,
   scoreBandBgClass,
 } from "./features/scoreBand.ts";
-import { RewriteButton } from "./RewriteButton";
+import { RewriteButton } from "./features/RewriteButton.tsx";
 
 interface ResultProps {
   result: CascadeResult;
--- a/src/components/Result.tsx
+++ b/src/components/Result.tsx
@@ -18,6 +18,7 @@ import {
   scoreBandTextClass,
   scoreBandBgClass,
 } from "./features/scoreBand.ts";
+import { RewriteButton } from "./RewriteButton";
 
 interface ResultProps {
   result: CascadeResult;
@@ -428,6 +429,7 @@ function BulletRow({ bullet }: { bullet: BulletObservation }) {
           #{bullet.index + 1}
         </span>
         {bullet.text}
+        <RewriteButton bullet={bullet.text} />
       </td>
       <td className="py-1 pr-2 text-right align-top tabular-nums">
         {bullet.startsWithActionVerb ? (
--- a/src/components/RewriteButton.tsx
+++ b/src/components/RewriteButton.tsx
@@ -1,0 +1,199 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+/**
+ * Per-bullet "Suggest a rewrite" CTA. Runs Qwen2-1.5B in the browser via
+ * WebLLM (see ../lib/webllm/).
+ *
+ * Non-negotiable rules from issue #3:
+ *   - On browsers without WebGPU, this component returns null. Not a
+ *     greyed-out button, not a "your browser isn't supported" banner. Just
+ *     gone. Silent degradation is worse than absence — please don't "fix"
+ *     this with a fallback message.
+ *   - The model weights (~1.2GB) download on click only. Never on mount,
+ *     never on hover, never on scroll-into-view. The cold-start UX is the
+ *     conversion killer, not inference cost.
+ */
+
+import { useCallback, useEffect, useState } from "react";
+import { detectWebGpu } from "../lib/webllm/capability.ts";
+import { loadEngine } from "../lib/webllm/web-llm.ts";
+import { rewriteBulletWithLlm } from "../lib/webllm/rewrite-bullet.ts";
+import type {
+  ProgressUpdate,
+  WebGpuCapability,
+} from "../lib/webllm/types.ts";
+
+interface RewriteButtonProps {
+  bullet: string;
+}
+
+type Status =
+  | { kind: "idle" }
+  | { kind: "loading"; progress: ProgressUpdate }
+  | { kind: "rewriting" }
+  | { kind: "done"; rewritten: string }
+  | { kind: "error"; message: string };
+
+export function RewriteButton({ bullet }: RewriteButtonProps) {
+  const [capability, setCapability] = useState<WebGpuCapability | null>(null);
+  const [status, setStatus] = useState<Status>({ kind: "idle" });
+  const [copied, setCopied] = useState(false);
+
+  useEffect(() => {
+    let cancelled = false;
+    void detectWebGpu().then((c) => {
+      if (!cancelled) setCapability(c);
+    });
+    return () => {
+      cancelled = true;
+    };
+  }, []);
+
+  const onClick = useCallback(async () => {
+    setCopied(false);
+    try {
+      setStatus({
+        kind: "loading",
+        progress: { progress: 0, text: "Starting…" },
+      });
+      const engine = await loadEngine((progress) => {
+        setStatus({ kind: "loading", progress });
+      });
+      setStatus({ kind: "rewriting" });
+      const rewritten = await rewriteBulletWithLlm(bullet, engine);
+      setStatus({ kind: "done", rewritten });
+    } catch (err) {
+      setStatus({
+        kind: "error",
+        message:
+          err instanceof Error ? err.message : "Couldn't load the rewrite model",
+      });
+    }
+  }, [bullet]);
+
+  const onCopy = useCallback(async () => {
+    if (status.kind !== "done") return;
+    try {
+      await navigator.clipboard.writeText(status.rewritten);
+      setCopied(true);
+    } catch {
+      setCopied(false);
+    }
+  }, [status]);
+
+  if (capability !== "available") return null;
+
+  const busy = status.kind === "loading" || status.kind === "rewriting";
+
+  return (
+    <div className="mt-1 flex flex-col gap-1.5">
+      <button
+        type="button"
+        onClick={onClick}
+        disabled={busy}
+        className="self-start rounded-md border border-neutral-300 bg-white px-2 py-1 text-[11px] font-medium text-neutral-700 hover:border-neutral-400 hover:bg-neutral-50 disabled:cursor-not-allowed disabled:opacity-60 dark:border-neutral-700 dark:bg-neutral-900 dark:text-neutral-200 dark:hover:bg-neutral-800"
+      >
+        {labelFor(status)}
+      </button>
+
+      {status.kind === "loading" && (
+        <LoadingPanel progress={status.progress} />
+      )}
+
+      {status.kind === "rewriting" && (
+        <p className="text-[11px] text-neutral-500 dark:text-neutral-400">
+          Rewriting…
+        </p>
+      )}
+
+      {status.kind === "done" && (
+        <RewriteResult
+          rewritten={status.rewritten}
+          copied={copied}
+          onCopy={onCopy}
+        />
+      )}
+
+      {status.kind === "error" && (
+        <p className="text-[11px] text-red-700 dark:text-red-300">
+          {status.message}
+        </p>
+      )}
+    </div>
+  );
+}
+
+function labelFor(status: Status): string {
+  switch (status.kind) {
+    case "loading":
+      return "Loading model…";
+    case "rewriting":
+      return "Rewriting…";
+    case "done":
+      return "Suggest another rewrite";
+    case "error":
+      return "Try again";
+    default:
+      return "Suggest a rewrite";
+  }
+}
+
+function LoadingPanel({ progress }: { progress: ProgressUpdate }) {
+  const pct = Math.max(0, Math.min(100, Math.round(progress.progress * 100)));
+  return (
+    <div className="flex flex-col gap-1 rounded border border-neutral-200 bg-neutral-50 p-2 dark:border-neutral-800 dark:bg-neutral-950">
+      <div className="flex items-center justify-between text-[11px] text-neutral-700 dark:text-neutral-300">
+        <span>Loading the bullet-rewrite model (~1.2GB, one-time download)</span>
+        <span className="font-mono">{pct}%</span>
+      </div>
+      <div className="h-1.5 w-full overflow-hidden rounded-full bg-neutral-200 dark:bg-neutral-800">
+        <div
+          className="h-full bg-emerald-500 transition-all"
+          style={{ width: `${pct}%` }}
+        />
+      </div>
+      {progress.text && (
+        <p className="font-mono text-[10px] text-neutral-500 dark:text-neutral-500">
+          {progress.text}
+        </p>
+      )}
+      <details className="mt-1">
+        <summary className="cursor-pointer text-[10px] text-neutral-600 hover:underline dark:text-neutral-400">
+          What's happening?
+        </summary>
+        <p className="mt-1 max-w-prose text-[10px] leading-relaxed text-neutral-600 dark:text-neutral-400">
+          A small open-source language model (Qwen2-1.5B) is downloading to
+          your browser. It runs entirely on your device — your bullet text
+          never leaves this tab. The download takes about a minute on a
+          typical connection and is cached for next time.
+        </p>
+      </details>
+    </div>
+  );
+}
+
+function RewriteResult({
+  rewritten,
+  copied,
+  onCopy,
+}: {
+  rewritten: string;
+  copied: boolean;
+  onCopy: () => void;
+}) {
+  return (
+    <div className="flex flex-col gap-1 rounded border border-emerald-200 bg-emerald-50/60 p-2 dark:border-emerald-900/60 dark:bg-emerald-950/30">
+      <p className="text-xs leading-snug text-neutral-900 dark:text-neutral-100">
+        {rewritten}
+      </p>
+      <button
+        type="button"
+        onClick={onCopy}
+        className="self-start text-[10px] font-medium text-emerald-800 hover:underline dark:text-emerald-200"
+      >
+        {copied ? "Copied" : "Use this — copy to clipboard"}
+      </button>
+    </div>
+  );
+}
--- a/src/components/features/RewriteButton.tsx
+++ b/src/components/features/RewriteButton.tsx
@@ -3,7 +3,7 @@
 
 /**
  * Per-bullet "Suggest a rewrite" CTA. Runs Qwen2-1.5B in the browser via
- * WebLLM (see ../lib/webllm/).
+ * WebLLM (see ../../lib/webllm/).
  *
  * Non-negotiable rules from issue #3:
  *   - On browsers without WebGPU, this component returns null. Not a
@@ -13,16 +13,20 @@
  *   - The model weights (~1.2GB) download on click only. Never on mount,
  *     never on hover, never on scroll-into-view. The cold-start UX is the
  *     conversion killer, not inference cost.
+ *
+ * Styling uses semantic tokens only (surface / content / border / feedback)
+ * per CLAUDE.md. The token CSS auto-adapts via prefers-color-scheme, so no
+ * `dark:` overrides are needed.
  */
 
 import { useCallback, useEffect, useState } from "react";
-import { detectWebGpu } from "../lib/webllm/capability.ts";
-import { loadEngine } from "../lib/webllm/web-llm.ts";
-import { rewriteBulletWithLlm } from "../lib/webllm/rewrite-bullet.ts";
+import { detectWebGpu } from "../../lib/webllm/capability.ts";
+import { loadEngine } from "../../lib/webllm/web-llm.ts";
+import { rewriteBulletWithLlm } from "../../lib/webllm/rewrite-bullet.ts";
 import type {
   ProgressUpdate,
   WebGpuCapability,
-} from "../lib/webllm/types.ts";
+} from "../../lib/webllm/types.ts";
 
 interface RewriteButtonProps {
   bullet: string;
@@ -92,7 +96,7 @@ export function RewriteButton({ bullet }: RewriteButtonProps) {
         type="button"
         onClick={onClick}
         disabled={busy}
-        className="self-start rounded-md border border-neutral-300 bg-white px-2 py-1 text-[11px] font-medium text-neutral-700 hover:border-neutral-400 hover:bg-neutral-50 disabled:cursor-not-allowed disabled:opacity-60 dark:border-neutral-700 dark:bg-neutral-900 dark:text-neutral-200 dark:hover:bg-neutral-800"
+        className="self-start rounded-md border border-border-light bg-surface-card px-2 py-1 text-[11px] font-medium text-content-secondary hover:border-border hover:bg-surface-hover disabled:cursor-not-allowed disabled:opacity-60"
       >
         {labelFor(status)}
       </button>
@@ -102,9 +106,7 @@ export function RewriteButton({ bullet }: RewriteButtonProps) {
       )}
 
       {status.kind === "rewriting" && (
-        <p className="text-[11px] text-neutral-500 dark:text-neutral-400">
-          Rewriting…
-        </p>
+        <p className="text-[11px] text-content-muted">Rewriting…</p>
       )}
 
       {status.kind === "done" && (
@@ -116,9 +118,7 @@ export function RewriteButton({ bullet }: RewriteButtonProps) {
       )}
 
       {status.kind === "error" && (
-        <p className="text-[11px] text-red-700 dark:text-red-300">
-          {status.message}
-        </p>
+        <p className="text-[11px] text-feedback-error-text">{status.message}</p>
       )}
     </div>
   );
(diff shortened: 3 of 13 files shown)


src/lib/webllm/web-llm.ts
```diff
@@ -0,0 +1,60 @@
...
+ */
+export function loadEngine(
+  onProgress: (update: ProgressUpdate) => void,
+): Promise<WebLlmEngine> {
+  if (cached) return cached;
+  if (!downloadStartedFired) {
+    downloadStartedFired = true;
+    trackWebllmDownloadStarted();
+  }
+  cached = (async () => {
```
s-annam: [Blocking]: If the engine load rejects (OOM, dropped network), `cached` permanently holds a rejected promise — every later click returns it via `if (cached) return cached` and re-rejects instantly. The "Try again" button (error branch of `labelFor`) then can't recover; only a reload does. Reset the cache on failure:
```ts
cached = (async () => { ... })();
cached.catch(() => { cached = null; }); // allow a real retry
```
This path is untested too — worth a `web-llm.ts` test for the reject-then-retry case.

src/components/RewriteButton.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+  const busy = status.kind === "loading" || status.kind === "rewriting";
+
+  return (
+    <div className="mt-1 flex flex-col gap-1.5">
+      <button
+        type="button"
+        onClick={onClick}
+        disabled={busy}
+        className="self-start rounded-md border border-neutral-300 bg-white px-2 py-1 text-[11px] font-medium text-neutral-700 hover:border-neutral-400 hover:bg-neutral-50 disabled:cursor-not-allowed disabled:opacity-60 dark:border-neutral-700 dark:bg-neutral-900 dark:text-neutral-200 dark:hover:bg-neutral-800"
```
s-annam: [Blocking]: Hardcoded raw Tailwind palette here (`border-neutral-300 bg-white text-neutral-700 …`, plus `text-red-700` L119 and `border-emerald-200 bg-emerald-50` L186/193). CLAUDE.md lists this under **"What NOT to do → ❌ Hardcoded colors"** for feature code — it's an enforced rule, not a preference. Use semantic tokens: error → `feedback-error-*`, result box → `feedback-success-{bg,border,text}`, all already defined and auto-adapting to dark via `@media (prefers-color-scheme)` in `styles.css` (so the `dark:` pairs become dead weight). `Chip`/`FeedbackControl`/`ContactCard` are the pattern.

Relatedly (non-blocking): the raw `<button className>` trips CLAUDE.md's "use the `<Button>` primitive" rule — but no `Button` primitive exists yet (only `Chip`). Fine to keep the raw element for this PR; flagging so we don't multiply inline button styling before a primitive lands.

src/components/Result.tsx
```diff
@@ -13,6 +13,7 @@ import { ScoreRing } from "./features/ScoreRing.tsx";
 import { VerdictHeader } from "./features/VerdictHeader.tsx";
 import { ContactCard } from "./features/ContactCard.tsx";
 import { FeedbackControl } from "./features/FeedbackControl.tsx";
+import { RewriteButton } from "./RewriteButton";
```
s-annam: [Blocking]: `RewriteButton` is wired to bullet domain data — a Feature component by CLAUDE.md's 3-tier definition — so it belongs in `src/components/features/`, alongside the siblings imported just above (ScoreRing, VerdictHeader, ContactCard, FeedbackControl). This is the enforced architecture, and the PR description's justification ("the `ui/` + `features/` tiers don't exist yet on main") is factually wrong — `features/` is right here and you're importing from it. Move to `features/RewriteButton.tsx`. (At 199 LOC it's also right at the ~200 split threshold — worth keeping an eye on.)

src/lib/webllm/rewrite-bullet.ts
```diff
@@ -0,0 +1,73 @@
...
+    ],
+    temperature: 0.3,
+    max_tokens: 120,
+  });
+  const raw = response.choices[0]?.message?.content ?? "";
+  const cleaned = postProcess(raw);
+  if (!firstRewriteFired) {
+    firstRewriteFired = true;
+    trackWebllmFirstRewrite();
+  }
```
s-annam, line 59: [Nit]: `trackWebllmFirstRewrite()` fires before checking whether `cleaned` is non-empty, so a null-content / empty rewrite still counts as a "first rewrite". Consider gating the event on a non-empty result.

s-annam (changes requested):
## PR Review: feat(webllm) per-bullet "Suggest a rewrite" pilot

> **Updated** — reclassified the hardcoded-color and component-placement findings from Suggestion to **Blocking**. Both are enforced rules in CLAUDE.md (the styling "What NOT to do" list and the 3-tier component architecture), not stylistic preferences.

### Summary
Clean, well-tested WebLLM pilot. Capability detection, lazy engine load, the pure rewrite function, and telemetry all match issue #3's spec. Verified locally on the branch: **207 tests pass** (13 new), typecheck clean, build code-splits `@mlc-ai/web-llm` into its own ~6 MB chunk (`index-CereMTgS.js`) with the entry chunk barely moving, and no `posthog` string in `dist/` when `VITE_POSTHOG_KEY` is unset. Three blocking items to address before merge: a broken retry path and two CLAUDE.md rule violations on the new component.

### Spec Alignment (issue #3)
|---|---|---|
| WebGPU detect, cache, 3-state | ✅ Implemented | `capability.ts`, cached promise, telemetry fires once |
| Pinned Qwen2-1.5B `MODEL_ID` | ✅ Implemented | single const, no inline string |
| Lazy load on click, own chunk | ✅ Verified | `index-CereMTgS.js` 6 MB; entry +5 KB |
| Hide CTA (not grey) w/o WebGPU | ✅ Implemented | returns `null`; header documents the stance |
| 4 snake_case events, env-gated | ✅ Verified | no `posthog` in dist without key |
| Tests pass | ✅ Verified | 207 (6 capability + 7 rewrite-bullet new) |

### Highlights
- Narrow `WebLlmEngine` contract so tests stub without importing the 6 MB lib — right call, keeps the type graph and test deps light.
- `BULLET_REWRITE_SYSTEM_PROMPT` as single source of truth, paraphrased in the user-facing "What's happening?" disclosure. Nice honesty.
- Capability + telemetry caching via module-level promise/flags is correct and well-commented.

### Key Findings

**1. [Blocking] Failed engine load caches a rejected promise — "Try again" can never recover.** `web-llm.ts:41`. `cached` is assigned the async IIFE with no `.catch`. If `CreateMLCEngine` throws (OOM, network drop), `cached` holds a *rejected* promise for the page lifetime; every later click hits `if (cached) return cached` and re-rejects instantly. The UI's "Try again" button (the `error` branch of `labelFor`) is therefore non-functional — only a page reload recovers. Reset the cache on failure so retry can re-attempt. This path is also untested (`web-llm.ts` has no test file).

**2. [Blocking] Hardcoded raw Tailwind palette in feature code.** `RewriteButton.tsx` (L95, L119, L186, L193). `border-neutral-300 bg-white text-red-700 border-emerald-200 bg-emerald-50` etc. CLAUDE.md lists this under **"What NOT to do → ❌ Hardcoded colors"** for feature components — enforced, not advisory. The token system already ships `feedback-success-{bg,border,text}` and `feedback-error-*` purpose-built for the green-result and red-error boxes here, and they auto-adapt to dark mode via `@media (prefers-color-scheme)` in `styles.css` — so the manual `dark:` pairs are redundant parallel theming. `Chip`/`FeedbackControl`/`ContactCard` are the pattern to follow.

**3. [Blocking] `RewriteButton` placed in the wrong architecture tier.** It's wired to bullet domain data, the CLAUDE.md definition of a Feature component, so it belongs in `src/components/features/` next to the siblings `Result.tsx` already imports (ScoreRing, VerdictHeader, ContactCard, FeedbackControl). The 3-tier architecture is enforced, and the PR description's justification — "the `ui/` + `features/` tiers don't exist yet on main" — is factually wrong; `features/` is right there. Move to `features/RewriteButton.tsx`.

**4. [Suggestion] Raw `<button>` vs the `<Button>` primitive.** CLAUDE.md says use the primitive — but none exists yet (only `Chip` in `ui/`). Keeping the raw element is fine for this PR; flagging so inline button styles don't proliferate before a `Button` primitive lands. Pairs with finding 2.

**5. [Nit] Entry chunk +5.01 KB** (559.46 → 564.47), a hair over the PR's "≤5 KB" claim. Trivial.

**6. [Nit] `webllm_first_rewrite` fires even on empty output** (`rewrite-bullet.ts`) when the model returns null content and `postProcess` yields `""`. Also `copied` never resets to false. Both minor.

### Verdict
**Action:** REQUEST_CHANGES
**Rationale:** One blocking correctness bug (advertised retry can't recover) plus two blocking CLAUDE.md rule violations on the new component (hardcoded palette, wrong tier). All three are cheap to fix and avoid baking drift into the first AI-capability component. Spec coverage, tests, bundle split, and telemetry gating are all in great shape.

Vaishnavi1709:
Thanks for the careful review @s-annam — all three blocking items + the empty-output nit are addressed. Rebased onto latest `main` (which had moved 4 commits, including the new `<tr>`/`<td>` BulletRow). PR is now MERGEABLE.

**1. Engine retry path** — `web-llm.ts` now clears `cached` on rejection so the UI's "Try again" branch can re-attempt. `downloadStartedFired` deliberately stays `true` after a failure so a retry of the same logical attempt doesn't double-fire `webllm_download_started`. New `web-llm.test.ts` mocks `@mlc-ai/web-llm` via a hoisted `vi.mock` and covers:
- reject-then-retry (the bug you flagged)
- concurrent-call dedupe (three parallel `loadEngine` calls share one `CreateMLCEngine` invocation)
- `MODEL_ID` is the value passed to the library
- `initProgressCallback` reports forward to the supplied `onProgress`

**2. Hardcoded palette** — every raw color in `RewriteButton.tsx` swapped for a semantic token (`border-border-light`, `bg-surface-card`, `bg-surface-subtle`, `text-content-{secondary,muted,tertiary,primary}`, `bg-feedback-success-bg`, `border-feedback-success-border`, `text-feedback-success-text`, `bg-feedback-success-icon` for the progress-bar fill, `text-feedback-error-text` for the error row). All `dark:` overrides removed — the token CSS auto-adapts via `prefers-color-scheme`. Pattern follows `Chip` / `FeedbackControl` / `ContactCard`.

**3. Component tier** — `git mv` to `src/components/features/RewriteButton.tsx`, imports rewritten as `../../lib/webllm/…`, `Result.tsx` now imports from `./features/RewriteButton.tsx`. You're right that the PR description's justification was factually wrong — `features/` was already there in the same Result.tsx import block, I should have looked harder.

**4. Empty-output telemetry** — `rewrite-bullet.ts` now gates `trackWebllmFirstRewrite()` on `cleaned.length > 0` so a null model response isn't counted as a funnel step. Two new mock-based tests assert no-fire on empty and fire-once on first non-empty output (uses `vi.hoisted` so the mock factory can reference the spy).

One nit I left as-is: **`copied` resetting to false.** Re-read the code and `setCopied(false)` is the first line of `onClick`, so it does reset on every new rewrite attempt. Post-copy state stays "Copied" until the next click — I read that as intended UX (the user has confirmed the action). Happy to add a timeout-based reset if you'd rather see it auto-revert.

### Rebase note

Main's new table-row BulletRow drops `CheckPill` for inline `<td>` cells. Conflict resolved by taking main's structure and placing `<RewriteButton bullet={bullet.text} />` inside the bullet-text `<td>` so the button sits under the bullet text in column 1 — the same visual relationship as before, in the new layout.

### Verification

- `npm run typecheck` clean
- `npm run test` — 213/213 pass (was 207; +4 `web-llm.test.ts`, +2 telemetry-gating in `rewrite-bullet.test.ts`)
- `npm run build` clean; `@mlc-ai/web-llm` still in its own ~6 MB chunk; `posthog` still absent from `dist/` with `VITE_POSTHOG_KEY` unset
- Manual click-flow + Firefox-hide verification still outstanding on my side; happy to take a follow-up review pass after that

Ready for re-review whenever you have a minute.

src/components/features/RewriteButton.tsx
```diff
@@ -0,0 +1,197 @@
...
+    case "loading":
+      return "Loading model…";
+    case "rewriting":
+      return "Rewriting…";
+    case "done":
+      return "Suggest another rewrite";
+    case "error":
+      return "Try again";
+    default:
+      return "Suggest a rewrite";
```
s-annam, line 140: [Nit]: This progress bar conveys state purely visually. Consider `role="progressbar"` with `aria-valuenow={pct}` / `aria-valuemin={0}` / `aria-valuemax={100}` on the track div so screen-reader users get the ~1-minute download progress, not silence.

src/components/features/RewriteButton.tsx
```diff
@@ -0,0 +1,197 @@
...
+      >
+        {labelFor(status)}
+      </button>
+
+      {status.kind === "loading" && (
+        <LoadingPanel progress={status.progress} />
+      )}
+
+      {status.kind === "rewriting" && (
+        <p className="text-[11px] text-content-muted">Rewriting…</p>
```
s-annam, line 109: [Nit]: The error message isn't announced to assistive tech. `role="alert"` (or `aria-live="polite"`) on this `<p>` would surface a failed load to SR users without a visual cue.

s-annam (dismissed):
## PR Review: feat(webllm) per-bullet "Suggest a rewrite" pilot — re-review

Follow-up commit `a70a642` resolves all three blocking items from the prior review, plus both nits. Verified locally on the PR head.

### Prior blocking items — all fixed
| Item | Status |
|---|---|
| **#1** Rejected-promise cache made "Try again" a no-op (`web-llm.ts`) | ✅ `pending.catch(() => { if (cached === pending) cached = null })` — identity-guarded so a slow failure can't null a later success's cache. New `web-llm.test.ts` covers reject-then-retry (re-invokes `CreateMLCEngine`, resolves). |
| **#2** Hardcoded Tailwind palette | ✅ All semantic tokens now (`feedback-success-{bg,border,text}`, `feedback-error-text`, `surface/content/border-*`). No raw palette, no redundant `dark:` pairs. |
| **#3** Wrong architecture tier | ✅ Moved to `src/components/features/RewriteButton.tsx`; `Result.tsx` imports `./features/RewriteButton.tsx`. No dangling old-path refs. |
| Nit: `webllm_first_rewrite` fired on empty output | ✅ Gated on `cleaned.length > 0`; test asserts no-fire on null content. |
| Nit: `copied` never reset | ✅ `setCopied(false)` at click start. |

### Local verification (PR head `a70a642`)
- `npm test` → **213 passed** (19 webllm: 6 capability + 9 rewrite-bullet + 4 web-llm)
- `npm run typecheck` → clean
- `npm run build` → `@mlc-ai/web-llm` code-split into its own ~6 MB chunk (`index-CereMTgS.js`); entry chunk 566 kB
- No `posthog` string in `dist/` with `VITE_POSTHOG_KEY` unset → env-gating + DCE hold

### Highlights
- Telemetry one-shot flags survive a retry by design (`downloadStartedFired` stays true) so a single logical attempt doesn't double-fire `webllm_download_started` — and that reasoning is in the doc comment. Good.
- Narrow `WebLlmEngine` contract keeps the 6 MB lib out of the test graph; concurrent-click sharing of one engine/one download is tested.

### Remaining — non-blocking nits only
1. **[Nit]** `LoadingPanel` progress bar has no `role="progressbar"` / `aria-valuenow` — screen-reader users get no load progress.
2. **[Nit]** Error `<p>` isn't announced (`role="alert"`).
3. **[Nit]** Component is 197 LOC — right at the ~200 split threshold; worth watching as the rewrite UX grows.

### Verdict
**Action:** APPROVE
**Rationale:** Zero blocking. The retry-recovery bug is fixed and now covered by a regression test, palette is token-clean, the component sits in the correct tier, and telemetry gating is sane. Tests green, typecheck clean, bundle split confirmed. Nits are optional polish.

s-annam (approved):
Re-approve after rebase onto main + a11y nits (progressbar/alert roles). Conflicts resolved, verify CI green, 243 tests pass.

Issue #71, opened 2026-06-13
Epic: Phone number handling across resumelint


s-annam:
## Epic — Phone number handling across resumelint

Track all work to make phone-number recognition, validation, and formatting correct and locale-aware throughout the project. Today phones are matched by a US-centric 10-digit regex (`PHONE_RE`, `src/lib/heuristics/regex.ts:19`), stored **verbatim**, displayed unformatted, and scored on **presence only**. This epic replaces that with a real phone-parsing library, location-derived locale context, and (optionally) validity-aware scoring.

### Why

- US-centric regex misses/mangles international and national-format numbers.
- No normalization → we echo back whatever messy substring the résumé contained.
- The candidate's region is already extracted (`location`) but unused for phone parsing.

### Direction

Adopt **`libphonenumber-js`** (MIT, ESM, tree-shakeable, lazy-loaded) — `findPhoneNumbersInText` for recall, `parsePhoneNumber` for validation + reformatting. Feed the extracted location as region context. Reformat to clean national/international form for display.

### Decisions (agreed)

- **Split** the work A/B/C (below) for reviewable, intern-sized PRs.
- **Default region = US** when no location is extracted or the country is unmapped.
- **Scoring stays presence-only** in the core work; validity-aware scoring is deferred to its own issue (carries `ATS_SCORE_ALGO_VERSION` bump + snapshot re-bake).

### Child issues

- [ ] #68 — Adopt `libphonenumber-js`: recognize, validate, reformat (core; default US region).
- [ ] #69 — Derive phone region from extracted location (locale-aware). *Blocked by #68.*
- [ ] #70 — Validity-aware phone scoring (deferred; score-version bump). *Blocked by #68.*
- [ ] #62 — Two-column fixture 7-digit phone → 10-digit synthetic fixture fix. *Independent fixture artifact; `PHONE_RE` stays 10-digit even after this epic.*

### Dependency order

#68 → (#69, #70). #62 is independent and can land anytime.

Issue #60, opened 2026-06-13
Two-column resumes: sidebar interleaves into experience — detection not wired to text assembly


s-annam:
## Summary

Two-column resumes parse with their text scrambled: the right-hand sidebar (skills, strengths, achievements) interleaves into the left-hand experience column, destroying role boundaries. The layout probe **detects** two-column (`two_column` trigger fires), but that knowledge is never used to fix the reading order — so the cascade still reads the page as a single interleaved stream.

Two committed fixtures demonstrate it (both score far below what their content warrants):

| Fixture | `two_column` | `experienceCount` | `bulletCount` | `skillsCount` | overall |
|---|---|---|---|---|---|
| `unknown/weasyprint-cairo-two-column.pdf` | ✅ | 1 | 21 | 22 | 43 |
| `unknown/chromium-two-column-sidebar.pdf` (new, 8 roles) | ✅ | **1** | 27 | **74** | 46 |

`experienceCount: 1` against 8 real roles, and `skillsCount: 74` against ~17 real skills, are the signature: every role collapses into one, and interleaved experience text gets misparsed as skills.

## Root cause

The probe and the text-assembly path are disconnected.

**Detection works.** `probeTwoColumn` (`src/lib/heuristics/pdf-layout.ts`) bins item x-starts, finds two dominant peaks separated by a gap, and sets `isTwoColumn` → pushes the `two_column` trigger.

**But that result is wired only to scoring/escalation, never to assembly:**
- `confidence.ts:102` — caps confidence
- `confidence.ts:293` — escalation hint → `"ner"`

`assembleTextFromLines` → `groupIntoLines` (`src/lib/heuristics/pdf-extract.ts`, `sections.ts`) is **column-blind**. It has only a *local* per-row gap split (`COLUMN_GAP_THRESHOLD = 50pt`, issue #9) — no page-level column model. Two concrete failure modes result:

**1. Row-interleaving (reading order).** `groupIntoLines` sorts items globally by `(page, y, x)`. A left-column row and a right-column row sharing a y-band are emitted adjacently and joined with `\n`, producing `L1, R1, L2, R2, …` instead of *all of left, then all of right*. The sidebar's section headers and skill tokens bleed into the middle of the experience block. Observed in extraction:
- `STRENGTHS` / `Leadership` injected between the `EXPERIENCE` header and the first role
[...]

PR #100, opened 2026-06-16
feat(phone): locale-aware region derivation + fix two-column fixture phone

### LINKED ISSUE #62: Two-column fixture phone (555-0142) not detected — 7-digit local format below PHONE_RE 10-digit floor
## Problem

The two-column sidebar fixture's phone number is present in the extracted plain text (`555-0142`) but the parser reports **"Phone not detected"** in the Contact card (3 of 5 detected). This drags Completeness scoring (the score's "Missing: phone, LinkedIn" line).

Surfaced during Playwright QA of the #60 two-column reading-order fix.

### Observed (live UI, `chromium-two-column-sidebar.pdf`)

- Extracted plain text contains: `Jamie Rivera Software Engineering Leader 555-0142 jamie.rivera@example.com …`
- Contact card shows **Phone — not detected** (⚠).

## Root cause

This is a **fixture artifact**, not a general parser gap. `PHONE_RE` in `src/lib/heuristics/regex.ts:19`:

```ts
export const PHONE_RE =
  /(?:(?:\+?\d{1,3})[\s.-]?)?\(?\d{3}\)?[\s.-]?\d{3}[\s.-]?\d{4}\b/g;
```

requires a full **10-digit** number (`\d{3}` area + `\d{3}` + `\d{4}`, optional country code). The fixture's `555-0142` is only **7 digits** (`\d{3}-\d{4}`) → no match. A real US résumé phone is 10 digits, so the regex is correct for production; the fixture simply carries an unmatchable 7-digit local number.

Broadening `PHONE_RE` to accept bare 7-digit runs is **not** advised — it would false-match zip+4, ID numbers, and stray `NNN-NNNN` runs (e.g. a `1985 - 1989`-style range with a dash).

## Proposed fix

**Primary — fix the fixture to a realistic 10-digit synthetic number.** Re-export `chromium-two-column-sidebar.pdf` with a full synthetic `555` phone that a real résumé would use and that `PHONE_RE` matches, e.g.:

- `(512) 555-0142`, or
- `512-555-0142`, or
- `555-555-0142`

(`555-01XX` is the reserved fictional block — keep the `555-0142` suffix, prepend a synthetic area+exchange so it's a valid 10-digit shape. Stays PII-free.)

Then re-bake the snapshot — Contact becomes 4/5 and Completeness improves accordingly.

**Decision to confirm:** should a 7-digit local number ever be surfaced? Recommendation: **no** — keep `PHONE_RE` 10-digit-only (production-correct, low false-positive), and fix the fixture. Document this in the fixture README so future synthetic personas use 10-digit numbers.

## Affected files

- `tests/fixtures/pdfs/unknown/chromium-two-column-sidebar.pdf` — re-export with a 10-digit synthetic phone.
- `tests/fixtures/pdfs/unknown/chromium-two-column-sidebar.expected.json` — re-bake (`npm run bake-fixtures`); Contact field count + Completeness shift.
- `tests/fixtures/pdfs/README.md` — add a one-line note: synthetic personas use full 10-digit `555` phone numbers so `PHONE_RE` matches.
- `src/lib/heuristics/regex.ts:19` — no change (documented as intentional 10-digit floor).

## Acceptance criteria

- [ ] `chromium-two-column-sidebar.pdf` re-exported with a 10-digit synthetic `555` phone; `pdftotext` confirms the new number and that name/email stay synthetic (Jamie Rivera / `@example.com`).
- [ ] Live parse surfaces the phone — Contact card shows 4 of 5 detected; `phone` no longer in the score's `missing` list for this fixture.
- [ ] `chromium-two-column-sidebar.expected.json` re-baked and committed alongside the PDF.
- [ ] Fixture README documents the 10-digit synthetic-phone convention.
- [ ] `PHONE_RE` left at 10-digit (no broadening); `npm run test` green.

## Notes

- Follow-up to #60. Low priority — cosmetic for one fixture, no production-code change.
- PII policy: verify the re-exported binary with `pdftotext tests/fixtures/pdfs/unknown/chromium-two-column-sidebar.pdf - | head -40` before commit (binary is the exposure surface, not the PR prose).

### LINKED ISSUE #69: Derive phone region from extracted location (locale-aware parsing)
## Problem

After Issue A, phone parsing uses a **fixed US default region**. But the résumé already tells us the candidate's region — `extractContact` extracts `location` in the *same scan* (`src/lib/heuristics/extract-fields.ts:268`, via `US_LOCATION_RE` / `INTL_LOCATION_RE`) — and we ignore it. A London or Bangalore résumé with a national-format number gets parsed as if US.

Use the extracted location as the region hint so national/local-format numbers parse and reformat correctly for the candidate's actual locale.

## Approach

1. Add a small `regionFromLocation(location: string): string | undefined` helper (likely in `phone.ts` or a sibling): map `US_LOCATION_RE` matches → `US`; map the country tail of `INTL_LOCATION_RE` → ISO 3166-1 alpha-2 (e.g. `United Kingdom`→`GB`, `India`→`IN`, `Canada`→`CA`). Keep the country→ISO table small and explicit; unmatched → undefined.
2. Thread the derived region into the `findPhoneNumbersInText` / `normalizePhone` call from Issue A as `defaultCountry`.
3. **Fallback to `US`** when no location is extracted or the country is unmapped (matches the Issue A default — confirmed decision).

## Depends on

Blocked by Issue A (the normalizer + `findPhoneNumbersInText` wiring must land first).

## Affected files

- `src/lib/heuristics/phone.ts` (or sibling) — add `regionFromLocation` + country→ISO table.
- `src/lib/heuristics/extract-fields.ts` — pass derived region instead of hardcoded `US`.
- Tests — `regionFromLocation` unit tests (US + 2–3 intl); a fixture demonstrating an intl national-format number reformatting correctly.

## Acceptance criteria

- [ ] `regionFromLocation` maps US + a small explicit set of intl countries to ISO codes; unmatched → undefined.
- [ ] Phone parsing uses the location-derived region; falls back to `US` when absent/unmapped.
- [ ] An international-locale fixture (national-format number) reformats to the correct national/international form.
- [ ] `npm run test` + `npm run typecheck` green.

## Notes

Part of the phone-handling epic. Builds on Issue A.

s-annam:
## Summary

Phone-handling epic (#71), two children landed together on one branch:

- **#69 — locale-aware region derivation.** Phone parsing no longer assumes a hardcoded US default. New `regionFromLocation()` (US + a small explicit country→ISO table) in `phone.ts` derives the region from the résumé's already-extracted `location`, threaded into `findFirstPhone()` at the `extract-fields` call site (location is now resolved *before* the phone scan). The `mightHavePhone` pre-filter is relaxed for non-US regions so national-format intl numbers reach libphonenumber. Falls back to `US` when location is absent or unmapped.
- **#62 — fix two-column fixture phone.** The `chromium-two-column-sidebar` fixture carried a 7-digit `555-0142`, unparseable by libphonenumber (a bad-fixture artifact, **not** a parser gap — we deliberately do **not** broaden to 7-digit). Replaced with a valid 10-digit synthetic `(512) 555-0142` and re-baked the snapshot: phone now detected, completeness 24→27, `two_column` trigger preserved.

### Note on #62 snapshot churn
In-place phone patching wasn't viable (CID-encoded fonts — no ASCII bytes to swap), so the PDF was regenerated from reconstructed HTML. The `skillsCount` (5→10), `experienceCount` (7→5), and `sectionSource` (markdown→regex) deltas are layout-detection artifacts of regeneration, **not** parser regressions. Two-column detection and the fixture's #60 purpose are intact. No production-code change in #62.

Resolves #69
Resolves #62

## Test plan

- [x] `npm run test` green (407/407)
- [x] Fixture persona verified synthetic — `Jamie Rivera` / `@example.com` / `(512) 555-0142`, no real PII (Step 3.5)


src/lib/heuristics/phone.ts
```diff
@@ -25,7 +26,107 @@ import {
...
+  location: string | undefined,
+): CountryCode | undefined {
+  if (!location) return undefined;
+
+  // Try the INTL table first — covers both full country names ("United Kingdom")
+  // and known 2-letter abbreviations like "UK" that would otherwise be caught
+  // by US_LOCATION_RE (which matches any 2-uppercase-letter token after a comma).
+  const intlMatch = INTL_LOCATION_RE.exec(location);
+  if (intlMatch) {
+    const countryTail = intlMatch[2].trim().toLowerCase();
```
rohithgollapalli, line 121: `regionFromLocation` only inspects the segment immediately after the *first* comma. `"Bengaluru, India"` maps to `IN`, but `"Bengaluru, Karnataka, India"` (City, State, Country) yields no map hit, so the caller falls back to `US` and an Indian national-format number won't parse. No regression (it degrades to current US-default behavior) and `City, Country` is the common form, but worth a one-line comment noting the country must be the segment right after the first comma - it sets expectations for the explicit table.

src/lib/heuristics/phone.ts
```diff
@@ -69,27 +170,39 @@ export function normalizePhone(
...
- * Pre-filter check: returns true if `text` might contain a phone number,
- * using two fast checks before invoking the heavier libphonenumber parser:
+ * Pre-filter check: returns true if `text` might contain a phone number.
+ *
+ * Two fast checks are applied before invoking the heavier libphonenumber parser:
  *   1. `PHONE_RE` — catches US/CA 10-digit shapes and common variants.
  *   2. `/\+\d/` — catches E.164 international numbers (`+44 …`, `+1 …`)
  *      whose space-separated groups fall outside PHONE_RE's US-biased pattern.
+ *
+ * For non-US regions the pre-filter is relaxed to any 6+ digit sequence,
```
rohithgollapalli: Doc/impl mismatch: this says the non-US filter is relaxed to "any **6+ digit sequence**," but the code (line 195) is `text.replace(/\D/g, "").length >= 7` - that's *7+ total digits across the string*, not a contiguous sequence (the inline comment there even argues against a sequence check). Suggest tightening this JSDoc to match: "7+ total digits across the string."

src/lib/heuristics/phone.ts
```diff
@@ -69,27 +170,39 @@ export function normalizePhone(
...
   PHONE_RE.lastIndex = 0;
   if (byUs) return true;
-  return /\+\d/.test(text);
+  if (/\+\d/.test(text)) return true;
+  // For non-US regions, accept text containing 7+ total digits as a candidate
+  // to pass to libphonenumber. National-format numbers (e.g. UK "020 7946 0958",
+  // India "098765 43210") are space-separated so no single run of 6+ digits
+  // exists; counting all digits is the reliable pre-filter. The heavier
+  // libphonenumber parser is the authoritative validity gate.
+  if (region !== "US") return (text.replace(/\D/g, "").length >= 7);
```
rohithgollapalli, line 198: Tier-1.5 asymmetry: this relaxed pre-filter only loosens for `region !== "US"`, but `regex-fallback.ts:77` still calls `findFirstPhone(rawText)` with the default `US` region. So for an intl resume where tier-1 misses the phone, the tier-1.5 recovery path benefits from neither region derivation nor this relaxation - an intl national-format number recoverable only at tier-1.5 would still drop. The common case (tier-1 finds it) is covered, so this is a deferral not a bug, but worth an explicit note/follow-up issue since the PR title implies full-pipeline locale-awareness.

rohithgollapalli (commented):
A few non-blocking notes from review. Overall this is clean and well-tested - fixture persona verified synthetic, two_column trigger and markdown coverage both intact. Approving direction; just some follow-up nits inline.

src/lib/heuristics/phone.ts
s-annam: Good catch. Added a JSDoc note to `regionFromLocation` clarifying that three-part strings ("City, State, Country") don't map — INTL_LOCATION_RE captures only the first comma-segment as the country tail, so the country must sit immediately after the first comma. Fixed in 9e59a01.

src/lib/heuristics/phone.ts
s-annam: You're right — doc/impl mismatch. Updated the JSDoc to "7+ total digits across the string" to match the `>= 7` check and the inline comment's own wording. Fixed in 9e59a01.

src/lib/heuristics/phone.ts
s-annam: Fixed rather than deferred. `regex-fallback.ts` already imports from `phone.ts`, so I imported `regionFromLocation` there too and derive region from `parsed.location` (set by tier-1) before the `findFirstPhone` call. No cascade.ts changes needed. Fixed in 9e59a01.

PR #102, opened 2026-06-16
fix(jd-match): decode numeric HTML entity references in htmlToPlaintext

### LINKED ISSUE #93: htmlToPlaintext misses numeric HTML entity references
## Context

  `src/lib/jd-match/fetch-jd.ts:177-184` decodes a fixed set of named entities (`&amp;`, `&lt;`, `&gt;`, `&quot;`, `&#39;`, `&nbsp;`) on its way from the ATS API HTML body to the matched plaintext.
  Numeric character references — `&#160;` (decimal nbsp), `&#x2013;` (hex en-dash), `&#8217;` (decimal curly apostrophe), etc. — pass through unchanged.

  Real JDs from various generators ship numeric entities (Lever in particular leans on them for typographic punctuation). When they survive the strip, they reach the JD-match passes as raw `&#…;` strings:
   the skill regex doesn't see "you'll" because the text says "you&#8217;ll", and the noun-pass acronym regex starts matching `#` fragments.

  ## Reproduce

  ```ts
  import { htmlToPlaintext } from "./src/lib/jd-match/fetch-jd.ts";
  htmlToPlaintext("<p>Use&nbsp;Kubernetes&#160;daily.</p>");
  // Actual:   "Use Kubernetes&#160;daily."
  // Expected: "Use Kubernetes daily."

  Suggested fix
  
  After the named-entity decode loop, run two more replacements:

  text = text
    .replace(/&#(\d+);/g, (_, n) => String.fromCodePoint(parseInt(n, 10)))
    .replace(/&#x([0-9a-f]+);/gi, (_, h) => String.fromCodePoint(parseInt(h, 16)));

  Optional but cheap: clamp to the BMP and skip values that don't decode to valid Unicode scalars.

  Acceptance criteria
  
  - Numeric decimal + hex entity references decode to their characters
  - New tests in src/lib/jd-match/fetch-jd.test.ts cover the common cases (&#160;, &#x2013;, &#8217;) and a malformed reference (e.g. &#x;) that should pass through unchanged
  - No regressions in the existing htmlToPlaintext test suite

rohithgollapalli:
## Summary

`htmlToPlaintext` (`src/lib/jd-match/fetch-jd.ts`) decoded only a fixed set of **named** entities, so **numeric** character references — decimal `&#160;`, hex `&#x2013;`, `&#8217;`, etc. — survived the strip and leaked raw `&#…;` fragments into the JD-match passes. The skill regex missed `you&#8217;ll` (vs `you’ll`) and the noun-pass acronym regex began matching `#` fragments. Generators such as Lever lean on numeric entities for typographic punctuation.

### Changes
- Add a decimal (`&#(\d+);`) + hex (`&#x([0-9a-f]+);`) numeric-reference decode pass after the named-entity loop.
- Route both through a new `decodeCodePoint` helper that:
  - leaves out-of-range / lone-surrogate values as the original `&#…;` text rather than throwing or emitting `U+FFFD`;
  - folds `0xA0` (nbsp) to a regular space so numeric `&#160;` / `&#xA0;` match the existing named `&nbsp;` decode path.
- Malformed refs like `&#x;` never match (the digit group requires 1+) and pass through unchanged.

### Note on nbsp
The issue's reproduce case expects `&#160;` → a plain space, but a faithful decode yields `U+00A0`. The existing named `&nbsp;` path already normalizes to a regular space, so nbsp is folded to a space inside `decodeCodePoint` for consistency — this also keeps the source ASCII-clean (no invisible nbsp bytes).

## Test plan

- [x] `npm run test` green (392/392; 16/16 in `fetch-jd.test.ts`)
- New tests in `fetch-jd.test.ts` cover decimal `&#160;`, hex `&#x2013;`, curly apostrophe `&#8217;`, a malformed `&#x;` (unchanged), and an out-of-range `&#1114112;` (unchanged, must not throw).
- No regressions in the existing `htmlToPlaintext` suite.

Closes #93


src/lib/jd-match/fetch-jd.ts
```diff
@@ -161,6 +161,28 @@ async function fetchRecruiteeJob(
...
+function decodeCodePoint(original: string, codePoint: number): string {
+  if (
+    !Number.isInteger(codePoint) ||
+    codePoint < 0 ||
+    codePoint > 0x10ffff ||
+    (codePoint >= 0xd800 && codePoint <= 0xdfff)
+  ) {
+    return original;
+  }
+  if (codePoint === 0xa0) return " ";
```
s-annam, line 187: [Suggestion]: Control code points decode through unfiltered — `&#0;`→U+0000, `&#13;`→CR, `&#8;`→backspace all pass this guard (cp ≥ 0, non-surrogate) and inject control chars into the JD-match plaintext. Real-world rare for ATS HTML, but a C0-range skip (e.g. `if (codePoint < 0x20 && codePoint !== 0x09 && codePoint !== 0x0a) return original;`) would keep the matched text clean. Non-blocking.

src/lib/jd-match/fetch-jd.ts
```diff
@@ -183,6 +205,17 @@ export function htmlToPlaintext(html: string): string {
     .replace(/&#39;/g, "'")
     .replace(/&nbsp;/g, " ");
 
+  // Decode numeric character references (decimal &#160; and hex &#x2013;).
+  // Generators such as Lever lean on these for typographic punctuation; left
+  // raw they leak `&#…;` fragments into the JD-match passes. Malformed refs
+  // like `&#x;` never match (the digit group requires 1+); out-of-range or
+  // surrogate code points are preserved by decodeCodePoint.
+  text = text
+    .replace(/&#(\d+);/g, (m, n: string) => decodeCodePoint(m, parseInt(n, 10)))
```
s-annam, line 228: [Nit]: The named `&amp;`→`&` pass runs before this numeric pass, so a literal `&amp;#160;` (the HTML encoding of the text `&#160;`) double-decodes into a space. Pre-existing decoder-ordering quirk, vanishingly unlikely in real ATS HTML — noting for completeness, no change needed.

s-annam (dismissed):
## PR Review: fix(jd-match): decode numeric HTML entity references in htmlToPlaintext

### Summary
Tight, well-tested fix. Decodes decimal + hex numeric character references in `htmlToPlaintext`, and hardens beyond the issue's suggested patch (out-of-range / lone-surrogate values preserved instead of throwing or emitting U+FFFD). Closes #93.

### Spec Alignment (issue #93)
| Acceptance criterion | Status | Notes |
|---|---|---|
| Decimal + hex numeric refs decode | ✅ Implemented | `&#(\d+);` + `&#x([0-9a-f]+);gi` |
| Tests for `&#160;`, `&#x2013;`, `&#8217;` + malformed `&#x;` | ✅ Implemented | Plus an out-of-range `&#1114112;` case |
| No regressions in `htmlToPlaintext` suite | ✅ Implemented | 392/392 green |
| Optional: clamp invalid Unicode scalars | ✅ Exceeded | `decodeCodePoint` guards range + surrogates |

### Highlights
- `decodeCodePoint` guard is the right call — `String.fromCodePoint(0x110000)` throws `RangeError`; returning the original text both avoids the throw and keeps malformed refs debuggable. Test pins it.
- nbsp (`0xA0`) folded to a plain space so numeric `&#160;`/`&#xA0;` match the existing named `&nbsp;` path — consistent, and keeps source ASCII-clean. Documented why.
- Regex ordering is correct: the decimal `\d+` pass can't eat the `x` in `&#x2013;`, so the hex pass handles it cleanly.

### Verdict
**APPROVE** — 0 blocking. Spec fully met and exceeded; typecheck clean, 392/392 green. Two non-gating polish items inline.

_Tested on a local checkout of the PR branch (`npm install` + `npm run typecheck` + `npm run test`)._

s-annam (approved):
Re-approve after the control-char hardening commit (`2b0fb3f`). Strips non-whitespace C0/DEL numeric refs (`&#0;`, `&#7;`…) instead of decoding them into invisible bytes; tab/LF/CR still decode as whitespace. Full suite 394/394, typecheck + lint clean, `verify` + `fallow` green. The earlier ordering nit is intentionally left as-is (pre-existing, won't occur in real ATS HTML).

Issue #110, opened 2026-06-17
L1 — Externalize section keyword config to a single file


s-annam:
## L1 — Externalize section keyword config to a single file

Part of the **Generalize resume section recognition** epic (link patched in after creation). This is the foundation child: it moves the section keyword data out of code into one editable config file, and adds the schema fields that L2 (head-noun anchors) and L3 (per-section flags) will consume.

### Background

Section keyword data is currently three structures hard-coded in `src/lib/heuristics/regex.ts`:

- `SECTION_KEYWORDS` (`regex.ts:87`) — the `{ sectionName: string[] }` exact-alias map.
- `SPLIT_LETTER_NORMALIZABLE_SECTIONS` (`regex.ts:181`) — a `Set<SectionName>` allowlist.
- `SectionName` (`regex.ts:160`) — `keyof typeof SECTION_KEYWORDS`, a literal union the rest of the codebase relies on as a discriminated key.

They sit alongside unrelated contact/date/degree regexes, so the "edit the section list" target is non-obvious. This child gives the list one canonical home.

### Goal

Move the data to **one** config file plus a thin typed loader. Behavior must not change (pure refactor) — the same aliases, the same `SectionName` union, the same exports.

### Implementation

1. **New `src/lib/heuristics/sections.config.json`** — the single editable source. One entry per section:

   ```json
   {
     "version": "1",
     "sections": {
       "experience": {
         "aliases": ["experience", "work experience", "professional experience",
                     "employment", "employment history", "work history",
                     "career", "career history"],
         "anchors": ["experience", "employment"],
         "splitLetterNormalizable": true,
         "anchorFallback": true
       },
       "skills": {
         "aliases": ["skills", "technical skills", "core competencies",
                     "competencies", "expertise", "technologies"],
         "anchors": ["skills", "competencies", "technologies", "expertise"],
         "splitLetterNormalizable": false,
         "anchorFallback": false
       }
       // …summary, education, projects, certifications, achievements, other
     }
   }
   ```

   - `aliases` ← current `SECTION_KEYWORDS` values (exact-match list).
   - `anchors` ← **new**, consumed by L2 (head-noun fallback). Populate per the epic's anchor table. Harmless to ship now (no reader yet until C2).
   - `splitLetterNormalizable` ← replaces the `SPLIT_LETTER_NORMALIZABLE_SECTIONS` set (true for summary/experience/education/certifications/projects).
[...]

Issue #109, opened 2026-06-17
Generalize resume section recognition (layered recognizer)


s-annam:
### Problem

Section-header detection is **exact-match only**. `matchSectionHeader` (`src/lib/heuristics/regex.ts:190`) classifies a line as a section header solely via `keywords.includes(normalized)` against the hand-maintained `SECTION_KEYWORDS` map (`regex.ts:87`). Real-world header variants that aren't in the list match nothing and the section silently drops out.

Live example (#108): a résumé with "Relevant Experience" and "Customer Service Experience" headings has **no work experience detected** — neither phrase is in `SECTION_KEYWORDS.experience`, so both blocks fall into `profile`/`other` and never reach the experience extractor.

### Why a finite list cannot be the answer

Corpus investigation (this is the headline finding): **there is no enumerable corpus of résumé section titles.** Career taxonomies (Purdue OWL, [Cornell career center](https://hecec.human.cornell.edu/2015/09/24/organizing-your-experiences-resume-headers)), academic work ([TSHD, Wiley 2023](https://onlinelibrary.wiley.com/doi/10.1155/2023/6044007); [IJEAST](https://www.ijeast.com/IJEAST-Paper?file=96-103%2C+Tesma1001%2CIJEAST.pdf)), and observed ATS-vendor behavior ([Resume Optimizer Pro](https://resumeoptimizerpro.com/blog/how-resume-parsers-actually-work)) all converge:

- "Experience" alone = Work / Professional / Relevant / Related / Career-Related / Employment History / Work History / Professional Background / Additional / Internship / **[Industry] Experience** ("Editorial Experience", "Customer Service Experience").
- Job seekers invent labels: "Career Journey", "My Story", "What I Have Done", "The Work".

A fixed keyword list — even one externalized to config — is the **Taleo-strict whitelist floor** (the behavior we audit *against*), not a solution to "support all industries and conventions."

### The path — layered recognizer

Robust parsers (including our own lineage, [OpenResume / Sconti](https://scontiresume.web.app/resume-parser)) converge on three layers. We adopt them, ranked by leverage:

| Layer | Mechanism | Covers |
|---|---|---|
| **L1 — exact alias** (config) | current keyword list, moved to a single config file | known labels; industry tuning without redeploy |
| **L2 — head-noun anchor** | last token of a header-shaped line ∈ a section's anchor set | "Customer Service Experience", "Editorial Experience" — **closes #108** |
[...]

PR #115, opened 2026-06-17
feat: section-recognition L2/L3 — anchor fallback + visual bounds

### LINKED ISSUE #108: "Relevant Experience" and "Customer Service Experience" section not recognized as work experience
[business_resume25.pdf](https://github.com/user-attachments/files/29022832/business_resume25.pdf)
When uploading this sample resume work experience was not detected even though there were two separate sections with experience titled "Relevant Experience" and "Customer Service Experience". 
Screenshot of missing work experience:
Image
Screenshot of the plain text box and pdf:

Image

### LINKED ISSUE #111: L2 — Head-noun anchor fallback for qualified section headers (closes #108)
## L2 — Head-noun anchor fallback for qualified section headers

Part of the **Generalize resume section recognition** epic (link patched in after creation). **Blocked by L1** (anchors live in the config file L1 introduces). **This child closes #108.**

### Problem

`matchSectionHeader` (`src/lib/heuristics/regex.ts:190`) only matches a header when the normalized line text is **exactly** one of the configured aliases. Qualified headers — a category noun preceded by a modifier — match nothing:

- "Relevant Experience", "Customer Service Experience" (#108)
- "Editorial Experience", "Industry Experience", "Leadership Experience", "Teaching Experience"

These are noun phrases whose **head noun** ("Experience") is the canonical category. That head noun is a small, closed set even though the qualifier is open-ended. This child adds a fallback keyed on the head noun.

### Approach — head-noun-LAST, not contains

After the existing exact-alias match and split-letter retry both fail in `matchSectionHeader`, add a final fallback:

> If the line is **header-shaped** and its **last token** ∈ a section's `anchors` set (from L1 config), classify it as that section.

Head-noun-**last** (not substring `contains`) is the actual grammar of section headers. It classifies "Customer Service **Experience**" while rejecting the prose false-positive class — "5 years of relevant experience leading teams" does not *end* in "experience" and is long-form, so it never triggers.

Anchor sets come from L1 config (epic table): experience → `[experience, employment]`, education → `[education]`, skills → `[skills, competencies, technologies, expertise]`, projects → `[projects]`, certifications → `[certifications, certificates, licenses]`, summary → `[summary, profile, objective]`, achievements → `[achievements, accomplishments, awards, honors, activities]`.

### FP guardrails (critical — `splitIntoSections` runs this on every PDF line)

The fallback fires only when **all** hold:

1. length ≤ 40 (already enforced at `regex.ts:192`)
2. word count ≤ 4 (qualifier + head noun; "Customer Service Experience" = 3)
3. last token ∈ the section's `anchors`
4. no terminal sentence punctuation (`.`, `!`, `?`)
5. not a bullet line
6. the section's `anchorFallback` flag is `true` — **`skills` anchors stay OFF on the raw-line path.** Two-column résumés flatten a sidebar "SKILLS" between experience entries; an anchored match there would open a section mid-experience and strand every following role. This is the same reason `SPLIT_LETTER_NORMALIZABLE_SECTIONS` excludes skills (`regex.ts:171-182`).

### Implementation

- In `matchSectionHeader` (`regex.ts:190`), after the split-letter block (`regex.ts:201-209`), add the anchor fallback using the config `anchors` + `anchorFallback` from L1's `sections.config.ts`.
- Tokenize the already-normalized text on whitespace; take the last token; check membership.
- The markdown / font-promoted paths (`splitIntoSectionsWithMarkdown`, `matchLeadingSectionKeyword`) inherit the improved `matchSectionHeader` automatically — they're tighter (heading-gated), so the same fallback is even safer there.

### Test fixture (PII — non-negotiable)

The #108 reporter attached `business_resume25.pdf`, which is a **real persona** — it cannot be committed verbatim (`tests/fixtures/pdfs/README.md` Privacy policy; the repo is public, the PDF binary is the exposure surface). Re-export an equivalent two-experience-section résumé with a **synthetic** persona:

- fake name, `@example.com` email
- phone = real area code + `555` exchange + `0100`–`0199` subscriber, e.g. `(312) 555-0123` (the only reserved-but-valid form; passes `libphonenumber-js` `isValid()`)
- headings "Relevant Experience" and "Customer Service Experience" preserved
- verify before commit: `pdftotext tests/fixtures/pdfs/<category>/<file>.pdf - | head -40`

### Acceptance criteria

- [ ] `matchSectionHeader("Relevant Experience")` → `"experience"`; `matchSectionHeader("Customer Service Experience")` → `"experience"`.
- [ ] Prose FP stays null: `matchSectionHeader("5 years of relevant experience leading teams")` → `null` (fails length / word-count / terminal-token).
- [ ] A raw-line "Skills" sidebar label does not open a section via the anchor path (anchorFallback off for skills); existing two-column behavior unchanged (`sections-column.test.ts` green).
- [ ] New synthetic #108 fixture added under the correct `tests/fixtures/pdfs/<category>/`, PII-verified via `pdftotext`, with its `*.expected.json` snapshot; the parsed result detects work experience.
- [ ] `corpus.test.ts` + `regex.test.ts` green; new unit tests pin both #108 headers and ≥2 FP cases.
- [ ] #108 closes (link via "Closes #108" in the implementing PR).

### Labels
improvement, bug, testing

### LINKED ISSUE #112: L3 — Visual-primary boundary detection + name/contact disambiguation
## L3 — Visual-primary boundary detection + name/contact disambiguation

Part of the **Generalize resume section recognition** epic (link patched in after creation). **Blocked by L1** (uses config for labeling). Independent of L2. This is the highest-value, highest-blast-radius child — split out so L1/L2 can ship first.

### Problem

Even with L1 (config) + L2 (head-noun anchors), recognition is still **keyword-gated** on the raw PDF path: `splitIntoSections` (`src/lib/heuristics/sections.ts:220`) opens a section only when `matchSectionHeader(line.text)` returns non-null. A résumé using an **invented** label ("Career Journey", "My Story", "What I Have Done") gets no boundary there — the block bleeds into the preceding section.

But résumés signal headers **visually**, and we already capture the signal: `PdfLine` carries `maxFontSize` and `allCaps` (`sections.ts:20-34`; the comment literally says *"drives name / header detection"*). `splitIntoSections` **ignores both**. The visual layer is half-wired — data present on every line, unused on the splitter. Our own lineage ([OpenResume / Sconti](https://scontiresume.web.app/resume-parser)) leads with this visual heuristic and treats keywords as the fallback; we do it backwards.

### Goal

Make a visually-distinct line open a section boundary **even with no keyword match**, then label it via the existing layers. An unrecognized header still **segments** correctly (boundary), even if we can't name the section.

### Approach

In the raw-line splitter `splitIntoSections` (`sections.ts:220`), treat a line as a header when **either**:

- **keyword path** (today): `matchSectionHeader(line.text)` is non-null → label = that section; **or**
- **visual path** (new): the line is *visually a header* AND *not a name/contact line* → open a boundary; label via L1 exact alias → L2 anchor → else `"other"` (the boundary-only sink that already exists at `regex.ts:146-157`, whose entire job is to terminate the prior section without rendering).

**Visual-header test** (all required, tune against corpus):
- `maxFontSize` meaningfully above the document body size (compute a body-size baseline like `markdown-emit.ts:computeBodyFontSize` does, `markdown-emit.ts:134`), or `allCaps` true;
- short (≤ ~40 chars, ≤ ~4 words);
- alone on its line (already true — `PdfLine` is one logical line);
- no terminal sentence punctuation, not a bullet.

**Name/contact disambiguation** (must-have, or the candidate's name becomes a header): the largest-font line at the **top** of the document is the name, not a section. Exclude a line that (a) is in the leading `profile` region before any section, and (b) matches name/contact shape (e.g. `EMAIL_RE`/`PHONE_RE`/`LINKEDIN_RE` present, or is the first large-font line above the first real section). Reference: external parsers call this out explicitly as the disambiguation layer.

### Why split from L1/L2

- Touches `splitIntoSections` **ordering / control flow**, not just a lookup — higher blast radius (`splitIntoSections` has no direct covering tests per codegraph).
- Needs an **FP corpus pass**: visual promotion can mistake a bolded job title or a large company name for a section header. Must validate against the full `tests/fixtures/pdfs/` corpus before enabling.
- Value is real (handles invented labels — the long tail no list reaches) but it should land on top of a settled L1+L2.

### Acceptance criteria

- [ ] A fixture with a visually-distinct but non-keyword header opens a section boundary at that line (block no longer bleeds into the prior section).
- [ ] A recognized visual header whose text maps via L1/L2 is labeled correctly; an unrecognized one is labeled `"other"` and still acts as a boundary.
- [ ] The candidate name line (top, large font) is **not** classified as a section header.
- [ ] Full corpus regression: `corpus.test.ts` green; no new false section splits on the existing fixtures (add assertions on section counts for a sample of fixtures).
- [ ] `splitIntoSections` gains direct unit coverage for the visual path (currently untested).

### Labels
improvement, feature

### LINKED ISSUE #114: Drift-guard comment in sections.config.ts overstates what it catches
Follow-up from PR #113 review (Rohith).

`sections.config.ts:45` comment claims the `_drift` assignment fails the build if the JSON "gains or loses a key." Only **missing** keys fail — TS excess-property checks don't apply to non-literal assignment (`rawConfig.sections`), so an *extra* JSON key passes structurally and would flow into `SECTION_KEYWORDS` at runtime mislabeled as a `SectionName`.

Fix: reword the comment to note the guard is one-directional (missing-key only). Optionally add a runtime check in `validate()` that the JSON key set equals the `SectionName` union if we want to catch extras too.

Non-blocking; doc-accuracy only.

s-annam:
## Summary

Lands L2 + L3 of the section-recognition epic (#109, on top of merged L1 #110), plus the #113-review follow-up #114.

- **#114** — reword the `sections.config.ts` `_drift` guard comment to its true one-directional nature (TS catches missing `SectionName` keys; extra JSON keys slip through structurally) + add a runtime `validate()` extra-key check.
- **#111 (L2)** — head-noun anchor fallback in `matchSectionHeader`: a qualified header ("Relevant Experience", "Customer Service Experience") matches when its last token is a section anchor, behind 7 FP guardrails (length/word-count/head-noun-last/no-terminal-punct/not-bullet/`anchorFallback` flag/header-casing). Exports `SECTION_ANCHORS` + `SECTION_ANCHOR_FALLBACKS`; `skills`/`other` stay off the fallback so a flattened two-column sidebar label can't strand experience. Synthetic Chromium fixture added.
- **#112 (L3)** — visual-primary boundary in `splitIntoSections`: a font-distinct line (ratio ≥ 1.2 over body baseline) opens a boundary even with no keyword, with name/contact disambiguation so the candidate name isn't promoted. First direct unit coverage for `splitIntoSections` (`sections.test.ts`). `allCaps` dropped as a visual signal after a corpus pass showed it was a net FP source (acronyms / skill tokens / sidebar labels); genuine all-caps headers are already caught by the keyword/anchor path.

Closes #108
Closes #111
Closes #112
Closes #114

## Test plan

- [x] `npm run lint` (eslint) clean
- [x] `npm run test` green — 438/438
- [x] Fixture persona verified synthetic — no real PII (Step 3.5)


src/lib/heuristics/regex.ts
```diff
@@ -115,6 +121,69 @@ function rejoinSplitLetters(text: string): string {
...
+  if (/[.!?]$/.test(normalized)) return null;
+  const tokens = normalized.split(/\s+/).filter((t) => t.length > 0);
+  // Guard 2: a qualified header is short (qualifier(s) + head noun).
+  if (tokens.length === 0 || tokens.length > 4) return null;
+  // Guard 7: header casing. Each word that starts with a letter must start
+  // with an uppercase letter (Title Case) — ALL CAPS satisfies this trivially.
+  // Rejects prose fragments like "i have experience" that end in a head noun.
+  const rawWords = raw.trim().split(/\s+/).filter((w) => w.length > 0);
+  for (const w of rawWords) {
+    const first = w[0];
```
Samhit21, line 177: [Suggestion]: Guard 7's casing check is bypassed by digit- or symbol-led tokens.

  `/[a-z]/.test(first)` returns false for any non-lowercase-letter — including digits, `+`, `&`, `#`, etc. So `"5 Years Experience"`, `"10+ Years Experience"`, `"3 Years Experience"` all clear the casing
  guard: first chars are `5`/`1`/`3`, none are lowercase letters. Combined with the other guards (last token `experience` ∈ anchors, 3 tokens ≤ 4, no terminal punct, not a bullet), these phrases match as
  a section header and would open a fresh `experience` boundary mid-summary/profile.
  
  The docstring at lines 170-176 says the guard separates "Relevant Experience" from "i have experience" — true for the lowercase-letter case, but doesn't address numeric qualifiers. None of the FP test
  fixtures in `regex.test.ts:59-95` cover this shape.

  Cheapest fix: require the first character of each word to be an *uppercase letter* (not just "not a lowercase letter"):

  ```ts
  for (const w of rawWords) {
    const first = w[0];
    if (!/[A-Z]/.test(first)) return null;
  }

  Trade-off: this also rejects a header like "& Skills" (& not uppercase) — probably fine, that's not a real heading shape. Adding "5 Years Experience" to the negative-case table in regex.test.ts would
  pin the guard.

src/lib/heuristics/sections.config.ts
```diff
@@ -42,13 +42,40 @@ interface SectionConfig {
...
+  "experience",
+  "education",
+  "skills",
+  "projects",
+  "certifications",
+  "achievements",
+  "other",
+]);
+
 function validate(cfg: Record<SectionName, SectionConfig>): void {
```
Samhit21, line 128: [Suggestion]: `validate()` doesn't cover the booleans or anchor uniqueness.

  The runtime check is excellent on the bits it covers — extra-JSON-keys, non-empty `aliases`, `anchors` being a string array. Two gaps worth a few more lines:

  1. **Boolean flags aren't type-checked.** `splitLetterNormalizable` and `anchorFallback` are read structurally; a JSON ship of `"true"` (string) or `1` (number) would still pass validation, then act
  truthy through `.filter(([, cfg]) => cfg.splitLetterNormalizable)` at lines 128 and 169. Not a problem today, but it's the same class of drift the new validate() exists to catch.

  2. **No cross-section anchor uniqueness check.** Today's JSON has no collisions, but `matchAnchorFallback` at `regex.ts:179-183` returns on the first iteration hit — if a future config introduces (say)
  `"projects"` in both `projects` and `other`, the section that happens to iterate first wins silently. Worth enforcing here while it's still cheap, since the fallback path is the load-bearing L2
  contract:
  
  ```ts
  const seenAnchors = new Map<string, SectionName>();
  for (const [name, section] of Object.entries(cfg) as Array<[SectionName, SectionConfig]>) {
    if (!section.anchorFallback) continue;
    for (const a of section.anchors) {
      const prior = seenAnchors.get(a);
      if (prior) throw new Error(
        `[sections.config] Anchor "${a}" appears in fallback-enabled sections "${prior}" and "${name}". Anchor sets across anchorFallback:true sections must be disjoint.`,
      );
      seenAnchors.set(a, name);
    }
  }

src/lib/heuristics/sections.ts
```diff
@@ -210,28 +216,196 @@ function mergeItemText(items: PdfTextItem[]): string {
...
+}
+
+/**
+ * True when a line carries name/contact shape — an email, phone, or LinkedIn
+ * URL. Used to keep a large contact line in the leading profile region from
+ * being promoted to a section boundary. The REs are stateful (`/g`), so reset
+ * `lastIndex` before each test.
+ */
+function hasContactShape(text: string): boolean {
+  EMAIL_RE.lastIndex = 0;
```
Samhit21: [Nit]: `hasContactShape` reaches into exported `/g` regex state — fragile, even if it works today.

  `EMAIL_RE`, `PHONE_RE`, `LINKEDIN_RE` are exported globals from `regex.ts` with the `/g` flag. The reset-and-test pattern is correct for a single synchronous caller — `splitIntoSections` runs once
  before contact extraction, and contact extraction uses `.matchAll()` which creates its own iterator. But the contract is implicit: any future code path that uses `.exec()`/`.test()` against these
  regexes after `splitIntoSections` runs would inherit a clean `lastIndex` only because this function happens to reset it. A renamed or reordered call could surface a state bug that's hard to localize.
  
  Two cheap mitigations: either (a) clone the regex inside `hasContactShape` so the original stays untouched, or (b) extract non-`/g` `*_TEST_RE` variants in `regex.ts` for boolean checks and keep the
  `/g` forms for iteration only. Either lands the same fix; (b) is the more durable answer.

Samhit21 (commented):
## PR Review: feat: section-recognition L2/L3 — anchor fallback + visual bounds

  ### Summary

  Clean L2 + L3 land. The 7-guardrail framing for the anchor fallback is the right shape, the name/contact disambiguation in `splitIntoSections` is a real fix (not a workaround), and dropping `allCaps` as
   a visual signal after the corpus pass is the kind of decision the reasoning round-trips back into the docstrings — well-evidenced. New direct unit coverage for `splitIntoSections` plus the synthetic
  Chromium fixture for the qualified-header path closes the right loops. `validate()`'s extra-key runtime check addresses the one-directional `_drift` gap from #114 cleanly.
  
  Prior review activity: two fallow bot comments about `SECTION_ANCHORS`/`SECTION_ANCHOR_FALLBACKS` being unused outside the module — already addressed in the in-file comment at `regex.ts:105-108` and not
   re-flagging.

  ### Findings

  1. **[Suggestion]** Guard 7's casing check at `regex.ts:172-176` lets digit-led tokens slip (`"5 Years Experience"`, `"10+ Years Experience"`). `/[a-z]/.test(first)` is false for digits, so the negative
   check passes when the positive check wouldn't. Trivial swap to `/[A-Z]/.test(first)` plus a fixture row in the FP table.
  2. **[Suggestion]** `validate()` in `sections.config.ts:68-100` doesn't type-check the booleans or enforce anchor uniqueness across `anchorFallback: true` sections. Future-config hazard for the
  load-bearing L2 contract.
  3. **[Nit]** `hasContactShape` in `sections.ts:306-311` mutates exported `/g` regex state. Works today; pattern is fragile if a future caller reorders or runs after these regexes have an in-flight
  iterator.
  4. **[Question]** `sections.config.json` routes the `"qualifications"` anchor to `education`. Common US-convention reading is closer to skills/profile; worth an inline comment pinning the intent
  (especially if the corpus you're tuning against is residency/CV-leaning).

  ### Verdict

  **Action: COMMENT** — none of the four are merge-blockers. #1 is the highest-value follow-up; the FP risk is real but contained to a few uncommon header shapes.

src/lib/heuristics/regex.ts
s-annam: Fixed in `e4ea6a5`. Guard now requires an uppercase lead char (`!/[A-Z]/.test(first)`) instead of "not lowercase", so digit/symbol-led tokens no longer slip through. Added `5 Years Experience` / `10+ Years Experience` / `3 Years Experience` to the negative-case table (regex.test.ts).

One real consequence worth flagging: this also closed an *incidental* match on `chromium-two-column-sidebar` — the column-flatten glued a sidebar `20%` onto the `Projects` header (`20% Projects`), which rode the exact same casing hole. It's structurally indistinguishable from the prose FP, so I accepted the loss and rebaked the snapshot (projectsCount 2→0). Recovering numeric-prefixed flattened headers properly would need the L3 visual path or sidebar-noise stripping, not the casing guard — happy to file a follow-up if you think that two-column recovery is worth chasing.

src/lib/heuristics/sections.config.ts
s-annam: Fixed in `e4ea6a5`. Both gaps closed in `validate()`:

1. Boolean flags (`splitLetterNormalizable`, `anchorFallback`) are now `typeof … === "boolean"` checked.
2. Cross-section anchor uniqueness enforced for `anchorFallback:true` sections — throws on a collision rather than letting iteration order decide silently.

Pulled both into `validateSection()` / `validateAnchorUniqueness()` helpers so `validate()` stays under the fallow complexity gate (it was tipping 15 cyclomatic with the additions inline).

src/lib/heuristics/sections.ts
s-annam: Fixed in `e4ea6a5`. Went with a variant of (a): module-level non-global clones derived from the exported `.source` — `new RegExp(EMAIL_RE.source, EMAIL_RE.flags.replace("g", ""))` etc. `hasContactShape` now `.test()`s those, so no shared `lastIndex` is touched and the pattern source stays single-sourced in regex.ts. Chose this over (b) to avoid widening the regex.ts export surface with `*_TEST_RE` variants that have a single consumer — same durability (stateless test), nothing new exported.

s-annam:
@Samhit21 all three review threads addressed and resolved in `e4ea6a5` — ready for final review.

- **Guard 7 casing** → require uppercase lead; `5 Years Experience` / `10+ Years Experience` / `3 Years Experience` pinned in `regex.test.ts`.
- **`validate()`** → boolean-type checks + cross-section anchor uniqueness, split into helpers to stay under the fallow gate.
- **`hasContactShape`** → non-global regex clones, no shared `lastIndex` mutation.

Your Guard 7 catch also closed an incidental `20% Projects` column-flatten match on `chromium-two-column-sidebar` (snapshot rebaked, projectsCount 2→0). Recovering that properly is its own concern — filed as #117 (lean toward the L3 visual-path attribution route), out of scope here.

Gates green: typecheck clean · 439/439 tests · `fallow audit` ✓ no issues in changed files. Requesting your approval when you get a chance. 🙏

PR #118, opened 2026-06-18
feat: column-gated sidebar-header recovery for two-column flattens

### LINKED ISSUE #117: Two-column flattened section headers with sidebar-noise prefixes go undetected (e.g. `20% Projects`)
## Problem

When a two-column résumé is flattened to a single text stream, a sidebar artifact can be glued onto a real section header. In `tests/fixtures/pdfs/unknown/chromium-two-column-sidebar.pdf` the skills-sidebar proficiency value `20%` lands on the `Projects` header line, so the extracted text reads:

```
20% Projects
```

PR #115 (L2 anchor fallback, #111) tightened Guard 7 to require an uppercase lead character on every word — correctly, because `20% Projects` is structurally indistinguishable from the prose false-positive class it targets (`5 Years Experience`, `10+ Years Experience`). As a result `matchSectionHeader("20% Projects")` now returns `null`, the `Projects` section is no longer opened, and the fixture's `projectsCount` dropped **2 → 0** (snapshot rebaked in #115).

Before #115 this matched only by *accident* — a digit lead char is neither lower- nor uppercase, so it slipped the old "not lowercase" guard. That was never a designed capability; it was the same hole that let prose through. So this issue tracks recovering the capability **properly**, not reverting the guard.

## Why it matters

Two-column layouts are a common résumé shape and already carry a layout-trigger score penalty. Silently dropping a whole section (Projects, here) on top of that penalty understates what a tolerant parser can recover, which is exactly the parser-audit signal ResumeLint exists to surface.

## Direction (not prescriptive)

The fix is **not** loosening Guard 7 — that reopens the prose FP. Options to explore:

1. **Sidebar-noise stripping** — detect and drop a leading pure-noise token (e.g. a standalone `\d+%` / bar-value token) *before* header classification, so `20% Projects` normalizes to `Projects` and matches the exact-alias keyword path. Must be narrow enough not to eat real qualifiers.
2. **L3 visual path attribution** — `isVisualHeader` (#112) already detects the font-distinct boundary, but currently opens a boundary-only `other` sink (terminates the prior section without a name). Teach it to recover the *section name* from the line's anchor token when the visual signal is present, so a font-distinct `20% Projects` opens `projects` rather than `other`.

Option 2 reuses the L3 machinery this epic already shipped and is probably the more general answer (it also helps non-numeric noise-prefixed headers).

## Acceptance

- `chromium-two-column-sidebar` recovers `projects` (projectsCount back to 2) **without** reopening the prose FP — `5 Years Experience` / `10+ Years Experience` / `20% Experience` must still return `null` (pin in `regex.test.ts`).
- New direct unit coverage for the chosen mechanism.

## Refs

- PR #115 (where the loss was introduced + the reviewer thread that surfaced it)
- Epic #109 (section recognition), L2 #111, L3 #112

s-annam:
## Summary

A two-column résumé flatten can glue a sidebar bar-value onto a real section header (e.g. `20% Projects`), producing a body-size line that the text-only `matchSectionHeader` path correctly rejects as prose. After #115 tightened Guard 7 (uppercase-lead), the accidental match was lost and `chromium-two-column-sidebar` dropped `projectsCount` 2 → 0.

This recovers the section **properly** without reopening the prose FP class #115 closed. The gate is **column membership**, not font: the per-page `columnBoundaries` map is threaded into `splitIntoSections`, and a new branch in `classifyLine` recovers the section name only when the line is body-size, header-shaped, and sits in the **secondary column** (`line.x >= columnSplitX`) of a detected two-column layout. Main-column prose (`5 Years Experience`, `20% Experience`) and single-column docs never trigger it; the text-only path is unchanged.

Closes #117

## Test plan

- [x] `npm run test` green (446 passed / 35 files)
- [x] Snapshot rebaked: `chromium-two-column-sidebar.expected.json` `projectsCount` 0 → 2 + `projects` added to `fieldsPopulated`; score block unchanged; **no other fixture moved**
- [x] AC pins: `matchSectionHeader("20% Experience" / "5 Years Experience" / "10+ Years Experience")` still return `null`
- [x] Column-gated unit coverage: (a) recover in secondary column, (b) main-column negative, (c) single-column negative

### DIFF
--- a/src/lib/heuristics/openresume.ts
+++ b/src/lib/heuristics/openresume.ts
@@ -73,7 +73,7 @@ export function parseHeuristic(
       sectionSource = "markdown";
     }
   }
-  if (!sections) sections = splitIntoSections(lines);
+  if (!sections) sections = splitIntoSections(lines, boundaries);
   return buildHeuristicResult(lines, sections, sectionSource, annotations);
 }
 
--- a/src/lib/heuristics/regex.test.ts
+++ b/src/lib/heuristics/regex.test.ts
@@ -2,7 +2,7 @@
 // Copyright 2026 The resumelint Authors
 
 import { describe, it, expect } from "vitest";
-import { matchSectionHeader } from "./regex.ts";
+import { matchSectionHeader, matchSectionAnchorToken } from "./regex.ts";
 
 describe("matchSectionHeader — split-letter headers (#56)", () => {
   it("matches a clean header unchanged", () => {
@@ -84,6 +84,11 @@ describe("matchSectionHeader — head-noun anchor fallback (#108 / #111)", () =>
     expect(matchSectionHeader("5 Years Experience")).toBeNull();
     expect(matchSectionHeader("10+ Years Experience")).toBeNull();
     expect(matchSectionHeader("3 Years Experience")).toBeNull();
+    // AC pin (#117): the text-only path stays unchanged when the L3 visual
+    // recovery lands — "20% Experience" must remain null here. The recovery
+    // for "20% Projects" lives in classifyLine's font-gated visual branch via
+    // matchSectionAnchorToken, never on this path.
+    expect(matchSectionHeader("20% Experience")).toBeNull();
   });
 
   it("rejects a header-shaped line ending in terminal punctuation", () => {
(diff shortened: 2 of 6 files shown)


s-annam:
Merging via admin bypass (no second reviewer available). This is a narrow follow-up fix to #115 / the section-recognition epic (#109): PR #115 correctly tightened Guard 7 to reject numeric-lead prose, which as a side effect dropped the accidental recovery of sidebar-glued headers like `20% Projects` (projectsCount 2→0 on the chromium-two-column-sidebar fixture). This restores that capability *properly* — column-gated recovery in the secondary column only, text-only `matchSectionHeader` untouched so every prose false-positive pin (`5/10+/20% Experience` → null) stays green. CI `verify` + `fallow` are green (full suite 446 pass); no open review threads.

PR #121, opened 2026-06-18
feat: expand experience section aliases + bullets completeness fallback

s-annam:
## Summary

- **Part A** (`sections.config.json`): Add volunteer/campus/activity headings to `sections.experience.aliases` — \"on campus involvement\", \"volunteer experience\", \"community service\", \"leadership\", \"involvement\", \"extracurricular\", \"extracurricular activities\", \"activities\", \"volunteer\", \"volunteering\". Remove \"volunteer\"/\"volunteering\" from `sections.other.aliases` and move \"activities\" from `sections.achievements` (where it lived in practice, not `other` as the issue body said — intent is clear, routing is correct).
- **Part C** (`score.ts`): Pass the experience completeness check when `bullets.length > 0` even if `expEntries` is empty, so resumes with parseable bullet content aren't penalised for unrecognised section headings.

Closes #19

## Snapshot changes (all genuine improvements)

| Fixture | Before | After | Why |
|---|---|---|---|
| `laverne` (word) | experienceCount=1, confidence=0, escalation=ocr | experienceCount=2, confidence=0.9, escalation=none | "on campus involvement" now routes to experience |
| `student-projects-activities-singlecol` | achievementsCount=4, experienceCount=2, confidence=0 | achievementsCount=0, experienceCount=3, confidence=0.86 | "activities" moved from achievements → experience |
| `google-docs-nonstandard-headers` + `weasyprint-nonstandard-headers` | experienceCount=2, educationCount=2, confidence=0 | experienceCount=3, educationCount=1, confidence=0.82 | involvement/volunteer headings routed out of education |
| `awesome-cv-resume` | score=63, missing=[\"skills\",\"work experience\"] | score=67, missing=[\"skills\"] | Part C bullets fallback clears the experience check |

Two-column fixtures (chromium-asymmetric-sidebar, chromium-two-column-sidebar, google-docs-two-column, weasyprint-two-column, two-column-achievements-sidebar): **unchanged** — no regression.

## Deviation from issue body

The issue says move \"activities\" from `sections.other` → `sections.experience`. In practice it was in `sections.achievements` (a drift since config externalisation). Moved from `achievements` → `experience` per the issue's clear routing intent. Noted in commit message.

## Test plan

- [x] `npm run typecheck` clean
- [x] `npm run test` green (454/454 — updated 2 hardcoded expectations: `sections.test.ts` nonstandard-headers count + `score.test.ts` missing-sections test now uses `rawText: ""` to correctly test the zero-content case)
[...]

Issue #119, opened 2026-06-18
Separator-less date ranges not recognized — Awesome-CV/LaTeX experience entries fail to parse (experienceCount 0)


s-annam:
## Problem

Experience entries fail to parse on the **Awesome-CV `resume.tex`** template family (and any LaTeX/template export that renders the date-range dash via a glyph pdfjs drops). The "Work Experience" section header **is** detected correctly, but every role parses to zero entries, so the reconstructed-resume UI shows an **empty experience section** while the score still floats high (~63).

Reproduces exactly against the committed fixture `tests/fixtures/pdfs/latex/awesome-cv-resume.pdf` (posquit0's Awesome-CV `resume.tex`, re-exported as the synthetic "Jane Smith" persona). Current snapshot `tests/fixtures/pdfs/latex/awesome-cv-resume.expected.json`:

```json
"experienceCount": 0,
"score": { "overall": 63, "completeness": { "missing": ["skills", "work experience"] } }
```

This is a sibling of #19 ("bullets counted but experienceCount: 0"), but a **distinct root cause**: #19 is about non-standard *section header keywords*; here the header matches fine and the failure is in **entry anchoring**.

## Root cause

Experience entries are split into blocks by anchoring on a **date range** (`src/lib/heuristics/entry-blocks.ts:104`, `EntryAnchor: "date_range"` → `isAnchorLine` tests `DATE_RANGE_RE`).

In the Awesome-CV text layer, the dash between the two dates is **absent** — pdfjs drops the glyph Awesome-CV uses for the range separator. The role date lines read (verified against the fixture markdown):

```
"DevOps Engineer Sep. 2023 Mar. 2024"
"Founding Member & Site Reliability Engineer & Infrastructure Team Lead Mar. 2021 Jun. 2023"
"Software Architect Jun. 2017 May. 2018"
```

Codepoint dump of the gap between the two dates: `2023␠Mar` — a single `U+0020` space, **no separator token**.

But `DATE_RANGE_RE` (`src/lib/heuristics/regex.ts:78-83`) **requires** a separator between the two anchors:

```
(${DATE_ANCHOR})\s*(?:–|—|-|to|through)\s*(${DATE_ANCHOR}|Present|Current|Now|Ongoing)
```

With only whitespace between the anchors, the regex doesn't match →
`collectAnchors` (entry-blocks.ts:132) returns `[]` →
`parseEntryBlocks` (entry-blocks.ts:156) returns `[]` →
[...]

Issue #38, opened 2026-06-11
Per-bullet feedback — redesign from flat dump to actionable model


s-annam:
## Per-bullet feedback — redesign from a flat dump to an actionable model

Part of **M5 · UI Improvements**.

### Problem

`PerBulletFeedback` in `src/components/Result.tsx` renders **one `BulletRow` per detected bullet**, each with three check pills (action verb / 8–30-word length / metric). On a real resume this means ~35 rows dumped in a single scroll. It's accurate but unusable: no hierarchy, no prioritization, and it buries the signal ("most of your bullets lack a metric") under 35 individually-rendered lines.

### Goal

Replace the flat list with a UI that:

1. **Leads with the high-level shape** — e.g. "12 of 35 bullets need attention" plus a per-check rollup ("28 missing a metric · 9 too long · 4 weak verb"). The summary line already exists in `PerBulletFeedback`; promote it to the primary content.
2. **Categorizes** bullets by failure mode (missing metric / length / weak verb), so the user works one category at a time instead of scanning 35 mixed rows.
3. **Encourages drill-down without forcing it** — collapse the per-bullet detail behind an expand affordance per category (or a "show flagged bullets" toggle); passing bullets shouldn't render at full weight at all.
4. **Gives the user a model to act on** — order by impact, show the worst offenders first, make it obvious what to fix next.

### Stretch (flag as a separate phase — the high-value play)

Auto-**rewrite** flagged bullets to address their issues (add a metric placeholder, tighten length, lead with a strong action verb). This is where the real value is — turning a diagnostic into a fix. Likely depends on the in-browser AI rewrite work (M4 · #3-area) for the generative side; the deterministic side can pre-fill structure (verb list, length trim) without a model. **Do not bundle the rewrite into the first pass** — ship the categorized/summarized read-only redesign first, then layer rewrite on top.

### Acceptance (first pass — read-only redesign)

- [ ] High-level summary (counts + per-check rollup) is the primary content, above any per-bullet detail
- [ ] Bullets grouped by failure category; passing bullets not rendered at full weight
- [ ] Per-bullet detail is drill-down (collapsed by default), not a flat 35-row dump
- [ ] Worst-first / impact ordering so the user knows what to fix next
- [ ] `npm run typecheck` + `npm run build` + `npm run test` green

### Out of scope (this issue)

- The auto-rewrite stretch — tracked as a follow-up once the read-only redesign lands.

### Files

[...]

PR #125, opened 2026-06-19
fix(heuristics): Word-template parsing + identity-link, education & experience extraction

### LINKED ISSUE #29: Word resume template — name, phone, and skills mis-parsed (completeness 12/30)
## Word resume template — name, phone, and skills mis-parsed (completeness 12/30)

Using a resume sample from a Microsoft Word "Office Manager" template, the phone number was visible in the extracted plain-text panel but did not appear in the score card. Deeper investigation shows the phone is one of **three** parsing failures on this document; overall completeness is **12/30**.

[Chanchal sharma sample.pdf](https://github.com/user-attachments/files/28768413/Chanchal.sharma.sample.pdf) (synthetic persona — PII-verified safe: body text, Info dict, XMP metadata, and fonts/ICC layers all checked clean)

[Screenshot 1 — phone present in plain-text panel](https://github.com/user-attachments/assets/8021bb14-471e-4225-a3c3-4abcdb01fd54)
[Screenshot 2 — phone absent from score card](https://github.com/user-attachments/assets/d926f555-bb3f-405b-8f2f-12379f01927d)

## Summary

Running the full cascade on the attached PDF yields `overall 12`, `completeness 12/30`, with name, phone, and skills all reported as missing. The text layer is **clean** (ASCII + one en-dash; verified at codepoint level — no ToUnicode/CMap mojibake). The failures are entirely in the Tier-1 heuristic parser, which makes three layout assumptions that Word table/text-box templates break simultaneously.

## Reproduction

1. Open the attached PDF in ResumeLint.
2. Observe completeness = **12/30**.
3. Score card shows name, phone, and skills missing (among others — see "Out of scope" below for the non-bugs).

Expected: name, phone, and skills all parsed and scored.

## Root cause

Word's resume templates lay out the header name and skills section in tables/text-boxes. This produces three independent assumption-breaks in the single-pass heuristic parser:

1. **Name** (`src/lib/heuristics/extract-fields.ts:173`) — The real name renders as two separate single-word lines (`Chanchal` / `Sharma`). `extractName` rejects any candidate with `words.length < 2`, so both name lines are individually disqualified. The two-word tagline "Office Manager" wins the slot at confidence 0.25 (below the anonymous contact floor of 0.5), and name is reported missing.

2. **Phone** (`src/lib/heuristics/phone.ts:187-200`) — The phone number is `(718) 555–0100` where the separator is an **en-dash U+2013**, not an ASCII hyphen. `mightHavePhone()` pre-gates on `PHONE_RE` whose separator class `[\s.-]?` excludes Unicode dashes, so `PHONE_RE.test()` returns false and `findPhoneNumbersInText` (PR #100, `libphonenumber-js`) is never invoked. `findPhoneNumbersInText` handles en-dashes natively — the fix is to make the pre-filter gate wide enough to let the call reach it, not to widen `PHONE_RE` itself.

3. **Skills** (`src/lib/heuristics/extract-fields.ts:469`) — The skills are in a 3-column table. `SKILL_SPLIT_RE = /[,;·•|/]+|\s{2,}/` splits correctly when pdfjs assembles a column row with 2+ spaces between entries. When a row is assembled with only a single space between columns, the entire row collapses to one token (`"Organization Problem-solving Management"`), which counts as 1 skill — below the minimum of 5 (`extract-fields.ts:512`) — so skills is reported missing.

**Note:** the text layer is not the problem. Fonts are subsetted Georgia + Calibri with a valid ToUnicode map; the raw text the cascade sees is byte-for-byte the visible text. ToUnicode/CMap mojibake is **not** the cause here.

## Scope — what to fix

- [ ] **Name: reconstruct two-word name from adjacent single-word lines.** When consecutive short lines at the top of the document together form a plausible full name (two words, title-cased), treat them as a single candidate before applying the `words.length < 2` guard. (`src/lib/heuristics/extract-fields.ts:165-205`)

- [ ] **Phone: expand `mightHavePhone`'s pre-filter to recognize Unicode dashes.** In `phone.ts:187-200`, widen the pre-gate so that en-dash (U+2013), em-dash (U+2014), and figure dash (U+2012) allow the string through to `findPhoneNumbersInText`. PR #100 already wires `findPhoneNumbersInText` (which handles these chars natively) — the only missing piece is that `mightHavePhone` gates it out before it runs. Do **not** widen `PHONE_RE` itself — the fix is the pre-filter gate, not the regex matcher.

- [ ] **Skills: context-detect a single-space multi-token column run and split accordingly.** When inside a recognized skills section, detect a run of tokens that are separated by single spaces and together form a plausible skills list, then split it. Do **not** widen `SKILL_SPLIT_RE` to `\s+` unconditionally — that causes false-positive splits in ordinary prose. (`src/lib/heuristics/extract-fields.ts:469`)

## Explicitly out of scope

- **Role dates showing as missing** — The template uses placeholder text `January 20XX` / `Sep 20XX`. `YEAR_RE = /\b(19|20)\d{2}\b/` correctly rejects `20XX`. This is a **fixture artifact** (template placeholder), not a product bug. Do **not** modify `YEAR_RE` to accept `20XX` as a valid year.

- **Summary missing** — The template genuinely has no Summary, Objective, or Profile section. Not a detection failure.

- **LinkedIn missing** — No LinkedIn URL appears in the document. Not a detection failure.

## Acceptance criteria

- [ ] `parsed.name` is the real name (not "Office Manager") for the Chanchal Word PDF.
- [ ] `parsed.phone` is `(718) 555-0100` (or equivalent normalized form) for the Chanchal Word PDF.
- [ ] `parsed.skills` contains at least the skills visible in the template columns for the Chanchal Word PDF.
- [ ] Completeness score rises above the 12/30 baseline for this document (role-dates, summary, and LinkedIn remain correctly absent — no strict numeric floor beyond the current baseline).
- [ ] The committed Word fixture's expected snapshot is updated to reflect the corrected parsed output.

## Test fixture

Commit the attached PDF to `tests/fixtures/pdfs/word/chanchal-sharma-sample.pdf`. The file has been verified PII-safe across all four layers (body text via `pdftotext`, Info dict via `pdfinfo`, XMP metadata, and fonts/ICC). After implementing fixes:

```
npm run bake-fixtures
```

Review and commit the updated expected snapshot. This fixture regression-guards all three fixes simultaneously.

### LINKED ISSUE #30: Skills section scored with experience-bullet rules (action verb / metric / length)
## Summary

Skills bullets are penalized by experience-oriented scoring rules — `startsWithActionVerb`, `hasMetric`, and `wellFormedLength` (8–30 words) — because the bullet extraction pipeline has no section awareness. A skill like "Project management" fails all three checks by design, not by mistake. This is a scoring bug; the fix is to make the bullet pool section-aware so skills entries never enter the experience-bullet scoring path or appear in per-bullet feedback.

A separate but related parsing gap exists in Word-table layouts: when a document places the bullet glyph (U+2022) in one table cell and the skill text in an adjacent cell, pdfjs extracts them at different y-coordinates. The current extractor requires the bullet marker at the **start of the same line** as the text, so zero bullets are detected even though the document is visually bulleted.

Completeness scoring is correctly bullet-independent — it checks `(parsed.skills?.length ?? 0) >= 3` against Tier-1 extracted skills, not the bullet pool (`score.ts:562–564`). Do not change completeness logic as part of this fix. If the reporter observed a low completeness score, it was caused by the genuinely absent summary and LinkedIn fields, not by this bug.

---

## Reproduction

Two PDFs are attached to the original report (no PII — Chanchal Sharma is an imaginary Microsoft template persona):

**Original (3-column Word table, no explicit bullets)**
`Chanchal sharma sample.pdf` — Skills are laid out in a 3-column Word table row (e.g., "Project management | Data analysis | Communication"). No bullet markers. The scoring bug does not surface here because no skills lines enter the bullet pool. Extraction is tracked separately by #29.

**Edited (Word bullet-formatted skills)**
`Edited dates and skills Chanchal Sharma Resume.pdf` — Reporter reformatted skills as bullet points. However, the Word table places the bullet glyph in one cell and the skill text in the adjacent cell. `pdftotext` shows:

```
•

Project management, Data analysis

•

Communication Organization, Problem-solving, Management
```

The `•` (U+2022) lands on its own line with a blank line before the text. `extractBulletsFromText` (`score.ts:508–521`) uses `BULLET_MARKER_RE = /^[\s ]*[-*•●–▪◦‣▶►·□]/` and requires the marker at the **start of the text line** — so it matches the lone glyph line but finds no text on that line, and the following text line does not start with a marker. Result: 0 bullets detected.

If any skills line *did* match (e.g., a plain-text `• Project management` on one line), it would be scored with:
- `startsWithActionVerb` — fails: "Project" is not in ACTION_VERBS (`score.ts:101–111`)
- `wellFormedLength` — fails: 2 words, below `BULLET_LENGTH_MIN_WORDS=8` (`score.ts:69`)
- `hasMetric` — fails: no digit, currency, or percentage

The `PerBulletFeedback` UI (`Result.tsx:266–329`) surfaces all three failures for every bullet in the pool with no skills-section override.

---

## Root cause

**Bug 1 — Global, section-unaware bullet pool (`score.ts:508–521`)**

`extractBulletsFromText(rawText)` scans the full document rawText for lines matching `BULLET_MARKER_RE`. There is no section filter. All matching lines — regardless of whether they belong to Experience, Skills, or any other section — are pooled together and scored with the same three experience-oriented checks:

- `startsWithActionVerb` (`score.ts:101–111`, ACTION_VERBS set)
- `wellFormedLength` (`score.ts:69`, 8–30 words)
- `hasMetric` (`score.ts:82–99`)

Skills entries are categorically incompatible with all three checks. The per-bullet feedback component (`Result.tsx:238–330`, esp. 266–329) then surfaces these failures to the user with no indication that different rules apply to skills.

**Bug 2 — Lone bullet glyph on its own line (Word-table layout)**

When a Word table puts the bullet marker and the skill text in separate cells, pdfjs/pdftotext extract them at different y-coordinates, producing a `•` on its own line followed (after a blank line) by the text. `extractBulletsFromText` never joins them, so visually-bulleted Word-table skills produce zero detected bullets.

---

## Scope — what to fix

- [ ] **Section-aware bullet pool**: modify `extractBulletsFromText` (or its call site) to exclude lines that fall within a detected skills section. Skills bullets must not enter the experience scoring pool and must not appear in `PerBulletFeedback`. Reference: `score.ts:508–521`; section detection available via the existing parsed sections structure.
- [ ] **Lone-bullet-line detect + merge**: when a line contains only a bullet marker (matches `BULLET_MARKER_RE` with no trailing text after trimming), look ahead to the next non-empty line and join them before entering the pool. This recovers Word-table bullet layouts. Reference: `score.ts:508–521` (same function).

---

## Explicitly out of scope

- **"Skills section was not detected" (3-column original PDF)** — This is a Tier-1 extraction failure tracked by **#29** (Word-template name/phone/skills extraction). Do not conflate with the scoring bug here.
- **Completeness scoring** — Completeness checks `parsed.skills?.length >= 3` from Tier-1 extraction, not the bullet pool (`score.ts:562–564`). It is correctly bullet-independent. Do not modify completeness logic.
- **Summary / LinkedIn completeness deductions** — These fields are genuinely absent from the Chanchal resume. Low completeness scores for those fields are correct behavior.

---

## Acceptance criteria

- Skills bullets are **not** run through `startsWithActionVerb`, `hasMetric`, or `wellFormedLength`.
- Skills entries do **not** appear in `PerBulletFeedback` in the UI.
- A Word-table PDF where the bullet glyph and text are on separate lines (as in the edited Chanchal PDF) yields correctly detected bullets after the lone-bullet merge.
- Completeness score is **unchanged** by this fix — same output before and after for resumes with/without skills bullets.

---

## Test fixture

Reuse the committed Chanchal Word fixture already added by #29 in `tests/fixtures/pdfs/word/`. Add expected-snapshot assertions covering:

1. Skills bullets are absent from the bullet scoring pool in the scored output.
2. The lone-bullet Word-table layout (`•` on its own line) produces detected bullets after the merge.
3. Completeness score is identical whether skills are bullet-formatted or not.

Run `npm run bake-fixtures` after adding assertions to regenerate snapshots.

### LINKED ISSUE #31: Partially redacted dates not identified as role dates
## Problem

When a resume contains partially redacted dates (e.g. `August 20XX – March 20XX`), the completeness check flags the role's date field as missing — but shows no explanation. The parser does not recognize common redaction tokens as date-like values, so the user gets no actionable guidance.

## Expected behavior

Roles with redacted dates should stay flagged as incomplete (the redaction does not satisfy the date requirement). However, the UI must surface a specific guidance message when a redaction pattern is detected, for example:

> "Dates appear redacted — use 4-digit years for best results."

The redacted state must be scored **distinctly** from a wholly-missing date. A role with a detected redaction token should receive a different (partial) contribution to the completeness score than a role with no date text at all.

## Redaction patterns to detect

At minimum, the parser must detect the following token families:

- `20XX` / `XXXX`
- `####`
- `20--` (replaced-year tokens with dashes)

Other clearly intentional year-substitution patterns may be included if they arise during implementation, but scope should be limited to these common variants.

## Out of scope (decided)

- Treating redacted dates as **valid/complete** — this option was considered and rejected. Redacted dates remain incomplete.
- Silent pass-through without a UI message — rejected.

## Acceptance criteria / definition of done

- [ ] Parser detects the redaction patterns listed above and distinguishes them from a wholly-missing date.
- [ ] Roles with a detected redaction token are flagged as incomplete in the completeness check.
- [ ] The completeness score reflects the redacted state distinctly from a wholly-missing date (different numeric/percentage contribution).
- [ ] The UI displays an actionable guidance message (e.g. "Dates appear redacted — use 4-digit years for best results.") when a redaction pattern is detected.
- [ ] A new parser unit test covers detection of at least the patterns above (`20XX`, `XXXX`, `####`, `20--`).

## Attachments

Original sample resume (redacted dates):
[Chanchal sharma sample.pdf](https://github.com/user-attachments/files/28769192/Chanchal.sharma.sample.pdf)

Edited version (4-digit years, problem resolved):
[Edited dates and skills Chanchal Sharma Resume .pdf](https://github.com/user-attachments/files/28769196/Edited.dates.and.skills.Chanchal.Sharma.Resume.pdf)

rohithgollapalli:
## Summary

Branch of contact / education / experience parsing fixes surfaced from
real-resume bug reports, on top of the earlier Word-template work (#29) and
scoring fixes (#30, #31). Extraction-layer only (`src/lib`); the sole UI
change is the contact card adapting its row count.

**Contact / identity links**
- LinkedIn & GitHub recovered **document-wide** (footer / "Links" blocks),
  promoted into the contact card, de-duplicated out of the rendered body, and
  kept **out of Skills** (`isSkillToken` rejects profile URLs + bare labels).
- LinkedIn matching now symmetric with GitHub: any `linkedin.com/<handle>`
  profile (incl. bare vanity URLs) resolves; `/company`, `/jobs`, etc. excluded.
- **GitHub is an optional contact row** — absent → no "not detected" gap, no
  penalty to the detected/total ratio.
- **DOCX header/footer hyperlinks extracted.** mammoth only converts the body,
  so a header-placed "LinkedIn | GitHub" row was lost; we read the
  `word/header*.xml` / `footer*.xml` parts from the DOCX zip (jszip, added as a
  direct dep) and resolve their relationship-based targets.

**Education** — multi-qualification sections now extract **every** degree (was:
only the first, often corrupted). One chunk per qualification via a dual
boundary; handles degree-first / institution-first orders and acronym schools
(MIT, UC Berkeley).

**Experience**
- Company vs. designation: a university/college (incl. plural "Schools") counts
  as the company; expanded title keywords (Assistant, Intern, …) so designations
  are recognized.
- Boundary integrity: a wrapped bullet's marker-less tail (indented past the
  bullet marker) no longer leaks into the next entry's header.

Snapshots re-baked where wrapped-bullet / doc-wide-link changes shift counts.

Closes #29
Closes #30
Closes #31

## Test plan

- [x] `npm run test` green (494 tests)
- [x] `npm run lint` clean
- [x] Manually verified in `npm run dev` (contact card, projects, skills, DOCX header links)
- [x] Fixture personas verified synthetic — no real PII (Step 3.5): Chanchal Sharma, @example.com, (718) 555-0100

## Known follow-ups (not in this PR)
- Per-entry experience `location`; `is_current` on en-dash "Present"
- Merging wrapped-bullet tails back into responsibility text (needs a corpus re-bake)


src/lib/heuristics/entry-blocks.ts
```diff
@@ -131,18 +131,55 @@ function isAnchorLine(line: PdfLine, anchor: EntryAnchor): boolean {
  */
 function collectAnchors(lines: PdfLine[], anchor: EntryAnchor): number[] {
   const anchors: number[] = [];
+  // Reference indent for the `first_line` anchor: the x of the bullet markers.
+  // Entry headers sit at (or left of) this margin, but when a long bullet wraps
+  // onto a second, marker-less line that continuation aligns with the bullet
+  // *text* — i.e. to the RIGHT of the marker. That x relationship (not an
+  // absolute point tolerance, which fails on tightly-indented layouts) is what
+  // separates a wrapped continuation from a real new header.
+  let bulletMarkerX = Infinity;
```
s-annam: [Nit]: this local `let bulletMarkerX = Infinity` (computed inline just below) shadows the module-level `bulletMarkerX(lines)` helper defined at l.163 and duplicates its leftmost-marker computation. Prefer `const bulletMarkerX = anchor === "first_line" ? <helper>(lines) : Infinity;` (or rename the local) so there's one source of truth and no name collision.

s-annam (changes requested):
## PR Review: Word-template parsing + identity-link, education & experience extraction

**Gates (run on checked-out branch):** `typecheck` ✅ · `test` ✅ 575 passed (40 files) · `lint` ✅ · fixture PII ✅ (synthetic: Chanchal Sharma, `chanchals@example.com`, `(718) 555-0100`; no Info-dict/author leak).

### Summary
Strong, well-commented extraction branch. Logic is sound, the cascade→scorer `skillsSectionText` wiring is fully threaded, contact-row optionality is a clean refactor, snapshots re-baked. One trivial-but-real blocker (a stray NUL byte); otherwise mergeable.

### Highlights
- `skillsSectionLines` → `skillsSectionText` correctly threaded `openresume` → `cascade` (both PDF + markdown paths) → `useResumeAnalysis` → scorer `buildSkillsExclusion`; the pure scorer never re-derives sections.
- `splitColumnCells` reuses the now-exported `mergeItemText` rather than a blind `\s+` split — multi-word skills survive.
- `URLISH_RE`'s path-slash requirement deliberately spares dotted skills (`Node.js`, `ASP.NET`).
- phone en-dash fold is pre-gate-only; `findPhoneNumbersInText` still parses Unicode natively — matches the `(718) 555–0100` fixture end-to-end.
- Education multi-degree chunking + `educationFromChunk` cleanly replaces the first-degree-only loop.

### Key Findings

**1. [Blocking] — `src/lib/ingest/docx.ts:129` contains a raw NUL byte (`U+0000`) in source.**
The dedup key is `` `${text}\x00${url}` `` with a *literal* NUL, not an escape. Consequence: git classifies the file as binary (the PR shows `Bin 2468 → 7846 bytes`), so the **entire ~5KB of new DOCX header/footer logic is invisible in the PR diff and to code-scanning**. NUL bytes in source are also an editor/formatter footgun. It works at runtime and tests pass — this is a reviewability/tooling blocker, not a correctness one. Trivial fix, identical runtime string:
```ts
const key = `${text}\u0000${url}`;   // keeps the file UTF-8 text + diffable
```
Confirmed three ways: `python3` byte dump, `file` reporting `data`, and the binary diff. Introduced by this PR (base was plain text).

**2. [Suggestion] — `extractBulletsFromText` cognitive complexity 20 (>15).** Already surfaced by the github-advanced-security bot on `score.ts:612`. CI's fallow step is report-only (non-blocking), so this won't fail `verify` — but lifting the lone-bullet merge into a small helper would clear it.

**3. [Nit] — `entry-blocks.ts:140`** — local `let bulletMarkerX` shadows the module-level `bulletMarkerX(lines)` (l.163) and re-inlines the identical computation; call the helper. (Inline below.)

### Verdict
**Action:** REQUEST_CHANGES
**Rationale:** Single blocking item (#1) — fix the NUL byte so `docx.ts` stays diffable in `main`'s history. Trivial change. #2/#3 optional. Everything else is solid.

s-annam:
Two follow-ups from the architecture pass — **neither blocks this PR**:

- **#126** — split `extract-fields.ts` (now 1060 LOC) into per-field modules. Behavior-preserving refactor; this PR's +400 lines are fine where they are.
- **#127** — centralize section segmentation so the scorer/extractors stop needing per-template side-channels (the `skillsSectionText` thread added here is the example). Relates to #109. This is the leverage item — better segmentation deletes whole classes of the downstream patches this branch adds one at a time.

The heuristic accretion in this PR is disciplined (pure named predicates + "non-trigger byte-identical" + fixtures), so this is about the next refactor, not this diff.

One note tied to the blocking NUL-byte item: because the literal NUL character (`U+0000`) makes git treat `docx.ts` as binary, the ~5KB of new DOCX header/footer logic never showed in the PR *diff* (I read the full file directly — logic is sound). Once the separator is written as the JavaScript escape `\u0000` (backslash-u-0000) instead of a raw NUL byte, the file becomes UTF-8 text and diffable; worth a normal diff pass on it then.

rohithgollapalli (commented):
## Code Review

> Self-review pass by the branch author — still needs a second reviewer's approval to merge.

### Overview
Extraction-layer fixes addressing a series of real-résumé bug reports: document-wide LinkedIn/GitHub recovery + DOCX header/footer hyperlink extraction (new `jszip` dep), multi-degree education extraction, experience company-vs-designation disambiguation, and wrapped-bullet-tail boundary integrity. Bundles the prior #29/#30/#31 commits and a merge of `main` (#119/#121/#123) with re-baked snapshots + algo bump to **1.2**.

### Correctness
- 🔴 **Found & fixed during review (`6ec3d8f`):** a stray **NUL byte** (U+0000) in `docx.ts` where a space belonged (a dedup-key separator). Behavior-equivalent so tests passed, but it made git classify the file **binary** → its diff was invisible in this PR. Now textual and reviewable.
- ✅ `new RegExp` from a URL slug in `openresume.ts` uses `escapeRegex` — no injection/ReDoS.
- ✅ Scanned all touched source: no other binary/NUL files, no `console.log`/`TODO`/`.only`/`debugger` leftovers.
- ✅ Merge resolution sound: `score.ts` keeps main's unexported `const` (dead-code gate), merges both changelogs, bumps 1.1→**1.2**; re-baked snapshots match merged behavior (`awesome-cv-cv` verified equal to main's #119 output).

### Code quality & conventions
- Adheres to repo norms: SPDX headers, `@design-system` imports, domain logic in `src/lib`, dynamic imports for `jszip`/`mammoth`/`turndown` to keep the entry chunk lean, hand-rolled minimal interfaces instead of `@types/jszip`.
- Helpers are small, well-named, and commented with the *why* (`looksLikeCompany`, `isWrappedContinuation`, `isLinkedinProfileUrl`).
- Minor: `anywhereOnDoc = () => true` is a no-op `band` predicate — reads slightly oddly but clear in context.

### Test coverage — strong
New unit tests for every behavior (contact rows, skills exclusion + dotted-skill guard, vanity LinkedIn, multi-degree education across both orderings + acronym schools, title/company disambiguation, x-positioned wrapped-bullet boundary, end-to-end DOCX-header zip round-trip). Full suite **575 green**, typecheck + lint clean.

### Risks / observations (non-blocking)
- ⚠️ **PR size:** 51 files bundling ~6 distinct fixes + 4 prior commits + a merge is a lot for one reviewer. Suggest a file-by-file skim.
- **XML-via-regex:** `parseHeaderFooterHyperlinks` parses DOCX header XML with regex, not a real parser — pragmatic and tested, fragile against unusual `<w:hyperlink>` nestings.
- **Heuristic constants:** `splitColumnCells` spacer width and `isWrappedContinuation` `markerX + 2` — commented, low risk.
- **Documented follow-ups (not in PR):** per-entry experience `location`, `is_current` on en-dash "Present", and merging wrapped-bullet tails back into responsibility text.

### Security
No secrets, no new network calls (jszip operates on in-browser DOCX bytes; everything stays client-side). Fixture PII preflight passed — synthetic personas.

### Verdict
Approve-worthy once a second reviewer signs off. One real defect (NUL byte) surfaced and fixed in this pass; the rest is well-tested and convention-clean. Recommend a reviewer skim the now-visible `docx.ts` and the `extract-fields.ts` education/disambiguation hunks specifically.

src/lib/ingest/docx.ts
```diff
@@ -55,6 +68,119 @@ async function loadTurndown(): Promise<new () => TurndownService> {
...
+
+const XML_ENTITIES: Record<string, string> = {
+  "&amp;": "&",
+  "&lt;": "<",
+  "&gt;": ">",
+  "&quot;": '"',
+  "&apos;": "'",
+};
+function decodeXml(s: string): string {
+  return s.replace(/&(amp|lt|gt|quot|apos);/g, (m) => XML_ENTITIES[m] ?? m);
```
Samhit21: [Suggestion]: `decodeXml` decodes named entities only — numeric character references survive unchanged.

  DOCX hyperlink XML can legitimately encode characters as numeric references — `&#38;` (decimal) and `&#xA0;` (hex) are valid XML and emitted by some Word save paths and a lot of templating tools,
  particularly inside URL `Target` attributes. The current regex only matches `&(amp|lt|gt|quot|apos);`, so a `Target="https://github.com/x?ref=a&#38;b"` round-trips as the literal string
  `https://github.com/x?ref=a&#38;b` and that's what lands in the contact card. The visible-text path has the same gap.

  It's a corner case in the wild but the round-trip is broken when it triggers, and the matching strip in `src/lib/jd-match/fetch-jd.ts` had the same gap until #117 closed it — same fix shape:
  
  ```ts
  function decodeXml(s: string): string {
    return s
      .replace(/&#(\d+);/g, (_, n) => String.fromCodePoint(parseInt(n, 10)))
      .replace(/&#x([0-9a-f]+);/gi, (_, h) => String.fromCodePoint(parseInt(h, 16)))
      .replace(/&(amp|lt|gt|quot|apos);/g, (m) => XML_ENTITIES[m] ?? m);
  }
  
  Numeric decode runs first so a &amp;#38; (double-encoded) still reduces correctly.

src/lib/heuristics/openresume.ts
```diff
@@ -94,6 +95,71 @@ export function parseHeuristicFromMarkdown(
...
+): string | undefined {
+  if (!text || slugs.length === 0) return text;
+  const kept = text
+    .split("\n")
+    .map((line) => {
+      let l = line;
+      for (const slug of slugs) {
+        l = l.replace(
+          new RegExp(
+            `(?:https?:\\/\\/)?(?:www\\.)?${escapeRegex(slug)}\\/?(?![\\w/])`,
```
Samhit21: [Suggestion]: The slug regex's negative lookahead `(?![\w/])` doesn't exclude `-` (or `.`), so a slug that's a prefix of a longer handle gets stripped mid-handle.

  `\w` is `[A-Za-z0-9_]` — hyphen is not in `\w` and not `/`, so the lookahead passes when the next char is `-`. Concrete failure: a user whose promoted GitHub slug resolves to `github.com/foo`, with a
  body line `Worked on github.com/foo-bar/awesome` (a different repo / handle that happens to share the prefix), gets the prefix chopped, leaving `Worked on -bar/awesome` — corrupt mid-handle. Same shape
  applies if a slug like `github.com/foo` appears immediately before a `.` in the body (extension, sentence end).

  Two small follow-ons in the same call site:
  1. The `new RegExp(...)` has no `/g` flag, so only the **first** occurrence per line is stripped. A line that mentions the identity link twice (e.g. `LinkedIn: linkedin.com/in/x — also at
  linkedin.com/in/x`) keeps the second copy. Rare, but the dedup intent is "every occurrence."
  2. Hot-path allocation: a fresh `RegExp` is compiled per slug per line per affected entry. Bounded but pre-compilable to one regex per slug, hoisted out of the `.map`.

  Fix shape:

  ```ts
  new RegExp(
    `(?:https?:\\/\\/)?(?:www\\.)?${escapeRegex(slug)}\\/?(?![\\w./-])`,
    "g",
  )

  Adding . and - to the lookahead class kills the prefix-stripping case; /g makes the dedup whole-line.

Samhit21 (commented):
## PR Review: fix(heuristics): Word-template parsing + identity-link, education & experience extraction

  ### Summary
  
  Re-read post the NUL-byte fix (`6ec3d8f`) so `docx.ts` is diffable. The architecture is clean — promoted-link dedup is well-isolated in `openresume.ts`, the optional-row pattern in `contact.ts` carries
  its own docstring, and `splitColumnCells` reusing `mergeItemText` is the right call. s-annam's review covered the load-bearing items (NUL byte, `extractBulletsFromText` complexity, `bulletMarkerX`
  shadowing) and split out the right follow-up architecture issues (#126 / #127). Two net-new suggestions below — both real, neither blocking.

  ### Findings

  1. **[Suggestion]** `decodeXml` at `docx.ts:88` misses numeric XML entity references (`&#38;`, `&#xA0;`). Same gap that #117 closed in `fetch-jd.ts`; cheap fix is two extra `.replace()` calls before the
   named-entity pass. Hyperlink URLs and visible text from some Word save paths round-trip corrupted today.
  2. **[Suggestion]** `stripPromotedUrls`' slug regex at `openresume.ts:144` uses lookahead `(?![\w/])`, which doesn't exclude `-` or `.` — a user whose GitHub slug is a prefix of another handle in the
  body (e.g. `foo` vs `foo-bar`) gets chopped mid-token. Same regex also lacks `/g`, so a line that mentions the same identity link twice keeps the second copy. Hoist+widen the lookahead and add `/g`.
  
  ### Verdict

  **Action: COMMENT** — neither finding is merge-blocking. #1 is the higher-value catch (closer parallel to a recently-fixed bug elsewhere in the lib); #2 is the more interesting reasoning bug but
  probability of triggering in the wild is low.

src/lib/heuristics/entry-blocks.ts
rohithgollapalli: Fixed in `1ccef64`. Replaced the local `let bulletMarkerX = Infinity` (and its inline scan) with `const markerX = anchor === "first_line" ? bulletMarkerX(lines) : Infinity;` — one source of truth, no shadow, and the non-`first_line` anchors still skip the scan.

src/lib/ingest/docx.ts
rohithgollapalli: Fixed in `1ccef64` using your exact shape — decimal + hex numeric refs decoded before the named-entity pass (so a double-encoded `&amp;#38;` still reduces). Added a unit test in `docx.test.ts` (`&#38;`→`&`, `&#x3D;`→`=` round-trip through a hyperlink `Target`). Good catch on the #117 parallel.

src/lib/heuristics/openresume.ts
rohithgollapalli: Fixed in `1ccef64`. Lookahead widened to `(?![\w./-])` so a longer handle sharing a prefix (`github.com/jane` vs `github.com/jane-doe`) is no longer chopped — covered by a new test in `openresume-markdown.test.ts`. Also hoisted the `RegExp` compilation out of the per-line `.map` (one per slug). Note: the `g` flag was already present (`"ig"`) on the current branch, so every-occurrence stripping already held — kept it.

src/lib/score/score.ts
```diff
@@ -530,18 +605,38 @@
  * resumes use markers, and grading paragraphs would either over-count
  * narrative summary lines or require the experience-section detection we
  * don't have without an LLM.
+ *
+ * `excludeStripped` (optional) is the marker-stripped skills-section line set;
+ * matching lines are dropped so bulleted skills never enter the pool (#30).
  */
-function extractBulletsFromText(text: string): string[] {
+function extractBulletsFromText(
```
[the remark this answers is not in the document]
rohithgollapalli: Deferred — `extractBulletsFromText` is pre-existing code from `main` (#121), not introduced by this PR, and fallow runs report-only so it won't fail `verify`. Leaving this open; the lone-bullet-merge extraction is a sensible follow-up cleanup but out of scope for this review pass.

rohithgollapalli:
### Revision pushed: `1ccef64`

Addressed the review feedback:

| Thread | Resolution |
|---|---|
| NUL byte in `docx.ts` (s-annam, **blocking**) | ✅ Fixed earlier in `6ec3d8f` — file is UTF-8 text + diffable now (Samhit21 confirmed) |
| `bulletMarkerX` shadow (s-annam, nit) | ✅ Fixed — calls the module-level helper |
| `decodeXml` numeric entities (Samhit21) | ✅ Fixed + test (decimal & hex refs) |
| `stripPromotedUrls` slug lookahead (Samhit21) | ✅ Fixed + test (longer prefix-sharing handle preserved); regex hoisted |
| `extractBulletsFromText` complexity (bot) | ↪️ Deferred — pre-existing from #121, fallow is report-only; left open |

Gates green on the branch: typecheck ✅ · test ✅ **577 passed** · lint ✅. Re-requesting review — thanks @s-annam @Samhit21.

s-annam:
@rohithgollapalli — re-reviewed the revision (`1ccef64`). Nice work clearing the four prior threads (NUL byte, `bulletMarkerX` shadow, `decodeXml` numeric entities, `stripPromotedUrls` lookahead) — all addressed and tested. Gates green locally: typecheck ✅ · test **616 passed** ✅ · lint ✅. Browser smoke on a Word fixture parses end-to-end (contact, experience, education, skills all populate, no crash).

Holding approval on one thing — the corpus regression surface. This PR regenerates ~25 `*.expected.json` snapshots, so a green corpus run proves "output matches the new golden," not "no regression vs main." I diffed every changed snapshot against `origin/main`. Most deltas check out:

- `bulletCount`/`totalBullets` ↓ → wrapped-bullet-tail merge (expected from this PR)
- `student-projects` `projectsCount` 4→2 → **verified correct** (PDF has exactly 2 real projects; 4 was an over-count)
- `skillsCount` ↑, `github_url` recovered → improvements
- `algoVersion` 1.1→1.2 → intended `ATS_SCORE_ALGO_VERSION` bump

One delta is a real field loss I want you to confirm:

- **`unknown/weasyprint-cairo-two-column`**: `achievementsCount` 1→0, `heuristic_achievements` dropped from `fieldsPopulated`. I checked it against the PDF (it has an AWARDS section) and ran the cascade on both branches: `main` extracts it as `[{title:"", year:"2021", description:"Acme Innovation Prize, 2023"}]` — mangled by the two-column layout, but present; this branch drops it entirely (text stays in `rawText`, no structured field).

Four questions before I approve:

1. **What corpus testing did you do beyond re-baking?** Did you diff old vs new `*.expected.json` and eyeball each delta, or regenerate and confirm green?
2. **`weasyprint-cairo-two-column` achievements drop — intended?** If yes, what's the rationale; if no, let's fix it or file a follow-up.
3. The other `projectsCount` 2→1 decreases (`google-docs-skia-proxy-nonstandard`, `weasyprint-cairo-nonstandard`, `google-docs-skia-proxy-two-column`) — please confirm each is a corrected over-count, not a dropped real section.
4. With 51 files / ~6 fixes in one PR, can you give a per-fix → affected-fixtures map? That lets us verify blast radius per change rather than as one blob.

Everything else looks solid — once these are answered I'll approve.

rohithgollapalli:
@s-annam — thanks for the careful diff; you found the one real problem. Answers below.

**1. Corpus testing beyond re-baking.** During development I diffed each changed snapshot against my pre-change `HEAD` and eyeballed the deltas, and live-verified the headline fixtures (e.g. dumped parsed projects for `student-projects` 4→2). What I did **not** do rigorously was diff the final merged state field-by-field against `origin/main` — which is exactly where the `weasyprint-cairo-two-column` achievement slipped through. Fair catch.

**2. `weasyprint-cairo-two-column` achievements drop — not intended; filed as #131.** Root cause: my new `collectAnchors` rule rejects first-line anchors at `x > bulletMarkerX`. This fixture is a **two-column flatten** with two bullet margins (awards at x≈48, experience at x≈249); the global `bulletMarkerX=48` makes the award's `"2021"` line (x=54) and the whole right column read as wrapped-bullet tails → the section yields 0 blocks. On `main` the same section extracts `{title:"", year:"2021", description:"Acme Innovation Prize, 2023"}` — an empty-title entry that mis-merges two distinct awards (the 2021 is "Globex…", the description is "Acme…"). So it's mangled-noise → nothing, not usable → lost. The principled fix is column-aware segmentation (your #127), which makes the section parse *correctly* rather than either count being right. I filed **#131** (linked to #127/#118) rather than bolt a fragile single-column special-case onto this PR. If you'd prefer I preserve the count now I can add a fallback (first_line yields 0 anchors → revert to the `prevIsBullet` rule), but that just restores the empty-title noise — I lean toward the follow-up.

**3. The other `projectsCount` 2→1 drops — confirmed corrected over-counts** (dumped the raw project sections):
- `google-docs-skia-proxy-nonstandard` & `weasyprint-cairo-nonstandard`: one real `tinydb` project; the dropped 2nd was the wrapped bullet tail `"hardware."` (x=65, past marker 59).
- `google-docs-skia-proxy-two-column`: one real `opentrace` project; dropped 2nd was the tail `"OpenTelemetry export."` (x=254).

All single-column wrapped tails that `main` mis-promoted to phantom headers. The surviving project per fixture is the real, clean one.

**4. Per-fix → affected-fixtures map:**

| Fix | Snapshot effect | Fixtures |
|---|---|---|
| #29 column-skills split | `skillsCount` ↑ | awesome-cv-cv (27→32), chromium-two-column-sidebar (10→18), two-column-achievements-sidebar (8→18); + new chanchal-sharma×2 |
| Wrapped-bullet phantom removal | `projectsCount` ↓ (corrected over-counts) | google-docs-skia-proxy-nonstandard, google-docs-skia-proxy-two-column, weasyprint-cairo-nonstandard (each 2→1), student-projects (4→2) |
| Boundary header-recovery (flip side) | `projectsCount` ↑ | chromium-two-column-sidebar (2→3) — real header after a wrapped bullet now opens an entry |
| Same rule on a 2-col flatten (**regression → #131**) | `achievementsCount` 1→0 | weasyprint-cairo-two-column |
| #30 / #31 scoring | specificity / structure / completeness | bulleted-skills + redacted-date fixtures |
| algoVersion bump | `algoVersion` 1.1→1.2 | all snapshots |
| doc-wide identity links | `github_url` in `fieldsPopulated` | deedy-resume-macfonts/openfonts |

Net: every count delta is an intended improvement or a corrected over-count, except the #131 two-column-achievements regression — which I'd like to take as a follow-up since the principled fix is segmentation (#127), not a per-section x-tolerance patch. Happy to switch to the fallback instead if you'd rather not reduce the count in this PR.

s-annam (approved):
## APPROVE

Re-reviewed the revision (`1ccef64` / merged `de04de6`). All four held questions answered, and I verified the load-bearing ones against the code:

- **`weasyprint-cairo-two-column` achievements 1→0 — not a real loss.** `collectAnchors` (`entry-blocks.ts:146`, `if (lines[i].x > markerX) continue;`) uses the global min `bulletMarkerX`, which can't represent a two-column flatten with two bullet margins — confirmed. `main`'s output there was mangled empty-title noise (`{title:"", year:"2021", description:"Acme Innovation Prize, 2023"}`, mis-merging two awards), so 0 ≥ noise. Correctly filed as **#131**; right fix is column-aware segmentation (**#127**, now wired as `blocked_by`). Accepting the deferral — a single-column fallback would just restore the noise.
- **Other `projectsCount` 2→1 drops** — confirmed corrected over-counts (wrapped-bullet tails `main` mis-promoted to phantom headers).
- **Per-fix → fixture map** — delivered; every count delta is an intended improvement or corrected over-count.

Gates green on the branch: typecheck ✅ · test **616 passed** ✅ · lint ✅ · fixture PII ✅ · `verify` + `fallow` CI ✅. Browser smoke on a Word fixture parses end-to-end.

One process note, not blocking: the per-fix → fixture blast-radius map should ship *in the PR body* next time, not after a reviewer digs for it — that's the cost of a 51-file / 6-fix bundle. Ties into the #126 split. Nice work clearing the threads.

Issue #124, opened 2026-06-18
design-system: extract shared inline-result chrome (Card variant vs InlineResult primitive)


Vaishnavi1709:
## Background

Two feature components currently hand-roll the same "inline result strip" chrome — a rounded, feedback-toned panel that sits inside a larger surface:

- `src/components/features/SectionRewrite.tsx` — `ProposedSection` (~line 292; success or warning tone, added in #123)
- `src/components/features/ResumeRewriteProposed.tsx` — top-level div (~line 49; success or warning tone)

Shared shape: `rounded border border-feedback-{tone}-border bg-feedback-{tone}-bg p-3`.

A stale code comment at `SectionRewrite.tsx:27` references `RewriteButton`/`RewriteResult` (neither exists); update that comment as part of this refactor.

#63 step 5 specifies the panel should sit "in a `shared/Card`," but `Card`'s chrome (`rounded-xl border-border-light bg-surface-card p-5`) is a different *concern* from a feedback-toned compact strip. Per the 3-tier "one primitive per concern" rule, feedback-toned result chrome gets its own primitive rather than polluting `Card` with tone and padding variants. `Card` stays scoped to neutral surface panel framing.

## Ratified approach — new `InlineResult` shared component

New `src/design-system/shared/InlineResult.tsx` owns the chrome+tone:

```tsx
<InlineResult tone="success" | "warning">
  {children}
</InlineResult>
```

Maps to: `rounded border border-feedback-{tone}-border bg-feedback-{tone}-bg p-3`.

**Gap/layout is caller-owned.** `InlineResult` owns only the chrome. The flex/gap wrapping (`flex flex-col gap-3`, `gap-4`, etc.) stays at each call-site, exactly as `Card` leaves layout to its callers. This preserves each call-site's existing spacing with no visual change — do not canonicalize gap inside `InlineResult`.

**`tone` supports `success | warning` only.** Neither current call-site uses an error tone. An `error` variant can be added when a real call-site needs it; do not ship an unused variant now.

## Acceptance criteria

- [ ] `src/design-system/shared/InlineResult.tsx` created with `tone: "success" | "warning"` prop; renders `rounded border border-feedback-{tone}-border bg-feedback-{tone}-bg p-3` with `{children}`.
- [ ] Re-exported from the `@design-system` barrel (`src/design-system/index.ts`).
[...]

PR #123, opened 2026-06-18
feat(webllm): section-level rewrite + number-preservation guardrail (#63)

Vaishnavi1709:
## Summary

Phase 1 of the in-browser AI rewrite epic. Each `RoleEntry` now exposes a **Rewrite section** button that asks Qwen2.5-1.5B to rewrite the whole role's bullets at once — the model owns count/order (can merge, dedupe, drop filler, reorder), and a deterministic, model-free regex guardrail compares numeric tokens in vs. out so the UI can name the specific dropped/invented metric inline ("⚠ AI altered a metric — removed `$5K`…").

Closes #63

## What's in the diff

**New webllm modules (all unit-tested, no real model fetched in CI)**
- [src/lib/webllm/post-process.ts](src/lib/webllm/post-process.ts) (+ test) — `cleanRewriteLine` factored out of `rewrite-bullet.ts`; strips `Rewritten:` echo, list markers, straight + smart quotes, paired bold/italic, prompt-echo lines.
- [src/lib/webllm/preserve-numbers.ts](src/lib/webllm/preserve-numbers.ts) (+ test) — single-pass ATOM regex classifies each numeric token exactly once. Handles `$/€/£/¥`, percent, magnitude (`5K`, `10MB`, `2GB`), comma groups, decimals, year ranges, and bare-integer headcounts in people-management context. Sign-flips (`15%` → `-15%`) are flagged. `of` deliberately NOT in the verb prefix (over-triggers on "1 of 5"). 22 tests.
- [src/lib/webllm/rewrite-section.ts](src/lib/webllm/rewrite-section.ts) (+ test) — `rewriteSectionWithLlm(bullets, engine) → SectionRewriteResult` with `max_tokens = min(60 × bullets.length, 768)`, post-processed output split on `\n`. 19 tests.

**New / modified UI**
- [src/components/features/SectionRewrite.tsx](src/components/features/SectionRewrite.tsx) — uses `<Button>` from `@design-system` (no raw `<button>`). Renders before/after side-by-side, green border when `numbersPreserved`, amber border + token-naming warning when not. Returns `null` when WebGPU is unavailable. Auto-dismisses the proposal when the underlying bullets change (so you can't see a "Original (2) | Proposed (3)" mismatch after an edit).
[...]

Issue #66, opened 2026-06-13
Epic: On-device section rewrite + multi-model support


s-annam:
## Epic

Evolve the on-device (WebGPU / WebLLM) résumé rewrite from a **per-bullet** trigger into a **section-level** rewrite, then make the underlying model **user-selectable** from a curated set, and finally pick the shipped default from a **measurement-backed eval** rather than taste.

All inference stays client-side — no résumé text leaves the browser. Apache-2.0; model choices weighed for license + download size + quality.

## Why

- **Per-line is a local optimum.** Rewriting one bullet at a time can't dedupe across bullets, balance verb variety, merge or drop weak bullets, or keep a section coherent. Those wins only exist at section granularity.
- **One pinned model is a guess.** `MODEL_ID = "Qwen2-1.5B-Instruct-q4f16_1-MLC"` was the pilot pick. Letting users try 2–3 models — and scoring them against a rubric — turns the default into a decision we can defend.

## Key decision (made during scoping)

**Output model = replace whole block.** The model owns bullet count + order; the UI swaps the old bullet list for the proposed one as a unit (section-level accept-all / reject, no per-bullet diff). This raises hallucination surface, so Phase 1 adds a deterministic **number-preservation guardrail** as a visible trust signal. Full reasoning in #63.

## Phases (sub-issues)

| Phase | Issue | Scope |
|-------|-------|-------|
| 1 | #63 | Section rewrite, single model, replace-whole-block + number-preservation guardrail (extract-and-diff mechanism). **Highest product value; do first.** |
| 2 | #64 | Tiered model registry + user picker (Low/Standard/High tiers, hardware-capability auto-routing, engine cache keyed by model, evict on switch, model-dimensioned telemetry). |
| 3 | #65 | Rewrite-quality eval harness (deterministic rubric) to choose `DEFAULT_MODEL_ID`. |
| 4 | #67 | Chain-of-sections whole-resume pipeline — see #67 for spec. **Future / deferred; not built now.** |

**Sequencing matters:** Phase 1 before Phase 2 — building model-swap infra against the per-line path that Phase 1 replaces is wasted work. Phase 3 pairs with Phase 2 (registry enumerates models) and tests Phase 1's rewrite fn. Phase 4 layers on top of a stable Phase 1–3 stack.

### Number-preservation guardrail — extract-and-diff mechanism

[...]

Issue #64, opened 2026-06-13
Phase 2: Multi-model registry + user model picker (Qwen / Gemma / …)


s-annam:
## Summary

Let the user choose which on-device model runs the rewrite, from a small curated set. Today the model is a single pinned constant; this issue generalizes that into a **typed model registry** + an engine cache keyed by model, plus a **picker UI** and model-dimensioned telemetry.

Phase 2 of the epic. Depends on Phase 1 (section rewrite) landing first — the picker drives whatever rewrite path exists, and there is no reason to build model-swap infra against the per-line path that Phase 1 replaces.

## Background

Current single-model wiring in `src/lib/webllm/web-llm.ts`:

- `export const MODEL_ID = "Qwen2.5-1.5B-Instruct-q4f16_1-MLC"` — one pinned constant.
- `let cached: Promise<WebLlmEngine> | null` — **one** engine cached for the page.
- One-shot telemetry flags (`downloadStartedFired`, `loadedFired`) keyed implicitly to "the model."

To support choice, every one of those single-valued slots becomes keyed by model id.

## Model registry

Replace the lone `MODEL_ID` with a typed registry at `src/lib/webllm/models.ts`. The canonical interface:

```typescript
export interface ModelMetadata {
  id: string;                                          // exact WebLLM prebuiltAppConfig model_id
  name: string;
  licenseType: 'Apache-2.0' | 'Restricted-Community';
  tier: 'Low' | 'Standard' | 'High';                  // advisory display grouping in the picker
  licenseUrl?: string;                                 // required for Restricted-Community models
  downloadSizeMb: number;                              // sourced from prebuiltAppConfig vram_required_MB
}
```

The three models in the initial registry (all valid in the pinned `@mlc-ai/web-llm` version — see Step 1 below):

| id | name | licenseType | tier | licenseUrl | downloadSizeMb |
|---|---|---|---|---|---|
| `Qwen2.5-1.5B-Instruct-q4f16_1-MLC` | Qwen 2.5 (1.5B) | Apache-2.0 | Low | — | ~1630 |
| `gemma-2-2b-it-q4f16_1-MLC` | Gemma 2 (2B) | Restricted-Community | Standard | https://ai.google.dev/gemma/terms | ~1895 |
| `Llama-3.2-3B-Instruct-q4f16_1-MLC` | Llama 3.2 (3B) | Restricted-Community | High | https://www.llama.com/llama3_2/license/ | ~2264 |

```typescript
[...]

PR #129, opened 2026-06-19
feat(webllm): typed model registry + keyed engine cache (PR A of #64)

Vaishnavi1709:
## Summary

**Phase 2 of the M4 epic, backend half.** Replaces the single pinned `MODEL_ID` constant with a typed registry, keys the engine cache + one-shot telemetry flags by model id, and threads the `model` dimension through every webllm event. **The picker UI, `Dialog` primitive, and Restricted-Community consent gate land in PR B.**

This is **PR A of 2** for #64. The two are split because:
- PR A is pure backend plumbing (registry + cache + telemetry). No user-visible changes, small reviewer surface.
- PR B introduces a new design-system primitive (`Dialog`) + a new feature component (`ModelSelector`) + a consent gate that needs Reuse-analysis review and human license vetting.

Refs #64 (PR B will close it). Refs #66 (Phase 2 of the epic).

## Steps shipped (1–4 of #64)

| Step | What | Surface |
|---|---|---|
| 1 | Exact-pin `@mlc-ai/web-llm` to `0.2.84` (was `^0.2.84`). Verified all 3 registry `model_id`s exist in its `prebuiltAppConfig.model_list` with `vram_required_MB` matching the registry's `downloadSizeMb` (1630 / 1895 / 2264). | `package.json`, `package-lock.json` |
| 2 | New typed [`models.ts`](src/lib/webllm/models.ts) — `ModelMetadata`, `MODEL_REGISTRY` (Qwen2.5-1.5B Apache-2.0, Gemma-2-2B Restricted-Community, Llama-3.2-3B Restricted-Community), `DEFAULT_MODEL_ID = Qwen2.5-1.5B`, `getModelById`, `isRegisteredModelId`. **Old `MODEL_ID` export removed.** 11 tests covering registry invariants. | `src/lib/webllm/models.ts` (+ test) |
| 3 | Engine cache becomes `Map<modelId, Promise<engine>>` with **evict-on-switch** via best-effort `.unload()`. Matches the updated spec ("call teardown/unload/dispose if it exposes one, otherwise delete the Map entry and let GC reclaim it" — so `unload?` is intentionally optional). Per-entry failure resets only that model's cache slot for retry. | `src/lib/webllm/web-llm.ts` (+ tests) |
| 4 | Every webllm telemetry event gains `{ model }` prop: `webllm_download_started`, `webllm_loaded`, `webllm_first_rewrite`, `webllm_section_rewrite_started`, `webllm_section_rewrite_completed`, `webllm_first_section_rewrite`. One-shot guard flags become per-model `Set<string>`s — a NEW model id re-arms the "first rewrite" event. `webllm_capability_detected` stays model-less (fires before any model is picked). | `src/lib/analytics.ts`, all rewrite paths |

## Spec updates reconciled (the live #64 update from 2026-06-18)

| # | Spec update | PR A impact |
|---|---|---|
| 1 | `Qwen2.5-1.5B` throughout body (was `Qwen2-1.5B`) | ✅ Aligned. `DEFAULT_MODEL_ID = Qwen2.5-1.5B` matches what #63 shipped. No deviation note needed. |
| 2 | Consent Cancel/Decline → revert to previously cached model | 🔜 PR B scope |
| 3 | `Dialog` lives at `src/design-system/primitives/Dialog.tsx`, re-exported via `@design-system` (not `src/components/ui/` as the original body said — the corrected path follows CLAUDE.md's 3-tier rule) | 🔜 PR B scope |
| 4 | Download-size warning fires only on fresh download | 🔜 PR B scope |
| 5 | Engine eviction: call teardown/unload/dispose if exposed, otherwise delete Map | ✅ Exactly what `evictAllExcept` does. `engine.unload?.().catch(() => {})` handles both presence and absence; the `.catch` keeps a flaky unload from surfacing as an unhandled rejection. |
| 6 | `ModelSelector` rendered inline near `SectionRewrite`, visible only in rewrite context | 🔜 PR B scope |

### Eviction-on-failure leaves no engine

Switching A → B evicts A and asks for its `.unload()` BEFORE B's load starts (memory invariant: peak VRAM stays at one model's footprint). If B then fails (OOM, dropped network), the user is left with no resident engine. The retry path is "click Y again" — starts fresh.

The alternative (defer eviction until B resolves) doubles peak VRAM and can itself OOM mid-swap on a 4 GB GPU. That's the failure we're guarding against, so we take the trade-off explicitly.

[`web-llm.ts:14-39`](src/lib/webllm/web-llm.ts#L14-L39) spells this out in detail. Pinned by a test that asserts **both halves**: B's `.unload()` was called (eviction happened) AND A's slot is cleared on failure (retry works).

> PR B's picker UX should surface this as "loading Y will unload X; if Y fails, click X to reload X."

### Concurrent cross-model loads bypass the eviction guard

Two `loadEngine(A, …)` and `loadEngine(B, …)` calls issued in the **same microtask** (before either populates the cache) each see an empty cache and each start a load — two engines downloading concurrently. The "at most one engine in memory" invariant is therefore "in the steady state."

**Unreachable in PR A** — both consumers (`RewriteButton`, `SectionRewrite`) only ever pass `DEFAULT_MODEL_ID`, so they always hit a single cache entry.

**PR B's picker MUST serialize cross-model `loadEngine` calls** — e.g., chain them through a single in-flight promise so the second call waits for the first to settle. There's a TODO comment in [`web-llm.ts`](src/lib/webllm/web-llm.ts) calling this out.

A pinning test for this race was attempted and removed: under vitest's parallel-worker load, `vi.mock` dispatch races with concurrent dynamic `import("@mlc-ai/web-llm")` calls inside `loadEngine`'s IIFEs, and one call falls through to the real `MLCEngine.reload()` (which then throws on `globalThis.location.origin` undefined in Node). The behavior is comprehensively documented in the file docstring instead; the removed test has a detailed comment in [`web-llm.test.ts`](src/lib/webllm/web-llm.test.ts) explaining why.

## Pre-merge gates (drafted locally, NOT posted by this PR)

Both require human action — drafts live in `.claude/scratch/issue-64-prep/` (gitignored, not in the commit):

1. **Version-pin verification comment on #64** (`version-pin-comment.md`) — mechanical: pin to `0.2.84`, all 3 model IDs confirmed in `prebuiltAppConfig`, with `vram_required_MB`. Notes one nuance for PR B: `gemma-2-2b-it-q4f16_1-MLC` requires the optional WebGPU `shader-f16` feature and is not `low_resource_required` — PR B's picker should surface a "may not work on your device" hint.

2. **License-vetting comment on #64** (`license-vetting-comment.md`) — needs human judgment: template for `permitted / permitted-with-conditions / blocked` verdicts on the Gemma Terms of Use and the Llama 3.2 Community License. Required pre-merge gate per #64.

## Out of scope (deferred to PR B)

- `Dialog` primitive at `src/design-system/primitives/Dialog.tsx` + re-export via `@design-system` barrel
- Restricted-Community consent modal + `localStorage` keyed per `licenseType` + per-model vendor URL display
- Consent Cancel/Decline → revert to previously cached model (or `DEFAULT_MODEL_ID` if none)
- `ModelSelector` feature component (rendered inline near `SectionRewrite`)
- Download-size warning before fresh download
- The Reuse-analysis paragraphs for `Dialog` and `ModelSelector`

## Reviewer-flag items left open (with rationale)

- **`LoadingPanel` duplicated between `RewriteButton.tsx` and `SectionRewrite.tsx`** — pre-existing from #123, not introduced by this PR. Same family of "extract shared chrome" as [#124 InlineResult](https://github.com/resumelint-org/resumelint/issues/124). Worth a follow-up issue; I'll file separately if you want.
- **`_resetEngineCacheForTesting` does not drain pending `unload()` promises queued by `evictAllExcept`** — no current test fails (tests use fresh `vi.fn()` engine spies per test). Documented as a known limitation in the function's docstring rather than refactored to async.
- **`as const` on `MODEL_REGISTRY`** — N=3, no real type-narrowing win. Skipped.

## Test plan

- [x] `npm run test` deterministically green — **555/555** across 3 consecutive full-suite runs (was 535 on `main`; +20 net: +11 models tests, +8 web-llm tests, +2 rewrite-bullet tests, +1 rewrite-section test, -1 dropped brittle Apache-2.0 test, -1 removed flaky concurrent-cross-model test)
- [x] `npm run build` green
- [x] Manually verified in `npm run dev`:
  - Per-bullet + section rewrite still work identically to #123 (regression check)
  - Cross-instance section rewrite lock still works ("Another rewrite running…" on sibling roles)
  - From DevTools: `(await import('/src/lib/webllm/models.ts')).MODEL_REGISTRY` returns the 3 registry entries
  - From DevTools: `(await import('/src/lib/webllm/models.ts')).DEFAULT_MODEL_ID` returns `"Qwen2.5-1.5B-Instruct-q4f16_1-MLC"`

### DIFF
--- a/src/lib/webllm/models.ts
+++ b/src/lib/webllm/models.ts
@@ -1,0 +1,118 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+/**
+ * Curated WebLLM model registry — what the user can pick from.
+ *
+ * Replaces the single `MODEL_ID` constant the per-bullet pilot shipped with.
+ * Every model exposed in the UI must have a row here so the engine cache,
+ * picker, telemetry, and consent gate can read off one source of truth.
+ *
+ * Three constraints shape this list:
+ *
+ *   1. License — resumelint is Apache-2.0. The default MUST be Apache-2.0
+ *      (Qwen2.5-1.5B) so every install works without surfacing a
+ *      Restricted-Community consent modal up front. Gemma + Llama ship as
+ *      opt-in extras gated by the consent modal (see #64 Step 5).
+ *   2. Download size — every entry is multi-GB; the picker surfaces the
+ *      `downloadSizeMb` so the user can opt out before the network blows up.
+ *   3. Tier — advisory display label only. Tiers do NOT auto-route. A
+ *      capability probe in `capability.ts` MAY annotate a "recommended for
+ *      your device" hint in the picker, but it never auto-loads a higher-
+ *      tier model. See #64 spec for why.
+ *
+ * Each `id` is the exact `model_id` in `@mlc-ai/web-llm`'s
+ * `prebuiltAppConfig.model_list` — verified against the pinned version
+ * (`@mlc-ai/web-llm@0.2.84`) before this file shipped. Bumping the web-llm
+ * pin requires re-verifying these IDs against the new `prebuiltAppConfig`,
+ * because MLC has renamed model_ids across minor releases before.
+ *
+ * `downloadSizeMb` is pulled from `vram_required_MB` in the same
+ * prebuiltAppConfig entry (rounded down to int). VRAM and download bytes
+ * track closely for 4-bit quantized models; close enough for the picker's
+ * "this download is ~1.6 GB" UI.
+ */
+
+export type LicenseType = "Apache-2.0" | "Restricted-Community";
+export type ModelTier = "Low" | "Standard" | "High";
+
+export interface ModelMetadata {
+  /** Exact `model_id` from `@mlc-ai/web-llm`'s `prebuiltAppConfig.model_list`. */
+  id: string;
+  /** Human-readable name for the picker. */
+  name: string;
+  /** SPDX-style license tag. `Restricted-Community` triggers the consent gate. */
+  licenseType: LicenseType;
+  /** Advisory display tier — Low/Standard/High. NOT a routing signal. */
+  tier: ModelTier;
+  /**
+   * Required for `Restricted-Community` models. The consent modal displays
+   * this link so the user can read the vendor's terms before accepting.
+   * Apache-2.0 models omit it (license is well-known).
+   */
+  licenseUrl?: string;
+  /**
+   * Approximate one-time download size in megabytes, surfaced in the picker
+   * before the user commits to a fresh download. Sourced from the pinned
+   * web-llm version's `vram_required_MB` (4-bit quantized models' VRAM and
+   * download bytes track within ~5% of each other).
+   */
+  downloadSizeMb: number;
+}
+
+/**
+ * The three models the user can pick from. Order matters: this is also the
+ * picker's display order (Apache-2.0 first, then Restricted-Community by
+ * tier).
+ *
+ * If you add a row, also:
+ *   1. Verify the `id` exists in the pinned `@mlc-ai/web-llm`'s
+ *      `prebuiltAppConfig.model_list`.
+ *   2. Confirm `vram_required_MB` from the same entry → `downloadSizeMb`.
+ *   3. If `licenseType: 'Restricted-Community'`, set `licenseUrl` AND record
+ *      a license-vetting result on the tracking issue before merging.
+ */
+export const MODEL_REGISTRY: readonly ModelMetadata[] = [
+  {
+    id: "Qwen2.5-1.5B-Instruct-q4f16_1-MLC",
+    name: "Qwen 2.5 (1.5B)",
+    licenseType: "Apache-2.0",
+    tier: "Low",
+    downloadSizeMb: 1630,
+  },
+  {
+    id: "gemma-2-2b-it-q4f16_1-MLC",
+    name: "Gemma 2 (2B)",
+    licenseType: "Restricted-Community",
+    tier: "Standard",
+    licenseUrl: "https://ai.google.dev/gemma/terms",
+    downloadSizeMb: 1895,
+  },
+  {
+    id: "Llama-3.2-3B-Instruct-q4f16_1-MLC",
+    name: "Llama 3.2 (3B)",
+    licenseType: "Restricted-Community",
+    tier: "High",
+    licenseUrl: "https://www.llama.com/llama3_2/license/",
+    downloadSizeMb: 2264,
+  },
+];
+
+/**
+ * The model loaded when the user has not made an explicit pick.
+ *
+ * MUST be Apache-2.0 — every install gets this without triggering a
+ * Restricted-Community consent modal. A user-persisted `localStorage`
+ * selection (PR B) is the ONLY override; no hardware auto-routing.
+ */
+export const DEFAULT_MODEL_ID = "Qwen2.5-1.5B-Instruct-q4f16_1-MLC";
+
+/** Look up a model by id. Returns undefined for unknown ids. */
+export function getModelById(id: string): ModelMetadata | undefined {
+  return MODEL_REGISTRY.find((m) => m.id === id);
+}
+
+/** True when the given model id is in the registry. */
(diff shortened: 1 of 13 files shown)


src/lib/webllm/models.ts
```diff
@@ -0,0 +1,118 @@
...
+ *   3. If `licenseType: 'Restricted-Community'`, set `licenseUrl` AND record
+ *      a license-vetting result on the tracking issue before merging.
+ */
+export const MODEL_REGISTRY: readonly ModelMetadata[] = [
+  {
+    id: "Qwen2.5-1.5B-Instruct-q4f16_1-MLC",
+    name: "Qwen 2.5 (1.5B)",
+    licenseType: "Apache-2.0",
+    tier: "Low",
+    downloadSizeMb: 1630,
```
s-annam, line 81: [Suggestion]: `downloadSizeMb` comes from `vram_required_MB`, but VRAM footprint and on-disk download bytes aren't the same — the actual weight download for a 4-bit model is usually smaller than its resident VRAM. The docstring discloses this honestly, so no change needed here. Flagging only so PR B's picker copy presents it as an approximate footprint, not a precise download size, to avoid over-stating the number to users.

s-annam (approved):
## PR Review: feat(webllm): typed model registry + keyed engine cache (PR A of #64)

### Summary
Clean, well-scoped backend plumbing. Replaces the single `MODEL_ID` constant with a typed registry, re-keys the engine cache + every one-shot telemetry flag by model id, and threads the `model` dimension through all webllm events. No user-visible change. Verdict: **APPROVE**.

### Gates (run on `vk/model-registry-issue-64`)
- `npm run typecheck` — clean
- `npm run test` — **555/555 passed** (41 files)
- `npm run build` — green (only the pre-existing >500kB chunk-size advisory)
- No orphaned `MODEL_ID` callers after the export removal (only doc-comment mentions remain)

### Spec Alignment (#64, steps 1–4)
|---|---|---|
| Exact-pin `@mlc-ai/web-llm@0.2.84` | ✅ | `package.json` + lockfile |
| Typed `models.ts` registry (`ModelMetadata`, `MODEL_REGISTRY`, `DEFAULT_MODEL_ID`, lookups) | ✅ | Interface matches the #64 canonical shape exactly |
| Engine cache keyed by model id + evict-on-switch | ✅ | `Map<modelId, Promise>`, `evictAllExcept`, best-effort `.unload()` |
| `{ model }` on every webllm event; per-model one-shot guards | ✅ | `capability_detected` stays model-less, correct |
| Picker UI / `Dialog` / consent gate | 🔜 | Explicitly PR B — out of scope here |

### Highlights
- Failure-during-switch and concurrent-cross-model-load trade-offs are documented in the file docstring **and pinned by tests** — the eviction+retry test asserts both halves (B unloaded, A's slot cleared). Good discipline.
- Telemetry retry semantics correct: `downloadStartedFiredFor` stays set on failure so a retry doesn't double-fire; `loadedFiredFor` only set after `CreateMLCEngine` resolves.
- Removed flaky concurrent-race test replaced with a thorough rationale comment rather than a silent drop. Reasonable for PR A (race unreachable while both consumers pass `DEFAULT_MODEL_ID` only).

### Key Findings
No blocking items.
1. **[Suggestion]** `downloadSizeMb` is sourced from `vram_required_MB`. The docstring is honest about the approximation, but actual on-disk weight download is typically smaller than VRAM footprint — PR B's picker copy should avoid presenting this as an exact "~1.6 GB download" figure. Inline note added.
2. **[Nit]** `LoadingPanel` duplication (RewriteButton/SectionRewrite) — pre-existing from #123, correctly out of scope. Endorse filing the follow-up issue you offered.

### Verdict
**Action:** APPROVE
**Rationale:** Scope is tight and matches the #64 spec; trade-offs are documented and test-pinned; all gates green. Findings are advisory only and most land in PR B.

PR #143, opened 2026-06-22
feat(webllm): model picker + consent gate + Dialog primitive

### LINKED ISSUE #64: Phase 2: Multi-model registry + user model picker (Qwen / Gemma / …)

### LINKED ISSUE #130: design-system: extract shared model-load progress panel (LoadingPanel dup between RewriteButton + SectionRewrite)
## Background

Two feature components hand-roll the same WebLLM model-download progress panel — same chrome, same WebGPU progress bar, same "what's happening" disclosure copy. Only the headline label string differs:

- [`src/components/features/RewriteButton.tsx`](src/components/features/RewriteButton.tsx) — `LoadingPanel` at [line 193](src/components/features/RewriteButton.tsx#L193). Label: `"Loading the bullet-rewrite model (~1.2GB, one-time download)"`.
- [`src/components/features/SectionRewrite.tsx`](src/components/features/SectionRewrite.tsx) — `LoadingPanel` at [line 244](src/components/features/SectionRewrite.tsx#L244). Label: `"Loading the rewrite model (~1.2GB, one-time download)"`.

Both render: an outer `<div>` with `border-border-light` / `bg-surface-subtle` chrome, a header row with the label + `pct%`, a `role="progressbar"` track + fill, an optional progress-text line in monospace, and (in `RewriteButton`'s case) the "What's happening?" `<details>` disclosure that explains the model is downloading locally.

This violates CLAUDE.md's reuse rule:

> Never hand-roll a parallel copy or an inline one-off. If a shared piece is missing a variant, add the variant to the shared piece.

Pre-existing duplication — shipped in #123 ([PR #123](https://github.com/resumelint-org/resumelint/pull/123)) and untouched by [PR #129](https://github.com/resumelint-org/resumelint/pull/129). Flagged in #129's review pass but explicitly deferred as out of scope for #64 PR A. This issue is that follow-up.

## Proposed surface

New shared component at `src/design-system/shared/ModelLoadProgress.tsx`:

```tsx
interface ModelLoadProgressProps {
  /** 0..1 fraction reported by WebLLM. */
  progress: number;
  /** Human-readable status from WebLLM (e.g. the weight file being fetched). */
  text?: string;
  /** Headline label. Defaults to "Loading the rewrite model (~1.2GB, one-time download)". */
  label?: string;
  /** Show the "What's happening?" disclosure (`RewriteButton` does; `SectionRewrite` currently doesn't). */
  showExplainer?: boolean;
}
```

Re-export from the `@design-system` barrel. `RewriteButton.LoadingPanel` and `SectionRewrite.LoadingPanel` both delete their local copy and use this.

Pros: one canonical model-download panel; the `"What's happening?"` disclosure can be enabled in `SectionRewrite` too without a second hand-roll; future model-load surfaces (PR B's `ModelSelector` will need this too — same progress bar, switching to a different model) reuse it.

Cons: minor — design-system needs a small new export; consumers thread two extra optional props.

## Relationship to #124

[#124](https://github.com/resumelint-org/resumelint/issues/124) tracks the OTHER duplication between these two files — the success/warning result strip (`RewriteResult` in `RewriteButton`, `ProposedSection` in `SectionRewrite`). Different chrome, different purpose. Both should land before the `Dialog` primitive in PR B of #64 so the new surfaces don't add a third copy of either pattern.

## Acceptance criteria

- [ ] New shared component at `src/design-system/shared/ModelLoadProgress.tsx`.
- [ ] Re-exported from the `@design-system` barrel (`src/design-system/index.ts`).
- [ ] `RewriteButton.LoadingPanel` removed; consumer imports the shared component.
- [ ] `SectionRewrite.LoadingPanel` removed; consumer imports the shared component.
- [ ] Default label preserved per-call site so existing copy ("bullet-rewrite model" vs "rewrite model") doesn't change unless intentionally consolidated.
- [ ] No visual regression — `npm run dev`, both rewrite paths look identical to current.
- [ ] `npm run test`, `npm run typecheck`, `npm run build` green.

## Out of scope

- Removing the `~1.2GB` literal in the label string (size depends on which model is loading; will be parameterized when PR B ships the model-dimensioned label).
- Consolidating with the success/warning result strip (that's #124).

Follow-up to [PR #123](https://github.com/resumelint-org/resumelint/pull/123) + [PR #129](https://github.com/resumelint-org/resumelint/pull/129).

Vaishnavi1709:
## Summary

PR B of the M4 in-browser AI rewrite epic, completing Phase 2 (#64). PR A (#129) shipped the backend — typed registry, keyed engine cache, model-dimensioned telemetry. This PR adds the user-facing picker (Steps 5–6 of the issue): a `Dialog` primitive, a `ConsentDialog` for Restricted-Community models, the `ModelSelector` feature component mounted in `ReconstructedResume`, and a `useModelSelection` hook that persists selection + per-`licenseType` consent to `localStorage` (with cross-tab `storage`-event sync).

Carries PR A's deferred TODO: cross-model `loadEngine` calls now serialize through a single in-flight `serialChain`, and the cache is split into `loadedEngines` (eviction candidates) and `pendingByModelId` (in-flight, never evicted). Also closes #130 by replacing the two near-identical loading panels in `RewriteButton` / `SectionRewrite` with a shared `ModelLoadProgress` component.

Closes #64
Closes #130

## Reuse analysis (per CLAUDE.md 3-tier rule)

- **`Dialog` (primitive, `src/design-system/primitives/Dialog.tsx`)** — no existing modal primitive in `@design-system`; built on the native `<dialog>` element so focus trap, Esc, and `::backdrop` come free. Controlled via the `open` prop. Re-exported via `@design-system`.
- **`ModelSelector` (feature, `src/components/features/ModelSelector.tsx`)** — no existing model-picker surface; the rewrite surfaces (`RewriteButton`, `SectionRewrite`) consume a model id but don't expose selection. Picker is mounted inline at the top of the Experience section in `ReconstructedResume`, visible only in the rewrite context (returns `null` when WebGPU is unavailable, same pattern as the rewrite buttons).
- **`ConsentDialog` (feature)** — composed on the new `Dialog` primitive; not a parallel modal implementation.
- **`ModelLoadProgress` (shared, `src/design-system/shared/`)** — promoted from two hand-rolled `LoadingPanel` functions previously duplicated across `RewriteButton.tsx` and `SectionRewrite.tsx`. Both call sites now import the shared one.

## Reviewer-pass fixes already applied

- Cache-probe race with WebLLM's in-flight IndexedDB writes (row could flip to "Downloaded" mid-download) — probe now triggers on a `lastCompletedAt` bumped only after the load resolves successfully.
- Consent freshness was depending on incidental re-renders — `consentMap` is now real React state, so `recordConsent` triggers the re-render that `hasConsent` needs.
- Removed `autoFocus` from Accept (Decline is the safe consent default).
- Inverted error fallback: friendly summary first, raw error in a `<details>` disclosure.
- `LICENSE_TYPES` derived from `MODEL_REGISTRY` (no hardcoded list to maintain).
- `console.warn` on unload failure so OOM-after-switch bug reports have something to investigate.

## Pre-merge gates (carry over from PR A — handled by author on the issue)

- [ ] Version-pin verification for `@mlc-ai/web-llm` (mechanical).
- [ ] License vetting for Gemma 2 Terms of Use + Llama 3.2 Community License (human judgment; load-bearing now that the consent gate fires at runtime). Drafts in `.claude/scratch/issue-64-prep/`.

## Test plan

- [x] `npm run test` green — 597/597 deterministic across 3 consecutive full-suite runs
- [x] `npm run build` green
- [ ] Manually verified in `npm run dev`: pick each model; Restricted-Community picks fire the consent dialog; Decline reverts; cached vs uncached row labels correct; cross-tab `storage` sync works


src/components/features/ModelSelector.tsx
```diff
@@ -0,0 +1,351 @@
...
+
+/**
+ * Probe `hasModelInCache` for every registry entry. Errors per-entry are
+ * swallowed (treated as "not cached") so a flaky storage check doesn't
+ * black out the picker.
+ */
+async function probeCachedIds(
+  models: readonly ModelMetadata[],
+): Promise<ReadonlySet<string>> {
+  const { hasModelInCache } = await import("@mlc-ai/web-llm");
```
s-annam, line 385: [Suggestion - perf/lazy-load]: This `await import("@mlc-ai/web-llm")` runs from the mount-time probe effect (line 105), so the multi-MB WebLLM chunk downloads for **any** WebGPU user who views a parsed resume with bullets (`ReconstructedResume.tsx:268`, gated only on `hasBullets`) — before any rewrite intent. `capability.ts` doesn't pull web-llm, so this is the first eager import and it defeats the project's dynamic-import design (CLAUDE.md: "entry chunk stays small and unused tiers don't ship"; web-llm was previously fetched only on Rewrite click). Consider deferring the probe — `requestIdleCallback`, lazy on picker expand, or a lighter Cache Storage check — so a user who never rewrites never pays the download.

src/design-system/primitives/Dialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+  useEffect(() => {
+    const dialog = ref.current;
+    if (!dialog) return;
+    if (open && !dialog.open) {
+      // showModal() is what gives us the focus trap + backdrop. Calling it
+      // on an already-open dialog throws (InvalidStateError), so guard.
+      dialog.showModal();
+    } else if (!open && dialog.open) {
+      dialog.close();
```
s-annam, line 79: [Suggestion - primitive footgun]: This programmatic `dialog.close()` fires the native `<dialog>` `close` event, which re-invokes `onClose={onClose}` (line 88). Not hit by current consumers (ConsentDialog unmounts rather than toggling `open`), but a reuse-the-primitive consumer that sets `open=false` inside its own `onClose` handler will get a redundant second `onClose`. Consider a ref flag to suppress the event during programmatic close, or document that consumers must unmount/toggle but not also act in `onClose`.

src/lib/webllm/web-llm.ts
```diff
@@ -60,104 +63,141 @@ interface CacheableEngine extends WebLlmEngine {
   unload?: () => Promise<void>;
 }
 
-const engineCache = new Map<string, Promise<CacheableEngine>>();
+const loadedEngines = new Map<string, CacheableEngine>();
+const pendingByModelId = new Map<string, Promise<WebLlmEngine>>();
```
s-annam, line 67: [Nit]: Fast-path A returns a loaded engine synchronously, but a queued cross-model load can subsequently `evictAllExcept` + `.unload()` it. Unreachable with the single-picker consumer (rows disabled mid-load), but worth a one-line docstring note since the trade-offs are otherwise enumerated exhaustively.

s-annam (commented):
## PR Review: model picker + consent gate + Dialog primitive

### Summary
High-quality PR B of the M4 epic. Picker + consent gate + `Dialog` primitive + shared `ModelLoadProgress` (closes #130), and it lands PR A's deferred cross-model serialization in `web-llm.ts`. Tests thorough, docstrings excellent, reuse analysis honored. No blocking issues. One notable lazy-load regression + one primitive footgun worth addressing.

**Gates (run on checked-out branch):** `npm run typecheck` clean · `npm run test` 668/668 green.

### Spec alignment (#64 steps 5-6)
| Requirement | Status |
|---|---|
| Picker UI mounted near rewrite surface | ✅ |
| Per-`licenseType` consent persistence + cross-tab sync | ✅ |
| Consent gate fires before Restricted-Community load; decline reverts | ✅ |
| Cached vs fresh-download row labels | ✅ |
| Cross-model `loadEngine` serialization (PR A TODO) | ✅ |
| Friendly-first error w/ raw detail disclosure | ✅ |

### Highlights
- `serialChain` + split `loadedEngines`/`pendingByModelId` is correct and cleanly reasoned — eviction now happens *at our turn* under serialization, killing the concurrent-cross-model race. The chain never rejects (inner try/catch resolves/rejects the separate `slot`), so a failed load can't wedge the queue.
- `useModelSelection` storage I/O split into pure, registry-validated functions; `safeGet`/`safeSet` swallow sandboxed-storage throws. Strong coverage incl. cross-tab + deprecated-id fallback.
- Genuine dedup of the two `LoadingPanel` copies into shared `ModelLoadProgress`.

### Key findings (all non-blocking)
1. **[Suggestion - perf]** `ModelSelector` eagerly imports `@mlc-ai/web-llm` on mount via `probeCachedIds`. The picker mounts for any WebGPU user viewing a parsed resume with bullets, so the multi-MB WebLLM chunk now downloads on resume render — before any rewrite intent. Defeats the project's lazy-load design (web-llm was previously fetched only on Rewrite click). See inline.
2. **[Suggestion]** `Dialog` primitive double-fires `onClose` on programmatic close. See inline.
3. **[Nit]** Stale-engine micro-race in `loadEngine` fast-path A. See inline.
4. **[Nit]** `ModelSelector` docstring mentions a "Try again" button that isn't rendered (retry = re-click the errored row).

### Verdict
**COMMENT** — no blocking items; tests green. Recommend addressing #1 before merge (real shipped-bytes regression against the lazy-load design); #2 hardens the new primitive for reuse; #3/#4 are cleanup.

src/components/features/ModelSelector.tsx
Vaishnavi1709: Fixed in `0f444ae`. Deferred `probeCachedIds` from the mount-time effect to first picker interaction — `onPointerEnter` + `onFocus` on the container, plus a defense-in-depth `setProbed(true)` in `onPick` for touch/keyboard paths that may bypass the container events. A WebGPU user who never engages the picker now never pulls the WebLLM chunk. Until first interaction every row reads "Will download ~N GB (one-time)" (graceful degradation for returning users — they see it as "will download" until they hover the picker); after interaction the existing `lastCompletedAt` re-probe keeps newly-cached rows flipping to "Downloaded · runs offline".

src/design-system/primitives/Dialog.tsx
Vaishnavi1709: Fixed in `0f444ae`. Added an `isProgrammaticCloseRef` flag set `true` inside the effect before `dialog.close()`; a new `handleClose` swallows the resulting native close event so consumers that toggle `open` and also do work in `onClose` (abort a request, free a resource) no longer see a spurious second invocation per user gesture. The Esc / cancel path is unchanged — `handleCancel` still `preventDefault`s and calls `onClose` once; the subsequent effect-driven close is the one being swallowed. Existing ConsentDialog (unmount-on-close) is unaffected.

src/lib/webllm/web-llm.ts
Vaishnavi1709: Done in `0f444ae`. Added a NOTE to the `loadEngine` fast-path A bullet calling out the post-return eviction trade-off explicitly: a queued cross-model load can subsequently `evictAllExcept` + `.unload()` the engine returned synchronously here. Unreachable today (rows are disabled mid-load, so two cross-model loads can't be in flight while a third caller is mid-stream on the loaded model), but a future multi-consumer caller would need to acquire a lock around `chat.completions.create()` to be safe.

s-annam (commented):
Reviewed on the checked-out branch — typecheck clean, 668/668 tests green, build per author. One critical functional bug plus supporting findings inline.

**Headline (blocks merge):** the picker selection never reaches the rewrite buttons in the same tab, so picking a model is a no-op for already-mounted `RewriteButton` / `SectionRewrite` instances. Detail on `useModelSelection.ts`.

src/hooks/useModelSelection.ts
```diff
@@ -0,0 +1,178 @@
...
+  recordConsent: (licenseType: LicenseType) => void;
+}
+
+function readAllConsent(): Record<LicenseType, boolean> {
+  const map = {} as Record<LicenseType, boolean>;
+  for (const t of LICENSE_TYPES) map[t] = hasPersistedConsent(t);
+  return map;
+}
+
+export function useModelSelection(): ModelSelectionState {
```
s-annam, line 204: 🔴 **Critical — picker selection never reaches the rewrite buttons in the same tab.**

This hook holds selection in **per-instance** `useState`. Three independent consumers call it — `ModelSelector`, `RewriteButton` (one per bullet), `SectionRewrite` (one per role) — each gets isolated state. There is no shared store / context / `useSyncExternalStore`.

Cross-instance sync relies only on the `storage` event (line 114). But `storage` **does not fire in the tab that made the write** — only in *other* same-origin tabs. So:

- User picks model X → `setSelectedModelId(X)` updates **only ModelSelector's** state + localStorage.
- `RewriteButton` / `SectionRewrite`, mounted with the resume before the pick, keep their mount-time snapshot (`DEFAULT_MODEL_ID`) and are never notified.
- Clicking "Rewrite" then loads `DEFAULT_MODEL_ID`, not X.

The PR's stated goal "all rewrite paths consume the persisted selection" holds across tabs but not within one. Fix: back selection with a module-level store consumed via `useSyncExternalStore` (or a context provider) so an in-tab write propagates to every consumer; the cross-tab `storage` listener folds into the same store.

src/lib/webllm/web-llm.ts
```diff
@@ -60,104 +63,148 @@ interface CacheableEngine extends WebLlmEngine {
...
- * from multiple `SectionRewrite` instances for the SAME model id) all share
- * one load.
+ * Fast paths (no chaining):
+ *   - Same model already loaded → returns the engine immediately. NOTE: a
+ *     queued cross-model load can subsequently `evictAllExcept` + `.unload()`
+ *     this engine before the caller awaits inference on it. Unreachable with
+ *     the current single-picker consumer (rows are disabled while a load is
+ *     in-flight, so two cross-model loads can't be in flight while a third
+ *     caller is mid-stream on the loaded model), but a future multi-consumer
+ *     caller would need to acquire a lock around `chat.completions.create()`
```
s-annam: 🟠 **Important — this safety claim is stale in PR B.** "unreachable with the current single-picker consumer (rows are disabled while a load is in-flight)" is no longer true: `RewriteButton` / `SectionRewrite` are separate consumers, not disabled by the picker's loading state. While the picker downloads model B, a user can trigger a bullet rewrite that fast-paths to loaded engine A (line 104), then the picker's chain runs `evictAllExcept(B)` → `A.unload()` mid-inference. The "would need a lock around `chat.completions.create()`" case you defer now exists. At minimum update the comment; ideally guard the in-flight engine from eviction.

src/hooks/useModelSelection.test.ts
```diff
@@ -0,0 +1,121 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+/**
+ * Tests for the pure I/O surface of `useModelSelection`. The React hook
+ * itself is a thin wrapper around these functions (matches the pattern in
+ * `useSectionRewriteLock.test.ts`); the Node test env doesn't run React
+ * effects, so we drive localStorage directly.
```
s-annam, line 8: 🟠 **Important — stateful layer untested, which is exactly where the critical bug lives.** These tests cover only the pure I/O functions (header admits the Node env doesn't run React effects). Multi-instance propagation, the `storage` listener, and `consentMap` re-render are all uncovered — that gap is why the in-tab stale-selection bug slipped through. Add a jsdom/RTL test rendering two consumers and asserting one's `setSelectedModelId` is observed by the other.

src/components/features/ModelSelector.tsx
```diff
@@ -0,0 +1,374 @@
...
+              loadState.kind === "loading" && loadState.modelId === model.id
+                ? loadState.progress
+                : null
+            }
+            onPick={() => onPick(model)}
+          />
+        ))}
+      </ul>
+
+      {consentModel && (
```
s-annam, line 268: 🟡 **Advisory — dialog unmounts while open instead of toggling `open=false`.** `{consentModel && <ConsentDialog open={true} …/>}` means decline/accept sets `consentModel=null` → the whole component unmounts. The `Dialog` close effect never runs `dialog.close()` on this path, so focus restoration to the previously-focused element may be lost. Consider keeping it mounted and driving the `open` prop.

s-annam (changes requested):
Requesting changes on the critical bug: the picker selection never reaches the rewrite buttons in the same tab (`useModelSelection` is per-instance `useState`; the `storage` event does not fire in the writing tab). Picking a model is a no-op for already-mounted `RewriteButton` / `SectionRewrite` instances — they load `DEFAULT_MODEL_ID`. See the inline thread on `src/hooks/useModelSelection.ts`. Fix = shared store (`useSyncExternalStore` or context). The two 🟠 (stale eviction claim + untested stateful layer) and 🟡 (dialog unmount) also stand.

src/hooks/useModelSelection.ts
Vaishnavi1709: Fixed in `b782984`. Moved selection + consent state to a module-level store consumed via `useSyncExternalStore`. Every write calls `notify()` synchronously so all subscribers — picker + every `RewriteButton` + every `SectionRewrite` — re-render with the new snapshot in the same tab. The cross-tab `storage` listener now folds into the same notify path (registered lazily once per page). Snapshot stability is preserved: `getModelIdSnapshot` returns a primitive (stable across calls); `getConsentMapSnapshot` returns the same object ref between writes — mutations replace the ref wholesale. Regression-proofed by a new jsdom integration test that mounts two consumers and asserts cross-instance propagation (see the reply on the test thread).

src/lib/webllm/web-llm.ts
Vaishnavi1709: Fixed in `b782984`. You're right — once 🔴 is fixed the rewrite consumers see the picker's selection, but the eviction race is independent of that: `RewriteButton` / `SectionRewrite` aren't gated by the picker's loading state, so a bullet rewrite can fast-path to engine A while the picker is downloading B and `evictAllExcept(B)` would call `A.unload()` mid-stream. Took the ideal path: added `acquireInference(modelId)` / `releaseInference(modelId)` + a `pendingUnload` Map. `evictAllExcept` parks the `.unload()` for any model with a positive inflight count; `releaseInference` drains the parked unload when the last in-flight call finishes. Both rewrite call sites (`rewrite-bullet.ts`, `rewrite-section.ts`) wrap `engine.chat.completions.create()` in `try { … } finally { releaseInference(modelId); }` so error paths still release. Updated the docstring on `loadEngine` accordingly.

src/hooks/useModelSelection.test.ts
Vaishnavi1709: Added in `b782984`. New file `src/hooks/useModelSelection.integration.test.tsx` runs under `@vitest-environment jsdom` (per-file pragma; project default stays Node so existing fast tests are unaffected). Uses `react-dom/client` + React 19's `act` directly — no RTL needed for two-consumer assertions. Two test cases: one mounts two Probe consumers and asserts `stateA.setSelectedModelId(X)` makes `stateB.selectedModelId` equal to X (and the DOM nodes update too, so it's a real re-render, not just a captured closure); a second case does the same for `recordConsent`. Required a `jsdom` devDep and extending vitest's include to `*.test.{ts,tsx}`. Tests pass 670/670.

src/components/features/ModelSelector.tsx
Vaishnavi1709: Fixed in `b782984`. Split `consentModel` (the data) from `consentOpen` (the visibility). Decline/accept now sets `consentOpen=false` while leaving `consentModel` in place — so the Dialog primitive's effect runs `dialog.close()` (which restores focus to the previously-focused element) before the React tree unmounts the `<dialog>` node. A follow-up `useEffect` keyed on `(consentOpen, consentModel)` defers the `setConsentModel(null)` into `requestAnimationFrame`, so by the time `consentModel && <ConsentDialog …/>` flips to `false` the close lifecycle has already run.

s-annam (approved):
**APPROVE** — model picker + consent gate + `Dialog` primitive. Clean, well-documented PR; reuse analysis is solid and the 3-tier placement (primitive/shared/feature) is correct.

**Gates:** `typecheck` ✅, `build` ✅, tests green on CI Node 20 (671/673 — the 2 local fails I hit are Node 25's native experimental `localStorage` shadowing jsdom, not a code defect).

**Verified the two riskiest spots — both correct:**
- `Dialog` drives the native `<dialog>` imperatively via `showModal()`/`close()` in an effect — so focus-trap, `::backdrop`, and Esc genuinely work (not the broken `<dialog open>` attribute path).
- `ConsentDialog` wires `onClose={onDecline}` and `Dialog`'s `onCancel` preventDefaults → Esc maps to Decline (the safe consent default). Initial focus on Decline.

**Non-blocking follow-ups (not required for merge):**
- **TOCTOU (narrow):** between `await loadEngine()` resolving on the fast path and `acquireInference()` firing inside `rewriteBulletWithLlm`, a concurrent picker switch can `evictAllExcept`→`unload()` the engine mid-use. Contained by `serialChain` + the UI's retry path; worth a tracked issue if cross-model switching during an active rewrite becomes common.
- A few silent `catch`/void-promise spots (`detectWebGpu`, probe, `safeSet`) could log for diagnosability.
- `ConsentDialog` renders `model.licenseUrl` as an `href` with no scheme guard — registry is static source so near-zero risk, but an `https://` check is cheap defense.

None block. Nice work.

Issue #127, opened 2026-06-19
Centralize section segmentation to retire downstream field-extractor special-cases


s-annam:
## Problem

A growing share of Tier-1 extraction bugs are really **segmentation failures wearing field-extraction costumes**. The document wasn't sliced into the right sections upstream, so we patch the symptom downstream in a field extractor — once per template. Recent examples:

- **#30** — bulleted skills were judged as experience bullets. Fixed by threading a `skillsSectionText` **side-channel** from the cascade into the scorer (`computeAnonymousAtsScore` → `buildSkillsExclusion`) so the scorer can exclude skills lines from its bullet pool.
- Profile links (LinkedIn/GitHub) leaking into the Skills token pool — fixed with `looksLikeContactLink` rejection inside `extractSkills`.
- Promoted identity links surviving as phantom project/achievement entries — fixed with `stripPromotedUrls` in `openresume.ts`.

Each fix is individually clean, but the **pattern** is the concern: the scorer and field extractors don't trust the section structure, so they re-filter raw text with template-specific predicates.

### The architectural tell

`computeAnonymousAtsScore` consumes `input.rawText` **plus** `input.skillsSectionText` as a side-channel (see `src/lib/score/score.ts` `AnonymousAtsScoreInput`, and the cascade wiring in `src/lib/heuristics/cascade.ts`). The scorer can't re-derive sections from `rawText`, so the cascade hands it a patched-in slice of one section. That's a leak: the scorer works on `rawText + N side-channels`, and every new "section X shouldn't be treated as Y" bug adds another side-channel or another downstream predicate instead of being solved once at the segmentation layer.

## Why now

This is the **leverage** issue, not a fix. Better/centralized section segmentation deletes whole *classes* of downstream special-cases rather than patching them one template at a time. The deterministic field-extractor file is already a god-module (tracked separately); reducing the *rate* of new heuristics there matters as much as splitting it.

Relates to **#109** (section-recognizer epic). This issue is the design/spike that frames how much downstream patching better segmentation can retire.

## Goal

A spike + design proposal (not a blind rewrite) answering:

1. **Inventory.** Which current Tier-1 field-extractor and scorer special-cases exist *only* because segmentation is imprecise? Catalog each with file:line and the template that triggered it (start from #30, the `looksLikeContactLink` skills rejection, `stripPromotedUrls`, and the experience/skills bleed cases).
[...]

PR #144, opened 2026-06-22
fix(heuristics): two-column flatten keeps flat-bullet awards as achievements (#131)

### LINKED ISSUE #131: Two-column flatten: first_line anchor drops achievements/projects when columns share a bullet margin
### Problem

`collectAnchors` (first_line anchor, `entry-blocks.ts`) rejects a header candidate when `line.x > bulletMarkerX`, where `bulletMarkerX` is the **global min** bullet-x of the section. This is correct for single-column sections (it strips marker-less wrapped-bullet tails), but it mis-fires on a **two-column flatten** where the section interleaves two bullet margins.

### Concrete case

`tests/fixtures/pdfs/unknown/weasyprint-cairo-two-column.pdf` — the Achievements section flattens to:

```
x=249 bullet  • Implemented authentication …        (right column: experience bullet)
x=48  bullet  • Globex Engineering Excellence,       (left column: real award)
x=254         200-user admin tool.                   (right column: wrapped tail)
x=54          2021                                    (left column: award year)
x=249 bullet  • Migrated legacy monolith …           (right column)
x=48  bullet  • Acme Innovation Prize, 2023          (left column: real award)
x=254         average response time by 55%.          (right column: wrapped tail)
```

`bulletMarkerX = min(249, 48) = 48`, so the award's own `"2021"` (x=54) and the entire right column (x≈254) are all `> markerX` → rejected as wrapped continuations → **0 anchors → the Awards section is dropped entirely** (`achievementsCount` 1→0 vs `main`).

> Note: `main`'s output here was also wrong — `{title:"", year:"2021", description:"Acme Innovation Prize, 2023"}`, an empty-title entry that mis-merges two distinct awards. So neither branch is correct; the section needs column-aware parsing.

### Root cause

A single global `bulletMarkerX` can't represent a two-column layout that has two distinct bullet margins. The right fix is **column-aware segmentation** so each column's entries parse independently, rather than a per-section x-tolerance heuristic.

### Relationship to other work

- Directly an instance of #127 (centralize section segmentation) — column-aware segmentation deletes this whole class of two-column-flatten patches.
- Related to #118 (column-gated sidebar-header recovery).

### Acceptance

- `weasyprint-cairo-two-column` Awards parse as two distinct achievements (`Globex Engineering Excellence — 2021`, `Acme Innovation Prize — 2023`), not 0 and not a mangled merge.
- No regression to the single-column wrapped-bullet-tail handling (e.g. `google-docs-skia-proxy-nonstandard` projects stays 1).

Surfaced in review of #125.

rohithgollapalli:
## What

Closes #131. A two-column résumé whose Awards/Achievements section is a **flat bullet list** (every item is itself a bullet, no name-led header line) was dropped entirely: `collectAnchors` found zero `first_line` anchors and `parseEntryBlocks` returned `[]`.

This adds a bullet-anchored fallback: when a `first_line` section yields no anchors, each **top-level** bullet (one sitting at the bullet-marker margin) becomes its own entry. Marker-less lines below it (a bare year, a wrapped award name) fold into that entry's title; deeper-indented sub-bullets become its body.

The fallback is strictly additive — only `first_line` sections that previously returned `[]` are affected. Sections with a real header keep ≥1 anchor and are untouched.

## Why this shape

The issue's original framing (two bullet margins interleaved in one section, global `bulletMarkerX` rejecting lines) was stale. `detectColumnBoundaries` already bands these fixtures into single-column reading order *before* sectioning, so each section is one column's clean list. The remaining bug was purely the zero-anchor flat list. This fix also re-seats cleanly on top of the #127 segmentation rework that landed on main (`bulletMarkerX`/`isWrappedContinuation`), reusing those helpers rather than duplicating them.

## Acceptance criteria

- ✅ `weasyprint-cairo-two-column` Awards parse as two distinct achievements (`Globex Engineering Excellence — 2021`, `Acme Innovation Prize — 2023`) — `achievementsCount` 0 → 2. Same for the sibling `google-docs-skia-proxy-two-column`.
- ✅ No single-column regression: `google-docs-skia-proxy-nonstandard` projects stays 1; `student-projects-activities-singlecol` unchanged; main's wrapped-bullet-tail handling preserved.

## Tests

- 2 new unit tests in `entry-blocks.test.ts` (flat-list split; sub-bullet-as-body).
- Re-baked corpus: only the 2 target two-column snapshots differ from `main`; all others byte-identical. No score changes.
- `npm run typecheck`, full `vitest` (628 passed), and `eslint` on changed files all clean.


src/lib/heuristics/entry-blocks.ts
```diff
@@ -196,12 +196,88 @@ export function parseEntryBlocks(
...
+  cfg: EntryBlockConfig,
+): EntryBlock {
+  const anchorIdx = anchors[a];
+  const nextIdx = a + 1 < anchors.length ? anchors[a + 1] : lines.length;
+
+  const titleParts = [stripBullet(lines[anchorIdx].text)];
+  const bodyBullets: PdfLine[] = [];
+  for (let i = anchorIdx + 1; i < nextIdx; i++) {
+    if (isBulletLine(lines[i])) bodyBullets.push(lines[i]);
+    else titleParts.push(lines[i].text.trim());
```
s-annam: [Suggestion]: This pushes every non-bullet line into `titleParts` regardless of where it sits — so a marker-less line appearing *after* a deeper sub-bullet (i.e. a wrapped continuation of that body bullet) gets folded into the entry title instead of staying with the body. Unreachable for the simple "name, year" award lists in the fixtures, but a wrapped sub-bullet tail would mis-route. Consider gating the title-fold on `!isWrappedContinuation(lines[i], markerX)` (or "before the first sub-bullet only"), mirroring how `buildEntryBlock` keeps wrapped tails out of the header. Non-blocking.

s-annam (dismissed):
## PR Review: two-column flatten keeps flat-bullet awards as achievements (#131)

### Summary
Clean, strictly-additive fix. The zero-anchor flat bullet list (awards where every item is a bullet, no header line) was dropped entirely; now each top-level bullet becomes an entry, marker-less lines fold into the title, and deeper sub-bullets become the body. Well-scoped, well-tested, and re-seats correctly on the #127 `bulletMarkerX`/`isWrappedContinuation` helpers rather than duplicating them.

**Gates (run on checked-out branch):** `npm run typecheck` clean · `npm run test` 628/628 · PR diff touches only the 2 target snapshots (no masked corpus regression).

### Spec alignment (#131)
| AC | Status |
|---|---|
| Two-column Awards parse as 2 achievements (`weasyprint-cairo`, `google-docs-skia-proxy`) | ✅ achievementsCount 0→2 |
| No single-column regression | ✅ only 2 snapshots differ, rest byte-identical |
| Reuse #127 helpers, no duplication | ✅ |
| Stay under cognitive-complexity bar (split `buildBulletEntry`) | ✅ |

### Highlights
- Trigger is tightly gated (`anchors.length === 0 && cfg.anchor === "first_line"`) so other anchors keep returning `[]` and real-header sections are untouched — genuinely additive.
- Top-level vs sub-bullet split via `x <= markerX + 2` matches the existing `isWrappedContinuation` tolerance — consistent margin logic.
- Two targeted unit tests (flat split + sub-bullet-as-body) plus a clean corpus rebake.

### Key findings (non-blocking)
1. **[Suggestion]** `buildBulletEntry` routes every marker-less line to the title regardless of position relative to sub-bullets — a wrapped continuation of a body sub-bullet would be pulled up into the title. See inline. Unreachable for the current award fixtures; forward-looking hardening.
2. **[Informational]** The fallback fires for any zero-anchor `first_line` section, projects included — not just achievements. Almost certainly an improvement (no fixture regressed), but the docstring frames it as awards-specific; a one-line note that it's generic to `first_line` would match the behavior.

### Verdict
**APPROVE** — no blocking items; correct, additive, gates green, snapshots clean.

src/lib/heuristics/entry-blocks.ts
rohithgollapalli: Good catch — fixed in c595f00. I went with the "before the first sub-bullet only" variant rather than gating on `!isWrappedContinuation`: the on-its-own-line award year (e.g. `2021` at x=54 vs marker x=48) is *itself* a wrapped continuation, so an `isWrappedContinuation` gate would drop the year from the title. Now a marker-less line folds into the title only before the first sub-bullet; after a sub-bullet it joins that bullet's text in the body. Added a unit test for the wrapped sub-bullet tail; full suite green (629).

s-annam (approved):
**APPROVE** — 0 blocking findings.

## Gates (checked-out branch, fresh `npm install`)
- ✅ `tsc -b --noEmit` clean
- ✅ vitest 629/629; new `entry-blocks.test.ts` 15/15
- ✅ Only the 2 target snapshots change (`weasyprint-cairo-two-column`, `google-docs-skia-proxy-two-column`), `achievementsCount` 0→2 — matches acceptance criteria
- ✅ No PDF fixtures touched → PII-safe (count-only snapshot deltas)
- ✅ 150/3 LOC, single concern, reuses #127 helpers (`bulletMarkerX`/`isBulletLine`) instead of duplicating

## Logic spot-check
`buildBulletEntry` title-vs-body state machine verified: marker-less line before the first sub-bullet folds into the title; after a sub-bullet it joins that bullet's tail. Windowing `anchorIdx+1 .. nextIdx` correct, no off-by-one. The 3 new unit tests cover title-fold, sub-bullet-as-body, and wrapped-tail.

## Non-blocking follow-ups (not gating this PR)
1. **Date-only bullet → empty-title achievement.** A bullet that is only a year (`• 2023`, no name) yields `{title:"", year:"2023"}` with score 0, emitted unfiltered (`achievementFromBlock`). This is a **pre-existing** systemic behavior shared by all three extractors, not introduced here. Cheap hardening if wanted: filter empty-`headerLines` blocks in `extractAchievements`/`extractProjects`. Low priority (unrealistic input).
2. **Greedy year-strip on joined titles.** `parseDateRange`'s `YEAR_RE` fallback consumes any bare 4-digit year mid-phrase (e.g. `• ACM Fellow, Class of 2019`). Pre-existing shared-parser limitation, now exercised at a new call site — worth a one-line JSDoc note.

The two `parseBulletList` `[]` guards (`!isFinite` markerX, zero top-level anchors) are unreachable given the caller gate + `markerX = min(x)` — safe defensive code, no action needed.

Ship it.

s-annam (approved):
All review comments addressed (wrapped sub-bullet tail fixed in c595f00, unit test added). Gates green. Approving for merge.

Issue #85, opened 2026-06-15
[Epic] Architecture pass: finish lib edges, establish primitives, enforce the rules


s-annam:
## What this epic is

An architecture pass over resumelint that turns "the core pipeline is healthy but the **edges** are unfinished" into a prioritized, intern-sized backlog. The cascade (Tier 0 extract+layout → Tier 1 openresume → Tier 1.5 regex) and the score math are in good shape. The findings cluster in two places: **lib paths that were built but never wired to UI**, and **primitives the project's own `CLAUDE.md` assumes exist but don't** (so the rules are quietly violated in-tree).

## How we went about it

1. **Mapped the real shape**, not the documented one — read `App.tsx`, `Result.tsx`, `cascade.ts`, `score.ts`, `types.ts`, the JD-match and webllm surfaces, and the editable-parse hook.
2. **Grep-verified each suspected gap** rather than trusting the prose: confirmed `onEvent` is never passed, `runCascadeFromMarkdown` has zero UI consumers, `fetch-jd.ts` has zero `.tsx` importers, there is no `Button` primitive, and overrides from `useEditableParse` are never consumed by scoring/JD.
3. **Ranked by impact × pattern-value**, not by size. Cheap wins that unlock data (telemetry) and pattern-setting primitives rank above feature breadth.
4. **Filed self-contained issues** with `file:line` anchors, an ordered plan, a Reuse analysis for each surface-adding issue, and testable acceptance criteria — so each is implementable after a `/clear`.

Smaller optional findings (parse cancel/abort, result persistence/share-link, folding JD into the overall score) were **deliberately dropped** for now — real but not pattern-setting. They live in the conversation if we want them later.

## Sub-issues

**A. Finish the paid-for lib paths (built, no UI) — highest ROI**
- [ ] #77 — Wire cascade telemetry seam (`onEvent`) into the app *(improvement)*
- [ ] #78 — DOCX upload: wire the built markdown-cascade path to the UI *(feature)*
- [ ] #79 — JD-from-URL: surface `fetch-jd.ts` in the JD input *(feature)*

**B. Establish the primitives the architecture already assumes**
- [ ] #80 — Shared UI primitives: Button, StatusBadge, ErrorState (+migrate) *(refactor)*
- [ ] #81 — React ErrorBoundary around the result surface *(feature)*

**C. Close the severed edit loop (product decision inside)**
- [ ] #82 — Edit → re-lint: make parse overrides authoritative (or honest) *(feature)*

[...]

Issue #83, opened 2026-06-15
Decompose App orchestration into a hook + split Result.tsx


s-annam:
## Problem

Two surfaces have outgrown the project's own structure rules. `App.tsx` carries orchestration logic inline (parse state machine, `handleFile`, JD memo, `formatBytes`) instead of delegating to a hook — violating CLAUDE.md "Keep business logic in `src/lib/`… Extract cross-cutting interaction state into `src/hooks/`." And `Result.tsx` is ~390 LOC, well past the ~200 LOC decomposition guideline, bundling six distinct concerns in one file.

## Current state

- `src/App.tsx` (231 LOC): `ParseState` union (L23-34), `handleFile` (L56-101), `jdMatch` memo (L46-54), `formatBytes` (L36-40), `reset` (L103). All interaction/orchestration state sits in the component.
- `src/components/Result.tsx` (390 LOC): branch + `StatusPill` + `ParsedCard` + `LimitedParsingCard` + `AtsScoreReadout` (L161-265) + `LayoutFlagsList` (L133-159) + `Dimension` (L267-316) + the inline Evidence panel (Source-PDF + extracted-text, L108-128).

## Plan

1. **`src/hooks/useResumeAnalysis.ts`** — owns `ParseState`, `handleFile` (parse → score → telemetry), `reset`. Returns `{ state, handleFile, reset }`. `App` becomes layout-only: drop zone, result region, JD card, footer. `formatBytes` moves to a small `lib/` util (or into the hook).
   - Note: if the edit-loop issue (option A) lands, override state + recompute live here too — coordinate so this hook is the single orchestration owner.
2. **Decompose `Result.tsx`:**
   - `src/components/features/EvidencePanel.tsx` — the Source-PDF + extracted-text two-pane block (currently L108-128).
   - `src/components/features/AtsScoreReadout.tsx` — the score readout incl. `Dimension` (L161-316).
   - `src/components/features/LayoutFlagsList.tsx` — (L133-159).
   - `Result.tsx` keeps only the `isFontsUnmappable` branch + `ParsedCard`/`LimitedParsingCard` composition. `StatusPill` is removed in favor of `shared/StatusBadge` (primitives issue).
3. Pure move/extract — **no behavior change**. Each extracted piece keeps the same props it reads today.

## Acceptance criteria

- [ ] `App.tsx` no longer contains the `ParseState` machine or `handleFile` body — it consumes `useResumeAnalysis`.
[...]

PR #149, opened 2026-06-23
feat(webllm): rewrite-quality eval harness (#65)

### LINKED ISSUE #65: Phase 3: Rewrite-quality eval harness (model + prompt comparison)
## Summary

Build a small, deterministic **eval harness** that scores rewrite quality across models and prompt variants, so the *default* model (and prompt) is chosen from measurement rather than vibes. A user-facing picker (Phase 2) yields anecdote; this gives the team a repeatable number to set `DEFAULT_MODEL_ID`.

Phase 3 of the epic. Depends on Phase 1 (the section-rewrite fn + number-preservation check are the unit under test) and pairs with Phase 2 (the registry enumerates the models to compare).

## Motivation

Issue history already flags that any model swap should be **measurement-backed** (see `MODEL_ID` comment in `web-llm.ts` referencing the original WebLLM pilot scope). "Gemma feels better" / "Qwen feels better" is not a basis for the shipped default. The rewrite task is narrow and checkable, so most of the rubric can be computed deterministically without a judge model.

## Design

A Node/Vitest harness (runs in CI's Node env, no WebGPU) that, for each `(model, prompt-variant, fixture)` triple, runs the section-rewrite and scores the output against a rubric.

1. **Fixture set** — `tests/fixtures/rewrite/` — synthetic résumé sections (arrays of bullets) spanning weak/strong/numeric/redundant cases. **PII policy applies** (synthetic personas only, `@example.com`, `555` phones) per the repo's non-negotiable fixture rule — these are bullet-text fixtures, not PDFs, but keep them persona-clean.
2. **Rubric (mostly deterministic, model-free):**
   - **Numbers preserved** — reuse `checkNumbersPreserved` from Phase 1 (hard signal; a drop/invent is a fail).
   - **One line per bullet** — no embedded newlines/wrapping after parse.
   - **Action-verb lead** — first token of each bullet is a strong verb (reuse the verb list already used by the scorer's specificity dimension if one exists; otherwise a curated list).
   - **Length sanity** — each bullet within a char band (not truncated, not bloated).
   - **No preamble leakage** — output contains none of the prompt's instruction phrases / "Rewritten:" echoes.
   - **Dedup effectiveness** — for fixtures with deliberately redundant bullets, output bullet count < input count.
   - (Optional, gated) **LLM-judge coherence** — a separate larger judge model behind a flag for a soft "is this actually better" score; never required for CI.
3. **Runner** — iterate the Phase-2 `MODELS` registry × prompt variants × fixtures, emit a scored table (model, variant, per-criterion pass rate, aggregate). Output a committed `*.json`/markdown report artifact. Because real model inference needs WebGPU, the harness runs against models in whatever environment can host them (local dev with WebGPU, or a stubbed-engine smoke mode in CI that exercises the **scoring** path on canned outputs). Keep the model-inference leg out of the default `npm run test` so CI stays fast and headless; expose it as a separate script (e.g. `npm run eval:rewrite`).

## Acceptance criteria

- [ ] `tests/fixtures/rewrite/` holds synthetic, PII-clean résumé-section fixtures covering weak / strong / numeric / redundant cases.
- [ ] A deterministic rubric scores each output on: numbers-preserved, one-line, action-verb lead, length sanity, no-preamble-leak, dedup-effectiveness — all without a judge model.
- [ ] A runner iterates `(model × prompt-variant × fixture)` and emits a scored comparison report (committed artifact).
- [ ] The scoring path has unit tests on canned good/bad outputs (no model needed) and runs in CI; the model-inference leg is a separate non-CI script (`npm run eval:rewrite`).
- [ ] Optional LLM-judge coherence score is flag-gated and off by default.
- [ ] Report output is human-readable enough to justify a `DEFAULT_MODEL_ID` choice.

## Out of scope

- Wiring eval results back into the app automatically — a human reads the report and sets the default.
- The picker UI (Phase 2) and section-rewrite mechanics (Phase 1).

Vaishnavi1709:
## Summary

Phase 3 of the in-browser AI rewrite epic. Adds a deterministic eval harness that scores section-rewrite outputs across (model × prompt variant × fixture) so the default model and prompt are picked from measurement rather than vibes.

- **Six deterministic rubric criteria** — numbers preserved (reuses `checkNumbersPreserved` from Phase 1), one line per bullet, action-verb lead, length sanity, no-preamble leakage, dedup effectiveness. All computed without a judge model.
- **Two execution legs:**
  - **Scoring** — pure logic + **31 unit tests** under `src/lib/webllm/eval/*.test.ts`; runs in default CI via `npm run test`.
  - **Inference** — dev-only `eval-rewrite.html` page driven by `npm run eval:rewrite`, **one model per tab** to avoid the WebGPU eviction-then-reload crash path on consumer GPUs.
- **DRYed verb list** — exports `ACTION_VERBS` from `src/lib/score/score.ts` so the eval extends the scorer's set instead of duplicating it.
- **LLM-judge flag-gated and off by default** — slot exists (`judgeEnabled` plumbed through runner + report), implementation deliberately a follow-up.

## Findings from the committed reports

All three `MODEL_REGISTRY` models ran against all four fixtures across three prompt variants. Aggregate best-variant scores:

| Model | Best variant | Aggregate |
|---|---|---|
| **Qwen 2.5 (1.5B)** | Baseline = Terse | **67%** 🏆 |
| Gemma 2 (2B) | Terse | 58% |
| Llama 3.2 (3B) | Baseline | 58% |

**The shipped `DEFAULT_MODEL_ID = Qwen2.5-1.5B-Instruct-q4f16_1-MLC` is confirmed.** Bigger ≠ better here — Llama and Gemma both hallucinate more metrics and mangle number notation.

## Follow-up bugs the eval surfaced (filed as separate issues)

These are real but **out of scope for this PR**:

- `cleanRewriteLine` doesn't strip Llama's `"Here are the rewritten bullets:"` preamble line.
- The `noPreambleLeak` rubric check has a blind spot when the preamble survives as a bullet (the bullet-stripping step erases it from the raw-text scan).
- `cleanRewriteLine` doesn't strip inline `**verb**` markdown bold delimiters that Gemma emits on individual tokens (the existing regex only handles whole-bullet wraps).

## Architecture notes

- **Engine-agnostic runner** — `RewriteFn` injection lets tests use stub outputs while the browser entry uses real WebLLM engines.
- **No production-bundle impact** — `eval-rewrite.html` is not in `build.rollupOptions.input`; the eval code is only reachable from that dev-only entry. Verified: `dist/` after `npm run build` contains no eval artifacts.
- **Eval skips analytics** — calls `engine.chat.completions.create` directly instead of `rewriteSectionWithLlm` so a local benchmark doesn't pollute `webllm_section_rewrite_*` counters.
- **Failure handling** — a thrown `RewriteFn` records an error on the cell and scoring continues; the row scores 0 so failures surface in the report instead of silently disappearing.

## Test plan

- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] `npm run test` — **704/704** passing (was 673 on main; +31 new tests under `src/lib/webllm/eval/`)
- [x] `npm run build` succeeds; `dist/` contains no eval code
- [x] `npm run eval:rewrite` opens the dev page, picks a model, runs all 12 cells, downloads JSON + Markdown reports — verified end-to-end against all three registry models on a real WebGPU device
- [x] Reports under `tests/fixtures/rewrite/reports/` are valid UTF-8 (mojibake in earlier paste-throughs was a clipboard artifact only — `file` confirms `Unicode text, UTF-8 text`)

Closes #65.

### DIFF
--- a/src/lib/webllm/eval/rubric.ts
+++ b/src/lib/webllm/eval/rubric.ts
@@ -1,0 +1,164 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+import { checkNumbersPreserved } from "../preserve-numbers.ts";
+import type {
+  FixtureKind,
+  PerBulletDiagnostic,
+  RawRewriteOutput,
+  RubricResult,
+} from "./types.ts";
+import { startsWithActionVerb } from "./verbs.ts";
+
+/**
+ * Deterministic rubric: takes the input bullets + a model's raw rewrite
+ * output and emits a per-criterion pass/fail record. No judge model.
+ *
+ * The six criteria are the issue #65 AC list:
+ *
+ *   1. numbersPreserved   — multiset of numeric tokens unchanged
+ *   2. oneLinePerBullet   — no embedded `\n` after the runner's split
+ *   3. actionVerbLead     — first token of each bullet in the curated set
+ *   4. lengthSanity       — each bullet in a sane char band
+ *   5. noPreambleLeak     — output doesn't echo prompt scaffolding
+ *   6. dedupEffective     — for `redundant` fixtures only: output < input
+ *
+ * Each criterion is computed independently — one failing does NOT
+ * short-circuit the others, because the report's per-criterion pass rate
+ * is more useful than a single composite verdict.
+ */
+
+/** Sanity band for a single bullet. Below MIN reads as a truncated
+ * fragment; above MAX reads as a run-on or multi-bullet collapse. The
+ * band intentionally covers strong real-world bullets (most of which
+ * land in the 60–180 range). */
+const BULLET_MIN_CHARS = 25;
+const BULLET_MAX_CHARS = 260;
+
+/**
+ * Phrases that indicate the model echoed prompt scaffolding into the
+ * output. Cleaned by `cleanRewriteLine` already, but the rubric also
+ * scans the RAW pre-split output to catch leakage that survived (e.g.
+ * spread across multiple lines, or with non-standard capitalization).
+ *
+ * The check is substring (case-insensitive) against the raw response
+ * with the bullet lines stripped, so a legitimate bullet that contains
+ * "the rules of engagement" doesn't trip the criterion.
+ */
+const PREAMBLE_LEAK_PHRASES = [
+  "rewritten bullets:",
+  "original bullets:",
+  "here are the rewritten",
+  "here is the rewritten",
+  "rules:",
+  "system:",
+  "as an ai",
+  "as a language model",
+];
+
+/**
+ * Returns the empty rubric used for an error row (RewriteFn threw, or
+ * returned an unparseable response). All criteria fail so the row
+ * surfaces in the report instead of being silently scored as a pass.
+ */
+export function emptyRubricForError(): RubricResult {
+  return {
+    numbersPreserved: false,
+    oneLinePerBullet: false,
+    actionVerbLead: false,
+    lengthSanity: false,
+    noPreambleLeak: false,
+    dedupEffective: null,
+    judgeCoherence: null,
+    perBullet: [],
+    droppedNumbers: [],
+    addedNumbers: [],
+  };
+}
+
+export interface ScoreRubricInput {
+  input: readonly string[];
+  output: RawRewriteOutput;
+  fixtureKind: FixtureKind;
+}
+
+export function scoreRubric({
+  input,
+  output,
+  fixtureKind,
+}: ScoreRubricInput): RubricResult {
+  const outputBullets = output.bullets;
+
+  // ── (1) Numbers preserved ─────────────────────────────────────────────
+  const preservation = checkNumbersPreserved(input, outputBullets);
+
+  // ── (2) One line per bullet ───────────────────────────────────────────
+  // The runner already split on `\n`, so an embedded `\n` here would
+  // only appear if the post-process kept a literal `\n` token (e.g. a
+  // Windows `\r` survived). Explicit check on each bullet keeps the
+  // criterion honest if the splitting strategy changes.
+  //
+  // Empty output is NOT vacuously a pass: the model produced no bullets
+  // at all, so the "every bullet is one line" claim has nothing to back
+  // it. Require at least one bullet for the criterion to be true.
+  const oneLinePerBullet =
+    outputBullets.length > 0 && outputBullets.every((b) => !/[\r\n]/.test(b));
+
+  // ── (3) Action-verb lead ──────────────────────────────────────────────
+  const verbResults = outputBullets.map((b) => startsWithActionVerb(b));
+  const actionVerbLead =
+    outputBullets.length > 0 && verbResults.every((v) => v);
+
+  // ── (4) Length sanity ─────────────────────────────────────────────────
+  const lengthResults = outputBullets.map(
+    (b) => b.length >= BULLET_MIN_CHARS && b.length <= BULLET_MAX_CHARS,
+  );
+  const lengthSanity =
+    outputBullets.length > 0 && lengthResults.every((v) => v);
+
+  // ── (5) No preamble leakage ───────────────────────────────────────────
+  // Scan the RAW response (pre-split) with the bullet text stripped out
+  // so a phrase like "rules:" inside a legitimate bullet doesn't trip.
+  // Lowercased substring match; the phrase list is conservative.
+  let rawMinusBullets = output.raw.toLowerCase();
+  for (const b of outputBullets) {
+    rawMinusBullets = rawMinusBullets.replace(b.toLowerCase(), "");
+  }
+  const noPreambleLeak = !PREAMBLE_LEAK_PHRASES.some((p) =>
+    rawMinusBullets.includes(p),
+  );
+
+  // ── (6) Dedup effectiveness ───────────────────────────────────────────
+  // Only meaningful for fixtures that explicitly stage redundancy. For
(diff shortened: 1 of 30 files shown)


src/lib/webllm/eval/verbs.ts
```diff
@@ -0,0 +1,66 @@
...
+  // Cross-discipline (research / ops / comms) IC verbs.
+  "completed", "conducted", "drafted", "identified", "owned",
+  "performed", "planned", "presented", "produced", "published",
+  "secured", "tracked",
+  // Present-progressive forms small models sometimes emit.
+  "building", "driving", "leading", "managing", "designing",
+  "shipping", "scaling", "owning",
+];
+
+export const ACTION_VERBS: ReadonlySet<string> = new Set([
```
s-annam: [Suggestion]: This `export` is never imported by another module — only `startsWithActionVerb` (below) is consumed externally. Dropping `export` here keeps `ACTION_VERBS` module-internal and clears both fallow alerts at once (#83 never-imported + #85 duplicate-name-across-modules with `score.ts`). One-word fix that turns the red `fallow` check green.

src/lib/webllm/eval/rubric.ts
```diff
@@ -0,0 +1,164 @@
...
+  const lengthSanity =
+    outputBullets.length > 0 && lengthResults.every((v) => v);
+
+  // ── (5) No preamble leakage ───────────────────────────────────────────
+  // Scan the RAW response (pre-split) with the bullet text stripped out
+  // so a phrase like "rules:" inside a legitimate bullet doesn't trip.
+  // Lowercased substring match; the phrase list is conservative.
+  let rawMinusBullets = output.raw.toLowerCase();
+  for (const b of outputBullets) {
+    rawMinusBullets = rawMinusBullets.replace(b.toLowerCase(), "");
```
s-annam, line 125: [Nit]: `String.prototype.replace` with a string arg strips only the *first* occurrence — if a bullet's text repeats in `raw`, later copies survive in `rawMinusBullets`. Edge-case-only, and adjacent to the `noPreambleLeak`-as-bullet blind spot you already filed as a follow-up. Fine to leave; flagging for the record.

s-annam (commented):
## PR Review: feat(webllm): rewrite-quality eval harness (#149)

### Summary
Clean, exceptionally well-documented Phase 3 eval harness. Engine-agnostic runner + deterministic 6-criterion rubric + snapshot-tested report renderer, with the inference leg quarantined to a dev-only HTML entry so zero eval code reaches the production bundle. 31 new tests, all green. Closes #65.

### Gate results (on checked-out PR branch)
| Gate | Result |
|---|---|
| `npm run typecheck` | ✅ pass |
| `npm run lint` | ✅ pass |
| `npm run test` | ⚠️ 702/704 — 2 failures in `useModelSelection.integration.test.tsx` (`localStorage.clear is not a function`) |
| CI `verify` | ✅ **pass** (55s) |
| CI `fallow` | ❌ red (report-only, non-blocking) |

**The 2 local test failures are NOT this PR's defect.** They reproduce identically on the #143 base branch this PR is stacked on (the test file comes from #143, untouched here), they're green in CI's `verify`, and the cause is a local jsdom/node `localStorage` env quirk. No action for #149.

### Highlights
- `RewriteFn` injection seam cleanly splits the Node-testable scoring leg from the WebGPU inference leg — `runner.ts` never imports `@mlc-ai/web-llm`.
- Sound non-vacuous guards throughout the rubric: empty output fails `oneLinePerBullet`/`actionVerbLead`, and zero-bullet output can't trivially "win" dedup (`outputBullets.length > 0 && < input.length`).
- Error rows score 0 and surface in the report rather than silently inflating an aggregate — genuinely defensive.
- Fixtures are synthetic, PII-clean (verified per repo policy).
- DRY done right: `ACTION_VERBS` anchored in `score.ts`, eval extends it.

### Key Findings (no blockers)
1. **[Suggestion]** `verbs.ts:43` — `export const ACTION_VERBS` is never imported by any other module (only `startsWithActionVerb` is consumed). Dropping the `export` keeps it module-internal and clears two fallow alerts at once (#83 never-imported + #85 duplicate-name-across-modules). Matches the team's standing practice of clearing fallow dead-export flags.
2. **[Nit]** `rubric.ts:125` — `replace(b.toLowerCase(), "")` strips only the first occurrence of each bullet. Edge-case-only and adjacent to the `noPreambleLeak`-as-bullet blind spot already filed as a follow-up; fine to leave.
3. **[Nit]** `report.ts:26` (`renderMarkdownReport` cognitive complexity 22) and `run-eval-browser.ts:199` (CRAP 30) — fallow report-only flags on a snapshot-tested renderer and a dev-only entry. Acceptable as-is.

### Verdict
**Action: COMMENT** — No blocking items. CI `verify` is green; the local test failures are inherited env noise, not a regression. Two small cleanups (esp. #1, a one-word fix that clears the red `fallow` check) before merge would be nice but aren't gating.

src/lib/webllm/eval/verbs.ts
s-annam: Addressed in ac25f72 — `export` dropped, `ACTION_VERBS` now module-internal. Cleared fallow #83/#84/#85 and the `fallow` check is green. Resolving.

s-annam (approved):
Re-reviewed at 0a5e390. Suggestion fixed (ac25f72, fallow now green), only the rubric.ts nit left non-blocking. CI verify + fallow green, typecheck/lint clean, 704/704 tests, build clean with no eval in dist. LGTM — shipping.

PR #154, opened 2026-06-23
fix(webllm): strip Llama preamble + leading-verb bold, close rubric blind spot (#150, #151, #152)

### LINKED ISSUE #150: fix(webllm): cleanRewriteLine should strip "Here are the rewritten bullets:" preamble line
## Surfaced by

The eval harness from #65 (PR #149). Llama 3.2 (3B) consistently emits a leading `"Here are the rewritten bullets:"` line under the **terse** and **examples-led** prompt variants. Example from `tests/fixtures/rewrite/reports/eval-rewrite-llama-3-2-3b-instruct-…json`:

```json
"perBullet": [
  { "index": 0, "text": "Here are the rewritten bullets:", "startsWithActionVerb": false, ... },
  { "index": 1, "text": "Spearheaded marketing task management, ...", ... },
  ...
]
```

The preamble survives `cleanRewriteLine` and ends up as `outputBullets[0]`. That inflates the output bullet count by 1, fails the action-verb check on cell 0, and dilutes every per-criterion aggregate for Llama.

## Root cause

`src/lib/webllm/post-process.ts` has a `PROMPT_ECHO_LINES` set that drops scaffolding lines like `"rewritten bullets:"` and `"rules:"`. Llama's exact phrase `"Here are the rewritten bullets:"` is not in that set, so the line passes through as a normal bullet.

## Fix

Two reasonable shapes:

1. **Substring match** — change the check from set-membership to substring-contains for a short list of known preamble openers (`"here are the rewritten"`, `"here is the rewritten"`, `"rewritten bullets:"`, `"original bullets:"`).
2. **Pattern match** — regex like `/^here (?:are|is) (?:the )?rewritten/i` plus the existing exact-match set.

Either works. (1) is simpler; (2) is more precise. Either way, add a unit test in `post-process.test.ts` that asserts the line is dropped.

## Tracking

- Surfaced by the eval reports committed in PR #149.
- Related: #65 (the eval harness itself).

### LINKED ISSUE #151: fix(eval): noPreambleLeak rubric has a blind spot when the preamble survives as a bullet
## Surfaced by

The eval harness from #65 (PR #149). The `noPreambleLeak` criterion in `src/lib/webllm/eval/rubric.ts` is reporting **100% pass** on Llama runs that visibly leaked the preamble `"Here are the rewritten bullets:"` as the first bullet — that's a false positive.

## Root cause

The current implementation (`rubric.ts`):

```ts
let rawMinusBullets = output.raw.toLowerCase();
for (const b of outputBullets) {
  rawMinusBullets = rawMinusBullets.replace(b.toLowerCase(), "");
}
const noPreambleLeak = !PREAMBLE_LEAK_PHRASES.some((p) =>
  rawMinusBullets.includes(p),
);
```

We strip every output bullet from `raw` before scanning for preamble phrases. The intent was to avoid false positives when a legitimate bullet happens to contain the phrase (e.g. a bullet about "the rules of engagement"). But when the preamble itself becomes a bullet (because `cleanRewriteLine` didn't strip it — see #150), our own bullet-stripping step erases the leaked preamble from the scanned text. The check then passes vacuously.

## Fix

Add a second check: any individual bullet whose text matches a preamble phrase should also fail `noPreambleLeak`. Pseudocode:

```ts
const anyBulletIsPreamble = outputBullets.some((b) =>
  PREAMBLE_LEAK_PHRASES.some((p) => b.toLowerCase().includes(p)),
);
const noPreambleLeak = !anyBulletIsPreamble && !PREAMBLE_LEAK_PHRASES.some((p) =>
  rawMinusBullets.includes(p),
);
```

This makes the rubric robust to both "preamble survived in raw text" and "preamble survived as a bullet" failure modes.

Add a unit test in `rubric.test.ts` that pins this: input with raw `"Here are the rewritten bullets:\nDrove a 4-touchpoint nurture sequence..."` should fail `noPreambleLeak` regardless of whether the preamble was line-split into a bullet or kept inline.

## Tracking

- Surfaced by the Llama report committed in PR #149.
- Related: #150 (the upstream fix in `post-process.ts` — once both land, this blind spot stops mattering in practice, but the rubric is still measuring the wrong thing today).
- Related: #65 (the eval harness itself).

### LINKED ISSUE #152: fix(webllm): cleanRewriteLine should strip inline **verb** bold delimiters from rewrite output
## Surfaced by

The eval harness from #65 (PR #149). Gemma 2 (2B) under the **terse** prompt variant emits per-bullet outputs that bold just the leading verb in markdown, e.g.:

```
**Increased** weekly active users by 1500% (120,000 to 1.8 million)…
**Spearheaded** a growth team of six individuals…
**Streamlined** the checkout process…
```

Example from `tests/fixtures/rewrite/reports/eval-rewrite-gemma-2-2b-it-q4f16-1-mlc-…json`:

```json
{
  "text": "**Triaged** an average of 200+ inbound customer support tickets per week…",
  "startsWithActionVerb": false
}
```

The `**Triaged**` token survives into the bullet text. The first-token check in `verbs.ts::startsWithActionVerb` strips non-`[a-z]` chars after lowercasing, so the verb match itself still works — but the visible bullet contains literal `**` delimiters that a user would see if this output were ever surfaced in product.

## Root cause

`src/lib/webllm/post-process.ts::cleanRewriteLine` strips markdown emphasis only when the **whole bullet** is wrapped:

```ts
const withoutEmphasis = withoutPrefix
  .replace(/^\*\*(.+)\*\*$/s, "$1")
  .replace(/^\*(.+)\*$/s, "$1")
  .replace(/^_(.+)_$/s, "$1");
```

Inline `**verb**` at the start of a longer bullet doesn't match `^\*\*(.+)\*\*$`, so the delimiters pass through.

## Fix

Add a narrow pre-strip that removes `**word**` only when it leads the bullet (a verb-bolding pattern), without touching legitimate mid-line emphasis:

```ts
const withoutLeadingBold = trimmed.replace(/^\*\*([A-Za-z][\w-]*)\*\*\s+/, "$1 ");
```

Match characteristics:
- `^\*\*` — must be at the start
- `([A-Za-z][\w-]*)` — captured word (single token, alphanumeric/hyphen)
- `\*\*\s+` — closing delimiter followed by at least one space (so we only match when there's body text after)

Apply before the existing whole-bullet emphasis strip so the order is: leading-verb-bold strip → whole-bullet emphasis strip → list-marker strip → quote strip.

Add unit tests in `post-process.test.ts`:

```ts
expect(cleanRewriteLine("**Increased** weekly active users by 1500%."))
  .toBe("Increased weekly active users by 1500%.");
expect(cleanRewriteLine("**Streamlined the** checkout"))   // not just one word
  .toBe("**Streamlined the** checkout");                    // unchanged — defensive
```

## Tracking

- Surfaced by the Gemma report committed in PR #149.
- Related: #65 (the eval harness itself).

Vaishnavi1709:
## Summary

Three small fixes surfaced by the rewrite-quality eval committed in #149 (now merged). All three were filed as follow-up issues from that PR; bundled here because they share two files and three days of reviewer time on three separate PRs would be worse than one.

**#150** — `cleanRewriteLine` drops chat-opener preambles like `"Here are the rewritten bullets:"` via a narrow regex (`/^here (?:are|is) (?:the )?(?:rewritten|new|updated)\b/i`). Anchored to start-of-line so a legitimate bullet that mentions "Here" mid-text is unaffected. Caught Llama 3.2 (3B) under the terse + examples-led variants emitting a leading opener that survived cleanup and inflated the output bullet count by one.

**#152** — `cleanRewriteLine` strips a leading `**Verb**` markdown bold when followed by body text (single-word capture by design — multi-word bolds are likely deliberate phrase emphasis). Runs **before** the whole-line emphasis strip, which only matches when the line both starts AND ends with `**`. Caught Gemma 2 (2B) under the terse variant bolding just the leading verb on every bullet.

**#151** — `scoreRubric.noPreambleLeak` now also fails when any output bullet itself contains a preamble phrase. Previously the rubric only scanned `raw - bullets`, so a preamble that survived AS a bullet had its text erased from the scan and the criterion falsely reported a pass. With #150 landing here, upstream cleanup also catches this — but the rubric was measuring the wrong thing regardless, and stays robust to any future preamble shape that slips past `cleanRewriteLine`.

## What changed

| File | Why |
|---|---|
| `src/lib/webllm/post-process.ts` | Two new patterns + ordering: chat-opener regex (drops the line), leading-bold-word regex (runs before the existing whole-line emphasis strip). |
| `src/lib/webllm/post-process.test.ts` | +11 tests across two `describe` blocks pinning #150 and #152 — including negative cases (legitimate bullet starting with "Here", multi-word bold left alone, mid-bullet emphasis untouched). |
| `src/lib/webllm/eval/rubric.ts` | `noPreambleLeak` now requires BOTH `!anyBulletIsPreamble` AND the existing raw-minus-bullets check. |
| `src/lib/webllm/eval/rubric.test.ts` | +2 tests pinning #151 — preamble-as-bullet with matching raw text, and the synthetic edge case where raw text is empty (per-bullet check must catch it independently). |

## Test plan

- [x] `npm run typecheck` passes
- [x] `npm run lint` passes
- [x] `npm run test` — **716/716** passing (was 704 on main post-#149; +12 new tests across the two `.test.ts` files)
- [x] `npm run build` succeeds; no production-bundle impact (these are pure-logic fixes)

## Closes

- Closes #150
- Closes #151
- Closes #152

### DIFF
--- a/src/lib/webllm/eval/rubric.ts
+++ b/src/lib/webllm/eval/rubric.ts
@@ -117,16 +117,37 @@ export function scoreRubric({
     outputBullets.length > 0 && lengthResults.every((v) => v);
 
   // ── (5) No preamble leakage ───────────────────────────────────────────
-  // Scan the RAW response (pre-split) with the bullet text stripped out
-  // so a phrase like "rules:" inside a legitimate bullet doesn't trip.
-  // Lowercased substring match; the phrase list is conservative.
+  // Two failure modes; we have to catch both.
+  //
+  // (a) Preamble survived in the raw response between/around bullets.
+  //     Scan the RAW response with the bullet text stripped so a phrase
+  //     like "rules:" embedded in a legitimate bullet doesn't trip.
+  //
+  // (b) Preamble survived AS a bullet — e.g. Llama emitting
+  //     "Here are the rewritten bullets:" as its own line, which the
+  //     line-splitter then treats as outputBullets[0]. Without an
+  //     explicit per-bullet check, the raw-minus-bullets scan in (a)
+  //     would erase the leaked preamble from rawMinusBullets and report
+  //     a false-positive pass.
+  //
+  // The per-bullet check uses `startsWith`, not `includes`: a preamble
+  // is always at the START of a line by shape (it's what the model
+  // emits BEFORE the real bullets). `includes` would false-positive on
+  // legitimate bullets like "Worked as an AI engineer on …" (which
+  // contains the "as an ai" phrase mid-bullet).
+  //
+  // Fix for #151. Both checks must pass for `noPreambleLeak` to be true.
   let rawMinusBullets = output.raw.toLowerCase();
   for (const b of outputBullets) {
     rawMinusBullets = rawMinusBullets.replace(b.toLowerCase(), "");
   }
-  const noPreambleLeak = !PREAMBLE_LEAK_PHRASES.some((p) =>
-    rawMinusBullets.includes(p),
-  );
+  const anyBulletIsPreamble = outputBullets.some((b) => {
+    const lower = b.trim().toLowerCase();
+    return PREAMBLE_LEAK_PHRASES.some((p) => lower.startsWith(p));
+  });
+  const noPreambleLeak =
+    !anyBulletIsPreamble &&
+    !PREAMBLE_LEAK_PHRASES.some((p) => rawMinusBullets.includes(p));
 
   // ── (6) Dedup effectiveness ───────────────────────────────────────────
   // Only meaningful for fixtures that explicitly stage redundancy. For
--- a/src/lib/webllm/eval/rubric.test.ts
+++ b/src/lib/webllm/eval/rubric.test.ts
@@ -126,6 +126,53 @@ describe("scoreRubric — canned bad outputs", () => {
     expect(r.noPreambleLeak).toBe(true);
   });
 
+  // ── #151: preamble-as-bullet blind spot ─────────────────────────────
+  // Llama 3.2 (3B) emits "Here are the rewritten bullets:" as its own
+  // line; the runner's line-splitter then makes it `outputBullets[0]`.
+  // Before #151, the raw-minus-bullets scan would erase the leaked
+  // preamble (because the leaked preamble IS one of the bullets) and the
+  // rubric would falsely report `noPreambleLeak = true`. Pin both shapes
+  // so a regression here is caught immediately.
+  it("flags a preamble line that survived as a bullet (#151)", () => {
+    const input = ["Worked on marketing."];
+    const output = out(
+      [
+        "Here are the rewritten bullets:",
+        "Drove a 4-touchpoint nurture sequence that lifted MQL conversion 12%.",
+      ],
+      "Here are the rewritten bullets:\nDrove a 4-touchpoint nurture sequence that lifted MQL conversion 12%.",
+    );
+    const r = scoreRubric({ input, output, fixtureKind: "weak" });
+    expect(r.noPreambleLeak).toBe(false);
+  });
+
+  it("flags a preamble-as-bullet even when raw text is empty (#151)", () => {
+    // Synthetic edge case: a test-supplied output that splits "weirdly"
+    // and produces a preamble bullet without any matching raw text. The
+    // per-bullet check must still catch it independently of the raw scan.
+    const input = ["A"];
+    const output = out(
+      ["Here is the rewritten bullets:", "Shipped a real bullet about something concrete."],
+      "",
+    );
+    const r = scoreRubric({ input, output, fixtureKind: "weak" });
+    expect(r.noPreambleLeak).toBe(false);
+  });
+
+  it("does NOT flag a legitimate bullet that contains a preamble phrase mid-text (#151)", () => {
+    // Defensive: the per-bullet check uses `startsWith`, not `includes`,
+    // so a bullet that mentions one of the leak phrases inside its body
+    // shouldn't trip the rubric. Without this anchor, "Worked as an AI
+    // engineer …" would false-positive on the "as an ai" phrase.
+    const input = ["W"];
+    const output = out([
+      "Worked as an AI engineer on the recommendation model pipeline for 3 teams.",
+      "Built the alerting system: 12 services, 99.9% uptime over the last 2 quarters.",
+    ]);
+    const r = scoreRubric({ input, output, fixtureKind: "weak" });
+    expect(r.noPreambleLeak).toBe(true);
+  });
+
   it("flags a redundant fixture whose output did NOT collapse", () => {
     const input = ["A", "B", "C"];
     const output = out([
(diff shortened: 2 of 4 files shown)


src/lib/webllm/eval/rubric.ts
```diff
@@ -117,16 +117,37 @@ export function scoreRubric({
...
+  //     a false-positive pass.
+  //
+  // The per-bullet check uses `startsWith`, not `includes`: a preamble
+  // is always at the START of a line by shape (it's what the model
+  // emits BEFORE the real bullets). `includes` would false-positive on
+  // legitimate bullets like "Worked as an AI engineer on …" (which
+  // contains the "as an ai" phrase mid-bullet).
+  //
+  // Fix for #151. Both checks must pass for `noPreambleLeak` to be true.
   let rawMinusBullets = output.raw.toLowerCase();
```
s-annam, line 140: [Suggestion]: `anyBulletIsPreamble` runs the **full** `PREAMBLE_LEAK_PHRASES` list (incl. `"rules:"`, `"system:"`) as `includes()` against whole bullet bodies. The (a) raw scan deliberately strips bullet text to avoid exactly this — see the comment a few lines up: *"so a phrase like 'rules:' embedded in a legit bullet doesn't trip."* A real bullet like `"Authored escalation rules: scope, cadence."` would now flag `noPreambleLeak=false`. #151's intent is *the whole bullet IS the preamble*, so an anchored `startsWith`, or restricting this per-bullet check to the chat-opener phrases only, would be tighter. Low impact (eval-harness criterion, not production scoring) — non-blocking.

s-annam (approved):
## PR Review: fix(webllm): strip Llama preamble + leading-verb bold, close rubric blind spot (#150, #151, #152)

### Summary
Three well-scoped post-processing fixes from the rewrite-quality eval (#149), bundled cleanly. All gates green on the checked-out branch.

### Verification
- `npm run typecheck` ✅
- `npm run lint` (ESLint) ✅
- `npm run test` ✅ **716/716** (matches PR claim)

### Spec Alignment
| Issue | Status | Notes |
|---|---|---|
| #150 chat-opener strip | ✅ Implemented | `CHAT_OPENER_PATTERN` anchored `^`, runs first, returns `""`. Negative test pins legit "Here,…" bullet. |
| #152 leading `**Verb**` bold | ✅ Implemented | Single-word capture, runs before whole-line emphasis strip. Multi-word + mid-bullet negatives pinned. |
| #151 rubric blind spot | ✅ Implemented | Per-bullet preamble check added alongside the raw-minus-bullets scan. |

### Highlights
- Transform ordering is reasoned and commented: chat-opener before any normalization; leading-bold before the whole-line emphasis strip (which requires a trailing `**$`). Verified the fall-through case `**Shipped Foo.**` → whole-line rule.
- Strong negative-test discipline (legit "Here" bullet, multi-word bold, mid-bullet emphasis all left untouched).
- Bundling justification is sound: shared files, no logic overlap.

### Key Findings
1. **[Suggestion]** `rubric.ts` per-bullet preamble check — see inline.

### Verdict
**Action:** APPROVE
**Rationale:** Zero blocking items. Three issues correctly fixed, tested, all gates green. The one suggestion is a narrow false-positive edge confined to the eval-harness rubric (not production scoring), so it does not gate merge.

PR #157, opened 2026-06-24
feat(score): validity-aware phone completeness credit

### LINKED ISSUE #70: Validity-aware phone scoring (deferred — score-version bump + snapshot re-bake)
## Problem

Phone contributes to the score on **presence only** (`src/lib/score/score.ts:272` — field filled or not). After Issues A+B we know whether a number is *valid* (`libphonenumber-js` `isValid()`), but a garbage digit-run that happens to match scores the same as a real, validated number. Completeness should reward a genuinely parseable phone over a coincidental match.

## Approach (deferred — deliberate later call)

Make the phone dimension validity-aware with three tiers of credit:

- **Validated phone** (`normalizePhone(...).isValid === true`) → **full credit**
- **Parsed-but-invalid phone** (raw-match-only, `isValid === false`) → **partial credit** (~0.5× the phone weight; exact factor TBD in implementation, but must be > 0 and < full)
- **Absent phone** (no match at all) → **zero**

Ordering: valid > raw-match-only > absent.

**Cost this issue exists to make explicit:**
- Changes score semantics → **bump `ATS_SCORE_ALGO_VERSION`**.
- **Re-bakes every `*.expected.json`** whose Completeness shifts (`npm run bake-fixtures`).

Because of the snapshot/version churn, this is intentionally **separate and deferred** from A/B so the core improvement ships without score noise.

## Depends on

Blocked by Issue A (needs `isValid` from the normalizer).

## Affected files

- `src/lib/score/score.ts` — validity-aware phone credit + `ATS_SCORE_ALGO_VERSION` bump.
- `src/lib/score/score.test.ts` — updated expectations.
- `tests/fixtures/pdfs/**/**.expected.json` — re-baked where Completeness changes.

## Acceptance criteria

- [ ] Phone score reflects validity tiers: validated phone = full credit; parsed-but-invalid (raw-match-only) phone = partial credit (~0.5× phone weight, exact factor TBD); absent phone = zero.
- [ ] `ATS_SCORE_ALGO_VERSION` bumped; rationale in the changelog/commit.
- [ ] All affected `expected.json` re-baked + committed.
- [ ] `npm run test` green.

## Notes

Part of the phone-handling epic. **Optional / deferred** — open for tracking; do not start until A+B land and the score-version bump is wanted.

s-annam:
## Summary

Plumbs \`libphonenumber\` \`isValid()\` from the extraction layer all the way to the anonymous scorer so a parsed-but-invalid phone earns **half** completeness credit instead of full. Valid phones and absent phones are unchanged.

Three-tier scoring: valid/unknown validity → full credit; `phoneIsValid===false` → `passed: false, credit: 0.5` (still surfaces in \`missing\`); absent/below confidence floor → zero credit. Mirrors the redacted-dates half-credit pattern.

Closes #70

## What changed

**Plumbing** (scorer cannot import `phone.ts` — entry-chunk budget):
- `contact.ts`: add `phoneIsValid?` to `ContactExtractionResult`; `scan()` sets it from `phoneResult.isValid`; `extractContact()` carries it from whichever scan supplied the phone.
- `types.ts`: add `phoneIsValid?` to `HeuristicParsedResume`.
- `openresume.ts`: spread `phoneIsValid` into `parsed` when phone is present.
- `regex-fallback.ts`: set `parsed.phoneIsValid` in the Tier 1.5 phone block.

**Scoring** (`score.ts`):
- Add `phoneIsValid?` to `AnonymousAtsScoreInput.parsed`.
- Add `PHONE_INVALID_CREDIT = 0.5` constant.
- Special-case phone in the contact-field loop with three tiers.
- Bump `ATS_SCORE_ALGO_VERSION` 1.3 → 1.4 + changelog entry.

**Tests**: 4 new `score.test.ts` cases; 25 corpus `expected.json` re-baked (only `algoVersion` and new `phoneIsValid` key changed — no completeness scores shifted, confirming all fixtures use valid 555-0100..0199 numbers per PII policy).

## Test plan

- [ ] `npm run test` green (722/722)

### DIFF
--- a/src/lib/edit/apply-overrides.test.ts
+++ b/src/lib/edit/apply-overrides.test.ts
@@ -82,6 +82,48 @@ describe("applyOverrides", () => {
     expect(out.full_name).toBeUndefined();
   });
 
+  it("clears a stale phoneIsValid flag when the phone is overridden (#70 review)", () => {
+    const parsed: HeuristicParsedResume = {
+      ...baseParsed(),
+      phone: "555-invalid",
+      phoneIsValid: false,
+    };
+    // User fixes the number → the old `false` must not survive, else the
+    // scorer keeps awarding half credit on the corrected phone.
+    const { parsed: out } = applyOverrides(
+      parsed,
+      "raw",
+      makeSections(),
+      { phone: "(312) 555-0123" },
+      {},
+      {},
+      [],
+    );
+    expect(out.phone).toBe("(312) 555-0123");
+    expect(out.phoneIsValid).toBeUndefined();
+    // Original untouched.
+    expect(parsed.phoneIsValid).toBe(false);
+  });
+
+  it("clears a stale phoneIsValid flag when the phone is cleared (#70 review)", () => {
+    const parsed: HeuristicParsedResume = {
+      ...baseParsed(),
+      phone: "555-invalid",
+      phoneIsValid: false,
+    };
+    const { parsed: out } = applyOverrides(
+      parsed,
+      "raw",
+      makeSections(),
+      { phone: "" },
+      {},
+      {},
+      [],
+    );
+    expect(out.phone).toBeUndefined();
+    expect(out.phoneIsValid).toBeUndefined();
+  });
+
   it("replaces experience header fields by index", () => {
     const parsed = baseParsed();
     const { parsed: out } = applyOverrides(
--- a/src/lib/edit/apply-overrides.ts
+++ b/src/lib/edit/apply-overrides.ts
@@ -211,6 +211,11 @@ export function applyOverrides(
     } else {
       nextParsed[key] = ov;
     }
+    // The original `phoneIsValid` flag is now stale — it described the parsed
+    // phone, not the user-supplied one. Drop it so the scorer re-grades the
+    // edited number as validity-unknown (backward-compatible full credit)
+    // instead of carrying the old false → permanent half credit. (#70 review)
+    if (key === "phone") delete nextParsed.phoneIsValid;
   }
 
   // ── Experience headers ──────────────────────────────────────────────────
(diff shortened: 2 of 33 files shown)


rohithgollapalli (commented):
## Review

Plumbing is clean and the scorer change is correct. The 25-fixture rebake touching only `algoVersion` + the new key confirms no score shifts. Two findings:

### 1. `src/lib/edit/apply-overrides.ts:204-214` — editing the phone carries a stale `phoneIsValid`

`nextParsed` is built with `{ ...parsed }` (line 196), copying the original `phoneIsValid`. The contact-override loop overwrites `nextParsed.phone` (line 212) but never refreshes `phoneIsValid`.

Scenario: a resume parses with an invalid phone (`phoneIsValid: false` → half credit, "phone" in `missing`). The user fixes the number in the editor → App.tsx re-grades with the override output → `phone` is the new valid string but `phoneIsValid` is still `false`, so the scorer keeps awarding half credit and keeps listing phone as missing. The correction has no effect on the score.

Fix: when `phone` is overridden, `delete nextParsed.phoneIsValid` (falls back to backward-compatible full credit) — mirroring the empty-string clear path.

### 2. `src/lib/score/score.test.ts:722` — half-credit test's upper bound is too weak

`expect(invalidResult.completeness.score).toBeLessThanOrEqual(validResult.completeness.score)` passes even if invalid phones were granted *full* credit — i.e. the test stays green if the feature breaks in exactly its core direction. Only the `> absentResult` assertion has teeth.

Fix: use `toBeLessThan(...)` (here 29 < 30, so it holds strictly).

Finding #1 is the substantive one — a real gap in the edit re-grade path this PR introduces.

s-annam:
Both review findings addressed in 1c7f5c8:

1. **`apply-overrides.ts` stale `phoneIsValid`** — `applyOverrides` now `delete`s `nextParsed.phoneIsValid` whenever the `phone` field is overridden *or* cleared, so a corrected phone re-grades as validity-unknown (backward-compatible full credit) instead of carrying the old `false`. Added two regression tests (override path + clear path).
2. **Weak upper bound** — `toBeLessThanOrEqual` → `toBeLessThan` on the half-credit case.

Typecheck clean, 724/724 green.

PR #159, opened 2026-06-24
fix: parse prose bodies, redacted dates, role-comma split (#158)

s-annam:
## Summary

Fixes the three defects the `chanchal-sharma-sample.pdf` referred-resume fixture (a Word/Office template with redacted `20XX` placeholder years and glyph-less prose descriptions) surfaced, plus the score-readout render collapse:

- **Render** — `AtsScoreReadout` score row no longer collapses when the completeness hint is long; the 3 dimension cards keep full width.
- **Parser — redacted years** — `DATE_ANCHOR`/`MONTH_YEAR_ANCHOR` admit `20XX` (and `XXXX`/`####` only when month-anchored), so `March 20XX – December 20XX` parses as a date range. All 3 roles now parse with correct title/company/dates.
- **Parser — prose bodies** — bullet-less prose descriptions are captured as the role body (R1 hybrid y-gap split on the PDF path, `isProseLine` fallback for flat-y DOCX/markdown), and a wrapped paragraph folds into one body unit.
- **Parser — R2** — guarded `"Role, Company"` single-line headers split into title + company (guarded against `Acme, Inc` and location tails), making all three roles map consistently.
- **Scoring** — pool per-role descriptions as a fallback bullet source when the accomplishment sections yield no marker bullets (glyph-less templates), mirroring the authed scorer. Specificity/Structure now grade the prose and the bullets render per role. `chanchal-sample` goes 1→3 roles, overall 24→79.
- `cascade.ts` counts the experience `team` field in `countExtractedChars` so the better field mapping doesn't trip the low-extraction-ratio hard-fail.

5 corpus snapshots re-baked — all improvements or negligible drift (chanchal-sample/-bulleted-skills 0→0.9; single-word-name-mononym 0→0.86; weasyprint-two-column 0→0.64; awesome-cv 0.9→0.89).

Closes #158

## Test plan

- [x] `npm run typecheck` clean
- [x] `npm run test` green (718 pass)
- [x] `npm run lint` clean
- [x] Manually verified in `npm run dev` on the target fixture (3 roles, descriptions render as bullets, score breakdown no longer collapses)

PR #161, opened 2026-06-24
fix(heuristics): drop empty-title (date-only) entry blocks before scoring (#145)

### LINKED ISSUE #145: Filter empty-title (date-only) entry blocks before they reach the score
## What

`parseEntryBlocks` can emit an entry whose `headerLines` is empty (`title === ""`), and the achievement / project extractors map it straight through to a `value[]` entry with `title: ""`. No filter drops it.

Surfaced during review of #144 (flat-bullet-list awards fallback), but the behavior is **pre-existing and shared across all three extractors** — not specific to that PR.

## How it happens

A header/anchor line whose entire text is a date token leaves nothing after `stripDateRange`:

- Flat-list path: a bullet like `• 2023` → `stripBullet` → `"2023"` → `parseDateRange` matches the `YEAR_RE` fallback → `stripDateRange` returns `""` → `headerLines: []`.
- Same shape reachable via `buildEntryBlock` for a date-only anchor line.

Downstream, `achievementFromBlock` (`src/lib/heuristics/extract/achievements.ts:58-81`) and the sibling `projectFromBlock` build `{ title: "", year, ... }` with `score 0` and push it into `value[]`. The zero score dilutes section confidence and a title-less item lands in output.

## Expected

A block with empty `headerLines` (no title) is not a real entry — drop it before it reaches the score, rather than emitting `{ title: "" }`.

## Proposed fix

Filter empty-title blocks in the two thin extractors so the fix covers every anchor path at once:

- `extractAchievements` (`src/lib/heuristics/extract/achievements.ts`)
- `extractProjects` (`src/lib/heuristics/extract/projects.ts`)

e.g. drop blocks where `liftHeaderLabel(block.headerLines).label` is empty before mapping/scoring. `extractExperience` should be checked for the same shape.

## Acceptance

- A date-only / title-less block produces **no** entry (not a `{title:""}` entry).
- A new unit test in `entry-blocks`/extractor tests covering a date-only bullet.
- Corpus snapshots re-baked; confirm no real fixture regresses (this should only remove phantom empty entries, if any).

## Notes

Low priority — the trigger (an award/project/role whose only text is a bare year) is an unrealistic résumé shape. Hardening + honest-output cleanup, not a live bug on real input.

rohithgollapalli:
## Summary

A header/anchor line whose entire text is a date leaves nothing after `stripDateRange`, so `parseEntryBlocks` emits a block with empty `headerLines`. All three extractors mapped that straight through to a `{ title: "" }` / `{ name: "" }` entry with score 0 — a title-less item in output that also diluted section confidence via `avgScore`. (Surfaced during #144 review; pre-existing and shared across the extractors.)

Adds a shared `finalizeEntries(built, hasLabel)` helper in `extract/shared.ts` that drops empty-label entries and packages the survivors as the standard `{ value, confidence }` pair — keeping the score-0 phantom out of the `avgScore` denominator. Routes `extractAchievements` / `extractProjects` / `extractExperience` through it (predicates: `title`, `name`, and `title || company` for experience). Filtering on the built entry's label rather than on empty `headerLines` also catches a URL-only header, which `liftHeaderLabel` collapses to an empty label — so every anchor path (`first_line` flat-list, `first_line` buildEntryBlock, `date_range`) is covered at once. Per-block mappers and scoring are unchanged.

Closes #145

## Test plan

- [x] `npm run test` green (723 passed; +5 new `extract-fields` cases: date-only bullet, date-only header, date-only project, date-only experience, and a no-dilution regression guard)
- [x] Corpus re-baked — **no fixture content changed** vs `main` (confirmed via `--ignore-cr-at-eol`); no real résumé carried a phantom empty entry, matching the issue's expectation
- [x] `eslint` clean on changed files

### DIFF
--- a/src/lib/heuristics/extract-fields.test.ts
+++ b/src/lib/heuristics/extract-fields.test.ts
@@ -888,3 +888,53 @@ describe("extractAchievements", () => {
     expect(extractAchievements(undefined)).toEqual({ value: [], confidence: 0 });
   });
 });
+
+// A header/anchor line whose entire text is a date leaves nothing after
+// stripDateRange, so the block carries no title. Such a phantom must be dropped
+// before it reaches the score, not emitted as a `{ title: "" }` entry (#145).
+describe("empty-title (date-only) block filtering (#145)", () => {
+  it("drops a date-only bullet in an achievements flat list", () => {
+    // Flat-list fallback path: a bullet whose only text is a year.
+    expect(extractAchievements(mkSection("achievements", [{ text: "• 2023" }]))).toEqual({
+      value: [],
+      confidence: 0,
+    });
+  });
+
+  it("drops a date-only header in achievements (buildEntryBlock path)", () => {
+    expect(extractAchievements(mkSection("achievements", [{ text: "2023" }]))).toEqual({
+      value: [],
+      confidence: 0,
+    });
+  });
+
+  it("drops a date-only project, leaving no name-less entry", () => {
+    expect(extractProjects(mkSection("projects", [{ text: "• 2023" }]))).toEqual({
+      value: [],
+      confidence: 0,
+    });
+  });
+
+  it("drops a date-only experience block with neither title nor company", () => {
+    expect(extractExperience(mkSection("experience", [{ text: "2020 - 2022" }]))).toEqual({
+      value: [],
+      confidence: 0,
+    });
+  });
+
+  it("keeps the real entry and confidence is not diluted by a dropped phantom", () => {
+    const section = mkSection("experience", [
+      { text: "Acme Corporation" },
+      { text: "Senior Software Engineer" },
+      { text: "Jan 2020 - Mar 2022" },
+      { text: "• Shipped the thing" },
+      { text: "2018 - 2019" }, // date-only phantom anchor, no header above
+    ]);
+    const { value, confidence } = extractExperience(section);
+    expect(value).toHaveLength(1);
+    expect(value[0].company).toBe("Acme Corporation");
+    expect(value[0].title).toBe("Senior Software Engineer");
+    // Phantom (score 0) is gone, so the average reflects only the real entry.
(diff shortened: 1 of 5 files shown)


s-annam (dismissed):
## PR Review: fix(heuristics): drop empty-title (date-only) entry blocks before scoring (#145)

### Summary
Correct root-cause fix. A date-only anchor reduces to an empty label after `stripDateRange`, producing a phantom `{ title: "" }` / `{ name: "" }` score-0 entry that also diluted section confidence via `avgScore`. The new shared `finalizeEntries` helper drops the phantom and packages survivors as the standard `{ value, confidence }` pair.

### Highlights
- DRY: one helper replaces the duplicated `blocks.length === 0 → avgScore(...)` tail across all three entry extractors.
- Filtering on the built entry's label (not on empty `headerLines`) also catches URL-only headers that `liftHeaderLabel` collapses to an empty label — covers every anchor path at once.
- Behavior-preserving for the empty case (empty `built` → `avgScore([])` → 0), so no regression vs the old early return.
- Predicates match the returned label fields (`title`, `name`, `title || company` for experience).

### Verification
- `npm run typecheck` clean
- `npm run test` — 723 passed (+5 new cases incl. no-dilution regression guard)
- Corpus unchanged vs main

### Verdict
**Action:** APPROVE
**Rationale:** No findings. Targeted, well-tested fix.

s-annam (approved):
Re-approve after rebase onto main + conflict resolution in experience.ts (import-line only: kept #159's US_LOCATION_RE/INTL_LOCATION_RE, swapped avgScore→finalizeEntries; avgScore now unused in file). Gates green post-rebase: typecheck clean, 732 tests passing.

s-annam (approved):
Re-approve after updating branch to base (160 merged; disjoint files, clean update — no conflict). Gates green on prior rebase: typecheck clean, 732 tests.

Issue #163, opened 2026-06-24
"Relevant Coursework" (unrecognized header) is swallowed into the previous entry's description instead of terminating the section


s-annam:
## Symptom

A **"Relevant Coursework"** section is not recognized. Worse than a missing label: because the header isn't recognized, the prior section never closes — the **entire coursework block, plus the header text itself, is swallowed into the previous entry's `description`**. The coursework items also leak into the bullet pool as 1–3-word junk bullets that tank Specificity/Structure.

Reported on a real student résumé. PII-free reproducer: `tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-multiline-bullets-coursework.pdf`.

## Root cause

`matchSectionHeader` (`src/lib/heuristics/regex.ts:254`) is **exact-alias** match against `SECTION_KEYWORDS`. `"relevant coursework"` / `"coursework"` is in **no** alias list (`src/lib/heuristics/sections.config.json`) and `"coursework"` is not an anchor-fallback head-noun (#111). So the sectionizer (`splitIntoSections` / `sectionizeMarkdownLines`) neither **opens** a new section nor **closes** the current one — every coursework line is appended to the last experience entry.

## Evidence (from the fixture)

Last experience entry ("Discussion Group Facilitator") `description`:

```
Planned meeting agendas and material for 20+ meetings
Led and moderated discussions, creating a welcoming environment for all participants
Relevant Coursework
Global Dimensions of ● Financial Accounting Microeconomics Business Fundamentals of Macroeconomics
Fundamentals of HR Operational Management ● Legal Environment of Management Business
```

Junk bullets pooled from the swallowed block: `[8] Global Dimensions of ● Financial Accounting`, `[9] Microeconomics`, `[10] Fundamentals of`, `[11] Macroeconomics`, `[12] Fundamentals of HR`.

### Sub-problem 1 — Add `coursework` / `relevant coursework` as recognized aliases

Add `"coursework"` and `"relevant coursework"` to `src/lib/heuristics/sections.config.json`, mapped to the **`education`** section type. This is the right bucket: coursework is an education-related signal, and student résumés frequently list it alongside their degree.

Scoring treatment: coursework lines contribute to **education completeness** (student-persona signal — a populated coursework section is treated as an education entry-level indicator). They must **not** pool into the accomplishment/specificity bullet pool.

### Sub-problem 2 — `isVisualHeader` fallback must fire on these renders (in scope for this issue)

[...]

PR #169, opened 2026-06-24
feat: surface extracted coursework into Education entries

### LINKED ISSUE #168: Surface extracted coursework into Education (data model + UI render)
## Problem

The parser extracts a "Relevant Coursework" block (it lands in an `education`-typed section, the keyword alias added in #163, and de-interleaves correctly after #164), but the coursework items never reach structured `ParsedResume.education` — so the reconstructed-resume UI shows the Education section with degrees only and silently drops every course.

Two root causes:

1. `ResumeEducation` had no field to hold coursework.
2. `extractEducation` discarded every bullet line (`!isBulletLine`) before chunking, so the coursework items (which are bullets) were thrown away.

Reproduced on a real student résumé (San Jose State, B.S. Business Admin): a 3-column coursework grid of 7 courses rendered as **nothing** under Education.

## Fix (this issue)

- **`ResumeEducation.coursework?: string[]`** — new optional field (`src/lib/score/types.ts`).
- **`extractEducation`** (`src/lib/heuristics/extract/education.ts`) now recovers coursework from bullet lines and attaches the list to the primary (first) entry, handling two real-grid wrinkles:
  - **Wrapped cells rejoined** — a de-interleaved cell like `● Global Dimensions of` + a following non-bullet `Business` is joined back into one item (`"Global Dimensions of Business"`), not truncated at the wrap, and the continuation line is excluded from entry detection so it isn't mistaken for an institution.
  - **Title-case guard** — a lowercase prose sub-note bullet (`-including courses taught in Japanese`) is dropped; course titles lead uppercase.
- **`EducationSection`** (`src/components/features/ReconstructedResume.tsx`) renders `Coursework: a · b · c` under the entry.
- 4 unit tests in `extract-fields.test.ts` (collect / wrap-rejoin / sub-note-drop / no-bullets → undefined). Full suite 765 green, typecheck clean.

Result on the repro: all 7 courses surface, wraps intact, sub-note gone.

## Known limitations / follow-ups (not in this fix)

- **Section-level attribution.** Coursework is always attached to the **first** entry. When the "Relevant Coursework" block belongs to a non-primary degree, it's misattributed — `findSection` merges multiple `education` sections via `flatMap`, losing the boundary that says which degree the block followed. A clean fix needs section-boundary-aware attribution or a separate coursework section type.
- **Title-case guard is lossy.** Legitimate lowercase-led course names (e.g. `iOS Development`) would be dropped. Rare; revisit if it bites.
- **Not editable.** Coursework is display-only; not wired to the `EditableField` inline-edit primitive.
- **No dedicated scoring credit.** Coursework contributes only via the raw-text completeness ratio, not as its own signal.

## Lineage

Sequel to #163 (coursework header → `education` alias) and #164 (column de-interleave). Those got the text into the right section; this gets it into the data model and onto the screen.

s-annam:
## Summary

Recover the "Relevant Coursework" block into structured education data and render it in the reconstructed résumé. The block already de-interleaved into the `education` section (#163/#164) but was discarded before reaching `ParsedResume`, so the UI showed degrees only and dropped every course.

- `ResumeEducation.coursework?: string[]`
- `extractEducation` collects bullet coursework onto the primary (first) entry, **rejoins wrapped grid cells** (`● Global Dimensions of` + `Business` → one item), and **drops lowercase prose sub-notes** via a Title-case guard; continuation lines are excluded from entry detection so they aren't mistaken for an institution.
- `EducationSection` renders `Coursework: a · b · c`.

Verified on a real student résumé (San Jose State, 3-column 7-course grid): all 7 surface, wraps intact, sub-note gone.

Known limitations (tracked in #168): coursework always attaches to the first entry; Title-case guard is lossy for lowercase course names; not yet editable; no dedicated scoring credit.

Closes #168

## Test plan

- [x] `npm run test` green (765)
- [x] Manually verified in `npm run dev` / `npm run preview`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,6 +14,7 @@
         "jszip": "^3.10.1",
         "libphonenumber-js": "^1.13.6",
         "mammoth": "^1.12.0",
+        "pdf-lib": "^1.17.1",
         "pdfjs-dist": "^4.10.38",
         "posthog-js": "^1.205.0",
         "react": "^19.1.1",
@@ -1855,6 +1856,24 @@
       "license": "Apache-2.0",
       "engines": {
         "node": ">=14"
+      }
+    },
+    "node_modules/@pdf-lib/standard-fonts": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/@pdf-lib/standard-fonts/-/standard-fonts-1.0.0.tgz",
+      "integrity": "sha512-hU30BK9IUN/su0Mn9VdlVKsWBS6GyhVfqjwl1FjZN4TxP6cCw0jP2w7V3Hf5uX7M0AZJ16vey9yE0ny7Sa59ZA==",
+      "license": "MIT",
+      "dependencies": {
+        "pako": "^1.0.6"
+      }
+    },
+    "node_modules/@pdf-lib/upng": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@pdf-lib/upng/-/upng-1.0.1.tgz",
+      "integrity": "sha512-dQK2FUMQtowVP00mtIksrlZhdFXQZPC+taih1q4CvPZ5vqdxR/LKBaFg0oAfzd1GlHZXXSPdQfzQnt+ViGvEIQ==",
+      "license": "MIT",
+      "dependencies": {
+        "pako": "^1.0.10"
       }
     },
     "node_modules/@pkgjs/parseargs": {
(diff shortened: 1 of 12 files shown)


s-annam:
Closing as resolved. The coursework→Education changes from this branch already landed in `main` via the #173 squash merge (gh-171 was stacked on this branch's commits, so its squash carried the coursework diff into main). I've synced this branch to `main` (merge commit `a20ccfc`) — its diff vs main is now empty, confirming the content shipped. Build green. Closing rather than merging an empty diff; any review feedback can be handled in a fresh follow-up.

Issue #176, opened 2026-06-24
Edit education entries and add/remove skills in the reconstructed resume


s-annam:
## Problem

The reconstructed resume makes contact, role headers, and bullets editable (#58, #82), but **Education and Skills are read-only** — there is no way to fix a mis-parsed degree/institution/date, and no way to add a skill the parser missed or drop one it hallucinated:

- `EducationSection` (`src/components/features/ReconstructedResume.tsx:408-445`) renders degree/institution/dates/coursework as static text.
- `SkillsSection` (`src/components/features/ReconstructedResume.tsx:447-460`) renders `skills.join(" · ")` as static text — no add, no remove.

Since the reconstructed resume is also what gets exported as a PDF (`useDownloadPdf`, `ReconstructedResume.tsx:515`), a parser miss in education or skills is uncorrectable and flows straight into the download.

## Desired behavior (UI/UX)

1. **Editable education entries.** For each education entry, let the user edit degree, institution, and dates inline — the same inline-edit affordance used for role headers. (Coursework editing can be a follow-up; in scope is the degree/institution/date line.)
2. **Add/remove skills.** Render skills as removable chips with an "Add skill" affordance:
   - removing a chip drops that skill,
   - adding accepts a typed skill (ideally with canonical-name suggestions — see Porting note).
3. Edits flow through the existing in-memory override model so the re-grade, score, and PDF export all see the corrected data.

## Where this lives

- `src/components/features/ReconstructedResume.tsx` — `EducationSection` and `SkillsSection` gain edit affordances (reuse `EditableField`; for skills, a chip + add-input pattern).
- `src/hooks/useEditableParse.ts` — extend the override model with **education field overrides** (keyed by education index, mirroring `experienceOverrides`) and a **skills override** (add/remove against `parsed.skills`). Follow the existing `setExperienceField` / `setBulletField` shape so App's `applyOverrides` re-grade picks them up.
- Whatever folds overrides back into the parse for scoring/PDF (the `applyOverrides` path referenced in `useEditableParse.ts` header comment) must include the new education/skills overrides.

### Porting note (skills canonicalization)

The "add skill" input benefits from canonical skill-name suggestions/normalization. **This logic can be ported largely as-is from Recruidea** — there is an existing canonical skills taxonomy we can reuse rather than rebuild. This issue is assigned to @s-annam to handle that port. (No need to specify the Recruidea internals here — the port is an implementation detail of the "Add skill" affordance.)

### Reuse analysis

[...]

Issue #175, opened 2026-06-24
Make role-header field editing (title, company, dates) discoverable and consistent


s-annam:
## Problem

In the reconstructed resume, a role header is **already editable** — `RoleHeader` edit mode exposes `EditableField` rows for **title, company, start date, and end date** (`src/components/features/ReconstructedRole.tsx:331-377`, wired by `ReconstructedResume` at `src/components/features/ReconstructedResume.tsx:558-567` via `experienceOverrides` + `onExperienceFieldChange`). But in practice users don't realize these are editable, for two UI/UX reasons:

1. **Low discoverability.** The edit affordance is a pencil that is `opacity-0` until you hover the exact field (`reserve` reveal mode, `EditableField.tsx:178-182`). With no persistent visual cue that the header is editable, the title/company/dates read as static text.
2. **Cramped, inconsistent edit box.** Each field edits into a one-line `<input>` (`EditableField.tsx:108`) — the same cramped box being upgraded for bullets in the companion issue. Title and company in particular can be long.

The result: the capability exists but feels absent, and it doesn't match the bullet-edit experience.

## Desired behavior (UI/UX)

A single, consistent inline-edit flow for the four role-header fields:

1. **Discoverable affordance.** Make it obvious a role header is editable — e.g. a persistent (not hover-only) edit cue on the header, or an "Edit role" toggle that reveals the field editors. The user must be able to tell at a glance that title/company/dates can be changed.
2. **Consistent editor.** Reuse the same edit-box treatment as the bullet editor (full-width where the field warrants it), so editing a title feels like editing a bullet.
3. **All four fields** (title, company, start date, end date) edit through this one flow, committing individually as today.

This is one cohesive flow over the existing `RoleHeader` edit mode — not new per-field surfaces.

## Where this lives

- `src/components/features/ReconstructedRole.tsx` — `RoleHeader` edit mode (L318–377) already renders the four `EditableField`s. Change is to the **discoverability/affordance** wrapper around them (a visible edit cue or an explicit edit toggle) and to adopt the consistent editor treatment.
- `src/design-system/primitives/EditableField.tsx` — adjust `revealOn` usage / add a more discoverable reveal so the affordance isn't hover-gated; reuse the multi-line/full-width variant from the bullet-editor issue where a field is long.
- `src/hooks/useEditableParse.ts` — `setExperienceField` already backs these edits; no model change expected.

[...]

Issue #174, opened 2026-06-24
Full-width bullet editor with Save + Rework actions (replace cramped single-line input)


s-annam:
## Problem

The reconstructed resume lets a user edit a bullet inline (#82), but the edit affordance is a **single-line `<input type="text">`** — see `EditableField` (`src/design-system/primitives/EditableField.tsx:108`). Bullets are long-form prose (we aim for 8–30 words), so editing one means typing into a one-line box that horizontally scrolls and hides most of the text. Both entry paths land in the same cramped box:

- clicking the bullet text itself (the `revealOn="hover"` value-as-trigger path), and
- clicking the hover-revealed pencil icon.

There is also no explicit **Save** action and no way to **Rework** (AI-rewrite) the bullet from edit mode — commit happens on blur/Enter, which is invisible, and the existing AI rewrite (`RewriteButton`) only appears on *flagged* bullets in read mode, not while editing.

## Desired behavior (UI/UX)

When a user edits a bullet, give them room to see and shape the whole sentence:

1. **Full-width, multi-line editor.** Edit mode for a bullet expands to a full-width auto-growing `<textarea>` (not a one-line input), showing the entire bullet without horizontal scroll.
2. **Explicit actions.** Render a small action row under the editor with **Save** and **Cancel**, plus a **Rework** button that runs the existing AI rewrite on the current draft and offers to replace it. (Enter-to-save / Esc-to-cancel keyboard shortcuts stay as accelerators.)
3. **Both entry points** (bullet-text click and pencil click) open this same full-width editor.

This is a refinement of the existing edit surface, not a new one.

## Where this lives

- `src/design-system/primitives/EditableField.tsx` — the ONE shared inline-edit primitive. Today its edit branch (L105–133) renders `<input type="text">`. Add a **multi-line variant** (e.g. a `multiline?: boolean` prop → renders an auto-growing `<textarea>` and an explicit Save/Cancel action row) so the single primitive grows a variant rather than forking. Keep semantic tokens only (no raw palette / hex) per CLAUDE.md.
- `src/components/features/ReconstructedRole.tsx` — `ResumeBulletRow` (L172–251) is the caller. It currently passes `display="inline"` + `revealOn="hover"` to `EditableField`. Switch the bullet to the multi-line variant and surface the **Rework** action wired to the existing rewrite path.
[...]

PR #180, opened 2026-06-25
fix(resume): make all role-header fields editable + fix PDF download race

s-annam:
## Summary

Two fixes on the reconstructed-resume surface:

1. **Role-header fields were not editable.** `RoleHeader` gated editability on `overrides !== undefined`, but the per-index experience-override map starts empty — so every un-edited role rendered the read-only composite "Title — Company · dates" line instead of inline edit fields. Bullets and Education escaped the bug (neither gates on a pre-existing override). Now gates on the commit handler alone, defaulting a missing override map to `{}` (mirrors `EducationEntry`/`ContactCard`). Title, company, start date, end date are all editable on first interaction.

2. **"Download PDF" silently no-op'd on slower/remote contexts and Firefox/Safari.** `URL.revokeObjectURL` ran synchronously in `finally`, immediately after `a.click()`. `click()` only *schedules* the download; the browser reads the `blob:` URL asynchronously afterward, so revoking that fast invalidated the URL before the fetch began. The URL is now handed off and revoked on a later task; the `finally` path still cleans up when no download started.

Also drops the last "pencil-on-hover" doc comments left over from the quiet-inline edit pivot.

Refs #174 #175 #176

## Test plan

- [x] `npm run test` green (802 tests)
- [x] `npm run lint` clean

### DIFF
--- a/src/components/features/ReconstructedEducationSkills.tsx
+++ b/src/components/features/ReconstructedEducationSkills.tsx
@@ -96,6 +96,12 @@ function EducationEntry({
   const { degree, institution, startDate, endDate, dates, coursework } =
     resolveEducationDisplay(edu, overrides);
 
+  // The editable start/end fields ARE the date display, so the compact `dates`
+  // string would duplicate them. Show it ONLY in the legacy year-only fallback
+  // (no start/end parsed, just a graduation `year`), where no editable field
+  // surfaces it otherwise.
+  const yearOnly = !startDate && !endDate && Boolean(dates);
+
   return (
     <li className="flex flex-col gap-0.5 text-sm">
       <div className="flex flex-wrap items-center gap-x-1.5 gap-y-0.5">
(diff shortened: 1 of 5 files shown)


s-annam:
### Manual verification

Exercised the full surface end-to-end against `npm run dev` (Playwright, real browser), including over a LAN IP on a non-secure-context origin to mirror the remote-access path:

- Dropped a fixture PDF → parsed in ~335 ms, reconstructed resume rendered, no console errors (only the unrelated favicon 404).
- **Role headers:** title, company, start date, end date all show inline edit affordances on first interaction and commit correctly — the regression this PR fixes.
- **Bullets / Contact / Education / Skills:** still editable via the same quiet-inline affordance.
- **Download PDF:** triggers the file download reliably (revoke-after-click race removed).
- Re-confirmed processing works over a non-localhost http origin (non-secure context) — same result as localhost.

Gates: `typecheck` clean, 802 tests green, `eslint` clean.

PR #212, opened 2026-06-26
fix(webllm): close loadEngine TOCTOU race by acquiring inference before await (#148)

### LINKED ISSUE #148: WebLLM: TOCTOU between loadEngine() resolve and acquireInference() lets a concurrent picker switch unload an engine mid-inference
## Summary

`loadEngine` / `acquireInference` have a time-of-check-to-time-of-use gap: an engine can be `.unload()`'d by a concurrent picker switch in the window between `loadEngine()` resolving and the caller calling `acquireInference()`. The in-flight-inference tracker that defers eviction (`inflightInferenceCount` / `pendingUnload` in `src/lib/webllm/web-llm.ts`) is exactly the mechanism meant to prevent tear-down-mid-use, but it has a hole at the boundary it guards.

Introduced in #143 (PR B of the M4 in-browser AI rewrite epic, #64).

## The race

Consumer flow (`src/components/features/RewriteButton.tsx:74-78`, same shape in `SectionRewrite.tsx`):

```ts
const engine = await loadEngine(selectedModelId, onProgress); // fast path: resolved engine A
const rewritten = await rewriteBulletWithLlm(bullet, engine, modelId);
//   ^ acquireInference(modelId) is called INSIDE rewriteBulletWithLlm, AFTER loadEngine resolved
```

`acquireInference` runs at `src/lib/webllm/rewrite-bullet.ts:60` (and `rewrite-section.ts`), inside the rewrite fn — not before the load resolves.

Interleaving:

1. Engine A already loaded. Rewrite caller hits fast-path A: `loadEngine(A)` returns `Promise.resolve(A)`. The `await` yields to the microtask queue.
2. The picker's queued `serialChain` entry (also a microtask) runs `evictAllExcept(B)`. It removes A from `loadedEngines`, sees `inflightInferenceCount[A] === 0` (acquire not called yet), and calls `A.unload()` immediately.
3. Rewrite caller's continuation resumes: `acquireInference(A)` (too late) → `A.chat.completions.create()` on a torn-down engine.

The `inflightInferenceCount`/`pendingUnload` deferral only protects calls that have *already* acquired. The fast-path resolve → acquire boundary is unguarded.

## Blast radius

Narrow. Requires a picker model-switch concurrent with an in-progress rewrite click on a different model. Recoverable: the failed inference surfaces a per-model error and the user retries (UI already supports this). `serialChain` keeps it from corrupting the cache. Not a data-loss or crash-the-app bug — a rare, recoverable inference failure. Deferred from #143 as a known follow-up, not a merge blocker.

## Fix options

1. **Acquire before the load resolves.** Have consumers `acquireInference(modelId)` *before* awaiting `loadEngine`, paired with `releaseInference` in `finally`. Then `evictAllExcept` sees a positive count and parks the unload. Cost: acquire references a model id whose engine may not exist yet — `evictAllExcept` would need to tolerate parking an id that isn't in `loadedEngines` (no-op park).
2. **Bracket inside `loadEngine`'s fast path.** Return a `{ engine, release }` handle from `loadEngine` that has already incremented the count, so there's no caller-side gap. Changes the `loadEngine` signature / all call sites.
3. **Re-check after acquire.** After `acquireInference`, verify the engine is still in `loadedEngines`; if evicted, re-load. Adds a retry hop but no signature change.

Option 1 or 2 closes the window cleanly; 3 is a smaller patch that tolerates the race instead of eliminating it.

## Acceptance

- [ ] A test reproduces the interleave: load A, start a rewrite on A, fire a picker switch to B in the same tick, assert A is not unloaded until the rewrite's `releaseInference`.
- [ ] Fix chosen and applied; existing web-llm cache/serialization tests stay green.

Found during review of #143.

Vaishnavi1709:
## Summary

Closes the time-of-check-to-time-of-use race between `loadEngine`'s fast-path resolve and `acquireInference()`. The `await` after `loadEngine` returns yielded to the microtask queue; in that gap a concurrent picker switch's chain entry could run `evictAllExcept(B)`, see `inflightInferenceCount[A] === 0` (acquire happened **inside** the rewrite primitive, after the load resolved), and call `A.unload()` immediately — leaving the rewrite caller's continuation about to run on a torn-down engine.

Fix per issue option 1: consumers acquire the inference guard **synchronously**, before any `await`, paired with release in `finally`. With the count already positive, `evictAllExcept` parks A in `pendingUnload` and the deferred `.unload()` only runs once the caller releases.

Closes #148

## What changed

| File | Change |
|---|---|
| `src/lib/webllm/web-llm.ts` | Tightened the `loadEngine` docstring — contract is now **mandatory** ("inference callers MUST acquire BEFORE awaiting"), with an explicit code example. Preloaders (ModelSelector) explicitly exempted. No behavioral change to the function itself. |
| `src/components/features/SectionRewrite.tsx` | `useSectionRewrite.onClick` wraps with outer `acquireInference` / `releaseInference`. `modelId` snapshotted at the top so the release ID matches the acquire ID. |
| `src/hooks/useResumeRewrite.ts` | `start()` callback — same outer pair, held across the **whole** chain-of-sections run (summary + every experience role) so no step boundary leaks. |
| `src/lib/webllm/eval/run-eval-browser.ts` | `runForModel()` — same outer pair. This path was the **most exposed**: `makeRealRewriteFn` calls `engine.chat.completions.create` directly, bypassing the rewrite primitives' inner defensive belt. |

The inner `acquireInference / releaseInference` inside `rewrite-section.ts` / `rewrite-summary.ts` stays as defensive belt for direct primitive callers (count rises to 2, falls to 0 — no behavioral change to the deferral semantics).

## Tests

3 new tests in `src/lib/webllm/web-llm.test.ts` under `describe("acquire-before-load TOCTOU (#148)")`:

1. **Negative** — demonstrates the bug shape: without the outer acquire, eviction unloads A immediately while the caller's continuation was about to use it.
2. **Positive** — verifies the fix contract: with the outer acquire, eviction parks A and `.unload()` defers until release.
3. **Nested** — verifies the inner acquire from rewrite primitives doesn't disrupt the deferral (count 1 → 2 → 1 → 0; unload only fires at 0).

## Acceptance criteria (#148)

- [x] A test reproduces the interleave: load A, start a rewrite on A, fire a picker switch to B in the same tick, assert A is not unloaded until `releaseInference` — covered by the positive + nested tests.
- [x] Fix chosen (Option 1) and applied; existing web-llm cache/serialization tests stay green — all 12 pre-existing tests in `web-llm.test.ts` pass alongside the 3 new ones.

## Test plan

- [x] `npm run test` green — **942 / 942 pass**
- [x] `npm run lint` clean
- [x] `npm run build` clean
- [x] Manual smoke not required — the race window is microseconds, not reliably triggerable by hand. Unit tests pin both bug and fix mechanically. (Happy-path regression for `Rewrite section` + `Rewrite full résumé` would only confirm no regression in the *non-racing* path, which the test suite already covers.)

## Reviewer-blocker scan

- [x] No raw `<button>` introduced, no hardcoded colors, semantic tokens only
- [x] SPDX header preserved on every touched file
- [x] No backwards-compat shims or commented-out code
- [x] `loadEngine` signature unchanged — no churn to ModelSelector / spike / other non-inference callers
- [x] `modelId` snapshotted at the top of each callback so the release ID matches the acquire ID even if `selectedModelId` changes mid-run
- [x] Release order in `finally`: `releaseInference` runs **before** the section-rewrite-lock release so the deferred unload drains while the lock is still held
- [x] ModelSelector correctly NOT wrapped (it's the preloader — the one *causing* the race, not a victim)
- [x] Inner `acquireInference / releaseInference` inside `rewrite-section.ts` / `rewrite-summary.ts` preserved as defensive belt for direct callers

### DIFF
--- a/src/components/features/SectionRewrite.tsx
+++ b/src/components/features/SectionRewrite.tsx
@@ -35,7 +35,11 @@
 import { useCallback, useEffect, useMemo, useState } from "react";
 import type { ReactNode } from "react";
 import { detectWebGpu } from "../../lib/webllm/capability.ts";
-import { loadEngine } from "../../lib/webllm/web-llm.ts";
+import {
+  acquireInference,
+  loadEngine,
+  releaseInference,
+} from "../../lib/webllm/web-llm.ts";
 import { rewriteSectionWithLlm } from "../../lib/webllm/rewrite-section.ts";
 import type {
   ProgressUpdate,
@@ -140,19 +144,27 @@ export function useSectionRewrite(
     // (and updates one render late, so it can't be relied on alone).
     const release = acquire();
     if (release === null) return;
+    // Snapshot the model id so the same id is released that we acquired —
+    // ModelSelector could in principle update `selectedModelId` mid-run.
+    const modelId = selectedModelId;
+    // Acquire the inference guard SYNCHRONOUSLY, before any await — closes
+    // the load→use TOCTOU window from #148. A concurrent picker switch's
+    // eviction will see the positive count and park `.unload()` until the
+    // matching release runs in `finally`.
+    acquireInference(modelId);
     try {
       setStatus({
         kind: "loading",
         progress: { progress: 0, text: "Starting…" },
       });
-      const engine = await loadEngine(selectedModelId, (progress) => {
+      const engine = await loadEngine(modelId, (progress) => {
         setStatus({ kind: "loading", progress });
       });
       setStatus({ kind: "rewriting" });
       const result = await rewriteSectionWithLlm(
         trimmedBullets,
         engine,
-        selectedModelId,
+        modelId,
       );
       if (result.bullets.length === 0) {
         setStatus({
@@ -169,6 +181,7 @@ export function useSectionRewrite(
           err instanceof Error ? err.message : "Couldn't load the rewrite model",
       });
     } finally {
+      releaseInference(modelId);
       release();
     }
   }, [trimmedBullets, acquire, selectedModelId]);
--- a/src/hooks/useResumeRewrite.ts
+++ b/src/hooks/useResumeRewrite.ts
@@ -30,7 +30,11 @@
 
 import { useCallback, useEffect, useMemo, useState } from "react";
 import { detectWebGpu } from "../lib/webllm/capability.ts";
-import { loadEngine } from "../lib/webllm/web-llm.ts";
+import {
+  acquireInference,
+  loadEngine,
+  releaseInference,
+} from "../lib/webllm/web-llm.ts";
 import {
   rewriteResumeWithLlm,
   type ResumeRewriteProgress,
@@ -155,18 +159,28 @@ export function useResumeRewrite(
   const start = useCallback(async () => {
     const release = acquire();
     if (release === null) return;
+    // Snapshot the model id so the same id is released that we acquired —
+    // ModelSelector could in principle update `selectedModelId` mid-run.
+    const modelId = selectedModelId;
+    // Acquire the inference guard SYNCHRONOUSLY, before any await — closes
+    // the load→use TOCTOU window from #148. Held across the WHOLE chain
+    // (summary + every experience role) so the engine cannot be torn down
+    // by a concurrent picker switch at any step boundary, not just at the
+    // first loadEngine. Released in `finally` whether the run completes
+    // successfully or errors out.
+    acquireInference(modelId);
     try {
       setStatus({
         kind: "loading",
         progress: { progress: 0, text: "Starting…" },
       });
-      const engine = await loadEngine(selectedModelId, (progress) => {
+      const engine = await loadEngine(modelId, (progress) => {
         setStatus({ kind: "loading", progress });
       });
       const result = await rewriteResumeWithLlm(
         rewriteableSections,
         engine,
-        selectedModelId,
+        modelId,
         (progress) => {
           setStatus({ kind: "running", progress });
         },
@@ -190,6 +204,7 @@ export function useResumeRewrite(
           err instanceof Error ? err.message : "Couldn't load the rewrite model",
       });
     } finally {
+      releaseInference(modelId);
       release();
     }
   }, [acquire, rewriteableSections, selectedModelId]);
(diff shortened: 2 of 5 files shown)


s-annam (approved):
## PR Review: close loadEngine TOCTOU race by acquiring inference before await (#148)

### Summary
Clean, correct fix for the load→use TOCTOU window. Inference callers now `acquireInference(modelId)` **synchronously before `await loadEngine`**, so a concurrent picker switch's `evictAllExcept` sees a positive in-flight count and parks `.unload()` in `pendingUnload` until the matching `releaseInference` in `finally`. Mechanism matches `web-llm.ts` semantics exactly.

### Highlights
- **Correct race closure** at all three inference call sites (`SectionRewrite`, `useResumeRewrite`, `run-eval-browser`). The eval path — the most exposed, since `makeRealRewriteFn` hits `engine.chat.completions.create` directly and bypasses the primitives' inner belt — is properly wrapped.
- **`modelId` snapshotted** at the top of each callback so the released id always matches the acquired id even if `selectedModelId` changes mid-run.
- **`finally` release on both success and error paths** — symmetric, no leaked count.
- **Inner `acquireInference`/`releaseInference` in `rewrite-section.ts`/`rewrite-summary.ts` preserved** as defensive belt; nested count 1→2→1→0 verified to still defer unload to the final release.
- **Non-inference preloader (ModelSelector) correctly NOT wrapped** — it's the cause of the race, not a victim, and the gap is harmless when no inference follows.
- **Tests pin both halves**: negative (demonstrates the bug shape), positive (engine survives until release), nested (inner belt doesn't disrupt deferral).

### Verification
Ran on the PR branch in an isolated worktree:
- `npm run test` → **942/942 pass**
- `npm run typecheck` → clean
- `npm run lint` → clean
- CI `verify` + `fallow` → pass

No raw `<button>`, no hardcoded colors, SPDX headers preserved, `loadEngine` signature unchanged.

### Verdict
**Action:** APPROVE
**Rationale:** 0 blocking, 0 suggestions, 0 nits. Fix is correct, minimal, well-tested, and all gates are green.

Issue #238, opened 2026-06-28
Parser: degree-less education entry dropped + year orphaned


s-annam:
## Summary

An education entry whose first line is a **program/certificate title without a degree keyword** is dropped entirely, and its year is mis-attributed to a surviving entry. Two stacked bugs:

- **C1 — entry dropped:** a degree-keyword-less education entry is not recognized as an entry and is lost.
- **C2 — year orphaned:** the dropped entry's year binds to a different (surviving) education entry that had no date of its own.

## Observed

Source EDUCATION (2 entries):
```
Applied Data Science Program: Leveraging AI for Effective Decision-Making        2023
Northfield Professional Education

Bachelor of Technology in Computer Science & Engineering
Riverside Institute Of Technology, Toronto
```

Parsed (1 entry):
```
degree="Bachelor of Technology" field="Computer Science & Engineering"
institution="Riverside Institute Of Technology, Toronto" year="2023"   ← 2023 belongs to the Northfield entry
```

The Northfield "Applied Data Science Program" entry is **gone**; its `2023` leaked onto the Riverside entry, which actually has **no** date.

## Root cause

Education entry recognition gates on a degree keyword (`B.S.`, `Bachelor`, `M.S.`, …). A program/cert-style line ("… Program: …" + an institution line) fails the gate and is not emitted as an entry. The free-floating year `2023` is then collected at section level and bound to the wrong entry.

Part of the dateless/keyword-less "valid entry lacking the field we anchor on" family (cf #31, #145).

## Fix direction

Recognize an education entry by **entry-boundary shape** (a header-like line followed by an institution line) rather than requiring a degree keyword; treat the degree keyword as enrichment. Bind a year to the entry it physically belongs to, never reassign an orphaned year to an unrelated entry. (Tracked under the anchor-on-shape epic.)

## Acceptance

- Both education entries survive (count = 2); Northfield "Applied Data Science Program" entry present.
- `2023` attaches to the Northfield entry; the Riverside entry has no fabricated year.
- Synthetic fixture (program-title + degree-line mix, no PII) + golden.

## Related

Child of the anchor-on-shape epic (with the dateless-role-fold issue). Family: #31, #145.

Issue #235, opened 2026-06-28
Epic: anchor entry recognition on header shape, not enrichment fields (date/degree)


s-annam:
## Problem

Multiple parser entry-extractors mis-recognize or mis-bound a valid entry because they **assume a rigid header field/position** instead of anchoring on the entry's header *shape*. Two related failure shapes:

**(a) Gate on an enrichment field that may be absent** — a perfectly valid entry is dropped or mis-bound when the field is missing:
- **Experience** anchors on a **date range** → a dateless role header folds into the previous role (#239).
- **Education** anchors on a **degree keyword** → a program/cert entry is dropped and its year orphans onto another entry (#238).

**(b) Assume a fixed header field ORDER** — the header window doesn't span the full multi-line header when the field order varies:
- **Experience role-first ordering** (`Role → Dates → Company → bullets`, company *below* the date anchor) → the company line shifts the body window, the role's bullets never enter the entry `description`, and they leak to "Other bullets" (#215).

Same shape already seen in #31, #145, and the role-first orphan-bullet fixture (commit `85c2bc9`). Each has been patched point-wise before; the recurrence shows the *rigid-header-assumption* design is the underlying defect.

## Goal

Adopt a single principle across the entry extractors: **anchor on entry-boundary SHAPE** (a header-like line that opens a block), and treat date / degree / other field signals as **enrichment**, not as the recognition gate. An entry exists when a header-shaped line introduces content; missing date/degree degrades the entry's fields, never drops the entry.

## Scope

- Unify entry-boundary detection in the experience and education extractors (and the shared entry-block primitive where applicable) around header-shape anchoring — recognizing the header regardless of which enrichment fields are present or in what order.
- The header window must span the full multi-line header (role/company/location/date) independent of field order.
- Never reassign an orphaned date/year to an unrelated entry.
- Preserve current behavior on the existing corpus (the dated/keyworded common case must not regress).

## Children

- [x] #238 — degree-less education entry dropped + year orphaned (`education.ts`)
- [x] #239 — dateless role folded into prior entry (`experience.ts`)
- [x] #215 — role-first ordering (company below date) orphans bullets to "Other bullets" (`entry-blocks.ts`)
- [x] #258 — institution name ending in a section anchor eaten as a section header (`regex.ts` / splitter layer; partial fix in #257)

## Acceptance

- Both child issues' acceptance criteria met.
[...]

PR #257, opened 2026-06-28
fix(parser): stop institution name being eaten as a section header; surface the major

s-annam:
## Summary

Fixes two distinct education-surfacing bugs that share one real-world résumé shape (a degree-less program entry plus an institution line whose name ends in a section-anchor word, e.g. `… Education`):

1. **Parser — root cause.** `matchAnchorFallback` misclassified an institution name ending in an anchor word (`"ACME Professional Education"`) as an education *section header*. The section splitter then consumed the line as a section boundary and dropped the whole entry **upstream of `education.ts`**, orphaning its graduation year onto a surviving entry. **Guard 8** rejects a line that mixes an ALL-CAPS acronym token with Title-case words — that shape is a proper-noun org entity, not a heading. Wholly-ALL-CAPS (`PROFESSIONAL EXPERIENCE`) and wholly-Title-case (`Relevant Experience`) headers still match.
   - This is the **section-splitter-layer recurrence** of #238's symptom. #238's extractor-layer fix (degree-less program recognition) was correct but masked: under real `pdfjs` extraction the line was eaten before the extractor ever saw it. Sibling to the epic #235 children (#238/#239/#215), at a layer they don't cover.

2. **Render.** `ResumeEducation.field` (the major / program title) was parsed but never rendered in the reconstructed view **or** the exported PDF. Now surfaced in both — comma-joined after the degree (`Degree, Major`), or alone in the primary slot for a degree-less program — and editable end-to-end (override type, `apply-overrides` write/clear, display resolver).

Refs #238
Refs #235

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green (1152 passed)
- [x] New regression tests: section-header guard (`regex.test.ts`), degree+major / degree-less header line (`ats-resume-model.test.ts`), `field` override write/clear (`apply-overrides.test.ts`), display resolver (`ReconstructedEducationSkills.test.ts`)
- [x] Verified on the originating PDF: two education entries recovered, full (non-truncated) program title shown, year no longer orphaned

## Notes

- Test data is fully synthetic (`ACME`/`QSU`/`Riverside`, `Applied Robotics Program`) — no real PII in this diff.

### DIFF
--- a/src/components/features/ReconstructedEducationSkills.test.ts
+++ b/src/components/features/ReconstructedEducationSkills.test.ts
@@ -62,6 +62,21 @@ describe("resolveEducationDisplay", () => {
     expect(d.institution).toBeUndefined();
   });
 
+  it("surfaces the major (field), passing it through and applying its override", () => {
+    const withMajor: ResumeEducation = {
+      ...base,
+      field: "Computer Science & Engineering",
+    };
+    expect(resolveEducationDisplay(withMajor, undefined).field).toBe(
+      "Computer Science & Engineering",
+    );
+    // Override replaces; an empty-string override clears the major.
+    expect(resolveEducationDisplay(withMajor, { field: "Data Science" }).field).toBe(
+      "Data Science",
+    );
+    expect(resolveEducationDisplay(withMajor, { field: "" }).field).toBeUndefined();
+  });
+
   it("collapses the dates string when both dates are cleared", () => {
     const noYear: ResumeEducation = { ...base, year: undefined };
     const d = resolveEducationDisplay(noYear, { start_date: "", end_date: "" });
--- a/src/components/features/ReconstructedEducationSkills.tsx
+++ b/src/components/features/ReconstructedEducationSkills.tsx
@@ -31,9 +31,11 @@ import { suggestSkills } from "../../lib/edit/skill-canonical.ts";
 import { Button, EditableField } from "@design-system";
 import { AddPill, RemoveButton } from "./ReconstructedAdd.tsx";
 
-/** Map an EducationEntry field name to the flat AddedEntry field it edits. */
+/** Map an EducationEntry field name to the flat AddedEntry field it edits.
+ *  `field` (major) is intentionally omitted — added entries carry no major slot,
+ *  so the major affordance renders on PARSED entries only and never routes here. */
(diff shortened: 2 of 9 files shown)


s-annam:
## Adversarial review — outcome

Ran a self-adversarial pass (manual + a reviewer agent) on this diff, focused on the new Guard 8 in `matchAnchorFallback` and the education render/PDF parity. Findings and resolutions:

| # | Sev | Finding | Resolution |
|---|-----|---------|------------|
| 1 | 🔴 | **Guard 8 false negative.** Rejecting *every* `ACRONYM + Title-case` line also dropped legit domain-qualified headers — `IT Experience`, `HR Experience`, `QA Qualifications` — folding their bullets into the prior section. | **Fixed** (`16161aa`). Guard now rejects only when a *non-acronym Title-case modifier* sits between the acronym and the head noun (`ACME **Professional** Education`, `QSU **Graduate** Academics`). An acronym paired directly with the head noun matches again. The original institution-name bug stays fixed — those lines all carry such a modifier. |
| 2 | 🟡 | **Void callback returning boolean.** `field !== "field" && onEntryField(...)` in `EducationSection.onFieldChange` — a `=> void` arrow returning `boolean`. | **Fixed** (`16161aa`). Rewritten as an explicit block; behavior identical, type intent clear. |
| 3 | 🟡 | EDUCATION_FIELD_MAP `Exclude<…,"field">` band-aid relies on the runtime guard. | Same fix as #2 — the block narrows `field` before indexing; no functional change needed. |
| 4 | ⚪ | **Residual hole (accepted).** A wholly Title-case institution that ends in an anchor word (`Harvard University Education`) or a wholly ALL-CAPS one (`ACME PROFESSIONAL EDUCATION`) is still indistinguishable from a header by shape alone, so it can still be eaten. | **Not fixed** — fundamental shape ambiguity; out of scope for this PR. Real-world institution lines carry descriptors that Guard 8 now catches; corpus is clean. Tracking-worthy if it recurs. |

Render↔PDF parity (`majorInPrimary`/`showMajor` vs `degreeField` join) verified consistent across (degree+field), (degree only), (degree-less program), (neither) — no issue.

**Gates after fixes:** `typecheck` clean · `lint` clean · **1152 tests pass**.

PR #259, opened 2026-06-29
fix(parser): stop institution name ending in a section anchor being eaten as a header (#258)

### LINKED ISSUE #258: Parser: institution name ending in a section anchor is eaten as a section header (splitter layer)
## Problem

The **section splitter** is the inverse failure shape of this epic's children. Where #238/#239 *drop a valid entry* because an extractor gates on an enrichment field, here `matchAnchorFallback` (`src/lib/heuristics/regex.ts`) *over-recognizes a header*: an **institution name whose trailing word is a section anchor** (`"<Org> Professional Education"`, `"<Org> Graduate Academics"`) is eaten as an `education` section header, splitting the document there and dropping the whole entry the line was supposed to introduce.

This is a recurrence of #238's symptom (degree-less education entry lost) at a **different layer** — #238's fix was in the extractor (`education.ts`); this line never reaches the extractor because the section splitter has already mis-classified it. Same root principle as the epic: a header-*shaped* line is not necessarily a header — header shape must be disambiguated from a proper-noun entity that merely ends in a category word.

## What shipped (partial, in #257)

PR #257 (squash `171e557`) added **Guard 8** to `matchAnchorFallback` as a point fix:

- Reject the anchor-fallback header match when a line mixes an ALL-CAPS acronym token (`/^[A-Z]{2,}$/`) with a Title-case **modifier word sitting between the acronym and the trailing head noun** — `ACME `**`Professional`**` Education`, `QSU `**`Graduate`**` Academics` (the modifier is the entity tell).
- An acronym paired *directly* with the head noun is a genuine domain-qualified header and still matches — `IT Experience`, `HR Experience`.

That fix is shape-heuristic and covers the acronym-bearing case only.

## Residual holes (the reason this issue exists)

Guard 8 is intentionally narrow. It does **not** cover:

1. **Wholly Title-case institution** ending in an anchor word — `Harvard University Education`, `Riverside College Academics`. No acronym token, so Guard 8 never fires; still eaten as a header.
2. **Wholly ALL-CAPS institution** — `ACME PROFESSIONAL EDUCATION`. The `allCaps` escape (needed so real `PROFESSIONAL EXPERIENCE` headers match) also lets an all-caps institution through.

Both are shape-indistinguishable from a real header by the current line-local heuristic.

## Goal

Disambiguate "header-shaped line" from "proper-noun entity ending in a category word" using **context the line-local guard lacks** — consistent with the epic's header-shape principle. Candidate signals (to evaluate, not prescribe):

- Following-line context: a real section header is followed by entry/bullet content; an institution line is itself part of an education entry (often adjacent to a degree/date/year line).
- Whether the same anchor section was already opened earlier (a second `Education` header mid-stream is suspicious).
- Position relative to a detected education block.

## Acceptance

- `Harvard University Education` and `ACME PROFESSIONAL EDUCATION` (and the Guard 8 cases) are recognized as **entry content, not section headers** — the entry is preserved.
- Genuine qualified headers still match: `Relevant Experience`, `PROFESSIONAL EXPERIENCE`, `IT Experience`.
- Full corpus snapshot green (no regression on the dated/keyworded common case).
- Regression fixtures for the wholly-Title-case and wholly-ALL-CAPS institution shapes.

## Non-goals

- LLM-path extraction. Deterministic heuristic parser only.
- The extractor-layer gates (#238/#239) and role-first ordering (#215) — sibling children, different code paths.

## Refs

- Parent epic: #235
- Partial fix already shipped: #257 (Guard 8, commit `16161aa`)
- Original symptom at the extractor layer: #238

s-annam:
## Summary

The section splitter over-recognized a header: an institution name whose trailing word is a section anchor (`Harvard University Education`, `ACME PROFESSIONAL EDUCATION`) was eaten as an `education` header in `matchAnchorFallback`, splitting the document there and **dropping the institution line**. Splitter-layer recurrence of #238, past #257's Guard 8.

Two complementary layers:

- **Layer A — line-local Guard 9 (`regex.ts`).** Reject the anchor-fallback match when an institution-type word sits in an **interior** slot (a proper-noun name precedes it) — `Harvard University Education`, `Riverside College Academics`. Uses a narrower `INSTITUTION_NAME_HINTS` set that drops `School` (a common interior header *qualifier* — `High School Coursework`, `Law School Experience` — not an org-name tell).
- **Layer B — splitter context (`sections.ts` + `markdown-lines.ts`).** New `matchSectionHeaderDetailed` exposes the L1-exact vs L2-anchor-fallback tier; the splitter suppresses an L2 open that re-matches the **currently-open** section (an institution entry under its own header — the wholly-ALL-CAPS case the line-local guards can't reach), retaining the line as content. Gated on the *current* section (not "ever opened"), so a genuine L2 header for a different section than the open one still opens — no cross-section bleed. Ported to the markdown/DOCX path for parity.

**Scoped out:** an all-caps institution ending in an anchor with no preceding section header of that type stays line-locally indistinguishable from a real first-occurrence header (documented in the issue's Non-goals spirit).

Closes #258

## Test plan

- [x] `npm run test` green (1156 passed; full corpus snapshot unchanged — no rebake)
- [x] `npm run lint` clean
- [x] New reproduction + regression tests: `regex.test.ts` (Title-case institution + `School`-qualifier headers), `sections.test.ts` (ALL-CAPS institution retained; cross-section L2 header still opens)

## Adversarial review (1 round + fix, converged)

An independent reviewer attacked the first implementation and surfaced **2 BLOCKING** regressions, both fixed here before the PR:

1. **Guard 9 over-fired on `… School …` headers** — `INSTITUTION_HINTS` includes `School`, so `High School Coursework` / `Law School Experience` wrongly returned `null`. Fixed with the narrower `INSTITUTION_NAME_HINTS` set + regression tests.
2. **Layer B cross-section bleed** — the first cut gated suppression on "ever opened", so a real `Relevant Experience` (L2) appearing after an `EDUCATION` block was suppressed and bled into `education`. Fixed by gating on the **currently-open** section + a regression test.

Non-blocking: fallow flags 3 `markdown-lines.ts` exports as "unused" — pre-existing exports re-attributed because the diff touched the file (known fallow false-attribution), plus pre-existing advisory complexity findings. Not introduced here.

### DIFF
--- a/src/lib/heuristics/markdown-lines.test.ts
+++ b/src/lib/heuristics/markdown-lines.test.ts
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+/**
+ * Unit tests for the markdown/DOCX section path (`sectionizeMarkdown`).
+ *
+ * Focus: the #258 Layer B parity port. The PDF splitter (`classifyLine`) and
+ * this markdown path share the same invariant — an L2 head-noun-anchor line
+ * that re-matches the CURRENTLY open section is an institution entry under its
+ * own header, not a new boundary, and must be retained as content. This file
+ * pins that behavior on the markdown path so the two never drift.
+ */
+
+import { sectionizeMarkdown } from "./markdown-lines.ts";
+
+describe("sectionizeMarkdown — institution name ending in a section anchor (#258 Layer B)", () => {
+  it("retains an all-caps institution line under an open education section instead of eating it as a 2nd header", () => {
+    const markdown = [
+      "**Dana Lopez**",
+      "",
+      "dana.lopez@example.com | (312) 555-0123",
+      "",
+      "**EDUCATION**",
+      "",
+      "ACME PROFESSIONAL EDUCATION",
+      "",
+      "M.S. Data Science  2018 - 2020",
+    ].join("\n");
+
+    const { sections } = sectionizeMarkdown(markdown);
+
+    // The institution line is retained as content inside an education section,
+    // not consumed as a boundary label (which would drop the institution name).
+    const inst = sections.find((s) =>
+      s.lines.some((l) => l.text.includes("ACME PROFESSIONAL EDUCATION")),
+    );
+    expect(inst).toBeDefined();
+    expect(inst!.name).toBe("education");
+
+    // Exactly one education section — the institution line did NOT open a second.
+    expect(sections.filter((s) => s.name === "education").length).toBe(1);
+  });
+
+  it("still opens a genuine L2 header for a DIFFERENT section than the one currently open", () => {
+    // Suppression is gated on the CURRENTLY-open section, not "ever opened": a
+    // real "Relevant Experience" (L2) header after an EDUCATION block must open
+    // its own experience section, not bleed into education.
+    const markdown = [
+      "**Dana Lopez**",
+      "",
+      "dana.lopez@example.com | (312) 555-0123",
+      "",
+      "**EDUCATION**",
+      "",
+      "B.S. Computer Science, MIT  2019",
+      "",
+      "Relevant Experience",
+      "",
+      "Mentor, Local Shelter  2022 - Present",
+    ].join("\n");
+
+    const { sections } = sectionizeMarkdown(markdown);
+
+    const mentor = sections.find((s) =>
+      s.lines.some((l) => l.text.includes("Mentor, Local Shelter")),
+    );
+    expect(mentor).toBeDefined();
+    expect(mentor!.name).toBe("experience");
+
+    const edu = sections.find((s) =>
+      s.lines.some((l) => l.text.includes("B.S. Computer Science")),
+    );
+    expect(edu!.name).toBe("education");
+    expect(edu!.lines.some((l) => l.text.includes("Mentor"))).toBe(false);
+  });
+});
--- a/src/lib/heuristics/markdown-lines.ts
+++ b/src/lib/heuristics/markdown-lines.ts
@@ -40,7 +40,7 @@
 
 import type { PdfLine, PdfSection } from "./sections.ts";
 import {
-  matchSectionHeader,
+  matchSectionHeaderDetailed,
   SECTION_KEYWORDS,
   SPLIT_LETTER_NORMALIZABLE_SECTIONS,
   SPLIT_LETTER_RE,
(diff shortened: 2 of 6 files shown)


s-annam:
## Adversarial review — what was found and fixed

One review round (independent reviewer subagent attacks the diff → fix the blocking findings → re-validate → re-review). It converged after one round. Two blocking findings, both real regressions the green suite did **not** catch:

### BLOCKING #1 — Guard 9 over-fired on common student-résumé headers
The first implementation reused `INSTITUTION_HINTS`, which includes `School`. That regex sits in an interior slot of legitimate qualified headers, so Guard 9 wrongly rejected:
- `High School Coursework` → should be `education`
- `Business School Experience` / `Law School Experience` → should be `experience`
- `Summer School Projects` → should be `projects`

**Fix:** introduced a deliberately narrower `INSTITUTION_NAME_HINTS` (`University | College | Institute | Academy | Polytechnic`, **drops `School``) used **only** by Guard 9. `School` is an interior header *qualifier*, not an org-name tell; the remaining words don't pattern that way. Pinned by 4 regression assertions in `regex.test.ts`.

### BLOCKING #2 — Layer B "ever-opened" gate caused cross-section bleed
The first splitter gate suppressed any L2 header re-matching a section that had **ever** been opened. That meant a genuine `Relevant Experience` (L2) header appearing after an EDUCATION block was swallowed into `education` — its roles bled into the wrong section.

**Fix:** gate on the **currently-open** section instead (`header.section === currentSection`). Only an L2 *repeat of the still-open section* folds in (unambiguously correct — the line already belongs there); an L2 header for a *different* section still opens its own. This is a stronger, simpler invariant than the ever-opened set. Pinned by the second `sections.test.ts` test (Mentor role lands in `experience`, education holds only the degree).

### NIT (fixed) — markdown/DOCX path lacked Layer B parity
The PDF splitter and the markdown path (`sectionizeMarkdownLines`) share the same boundary logic. Ported the current-section gate to the markdown path so DOCX résumés get the same fix and the two paths can't drift.

### fallow (non-blocking)
fallow's 3 findings are pre-existing `markdown-lines.ts` exports re-attributed to this PR by its touched-file diff heuristic + an advisory complexity note — none introduced here.

## Confidence

**High.** Layered, each layer pinned by a test at the right seam:
- **Layer A (line-local, `regex.ts` Guard 9):** rejects the Title-case institution case (`Harvard University Education`) — unit-tested in `regex.test.ts`, incl. the 4 `School`-qualifier non-regressions.
- **Layer B (splitter context, `sections.ts` + `markdown-lines.ts`):** catches the all-caps case (`ACME PROFESSIONAL EDUCATION`) that is shape-indistinguishable line-locally — tested on **both** the PDF path (`sections.test.ts`) and now the markdown path (`markdown-lines.test.ts`, added in f1edd11 to close the one coverage gap this review surfaced).

**Tests/fixtures — right seam, no new fixture needed.** Tests use the synthetic `build([...])` line builder and raw markdown strings — the exact granularity these two layers operate at. A binary PDF fixture would be heavier, slower, and a worse regression signal than the line-level tests, and the corpus snapshot suite (`corpus.test.ts`) already guards end-to-end against real PDFs with **zero snapshot diff** from this change. No PII surface added.

**Gates:** `typecheck` ✓ · `npm run test` (1156+) ✓ · `lint` ✓ · `verify` CI ✓ · `fallow` ✓ · corpus snapshot unchanged.

Issue #243, opened 2026-06-28
Degenerate-case LLM escape hatch — wire suggestedEscalation:"llm"


s-annam:
Parent epic: #240
Depends on: Foundation — LLM structured-parse provider (child #1)

## Goal

When the deterministic cascade returns a degenerate result — zero experiences,
low extraction ratio, or an existing `suggestedEscalation: "llm"` — offer the
user a WebLLM recovery pass instead of showing an empty/broken parse. This is
the "escape hatch" the architecture already reserved a slot for.

## Scope

The cascade already computes `suggestedEscalation` and supports an `"llm"`
value (`types.ts:170`), and `FinalSource` already has `"llm_fallback"`
(`types.ts:257`). This child **wires those up** end to end.

- In `cascade.ts`, ensure `suggestedEscalation` is set to `"llm"` on the
  degenerate conditions (zero experiences found, low extraction ratio that
  isn't a scanned/OCR case).
- UI: when `result.suggestedEscalation === "llm"` and WebGPU is available,
  show an opt-in CTA: *"We couldn't read much of this resume. Try a local
  AI pass? (runs in your browser, ~1.2GB one-time download)"*.
- On opt-in: run `parseResumeWithLlm` (child #1), then re-render the result
  from the LLM parse, marking provenance (`final_source: "llm_fallback"`).
- Reuse the #3 download-progress UI pattern — do not rebuild it.

## Acceptance criteria

- [ ] `suggestedEscalation` set to `"llm"` on zero-experience / low-extraction
      (non-scanned) cases, with unit coverage in the cascade tests
- [ ] CTA shown only when escalation is `"llm"` AND WebGPU available; hidden
      otherwise (no fallback message)
- [ ] On opt-in, result re-renders from the LLM parse with clear provenance
      ("recovered with on-device AI")
- [ ] `final_source: "llm_fallback"` + `llm_ran: true` reported in telemetry
- [ ] No network calls after model download
- [ ] Reuses #3 progress UI (no duplicate progress component)

## Out of scope

- OCR / scanned recovery — `suggestedEscalation: "ocr"` stays a separate path
- Auto-running without opt-in

PR #279, opened 2026-07-01
fix(score): drop parser blame from content-only verdict copy (#214)

### LINKED ISSUE #214: Fix grading message about a text extractor struggling
When a resume gets a low score, in this case 50/100, This message pops up.
"Needs Work

A generic extractor struggles here — add metrics like numbers, %, or $ to more bullets so each one shows measurable impact." 

However there were no layout flags. This message "a generic extractor struggles here" leads people to think their resume may be breaking an ATS due to not having metrics when that is more of a content grading issue. Suggestion for a solution is to change the message to not mention the extractor under the content score.
Screenshot below:
Image
Sample resume used: 

[music_resume25.pdf](https://github.com/user-attachments/files/29396451/music_resume25.pdf)

Vaishnavi1709:
## Summary

The score verdict's band-opening copy (`BAND_OPENER` in `src/lib/score/recommendation.ts`) blamed parsers/extractors on the **content-only** code path — i.e. when the resume parsed cleanly, no layout flags fired, and the low score was purely a content-quality issue like weak metrics or missing sections. A 50/100 resume with a clean parse read *"A generic extractor struggles here — add metrics…"*, misattributing a real content weakness to a parsing failure and misleading users into thinking their resume was breaking ATSs when it actually parsed fine.

Reworded all three tiers to speak to content only. Deliberately scoped: the scanned short-circuit and the layout-penalty branch still name parsers where the parser really is the problem.

Closes #214.

## Before / after

| Tier | Before (misleading on content-only path) | After |
| --- | --- | --- |
| `high` | "Most generic parsers should read this cleanly" | **"This is in solid shape"** |
| `medium` | "A generic parser gets most of this" | **"This is close"** |
| `low` | "A generic extractor struggles here" | **"This has clear gaps to close"** |

Each opener composes with the concrete dimension step after the em-dash:

- `— add metrics like numbers, %, or $ to more bullets so each one shows measurable impact.`
- `— tighten each bullet to a single line that opens with an action verb.`
- `— check that phone and location extract as plain text.`
- `— use real 4-digit years on your roles — the dates currently read as redaction stubs.`
- `— round out the remaining contact and section fields.`

All 15 opener × step permutations were sanity-checked to read naturally.

## Scope discipline — what was intentionally NOT touched

Every other place that mentions parsers/extractors was reviewed and left alone because the mention is correct in that context:

| Path | Copy left as-is | Reason |
| --- | --- | --- |
| Scanned short-circuit in `getScoreRecommendation` | *"This reads as a scanned image, not selectable text — export a real text-based PDF before anything else."* | Text really is unreadable; parser mention is accurate. |
| Layout-penalty branch (`multiplier < 1`) | *"Your content scored X/100, but a `<trigger>` will scramble it for many parsers — fix that layout first."* | The layout really is causing a parse problem; parser mention is accurate. |
| `TRIGGER_PHRASE.fonts_unmappable` | *"font encoding the parser can't read"* | Same as above. |
| LLM escape-hatch banner (`useLlmEscapeHatch`) | *"We couldn't read much of this resume. Try a local AI pass?"* | Different feature (#243), different code path, out of scope per #214. |

## Reviewer notes

**(a) Low-tier wording choice.** The issue's clarification suggested `low` → "This reads thin on impact" ("e.g., final wording is the implementer's call"). I tried that first and switched to **"This has clear gaps to close"** because "thin on impact" is specificity-flavored and composes awkwardly with the structure / redacted-dates / missing-fields / round-out steps — e.g. *"This reads thin on impact — check that phone extracts as plain text"*. The neutral phrasing reads uniformly across all five downstream steps. All three openers are dimension-agnostic and tab-agnostic by construction.

**(b) Doc comment discipline.** The new comment on `BAND_OPENER` explains the WHY (content-only path, no parser blame, dimension-agnostic, tab-agnostic) but does NOT reference #214 — per CLAUDE.md's rule that fix references belong in the PR description rather than in code comments where they rot.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — **1297/1297 tests pass**, including:
  - 5 existing assertions updated to expect the new strings.
  - New regression guard `keeps content-only messages free of parser/extractor language on the no-layout-flag path (issue #214)` — parametrizes **3 tiers × 3 weakest-dimension paths** (9 combinations) and asserts the rendered sentence never matches `/parser|extractor/i` on the content-only path. Locks the fix against a future edit reintroducing parser vocabulary in either `BAND_OPENER` or any `dimensionStep()` branch.
  - Existing `flags a scanned PDF as the hard blocker` and `leads with the layout penalty` tests still pass — confirming the untouched branches still emit their (correct) parser mentions.
- [x] Manually verified in `npm run dev` using the exact `music_resume25.pdf` from the issue report. The score renders 50/100 (Needs Work band), no layout flags, and the sentence beneath the ring now reads *"This has clear gaps to close — add metrics like numbers, %, or $ to more bullets so each one shows measurable impact."* — the old *"A generic extractor struggles here"* wording is gone.

## Diff size

Two files, +61 / −9 lines total.

### DIFF
--- a/src/lib/score/recommendation.test.ts
+++ b/src/lib/score/recommendation.test.ts
@@ -110,7 +110,7 @@ describe("getScoreRecommendation", () => {
       }),
     );
     expect(msg).toMatch(/add metrics/i);
-    expect(msg).toContain("A generic parser gets most of this"); // overall 70 → medium tier
+    expect(msg).toContain("This is close"); // overall 70 → medium tier
   });
 
   it("points at Structure when it is the weakest gradable dimension", () => {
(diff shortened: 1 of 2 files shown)


s-annam (approved):
## PR Review: fix(score): drop parser blame from content-only verdict copy (#214)

### Summary
Surgical, correctly-scoped copy fix. `BAND_OPENER` no longer misattributes content-quality weakness to a parser failure on the content-only path. Verified against the full `recommendation.ts` — the only remaining parser/extractor mentions live in the scanned short-circuit, the layout-penalty branch, and `TRIGGER_PHRASE.fonts_unmappable`, where a parser really is the problem. Nothing else touched.

### Spec Alignment
| Requirement (#214) | Status | Notes |
|---|---|---|
| Content-only low score must not blame the extractor | ✅ Implemented | `low` → "This has clear gaps to close" |
| Leave layout/scanned parser mentions intact | ✅ Implemented | scanned + `multiplier < 1` branches unchanged |

### Highlights
- Regression guard parametrizes 3 tiers × 3 weakest-dimension paths and asserts `/parser|extractor/i` never matches on the content-only path — locks the fix against reintroduction in either `BAND_OPENER` or any `dimensionStep()` branch.
- Doc comment explains the WHY (content-only path, no parser blame, dimension/tab-agnostic) without an issue ref that would rot — matches CLAUDE.md.
- Wording deviation from the issue's suggested `low` copy is justified in the PR (composes uniformly across all downstream steps).

### Key Findings
None blocking. CI green (verify + fallow both pass), 1297/1297 tests.

### Verdict
**Action:** APPROVE
**Rationale:** Correct fix, tightly scoped, regression-guarded, all gates green.

PR #286, opened 2026-07-02
fix: page-furniture strip on experience path, preserve source headings, PDF round-trip

### LINKED ISSUE #283: Page running-header/footer bleeds into experience roles across page breaks
## Summary

A page **running-header/footer** line (e.g. `June 3, 2026  Jane Smith · Résumé 1`) that falls on a page break **inside the Work Experience section** is not stripped on the experience path, so it is absorbed as the **company/title of the first role on the next page** — corrupting that role and dropping its real title.

A furniture filter for exactly this shape **already exists** but is scoped to the achievements extractor only. This issue generalizes it to the experience (and other entry) paths.

## Reproduction

Reproduces on an **existing synthetic in-tree fixture** — `tests/fixtures/pdfs/latex/awesome-cv-resume.pdf` (persona: Jane Smith / `jane.smith@example.com` / `(+1) 312-555-0123`, 3 pages, `Jane Smith · Résumé N` running footer). No new PDF is needed and **no real-person PII is involved**; the original real-PII report PDF (Awesome-CV's demo ships the template author's real contact info) is **not** committed and must never be.

```ts
const bytes = readFileSync("tests/fixtures/pdfs/latex/awesome-cv-resume.pdf");
const c = await runCascade(new Uint8Array(bytes));
// c.parsed.experience[3] is the corrupted Kasa role
```

Observed `runCascade(...).parsed.experience` for the first role on page 2 (verified, not hypothetical):

```json
{
  "title": "Résumé 1",
  "company": "June 3, 2026 Jane Smith",
  "location": "Kasa Seoul, S.Korea",
  "start_date": "Jun. 2018",
  "end_date": "Jan. 2021"
}
```

Expected:

```json
{
  "title": "Founding Member & Director of Infrastructure Division",
  "company": "Kasa",
  "location": "Seoul, S.Korea",
  "start_date": "Jun. 2018",
  "end_date": "Jan. 2021"
}
```

The real title is **dropped entirely**; the footer's two halves (split on `·`) become company + title; the real company line is demoted to `location`.

## Root cause (traced)

1. **`src/lib/heuristics/pdf-extract.ts`** assembles page text strictly page-by-page and inserts a blank line at each page transition (`assembleTextFromLines`, `pdf-extract.ts:44-48`). So the page-1 footer lands in reading order **between** the last page-1 role and the first page-2 role. Footers are **not** detected or stripped here — the only item-level filter is `dropDecorativeGlyphs` (PUA icon fonts), unrelated to page furniture.

2. **`src/lib/heuristics/openresume.ts`** applies no furniture filtering before handing the experience `PdfSection` to `extractExperience` (`openresume.ts:222,241`). `stripConsumedLines` (`:112-121`) removes only contact-consumed lines. The footer stays in the section when the section spans a page break.

3. **`src/lib/heuristics/entry-blocks.ts`** — `isEntryHeaderShape` (`entry-blocks.ts:101-110`) returns **true** for the footer: it leads with a capital, is not a date-only line (`June 3, 2026` is a single date, not a range, so `isDateOnlyLine` does not fire — it strips the month/year, but `Jane Smith · Résumé 1` remains, so the line is non-date), not prose, not a program note.
   - `buildEntryBlock`'s above-anchor lookback walk (`entry-blocks.ts:972-994`) claims the footer as company/title of the role whose date anchor (`Jun. 2018 - Jan. 2021`) sits just below it. Rejection filters there (`isProseLine`, `isLocationLine`, `isWrappedContinuation`) do **not** recognize page furniture.

4. **`src/lib/heuristics/extract/experience.ts`** — `disambiguateCompanyTitle` (`experience.ts:213-343`) splits the folded header on the `·` separator, carving `June 3, 2026 Jane Smith`→company and `Résumé 1`→title, pushing the real `Kasa Seoul, S.Korea` line into `location`.

## The fix already half-exists

`src/lib/heuristics/extract/achievements.ts` already has the exact filter (added in #225 for the achievements path):

```ts
// achievements.ts:30-44
const PAGE_FURNITURE_RE =
  /(?<![A-Za-z])(r[ée]sum[ée]|curriculum\s+vitae)(?![A-Za-z])/i;
const CV_FURNITURE_RE = /(?:^|\s)cv(?:$|\s)/i;
function isPageFurniture(line: PdfLine): boolean {
  return PAGE_FURNITURE_RE.test(line.text) || CV_FURNITURE_RE.test(line.text);
}
```

It is a **non-exported local**, applied only in `achievements.ts:86` (`lines.filter((l) => !isPageFurniture(l))`). The experience, education, projects, and summary paths never call it.

## Implementation plan

1. **Promote the furniture predicate to a shared module.** Move `PAGE_FURNITURE_RE`, `CV_FURNITURE_RE`, and `isPageFurniture` out of `extract/achievements.ts` into a shared home — `line-primitives.ts` is the natural fit (already the shared home for `isBulletLine`/`isProseLine`/`stripBullet` etc.). Export `isPageFurniture(line: PdfLine): boolean`. Update `achievements.ts` to import it (behavior there must stay byte-identical — the existing achievements tests must still pass).

2. **Filter furniture on the experience path.** In `entry-blocks.ts` `parseEntryBlocks` (or upstream in `extractExperience`), drop lines for which `isPageFurniture` is true **before** anchor detection (`collectAnchors`), so the footer never becomes an above/below header line or mints a dateless anchor. Filtering in `parseEntryBlocks` covers all `date_range` sections at once.
   - Caution: `CV_FURNITURE_RE` matches a bare `cv` between whitespace — verify it does not strip a legitimate bullet/company containing "CV" (the regex is already conservative for this; keep the guard). Prefer filtering only lines that ALSO look like furniture position (short line, page-number tail) if over-matching shows up in the corpus run.

3. **Consider applying to the other entry paths** (education, projects) via the same shared predicate, since they share `parseEntryBlocks`. If `parseEntryBlocks` does the filtering centrally, this is free.

4. **Reuse the existing fixture — do not add a new PDF.** `tests/fixtures/pdfs/latex/awesome-cv-resume.pdf` already spans a page break with the `Jane Smith · Résumé N` footer and reproduces the corruption. ⚠️ **Its golden snapshot currently bakes in the bug and masks it:** `awesome-cv-resume.expected.json` was baked *with* the corrupted Kasa role, so the corpus test is green today despite the defect (see the corpus-snapshot golden-regen caveat — regenerating a golden can silently ratify a regression). After the fix, `npm run bake-fixtures` will re-bake this snapshot; the reviewer must **eyeball the Kasa role in the diff** to confirm `title`/`company`/`location` corrected rather than rubber-stamping the regen. Optionally add a focused unit assertion on `experience[3]` so the correct mapping is enforced independently of the lossy golden.

## Acceptance criteria

- [ ] `isPageFurniture` (+ its two regexes) lives in one shared module and is imported by both `achievements.ts` and the experience path — no duplicated copy.
- [ ] A résumé whose Work Experience spans a page break with a `Name · Résumé N` footer parses the first page-2 role with its **real** title/company/location; the footer text appears in **no** experience field.
- [ ] The footer no longer mints a spurious dateless role (role count is correct).
- [ ] Existing achievements furniture behavior is unchanged (achievements tests green).
- [ ] On `tests/fixtures/pdfs/latex/awesome-cv-resume.pdf`, `experience[3]` re-parses as `{title: "Founding Member & Director of Infrastructure Division", company: "Kasa", location: "Seoul, S.Korea"}` — the `Jane Smith · Résumé 1` footer appears in no field. Its `*.expected.json` snapshot is re-baked and the reviewer confirmed the Kasa role corrected in the diff (golden regen not rubber-stamped).
- [ ] No new PDF fixture is added (existing synthetic fixture reused); no real-person PII enters the repo.

## Related

- #225 — introduced `isPageFurniture` for the achievements path (the code this generalizes).
- Blocks a clean fix for the "Download PDF" round-trip bug (companion issue): the corrupted role is otherwise re-rendered into the downloaded PDF as permanent text.

### LINKED ISSUE #284: "Download PDF" reconstructed résumé doesn't round-trip through the parser — loses roles
## Summary

A résumé exported via the app's **"Download PDF"** button (the reconstructed, ATS-safe résumé) does **not** round-trip cleanly through our own parser: re-feeding the downloaded PDF into `runCascade` **loses work-experience roles** (they get merged into the previous role's description). The aggregate score barely moves, which hides the regression.

The original theory — "the download PDF drops bold / font-size signals" — is **only partly right**: the renderer *does* emit bold + distinct font sizes, but (a) the parser ignores font signals entirely, and (b) the renderer's text layout is itself hard for the text-only parser to re-segment. See root cause.

## Reproduction

Parse a multi-role résumé → `buildAtsResumeModel` → `renderAtsResumePdf` → re-parse the rendered bytes with `runCascade`. Use the **in-tree synthetic fixture** `tests/fixtures/pdfs/latex/awesome-cv-resume.pdf` (Jane Smith persona, 8 roles) as the round-trip input — no new PDF and no real-person PII needed. Observed:

| metric | original | re-parsed download |
|---|---|---|
| overall score | 71 | 71 (unchanged) |
| confidence | 0.957 | 0.962 (slightly up) |
| `experience[]` role count | 8 | **5** |

Three roles (the description-less ones and their neighbors) vanish as standalone entries — each folded into the previous role's `description` as a trailing line, e.g. the Danggeun Pay role's description ends with `Site Reliability Engineer · Danggeun Market Inc. Seoul, S.Korea` (the swallowed next role).

## Root cause (traced)

**Claim to ratify — "we don't re-introduce bold/font-size signals": FALSE as stated.** `src/lib/pdf/render-ats-pdf.ts` emits Helvetica-Bold and a real type scale:

```
SIZE_NAME 18 (bold) · SIZE_SECTION 11 (bold, uppercase, + rule) ·
SIZE_HEADER 10.5 (bold, entry header) · SIZE_SUB 9.5 (muted, date) ·
SIZE_BODY 10 (bullets with "• ")
```

The bold/size signals **are present** in the downloaded PDF. So signal loss is not the direct cause. The actual causes:

1. **The parser is blind to font signals.** No tier uses bold/weight to classify a role title vs a bullet. Font size is consumed only for section-header promotion (`markdown-emit.ts` `H1_RATIO`, font-change transitions) and name detection (`sections.ts`). Within the experience grouping path (`extract/experience.ts`, `entry-blocks.ts`), title/company/bullet classification is **purely text + x/y-geometry** — `PdfTextItem.fontSize`/`fontName` are collapsed away by `groupIntoLines` before `parseEntryBlocks` runs. So re-introducing bold buys nothing: nothing downstream reads it.

2. **The download layout is hard to re-segment.** `renderAtsResumePdf` draws each role header as a **single combined line** `Title · Company Location` (`render-ats-pdf.ts:236`, `drawEntry` → `entry.headerLine`) followed by the date on its own line (`:240`, `entry.subLine`). The original résumés this parser was tuned for stack company and title on **separate** lines with a right-aligned date on the company line. The `date_range` anchor + 2-line lookback (`entry-blocks.ts` `buildEntryBlock`) can't reliably separate a **bulletless** role (header + date, no body) from its neighbor in the regen layout, so roles alternate getting swallowed.

3. **Companion bug #283 is baked in.** The reconstructed model faithfully re-renders the corrupted page-2 role (see the page-footer bleed issue) as literal text, so that corruption becomes **permanent content** in the downloaded file and re-parses as garbage again.

## Implementation plan

This bug has three related fronts; sequence them:

### Part 1 — make the Download-PDF layout parser-friendly (primary)
Adjust `src/lib/pdf/render-ats-pdf.ts` (and `ats-resume-model.ts` if the model shape needs it) so the emitted role header is re-segmentable by the text-only parser. Options to evaluate (pick during implementation, verify with the round-trip harness):
- Emit **company and title on separate lines** (matching the stacked shape the parser expects), instead of `Title · Company` on one line.
- Ensure every role's header carries a **date line the parser anchors on**, and a clear boundary even when the role has no bullets.
- Re-run the round-trip: `runCascade(original)` role count must equal `runCascade(renderAtsResumePdf(buildAtsResumeModel(original)))` role count.

### Part 2 — decide the font-signal stance (document or scope out)
The parser ignoring bold/size is a **known limitation**, not necessarily this issue's fix. Either:
- Document it (a code comment + a note in the pipeline docs) and rely on Part 1's text-layout fix, **or**
- File a separate follow-up to teach the parser to consume `PdfTextItem.fontSize`/bold as a role-header signal (larger change — touches `groupIntoLines` retention + `entry-blocks` anchoring).

Recommendation: Part 1 fixes the user-visible bug; treat Part 2 as documentation here and a separate issue if we later want font-aware parsing.

### Part 3 — depends on issue #283
Once the page-footer-bleed fix lands, the corrupted role stops being re-rendered into the download. No separate work here beyond confirming the round-trip is clean on a multi-page input.

## Acceptance criteria

- [ ] Round-trip role count is preserved: for a representative multi-role résumé, `runCascade(download).parsed.experience.length === runCascade(original).parsed.experience.length`.
- [ ] No role's `description` ends with a line that is actually the **next** role's header (`Title · Company Location`).
- [ ] The reconstructed PDF's per-role `title` / `company` / `location` re-parse into the correct fields (spot-checked against the original parse).
- [ ] A regression test exercises the full round-trip (build model → render → re-parse) and asserts role-count + field-mapping preservation, using the existing synthetic `awesome-cv-resume.pdf` as input — no new PDF, no real-person PII.
- [ ] Font-signal limitation is either documented in-repo or split into a tracked follow-up issue (state which).

## Related

- Companion issue #283 (page running-header/footer bleed) — its corruption is otherwise baked into the download; land it first or in parallel.
- `src/hooks/useDownloadPdf.ts`, `src/lib/pdf/ats-resume-model.ts`, `src/lib/pdf/render-ats-pdf.ts` — the Download-PDF path.

### LINKED ISSUE #285: Preserve original section titles from the source PDF in reconstructed résumé + Download PDF
## Problem

When resumelint reconstructs a résumé and exports the "Download PDF", every section heading is a **hardcoded canonical string** — "Experience", "Skills", "Education", "Projects", "Achievements", "Summary". The original heading text from the source PDF (e.g. "Work History", "Technical Proficiencies", "Professional Experience", "ADDITIONAL", "Volunteering") is **discarded** the moment it matches a canonical section type, so the reconstructed résumé and the exported PDF re-label the user's sections regardless of what they actually wrote.

Two costs:

1. **Fidelity** — resumelint positions itself as a *parser audit*, but silently rewriting a user's headings misrepresents the source. Showing the verbatim heading is more honest to what the parser actually read.
2. **Lost sections** — unrecognized headers land in an `other` bucket that keeps the body lines but drops the header string. Real sections ("Publications", "Volunteering", "Certifications" under an unrecognized synonym) vanish with no label. This compounds the synonym-recognition work (#282): even headers we don't canonicalize could still render with their real label.

## Root cause — original heading is dropped at the section boundary

The header text survives only transiently as the argument to `matchSectionHeaderDetailed`. Once it matches, the pipeline keeps only the canonical `SectionName` enum; the literal line is neither stored on `PdfSection` nor pushed into any section's `lines`.

**Trace (verified):**

1. **`src/lib/heuristics/sections.ts:57-61`** — the `PdfSection` model has only a canonical `name` and body `lines`, no field for the raw heading:
   ```typescript
   export interface PdfSection {
     /** "profile" covers anything above the first recognized header. */
     name: SectionName | "profile";
     lines: PdfLine[];
   }
   ```

2. **`src/lib/heuristics/sections.ts:837-883`** — in `splitIntoSections`, when a line is classified as a header (`action.kind === "open"`), the code pushes a **new empty section** and `continue`s (lines ~871-876) — the header line is never stored anywhere. The markdown path (`src/lib/heuristics/markdown-lines.ts:346-349`) does the same.

3. **`src/lib/heuristics/regex.ts:406-443`** — `matchSectionHeaderDetailed` returns `{ section: SectionName; viaAnchorFallback: boolean } | null` and **never returns the matched string**. It lowercases/normalizes input for membership testing, then discards it. The thin wrapper `matchSectionHeader` (441-442) drops even the tier flag.

4. **`src/lib/heuristics/openresume.ts:198-318`** — `buildHeuristicResult` locates sections by canonical name (`findSection(ownedSections, "experience")`) and assembles a flat `HeuristicParsedResume` keyed by canonical field. No section-label field is carried.

5. **UI headings are hardcoded literals:**
   - `src/components/features/ReconstructedResume.tsx:425` — `<SectionHeading>Experience</SectionHeading>`
   - `src/components/features/ReconstructedResume.tsx:523` — `<SectionHeading>Projects</SectionHeading>`
   - `src/components/features/ReconstructedResume.tsx:618` — `<SectionHeading>Achievements</SectionHeading>`
   - `src/components/features/ReconstructedEducationSkills.tsx:233` — `<SectionHeading>Education</SectionHeading>`
   - `src/components/features/ReconstructedEducationSkills.tsx:424` — `<SectionHeading>Skills</SectionHeading>`

6. **Export headings are hardcoded literals:** `src/lib/pdf/ats-resume-model.ts:291-302` assigns `{ heading: "Experience" }`, `"Projects"`, `"Achievements"`, etc.; `"Summary"` at `src/lib/pdf/render-ats-pdf.ts:209`. The `AtsSection` type (`ats-resume-model.ts:59-62`) already has a `heading: string` field — it's just always fed a literal. The renderer uppercases each `section.heading` (`render-ats-pdf.ts:215-232`).

7. **`other` bucket** — unrecognized headers open a section via `classifyLine` → `{ kind: "open", name: "other" }` (`sections.ts:~956`). Body lines are kept; the header line itself was already dropped at the boundary, so the original header string is **not** retained.

## Design decisions (baked into this issue)

- **Preserve original, canonical as fallback.** Store the verbatim source heading and render it. When absent (e.g. `profile` / synthesized sections), fall back to the current hardcoded canonical word. Scoring stays keyed on canonical `SectionName` — only the *display* layer reads the raw heading. No scorer changes.
- **Preserve source casing.** Store the heading as it appeared. The export renderer already uppercases at draw time (`render-ats-pdf.ts:215-232`); leave that as-is so the exported PDF stays visually consistent while the *stored* value is faithful. The on-screen reconstructed résumé renders the raw heading as-stored (title/mixed case preserved).
- **`other` sections become visible.** With a raw heading preserved, an `other` section can render under its real label instead of vanishing. (Surfacing `other` sections in the UI/export can be a follow-up if it widens scope — the field plumbed here is the prerequisite.)

### Step 1 — Add `rawHeading` to the section model
`src/lib/heuristics/sections.ts:57-61` — add an optional field:
```typescript
export interface PdfSection {
  name: SectionName | "profile";
  /** Verbatim heading text as it appeared in the source PDF, when this
   *  section was opened by a recognized/other header. Absent for "profile"
   *  (content above the first header) and synthesized sections. */
  rawHeading?: string;
  lines: PdfLine[];
}
```

### Step 2 — Capture the heading text at the boundary
- `src/lib/heuristics/sections.ts` `splitIntoSections` (~837-883) — when opening a new section (`action.kind === "open"`), set `rawHeading` on the newly-pushed section to the trimmed original line text (the same string passed to the matcher, before normalization).
- `src/lib/heuristics/markdown-lines.ts:346-349` — mirror the same capture on the markdown path.
- Verify `matchSectionHeaderDetailed` is called with the un-normalized line so the raw text is available at the call site; if only a normalized copy is in scope, thread the original through. Do **not** change what the matcher returns unless necessary — capture happens at the caller.

### Step 3 — Thread `rawHeading` to the parsed model
- `src/lib/heuristics/sections.ts:76-88` — carry `rawHeading` through `SectionedResume.byName` entries (or expose it alongside so downstream can read it).
- `src/lib/heuristics/openresume.ts:198-318` — when `buildHeuristicResult` locates each section via `findSection`, propagate its `rawHeading` onto the parsed result so the UI/export can read a per-section label. Add the field to `HeuristicParsedResume` (per-section, e.g. a `headings: Partial<Record<SectionName, string>>` map, or a label on each section group — implementer's choice, but it must reach both the UI and the export builder).

### Step 4 — Render raw heading in the reconstructed résumé (fallback to canonical)
- `src/components/features/ReconstructedResume.tsx:425,523,618` — replace literal `<SectionHeading>Experience</SectionHeading>` etc. with `<SectionHeading>{rawHeading ?? "Experience"}</SectionHeading>` (canonical word as fallback).
- `src/components/features/ReconstructedEducationSkills.tsx:233,424` — same for Education / Skills.

### Step 5 — Use raw heading in the exported PDF (fallback to canonical)
- `src/lib/pdf/ats-resume-model.ts:291-302` — feed the section's `rawHeading` into the `heading` field: `heading: rawHeading ?? "Experience"` (and the corresponding canonical fallbacks for Projects / Achievements / Education / Skills).
- `src/lib/pdf/render-ats-pdf.ts:209` — same for the Summary heading.
- Leave the uppercase-at-draw behavior (`render-ats-pdf.ts:215-232`) unchanged.

### Step 6 — Tests
- Unit: a fixture whose section reads "Work History" (mapped to `experience`) yields `PdfSection.rawHeading === "Work History"` after `splitIntoSections`; a synonym for skills ("Technical Proficiencies") likewise round-trips.
- Unit: `profile` and any synthesized section have `rawHeading === undefined` and the UI/export fall back to canonical words.
- Snapshot/corpus: confirm the reconstructed model now carries per-section headings without changing any `*.expected.json` score (scoring is canonical-keyed; regenerate goldens only if a key/count legitimately changes, never to mask a regression).

## Acceptance criteria

- [ ] `PdfSection` carries an optional `rawHeading` populated with the verbatim source heading for recognized **and** `other` sections; `undefined` for `profile`/synthesized sections.
- [ ] `splitIntoSections` (text path) and `markdown-lines.ts` (markdown path) both capture the original heading string, casing preserved.
- [ ] `rawHeading` is threaded through `SectionedResume` / `HeuristicParsedResume` to the UI and export layers.
- [ ] Reconstructed résumé (`ReconstructedResume.tsx`, `ReconstructedEducationSkills.tsx`) renders the raw heading when present, falling back to the canonical word when absent.
- [ ] Exported "Download PDF" labels sections with the raw heading when present, canonical fallback otherwise; existing uppercase-at-draw behavior unchanged.
- [ ] A résumé titled "Work History" reconstructs and exports with the heading "Work History" (uppercased in the PDF), not "Experience".
- [ ] Anonymous ATS score is unchanged for all existing fixtures (scoring stays canonical-keyed; no `*.expected.json` score drift).
- [ ] `npm run verify` passes (typecheck → lint → coverage → build → fallow).

## Notes / relationship to other work

- Complements #282 (section-title synonym recognition): synonyms still map to canonical types for scoring, but now the user sees their own wording.
- Follow-up candidate (out of scope here): actually surfacing `other` sections in the reconstructed UI/export under their preserved label. This issue plumbs the field that makes that possible.

### LINKED ISSUE #287: Two-column location folds into company field on the experience path
## Problem

In two-column résumé layouts (e.g. the Awesome-CV fixture `tests/fixtures/pdfs/latex/awesome-cv-resume.pdf`), the right-column location is folded onto the company line for every role, so `experience[N].company` comes back as e.g. `"Kasa Seoul, S.Korea"` instead of `company: "Kasa"` + `location: "Seoul, S.Korea"`.

This is a pre-existing limitation independent of the page-footer bleed (#283). #283 fixed the footer displacing the title/company; its AC#5 idealized the split (`company:"Kasa"`, `location:"Seoul, S.Korea"`) but that split was consciously deferred here — the repro test honestly asserts `company.startsWith("Kasa")`.

## Scope

Teach the experience entry path to separate a folded right-column location from the company when the source layout is two-column (`disambiguateCompanyTitle` / `stripLocationSuffix` in `src/lib/heuristics/extract/experience.ts`). The lossy corpus golden cannot catch a company/location swap, so any fix needs a focused unit assertion on the split.

## Related
- Follow-up to #283 (page-footer bleed) — surfaced during PR #286 review.

s-annam:
## Summary

Epic across four M1 · Parser Hardening issues, accumulated on one branch:

- **#283 (fix)** — Generalize the page running-header/footer furniture strip from the achievements-only path to the shared experience/entry path. `isPageFurniture` promoted to `line-primitives.ts`; a **separator-adjacent-to-keyword** positional guard keeps legit `Company | Title` / `Resume Parser | …` headers while stripping `Name · Résumé N` footers. Fixes the footer bleeding into the first page-2 role's title/company.
- **#285 (feat)** — Preserve verbatim source section headings (`rawHeading` on `PdfSection`, `sectionHeadings` map on `SectionedResume`); render them in the reconstructed résumé + Download PDF, falling back to canonical words. Scoring stays canonical-keyed — no score drift.
- **#284 (fix)** — Make the Download-PDF reconstructed résumé round-trip through the parser by emitting role headers in the stacked `title` / `company·location + dates` shape (role count was **8→5**, now **8→8**). Font-signal blindness documented as an out-of-scope limitation.
- **#287 (fix)** — Split a two-column folded right-column location off the company on the experience path. Awesome-CV folds the location onto the company line with no comma before the city (`Kasa Seoul, S.Korea`), so it landed wholesale in `company`. Adds **Pass D** to `stripLocationSuffix` (space-delimited intl `…Company City, Country`, single-token city + closed-gazetteer country, mirroring Pass B's US guards) and a Pass C guard so a company-internal comma (`Co., Ltd.`, `Command, MND`) is not mistaken for the city boundary. Follow-up to #283, surfaced during this PR's review. All 8 fixture roles now split cleanly.

Resolves #283
Resolves #285
Resolves #284
Resolves #287

## Location-fold behavior (Pass C vs Pass D)

The two-column split is boundary-driven — where the company ends and the city begins is only knowable from a **comma** or a **city gazetteer** (we carry a *country* gazetteer, not a city one):

| Input shape | Handler | Result |
|---|---|---|
| `Google, Mexico City, Mexico` (comma before city) | Pass C | `company: Google` + `location: Mexico City, Mexico` ✅ |
| `Kasa Seoul, S.Korea` (space-fold, single-token city) | Pass D | `company: Kasa` + `location: Seoul, S.Korea` ✅ |
| `Google Mexico City, Mexico` (space-fold, **multi-word** city) | Pass D | **deferred** — stays `company: Google Mexico City`, no location |

The last row is the deliberate limitation: with no comma, `Google Mexico City` is structurally ambiguous (company `Google` + city `Mexico City`, vs. subsidiary `Google Mexico` + city `City`). Pass D captures a single token, so a locality-generic tail (`City`, `Beach`, `Town`, …) is the truncated end of a multi-word place — `LOCALITY_SUFFIX_RE` **defers** those rather than emit a wrong split. Not tracked as an issue; will revisit if it shows up on a real résumé. Comma-delimited multi-word cities (`…, Mexico City, Mexico`) are unaffected — Pass C handles them.

## Test plan

- [x] `npm run test` green (1390 tests)
- [x] `npm run verify` green (typecheck → lint → coverage → build → fallow) — passed in pre-push hook
- [x] New regression tests: page-furniture footer strip, furniture false-positive guards, full build→render→re-parse round-trip, two-column location split (Pass D), Pass D locality-tail deferral
- [ ] Manually spot-check reconstructed résumé + Download PDF headings in `npm run dev`

No new fixture binaries — existing synthetic `awesome-cv-resume.pdf` reused; no PII added.

## Adversarial review

Bounded pre-PR adversarial review (independent reviewer → fix → re-verify → re-review), **2 rounds, converged** (covered #283/#284/#285).

**Round 1 — 1 blocking, fixed.** `FURNITURE_POSITION_RE` (#283) over-matched ordinary `Company | Title` / `Company · Location` headers that merely contain "Resume"/"CV": a bare `|`/`·`/`•`/`‣` alternative combined with the bare-keyword `isPageFurniture` silently deleted the whole entry before anchor detection. Concretely reproduced — `Resume Parser | Python, React  Jan 2024 - Present` → **0** experience entries (vs 1 for `Task Tracker | …`). Real hazard given this tool's own audience (résumé-tooling names: "Resume Parser", "CV Builder", …).
  - **Fix:** the positional guard now requires the separator **adjacent** to the résumé/CV/page-number token (rebuilt from a shared `FURNITURE_KEYWORD` fragment). Legit headers survive; the real `Jane Smith · Résumé 1` footer is still stripped (satisfies both the `· Résumé` adjacency and the `Résumé 1` keyword+digit tell). +2 regression tests (false-positive survives, footer still dropped).

**Round 2 — clean.** Over-match closed (4 legit headers verified surviving), footer still stripped, regex construction sound (`i` flag, escaped metacharacters, correct interpolation), tests non-vacuous. `#285` (score stays canonical-keyed, fallbacks correct, no injection surface) and `#284` (stacked-shape composes with #285's heading logic; anchor still matches) cleared. Fallow's 3 "unused exports" in `markdown-lines.ts` confirmed a pre-existing diff-attribution artifact (not new dead code); duplication clones are pre-existing structural parallelism.

**#287 addendum (added after the 2-round review).** Implementation self-review caught the real hazard: once the abbreviated `s.korea` alias makes Pass C's country match, Pass C's greedy multi-word city over-captured across a company-internal comma (`Omnious. Co., Ltd. Seoul, S.Korea` → city `Ltd. Seoul`, `R.O.K Cyber Command, MND Seoul, S.Korea` → city `MND Seoul`). Fixed with the `cityStartsWithCompanyText` guard (reject a "city" whose first token is a legal suffix or all-caps acronym) so Pass D peels only the real single-token city. Verified against all 8 fixture roles; comma-delimited multi-word cities (`…, Mexico City, Mexico`) still handled by Pass C. Full suite green; corpus golden unchanged.

## PR #286 review (Rohith) — 2 findings, both fixed in `3c698ee`

**1. Pass D mis-split a multi-word city with a locality-generic tail.** `Google Mexico City, Mexico` → `company: "Google Mexico"` + `location: "City, Mexico"` — both wrong. The `#287` addendum *claimed* these stayed deferred, but nothing enforced it: `SPACE_INTL_RE` grabbed `City` as the single-token city. Rohith's literal guard (defer when the remainder ends in a capitalized non-legal-suffix word) would also kill the shipped `Kasa Seoul` / `Sigma Analytics Berlin` cases — structurally a company's last word and a multi-word-city lead are indistinguishable. **Fix:** added `LOCALITY_SUFFIX_RE` to defer specifically the locality-*generic* tails (`City`/`Beach`/`Town`/…) that are never standalone cities, so `Google Mexico City` stays whole while single-token folds still split. Proper-noun compounds (`Buenos Aires`) remain a documented gazetteer-bound limitation. +1 regression test.

**2. Furniture strip dropped real bullets mentioning "resume"/"CV".** `isEntryPageFurniture` combined the bare résumé/CV keyword with `FURNITURE_POSITION_RE`, whose **keyword-independent** `\bpage\s+\d` and `\b\d+\s*(?:/|of)\s*\d+\b` alternatives fire on real content — e.g. `• Rebuilt the resume parser, improving 3 of 5 core metrics` was silently deleted. Especially likely on this project's own domain and on recruiting résumés. **Fix (Rohith's lever 2):** dropped those two keyword-independent alternatives; the positional tell must now sit adjacent to the résumé/CV keyword. Genuine footers keep matching (the #283 fixture `Jane Smith · Résumé 1` hits the separator-adjacency and keyword+digit branches; no test relied on the bare counters). +1 regression test.

**Surviving nit (LOW, no action):** a separator-less footer like `Jane Smith Resume` (no `·`, no page digit) escapes the entry-path guard — a documented, deliberate precision-over-recall tradeoff of the furniture filter.

### DIFF
--- a/src/lib/heuristics/entry-blocks.ts
+++ b/src/lib/heuristics/entry-blocks.ts
@@ -40,6 +40,7 @@ import {
   parseDateRange,
   stripDateRange,
   isBulletLine,
+  isPageFurniture,
   isProseLine,
   stripBullet,
 } from "./line-primitives.ts";
@@ -98,6 +99,39 @@ export function isDateOnlyLine(text: string): boolean {
  *     "Minor in Economics", "Relevant Coursework: …" are properties of the entry
  *     above, not a new entry head.
  */
+/**
+ * A running-header/footer POSITION signal, required on the entry paths in
+ * addition to the {@link isPageFurniture} keyword before a line is stripped as
+ * furniture (#283). The keyword test alone over-matches on entry sections, where
+ * a legitimate project or role title routinely CONTAINS "Resume"/"CV" — "Resume
+ * Linter", "CV Toolkit", even "Resume Linter 2024 - 2025". A genuine page footer
+ * carries a structural tell such a title never does: a name↔label separator
+ * ("Jane Smith · Résumé"), the résumé/CV keyword immediately followed by a page
+ * number ("Résumé 1", "Resume 2"), or a "Page N" / "N of M" counter. The
+ * achievements path (#225) keeps the bare keyword test — an award line is far
+ * less likely to embed the word — so its behavior is unchanged.
+ */
+const FURNITURE_KEYWORD = "(?:r[ée]sum[ée]|resume|cv|curriculum\\s+vitae)";
+const FURNITURE_POSITION_RE = new RegExp(
+  // separator ADJACENT to the résumé/CV keyword ("Jane Smith · Résumé",
+  // "Résumé | Jane Smith") — a bare separator anywhere is NOT enough, or a
+  // legit "Resume Parser | Python" / "Company · Title" header would be stripped.
+  `${FURNITURE_KEYWORD}\\s*[·•‣|]|[·•‣|]\\s*${FURNITURE_KEYWORD}` +
+    // the résumé/CV keyword immediately followed by a page number ("Résumé 1")
+    `|${FURNITURE_KEYWORD}\\s+\\d` +
+    // an explicit page counter ("Page 2", "1 of 3", "2/5")
+    `|\\bpage\\s+\\d|\\b\\d+\\s*(?:\\/|of)\\s*\\d+\\b`,
+  "i",
+);
+
+/** True when a line is page running-header/footer furniture on an ENTRY path —
+ *  it carries both the résumé/CV keyword ({@link isPageFurniture}) AND a footer
+ *  position signal ({@link FURNITURE_POSITION_RE}), so a real project/role title
+ *  that merely contains "Resume"/"CV" is not stripped (#283). */
+function isEntryPageFurniture(line: PdfLine): boolean {
+  return isPageFurniture(line) && FURNITURE_POSITION_RE.test(line.text);
+}
+
 export function isEntryHeaderShape(text: string): boolean {
   const t = text.trim();
   if (!t) return false;
@@ -756,6 +790,18 @@ export function parseEntryBlocks(
 ): EntryBlock[] {
   if (!section || section.lines.length === 0) return [];
 
+  // Strip page running-header/footer furniture (a repeated "Name · Résumé N"
+  // line a continuation page carries) BEFORE any anchor detection or header
+  // folding (#283, generalizing the achievements-only strip of #225). When an
+  // entry section spans a page break, that footer lands in reading order between
+  // the last role on one page and the first on the next; left in, it becomes the
+  // company/title of that first page-2 role (or mints a spurious dateless role),
+  // dropping the real title. Every entry path (experience `date_range`, projects
+  // / achievements `first_line`) filters here at once. Achievements already
+  // pre-filters its lines, so this is idempotent there.
+  const furnitureFiltered = section.lines.filter((l) => !isEntryPageFurniture(l));
+  if (furnitureFiltered.length === 0) return [];
+
   // Fold wrapped multi-line role headers (an org/date that spilled onto extra
   // physical rows) back into one logical header BEFORE anchor detection, so a
   // header whose closing date-year wrapped still opens a `date_range` entry
@@ -763,8 +809,8 @@ export function parseEntryBlocks(
   // hint / first line, not a date range that can wrap incomplete.
   const lines =
     cfg.anchor === "date_range"
-      ? mergeWrappedHeaderRows(section.lines, cfg.headerLookback || 2)
-      : section.lines;
+      ? mergeWrappedHeaderRows(furnitureFiltered, cfg.headerLookback || 2)
+      : furnitureFiltered;
   const anchors = collectAnchors(lines, cfg.anchor);
   if (anchors.length === 0) {
     // A `first_line` section with no anchorable header line is a flat bullet
--- a/src/lib/heuristics/entry-blocks.ts
+++ b/src/lib/heuristics/entry-blocks.ts
@@ -106,8 +106,11 @@ export function isDateOnlyLine(text: string): boolean {
  * a legitimate project or role title routinely CONTAINS "Resume"/"CV" — "Resume
  * Linter", "CV Toolkit", even "Resume Linter 2024 - 2025". A genuine page footer
  * carries a structural tell such a title never does: a name↔label separator
- * ("Jane Smith · Résumé"), the résumé/CV keyword immediately followed by a page
- * number ("Résumé 1", "Resume 2"), or a "Page N" / "N of M" counter. The
+ * ("Jane Smith · Résumé"), or the résumé/CV keyword immediately followed by a
+ * page number ("Résumé 1", "Resume 2"). The positional tell must sit ADJACENT to
+ * the keyword — a bare "Page N" / "N of M" counter anywhere on the line is NOT
+ * enough, or a real bullet that merely mentions our own domain would be dropped
+ * ("Rebuilt the resume parser, improving 3 of 5 core metrics", #286 review). The
  * achievements path (#225) keeps the bare keyword test — an award line is far
  * less likely to embed the word — so its behavior is unchanged.
  */
@@ -118,9 +121,7 @@ const FURNITURE_POSITION_RE = new RegExp(
   // legit "Resume Parser | Python" / "Company · Title" header would be stripped.
   `${FURNITURE_KEYWORD}\\s*[·•‣|]|[·•‣|]\\s*${FURNITURE_KEYWORD}` +
     // the résumé/CV keyword immediately followed by a page number ("Résumé 1")
-    `|${FURNITURE_KEYWORD}\\s+\\d` +
-    // an explicit page counter ("Page 2", "1 of 3", "2/5")
-    `|\\bpage\\s+\\d|\\b\\d+\\s*(?:\\/|of)\\s*\\d+\\b`,
+    `|${FURNITURE_KEYWORD}\\s+\\d`,
   "i",
 );
 
--- a/src/lib/heuristics/extract/experience.ts
+++ b/src/lib/heuristics/extract/experience.ts
@@ -150,6 +150,19 @@ function cityStartsWithCompanyText(city: string): boolean {
   return LEGAL_SUFFIX_RE.test(first) || /^[A-Z]{2,}$/.test(first);
 }
 
+/** A single-token Pass-D "city" that is a locality-TYPE generic ("City", "Beach",
+ *  "Springs", "Heights", "Town", …) is never a standalone city — it is the
+ *  truncated tail of a multi-word place ("Mexico City", "Long Beach", "Cape
+ *  Town") whose earlier words the single-token space-fold regex left glued to the
+ *  company. Peeling it would mis-split "Google Mexico City, Mexico" into company
+ *  "Google Mexico" + location "City, Mexico" — both wrong (#286 review). Defer:
+ *  leave the whole string as company rather than fragment a real multi-word city.
+ *  (A proper-noun compound like "Buenos Aires" has no generic tell and stays a
+ *  known limitation — distinguishing it from a real company+city fold needs a
+ *  city gazetteer we don't carry.) */
+const LOCALITY_SUFFIX_RE =
+  /^(?:city|town|beach|springs?|heights|falls|hills|park|bay|harbou?r|grove|gardens?|valley|shores?)$/i;
+
 function stripLocationSuffix(s: string): {
   text: string;
   location: string | undefined;
@@ -191,9 +204,11 @@ function stripLocationSuffix(s: string): {
     // so Pass C (which needs a comma before the city) misses it. This is the
     // space-delimited intl case Pass C deferred: it's admissible here only under
     // the same tight guards Pass B uses for US "City, ST" —
-    //   - single-token city (space boundary; a multi-word space-fold city like
-    //     "Google Mexico City" is genuinely ambiguous with a subsidiary name and
-    //     stays deferred), and
+    //   - single-token city (space boundary). A multi-word space-fold city whose
+    //     last token is a locality generic ("…City", "…Beach") would otherwise be
+    //     truncated to that generic tail and mis-split, so LOCALITY_SUFFIX_RE
+    //     defers it ("Google Mexico City, Mexico" stays company, not fragmented),
+    //     and
     //   - a closed-vocabulary country (COUNTRY_GAZETTEER), not any capitalized
     //     word — so it fires on a real "City, Country" fold, not on a company
     //     that merely ends in a comma tail.
@@ -204,7 +219,11 @@ function stripLocationSuffix(s: string): {
     const SPACE_INTL_RE =
       /\s+([A-Z][A-Za-z.\-]+),\s*([A-Z][A-Za-z.\-]+(?:\s+[A-Z][A-Za-z.\-]+)*)$/;
     const mSpaceIntl = s.match(SPACE_INTL_RE);
-    if (mSpaceIntl && COUNTRY_GAZETTEER.has(mSpaceIntl[2].toLowerCase())) {
+    if (
+      mSpaceIntl &&
+      COUNTRY_GAZETTEER.has(mSpaceIntl[2].toLowerCase()) &&
+      !LOCALITY_SUFFIX_RE.test(mSpaceIntl[1])
+    ) {
       const before = stripDanglingSeparator(s.slice(0, mSpaceIntl.index));
       if (before)
         return { text: before, location: `${mSpaceIntl[1]}, ${mSpaceIntl[2]}` };
--- a/src/lib/heuristics/extract/experience.ts
+++ b/src/lib/heuristics/extract/experience.ts
@@ -116,8 +116,11 @@ function looksLikeLocationTail(after: string): boolean {
  *   Pass C — comma-delimited international "…, City, Country": validates the
  *     trailing token against `COUNTRY_GAZETTEER` (closed ~249-entry ISO set +
  *     colloquial aliases) to avoid false-matching any capitalized word.
- *     Space-delimited intl is NOT attempted here — multi-word countries make
- *     space boundaries too ambiguous (deferred to a follow-up issue).
+ *   Pass D — space-delimited international "…Company City, Country" (#287): the
+ *     two-column fold where the location is glued onto the company line with no
+ *     comma before the city ("Kasa Seoul, S.Korea"). Single-token city only and
+ *     a closed-vocabulary country, mirroring Pass B's US guards; tried last so
+ *     the comma-delimited passes win first.
  *
  * Conservative design choices shared by all passes:
  *   - The state/country suffix must be in a closed vocabulary (US_STATE_CODE_RE
@@ -133,6 +136,18 @@ function looksLikeLocationTail(after: string): boolean {
  *  artifact, never real company text. */
 function stripDanglingSeparator(s: string): string {
   return s.replace(/[\s,–—\-|·]+$/, "").trim();
+}
+
+/** A comma-delimited intl "city" (Pass C group 1) whose FIRST token is a legal
+ *  suffix ("Ltd.") or an all-caps org acronym ("MND") is company text carried
+ *  past a company-internal comma — not a city — so that comma is not the
+ *  company/city boundary. Reject it in Pass C so the space-delimited Pass D
+ *  peels only the real trailing single-token city:
+ *  "Omnious. Co., Ltd. Seoul, S.Korea" → company "Omnious. Co., Ltd." +
+ *  "Seoul, S.Korea", not company "Omnious. Co." + "Ltd. Seoul, S.Korea" (#287). */
+function cityStartsWithCompanyText(city: string): boolean {
+  const first = city.split(/\s+/)[0];
+  return LEGAL_SUFFIX_RE.test(first) || /^[A-Z]{2,}$/.test(first);
 }
 
 function stripLocationSuffix(s: string): {
@@ -161,9 +176,38 @@ function stripLocationSuffix(s: string): {
     const INTL_SUFFIX_RE =
       /,\s*([A-Z][A-Za-z.\-]+(?:\s+[A-Z][A-Za-z.\-]+)*),\s*([A-Z][A-Za-z.\-]+(?:\s+[A-Z][A-Za-z.\-]+)*)$/;
     const mIntl = s.match(INTL_SUFFIX_RE);
-    if (mIntl && COUNTRY_GAZETTEER.has(mIntl[2].toLowerCase())) {
+    if (
+      mIntl &&
+      COUNTRY_GAZETTEER.has(mIntl[2].toLowerCase()) &&
+      !cityStartsWithCompanyText(mIntl[1])
+    ) {
       const before = stripDanglingSeparator(s.slice(0, mIntl.index));
       if (before) return { text: before, location: `${mIntl[1]}, ${mIntl[2]}` };
+    }
+
+    // Pass D — space-delimited international "…Company City, Country" (#287).
+    // Two-column templates (Awesome-CV) fold the right-column location onto the
+    // company line with no comma between company and city ("Kasa Seoul, S.Korea"),
+    // so Pass C (which needs a comma before the city) misses it. This is the
+    // space-delimited intl case Pass C deferred: it's admissible here only under
+    // the same tight guards Pass B uses for US "City, ST" —
+    //   - single-token city (space boundary; a multi-word space-fold city like
+    //     "Google Mexico City" is genuinely ambiguous with a subsidiary name and
+    //     stays deferred), and
+    //   - a closed-vocabulary country (COUNTRY_GAZETTEER), not any capitalized
+    //     word — so it fires on a real "City, Country" fold, not on a company
+    //     that merely ends in a comma tail.
+    // Tried after Pass C so a genuine comma-delimited multi-word city
+    // ("…, Mexico City, Mexico") is still captured whole; Pass D then peels the
+    // single-token fold Pass C skipped (including the company-internal-comma case
+    // "…Co., Ltd. Seoul, S.Korea" that Pass C's guard rejected).
+    const SPACE_INTL_RE =
+      /\s+([A-Z][A-Za-z.\-]+),\s*([A-Z][A-Za-z.\-]+(?:\s+[A-Z][A-Za-z.\-]+)*)$/;
+    const mSpaceIntl = s.match(SPACE_INTL_RE);
+    if (mSpaceIntl && COUNTRY_GAZETTEER.has(mSpaceIntl[2].toLowerCase())) {
+      const before = stripDanglingSeparator(s.slice(0, mSpaceIntl.index));
+      if (before)
+        return { text: before, location: `${mSpaceIntl[1]}, ${mSpaceIntl[2]}` };
     }
   }
 
--- a/src/lib/heuristics/page-furniture-experience.repro.test.ts
+++ b/src/lib/heuristics/page-furniture-experience.repro.test.ts
@@ -1,0 +1,80 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+/**
+ * Repro regression for #283 — a page running-header/footer line
+ * ("June 3, 2026  Jane Smith · Résumé 1") that falls on a page break INSIDE the
+ * Work Experience section used to be absorbed as the company/title of the first
+ * role on the next page, dropping that role's real title.
+ *
+ * The furniture strip already existed for the achievements path (#225); #283
+ * promotes `isPageFurniture` to `line-primitives.ts` and filters it centrally in
+ * `parseEntryBlocks`, so every entry path (experience/projects/achievements)
+ * drops the footer before anchor detection.
+ *
+ * Reproduces on the in-tree synthetic fixture
+ * `tests/fixtures/pdfs/latex/awesome-cv-resume.pdf` (persona: Jane Smith /
+ * jane.smith@example.com, a `Jane Smith · Résumé N` running footer). No new PDF
+ * and no real-person PII — the fixture is synthetic. The lossy `*.expected.json`
+ * golden records only counts and cannot catch a title/company swap, so this
+ * asserts the exact field mapping directly.
+ */
+
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+import { describe, it, expect, beforeAll } from "vitest";
+import { runCascade } from "./cascade.ts";
+import type { HeuristicParsedResume } from "./types.ts";
+
+const HERE = dirname(fileURLToPath(import.meta.url));
+const FIXTURE = join(
+  HERE,
+  "../../..",
+  "tests/fixtures/pdfs/latex/awesome-cv-resume.pdf",
+);
+
+// Footer tokens that must never leak into any experience field once furniture is
+// stripped: the running-footer date, name, "Résumé", and the page number tail.
+const FOOTER_TOKENS = ["Résumé", "Resume", "Jane Smith", "June 3, 2026"];
+
+describe("#283 — page footer bleeds into experience roles across page breaks", () => {
+  let parsed: HeuristicParsedResume;
+
+  beforeAll(async () => {
+    const bytes = readFileSync(FIXTURE);
+    const c = await runCascade(new Uint8Array(bytes));
+    parsed = c.parsed;
+  });
+
+  it("parses the first page-2 role (Kasa) with its real title, footer stripped", () => {
+    const exp = parsed.experience ?? [];
+    // The footer corrupted an existing role rather than minting one, so the
+    // count is unchanged — no spurious dateless role.
+    expect(exp).toHaveLength(8);
+
+    const kasa = exp[3];
+    expect(kasa.title).toBe(
+      "Founding Member & Director of Infrastructure Division",
+    );
+    // Company leads with the real org "Kasa" (this two-column fixture folds the
+    // right-column location onto the company line for every role; the #283 fix
+    // is about the footer no longer displacing it, not about location splitting).
+    expect(kasa.company?.startsWith("Kasa")).toBe(true);
+    expect(kasa.start_date).toBe("Jun. 2018");
+    expect(kasa.end_date).toBe("Jan. 2021");
+  });
+
+  it("leaves no footer text in any experience field", () => {
+    const exp = parsed.experience ?? [];
+    for (const role of exp) {
+      const blob = [role.title, role.company, role.location, role.description]
+        .filter(Boolean)
+        .join(" ");
+      for (const token of FOOTER_TOKENS) {
+        expect(blob).not.toContain(token);
+      }
+    }
+  });
+});
--- a/src/lib/heuristics/page-furniture-experience.repro.test.ts
+++ b/src/lib/heuristics/page-furniture-experience.repro.test.ts
@@ -58,10 +58,12 @@ describe("#283 — page footer bleeds into experience roles across page breaks",
     expect(kasa.title).toBe(
       "Founding Member & Director of Infrastructure Division",
     );
-    // Company leads with the real org "Kasa" (this two-column fixture folds the
-    // right-column location onto the company line for every role; the #283 fix
-    // is about the footer no longer displacing it, not about location splitting).
-    expect(kasa.company?.startsWith("Kasa")).toBe(true);
+    // This two-column fixture folds the right-column location onto the company
+    // line for every role. #283 stopped the footer from displacing it; #287 then
+    // split the folded location off, so company is the clean org and the folded
+    // "Seoul, S.Korea" lands in `location` (Pass D of stripLocationSuffix).
+    expect(kasa.company).toBe("Kasa");
+    expect(kasa.location).toBe("Seoul, S.Korea");
     expect(kasa.start_date).toBe("Jun. 2018");
     expect(kasa.end_date).toBe("Jan. 2021");
   });
--- a/src/lib/pdf/ats-resume-model.ts
+++ b/src/lib/pdf/ats-resume-model.ts
@@ -64,6 +64,10 @@ export interface AtsSection {
 export interface AtsResumeModel {
   contact: AtsContact;
   summary?: string;
+  /** Verbatim source heading for the Summary section (#285); falls back to
+   *  "Summary" at draw time when absent. Only meaningful when `summary` is
+   *  set — the Summary heading is drawn separately from `sections`. */
+  summaryHeading?: string;
   sections: AtsSection[];
 }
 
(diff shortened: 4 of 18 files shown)


src/lib/heuristics/page-furniture-experience.repro.test.ts
```diff
@@ -0,0 +1,80 @@
...
+    expect(exp).toHaveLength(8);
+
+    const kasa = exp[3];
+    expect(kasa.title).toBe(
+      "Founding Member & Director of Infrastructure Division",
+    );
+    // Company leads with the real org "Kasa" (this two-column fixture folds the
+    // right-column location onto the company line for every role; the #283 fix
+    // is about the footer no longer displacing it, not about location splitting).
+    expect(kasa.company?.startsWith("Kasa")).toBe(true);
```
s-annam: [Suggestion]: #283 AC#5 as written expects `company: "Kasa"` + `location: "Seoul, S.Korea"`, but the location stays folded onto the company line (`startsWith("Kasa")` here is honest about that). The load-bearing bug — the footer displacing title/company — is fixed, and this two-column fold is a pre-existing concern out of #283's scope. Since the lossy golden can't catch a company/location swap, worth a one-line note in the PR that the idealized field split is consciously deferred, so a merger doesn't read AC#5 as met verbatim.

src/lib/pdf/ats-resume-model.ts
```diff
@@ -226,16 +230,36 @@ export function buildAtsResumeModel(
...
+  // sentence break), so the parser folded the header into the previous role's
+  // body and dropped the role — and a bulletless role had no anchor at all.
+  const experienceEntries: AtsEntry[] = experiences.map((exp, i) => {
+    const title = (exp.title ?? "").trim();
+    // Company · Location joined with the dot; the date range is appended after a
+    // whitespace gap (the natural right-aligned-date shape) rather than another
+    // " · ", so stripping the date anchor leaves a clean "Company · Location"
+    // with no dangling separator leaking into the parsed company field.
+    const org = joinHeader([exp.company, exp.location], " · ");
+    const orgLine = [org, experienceDateRange(exp)].filter(Boolean).join("  ");
```
s-annam, line 251: [Nit]: The comment above (lines 245-247) says the date range is "appended after a ` · `", but this joins with `"  "` (two spaces). The two-space join is the correct behavior — it's what lets the re-parser strip the date anchor and leave a clean `Company · Location`. Just update the comment to match the code.

s-annam (commented):
## PR Review: page-furniture strip, preserve headings, PDF round-trip

### Summary
Focused, well-tested epic across three M1 · Parser Hardening issues. CI green (`verify` + `fallow`). Correctness hardened by a 2-round pre-PR adversarial pass (see the PR body's Adversarial review section). No blocking items.

### Spec alignment
| Issue / AC | Status | Notes |
|---|---|---|
| #283 shared `isPageFurniture`, both paths | ✅ | promoted to `line-primitives.ts`, no dup |
| #283 footer in no experience field, no spurious role | ✅ | repro test asserts; count stays 8 |
| #283 `experience[3]` → `{company:"Kasa", location:"Seoul, S.Korea"}` | ⚠️ Partial | footer fixed + real title restored; **location stays folded into `company`** (two-column fold, pre-existing, out of scope). Test honestly uses `startsWith("Kasa")` |
| #285 `rawHeading` plumbed, canonical fallback, score unchanged | ✅ | every site `heading ?? "…"`; corpus goldens no drift |
| #284 round-trip role count 8→8 | ✅ | `render-roundtrip.repro.test.ts` locks it |
| #284 font-signal limitation documented | ✅ | in-code comment; matches issue recommendation |

### Highlights
- #283 adjacency guard (résumé/CV keyword *adjacent* to a separator) precisely closes the over-match — legit `Resume Parser | Python` headers survive, `Name · Résumé N` footers strip. Clear rationale comment.
- Furniture predicate DRY'd into one shared module; round-trip + false-positive regression tests lock both bug classes.

### Verdict
**COMMENT** — no blocking items; CI green; one spec-nuance suggestion + one comment nit. (Author review — cannot self-approve; a second approving review is still required to merge.)

src/lib/heuristics/page-furniture-experience.repro.test.ts
s-annam: Deferred and now tracked: filed #287 for the two-column location→company fold. #283's actual fix (footer no longer displacing title/company) stands; the idealized AC#5 field split (`company:"Kasa"`+`location:"Seoul, S.Korea"`) is the separate two-column concern in #287. Test stays honest with `startsWith("Kasa")`. Leaving this thread open for you to close.

src/lib/pdf/ats-resume-model.ts
s-annam: Good prompt to look closer — the detailed comment at 245-248 is actually correct (it says the date is appended after a whitespace gap, contrasting *against* another ` · `). The genuinely misleading bit was the line-235 shorthand `"Company · Location · Dates"`, which implied a third ` · ` before the dates. Fixed that shorthand to match the two-space join in 0d84fae.

rohithgollapalli (commented):
Bundles four M1 parser-hardening fixes: generalizing the page furniture strip to the shared entry path, preserving verbatim source headings, the stacked round-trip role shape, and Pass D for two-column folded international locations. The heading-preservation and round-trip work look solid and well-tested. Two findings below are worth acting on before merge.

### 1. Pass D mis-splits a genuine multi-word space-folded city (`experience.ts`)

`SPACE_INTL_RE` captures the city as a **single** token immediately before the country comma. For a plausible two-column fold like `"Google Mexico City, Mexico"` (a role at Google in Mexico City), the regex matches `" City, Mexico"` → `city="City"`, `before="Google Mexico"`, giving:

- `company = "Google Mexico"`
- `location = "City, Mexico"`

Both wrong. Before this PR the string was left whole (company unsplit, no location); Pass D introduces a **new** incorrect split. The PR description explicitly says these cases "stay deferred as genuinely ambiguous," but Pass D actively fires on them — the `cityStartsWithCompanyText` guard is applied to Pass C only, not Pass D, so nothing checks that the token before the captured city (`Mexico`) actually belongs to the location rather than the company.

Suggestion: gate Pass D when the remainder still ends in a capitalized non-legal-suffix word, or require the pre-city token to look location-like, so a multi-word city genuinely stays deferred instead of being fragmented.

### 2. Entry-path furniture filter can silently drop legitimate bullet/content lines (`entry-blocks.ts`)

The new filter runs over **all** lines in every entry section (experience / projects / education), including bullets — not just header lines:

```ts
const furnitureFiltered = section.lines.filter((l) => !isEntryPageFurniture(l));
```

`isEntryPageFurniture` requires `isPageFurniture` (matches a standalone `resume`/`résumé`/`cv`) **and** `FURNITURE_POSITION_RE`, which has keyword-independent alternatives `\b\d+\s*(?:\/|of)\s*\d+\b` and `\bpage\s+\d`. So a real bullet like:

- `"Rebuilt the resume parser, improving 3 of 5 core metrics"`
- `"Screened CV submissions, 3 of 5 advanced"`
- `"Shipped the CV landing page 3 redesign"`

matches both conditions and is stripped, losing that bullet from the role's body. This is especially likely on this project's own domain (résumé tooling) and on HR/recruiting resumes — the same "résumé-tooling names" hazard the PR flagged for headers applies to bullet content here. The prior achievements strip used the bare keyword only and was scoped to award lines; generalizing the positional tell to every entry line broadens the false-positive surface.

Suggestion: apply the furniture filter only to non-bullet / anchor-candidate lines, or require the résumé/CV keyword itself to participate in the positional tell (drop the keyword-independent `N of M` / `page N` alternatives, or require adjacency to the keyword).

---

Both are confirmed by tracing the diffed regexes; #1 is deterministic, #2 requires the keyword+counter co-occurrence on one line. The rest of the PR (rawHeading threading, the stacked round-trip shape, the `s.korea` gazetteer aliases, and the new regression tests) reads cleanly.

rohithgollapalli (changes requested):
Requesting changes on the two findings from my review above: (1) Pass D mis-splits a multi-word space-folded city (`experience.ts`) — deterministic incorrect split, contradicts the PR's "stays deferred" claim; (2) the entry-path furniture filter can silently drop legitimate bullet lines mentioning resume/CV alongside an `N of M` / `page N` token (`entry-blocks.ts`). Both need a fix or explicit justification before merge; the rest of the PR looks good.

src/lib/heuristics/extract/experience.ts
```diff
@@ -161,10 +176,39 @@ function stripLocationSuffix(s: string): {
...
+    //     word — so it fires on a real "City, Country" fold, not on a company
+    //     that merely ends in a comma tail.
+    // Tried after Pass C so a genuine comma-delimited multi-word city
+    // ("…, Mexico City, Mexico") is still captured whole; Pass D then peels the
+    // single-token fold Pass C skipped (including the company-internal-comma case
+    // "…Co., Ltd. Seoul, S.Korea" that Pass C's guard rejected).
+    const SPACE_INTL_RE =
+      /\s+([A-Z][A-Za-z.\-]+),\s*([A-Z][A-Za-z.\-]+(?:\s+[A-Z][A-Za-z.\-]+)*)$/;
+    const mSpaceIntl = s.match(SPACE_INTL_RE);
+    if (mSpaceIntl && COUNTRY_GAZETTEER.has(mSpaceIntl[2].toLowerCase())) {
```
rohithgollapalli: Pass D mis-splits a multi-word space-folded city. `SPACE_INTL_RE` captures the city as a single token immediately before the country comma, so `"Google Mexico City, Mexico"` matches `" City, Mexico"` → `company="Google Mexico"`, `location="City, Mexico"` — both wrong. Before this PR the string was left whole; Pass D introduces a **new** incorrect split. The PR description says these stay "deferred as genuinely ambiguous," but nothing here defers them — `cityStartsWithCompanyText` guards Pass C only, not Pass D. Suggestion: gate Pass D when the remainder still ends in a capitalized non-legal-suffix word (i.e. the pre-city token still looks like part of the location), so a real multi-word city stays deferred instead of being fragmented.

src/lib/heuristics/entry-blocks.ts
```diff
@@ -756,15 +790,27 @@ export function parseEntryBlocks(
...
+  // Strip page running-header/footer furniture (a repeated "Name · Résumé N"
+  // line a continuation page carries) BEFORE any anchor detection or header
+  // folding (#283, generalizing the achievements-only strip of #225). When an
+  // entry section spans a page break, that footer lands in reading order between
+  // the last role on one page and the first on the next; left in, it becomes the
+  // company/title of that first page-2 role (or mints a spurious dateless role),
+  // dropping the real title. Every entry path (experience `date_range`, projects
+  // / achievements `first_line`) filters here at once. Achievements already
+  // pre-filters its lines, so this is idempotent there.
+  const furnitureFiltered = section.lines.filter((l) => !isEntryPageFurniture(l));
```
rohithgollapalli, line 803: This filters **all** lines in every entry section (experience/projects/education), including bullets — not just header lines. `isEntryPageFurniture` requires `isPageFurniture` (a standalone `resume`/`résumé`/`cv`) AND `FURNITURE_POSITION_RE`, whose keyword-independent alternatives `\b\d+\s*(?:\/|of)\s*\d+\b` and `\bpage\s+\d` can fire on real content. So legitimate bullets are silently dropped, e.g.:

- `"Rebuilt the resume parser, improving 3 of 5 core metrics"`
- `"Screened CV submissions, 3 of 5 advanced"`
- `"Shipped the CV landing page 3 redesign"`

Especially likely on this project's own domain (résumé tooling) and HR/recruiting resumes — the same "résumé-tooling names" hazard flagged for headers applies to bullet content here. The prior achievements strip was scoped to award lines with the bare keyword; generalizing the positional tell to every entry line broadens the false-positive surface. Suggestion: apply the filter only to non-bullet / anchor-candidate lines, or require the résumé/CV keyword itself to participate in the positional tell (drop the keyword-independent `N of M` / `page N` alternatives, or require adjacency to the keyword).

src/lib/heuristics/extract/experience.ts
s-annam: Fixed in `3c698ee`. Your literal guard (defer when the remainder ends in a capitalized non-legal-suffix word) would also kill the shipped #287 cases `Kasa Seoul, S.Korea` (remainder `Kasa`) and `Sigma Analytics Berlin, Germany` (remainder `Analytics`) — structurally a company's last word and a multi-word-city lead are indistinguishable without a city gazetteer. The false positive you cited is specifically a locality-*generic* tail (`City`, `Beach`, `Town`, …) that is never a standalone city, so I added `LOCALITY_SUFFIX_RE` to defer exactly those: `Google Mexico City, Mexico` now stays whole instead of fragmenting. A proper-noun compound like `Buenos Aires` has no generic tell and remains a documented known limitation (needs a city gazetteer we don't carry). Regression test added.

src/lib/heuristics/entry-blocks.ts
s-annam: Fixed in `3c698ee` via your lever 2 (drop the keyword-independent alternatives). Removed `\bpage\s+\d` and `\b\d+\s*(?:/|of)\s*\d+\b` from `FURNITURE_POSITION_RE`; the positional tell must now sit adjacent to the résumé/CV keyword (separator-adjacent, or keyword+number). Those bare branches only ever fired *in combination with* a résumé/CV mention elsewhere on the line (a pure `Page 2 of 5` footer has no keyword, so `isPageFurniture` already gates it out), which is exactly your false-positive surface — all three examples (incl. the non-bullet-prose form) now survive. Genuine footers keep matching: the #283 fixture footer `Jane Smith · Résumé 1` hits both the separator-adjacent and keyword+number branches, and no test relied on the bare counters. Regression test added.

src/lib/heuristics/page-furniture-experience.repro.test.ts
s-annam: Superseded: the idealized field split this asked to document-as-deferred is now actually performed — Pass D of `stripLocationSuffix` (#287, a0e6b41) splits the two-column fold, so `company: "Kasa"` + `location: "Seoul, S.Korea"` per AC#5. No PR-description caveat needed anymore. Resolving as outdated.

src/lib/heuristics/extract/experience.ts
rohithgollapalli: Resolved in 3c698ee — `LOCALITY_SUFFIX_RE` now defers Pass D when the single-token city is a locality generic ("City", "Beach", …), so "Google Mexico City, Mexico" stays with the company instead of fragmenting. Verified + regression test added. Thanks!

src/lib/heuristics/entry-blocks.ts
rohithgollapalli: Resolved in 3c698ee — the keyword-independent `N of M` / `page N` alternatives were dropped, so the positional tell must now involve the résumé/CV keyword. Real bullets like "improving 3 of 5 core metrics" survive; the "· Résumé N" footer is still stripped. Verified + regression test added. Thanks!

rohithgollapalli (approved):
Both blocking findings from my earlier review are resolved in 3c698ee with targeted fixes and regression tests:

- **Pass D mis-split** — `LOCALITY_SUFFIX_RE` defers a single-token city that's a locality generic, so `"Google Mexico City, Mexico"` stays with the company instead of fragmenting into `"City, Mexico"`.
- **Furniture strip dropping real bullets** — the keyword-independent `N of M` / `page N` alternatives are gone; the positional tell now requires the résumé/CV keyword, so real bullets survive while the `· Résumé N` footer is still stripped.

Verified no regressions to the existing Pass D cases or the #283 footer strip. The rest of the PR (rawHeading threading, stacked round-trip shape, gazetteer aliases) was already clean. LGTM 🎉

PR #289, opened 2026-07-02
fix(webllm): drop parser blame from escape-hatch banner headline (#281)

### LINKED ISSUE #281: Escape-hatch banner claims "couldn't read much" on resumes that parsed cleanly
## Problem

The LLM escape-hatch banner (`LlmEscapeHatchBanner`, from #243) renders at the top of the results page with the copy:

> **We couldn't read much of this resume.** Try a local AI pass? …

It fires whenever the cascade sets `result.suggestedEscalation === "llm"` and WebGPU is available. But on at least one real resume the banner appears while the parser actually recovered most of the content — so the copy tells the user something demonstrably false, in the same way #214 did.

## Reproducer

Same fixture as #214 (PDF attached on that issue).

On that resume, the banner fires — but the parser recovered:

- Name, email, phone, and location fields
- 5 experience entries (company + title + bullets each)
- 2 education entries
- 12 skills
- Enough content to score 50/100 with all three dimensions gradable (13/40 specificity, 16/30 structure, 21/30 completeness)

That's emphatically not a case where the parser "couldn't read much."

## Why this is the same class of bug as #214

Both messages tell the user the parser failed when the parser did its job:

| Bug | Surface | Copy | Reality |
| --- | --- | --- | --- |
| #214 (fixed in #279) | Verdict sentence under the score ring | *"A generic extractor struggles here — …"* | Parser worked; score was low because content was thin |
| **This issue** | Escape-hatch banner above the score card | *"We couldn't read much of this resume."* | Parser recovered most of the resume; cascade over-triggered escalation |

Same misattribution pattern, different code path. Different feature, so needs its own PR — but the underlying user-facing bug reads identically.

## Fix direction (choose at triage)

Two hypotheses, either or both may be at play:

1. **Copy is too absolute.** Even when the escape-hatch is a legitimate offer, *"We couldn't read much of this resume"* is a false claim in the common case where the cascade recovered plenty. Rewording to something less parser-blaming (e.g. *"Try a local AI pass for a second opinion?"*) would remove the misleading claim without changing the firing rule.
2. **Firing rule is over-eager.** `suggestedEscalation === "llm"` is being set on resumes that clearly parsed successfully (all three dimensions gradable, real score computed, most fields populated). Whatever heuristic in the cascade sets that flag is either mis-calibrated or being applied to inputs where the deterministic parse is objectively fine.

## Acceptance criteria

- [ ] On the #214 reproducer PDF, the escape-hatch banner either (a) does not fire, or (b) fires with copy that does not falsely claim the parser couldn't read the resume.
- [ ] No regression on the degenerate cases the escape hatch was designed for (#243): PDFs where the deterministic parse really is degraded should still get the offer, so users don't miss the recovery affordance.
- [ ] PII / no-network invariants intact (same as #243).

## Related

- #243 — original escape-hatch feature.
- #214 — same class of "parser blame on a clean parse" bug on a different surface; fixed by #279.

Vaishnavi1709:
## Summary

The LlmEscapeHatchBanner headline read *"We couldn't read much of this resume"* on every path that fires `suggestedEscalation === "llm"`. But `chooseEscalation` in `confidence.ts` routes there from **both** hard failures (missing email, low extraction ratio, low name confidence, zero experience on non-student) AND soft confidence dips below `CANONICAL_CONFIDENCE_THRESHOLD` (0.85). On the soft path — the reproducer from #214 is the exemplar — the parser had recovered name/email/phone/all experience entries/education/skills with all three score dimensions gradable at 50/100. Claiming *"couldn't read much of this resume"* there was a false statement about the parse.

Same class of bug as #214 (fixed in #279): parser blame on a case where the parser did its job. Different surface (banner, not verdict), different code path (`useLlmEscapeHatch` / `#243`, not `getScoreRecommendation`).

Closes #281.

## Before / after

Only the headline line changes. Subheadline and CTA button stay put.

| Element | Before | After |
| --- | --- | --- |
| Banner headline | *"We couldn't read much of this resume."* | **"Not everything parsed cleanly."** |
| Banner subheadline | *"Try a local AI pass? Runs entirely in your browser — nothing leaves this tab. One-time ~1.2 GB download, cached for next time."* | *(unchanged)* |
| CTA button | *"Try a local AI pass"* | *(unchanged)* |
| Banner position, style, `role`, `aria-label` | — | *(unchanged)* |

## Why the new headline works on every firing path

`chooseEscalation` in `confidence.ts` routes to `"llm"` from **five distinct causes**. The new copy is honest across all of them:

| Firing cause | Old headline | New headline |
| --- | --- | --- |
| `no_email` hard failure | Overstated but plausible | ✅ Honest |
| `zero_experience_non_student` hard failure | Overstated but plausible | ✅ Honest |
| `low_extraction_ratio` (< 0.5) hard failure | ✅ Accurate | ✅ Honest |
| `low_name_confidence` hard failure | Overstated but plausible | ✅ Honest |
| Soft: `confidence < 0.85` (the #214 reproducer's case) | ❌ **Lie** — parser recovered most fields | ✅ Honest |

## Scope discipline — what was intentionally NOT touched

The issue's acceptance criteria explicitly says *"No regression on the degenerate cases the escape hatch was designed for."* So I deliberately did not narrow the firing rule:

| Deliberate non-change | Why |
| --- | --- |
| **`chooseEscalation` firing rule in `confidence.ts`** | On the #214 reproducer, the parser actually did have real gaps ("date not detected" on every role, "location not detected" on every role, "degree not detected" on one education entry). The confidence rule correctly identifies these as a marginal parse worth offering LLM recovery on. Narrowing the rule would hide the offer on marginal resumes that genuinely benefit from it. |
| **`CANONICAL_CONFIDENCE_THRESHOLD`, `NAME_CONFIDENCE_FLOOR`, `EXTRACTION_RATIO_FLOOR`** | Same reason — these are the knobs that control WHEN the offer appears, not what it SAYS. |
| **Subheadline copy** (`"Try a local AI pass? Runs entirely in your browser…"`) | Neutral, describes the offer, no parser blame. |
| **CTA button label** (`"Try a local AI pass"`) | The CTA names the action, doesn't blame the parser. |
| **`aria-label`s** (button + region) | Neutral, no parser blame. |

## Reviewer notes

**(a) Alternate wording tried mentally.** I considered *"Some fields didn't parse cleanly"*, *"This parse has some gaps"*, and *"Not everything came through cleanly"*. Picked **"Not everything parsed cleanly"** because (i) it reads as one clean line without stumbling, (ii) *"parsed"* names the action directly (matches app positioning: *"parser audit, not a judge"*), (iii) it composes naturally with the subheadline that follows without repeating vocabulary.

**(b) Doc comment discipline.** Added an in-source comment above the headline explaining the WHY (must stay honest across every firing path), but did NOT reference `#281` in it — per CLAUDE.md's rule that fix numbers belong in the PR description rather than code where they rot.

**(c) No `#281` in code strings.** The `it()` test title says *"(issue 281)"* without the `#` because the lint rule `no-restricted-syntax` treats `#281` as a hex color when it appears inside a code string. Same phrase in a JSDoc block is fine (rule skips comments); this is code, so the sigil is dropped.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — **1391/1391 tests pass**, including:
  - New regression-guard test *"keeps the idle headline free of overstated parser-failure claims (issue 281)"* — asserts the banner never contains `/couldn'?t read much/i` or `/couldn'?t read this resume/i`. Locks the fix against a future edit reintroducing the same class of overclaim.
  - Existing `renders the idle CTA` test still passes — confirming the CTA button label is unchanged.
  - Existing `renders loading, running, and error states` test still passes — confirming the banner's status lifecycle is unchanged.
  - Existing `fires onRecovered when the pass completes` test still passes — confirming the done-state contract with the parent is unchanged.
- [x] Manually verified in `npm run dev` using the exact `music_resume25.pdf` from #214. The blue banner now reads *"Not everything parsed cleanly."* — the old *"We couldn't read much of this resume."* wording is gone. Subheadline, CTA button, position, and styling are unchanged. Screenshot verified in the working session.

## Acceptance criteria mapping (from #281)

- [x] On the #214 reproducer PDF, the escape-hatch banner **fires with copy that does not falsely claim the parser couldn't read the resume**. New headline *"Not everything parsed cleanly"* is truthful (some fields — dates, locations, one degree, one title — genuinely didn't parse cleanly).
- [x] **No regression on the degenerate cases the escape hatch was designed for** (#243). The firing rule (`chooseEscalation`) is untouched. Every resume that got the offer before still gets it. Only the misleading headline is fixed.
- [x] **PII / no-network invariants intact** (same as #243). No new network calls, no new schema fields, no new egress paths — this is a two-word copy change plus a doc comment plus a test.

## Diff size

Two files, +27 / -1 lines. The runtime change is a single-line copy fix; the rest is the explanatory doc comment (safety) and the regression test (durability).


s-annam (approved):
## PR Review: fix(webllm): drop parser blame from escape-hatch banner headline (#281)

### Summary
Headline copy fix removing a false "couldn't read much" claim on the soft-confidence firing path, backed by a WHY comment and a regression guard. Correct, minimal, honestly scoped.

### Spec Alignment (#281)
|---|---|---|
| Copy no longer falsely claims parser failure | ✅ | "Not everything parsed cleanly" honest across all 5 `chooseEscalation` firing causes |
| No regression on degenerate cases (#243) | ✅ | Firing rule + thresholds untouched — chose fix direction (a) |
| PII / no-network invariants intact | ✅ | Copy + comment + test only; no new egress/schema |

### Highlights
- Picked copy fix over narrowing the firing rule — right call; narrowing would hide the offer on genuinely marginal parses.
- Doc comment explains *why* the copy must stay honest without embedding a rot-prone fix number.

### Follow-up applied on the branch
Broadened the regression guard (was matching only the two exact prior phrasings) to catch any future rewrite attributing a parser failure to the resume (`couldn't/can't/unable to/failed to` + `read/parse`). Pushed as a follow-up commit; test now 4/4 green, lint clean.

### Verdict
**APPROVE** — meets all acceptance criteria, no blocking issues, tests green, exemplary scope discipline.

Issue #305, opened 2026-07-03
Reduce complexity in parser/score hotspots (fallow health backlog)


s-annam:
## Backlog: reduce complexity in parser/score hotspots (fallow health)

Tracking issue for the genuine high-complexity functions surfaced by `fallow health` (the static-analysis gate, #15). These are **real** maintainability signal, not false positives — the false positives and dead code were cleaned up separately (config + trims), so `fallow dead-code` is now clean and the `fallow audit` diff-gate passes with zero introduced findings. What remains is inherited complexity debt, which the new-only audit gate does **not** block — so it belongs in the backlog, tracked here rather than silently ignored.

Repo maintainability score: **91.9 (good)**. 43 functions exceed the default thresholds (cyclomatic 20 / cognitive 15 / CRAP 30). Top targets, worst-first by cognitive complexity:

| Function | File:line | cyclo | cog | LOC | cov% | sev |
|---|---|---|---|---|---|---|
| `applyOverrides` | src/lib/edit/apply-overrides.ts:279 | 56 | 103 | 236 | 97 | critical |
| `detectColumnBoundaries` | src/lib/heuristics/pdf-layout.ts:148 | 30 | 61 | 93 | 97 | critical |
| `extractEducation` | src/lib/heuristics/extract/education.ts:632 | 55 | 55 | 234 | 100 | critical |
| `buildEntryBlock` | src/lib/heuristics/entry-blocks.ts:1000 | 43 | 52 | 185 | 100 | critical |
| `RoleHeader` | src/components/features/ReconstructedRole.tsx:304 | 36 | 42 | 139 | 0 | critical |
| `extractFromPdfBytes` | src/lib/heuristics/pdf-extract.ts:165 | 27 | 37 | 123 | 0 | high |
| `disambiguateCompanyTitle` | src/lib/heuristics/extract/experience.ts:325 | 50 | 36 | 209 | 100 | critical |
| `computeAnonymousAtsScore` | src/lib/score/score.ts:699 | 27 | 34 | 167 | 0 | high |
| `scoreCompleteness` | src/lib/score/score.ts:340 | 17 | 31 | 66 | 0 | high |
| `buildAtsResumeModel` | src/lib/pdf/ats-resume-model.ts:175 | 30 | 27 | 216 | 99 | high |

(Full 43-row list: `npx fallow health --coverage coverage/coverage-final.json`.)

### Notes / how to approach
[...]

Issue #15, opened 2026-06-06
Adopt fallow static analysis (dead-code, dupes, complexity gate)


s-annam:
## Summary

Adopt [**fallow**](https://github.com/fallow-rs/fallow) — a deterministic, Rust-based static analyzer for TypeScript/JavaScript — as resumelint's standing code-intelligence tool. resumelint is an ideal target: it is **100% TS/React** (Vite, React 19, vitest; 48 `.ts`/`.tsx` files under `src/`, no backend), so fallow covers the entire codebase with no language gaps. fallow is fast (sub-second, no AI internally) and zero-config.

This issue covers the **CI lane only** (Steps 1, 3, and 4): install the devDependency, seed a baseline, and add the CI risk gate.

**Prerequisite:** #36 (add fallow-stop.sh to repo) must be merged before Step 2 (registering the Stop hook) can be wired. Step 2 is tracked entirely in #36.

## Background

fallow ships several relevant subcommands:

| Subcommand | What it finds |
|---|---|
| `fallow dead-code` | Unused files/exports/types/enum members, unused deps, unresolved imports, stale suppressions |
| `fallow dupes` | Clone families — modes: `strict` (exact), `mild` (AST, default), `weak` (diff literals), `semantic` (renamed vars) |
| `fallow health` | Cyclomatic + cognitive complexity, CRAP score, per-file maintainability index |
| `fallow audit --changed-since <ref>` | PR risk gate over the diff |
| `fallow fix --dry-run` | Preview cleanup |

Output formats include JSON (typed contracts), **SARIF** (stable schema, GitHub code-scanning), Markdown, and GitHub annotations.

## Current state in resumelint

- `package.json` — `lint` is just `tsc -b --noEmit`; no dead-code / duplication / complexity analysis exists.
- `.github/workflows/ci.yml` runs typecheck/test; no static-analysis gate.
- fallow is **not installed** (`fallow not found`), so Step 1 below is a prerequisite for everything in this issue.

### 1. Install fallow as a devDependency

resumelint uses **npm** (`package-lock.json`). From the repo root:

```bash
npm i -D fallow
```

Verify: `npx fallow --version`.

### 2. Seed a baseline and triage

```bash
npx fallow audit --format json > .fallow/baseline.json 2>&1 || true
```

[...]

PR #306, opened 2026-07-03
chore(fallow): tune out false positives + clean dead code (#15)

s-annam:
## Summary

No `.fallowrc` existed, so fallow ran on raw defaults — false positives never got tuned out and the reporting felt like noise. This adds `.fallowrc.jsonc` that models the project (dev HTML entries, the `@design-tokens` Vite alias, public-API barrels, dev-only harness dirs) and removes the genuine dead code fallow surfaced.

**Result:** `fallow dead-code` **93 issues → 0**, and the diff-scoped `fallow audit` gate passes with **zero introduced findings** (`✓ No issues in 15 changed files`). This PR also exists to confirm the **cloud fallow / code-scanning** run is green again on a real PR — the surfaced-but-unfixed warnings from #304 are what prompted it.

### Config (`.fallowrc.jsonc`)
- **entry** — register the 3 dev-only HTML harnesses (`jd-spike` / `eval-rewrite` / `parse-eval`) so their browser modules aren't flagged unreachable. They are real entries, just excluded from the production rollup (`vite.config.ts:152`), so fallow couldn't see them.
- **ignoreDependencies** — `@design-tokens` is a Vite CSS-import alias, not an npm package (false "unlisted dependency").
- **health.ignore** — webllm `spike`/`eval`/`parse-eval` dev dirs (high CRAP only from zero coverage by design, not shipped-code debt).
- **ignoreExports** — the two public-API barrels (`heuristics/index.ts`, `jd-match/index.ts`) that deliberately re-export result-shape types; the design-system primitive variant types; and two documented keep-alive seams (`ParseMetadata` telemetry contract, `getThresholdsFor` extensibility stub).

### Source cleanup
- Rename render-side `PdfLine` → `RenderLine` (`markdown-emit.ts`) — resolves a duplicate export colliding with the parse-side `PdfLine` in `sections.ts`.
- Drop `export` on internal-only symbols; delete genuinely-dead surface (`StructuredAmbiguity` cluster, unused `AtsScoreDimensions` re-export, `BOILERPLATE_ANCHORS` barrel leak).
- Remove 9 stale suppressions in `webllm/parse-eval` (redundant once the HTML entry is registered).

### Not in scope
Inherited complexity hotspots (`applyOverrides` cog 103, etc.) are **real** signal but don't block the new-only gate — tracked in **#305** rather than silently suppressed.

Refs #15, #304, #305

## Test plan

- [x] `npm run test` green (1482 pass)
- [x] `npm run verify` green (full local CI mirror, incl. fallow audit)
- [x] Cloud fallow / code-scanning check green on this PR

### DIFF
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -1,0 +1,74 @@
+{
+  // fallow static-analysis config (issue #15). Tunes out false positives so the
+  // diff-scoped `fallow audit` gate stays high-signal. See docs.fallow.tools.
+
+  // Dev-only browser harnesses (jd-spike.html / eval-rewrite.html / parse-eval.html)
+  // are real entry points kept in-repo but excluded from the production rollup
+  // `input` (vite.config.ts) so they never ship to dist/. Register them here so
+  // fallow follows them and stops reporting their browser `.ts` modules
+  // (jd-spike-browser.ts, run-eval-browser.ts, parse-eval-browser.ts and their
+  // spike/eval helpers) as unreachable dead code.
+  "entry": [
+    "jd-spike.html",
+    "eval-rewrite.html",
+    "parse-eval.html"
+  ],
+
+  // `@design-tokens` is a Vite CSS-import alias (vite.config.ts resolve.alias),
+  // not an npm package — it's `@import`ed from src/styles.css. Fallow's dependency
+  // scanner has no view of the Vite alias, so it reports an "unlisted dependency".
+  // Teach it the specifier is intentional.
+  "ignoreDependencies": [
+    "@design-tokens"
+  ],
+
+  // Complexity/CRAP thresholds only apply to shipped product code. The dev-only
+  // WebLLM harnesses (spike/eval/parse-eval, loaded via the dev HTML entries and
+  // never in the production rollup) score high CRAP purely because they carry no
+  // test coverage — that's not maintainability debt worth tracking.
+  "health": {
+    "ignore": [
+      "src/lib/webllm/spike/**",
+      "src/lib/webllm/eval/**",
+      "src/lib/webllm/parse-eval/**"
+    ]
+  },
+
+  "ignoreExports": [
+    // Design-system primitive variant types are the deliberate public surface of
+    // the `@design-system` seam (3-tier component architecture, CLAUDE.md). They
+    // are intended API for consumers even when no in-tree caller references the
+    // union type directly — not dead code.
+    { "file": "src/design-system/primitives/Button.tsx", "exports": ["ButtonVariant", "ButtonSize"] },
+    { "file": "src/design-system/primitives/Chip.tsx", "exports": ["ChipTone"] },
+    { "file": "src/design-system/shared/ErrorState.tsx", "exports": ["ErrorStateTone"] },
+
+    // Public-API barrels. Each file documents itself as THE consumer boundary
+    // for its module ("Public surface is intentionally small… Consumers should
+    // not import internals directly") and deliberately re-exports the result-shape
+    // types a caller needs to type `runCascade` / `extractJdTerms` — even where
+    // the in-tree app imports those types from deeper paths. Re-exports with no
+    // in-repo consumer are the intended external surface, not dead code.
+    { "file": "src/lib/heuristics/index.ts", "exports": ["*"] },
+    { "file": "src/lib/jd-match/index.ts", "exports": ["*"] },
+
+    // `ParseMetadata` is the telemetry metadata contract the heuristics barrel
+    // re-exports so a host can type the `ParseEvent` payloads it wires to the
+    // `onEvent` sink. It has no in-tree consumer (the app doesn't inspect
+    // metadata), and a barrel re-export whose barrel is itself ignored doesn't
+    // register as a consumer edge — so the source export must be marked here too.
+    { "file": "src/lib/heuristics/types.ts", "exports": ["ParseMetadata"] },
+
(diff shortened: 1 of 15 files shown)


.fallowrc.jsonc
```diff
@@ -0,0 +1,74 @@
...
+    { "file": "src/design-system/shared/ErrorState.tsx", "exports": ["ErrorStateTone"] },
+
+    // Public-API barrels. Each file documents itself as THE consumer boundary
+    // for its module ("Public surface is intentionally small… Consumers should
+    // not import internals directly") and deliberately re-exports the result-shape
+    // types a caller needs to type `runCascade` / `extractJdTerms` — even where
+    // the in-tree app imports those types from deeper paths. Re-exports with no
+    // in-repo consumer are the intended external surface, not dead code.
+    { "file": "src/lib/heuristics/index.ts", "exports": ["*"] },
+    { "file": "src/lib/jd-match/index.ts", "exports": ["*"] },
```
Samhit21: [Question]: Wildcard-ignoring the two barrels masks *future* dead re-exports, unlike the enumerated entries above it.

Every other keep-alive in this list is named explicitly — `ParseMetadata`, `getThresholdsFor`, the `Button`/`Chip`/`ErrorState` variant types — so a *new* accidentally-dead export elsewhere in those files still gets caught. The two barrels use `["*"]`, which turns off the unused-export check for the whole file: if someone later adds `export { somethingDead }` to `heuristics/index.ts` or `jd-match/index.ts` and nothing consumes it, fallow stays silent.

I confirmed the `*` is load-bearing today — `ExtractedTerm`, `ExtractJdTermsResult`, and `stripBoilerplate` are re-exported by `jd-match/index.ts` but every in-repo consumer imports them from the deep `./extract-jd-terms.ts` path, so the barrel copies genuinely have no in-repo edge. So `*` isn't wrong.

The question is whether it's worth enumerating the intended public names instead, to keep catching new dead re-exports:

```jsonc
{ "file": "src/lib/jd-match/index.ts",
  "exports": ["ExtractedTerm", "ExtractJdTermsResult", "stripBoilerplate", ...] },

If the barrels' public surface churns enough that enumerating would be constant upkeep, * is a fair call — just flagging the signal you're trading away. Your call.

Samhit21 (commented):
## PR Review: chore(fallow): tune out false positives + clean dead code (#15)

### Summary
Clean, unusually well-documented chore PR — every `.fallowrc` entry carries a rationale, and the dead-code removals are surgical. No prior reviews, so all findings are net-new. I focused verification on the PR's highest risk: an "un-export" or deleted type that actually has a live consumer. It checks out —

- All un-exported symbols (`EDITABLE_KEYS`, `mkItem`/`mkPage`, `preprocessMarkdown`, `markdownToPseudoLines`, `sectionizeMarkdownLines`, `rankNounHits`, `BOILERPLATE_ANCHORS`, `NOUN_PASS_CAP`, `WEIGHTS`, `SPECIFICITY_TARGET_RATIO`) are consumed only within their own file.
- The deleted `StructuredAmbiguity` cluster has zero references anywhere in `src/`.
- The dropped `AtsScoreDimensions` re-export from `score.ts` had no consumer (the type still lives in `types.ts`, used there).
- The `PdfLine` → `RenderLine` rename is complete; `markdown-lines.ts` uses the *distinct* parse-side `PdfLine` from `sections.ts`, not the renamed render-side type — no dangling import.
- The `entry`-registered HTML harnesses all exist at repo root; dead-code dropping 93→0 (not exploding) confirms `index.html` is still detected, so the `entry` list is additive.

### Findings
1. `.fallowrc.jsonc:52-53` — `["*"]` on the two public-API barrels disables the unused-export check file-wide, so a *future* dead re-export added to either barrel won't be flagged. Consistent with the author's stated intent and defensible given barrel churn; raised as a [Question] on whether to enumerate instead. Not blocking.

### Verdict
**Action: COMMENT** — no correctness or safety issues; the one finding is a signal-vs-upkeep trade-off for the author to decide.

.fallowrc.jsonc
s-annam: Good signal to catch — agreed the enumerate-vs-`*` trade is worth taking here. Switched both barrels from `["*"]` to their enumerated public export names in `aff68f0`, so a future accidentally-dead re-export in either barrel still trips fallow instead of being silently masked. These two barrels are stable public boundaries, so the sync upkeep is cheap. Added a comment above the entries explaining the choice. Verified `dead exports 0.0%` after the change (fallow audit clean, typecheck + 1482 tests green).

s-annam: Addressed, please take a look at the latest commit.

Samhit21 (approved):
## PR Review: chore(fallow): tune out false positives + clean dead code (#15)

### Update — re-reviewed after aff68f0
My one open question (the `["*"]` barrel ignores masking future dead re-exports) is fully addressed. Both barrels now enumerate their public export names, and I verified the lists match each barrel exactly — heuristics 38/38, jd-match 19/19 — so nothing's missing (no false positive returns), nothing's stale, and a future accidentally-dead re-export will now trip fallow instead of being silently swallowed.

### Verified earlier
- All un-exported symbols are consumed only within their own file.
- Deleted `StructuredAmbiguity` cluster has zero references in `src/`.
- Dropped `AtsScoreDimensions` re-export had no consumer (type still lives in `types.ts`).
- `PdfLine` → `RenderLine` rename is complete; no dangling imports.
- `entry`-registered HTML harnesses exist at root; dead-code 93→0 confirms `index.html` is still detected.

### Verdict
**Approve** — clean, well-documented, and the review feedback landed. 👍

PR #316, opened 2026-07-03
feat: drag-and-drop resume replacement on results view

### LINKED ISSUE #315: Allow drag-and-drop to replace a resume on the results view
## Problem

Once a resume is parsed and the results view (`state.phase === "done"`) is showing the reconstructed resume, there is **no way to drop a new resume onto the page**. The inline `DropZone` is only rendered while `state.phase !== "done"`, so in the results view it is unmounted and no drop handler exists.

Because the browser has no page-level `dragover`/`drop` handler in that state, dropping a PDF **triggers the browser default action — it navigates away and opens the raw PDF**, silently discarding the parse. The only path to analyze another resume is the "Try another file" reset button.

This is a UX gap, not a blocker: users naturally expect drag-and-drop to keep working after the first parse.

## Desired behavior

While viewing a parsed result, dragging a file over the window should:

1. Show a full-page **drop affordance overlay** ("Drop to analyze a new resume") so the drop target is obvious.
2. On drop of an accepted file, show a **confirmation dialog** before discarding — replacing clears the current parse *and any inline edits*, so it's a destructive action and must be confirmed (per CLAUDE.md / UX `confirmation-dialogs`).
3. On confirm, parse the new file (reuse the existing `handleFile` path). On cancel, keep the current result untouched.

Armed **only** in `done`; `idle`/`error` already render the inline `DropZone`, which owns drops there.

## Implementation plan

Three-tier component architecture (per CLAUDE.md): shared accept logic in `lib/`, state machine in a hook, presentational overlay in `features/`.

### 1. `src/lib/file-accept.ts` (new) — one source of truth for "is this a resume file"

Extract the accept predicate out of `DropZone` so both drop surfaces agree:

```ts
export const RESUME_ACCEPT_ATTR = `application/pdf,.pdf,${DOCX_MIME},.docx`;
export const RESUME_REJECT_HINT = "That doesn't look like a PDF or DOCX. …";
export function isAcceptedResumeFile(f: File): boolean; // isPdf(f) || isDocx(f)
```

`DOCX_MIME` stays **module-private** (only used to build `RESUME_ACCEPT_ATTR`) — a new `export const` would trip the fallow dead-code gate on the diff.

### 2. `src/hooks/useReplaceResumeOnDrop.ts` (new) — window-level, confirm-gated state machine

Signature:

```ts
useReplaceResumeOnDrop({ enabled: boolean; onFile: (f: File) => void }): {
  isDragging: boolean;
  pendingFile: File | null;
  confirmReplace: () => void;
  cancelReplace: () => void;
}
```

- Window listeners for `dragenter`/`dragover`/`dragleave`/`drop`, effect gated on `enabled`.
- File-drag gate: `Array.from(e.dataTransfer?.types ?? []).includes("Files")` — ignore text/selection drags.
- **Drag-depth counter ref** so the overlay doesn't flicker as the drag crosses nested DOM elements (enter++ / leave--; visible when depth > 0).
- `preventDefault()` on `dragover` **and** `drop` — this is what stops the browser opening the file.
- Accepted drop → park file in `pendingFile` (does *not* call `onFile` yet). `confirmReplace()` → `onFile(f)` then clear. `cancelReplace()` → clear.

### 3. `src/components/features/ReplaceResumeDropOverlay.tsx` (new) — presentational

Props `{ isDragging, pendingFile, onConfirm, onCancel }`.
- Overlay: `pointer-events-none fixed inset-0 z-50 …` (must not intercept the drop — the window listener catches it), dashed card "Drop to analyze a new resume".
- Confirmation: reuse the `Dialog` + `Button` primitives from `@design-system` (native `<dialog>` gives Esc + focus-trap free). Body names `pendingFile?.name`; Cancel (ghost) / Replace (primary).

### 4. `src/components/DropZone.tsx` (edit) — rewire to shared module

Remove the private `isPdf`/`isDocx`/`isAccepted`; import `isAcceptedResumeFile`, `RESUME_ACCEPT_ATTR`, `RESUME_REJECT_HINT` from `../lib/file-accept.ts`. No behavior change.

### 5. `src/App.tsx` (edit) — wire it up

```tsx
const replaceDrop = useReplaceResumeOnDrop({
  enabled: state.phase === "done",
  onFile: handleFile,
});
// …rendered inside PageShell, after the ErrorBoundary:
<ReplaceResumeDropOverlay
  isDragging={replaceDrop.isDragging}
  pendingFile={replaceDrop.pendingFile}
  onConfirm={replaceDrop.confirmReplace}
  onCancel={replaceDrop.cancelReplace}
/>
```

## Reuse analysis

- **Capability:** accept + ingest a resume file (drag/drop + confirm) on the results view.
- **Existing surfaces found:** `src/components/DropZone.tsx` (`isPdf`/`isDocx` accept logic + `onFile`); `@design-system` `Dialog`/`Button` primitives; `useAnalyzedResume().handleFile` (the parse entry point).
- **Decision:** **extend** — the accept predicate is lifted into shared `lib/file-accept.ts` and consumed by both `DropZone` and the new overlay (no parallel accept logic); the confirm UI reuses the shared `Dialog`/`Button` primitives; ingestion reuses `handleFile`. The new hook + overlay are additive only because the results view has a genuinely different interaction model (window-level drag + destructive-replace confirm) than the inline landing `DropZone` — but they own **no** duplicated accept or ingest logic.

## Acceptance criteria

- [ ] In `done` mode, dragging a file over the window shows the drop overlay; it disappears when the drag leaves or ends.
- [ ] Overlay stays visible across nested `dragenter`/`dragleave` (no flicker) — depth counter.
- [ ] Dropping a PDF/DOCX **does not** navigate the browser / open the file (`preventDefault` on `dragover` + `drop`).
- [ ] An accepted drop parks the file and opens a confirmation dialog naming the file; it does **not** parse yet.
- [ ] Confirm → new file parses via `handleFile`, replacing the current result + edits. Cancel → current result untouched.
- [ ] A dropped non-resume file (e.g. `.txt`) is ignored (no dialog, no parse).
- [ ] Non-file drags (selected text) are ignored (no overlay).
- [ ] The hook is inert while `enabled` is false (`idle`/`error` still owned by the inline `DropZone`).
- [ ] `DropZone` accept behavior unchanged after the `lib/file-accept.ts` extraction.
- [ ] New hook covered by a jsdom test (raw `createRoot` + `act`, no `@testing-library/react` — matches the repo's other hook tests).
- [ ] `npm run verify` green (typecheck, lint, coverage, build, fallow).

## Known minor gap (out of scope, low priority)

Dropping a **non-resume** file onto the overlay in `done` mode is silently ignored — no error toast like the inline `DropZone` shows. Acceptable for a non-blocker; can add a rejection hint follow-up if desired.

s-annam:
## Summary

In `done` mode the inline `DropZone` is unmounted, so dropping a PDF hit the browser default and **opened the raw file**, discarding the parse. This restores drag-and-drop on the results view with a window-level, confirm-gated handler armed only in `done`.

Drag a file over the results → full-page "Drop to analyze a new resume" overlay → drop → confirm dialog ("…will clear the current result and any edits") → **Replace** parses / **Cancel** keeps current.

- `lib/file-accept.ts` — shared accept predicate (one source of truth for `DropZone` + the new overlay); `DOCX_MIME` stays module-private (avoids fallow dead-export flag).
- `hooks/useReplaceResumeOnDrop.ts` — window drag listeners, drag-depth counter (no flicker), `preventDefault` on `dragover`+`drop` (stops the browser opening the file), dropped file parked as `pendingFile` for confirmation.
- `features/ReplaceResumeDropOverlay.tsx` — `pointer-events-none` drop affordance + `Dialog`/`Button` confirm (reuses `@design-system` primitives; native `<dialog>` → Esc + focus-trap free).
- `DropZone.tsx` — rewired to the shared accept module (no behavior change).
- `App.tsx` — wired hook + overlay, `enabled: state.phase === "done"`.

Closes #315

## Test plan

- [x] `npm run test` green — new 9-case jsdom hook test (`useReplaceResumeOnDrop.test.tsx`): overlay toggle, depth counter, preventDefault, park/confirm/cancel, reject non-resume, ignore text drags, inert-while-disabled
- [x] `npm run verify` green (typecheck, lint, coverage, build, fallow — "No issues in 6 changed files")
- [x] Manually verified in `npm run dev` — drag/drop replace on results view (live-browser check blocked locally by HTTPS self-signed cert; covered end-to-end in jsdom)

## Known minor gap (out of scope)

Dropping a non-resume file onto the overlay in `done` mode is silently ignored — no error toast like the inline `DropZone` shows. Acceptable for a non-blocker; follow-up if desired.


s-annam:
### Round 2 — review fixes (`80c027f`)

Addressed both items from review:

**1. Replace dialog rendered in the top-left corner → now centered.**
Root cause: Tailwind's preflight zeroes the native `<dialog>`'s UA `margin: auto`, so a modal drops to the top-left unless it's explicitly positioned. The `Dialog` primitive delegates positioning to the caller by design (per its docstring), and this overlay was passing only a width. Added the same `fixed left-1/2 top-1/2 -translate-x-1/2 -translate-y-1/2` centering the three other `Dialog` callers already use (`ResumeRewrite`, `WebGpuUnavailableNotice`, `ConsentDialog`).

**2. Confirmation dialog copy is now more descriptive and friendly.**
Reworked the microcopy (consulted `ui-ux-pro-max`: confirmation-dialogs / destructive-emphasis / error-clarity) so the body names the consequence *and* reassures that the on-disk file is safe, and the buttons carry action labels instead of bare Cancel/Replace:

| | before | after |
|---|---|---|
| title | Replace this resume? | Analyze a different resume? |
| body | Analyzing {file} will clear the current result and any edits you've made. | We'll run a fresh analysis on {file} and clear the current result along with any edits you've made here. **Your original file stays untouched.** |
| buttons | Cancel / Replace | Keep current / Analyze new resume |
| overlay subtext | This replaces the resume you're looking at now. | We'll swap out the one you're viewing now. |

**Gates:** `npm run typecheck` clean · 1491 tests pass · fallow clean. Scope: one file (`ReplaceResumeDropOverlay.tsx`), +8/-7.

Issue #326, opened 2026-07-03
Decision (wontfix by-design): #295 toWinAnsi() lossy → to -> transliteration on round-trip (skia-classic, weasyprint-classic)


s-annam:
## Purpose

**Record an accepted-by-design decision** so two baselined round-trip fixtures are tracked as a deliberate tradeoff, not silent debt. This is not a bug to fix under current scope — it documents *why* the baseline exists and *what* a real fix would require.

## The artifact

On the Download-PDF round-trip (`parse → renderAtsResumePdf → re-parse`), two fixtures differ by a **single Unicode glyph** the `#295` `toWinAnsi()` render sanitizer rewrites lossily. A role title:

```
… Intern → Junior Engineer      (→ = U+2192 RIGHTWARDS ARROW)
```

round-trips as:

```
… Intern -> Junior Engineer     ("->")
```

Affected fixtures, baselined in the round-trip corpus ratchet:

- `src/lib/heuristics/corpus-roundtrip.test.ts:108` — `"google-docs/google-docs-skia-proxy-classic.pdf": ["experience"]`
- `src/lib/heuristics/corpus-roundtrip.test.ts:109` — `"unknown/weasyprint-cairo-classic.pdf": ["experience"]`

## Why it is by-design

`#295` fixed a hard crash: `renderAtsResumePdf` threw on any non-WinAnsi glyph (`→`, `‐`, or a `µ` that case-expands to `Μ` under uppercasing), killing the entire Download-PDF path. The `toWinAnsi()` sanitizer runs **after** uppercasing so `drawText` can never throw — worst case it **degrades** a glyph (`→` → `->`), never crashes.

Trading a lossy glyph substitution for a guaranteed no-crash render is the **correct** tradeoff for a client-side PDF export. The exact-match round-trip gate flags the substitution, but the substitution is intentional.

## What a real fix would require

Not a bug fix — **new functionality**: embed a Unicode-capable font in the render path so `renderAtsResumePdf` can draw `→` (and other non-WinAnsi glyphs) directly, removing the need for `toWinAnsi()` degradation. That is a font-embedding feature with bundle-size and licensing considerations, out of scope for the round-trip fidelity work. If/when pursued, it should be its own `Feature` issue.

## Decision

- **Status: wontfix (by-design).** The two baselined `experience` lines at `corpus-roundtrip.test.ts:108-109` stay, now backed by this recorded decision rather than an unexplained baseline.
[...]

Issue #325, opened 2026-07-03
Round-trip: empty-title role emits Company·Location header, re-parse assigns location as title (openresume-laverne-word-quartz)


s-annam:
## Problem

On the Download-PDF round-trip (`parse → renderAtsResumePdf → re-parse`), the `word/openresume-laverne-word-quartz.pdf` fixture mis-assigns a role's title.

Page 1 has a role with an **empty title**. The reconstructed emit puts

```
Company · Location  Dates
```

on the header line. On re-parse, with no title text present, the parser assigns the bare `City, ST` location string as the **title**, corrupting the round-tripped role.

This fixture is currently **baselined** in the round-trip corpus ratchet:

- `src/lib/heuristics/corpus-roundtrip.test.ts:110` — `"word/openresume-laverne-word-quartz.pdf": ["experience"]`

## Distinct from #145

Closed issue **#145** ("Filter empty-title (date-only) entry blocks before they reach the score") is a **score-path** concern — it strips empty-title/date-only blocks so they don't pollute the ATS score. This issue is on the **round-trip emit path**: the reconstruction emits a header for the empty-title role, and re-parse then mis-reads the location as the title. Different path, different fix — the score-side filter does not touch what `renderAtsResumePdf` emits.

## Root cause

The empty-title-role emit edge: when a role has no title, the reconstructed header line is `Company · Location  Dates` with nothing above it. The parser's title/company disambiguation then has no title line to anchor on and falls back to treating the bare location as the title.

Two candidate fixes (pick during implementation):
1. **Emit side** — for an empty-title role, emit a header shape the re-parser reads back as empty-title (not one where the location is promotable to title).
2. **Parse side** — in `disambiguateCompanyTitle`, refuse to assign a bare `City, ST` location string as a title when it matches the location on the company sub-line.

## Where to look

- `src/lib/pdf/ats-resume-model.ts` — `buildAtsResumeModel` (composes the reconstructed role header; where the empty-title header shape originates).
- `src/lib/pdf/render-ats-pdf.ts` — `renderAtsResumePdf`.
- `src/lib/heuristics/extract/experience.ts` — `disambiguateCompanyTitle` (the re-parse title/company assignment).
- `src/lib/heuristics/entry-blocks.ts` — `buildEntryBlock` / header-shape detection.
- Fixture: `tests/fixtures/pdfs/word/openresume-laverne-word-quartz.pdf`.

## Reproduction

```bash
[...]

Issue #324, opened 2026-07-03
Round-trip: 20% sidebar token pollutes PROJECTS heading in reconstructed PDF, leaks 2 phantom roles into Experience (chromium-two-column-sidebar)


s-annam:
## Problem

On the Download-PDF round-trip (`parse → renderAtsResumePdf → re-parse`), the `unknown/chromium-two-column-sidebar.pdf` fixture inflates its experience-role count from **5 → 7**. Two phantom roles appear.

Root cause is on the **render side**, not segmentation. In the reconstructed PDF, a stray `20%` sidebar-proficiency token renders **glued onto** the `PROJECTS` section heading, so the re-parsed text reads:

```
20% PROJECTS
```

Section recognition then misses the polluted heading, and the entire `PROJECTS` block leaks into `EXPERIENCE` as 2 phantom roles.

This fixture is currently **baselined** in the round-trip corpus ratchet:

- `src/lib/heuristics/corpus-roundtrip.test.ts:99` — `"unknown/chromium-two-column-sidebar.pdf": ["experience"]`

The ratchet (`KNOWN_FAILURES`) only shrinks: when this bug is fixed, the `experience` category will round-trip cleanly, the stale-entry check trips, and the baseline line must be deleted.

## Distinct from #117

Closed issue **#117** ("Two-column flattened section headers with sidebar-noise prefixes go undetected, e.g. `20% Projects`") addressed the **parse side** — it tightened the L2 anchor-fallback Guard 7 to require an uppercase lead character on every word, so already-flattened `20% Projects` input text is handled.

This issue is a **different layer**: the reconstructed **emit** (`renderAtsResumePdf` / the ATS-resume model) is what glues `20%` onto the heading in the first place, during the Download-PDF reconstruction. #117 is COMPLETED, yet this fixture still fails the ratchet because the render-side pollution is upstream of the parse-side guard. Fixing it means preventing the `20%` sidebar token from landing on the heading line in the reconstructed PDF (or ensuring the heading re-parses cleanly despite it).

## Where to look

- `src/lib/pdf/render-ats-pdf.ts` — `renderAtsResumePdf` (the Download-PDF render surface).
- `src/lib/pdf/ats-resume-model.ts` — `buildAtsResumeModel` (the model the render consumes; where section/heading text is composed).
- `src/lib/heuristics/pdf-layout.ts` — `detectColumnBoundaries` / two-column flattening (how the sidebar `20%` token gets adjacent to the heading in the first place).
[...]

PR #328, opened 2026-07-04
fix(roundtrip): sidebar-heading + empty-title role fidelity (#324, #325, #326)

s-annam:
## What

Round-trip ratchet cleanup — one accumulated commit closing three issues on the Download-PDF `parse → renderAtsResumePdf → re-parse` fidelity path.

| Issue | Fix |
|---|---|
| **#324** | Two-column sidebar noise (e.g. `20%`) no longer glues onto the recovered section heading. `sections.ts` `stripSidebarNoisePrefix` cleans the stored `rawHeading` at capture time, so the reconstructed PDF heading re-parses cleanly and the `PROJECTS` block stops leaking into `EXPERIENCE` as two phantom roles (5→7 role inflation gone). |
| **#325** | An empty-title role header (`Company · City, ST  Dates`) no longer re-parses with the location assigned as the title. `disambiguateCompanyTitle` rescues a bare-location `title → location`, gated on a **closed-vocab** predicate (`isBareLocationString`: valid USPS state code or country gazetteer), not a generic `"TitleCase, TitleCase"` shape. |
| **#326** | Recorded wontfix-by-design decision for the `toWinAnsi()` lossy glyph substitution (`→` → `->`). The `#326` cross-ref is already present in `KNOWN_FAILURES`; both baselines retained (no code change). |

Removed the `chromium-two-column-sidebar` and `openresume-laverne-word-quartz` round-trip baselines from `KNOWN_FAILURES` — the ratchet stays green.

## Verification

- `npm run verify`: **1490 passed, 1 skipped**, build clean.
- New regression guards in `experience.role-comma.test.ts` (title-path + team-path false-positive class).

## Adversarial review

Ran a 2-round adversarial review loop over the accumulated diff before opening this PR. Two blocking findings found and fixed:

1. **Round 1 — #325 over-broad rescue.** The first cut of the empty-title fix leaned on `INTL_LOCATION_RE` (a generic `"TitleCase, TitleCase"` shape matcher) with only a finite `looksLikeTitle` keyword guard. Real comma-formatted titles outside that list (`Buyer, Home Goods`, `Merchandiser, Footwear`, `Barista, Downtown Store`, `Product Owner, Growth Team`) were erased into `location`. **Fixed** by tightening `isBareLocationString` to require a closed vocabulary (valid USPS state code or country gazetteer) on the comma tail.
2. **Round 2 — same class survived in step-3b.** The `team → location` rescue (`experience.ts` step 3b) still used the raw shape regexes inline instead of the tightened helper, so the identical erasure was reachable via a 3-segment header (`Merchandising · Acme Corp · Buyer, Home Goods`). **Fixed** by routing 3b through the shared `isBareLocationString(team)` — one predicate behind all three rescue paths (also net-simplified `disambiguateCompanyTitle`: 49→46 cyclomatic).

Both fixes carry permanent regression guards (title-path + team-path), and both directions are proven — the legitimate rescues (`Austin, TX`; `Bangalore, India`; the original word-quartz empty-title round-trip) still fire.

**Remaining nits** (non-blocking, for the human reviewer): fallow flags 4 test-scaffolding clone groups (all pre-existing shared parse-harness convention across `experience.*.test.ts`, plus one 6-line intra-file repeat in the new guard file) and `disambiguateCompanyTitle` complexity (100% tested, inherited). No production duplication.


src/lib/heuristics/sections.ts
```diff
@@ -911,9 +911,30 @@ export function splitIntoSections(
...
+ * cleaned heading, re-emitted verbatim into the reconstructed single-column
+ * PDF (`ats-resume-model.ts`), no longer carries a digit-lead token that
+ * would fail Guard 7 on re-parse (#324) — the column signal that justified
+ * the unguarded recovery is gone in the reconstruction, so the stored
+ * heading must be guard-clean on its own.
+ */
+function stripSidebarNoisePrefix(raw: string): string {
+  const words = raw.trim().split(/\s+/).filter((w) => w.length > 0);
+  let i = 0;
+  while (i < words.length - 1 && !/^[A-Z]/.test(words[i])) i++;
```
Samhit21: [Nit]: `/^[A-Z]/` is ASCII-only, so a section heading whose first real word starts with an accented or non-Latin uppercase letter (e.g. "Éducation", "Проекты") is treated as sidebar noise and stripped.

Recovered anchors are almost always ASCII ATS section names, so this is very unlikely to bite in practice — flagging it only because the same casing assumption silently drops the leading word rather than failing loudly. `\p{Lu}` with the `u` flag would match any Unicode uppercase if you ever want to harden it:

    while (i < words.length - 1 && !/^\p{Lu}/u.test(words[i])) i++;

Non-blocking.

Samhit21 (commented):
## PR Review: fix(roundtrip): sidebar-heading + empty-title role fidelity (#324, #325, #326)

### Summary
Tight, well-tested round-trip cleanup. The empty-title rescue (#325) is the delicate part, and it's handled correctly: `isBareLocationString` gates every rescue path (title + team) on a closed vocabulary (valid USPS code / gazetteer country), not a bare "CapWords, CapWords" shape, so comma-formatted titles like "Buyer, Home Goods" keep their title. I verified the regex capture groups line up with the state/country checks, that the new predicate is strictly narrower than the old shape-only test, and that both rescue directions are covered by the new regression guards. `stripSidebarNoisePrefix` (#324) is scoped to the recovery branch and is a no-op on clean headings. No prior reviews; all findings here are net-new.

### Findings
1. `sections.ts:934` — [Nit, non-blocking] `/^[A-Z]/` is ASCII-only; an accented/non-Latin uppercase heading word would be stripped as noise. Realistically won't bite (recovered anchors are ASCII ATS names). Optional `\p{Lu}/u` hardening.

(Minor: the retained `KNOWN_FAILURES` comment lead-in still reads "each warrants its own follow-up:" but now lists only the #326 entry — trivial doc staleness, ignore or tidy.)

### Verdict
**Action: COMMENT** — no correctness issues found; the single nit is optional. Comfortable seeing this merged as-is.

src/lib/heuristics/sections.ts
s-annam: Fixed in `a83540d` — swapped `/^[A-Z]/` for `/^\p{Lu}/u` so a non-Latin uppercase lead word (Éducation, Проекты) is kept rather than stripped. Thanks for the catch.

Issue #301, opened 2026-07-03
Skills: multi-word skill token splits at line-wrap boundary on round-trip


s-annam:
## What

A **multi-word skill token splits into two** on the parse→export→parse round-trip. The reconstructed "Download PDF" re-parses with a higher skills count than the source, because a skill containing spaces is broken at an interior word boundary.

Surfaced by the round-trip probe (recruidea-internal #29) on a real résumé; reproduced here with a synthetic persona.

## Repro

Source résumé skills line contains a three-word skill, e.g.:

```
Skills: Python · Kubernetes · Cloud Data Warehousing · Terraform · ...
```

Round-trip it (parse1 → render → parse2):

```
RL_RT_PDF=/abs/path/synthetic.pdf npx vitest run src/lib/heuristics/corpus-roundtrip.test.ts
```

**Observed:**

```
skills:
  count 13 → 14
  removed: ["Cloud Data Warehousing"]
  added:   ["Cloud Data", "Warehousing"]
```

The single skill `Cloud Data Warehousing` re-parses as two tokens (`Cloud Data` + `Warehousing`) — split at the **last** interior space, i.e. exactly where the rendered skills line wrapped.

## Suspected mechanism

- Reconstruction joins all skills into **one** middot-delimited header line: `skills.join(" · ")` — `src/lib/pdf/ats-resume-model.ts:325`.
- The renderer word-wraps that line (`text.split(/\s+/)` — `src/lib/pdf/render-ats-pdf.ts:114`). When the wrap boundary falls **inside** a multi-word skill (between `Data` and `Warehousing`), the emitted PDF has that skill broken across two rendered lines.
- On re-parse, the skills parser treats the wrapped continuation as a new token, so one skill becomes two.

Net: any skill whose interior space coincides with the line-wrap point is split. Same family as the existing baselined skills token-split fixtures (`latex/deedy-resume-*.pdf`, `unknown/openresume-react-pdf.pdf`).

## Definition of done

- A three-word skill that wraps mid-token round-trips as **one** skill, not two.
- A synthetic fixture reproducing the wrap-boundary split is added and its `skills` line removed from `KNOWN_FAILURES` (or never baselined) in `src/lib/heuristics/corpus-roundtrip.test.ts`.

Relates to #295–#299 (round-trip corruption family).

s-annam:
**Confirmation** — reproduced independently during the #302/#300/#297 education round-trip epic (PR #303, now merged).

[...]

Issue #166, opened 2026-06-24
Multi-line (3-line) wrapped role header not recognized → role's bullets fall into "Other"


s-annam:
## Symptom

On a role whose header wraps across **three physical lines**, the role never forms an `experience[]` entry, so its bullets fall into the unmatched **"Other"** group instead of attributing to the role.

Surfaced (but explicitly left out of scope) during #162 on the reproducer `tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-multiline-bullets-coursework.pdf`. The "Docent" role header extracts as three lines:

```
Docent … May 2023 - June
Museum
2024
```

Because the title + org + date span three wrapped lines, the entry-block parser doesn't recognize a single role header, so no `experience[]` entry is built for it — and `groupBulletsByExperience` then has no role description to match the Docent bullets against, dropping them into **"Other"**.

## Verification

Confirmed pre-existing and independent of the #162/#163/#164 epic: reverting the epic's edits reproduces the identical "Other" attribution for the Docent bullets. It is **not** caused by — nor fixed by — the wrap-merge (#162), the coursework section boundary (#163), or the column de-interleave (#164).

## Root cause (direction)

`buildEntryBlock` / `parseEntryBlocks` (`src/lib/heuristics/entry-blocks.ts`) recognizes a role header on a bounded number of lines (title line + date). A header that wraps the org name and/or date onto additional physical lines isn't folded back into one logical header, so the block never opens. This is the entry-header analogue of the bullet-wrap problem #162 fixed for bullet bodies — but at the **role-header** layer, where `mergeWrappedContinuations` does not apply (it folds continuation lines onto a prior *bullet*, not onto a role header).

## Fix direction

Fold wrapped role-header continuation lines into the logical header before entry detection — likely an extension of the wrapped-continuation handling, scoped to the header region of an entry block (org/date wraps), so the title, org, and date reassemble into one header and the block opens normally. Then the Docent bullets attribute to the role instead of "Other".

## Acceptance criteria

- [ ] On the `google-docs-skia-proxy-multiline-bullets-coursework.pdf` fixture, a "Docent" `experience[]` entry is formed (title + org + dates reassembled from the three wrapped lines).
- [ ] Its bullets attribute to the Docent role via `groupBulletsByExperience` (no longer in "Other").
- [ ] Targeted test asserting the multi-line wrapped role header reassembles into one entry.
- [ ] No regression in existing corpus snapshot tests.

[...]

PR #329, opened 2026-07-04
fix(roundtrip): achievement wrap, achievement bullet-note, dateless experience (#307, #308, #309)

### LINKED ISSUE #307: Round-trip: Download PDF fragments multi-segment achievement headers into orphan lines
## Summary

The **Download PDF** path fragments any entry header line that packs 3+ middot-joined segments — most visibly an **Achievements** item — emitting each `·`-separated segment on its own line. The keyword and the year get orphaned onto separate lines. The on-screen Reconstructed Resume renders the same content correctly (word-wraps a single `<h3>`), so the corruption is Download-PDF-only. Surfaced via `/roundtrip-probe` on a real résumé; the parse **round-trips clean** — this is purely a rendering defect.

## Observed (rendered Download PDF, synthetic repro)

An achievement `Patent · Issued Patent US0000000B1; long statement that overflows one line at body width · 2019` renders as:

```
Patent
Issued Patent US0000000B1; long statement that overflows one line at body width
2019
```

`Patent` (keyword) and `2019` (year) are each stranded on their own line. A **shorter** achievement that fits one line shows no wrap and looks fine — which hides the bug on most inputs.

## Expected

Word-wrap the header like the on-screen surface does, e.g.:

```
Patent · Issued Patent US0000000B1; long statement that
overflows one line at body width · 2019
```

## Root cause

`src/lib/pdf/render-ats-pdf.ts` — `Layout.wrap()` (≈ line 337) routes **any** line containing `" · "` through `wrapSegmentsToLines`:

```ts
private wrap(text, font, size, maxWidth): string[] {
  if (text.includes(MIDDOT_SEGMENT_SEP)) {
    return wrapSegmentsToLines(text.split(MIDDOT_SEGMENT_SEP)..., ...);
  }
  return wrapWordsToLines(...);
}
```

`wrapSegmentsToLines` keeps each middot segment **atomic** — it was built for the **skills list** (#301), where a multi-word skill (`Cloud Data Warehousing`) must not split mid-segment or it re-parses as two skills. When a line overflows, atomic packing places whole segments on separate lines.

Achievement `headerLine` is assembled in `src/lib/pdf/ats-resume-model.ts:279`:

```ts
headerLine: joinHeader([ach.title, ach.year], " · ") || "Achievement",
```

and `ach.title` **itself** already contains `" · "` (e.g. `"Patent · Issued Patent …"`). So the header carries `keyword · statement · year` — 3 middot segments of **different semantic meaning** (keyword separator vs. year separator). When it overflows one line, atomic wrapping strands `keyword` and `year`.

For header/entry lines the middot is a display joiner, not a re-parse-critical boundary — atomic wrapping helps nobody here and looks broken.

## Fix direction

Apply segment-atomic wrapping **only to the skills entry** (where multi-word-segment integrity is required for re-parse, #301); word-wrap every other header/entry line.

Suggested shape: give `drawText` an explicit wrap mode (e.g. `atomicSegments?: boolean`, default `false`, threaded into `Layout.wrap`), and set it `true` only for the skills entry draw in `render-ats-pdf.ts`. Do **not** switch the decision on `text.includes(" · ")` — that's the bug.

## Verify round-trip stays clean

The affected résumé already round-trips clean (`RL_RT_PDF=… npx vitest run src/lib/heuristics/corpus-roundtrip.test.ts` → `clean`); this change is cosmetic and must keep it clean. The skills round-trip guard `src/lib/pdf/skills-wrap-roundtrip.test.ts` must still pass (skills still use atomic wrap).

## Acceptance criteria

- [ ] Achievement header with `keyword · statement · year` that overflows one line **word-wraps** in the Download PDF — no segment (keyword/year) is stranded alone unless the line genuinely fits on one line.
- [ ] Skills entry still wraps atomically (multi-word skills never split mid-segment) — `skills-wrap-roundtrip.test.ts` passes unchanged.
- [ ] Experience / education / project header + sub-lines are unaffected (they already fit; regression-check they still render one entry per role).
- [ ] Corpus round-trip gate (`corpus round-trip invariants (#293)`) stays green.
- [ ] New unit test over `wrap`/`drawText` (or `renderAtsResumePdf`) asserting a 3-middot-segment header word-wraps rather than emitting one segment per line, using a **synthetic** achievement (no candidate PII).

## Notes

Sibling of the round-trip fidelity batch #295–#299. No candidate PII in this issue — synthetic repro only.

### LINKED ISSUE #308: Reconstructed resume shows 'No bullet-shaped lines detected.' under every achievement (wrong copy)
## Summary

The Reconstructed Resume prints **"No bullet-shaped lines detected."** under **every** achievement. Achievements are single-statement items by design (the statement *is* the content — there are no bullets to detect), so the note is wrong copy on this surface. It reads as a warning where nothing is wrong.

## Observed

Every parsed achievement renders its title/year header followed by the grey note:

```
Patent · Issued Patent US0000000B1; … · 2019
No bullet-shaped lines detected.

Publication · Co-authored N publications on <topic>. · 2009
No bullet-shaped lines detected.
```

## Expected

No such note under a parsed achievement — the achievement statement standing alone is the correct, complete rendering.

## Root cause

`src/components/features/ReconstructedResume.tsx` — `AchievementsSection` (≈ lines 656–672) copies the empty-bullet note from `RoleEntry`:

```tsx
{group && group.bullets.length > 0 ? (
  <ul>…</ul>
) : (
  !added && (
    <p className="text-sm text-content-tertiary">
      No bullet-shaped lines detected.
    </p>
  )
)}
```

`HeuristicAchievement` (`src/lib/score/types.ts:199`) is `{ title, year?, url?, description? }` — a heuristic achievement has **no bullets by design**. The note is a meaningful diagnostic in `RoleEntry` (`src/components/features/ReconstructedRole.tsx:554`), where a parsed experience *role* with zero graded bullets is a real signal the surface exists to expose. On an achievement, empty bullets is the normal case, so the note is noise.

Note the `!added` guard already suppresses the note for user-**added** achievements (an added entry's CTA is the "+ Add bullet" affordance) — parsed achievements just also need it gone.

## Fix direction

Drop the "No bullet-shaped lines detected." note from `AchievementsSection`. Render the bullet `<ul>` only when `group.bullets.length > 0`; otherwise render nothing (still keep the "+ Add bullet" affordance for added entries). Leave `RoleEntry`'s note untouched — it stays correct for experience roles.

## Acceptance criteria

- [ ] A parsed achievement with no bullets renders **only** its title/year header (plus its edit/remove affordances) — no "No bullet-shaped lines detected." note.
- [ ] An achievement that *does* carry bullets (via `description` / added bullets) still renders them.
- [ ] `RoleEntry` (Experience section) still shows "No bullet-shaped lines detected." for a parsed role with zero bullets — unchanged.
- [ ] Added-achievement path unchanged (already suppressed via `!added`).

## Notes

Surfaced alongside the Download-PDF achievement-wrap fidelity issue during a `/roundtrip-probe` run. UI-copy only; no candidate PII.

### LINKED ISSUE #309: Round-trip: dateless experience section collapses to 0 roles after Download-PDF re-parse
## Summary

A résumé whose **experience roles carry no dates** loses its **entire experience section** on the round-trip: `parse1` reads N roles, the reconstructed **Download PDF** is re-parsed, and `parse2` reads **0**. Found via `/roundtrip-probe` on a real (dateless, student-style) résumé; reproduced structurally below with a synthetic persona.

```
hop 1 (parse1 → parse2):  experience: role count 6 → 0
hop 2 (parse2 → parse3):  clean          # nothing left to lose
```

This is a **full-section collapse**, the highest-severity round-trip failure class the probe hunts (sibling of #295–#299) — worse than a field swap.

## Root cause

`extractExperience` (`src/lib/heuristics/extract/experience.ts:33`) splits the section into entry blocks with **only** the `date_range` anchor:

```ts
const blocks = parseEntryBlocks(experience, {
  anchor: "date_range",
  collectBody: true,
  headerLookback: 2,
});
```

`parseEntryBlocks` (`src/lib/heuristics/entry-blocks.ts`) anchors a new role on a line **containing a date range**. When a section has **zero** date-bearing lines, it yields **zero** blocks → **zero** roles.

Why `parse1` still gets roles but `parse2` doesn't:

- The **source** résumé is a denser/two-column layout. `parse1` segments the dateless roles from structural/layout cues (line grouping, header/company/bullet shape) upstream of the date-anchored splitter.
- The reconstructed **Download PDF** normalizes to a flat single-column stack of `title` / `company` / `bullets` — **no dates, no layout cues**. On re-parse, the only role signal left is the date-anchored splitter, which finds nothing.

Confirmed empirically: every parsed role in the repro had `start_date`/`end_date` = `undefined` in **both** `parse1` and the rebuilt model, and the rendered experience text contained no date tokens.

## Fix direction

`parseEntryBlocks` already defines a `"first_line"` anchor (`entry-blocks.ts:167`) built exactly for **date-optional** entries (a non-bullet header line followed by a bullet group starts a new block — used by the projects design). It is currently unexercised.

Make `extractExperience` **fall back to `"first_line"` anchoring when the section carries no date ranges**:

- If `parseEntryBlocks(..., "date_range")` returns 0 blocks (or the section has no date-range line at all), re-run with `anchor: "first_line"` so each `header + bullets` group becomes one dateless role.
- Keep `date_range` as the primary path — dated résumés must not regress (a role's dates are still the strongest anchor when present).
- Preserve the existing date-only-phantom drop and the `title || company` non-empty filter.

## Acceptance criteria

- [ ] A reconstructed experience section with **no dates** (header + bullets per role) re-parses to the **same role count** as `parse1` — round-trip `clean` on the experience category.
- [ ] Dated experience résumés are unchanged (existing corpus round-trip fixtures stay green; no new `KNOWN_FAILURES` line).
- [ ] Unit test in `extract-fields.test.ts`: a dateless experience section with 3 `title` + `company` + bullet groups → 3 `ResumeExperience` rows.
- [ ] Corpus gate: add/keep a synthetic **dateless** experience fixture so this can't silently regress.

## Repro (synthetic — no candidate PII)

Build an experience section of dateless roles and round-trip it:

```
RELEVANT EXPERIENCE
Production Intern
Example Opera Company
• Supported ticketing and stage assignments
• Maintained audio systems for performance halls

Retail Sales Associate
Example Music Retailer
• Matched customers to products against their needs
• Completed the in-house product certification program
```

Expected: 2 roles after the Download-PDF round-trip. Actual: **0**.

_Found via `/roundtrip-probe`; the triggering résumé is a real local file and is **not** committed (PII policy). The corpus fixture added for the acceptance criterion must use a synthetic persona per `tests/fixtures/pdfs/README.md`._

s-annam:
## Summary

Three round-trip fidelity fixes surfaced via `/roundtrip-probe`, accumulated into one commit:

- **#307** (bug) — Download PDF fragmented multi-segment achievement headers (`keyword · statement · year`) into one segment per line on overflow. Fix: added an opt-in `atomicSegments` flag to `AtsEntry`, threaded through `Layout.wrap`/`drawText`; segment-atomic wrapping now applies **only** to the skills entry (where multi-word skills must not split mid-segment, #301), and every other header word-wraps. New pdf unit test; skills atomic-wrap round-trip unchanged.
- **#308** (bug) — Reconstructed resume printed "No bullet-shaped lines detected." under *every* achievement (wrong copy — achievements are single-statement by design). Fix: dropped the note from `AchievementsSection`; bullets render only when present; the `+ Add bullet` affordance is preserved and `RoleEntry`'s experience-section note is untouched.
- **#309** (bug) — Dateless experience sections collapsed to 0 roles after the Download-PDF re-parse. Fix: `extractExperience` falls back to the `first_line` anchor when the `date_range` pass yields 0 blocks; dated résumés never reach the fallback (no regression). New unit test + synthetic PII-free dateless corpus fixture.

Resolves #307
Resolves #308
Resolves #309

## Test plan

- [x] `npm run test` — affected vitest suites green (one pre-existing `dateless-entries` test updated to the new dateless→roles contract this fix overturns)
- [x] `npm run lint` clean
- [x] `verify` pre-push hook green (fallow report-only — all findings pre-existing/re-attributed, no new dupes)
- [x] Fixture personas verified synthetic — no real PII (Step 3.5): Jordan Rivera / `jordan.rivera@example.com` / `(312) 555-0142`, clean PDF metadata

## Adversarial review

`ecc:code-reviewer` ran an independent adversarial pass over the whole accumulation — **1 round, 0 blocking findings, 0 nits → APPROVE**. It did hands-on round-trip repro of the two highest-risk paths:

- **#307 wrap-atomicity narrowing** — confirmed the `subLine "Company · Dates"` path still round-trips dates correctly (the pre-existing `mergeWrappedHeaderRows` fold, #166, reassembles a wrapped date range before anchor detection); education headers join dates with double-space, never depended on atomic wrapping.
- **#309 fallback trigger** — confirmed the fallback fires only when the `date_range` pass yields exactly 0 blocks, so any résumé with ≥1 dated role never reaches it; date-only-phantom drop and `title || company` filter still apply on the fallback path.

No correctness bugs, regressions, or unmet acceptance criteria found.


src/lib/pdf/render-ats-pdf.ts
```diff
@@ -476,7 +495,11 @@ function drawSectionHeading(layout: Layout, heading: string) {
 
 function drawEntry(layout: Layout, entry: AtsEntry, mutedColor: RGB) {
   if (entry.headerLine) {
-    layout.drawText(entry.headerLine, { bold: true, size: SIZE_HEADER });
+    layout.drawText(entry.headerLine, {
+      bold: true,
+      size: SIZE_HEADER,
+      atomicSegments: entry.atomicSegments,
```
rohithgollapalli, line 503: Low-severity / non-blocking (note applies to the `entry.subLine` `drawText` just below, line ~506, which is unchanged so I could not anchor there): `atomicSegments` is threaded to the header line but not the sub-line, so `"Company · Location"` / `"Institution · Location"` sub-lines — called out explicitly in the old `wrap()` doc comment — now word-wrap where they were previously atomic. Harmless in the common case (sub-lines rarely exceed one line), and the date is still reassembled by `mergeWrappedHeaderRows` (#166); but a long overflowing sub-line could break mid-location and re-parse the location into fragments. No test covers a wrapping sub-line, so this is an under-tested edge, not a demonstrated bug. Consider a one-line note that the narrowing is intentional for sub-lines too, or a guard test to lock it in.

rohithgollapalli (dismissed):
Three well-constructed, independently-tested round-trip fixes. I verified the highest-risk paths and found no blocking correctness issues:

- **#309 (dateless fallback)** does not double-extract — there's no separate dateless source path (`dateless-entries` is test-only) and `extractExperience` has a single caller (`openresume.ts:240`). The fallback fires only when the `date_range` pass yields exactly 0 blocks, so any résumé with ≥1 dated role never reaches it; the `title || company` filter and date-only-phantom drop apply on both paths.
- **#307 (wrap atomicity)** preserves the #301 skills fix — all entries render via `drawEntry`, and the skills entry (the only `atomicSegments: true` caller) still wraps atomically while every other header word-wraps. The new test correctly asserts no stranded lone keyword/year segments.
- **#308** is a clean JSX simplification; `added` is still referenced by the add affordance.
- **Fixture** is synthetic and PII-compliant — I extracted the binary with `pdftotext` and eyeballed the persona (Jordan Rivera / jordan.rivera@example.com / (312) 555-0142, `Example …` companies), per the non-negotiable policy.

One low-severity, non-blocking note left inline on `drawEntry`: `atomicSegments` is threaded to the header line but not the sub-line, so `Company · Location` sub-lines now word-wrap where they were previously atomic — harmless in the common case but under-tested for a long overflowing sub-line. Not a blocker.

LGTM 🎉

src/lib/pdf/render-ats-pdf.ts
s-annam: Fixed in `5693698` — good catch, this was a real #301 regression, not just an under-tested edge. #307 flipped middot atomic-wrapping to opt-in (default off) so 3-segment achievement headers stop stranding a lone keyword/year, but the sub-line `drawText` call passed no `atomicSegments`, so `Company · Location  Dates` / `Institution · Location  Dates` sub-lines lost the #301 protection you flagged.

Sub-lines are always semantic org·location/org·date lines (never a 3+ display-joiner list), so I opted them into `atomicSegments` unconditionally — the inverse decision from the achievement header. Also corrected the `wrap()` doc comment that wrongly grouped "Company · Location" sub-lines with the display-joiner headers, and added a guard test that an overflowing multi-word location (`San Francisco Bay Area`) stays intact on one wrapped line.

PR #331, opened 2026-07-05
fix: de-track letter-spaced name headings before parsing

### LINKED ISSUE #330: Letter-spaced name heading fails full_name parse (name exploded to per-glyph tokens)
## Summary

A résumé whose **name heading uses wide letter-spacing** (tracked-out type — a
common design choice) fails to parse `full_name`. The name comes back empty or,
worse, a wrong fallback token (e.g. a skill) is picked as the name.

Contact scalars on the same résumé (email, phone, location) parse correctly —
**only the name drops.**

## Reproducer

Synthetic, PII-free fixture: `tests/fixtures/pdfs/unknown/letter-spaced-name-heading.pdf`
(persona `JORDAN REYES` / `jordan.reyes@example.com` / `(312) 555-0123` /
New York, NY). Its `.expected.json` shows `full_name` **absent** from
`fieldsPopulated` — the visible reproducer.

Triage with the `probe-contact` skill:

```
RL_CONTACT_PDF=/abs/path/to/letter-spaced-name-heading.pdf \
  npx vitest run src/lib/heuristics/probe-contact.test.ts
```

Output (abridged):

```
Extracted fields:
  full_name   null @ 0
  email       "jordan.reyes@example.com" @ 0.98
  phone       "(312) 555-0123" @ 0.85
  location    "New York, NY" @ 0.75

Profile region scanned:
  | J O R D A N R E Y E S          <- name exploded to per-glyph tokens
  | jordan.reyes@example.com | (312) 555-0123 | New York, NY
  ...
```

## Root cause — two layers

1. **Line assembly** — `src/lib/heuristics/sections.ts` `mergeItemText`: a space
   is inserted whenever `gap > avgCharW * 0.4`. A tracked-out heading renders
   each glyph as a separate pdfjs text item whose inter-glyph advance exceeds
   `0.4 * charWidth`, so a space is inserted between **every glyph**:
   `JORDAN REYES` → `J O R D A N R E Y E S`. (This is a real glyph-advance
   artifact — `pdftotext` explodes the same heading — not solely our bug.)

2. **Name heuristic** — `src/lib/heuristics/extract/name.ts` `nameCandidateWords`:
   `words.length > 5 → null`. The exploded heading is 8–10 single-char tokens, so
   it is rejected outright; `extractName` then either returns nothing or scores a
   wrong nearby line as the name.

## Suggested fix directions

- **Preferred (layer 1):** detect a run of ≥N consecutive single-character
  tokens on one line during/after line assembly and collapse it
  (`J O R D A N` → `JORDAN`), so every downstream extractor — not just the name
  — sees the de-tracked text. A letter-spaced heading is the general case; the
  name is just where it bites first.
- **Narrow (layer 2):** before the `>5 words` reject in `nameCandidateWords`,
  collapse an all-single-char token sequence into one word. Fixes the name only.

Include a regression test asserting `full_name === "Jordan Reyes"` on the fixture,
and drop the fixture's `.expected.json` from any known-failure baseline once fixed.

## Not in scope

Contact scalars (email/phone/location) already parse correctly on this layout;
this issue is scoped to the name field / the letter-spacing artifact.

s-annam:
## Summary

A résumé whose name heading uses wide `letter-spacing` reaches the parser as
single-letter runs joined by spaces inside one pdfjs text item
(`"J O R D A N"`), so the `full_name` heuristic reads per-glyph tokens and drops
the name. This de-tracks such runs before parsing.

- add `collapseLetterSpacing()` in `sections.ts`: collapses a maximal run of
  ≥4 single letters (Unicode `\p{L}`, so accented names de-track too) separated
  by exactly one space back into one word, anchored so a trailing multi-char
  word isn't swallowed
- apply it per-item in `mergeItemText()` so real word boundaries (separate
  items, or a ≥2-space gap within an item) are preserved while each tracked-out
  word de-tracks; item geometry is untouched
- unit tests for the collapse + its boundary/accent/threshold cases
- remove the now-fixed `KNOWN_FAILURES` round-trip baseline for this case and
  re-bake the `letter-spaced-name-heading` fixture snapshot (`full_name` now
  populated)

Known scope limit (commented in code): a whole multi-word heading emitted as a
*single* item with a *single*-space word gap can't be split from the string
alone — not observed for pdfjs on the corpus.

Closes #330

## Test plan

- [ ] `npm run test` green
- [ ] `npm run verify` green (ran locally: EXIT=0, 1506 passed)

## Adversarial review

One skeptic pass (opus) on the diff, pre-PR. 2 blocking findings, both fixed; 2 nits accepted with scoping.

- **blocking (fixed)** — ASCII-only `[A-Za-z]` corrupted accented names (`A N D R É S` → `ANDR É S`). Switched the run regex to Unicode `\p{L}` + `u` flag; added accented-name tests.
- **blocking (downgraded on verification)** — claimed a single-item single-space heading welds both words. True only for the pure single-item + single-space-word-gap shape; the regex requires exactly one space per pair, so a ≥2-space word gap already ends the run and is preserved. Added a boundary test + a scoping comment for the genuinely-ambiguous residual (not observed for pdfjs on the corpus).
- **nit (accepted)** — double-space *intra-word* tracking won't collapse; and a genuine ≥4 single-letter run (e.g. musical notes `C D E F G`) welds. Threshold 4 is a defensible precision/recall tradeoff; documented.

### DIFF
--- a/src/lib/heuristics/sections.ts
+++ b/src/lib/heuristics/sections.ts
@@ -487,6 +487,47 @@ function groupLinesSingle(bandItems: PdfTextItem[]): PdfLine[] {
 }
 
 /**
+ * Minimum single-letter run length that reads as letter-spacing (tracked-out
+ * type) rather than genuine single-char tokens. Four keeps initials ("J R R"),
+ * roman numerals ("I V X"), and short spaced acronyms out of the collapse.
+ */
+const LETTER_SPACING_MIN_RUN = 4;
+
+/**
+ * Regex for a maximal run of ≥`LETTER_SPACING_MIN_RUN` single letters each
+ * separated by exactly one space (`J O R D A N`). `\p{L}` (Unicode letter, `u`
+ * flag) so accented names de-track too (`A N D R É S` → `ANDRÉS`), not just
+ * ASCII. Anchored on both sides by a non-letter (or string edge) so a trailing
+ * multi-char word isn't swallowed ("J O R D A N Reyes" → only "J O R D A N"
+ * collapses). Requiring exactly one space per pair means a wider (≥2-space)
+ * inter-word gap ends the run, preserving that word boundary even inside one
+ * item (`"J O R D A N  R E Y E S"` → `"JORDAN REYES"`).
+ */
+const LETTER_SPACED_RUN = new RegExp(
+  `(?<!\\p{L})(?:\\p{L} ){${LETTER_SPACING_MIN_RUN - 1},}\\p{L}(?!\\p{L})`,
+  "gu",
+);
+
+/**
+ * Collapse letter-spaced (tracked-out) runs inside one pdfjs item string.
+ * A heading rendered with wide `letter-spacing` reaches us as glyphs joined by
+ * spaces *within a single item* (`"J O R D A N"`), while genuine word breaks
+ * arrive as separate items — so collapsing per item de-tracks each word yet
+ * preserves the real word boundary between items (#330). Every downstream
+ * extractor (name, contact, sections) then sees `"JORDAN"`, not `"J O R D A N"`.
+ *
+ * Scope: this recovers the word boundary when it surfaces as a separate item
+ * (the observed pdfjs shape) or as a ≥2-space gap within an item. The one case
+ * it cannot resolve is a whole multi-word heading emitted as a *single* item
+ * with a *single*-space word gap — then intra- and inter-word gaps are
+ * indistinguishable from the string alone and the words would weld. Not
+ * observed for pdfjs on the #330 corpus; a gap-magnitude split would be needed.
+ */
+export function collapseLetterSpacing(str: string): string {
+  return str.replace(LETTER_SPACED_RUN, (run) => run.replace(/ /g, ""));
+}
+
+/**
  * Concatenate items on a line, inserting a space when the horizontal gap
  * between runs is large enough to imply a word boundary. pdfjs emits each
  * glyph run as a separate item, so naively joining with spaces over-pads
@@ -494,7 +535,11 @@ function groupLinesSingle(bandItems: PdfTextItem[]): PdfLine[] {
  */
 export function mergeItemText(items: PdfTextItem[]): string {
   if (items.length === 0) return "";
-  let out = items[0].str;
+  // De-track each item first (geometry left untouched — gap math below still
+  // uses the original item widths). Collapsing per item keeps the word-boundary
+  // items intact, so `"J O R D A N"` + `" "` + `"R E Y E S"` → `"JORDAN REYES"`.
+  const strs = items.map((it) => collapseLetterSpacing(it.str));
+  let out = strs[0];
   for (let i = 1; i < items.length; i++) {
     const prev = items[i - 1];
     const cur = items[i];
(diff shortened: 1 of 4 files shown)


src/lib/heuristics/sections.ts
```diff
@@ -494,18 +535,22 @@ function groupLinesSingle(bandItems: PdfTextItem[]): PdfLine[] {
  */
 export function mergeItemText(items: PdfTextItem[]): string {
   if (items.length === 0) return "";
-  let out = items[0].str;
+  // De-track each item first (geometry left untouched — gap math below still
+  // uses the original item widths). Collapsing per item keeps the word-boundary
+  // items intact, so `"J O R D A N"` + `" "` + `"R E Y E S"` → `"JORDAN REYES"`.
+  const strs = items.map((it) => collapseLetterSpacing(it.str));
```
Samhit21, line 541: [Question]: De-tracking is applied to every item in every line via `mergeItemText`, not just heading-band items — was global scope intentional over gating on font size?

The threshold-4 guard keeps this safe in practice (I checked: `AWS GCP SQL` and normal prose are untouched; only genuine ≥4 single-letter runs weld, e.g. `C R U D` → `CRUD`, which is usually what you'd want anyway). So this isn't a bug — just confirming you preferred the broader de-tracking over scoping to large-font/heading bands, since letter-spacing overwhelmingly shows up in headings. Global + threshold-4 is a defensible call; leaving it to you.

Samhit21 (commented):
## PR Review: fix: de-track letter-spaced name headings before parsing

### Summary
Clean, well-tested fix for #330. The `collapseLetterSpacing` regex is correctly anchored — verified the trailing-word anchor (`"J O R D A N Reyes"` → `"JORDAN Reyes"`), the `\p{L}`+`u` accent handling, the ≥2-space boundary preservation, and that item geometry is untouched (gap math still uses original widths; only output text + ws-checks use the de-tracked strings). Ran the regex against realistic body text: multi-char acronyms and prose are untouched, so the threshold-4 precision/recall tradeoff the author documented holds up. The fixture re-bake is consistent. No prior reviews; all findings net-new.

### Findings
1. `sections.ts:541` — [Question, optional] De-tracking runs on all items, not just heading bands. Safe given threshold-4; raised only to confirm global scope is intentional vs font-size gating. Not blocking.

### Verdict
**Action: COMMENT** — no correctness issues; the single item is an optional design question. Comfortable seeing this merged as-is.

src/lib/heuristics/sections.ts
s-annam: Confirmed — global de-tracking via `mergeItemText` (not gated on font size / heading band) was intentional. Rationale: letter-spacing that trips the parser shows up in more than just large-font headings (invented section labels, sidebar name blocks, some contact lines all render tracked-out), and font-size gating would need a reliable heading-band classifier at this layer that we don't have here. The threshold-4 `LETTER_SPACING_MIN_RUN` guard is exactly what keeps the broad scope safe — `AWS GCP SQL` and prose are untouched; only genuine ≥4 single-letter runs weld (`C R U D` → `CRUD`), which is the desired recovery. Good check on the blast radius; keeping global + threshold-4.

PR #336, opened 2026-07-05
feat: download-PDF guardrail, Poppins embed, experience header fix

### LINKED ISSUE #310: Parser absorbs a second experience-category section header as a role's company
## Summary

When a résumé has **two experience-category sections** (e.g. `RELEVANT EXPERIENCE` followed by `INVOLVEMENT EXPERIENCE`), the parser absorbs the **second section header as a role's company**. The heuristic role that starts right after the second header takes the header line above it as its `company` (via the header-lookback disambiguation), instead of treating that line as a section boundary.

Downstream effect on the Download-PDF round-trip: that company field re-renders as plain body text, and on re-parse the string is recognized as a section header again (anchor fallback → `experience`), corrupting section segmentation.

Surfaced via `/roundtrip-probe` alongside #309 (the dateless-experience collapse). This is a **separate** defect — scrubbing the mis-attributed company alone did **not** restore the roles, so it is not the collapse cause, but it is a real parse mis-attribution worth fixing on its own.

## Root cause

Role entry blocks map their header lines to `title`/`company` via `disambiguateCompanyTitle` with a 2-line lookback (`headerLookback: 2`, `src/lib/heuristics/extract/experience.ts:36`). The lookback treats **any** non-bullet line above the anchor as a title/company candidate — including a line that is itself a **recognized section header**.

`matchSectionHeaderDetailed` (`src/lib/heuristics/regex.ts:413`) already recognizes `INVOLVEMENT EXPERIENCE` (and `RELEVANT EXPERIENCE`, `PROFESSIONAL EXPERIENCE`, etc.) as an `experience` header via the head-noun anchor fallback. But the entry-block header lookback never consults it, so a second same-category header directly above a company-less role gets consumed as that role's company.

Observed in the model: the role's `company`/subLine literally equaled the second section header string.

## Fix direction

In the header-lookback path (`disambiguateCompanyTitle` / `parseEntryBlocks` header collection), **reject a candidate header line that `matchSectionHeader()` recognizes as a section header** — it is a boundary, not a company/title. A rejected line should:

- not be pulled into the role below it, and
- ideally split the section so a second experience-category header starts a fresh entry group rather than being swallowed.

This composes cleanly with #309 (the `first_line` dateless fallback): once dateless roles segment on header/bullet structure, honoring section-header boundaries there keeps a second `… EXPERIENCE` header from contaminating the role beneath it.

## Acceptance criteria

- [ ] A role immediately following a second experience-category section header does **not** take that header as its `company`.
- [ ] The second section header does not appear as body/company text in the reconstructed Download PDF.
- [ ] Unit test in `extract-fields.test.ts`: an experience section containing an inner `INVOLVEMENT EXPERIENCE` line above a company-less role → the role's `company` is empty (not the header string).
- [ ] No regression to single-header experience sections (existing corpus round-trip fixtures stay green).

## Repro (synthetic — no candidate PII)

```
RELEVANT EXPERIENCE
Composer
Example Festival
• Composed for an interactive-technology performance group

INVOLVEMENT EXPERIENCE
Student Composers Association
• Created work ranging from symphonic to electronic music
```

Expected: the `Student Composers Association` role has an empty `company`. Actual: its `company` = `"INVOLVEMENT EXPERIENCE"`.

_Found via `/roundtrip-probe` on a real local résumé that is **not** committed (PII policy). Any fixture added must use a synthetic persona per `tests/fixtures/pdfs/README.md`._

### LINKED ISSUE #312: Warn before Download PDF when critical fields (name / contact / experience) are missing
## Problem

"Download PDF" on the reconstructed résumé exports unconditionally, even when the parse dropped critical fields (no name, no contact method, no experience). A user can download an ATS-friendly PDF that is missing their own name and never be told. The only current gate is `disabled={isGenerating}` (a double-click guard) at `ReconstructedResume.tsx:884`; `useDownloadPdf.ts` has no pre-checks and silently falls back to `resume-ats.pdf` when `full_name` is empty (`useDownloadPdf.ts:35`).

This is a **feature / UX guardrail**, not a parser bug. The parse is correctly reporting what a generic ATS reads back — a missing name is often *the finding itself*. So the guard must **inform and offer a fix, not block** the "see what the ATS reads" message.

## Behavior — pre-download checklist popover

On Download click, if any critical field is missing, open a small popover instead of downloading immediately:

```
Missing before download:
  • Name
  • Contact (email or phone)
  ( Fix now )        ( Download anyway )
```

- **Fix now** (secondary) — dismiss popover, scroll/focus the first gated `EditableField` so the user corrects it inline. Re-clicking Download re-checks.
- **Download anyway** (primary) — proceed with the existing `download()` path unchanged.
- If nothing critical is missing, Download click behaves exactly as today (no popover, immediate export).

## Critical set

A field is "missing" if gated `absent` / `low_confidence` (name) or empty (contact/experience):

1. **Name** — `full_name` absent or low-confidence. NOTE: there is no `first_name`/`last_name` in the data model; identity is a single `full_name` string (`score/types.ts:217`). Reuse the existing gated-Name signal.
2. **At least one contact method** — email OR phone present. Flag only if *both* absent.
3. **At least one experience entry** — `result.parsed` experience array non-empty.

## Reuse — signal already exists

The missing-field detection is largely done; this is mostly wiring:

- `src/lib/contact.ts:144-153` — `contactCompleteness(...)` returns `missing` = gated required rows (Name included).
- `src/components/features/ReconstructedResume.tsx:802-804` — `contactMissing` already computed on the same override-applied path the ATS model uses, surfaced today in `AttentionStrip` (`898-899`). The download handler just doesn't consult it.
- `src/lib/heuristics/types.ts:132-151` — `FieldConfidence` map (per-field 0..1, includes `full_name`).
- `EditableField` (existing inline-edit primitive) is the "Fix now" target — no new edit affordance needed.

Gap to build: (a) extend the completeness check to also cover the email-OR-phone rule and the experience-non-empty rule (name is already there); (b) a popover/dialog surface — reuse the existing Dialog primitive from `@design-system`, do not hand-roll; (c) gate `download()` in `useDownloadPdf` / the button `onClick` behind the check.

## Out of scope

- No hard block / disabled Download button — rejected as fighting the product premise.
- No change to the parse, scoring, or the PDF render itself.
- Definition of "critical" is fixed to the three above; no per-field config UI.

## Acceptance

- [ ] Download click with a missing critical field opens the checklist popover listing exactly the missing items.
- [ ] "Download anyway" produces the identical PDF the current path produces.
- [ ] "Fix now" focuses the first gated field inline; correcting it and re-clicking Download exports with no popover.
- [ ] Download click with all three critical fields present exports immediately (no regression, no popover).
- [ ] Popover reuses the `@design-system` Dialog primitive and `Button` primitive — no raw `<button>`, no hardcoded colors.
- [ ] Unit test for the completeness predicate (name / one-contact / experience combinations).

### LINKED ISSUE #314: Embed Poppins in the Download PDF for higher-quality output (zero-egress, Helvetica fallback)
## Problem

The reconstructed-résumé **Download PDF** (`src/lib/pdf/render-ats-pdf.ts`) draws every glyph with pdf-lib's built-in **Helvetica / Helvetica-Bold** (`StandardFonts`). No custom font is embedded. Poppins is already the product's brand font on screen (`@fontsource/poppins`, loaded in `src/main.tsx`), so the downloaded PDF is the **only** surface still rendering in Helvetica — it looks visibly off-brand and lower-quality than the rest of the app.

Embedding Poppins uses the **same pdf-lib engine** already in use; only the font handles change. The font embed is the single biggest readability lever — everything else (layout) is secondary polish tracked as follow-ups (see Out of scope).

**This issue is Phase A only: embed Poppins in the Download PDF, with a safe Helvetica fallback.** It is deliberately scoped small and low-risk.

## Non-negotiable constraint: zero egress

resumelint's core promise is **no bytes leave the browser** (documented in `CLAUDE.md` / README Telemetry). A tempting shortcut is to fetch Poppins from a font CDN (e.g. `fonts.gstatic.com`) at download time — **do NOT do that.** That is network egress + IP disclosure to a third party and would break resumelint's guarantee.

**The Poppins TTF bytes must be bundled locally** (Vite asset), never fetched from a CDN.

### Font-file gotcha

`@fontsource/poppins` is already a dependency, **but it ships only `.woff` / `.woff2`** (verified: `node_modules/@fontsource/poppins/files/` has no `.ttf`). **fontkit cannot parse woff2** (Brotli-compressed). So the existing dep does not give us usable bytes for pdf-lib — we must vendor the raw **TTF** files.

## Current behavior (files + lines)

- `src/lib/pdf/render-ats-pdf.ts:418-424` — `renderAtsResumePdf()` calls `doc.embedFont(StandardFonts.Helvetica)` / `HelveticaBold`. No `registerFontkit`, no custom embed.
- `src/lib/pdf/render-ats-pdf.ts:166-205` — `toWinAnsi()` sanitizes all text to the Windows-1252 subset because StandardFonts can only encode WinAnsi; any glyph outside it degrades to `?`. Embedded Poppins encodes full Latin-Extended, so this degradation is unnecessary on the embedded path (bonus Unicode fidelity).
- `src/lib/pdf/load-pdf-lib.ts` — the `loadPdfLibOnce()` dynamic-import shim for pdf-lib. Fontkit will need to be added here.
- `package.json` — has `pdf-lib@^1.17.1`, `@fontsource/poppins@^5.2.7`. **Missing `@pdf-lib/fontkit`.**

## Embed pattern

Standard pdf-lib custom-font embedding — register fontkit, embed the vendored bytes, fall back to StandardFonts on any failure:

```ts
doc.registerFontkit(fontkit);
if (fontBytes) {
  try {
    const regular = await doc.embedFont(fontBytes.regular);
    const bold = await doc.embedFont(fontBytes.bold);
    return { regular, bold };
  } catch (err) {
    console.warn("custom font embed failed, falling back to Helvetica:", err);
  }
}
// fallback
const regular = await doc.embedFont(StandardFonts.Helvetica);
const bold = await doc.embedFont(StandardFonts.HelveticaBold);
return { regular, bold };
```

## Implementation plan

1. **Add dependency.** `npm install @pdf-lib/fontkit` (runtime dep). It ships no `.d.ts` — type as `unknown` and pass to `registerFontkit`.

2. **Vendor Poppins TTFs.** Add `Poppins-Regular.ttf` (weight 400) and `Poppins-Bold.ttf` (weight 700) under `src/assets/fonts/`. Source: the canonical Google Fonts / Poppins GitHub release TTFs (OFL-1.1). ~340 KB total for the two.
   - Add the Poppins **OFL-1.1 license file** alongside (`src/assets/fonts/Poppins-OFL.txt`).
   - Add a NOTICE line crediting Poppins (OFL-1.1, © Indian Type Foundry / Jonny Pinhorn).

3. **Load the bytes as a bundled asset.** Import each TTF via Vite `?url` (same mechanism `src/main.tsx` uses for the pdfjs worker) and `fetch()` the local URL to `ArrayBuffer` at download time — the fetch hits the app's own origin (the bundled asset), **not** an external host, so zero-egress holds. Cache the bytes module-scoped so repeat downloads don't re-fetch.

4. **Add fontkit to the pdf-lib shim.** Extend `loadPdfLibOnce()` in `src/lib/pdf/load-pdf-lib.ts` to also dynamic-import `@pdf-lib/fontkit` and expose it on `PdfLibParts`, so the entry chunk stays small (fontkit only loads when a download happens).

5. **Embed in the renderer.** In `renderAtsResumePdf()` (`render-ats-pdf.ts:415`), replace the two `embedFont(StandardFonts.*)` calls with a `loadFonts()` helper: `registerFontkit`, try embedding the vendored Poppins bytes, fall back to Helvetica on any failure. Keep the `{ regular, bold }` shape the `Layout` class already consumes — no draw-code changes.

6. **Gate `toWinAnsi` on the embedded path.** When Poppins embedded successfully, skip `toWinAnsi()` (Poppins encodes the glyphs directly). When falling back to Helvetica, keep `toWinAnsi()` (StandardFonts still require it). Simplest: thread an `isEmbedded` / `sanitize` flag from `loadFonts()` into `Layout.drawText`. Do **not** remove `toWinAnsi` — the fallback path still needs it.

7. **Round-trip check.** `render-ats-pdf.ts` output is re-parsed by our own cascade (the #284 round-trip constraint). A pure font swap should not change text geometry/order, but confirm with `/roundtrip-probe` on a few fixtures that the parse-back is unchanged. Update `render-ats-pdf.test.ts` / `render-roundtrip.repro.test.ts` expectations only if the embed legitimately shifts wrap widths (Poppins metrics differ from Helvetica → wrap points can move; adjust wrap-related assertions, not field values).

## Acceptance criteria

- [ ] `@pdf-lib/fontkit` added to `package.json` dependencies.
- [ ] `Poppins-Regular.ttf` + `Poppins-Bold.ttf` vendored under `src/assets/fonts/` with the OFL-1.1 license file; NOTICE updated.
- [ ] A downloaded PDF, checked with `pdffonts <file>.pdf`, reports embedded `Poppins` (Regular + Bold), **not** Helvetica.
- [ ] **Zero egress preserved:** no new external network request during download. Verify via DevTools Network tab (or a test asserting the font URL resolves to the app origin / bundled asset, not an external host). No `fonts.gstatic.com` (or any font-CDN) reference anywhere in `src/`.
- [ ] Poppins load failure falls back to Helvetica and still produces a valid PDF (no throw) — covered by a test that forces the embed to fail.
- [ ] `toWinAnsi()` still applied on the Helvetica fallback path; skipped on the embedded path. A Latin-Extended glyph (e.g. `→` or `ł`) that Helvetica degrades to `?` renders correctly when Poppins is embedded.
- [ ] `/roundtrip-probe` on the standard fixtures shows unchanged parse-back (no regression in `corpus-roundtrip` / `corpus` snapshot tests beyond intentional wrap-metric adjustments).
- [ ] fontkit + Poppins bytes are dynamic-imported / lazily fetched — the entry chunk size does not grow (they load only on download).
- [ ] `npm run verify` green (typecheck → lint → coverage → build → fallow).

## Out of scope (follow-up issues)

Layout/alignment polish — **not** in this issue, file separately after Phase A lands, each gated by `/roundtrip-probe` since they touch parse-back:

- Centered header block (name / title / `City | email | phone` / links) vs. current left-aligned.
- Right-aligned date row (`Role · Company, Loc ······ 01/2025 – Present`) + a team→company drop wrap ladder.
- Inline `**bold**` markdown segments inside bullets.
- Clickable URI link annotations on contact URLs.
- Keep-together pagination (measure entry height, page-break before an entry that won't fit).

s-annam:
## Summary

Epic bundling three issues into one accumulated commit (run via `/implement-epic`), each reviewed by an adversarial pass before this PR:

- **#310 (fix)** — Parser absorbed a second experience-category section header (e.g. `INVOLVEMENT EXPERIENCE`) as the following role's `company` via the entry-block header lookback. Fixed with a **contiguous leading-run** boundary scan: only a section-header line that *leads* the entry block is treated as a boundary and stripped, so a genuine 2nd header is caught while a mid-block role title that happens to be a short anchor-noun phrase (e.g. `Clinical Research Experience`, company above it) is preserved.
- **#312 (feat)** — Pre-download checklist popover (`DownloadGateDialog`, built on `@design-system` `Dialog` + `Button`) surfaces when name / contact (email OR phone) / experience is missing from the parse. `criticalDownloadGate` in `contact.ts` backs the check, gated onto the Download button's `onClick`. Non-blocking by design — "Download anyway" produces the byte-identical PDF; "Fix now" focuses the first gated field.
- **#314 (feat)** — Embed Poppins in the downloaded PDF with a Helvetica fallback. TTFs are bundled locally (Vite `?url`) and fontkit is lazy-loaded, so **zero egress** holds (no font-CDN request). `toWinAnsi` sanitization is skipped on the embedded path, kept on the fallback path. NOTICE updated for the OFL-1.1 Poppins font.

Resolves #310
Resolves #312
Resolves #314

## Test plan

- [x] `npm run test` green (1519 passed, 2 skipped)
- [x] `npm run verify` green on pre-push (typecheck → lint → coverage → build → fallow)
- [x] Manually verified Download-PDF flow in `npm run dev` / `npm run preview`

## Adversarial review

Run via `/implement-epic` with a bounded adversarial review loop (2 rounds) over the accumulated diff before this PR was opened.

**Round 1 — 1 blocking finding (fixed):**
- **#310** — the initial fix rejected *any* `matchSectionHeader()` hit as a section boundary. That matcher falls through to the fuzzy anchor-fallback (any ≤4-word Title-Case/ALL-CAPS phrase ending in an anchor noun), so a legitimate role title like `Clinical Research Experience` (company line above, dates below) was misclassified as a boundary and the **whole role was dropped**. Reviewer verified with a before/after repro.
- **Fix:** replaced the "slice at the last match anywhere" scan with a **contiguous leading-run** scan — only header lines contiguous from index 0 are boundaries. Distinguishes a genuine leading 2nd header from a mid-block title. Added a negative test (role preserved) + a single-header no-regression test.

**Round 2 — clean:** round-1 finding confirmed resolved; no new false-negative introduced (stacked-header and negative-`anchorIdx` re-basing cases traced and safe). #312 (download-gate logic, byte-identical "Download anyway" path) and #314 (no external-host fetch path, exception-safe Helvetica fallback, correct `toWinAnsi` gating, lazy fontkit) reviewed clean in round 1 and unchanged since.

**Surviving nit (non-blocking, LOW):** a genuine section header at block index 1+ when index 0 is non-header content is intentionally *not* stripped by the leading-run scan — the windower's construction doesn't produce that shape, so it's out of scope, not a regression. Also noted: `loadPoppinsBytes()` caches a rejected promise for the session on a transient font-fetch failure (correctness + zero-egress unaffected).


src/lib/pdf/render-ats-pdf.ts
```diff
@@ -209,6 +214,69 @@ type Doc = Awaited<ReturnType<PdfLibParts["PDFDocument"]["create"]>>;
...
+  parts: PdfLibParts,
+): Promise<{ regular: PdfFont; bold: PdfFont; isEmbedded: boolean }> {
+  try {
+    // `@pdf-lib/fontkit` ships no usable default-export `.d.ts` shape, so it
+    // is typed `unknown` in `PdfLibParts` and cast here at the one call site
+    // that hands it to pdf-lib's `registerFontkit` — the narrowest possible
+    // untyped surface, rather than threading `any` through load-pdf-lib.ts.
+    doc.registerFontkit(parts.fontkit as Parameters<Doc["registerFontkit"]>[0]);
+    const bytes = await loadPoppinsBytes();
+    const regular = await doc.embedFont(bytes.regular);
```
Samhit21: [Suggestion]: Embed the Poppins TTFs subsetted to avoid bloating every exported PDF.

`doc.embedFont(bytes.regular)` / `bytes.bold` embed the *entire* font files (no `subset` option defaults to `false` in pdf-lib), so each downloaded résumé carries the full Poppins Regular + Bold — on the order of a few hundred KB added to every PDF, for a document that uses maybe 60–80 glyphs. fontkit is already registered, so subsetting is a one-word change and strictly smaller output:

```ts
const regular = await doc.embedFont(bytes.regular, { subset: true });
const bold = await doc.embedFont(bytes.bold, { subset: true });

Subsetting is orthogonal to the skip-toWinAnsi path — it just prunes unused glyphs, so nothing about the embedded-encoding logic changes.

File: src/lib/heuristics/extract/experience.ts [hunk 2]
```diff
@@ -390,9 +396,43 @@ function disambiguateCompanyTitle(
...
+  // inner section header LEADS the entry block (the role's own title/company/date
+  // follow below it). Matching only the contiguous leading run of header lines
+  // distinguishes the two: the genuine boundary is stripped, the mis-flagged
+  // mid-block title is kept so its role is not dropped (#310 false-positive).
+  // Everything at or above that leading boundary is the previous entry or the
+  // boundary itself, so keep only the lines below it as this role's header.
+  let leadsFreshEntry = false;
+  let lastBoundary = -1;
+  for (let i = 0; i < filtered.length; i++) {
+    if (matchSectionHeader(filtered[i]) === null) break;
```
Samhit21, line 426: [Question]: Does the leading-run strip still drop a legitimate role title that *leads* its block and happens to read like a section header?

The round-1 fix (leading-run vs match-anywhere) correctly rescues the mid-block case — `Clinical Research Experience` sitting *below* its own company line. But the guard's premise is "a legit title that matches the anchor-fallback is always mid-block." This function also handles the stacked `Title \n Company · Dates` shape (reconstructed export / two-column source), where the title *leads* at index 0. A real title that is a ≤4-word Title/ALL-CAPS phrase ending in an anchor noun — e.g. `Research Experience`, `Teaching Experience` — would match `matchSectionHeader` at `filtered[0]` and get stripped, dropping the title.

Narrow (few real titles end in experience/education/employment), but it's the exact class round-1 was about. Is there a corpus/unit fixture for a title-*leading* stacked block whose title matches the fallback? If not, worth one negative test to pin the boundary.

Samhit21 (commented):
## PR Review: download-PDF guardrail, Poppins embed, experience header fix

### Summary
Clean, well-documented epic. I verified the three changes against the checked-out branch: the `anchorIdx` re-basing after the leading-run slice is safe (downstream `hasAbove`/`anchorHasReconstructedSignature` both guard `>0`/`>=0`, so a negative index is inert); zero-egress holds for the font embed (same-origin `?url` asset fetch, not a CDN); and the download-gate "Fix now" wiring is correct — gated name/email still render the `EditableField`, so `[aria-label="Edit Name"|"Edit Email"]` resolves even when the field is absent. No blocking issues. Two suggestions and two notes below.

### Findings
1. **[Suggestion] render-ats-pdf.ts:266** — embed Poppins with `{ subset: true }`; un-subsetted embed adds a few hundred KB of unused glyphs to every exported PDF.
2. **[Question] experience.ts:426** — the leading-run strip still removes a title that *leads* its block and reads like a section header (`Research Experience`, `Teaching Experience`); confirm a negative test exists for the title-leading stacked shape.
3. **[Note, already self-flagged]** `loadPoppinsBytes()` caches a *rejected* promise on a transient fetch failure, permanently forcing Helvetica for the rest of the session. You flagged it as a LOW nit — worth actually fixing (null the module-scope promise in the `.catch`) since it's a one-liner and silently degrades the brand font.
4. **[Note, minor]** Skipping `toWinAnsi` on the embedded path means exotic glyphs Poppins lacks (→, CJK, emoji) render as `.notdef` boxes rather than the transliteration/`?` the Helvetica path gives. Won't crash (custom fonts don't throw on missing glyphs), and it's a reasonable trade for Latin-Extended fidelity — just calling out the behavior difference.

### Verdict
**Action: COMMENT**

src/lib/pdf/render-ats-pdf.ts
s-annam: Fixed in `245bd09`. Both `embedFont` calls now pass `{ subset: true }` — fontkit is already registered so the downloaded PDF carries only the ~60–80 glyphs used instead of the full Poppins Regular + Bold. Confirmed orthogonal to the skip-`toWinAnsi()` path (encoding logic untouched); render round-trip tests stay green.

src/lib/heuristics/extract/experience.ts
s-annam: Good catch — verified the title-leading case and it does get stripped: `Research Experience` (leads) over `Northern Trust · Jan 2020 - Dec 2021` yields title=`Northern Trust`, company=``. That's the **intended** precedence, not a regression: a leading ≤4-word anchor-noun phrase (`Research Experience`, `Teaching Experience`) is a section header far more often than a real job title — real titles read `Research Assistant`, not `Research Experience` — so the strip drops it as a boundary and the company-less entry below becomes the role (its name promoted to title via the `leadsFreshEntry` flip, same as #310). Retaining a leading anchor-noun phrase as a title would re-break #310. Pinned this boundary with a negative test (the mirror of the existing #310 mid-block false-positive) in `245bd09` — `extract-fields.test.ts`, 'strips a LEADING anchor-noun phrase as a section boundary, not a role title'.

Samhit21 (approved):
## Approving ✅

Both review items resolved and verified locally on `245bd09`:
- Font subsetting — `{ subset: true }` on both `embedFont` calls; typecheck clean, font tests green.
- Leading anchor-noun boundary — resolved as intended + pinned with a negative test; agree with the call (retaining it would re-break #310).

`tsc` clean, targeted suites 97/97, CI green (verify / fallow / Pages).

Optional non-blocking follow-up: null the `loadPoppinsBytes` promise on fetch failure so a transient error doesn't force Helvetica for the session.

PR #338, opened 2026-07-06
fix(test): posix-normalize corpus-roundtrip fixture keys on Windows

rohithgollapalli:
## Summary

On any **Windows checkout**, `corpus-roundtrip.test.ts` fails 4 tests: `relative()` yields `\`-separated paths while `KNOWN_FAILURES` is keyed with `/`, so

1. the stale-key check (`every KNOWN_FAILURES key names a real fixture`) fails, and
2. every known-failure exemption is silently voided — the three documented fixtures (`google-docs-skia-proxy-classic`, `openresume-react-pdf`, `weasyprint-cairo-classic`) report as `experience` regressions.

Fix: one `relKey()` helper that posix-normalizes the relative path (`split(sep).join("/")`), used by both the stale-key check and the per-fixture gate. No behavior change on macOS/Linux (where `sep` is already `/`).

## Test plan

- [x] `corpus-roundtrip.test.ts` on Windows: 4 failed → **39 passed / 1 skipped**
- [x] Full suite on Windows: **1514 passed, 0 failed**
- [x] `tsc -b --noEmit` + `eslint .` clean; `vite build` OK; fallow: no issues in the changed file

## Note: a second Windows tooling gap (not fixed here)

The `verify` npm script itself cannot run on Windows — its `(fallow audit … || echo '…')` subshell is POSIX syntax, and npm's default script shell on Windows is cmd.exe, which dies with `') was unexpected at this time.'` before anything runs. That also breaks the pre-push hook. This PR's gate was run step-by-step manually and the push used the documented `RESUMELINT_SKIP_HOOKS=1` bypass. Happy to follow up making `verify` cross-platform (e.g. move the fallow fallback into a small node script) if wanted — flagging rather than scope-creeping it here.

### DIFF
--- a/src/lib/heuristics/corpus-roundtrip.test.ts
+++ b/src/lib/heuristics/corpus-roundtrip.test.ts
@@ -47,7 +47,7 @@
  */
 
 import { readFileSync, readdirSync, mkdirSync, writeFileSync } from "node:fs";
-import { dirname, join, relative, basename } from "node:path";
+import { dirname, join, relative, basename, sep } from "node:path";
 import { fileURLToPath } from "node:url";
 
 import { describe, it, expect } from "vitest";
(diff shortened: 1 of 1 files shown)


rohithgollapalli:
Consolidating into #339 per review preference — the identical commit (cherry-picked as 8cb1a67) now rides there, and the combined branch is fully green on Windows (1521/1521). Closing this one.

Issue #319, opened 2026-07-03
job-search: provider adapter interface + keyless CORS-open feeds + fit ranking


s-annam:
Part of the job-search epic (see parent issue). Second slice: fetch real postings from keyless CORS-open job feeds and rank them against the resume using the existing jd-match machinery. Blocked by the query-builder issue (consumes `JobQuery`).

## Problem

Deep links (prior slice) hand the user off to external boards. This slice brings ranked results in-app: fetch postings from free, keyless, CORS-open job APIs and score each against the parsed resume.

### Provider adapter interface — `src/lib/job-search/types.ts`

```ts
export interface JobPosting {
  id: string;            // provider-prefixed for dedup, e.g. "remotive:12345"
  title: string;
  company: string;
  location: string;      // often "Remote"
  url: string;           // canonical external listing URL
  description: string;   // plaintext (strip HTML via existing htmlToPlaintext from jd-match)
  postedAt?: string;     // ISO date if provider supplies it
  source: string;        // provider display name
}

export interface JobProvider {
  id: string;
  label: string;
  search(query: JobQuery, signal: AbortSignal): Promise<JobPosting[]>;
}
```

### Providers — `src/lib/job-search/providers/`

Candidate keyless CORS-open feeds (verify CORS from a browser origin **first**; drop any that fail):

- **Remotive** — `https://remotive.com/api/remote-jobs?search=<kw>`
- **Arbeitnow** — `https://www.arbeitnow.com/api/job-board-api?search=<kw>`
- **Jobicy** — `https://jobicy.com/api/v2/remote-jobs?tag=<kw>`

One adapter file per provider, each mapping its response shape to `JobPosting`. Mock-fetch unit tests per adapter (no live network in vitest).

### Orchestration + ranking — `src/lib/job-search/search.ts`, `rank.ts`

- Fan out `Promise.allSettled` across enabled providers; a failing provider degrades gracefully (its results are absent, UI notes it), never rejects the whole search.
- Dedup by normalized `title+company`.
- Rank each posting: `extractJdTerms(posting.description)` + `computeCoverage(...)` from `src/lib/jd-match/` → fit score. Sort descending. Providers/rank tiers are dynamic-imported (same pattern as cascade tiers) so the entry chunk stays small.

### UI

- Extend the "Find Jobs" panel: explicit **Search** button (never auto-fetch on drop) → loading state → ranked job cards.
- Card: title, company, source, fit %, top matched + missing skills, external link.
[...]

Issue #201, opened 2026-06-26
JD-match: LLM evidence judge (requirements + resume → met/partial/missing, batched)


s-annam:
Part of the WebLLM-native JD-matching redesign (anchor: #156, milestone **M6 · JD Matching v2**).

## Goal
LLM call #2: judge resume evidence per requirement.

## Scope
New `src/lib/jd-match/llm/judge-evidence.ts`:
- `judgeEvidence(requirements, parsed, engine): Promise<RequirementVerdict[]>`
  ```ts
  RequirementVerdict {
    requirement: JdRequirement;
    status: "met" | "partial" | "missing";
    reason: string;     // one sentence, grounded in resume text
    evidence?: string;  // cited resume snippet
  }
  ```
- **Batch** requirements (~6–8 per call) the way `rewrite-resume.ts` chains sections; each prompt carries a flattened/deduped resume projection (reuse `buildCorpus`-style field selection, **not** raw PDF text) in the system message as reference-only.
- Bracket every `chat.completions.create()` with `acquireInference` / `releaseInference` (`web-llm.ts`) — copy the `rewrite-section.ts` guard pattern.

## Constraints
- Reconcile verdicts against the extracted requirement ids (injected "all met" can't invent requirements).
- Canned-stub unit tests; batch-boundary test.

## Done when
Given requirements + a parsed resume + stub engine, returns batched verdicts with status+reason; inference guard bracketed; tests green.

Issue #200, opened 2026-06-26
JD-match: LLM requirement extractor (JD text → JdRequirement[]) + tolerant JSON parser


s-annam:
Part of the WebLLM-native JD-matching redesign (anchor: #156, milestone **M6 · JD Matching v2**).

## Goal
LLM call #1: JD text → structured requirements (`JdRequirement[]`).

This is the **production build** of what issue #198 prototyped in `src/lib/webllm/spike/`. Promote the spike's already-tuned artifacts into the prod path — do NOT rewrite them from scratch. The spike prompt is tuned against real model output; preserve that tuning exactly.

### `src/lib/jd-match/llm/extract-requirements.ts` — new prod file

```ts
export function extractRequirements(
  jdText: string,
  engine: WebLlmEngine,
): Promise<JdRequirement[]>
```

`WebLlmEngine` is imported from `src/lib/webllm/types.ts` (the existing narrow interface over `@mlc-ai/web-llm`).

### `src/lib/jd-match/llm/prompts.ts` — PROMOTED from spike

Promote the **extract-half** of `src/lib/webllm/spike/prompts.ts` into this file:
- `EXTRACT_SYSTEM_PROMPT` (exported const)
- `buildExtractUserPrompt(jdText: string): string` (exported function)

Keep both exported so the eval harness can import and compare variants.

**Delete the spike copies** once the prod file is landed. The judge-half (`JUDGE_SYSTEM_PROMPT`, `buildJudgeUserPrompt`, `JUDGE_BATCH_SIZE`) stays in the spike until PR #201.

### `JdRequirement` type — PROMOTED from spike

Promote `JdRequirement` from `src/lib/webllm/spike/types.ts` into `src/lib/jd-match/llm/` (or re-export from `src/lib/jd-match/types.ts` — implementer's call, but pick one canonical location).

Correct field shape (field is `years`, **not** `yearsRequired`):

```ts
export interface JdRequirement {
  /** Stable kebab-case id, e.g. "req-1". Sequential, 1-based. */
  id: string;
  /** Semantic category. */
  kind: "skill" | "experience" | "responsibility" | "qualification";
  /** Concise requirement text (one sentence). */
  text: string;
  /**
   * Minimum years stated in the requirement, as an integer.
   * Omit when the JD does not give a numeric year count.
   */
  years?: number;
}
```

### JSON parsing — EXTEND `src/lib/webllm/json-repair.ts` (do NOT create `parse-json.ts`)

[...]

Issue #156, opened 2026-06-24
Epic — WebLLM-native JD matching (semantic requirement/evidence matching, keyword fallback)


sriyau64:
**Epic** · Milestone **M6 · JD Matching v2** (post-v1.0)

Rebuild JD matching as a **WebLLM-native semantic capability**, with the current deterministic keyword-coverage retained as the no-WebGPU fallback. This issue is the tracking anchor; the work is the linked sub-issues (#198–#207).

## Problem

Today JD matching is a **keyword-coverage** model (`src/lib/jd-match/`): extract JD terms (a dictionary pass over `skills.ts` + a capitalized noun-phrase pass) and string-match them against a corpus of typed resume fields. The original report (a music-intern JD, preserved below) exposed that this is structurally wrong, not a single bug:

1. **`skills.ts` is engineering-only (~120 entries).** For any non-tech role the dictionary pass finds ~nothing, so noun-pass noise *is* the output. A dictionary per domain is unbounded maintenance.
2. **Literal string match has no semantics.** "Led ensemble rehearsals" never matches "conducting experience" — we miss the real signal.
3. **Section-heading filtering is whack-a-mole.** Every new JD shape adds stop-phrases (e.g. the `isSectionHeading()` guard already added for "Minimum Qualifications" / "Physical Demands"); never complete.
4. **Corpus is typed-fields-only.** Creative sections the parser buckets under `other` are invisible regardless of match quality.

## Decision

- **Semantic matching on the planned in-browser LLM layer:** the model extracts JD **requirements** (skill / experience / responsibility / qualification, + years) and judges resume **evidence** per requirement → **met / partial / missing, with a reason**. Domain-agnostic, no per-domain dictionaries, honors the no-bytes-leave-the-browser constraint, fits the local-LLM + HITL direction.
- **Keep the deterministic keyword path as the no-WebGPU fallback** — never deleted; it is the v1 "always render something" return hook.

## Architecture

- **Two-call contract** (bounded for the small model's context): (1) JD text → `JdRequirement[]`; (2) `(requirements, flattened resume projection)` → `RequirementVerdict[]`, batched ~6–8/call.
- **One path-agnostic API** behind `JdMatch.tsx`: a discriminated union `{ path: "semantic" | "keyword" }`. `detectWebGpu()` + an explicit opt-in toggle select the path; UI is path-agnostic.
[...]

PR #339, opened 2026-07-06
feat(jd-match): semantic orchestrator + keyword fallback; fix Windows corpus-test paths (#202)

### LINKED ISSUE #202: JD-match: orchestrator + WebGPU path selection (semantic, keyword fallback)
Part of the WebLLM-native JD-matching redesign (anchor: #156, milestone **M6 · JD Matching v2**).

## Goal
Orchestrate the semantic path and the path selection behind the stable API.

## Scope
New `src/lib/jd-match/llm/run-llm-match.ts`:
- `runLlmMatch(jdText, parsed, engine, onProgress): Promise<JdMatchResult>` → `detectWebGpu` gate already passed by caller → `loadEngine` → `extractRequirements` → `judgeEvidence` → assemble `{ path: "semantic", verdicts, summary }`.
- **Fallback discipline:** any failure (no WebGPU, load error, parse hard-failure, inference error) resolves to the keyword path (`{ path: "keyword" }`) — never a blank panel.
- Reuse `useModelSelection` selected model id; reuse `ConsentDialog` gate if the user is on a restricted model.

## Done when
Orchestrator returns a `semantic` result on the happy path and degrades to `keyword` on every failure mode; covered by stub tests for both branches.

rohithgollapalli:
## Summary

Two changes on one branch (per review consolidation): the semantic JD-match **orchestrator** (#202) and the **Windows corpus-test fix** it surfaced (originally opened separately as #338, now folded in here).

### 1. Semantic orchestrator + path selection (#202)

Adds the semantic JD-match orchestrator behind the stable `JdMatchResult` API (M6 anchor #156; follows #200 extractor and #201 judge, whose first non-test consumer this is).

- **New `src/lib/jd-match/llm/run-llm-match.ts`** — `runLlmMatch(jdText, parsed, modelId, onProgress)`: `loadEngine` → `extractRequirements` (#200) → `judgeEvidence` (#201) → `{ path: "semantic", verdicts, summary }`. `modelId` is the `useModelSelection` selected id, threaded to both the loader and the judge's inference guard; the `detectWebGpu` and ConsentDialog gates are documented as the caller's job (this module never prompts). *Signature note:* the issue sketch wrote `engine` as the third param, but `loadEngine` is a step **inside** the orchestrator (that's what `onProgress` reports), and the judge needs the model id for its inference guard — so the param is `modelId`.
- **Fallback discipline** — never rejects, never a blank panel: engine-load failure (also how a missing-WebGPU environment manifests), `RequirementExtractionError`, an unexpected judge error, **and a valid-but-empty extraction** all resolve to `{ path: "keyword" }`, built with the same `extractJdTerms` + `computeCoverage` composition the JD-fit surface runs today. (`judgeEvidence` never throws by contract — its per-batch failures stay on the semantic path as `missing` verdicts, by design.)
- **`types.ts`** — fleshes out the provisional semantic arm exactly as its comment promised: real judge `RequirementVerdict` (type-only import → no runtime edge into the WebLLM chunk) and a new exported `SemanticMatchSummary` (`met/partial/missing/total` — per-status counts so the renderer never re-tallies; the provisional `matched/total` couldn't represent `partial`).
- **Chunk discipline** — deliberately NOT exported from the `jd-match` barrel (statically imported by the JD-fit entry); consumers dynamic-import this module (cascade-tier pattern) so WebLLM stays out of the entry chunk.

### 2. Windows corpus-roundtrip fix (was #338)

On any Windows checkout, `corpus-roundtrip.test.ts` failed 4 tests: `relative()` yields `\`-separated paths while `KNOWN_FAILURES` is keyed with `/`, so the stale-key check failed and every known-failure exemption was silently voided (the three documented fixtures reported as `experience` regressions). Fixed with one `relKey()` helper that posix-normalizes the relative path (`split(sep).join("/")`), used by both the stale-key check and the per-fixture gate. No behavior change on macOS/Linux.

Resolves #202

## Test plan

- [x] 7 new stub tests (`run-llm-match.test.ts`): happy-path assembly + summary tally, collaborator arg threading, all four failure modes → keyword (with judge-not-called assertions), and a **fallback-parity** test that independently recomputes the keyword result (#319-style)
- [x] One-line fixture update in `JdMatch.test.ts` (old provisional `{matched,total}` summary shape); renderer still renders nothing for the semantic path until its UI slice
- [x] `corpus-roundtrip.test.ts` on Windows: 4 failed → 39 passed / 1 skipped
- [x] **Full suite on Windows: 1521 passed / 0 failed** (fully green with both changes combined)
- [x] `tsc -b --noEmit` + `eslint .` clean; `vite build` OK; fallow report-only clean on changed files

## Note: a second Windows tooling gap (not fixed here)

The `verify` npm script itself cannot run on Windows — its `(fallow audit … || echo '…')` subshell is POSIX syntax, and npm's default script shell on Windows is cmd.exe, which dies with `') was unexpected at this time.'` before anything runs. That also breaks the pre-push hook. This PR's gate was run step-by-step manually and the push used the documented `RESUMELINT_SKIP_HOOKS=1` bypass. Happy to follow up making `verify` cross-platform (e.g. move the fallow fallback into a small node script) — flagging rather than scope-creeping it here.

### DIFF
--- a/src/components/features/JdMatch.test.ts
+++ b/src/components/features/JdMatch.test.ts
@@ -127,7 +127,7 @@ describe("JdMatch", () => {
     const result: JdMatchResult = {
       path: "semantic",
       verdicts: [],
-      summary: { matched: 0, total: 0 },
+      summary: { met: 0, partial: 0, missing: 0, total: 0 },
     };
     const html = renderToStaticMarkup(createElement(JdMatch, { result }));
     expect(html).toBe("");
--- a/src/lib/heuristics/corpus-roundtrip.test.ts
+++ b/src/lib/heuristics/corpus-roundtrip.test.ts
@@ -47,7 +47,7 @@
  */
 
 import { readFileSync, readdirSync, mkdirSync, writeFileSync } from "node:fs";
-import { dirname, join, relative, basename } from "node:path";
+import { dirname, join, relative, basename, sep } from "node:path";
 import { fileURLToPath } from "node:url";
 
 import { describe, it, expect } from "vitest";
@@ -59,6 +59,14 @@ import { renderAtsResumePdf } from "../pdf/render-ats-pdf.ts";
 
 const HERE = dirname(fileURLToPath(import.meta.url));
 const FIXTURE_ROOT = join(HERE, "../../..", "tests/fixtures/pdfs");
+
+/** Fixture path relative to `FIXTURE_ROOT`, posix-separated on every platform.
+ *  `KNOWN_FAILURES` is keyed with `/`; Windows `relative()` yields `\`, which
+ *  both failed the stale-key check and silently voided every known-failure
+ *  exemption there (the three documented fixtures reported as regressions). */
+function relKey(fixture: string): string {
+  return relative(FIXTURE_ROOT, fixture).split(sep).join("/");
+}
 
 type Category =
   | "contact"
@@ -261,13 +269,13 @@ describe("corpus round-trip invariants (#293)", () => {
   });
 
   it("every KNOWN_FAILURES key names a real fixture", () => {
-    const rel = new Set(fixtures.map((f) => relative(FIXTURE_ROOT, f)));
+    const rel = new Set(fixtures.map(relKey));
     for (const key of Object.keys(KNOWN_FAILURES))
       expect(rel.has(key), `stale KNOWN_FAILURES key: ${key}`).toBe(true);
   });
 
   for (const fixture of fixtures) {
-    const rel = relative(FIXTURE_ROOT, fixture);
+    const rel = relKey(fixture);
     it(`round-trips: ${rel}`, async () => {
       const p1 = await runCascade(new Uint8Array(readFileSync(fixture)));
       const model = buildAtsResumeModel(p1, scoreFor(p1));
--- a/src/lib/jd-match/llm/run-llm-match.test.ts
+++ b/src/lib/jd-match/llm/run-llm-match.test.ts
@@ -1,0 +1,216 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+/**
+ * Unit tests for runLlmMatch (#202) — the semantic orchestrator's two
+ * branches, driven by stubs (no WebGPU): the happy path assembles a
+ * `semantic` result from the collaborators' outputs, and EVERY failure mode
+ * (engine load error, extraction hard-failure, empty extraction, unexpected
+ * judge error) resolves to a keyword result that matches a fresh
+ * `extractJdTerms` + `computeCoverage` computation exactly — fallback parity,
+ * proven the same way rank parity is (#319-style independent recompute).
+ *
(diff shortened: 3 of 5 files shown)


s-annam (approved):
## PR Review: semantic orchestrator + keyword fallback; Windows corpus-test fix (#202)

### Summary
Clean, well-tested, well-documented. Both changes land green. Verified the orchestrator's collaborator wiring against the real modules, confirmed chunk discipline, and reran the full gate locally. **APPROVE.**

### Spec alignment (#202)
|---|---|---|
| `runLlmMatch` behind stable `JdMatchResult` API | ✅ | `loadEngine → extractRequirements → judgeEvidence → { semantic }` |
| Never rejects / never blank panel | ✅ | engine-load, `RequirementExtractionError`, empty extraction, unexpected judge error → keyword |
| Empty-but-valid extraction degrades | ✅ | guarded before judge, no warn (correct — not a failure) |
| `modelId` threaded to loader + judge guard | ✅ | signature deviation from issue sketch is justified in the body |
| `types.ts` semantic arm fleshed out, type-only import | ✅ | no runtime edge into the WebLLM chunk |
| Chunk discipline (not in barrel) | ✅ | barrel exports only the `JdMatchResult` **type** |

### Verification I ran
- Collaborator signatures match exactly: `loadEngine(modelId, onProgress)`, `extractRequirements(jdText, engine)`, `judgeEvidence(requirements, parsed, engine, modelId)`.
- Barrel (`index.ts`) does **not** re-export `run-llm-match`; `types.ts` `RequirementVerdict` import is `type`-only → WebLLM stays out of the JD-fit entry chunk. ✓
- Full suite **1521 passed / 2 skipped**, `tsc -b --noEmit` clean, `eslint .` clean.
- Fallow (report-only gate): **dead code 0 · complexity 0**; `runLlmMatch` is not flagged as an orphan export (its test consumes it). Duplication is warn-only.

### Highlights
- Fallback-parity test independently recomputes `extractJdTerms + computeCoverage` (#319-style) — proves the keyword arm is byte-identical to what the JD-fit surface renders today, not just "a keyword result."
- `summarize()`'s `else`-buckets-missing keeps the `met + partial + missing === total` invariant the headline renders from — deliberately robust against an unknown status inflating `total`.
- Semantic panel can never render blank when reached: judge returns one verdict per requirement, and the empty-extraction case degrades to keyword.
- Windows fix (`relKey` posix-normalize) is a genuine correctness bug — the stale `\`-vs-`/` key mismatch silently voided every `KNOWN_FAILURES` exemption; no-op on macOS/Linux.

### Non-blocking observations (no change requested)
- `run-llm-match.test.ts` — the two failure-mode tests share an 8-line block (fallow `dup:3c0c0896`). Warn-only; keeping the four failure tests un-parameterized reads clearer, so leaving as-is is fine.
- The Windows `verify`/pre-push cross-platform gap you flagged is the right call to defer — good separation of concerns.

### Verdict
**Action:** APPROVE
**Rationale:** Zero blocking findings. Signatures, chunk discipline, and fallback semantics all verified against the real collaborators; full suite green. Nothing needs a follow-up commit.

Issue #349, opened 2026-07-06
Name heuristic promotes an education institution as full_name when the name is absent from the text layer (Deedy)


rohithgollapalli:
## Summary

**Category: latex, two-column (Deedy).** On `tests/fixtures/pdfs/latex/deedy-resume-openfonts.pdf`, the persona's name never reaches the text layer (the template renders it in display fonts; `rawText` starts at the contact line). With **no name present**, the name heuristic **promotes an education content line as `full_name`**:

```
full_name = "CORNELL UNIVERSITY"
```

A **false-positive name is worse than an empty one**: it renders a wrong name in the reconstructed resume, and it earns undeserved completeness credit in the score (an absent name should surface as a `completeness.missing` entry, not silently pass).

Found via the `/probe-contact` sweep. The profile band contains only the contact line (`profileLines = 1`), so the scorer's candidate pool must have reached past the profile into the `EDUCATION` region and accepted an all-caps institution line. Related: #330 (closed) was a name *present but exploded*; this is the inverse — name absent, heuristic hallucinates one.

## Fix direction

In `src/lib/heuristics/extract/name.ts`: candidate lines that (a) sit inside a recognized non-profile section, or (b) exactly match a known-institution/organization pattern already recognized by the education extractor, should be rejected. If no candidate survives, return empty and let completeness report the missing name.

## Acceptance criteria

- [ ] `deedy-resume-openfonts.pdf` parses with empty `full_name` (and `completeness.missing` includes the name) — OR the actual name when the text layer carries it
- [ ] No candidate line from inside the `education` region is ever promoted to `full_name`
- [ ] Snapshot re-baked; sibling `deedy-resume-macfonts.pdf` checked for the same symptom

Issue #346, opened 2026-07-06
Two-column (Chromium): title/company inverted + role location dropped when company line carries a comma descriptor


rohithgollapalli:
## Summary

**Category: unknown / Chromium-print, two-column.** On the committed fixture `tests/fixtures/pdfs/unknown/chromium-two-column-sidebar.pdf`, a role whose **company line carries a comma descriptor** parses with **title and company inverted**, and its **role location is dropped**. First-parse defect — distinct from closed #298, which was the *round-trip* re-segmentation swap.

Found via the `/probe-experience` sweep (probe added in `src/lib/heuristics/probe-experience.test.ts`).

## Evidence (synthetic fixture — safe to quote)

Experience region scanned (INPUT), role 2:

```
Sr. Engineering Manager
Globex Assistant, Assistant on Living-Room TV
06/2017 - 04/2021 Austin, TX
```

Parsed entry (OUTPUT):

```
title   = "Globex Assistant"
company = "Sr. Engineering Manager"
location = (empty)
```

Expected: `title = "Sr. Engineering Manager"`, `company = "Globex Assistant, Assistant on Living-Room TV"` (or company + team split), `location = "Austin, TX"`.

Role 1 in the same fixture has the identical title-first shape but a **comma-less** company line (`Northwind Systems - Customer Experience Engineering`) and maps correctly, including location. So the inversion is triggered by the comma descriptor in the company line, and the location drop co-occurs only on the inverted role — `disambiguateCompanyTitle` (`src/lib/heuristics/extract/experience.ts`) appears to prefer the comma-segmented line as the title and then mis-consumes the header stack.

Note: this wrong parse is **baked into the corpus snapshot** (`chromium-two-column-sidebar.expected.json` passes CI with the inversion), so the corpus gate cannot catch it — it normalizes the bug.

## Acceptance criteria

- [ ] Role 2 of `chromium-two-column-sidebar.pdf` parses with title `Sr. Engineering Manager` and the Globex line as company (± team split)
- [ ] Its `location` = `Austin, TX` like the sibling roles
- [ ] Snapshot re-baked; roles 1/3/4/5 unchanged
- [ ] Unit case in `extract-fields.test.ts`: title line + comma-descriptor company line + `MM/YYYY - MM/YYYY City, ST` line → no inversion

Issue #342, opened 2026-07-06
Single-column layout: employment role (title) not detected properly on experience entries


rohithgollapalli:
## Summary

On a **single-column résumé**, the parser does not detect the **employment role (job title)** properly for experience entries. The entry itself is found, but the role/title field comes out wrong (missing, or mis-assigned from another header line).

Observed during manual testing on a local single-column résumé (not committed — PII policy). **Needs a synthetic repro fixture** per `tests/fixtures/pdfs/README.md`, plus a `/probe-contact`-style localization of which layer drops it (line assembly vs entry-block header mapping vs `disambiguateCompanyTitle`).

## Symptom

- Input: plain single-column layout, standard experience section with role headers.
- Expected: each experience entry carries the correct `title` (employment role).
- Actual: role/title not detected properly — empty or wrong value on the parsed entry.

## Related (all closed — adjacent failure modes)

- #217 — ·-delimited single-line headers not split into title vs company
- #166 — multi-line (3-line) wrapped role header not recognized
- #239 — dateless role header folded into previous entry
- #310 (open) — second experience-category section header absorbed as a role's company (different defect: that one is about section headers, this one is about the role header itself on ordinary single-column input)

Likely area: header-line → `title`/`company` mapping in `parseEntryBlocks` / `disambiguateCompanyTitle` (`src/lib/heuristics/extract/experience.ts`).

## Acceptance criteria

- [ ] Synthetic single-column fixture (fake persona, `555-01xx` phone) reproducing the miss added under `tests/fixtures/pdfs/`
- [ ] Each experience entry in the fixture parses with the correct `title`
- [ ] Unit coverage in `extract-fields.test.ts` for the failing header shape once localized
- [ ] No regression on existing corpus fixtures

rohithgollapalli:
Committed synthetic repro found via the new `/probe-experience` sweep: `tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-role-first-experience.pdf` (safe to quote — synthetic persona).

Role 2 parses with the role title mis-assigned:

```
title   = "Multicultural Engineering Program – State Polytechnic University"
company = "Peer Tutor (15 to 20 hours per week)"
```

Expected: `title = "Peer Tutor (…)"` with the program/university line as the org. Role 1 in the same fixture (same role-first layout, but followed by an `Org – City, ST` line) maps correctly, so the swap is in the header-stack mapping (`disambiguateCompanyTitle`), not section routing.

[...]

Issue #341, opened 2026-07-06
Two-column layout: second employment entry not recognized - demoted to bullets under the first role


rohithgollapalli:
## Summary

On a **two-column résumé** with two employment-history entries, the parser recognizes only **one of the two roles**. The second entry's content is not segmented as its own role — its lines are instead represented as **bullet points** (attached under the recognized role / the "Other bullets" pool), so `experienceCount` under-reports and the reconstructed resume shows one role with an inflated bullet list.

Observed during manual testing on a local two-column résumé (not committed — PII policy). **Needs a synthetic repro fixture** per `tests/fixtures/pdfs/README.md` before the fix lands.

## Symptom

- Input: two-column layout, experience section contains 2 employment entries.
- Expected: 2 parsed roles, each with its own title/company/dates and bullets.
- Actual: 1 parsed role; the second entry's header line(s) are not recognized as a role anchor, and its content is demoted to bullet lines under the first role (or the "Other" pool).

## Related (all closed — different symptoms)

- #60 — sidebar interleaves into experience (text-assembly ordering, not entry segmentation)
- #131 — two-column flatten drops achievements/projects when columns share a bullet margin
- #298 — round-trip title/company re-segmentation swap in dense/two-column layouts
- #239 — dateless role header folded into previous entry (single-column variant of the same "entry not anchored" failure class)
- #235 — epic: anchor entry recognition on header shape, not enrichment fields

This looks like the two-column flattening path breaking the shape cues (`parseEntryBlocks` anchors in `src/lib/heuristics/extract/experience.ts`) that entry segmentation relies on — e.g. the flattened line order or merged column text keeps the second entry's header from matching a role anchor.

## Acceptance criteria

- [ ] Synthetic two-column fixture (fake persona, `555-01xx` phone) with 2 employment entries added under `tests/fixtures/pdfs/`
- [ ] Both roles parse as distinct experience entries (`experienceCount: 2`), each with its own bullets
- [ ] No entry content leaks into the other role's bullet list or the "Other" pool
- [ ] Existing two-column corpus fixtures stay green

rohithgollapalli:
Reproducer landed in #352 — synthetic two-column fixture (`google-docs-skia-proxy-two-column-long-role-header.pdf`) that reproduces the demote-to-Other-bullets symptom through the real parser (`experienceCount: 1`, probe verdict UNDER-SEGMENTED), confirmed matching the real résumé it was modeled on.

[...]

PR #351, opened 2026-07-06
feat(probe): add probe-experience skill + RL_EXPERIENCE_PDF vitest harness

rohithgollapalli:
## Summary

Adds `probe-experience`, the third `probe-*` parser probe (siblings: `probe-contact`, `probe-roundtrip`): a `RL_EXPERIENCE_PDF`-gated vitest harness (`src/lib/heuristics/probe-experience.test.ts`) plus its skill doc (`.claude/skills/probe-experience/SKILL.md`). It runs any résumé PDF — including a real, PII-bearing one — through the real cascade and prints the score, section overview, parsed role entries (title/company/location/dates/bullets), the experience region the segmenter scanned (INPUT vs OUTPUT), and an `ok / UNDER-SEGMENTED / PARSER-MISS` verdict from a date-range lower-bound oracle, so a dropped/merged role is localizable to line-assembly vs section-routing vs entry segmentation.

Same guardrails as the sibling probes: inert in CI (env-gated, never reddens the suite), JSON report written to the gitignored `internal/experience/`, PII rules documented in the SKILL.md.

Proven in use: this harness ran the 10-fixture parse-accuracy sweep that filed #346–#349 and produced the committed-fixture repro comment on #342.

Refs #341, #342 (triage tooling for those lanes; does not resolve them).

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green (probe skips when `RL_EXPERIENCE_PDF` unset — verified inert)
- [x] `npm run build` clean
- [x] Harness exercised end-to-end against 10 corpus fixtures across 4 provenance categories (reports in gitignored `internal/experience/`)

### DIFF
--- a/src/lib/heuristics/probe-experience.test.ts
+++ b/src/lib/heuristics/probe-experience.test.ts
@@ -64,17 +64,7 @@ describe.runIf(process.env.RL_EXPERIENCE_PDF)(
       const p = cascade.parsed;
 
       const score = computeAnonymousAtsScore({
-        parsed: {
-          full_name: p.full_name,
-          email: p.email,
-          phone: p.phone,
-          location: p.location,
-          linkedin_url: p.linkedin_url,
-          summary: p.summary,
-          skills: p.skills,
-          experience: p.experience,
-          education: p.education,
-        },
+        parsed: { ...p },
         fieldConfidence: cascade.fieldConfidence,
         triggers: cascade.triggers,
         rawText: cascade.rawText,
--- a/src/lib/heuristics/probe-experience.test.ts
+++ b/src/lib/heuristics/probe-experience.test.ts
@@ -1,0 +1,169 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+/**
+ * Experience-section dev probe — inert in CI, runs ONLY when
+ * `RL_EXPERIENCE_PDF=<path>` is set:
+ *
+ *   RL_EXPERIENCE_PDF=/abs/path/to/resume.pdf npx vitest run \
+ *     src/lib/heuristics/probe-experience.test.ts
+ *
+ * Extracts + VERIFIES the experience section of one arbitrary PDF so a real
+ * (uncommitted, possibly PII-bearing) résumé can be triaged WITHOUT being
+ * committed as a fixture. This is the execution vehicle for the
+ * `probe-experience` skill — sibling of `probe-contact` / `probe-roundtrip`;
+ * the pattern is: run the real parser via runCascade (the pdfjs worker only
+ * resolves under the vitest transform — no standalone script), then show the
+ * section extractor's INPUT (the experience region it scanned) next to its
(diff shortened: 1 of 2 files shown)


src/lib/heuristics/probe-experience.test.ts
```diff
@@ -0,0 +1,169 @@
...
+    it("extracts + verifies the experience section for RL_EXPERIENCE_PDF", async () => {
+      const path = process.env.RL_EXPERIENCE_PDF!;
+      const outDir =
+        process.env.RL_EXPERIENCE_OUT ??
+        join(HERE, "../../..", "internal/experience");
+
+      const cascade = await runCascade(new Uint8Array(readFileSync(path)));
+      const p = cascade.parsed;
+
+      const score = computeAnonymousAtsScore({
```
s-annam, line 66: [Nit]: Consider `parsed: { ...p }` (or `{ ...cascade.parsed }`) instead of hand-listing the 9 fields — matches the pattern in `corpus-roundtrip.test.ts` `scoreFor()` and won't silently drop a field if the parsed shape grows. Non-blocking; tsc confirms the current list is complete.

s-annam (dismissed):
## PR Review: feat(probe): add probe-experience skill + RL_EXPERIENCE_PDF vitest harness

### Summary
Third `probe-*` dev harness, faithfully modeled on the two merged siblings (`probe-contact`, `probe-roundtrip`). Env-gated, inert in CI, PII guardrails documented. Verified locally: `typecheck`/`lint` clean, test skips when `RL_EXPERIENCE_PDF` unset (1 skipped — never reddens the suite), `internal/experience/` gitignored, all imported symbols exist and are used correctly.

### Highlights
- INPUT-vs-OUTPUT split + date-range lower-bound oracle is a genuinely useful triage design; verdict strings (`ok` / `UNDER-SEGMENTED` / `PARSER-MISS`) match the SKILL doc exactly.
- Correctly notes `DATE_RANGE_RE` is non-global, so the per-line `.test()` in a filter is stateless-safe — confirmed (`"i"` flag only).
- Consistent guardrails with siblings: env-gated, gitignored output, cite-corruption-by-category PII rule.

### Key Findings
1. **[Nit]** The `computeAnonymousAtsScore({ parsed: {…} })` call hand-lists 9 fields; `corpus-roundtrip.test.ts` uses `parsed: { ...cascade.parsed }`. Spread is simpler and won't silently drop a field if the parsed type grows. Optional.

### Verdict
**APPROVE** — 0 blocking, 1 nit. Well-constructed, gates verified green locally, matches established probe conventions.

src/lib/heuristics/probe-experience.test.ts
rohithgollapalli: Fixed in 9aca776 — swapped the hand-listed 9 fields for `parsed: { ...p }`, matching `corpus-roundtrip.test.ts` `scoreFor()`. Verified behavior-identical with a live probe run (same score on the classic fixture); typecheck + full suite green.

s-annam (approved):

PR #352, opened 2026-07-07
fix(parser): two-column role header demoted to bullets — 'Company. City' prose false-positive (#341)

rohithgollapalli:
## Summary

Fixes **#341** (two-column: second employment entry demoted / role lost) and lands its synthetic reproducer fixture in the same PR. Confirmed reproducing in `npm run dev` (drop the fixture → one role parsed, the other's content in "Other bullets") and via `/probe-experience`, then fixed at the root.

## Root cause

`isProseLine`'s `SENTENCE_BREAK_RE` (`word. Capital`) false-matched a résumé header's **company/location period delimiter**: `Platform Engineering Intern, Northwind Technology. San Jose, California` — the `Technology. San` reads as a sentence boundary. On a two-column résumé that stacks `Title, Company. City, State` over a bare date line, an **8+-word** header was mis-classified as a description → its entry block was dropped as a date-only phantom → the role's bullets fell through to a neighbor / the "Other" pool. Only the shorter (≤7-word) role survived (that role clearing the 8-word prose floor by luck, not design).

## Fix

Tighten `SENTENCE_BREAK_RE` to require a **running clause** after the break — a later lowercase-initial word — which a real second sentence has (`flow. Reduced churn by…`) but an all-Title-Case `City, State` location tail does not (`Technology. San Jose, California`). One line:

```
-const SENTENCE_BREAK_RE = /[a-z]{2}\.\s+[A-Z]/;
+const SENTENCE_BREAK_RE = /[a-z]{2}\.\s+[A-Z][^.]*\s[a-z]/;
```

Real multi-sentence glyph-less descriptions still classify as prose (verified); the header is now claimed and both roles segment, each keeping its own bullets.

## The reproducer fixture

`tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-two-column-long-role-header.pdf` — Chromium `--print-to-pdf` Skia export (`google-docs-skia-proxy-*` convention). Persona synthetic + `pdftotext`-verified: **Jordan Avery / jordan.avery@example.com / (614) 555-0137**. Its snapshot moves `experienceCount` **1 → 2** as the red→green proof.

## Corpus impact (net positive)

| Fixture | experienceCount | Why |
|---|---|---|
| `…two-column-long-role-header` (repro) | 1 → **2** | the fix |
[...]

PR #361, opened 2026-07-07
fix(parser): two-column role header demoted to bullets — 'Company. City' prose false-positive (#341)

### LINKED ISSUE #341: Two-column layout: second employment entry not recognized - demoted to bullets under the first role

rohithgollapalli:
## Summary

Fixes **#341** (two-column: second employment entry demoted to bullets). The synthetic reproducer fixture already landed via #352 (`experienceCount: 1`, recording the bug); this PR is the **fix** and flips that snapshot to `2`.

> Context: #352 was merged while it was still the fixture-only reproducer, so this fix (originally staged on that branch) didn't ride along. Re-based onto current `main` (post-#353) here.

## Root cause

`isProseLine`'s `SENTENCE_BREAK_RE` (`word. Capital`) false-matched a résumé header's **company/location period delimiter**: `Platform Engineering Intern, Northwind Technology. San Jose, California` — the `Technology. San` reads as a sentence boundary. On a two-column résumé that stacks `Title, Company. City, State` over a bare date line, an **8+-word** header was mis-classified as a description → its entry block dropped as a date-only phantom → the role's bullets demoted to a neighbor / the "Other" pool. Only the shorter (≤7-word) role survived.

## Fix (one line)

```
-const SENTENCE_BREAK_RE = /[a-z]{2}\.\s+[A-Z]/;
+const SENTENCE_BREAK_RE = /[a-z]{2}\.\s+[A-Z][^.]*\s[a-z]/;
```

Require a **running clause** after the break — a later lowercase-initial word — which a real second sentence has (`flow. Reduced churn by…`) but an all-Title-Case `City, State` location tail does not. Real multi-sentence glyph-less descriptions still classify as prose; the header is now claimed and both roles segment, each keeping its own bullets.

## Corpus impact (net positive)

| Fixture | experienceCount | Why |
|---|---|---|
| `…two-column-long-role-header` (#352 repro) | 1 → **2** | the fix |
| `latex/awesome-cv-cv` | 16 → **17** | **same root fix recovers a real role** this CV was dropping (an `Undergraduate Research, … Lab(Prof. …)` header the old false-positive swallowed) — every other role byte-identical |

The recovered `awesome-cv-cv` role has an inline-abbreviated-date header (`Researcher … Mar. 2016 Exp. Jun. 2017`) that doesn't round-trip cleanly yet — a **separate** header-shape weakness — so its experience round-trip is baselined in `KNOWN_FAILURES` (the sanctioned ratchet) as a documented follow-up. Recovering the role beats dropping it.

## Test plan

- [x] `/probe-experience` on the #352 fixture: UNDER-SEGMENTED → `Verify: ok`, 2 entries, each with its own bullets
- [x] New unit test: `extractExperience` two-role stacked-header case (`extract-fields.test.ts`)
- [x] Full suite **1623 passed / 0 failed** on current main; `tsc -b --noEmit` + `eslint .` clean; `vite build` OK
- [x] Snapshot changes are real content diffs (both `experienceCount`), not line-ending churn

## Out of scope (separate issues)

- **Title/company split** for the surviving role (whole `Title, Company. City, State` still lands in `title`) — the **#342**-class defect (needs US-full-state-name location handling), not required by #341's ACs.
- **⚠️ PII:** while probing I found `latex/awesome-cv-cv.pdf` appears to be the Awesome-CV author's **real CV** (real employers) — exactly what the repo PII policy warns about. Not touched here; worth its own issue + `git filter-repo` remediation. The re-baked snapshot is lossy (counts only), so no new PII surface.

Closes #341


src/lib/heuristics/line-primitives.ts
```diff
@@ -34,18 +34,30 @@ export function stripBullet(text: string): string {
...
  * Glyph-less descriptions WITHOUT a sentence period (e.g. indented one-line
  * bullets) are left to the bullet/indent path, unchanged by this predicate.
  */
 const PROSE_MIN_WORDS = 8;
-const SENTENCE_BREAK_RE = /[a-z]{2}\.\s+[A-Z]/;
+// `word. Capital…` (a sentence break) followed, before the next period, by a
+// space + lowercase letter (a real second clause). The trailing lowercase is
+// what separates a running sentence from a "Company. City, State" location tail
+// (all Title-Case, no lowercase word → not prose). See #341.
+const SENTENCE_BREAK_RE = /[a-z]{2}\.\s+[A-Z][^.]*\s[a-z]/;
```
s-annam, line 60: NIT (non-blocking): the tightened regex is well-scoped and the `[a-z]{2}\.` prefix nicely defends against the `Co.`/`Inc.` abbreviation false-positive. One residual over-tightening tradeoff worth noting: a genuine two-sentence description whose *second* sentence is entirely Title-Case up to its period (e.g. `…relations. Improved Delivery Times Across Regions.`) now fails the lowercase-continuation guard and reads as a header. That shape is rare in real descriptions (they almost always contain an article/verb) and is a strictly better failure mode than the #341 bug, so this is fine to ship — just flagging the boundary. Verified empirically alongside the fix.

src/lib/heuristics/corpus-roundtrip.test.ts
```diff
@@ -123,6 +123,16 @@ const KNOWN_FAILURES: Record<string, Category[]> = {
...
 
+  //   - awesome-cv-cv: the #341 isProseLine fix RECOVERS a real role this CV was
+  //     dropping (an "Undergraduate Research, … Lab(Prof. …)" header the old
+  //     "Company. City" prose false-positive had swallowed). Net +1 real role
+  //     (16 → 17). That recovered role's header packs inline abbreviated dates
+  //     ("Researcher … Mar. 2016 Exp. Jun. 2017"), so its title/start_date don't
+  //     round-trip cleanly yet — a SEPARATE header-shape weakness tracked as a
+  //     follow-up. Recovering the role beats dropping it, so the experience
+  //     round-trip is baselined here rather than blocking the #341 fix.
+  "latex/awesome-cv-cv.pdf": ["experience"],
```
s-annam, line 134: NIT (non-blocking): this KNOWN_FAILURES baseline is the right call — recovering a real role (16→17) beats dropping it, and the comment documents the abbreviated-date round-trip weakness clearly. Only gap: the comment says "tracked as a follow-up" but cites no issue number, so the ratchet is orphaned. Please file (or link) a GitHub issue for the `Researcher … Mar. 2016 Exp. Jun. 2017` inline-abbreviated-date header-shape and reference it here, so this baseline can be un-ratcheted later. Same for the `awesome-cv-cv.pdf` real-PII concern you flagged in the PR body (pre-existing; the binary is unchanged here, so not blocking this PR).

s-annam:
**Review: APPROVE** ✅

Verified the root cause and fix end to end. The `SENTENCE_BREAK_RE` tightening (requiring a lowercase continuation after the `word. Capital` break) correctly stops a `Company. City, State` header tail from reading as a sentence, while genuine multi-sentence prose still classifies as prose — and the `[a-z]{2}\.` prefix keeps `Co.`/`Inc.` abbreviations from false-positiving. Fix is one line, narrowly scoped, and well-documented.

- Full suite green: **1623 passed / 0 failed**; typecheck + lint clean.
- Corpus deltas are real content diffs (`experienceCount` 1→2 on the #352 repro, 16→17 on `awesome-cv-cv`), and the recovered awesome-cv role is honestly baselined in `KNOWN_FAILURES` rather than masked.
- New unit test covers the two-role stacked-header case.

I pushed one nit commit (`32d6a93`): a positive-prose regression test that locks in the *other* side of the tightening (a genuine glyph-less multi-sentence description must still be captured as prose) — previously only covered indirectly via the Word corpus fixtures.

Two non-blocking nits left inline: (1) the `KNOWN_FAILURES` baseline should cite a tracked follow-up issue number; (2) the `awesome-cv-cv.pdf` real-PII concern (pre-existing, binary unchanged here) deserves its own issue.

s-annam (dismissed):
APPROVE — correct, narrowly-scoped fix; full suite green (1623/0). Details in the summary comment. Pushed one nit test commit (32d6a93); two non-blocking nits left inline.

s-annam (approved):
Re-approve after rebasing onto main (#362 merged). Conflict was purely additive in extract-fields.test.ts — kept both the #341 tests and the incoming #342 test (keep-both); experience.ts auto-merged clean and both #341/#342 feature tests pass together. Also references #386 for the previously-orphaned abbreviated-date round-trip ratchet. Full suite green.

Issue #298, opened 2026-07-03
Round-trip: experience header re-segmentation (title/company swap) in dense/two-column layouts


s-annam:
## Symptom

On re-parse of the reconstructed résumé, **experience entries mis-segment**: title/company (and sometimes dates) swap or shift, or a single role splits / two roles merge. Shows up in denser, multi-role, and two-column source layouts. Extends the #284 experience-stacking fix — those cases stack correctly but the re-parser still re-segments the reconstructed output wrong.

## Affected fixtures (round-trip gate `experience` category)

Experience only:
- `google-docs/google-docs-skia-proxy-certifications.pdf`
- `latex/awesome-cv-cv.pdf`
- `unknown/chromium-two-column-sidebar.pdf`
- `unknown/two-column-achievements-sidebar.pdf`

Experience + skills (skills part → #E):
- `google-docs/google-docs-skia-proxy-role-first-experience.pdf`
- `latex/deedy-resume-macfonts.pdf`
- `latex/deedy-resume-openfonts.pdf`
- `unknown/openresume-react-pdf.pdf`

Experience + education + skills (experience part here):
- `google-docs/google-docs-skia-proxy-programs-skills-software.pdf`

## Root cause (hypothesis)

The reconstructed experience block emits title/company/date on lines whose ordering or whitespace the re-parser's header detection (`openresume.ts` / `regex.ts` anchors) re-segments differently than parse-1 — e.g. company-first vs role-first, or a date landing where a header anchor expects it. Two-column source layouts feed a different parse-1 shape that `buildAtsResumeModel` normalizes in a way the re-parser then mis-reads.

## Fix direction

- Harness-trace one pure-experience fixture (`certifications` or `two-column-achievements-sidebar`); compare parse-1 experience entries vs the reconstructed lines vs parse-3.
- Make the reconstructed experience header shape unambiguous to our own re-parser (title on bold header, `Company · Location  Dates` sub-line — mirror the #291 education stacking).

## Definition of done

- Experience count + title/company/start_date/end_date round-trip for all listed fixtures.
- Delete the `experience` category from each listed line in `KNOWN_FAILURES`; corpus gate (#293) stays green.

Note: `role-first`, `deedy×2`, `openresume-react-pdf`, `programs-skills-software` also carry a `skills` failure tracked in #299 — leave those skills entries until #E lands. If the experience fix here incidentally clears the skills split, close #299 as covered.

[...]

PR #362, opened 2026-07-07
fix(experience): role title from the line below a company+dates anchor (#342)

### LINKED ISSUE #342: Single-column layout: employment role (title) not detected properly on experience entries

rohithgollapalli:
## Summary

Fixes **#342** (single-column: employment role/title not detected). Modeled off a real résumé (probed locally, PII-free synthetic recreation committed).

## Root cause — header→field mapping, not segmentation

Single-column "date-first" résumés put `Company [— Dept] — Location  Dates` on the anchor (date) row and the job **title on the line BELOW it**:

```
Brightwave Analytics LLC — Remote, USA   Jul 2024 - Present   ← anchor (company + location + dates)
Data Analyst Intern                                           ← title, BELOW the anchor
• bullets…
```

Entry segmentation is already **correct** — the title line IS captured into `headerLines` below the anchor. The miss is entirely in `disambiguateCompanyTitle`, which only knew the *stacked* shape (`Title` above / `Company+Dates` on the anchor — #298). For the **inverse** layout it filled the title slot from an anchor-row segment and dropped the real below-anchor title:

- role 1 → `title: ""` — the anchor-row location `Remote, USA` took the title slot; `Data Analyst Intern` was demoted to `team`;
- role 2 → `title: "IT Service Desk"` — the **department** on the anchor row was taken as the title, dropping the real `Data Analyst` below.

## Fix

In the company-found branch, prefer a **below-anchor** split (`source > anchorIdx`) that `looksLikeTitle` as the role title — symmetric to the existing "title above" tiebreak, for the title-below-anchor layout:

```ts
const belowTitle = anchorIdx !== undefined
  ? others.find((s) => s.source > anchorIdx && looksLikeTitle(s.text))
  : undefined;
if (belowTitle) { title = belowTitle.text; team = others.find((s) => s !== belowTitle)?.text; }
else { title = others[0]?.text; team = others[1]?.text; }  // unchanged prior mapping
```

**Additive & low-risk:** the `else` path is the exact prior behavior, so nothing changes unless there's a below-anchor title-shaped split — precisely the case the old code got wrong. Existing shapes (stacked `Title above`, single-line headers) have no below-anchor title split, so they're untouched. Guarded with `looksLikeTitle` (not a weaker not-a-company/not-a-location test) to avoid promoting a stray line.

## The fixture

`tests/fixtures/pdfs/unknown/single-column-title-below-anchor.pdf` — synthetic, `pdftotext`-verified persona: **Jordan Avery / jordan.avery@example.com / (614) 555-0137**. Single column (`triggers: []`), two title-below-anchor roles (one with an extra `— Dept —` segment). Unit test in `extract-fields.test.ts` pins both titles (`Data Analyst Intern`, `Data Analyst`) and rejects the department-as-title.

## Test plan

- [x] Reproduced from the real résumé via `/probe-experience`, then fixed; both titles now correct
- [x] Unit test pins the fix (red without it: role 1 `title=""`, role 2 `title="IT Service Desk"`)
- [x] **Full corpus unchanged — no snapshot moved** (the fix changes title *values*, which the lossy snapshots don't record; counts stable)
- [x] Full suite **1625 passed / 0 failed**; `tsc -b --noEmit` + `eslint .` clean; `vite build` OK
- [x] Fixture persona synthetic + phone-policy-compliant

## Out of scope

The role-2 location `Columbus, Ohio` (full state name, not `OH`) still isn't peeled into `location` — a **pre-existing** full-state-name location gap (also seen in #341's fixtures), independent of #342's title bug.

Closes #342


src/lib/heuristics/extract/experience.ts
```diff
@@ -508,8 +508,25 @@ function disambiguateCompanyTitle(
...
+    // `team` or dropped. Symmetric to the "title above" tiebreak (#298), for the
+    // inverse title-below-anchor layout. Additive: when no such split exists the
+    // `else` keeps the exact prior mapping, so only the broken case changes.
+    const belowTitle =
+      anchorIdx !== undefined
+        ? others.find((s) => s.source > anchorIdx && looksLikeTitle(s.text))
+        : undefined;
+    if (belowTitle) {
+      title = belowTitle.text;
+      team = others.find((s) => s !== belowTitle)?.text;
```
s-annam, line 525: NIT (non-blocking, already scoped-out in the PR): `team` takes the first remaining split, which for role 1 is the anchor-row segment `Remote, USA   Jul 2024 - Present` — location + the raw date range. For role 2 the date-bearing `Columbus, Ohio ...` segment is instead dropped entirely (team becomes the department `IT Service Desk`). So the below-anchor title is now correct, but where the date/location lands is inconsistent between the two shapes. This is the pre-existing full-state-name location gap you flagged as out of scope (also #341), not a regression from this diff — the unit test correctly pins only `company`/`title`. Flagging only so the follow-up location work knows `team` can currently carry a date string.

s-annam (approved):
Reviewed the title-below-anchor fix end to end. Correct, additive, and well-guarded; approving. One non-blocking note inline. Detail in the top-level summary.

s-annam:
**Review: APPROVE** ✅

Verified the fix against #342: in the company-found branch, `disambiguateCompanyTitle` now prefers a below-anchor split that `looksLikeTitle` as the role title (symmetric to the #298 "title above" tiebreak). Traced both fixture roles by hand — role 1 title `Data Analyst Intern` (was `""`), role 2 title `Data Analyst` (was the department `IT Service Desk`). The `else` branch is byte-identical to the prior mapping, so only the previously-broken shape changes.

- Correctness: sound and additive; guarded with `looksLikeTitle` (not a weaker not-a-company test).
- Tests: new unit test in `extract-fields.test.ts` pins both titles + rejects department-as-title. `extract-fields` + `corpus` suites green. `tsc -b` and `eslint .` clean.
- Fixture: `unknown/single-column-title-below-anchor.pdf` — PII-verified synthetic (Jordan Avery / (614) 555-0137 [real Columbus area code, 555 exchange, 0137 subscriber] / @example.com). Under `unknown/` so the HeadlessChrome/Skia producer carries no miscategorized-provenance issue.
- One non-blocking NIT left inline re: `team` carrying a date-bearing location segment (pre-existing, already scoped out).

Note on the full suite: 3 PDF-render round-trip tests (`group-bullets`, `multi-experience-roundtrip`, `corpus-roundtrip`) intermittently hit the 5s per-test timeout under parallel load — a *different* test each run — and all pass cleanly in isolation (71/71). Pre-existing flakiness, unrelated to this diff.

PR #385, opened 2026-07-07
fix(parser): stop promoting an education institution to full_name (#349)

### LINKED ISSUE #349: Name heuristic promotes an education institution as full_name when the name is absent from the text layer (Deedy)

Vaishnavi1709:
## Summary

On the Deedy two-column fixture, the name heuristic used to promote **`CORNELL UNIVERSITY`** to `full_name` @ 0.5 — the centred top-of-page name straddles the column split, so the column-ordered reading-order flatten pushes it out of the profile band while the contact line stays; the first title-case-shaped line remaining in front of the scanner is an all-caps education entry. A false-positive name is worse than a missing one: it renders wrong in the reconstructed résumé and earns undeserved completeness credit.

Two-part fix:

1. **Reject institution + job-title lines as name candidates** — in both `extract/name.ts` (Tier 1) and `regex-fallback.ts` (Tier 1.5), any candidate matching `INSTITUTION_HINTS` (`University|College|Institute|School|Academy|Polytechnic`) is rejected; the fallback additionally rejects `looksLikeTitle` candidates so the reconstructed name-less Download PDF doesn't round-trip "Software Engineer" as a wrong name. Real names never carry these words, so the guards only ever reject false positives.
2. **Recover the top-of-page name when column reorder hid it** — when the primary `extractName` returns nothing AND the layout is two-column, rebuild the profile section from the raw items *without* column reordering and retry just the name extractor. Read-only, name-only: every other extractor still runs against the column-ordered primary sections.

Together these hit the issue's better acceptance option — `deedy-resume-openfonts.pdf` now correctly parses `full_name = "Jane Smith"` (recovered from the text layer). Sibling `deedy-resume-macfonts.pdf` verified the same. Fixture snapshots re-baked; round-trip invariants preserved.

Closes #349

## Test plan

- [x] `npm run verify` clean (1629 tests pass, 3 skipped)
- [x] Regression tests added for Tier 1 (`extract-fields.test.ts`), Tier 1.5 (`regex-fallback.test.ts`), and the two-column recovery path (`openresume.test.ts`)
- [x] Manually verified in `npm run dev` — dropped `deedy-resume-openfonts.pdf`, header now shows "Jane Smith" (was "Name not detected" after step 1, "CORNELL UNIVERSITY" before this fix)
- [x] Corpus snapshots updated for both Deedy fixtures — `full_name` / `given_name` / `family_name` back in `fieldsPopulated`, "name" out of `completeness.missing`, overall score 43 → 45, cascade confidence 0 → 0.64

### DIFF
--- a/src/lib/heuristics/extract-fields.test.ts
+++ b/src/lib/heuristics/extract-fields.test.ts
@@ -483,6 +483,52 @@ describe("extractName — stacked single-word name lines (#29)", () => {
   });
 });
 
+describe("extractName — institution-line rejection (issue #349)", () => {
+  // Two-column layouts (Deedy) can flatten the reading order so an education
+  // institution ends up in the profile-adjacent candidate pool with the visible
+  // name still centred in the header. An all-caps institution line ("CORNELL
+  // UNIVERSITY") passes the loose title-case gate on its own and used to be
+  // promoted to `full_name` — a false-positive worse than an empty one. The
+  // fix rejects any candidate carrying an institution-type word, using the
+  // same INSTITUTION_HINTS the education extractor recognizes.
+  it("rejects 'CORNELL UNIVERSITY' as a name candidate and returns empty when nothing survives", () => {
+    const { profile } = buildContext([
+      { text: "jane.smith@example.com | (312) 555-0123", fontSize: 10 },
+      { text: "CORNELL UNIVERSITY", fontSize: 12 },
+      { text: "MEng in Computer Science", fontSize: 10 },
+    ]);
+    const result = extractName(profile);
+    expect(result.value).toBeUndefined();
+    expect(result.confidence).toBe(0);
+  });
+
+  it("still picks a real name over a nearby institution line", () => {
+    const { profile } = buildContext([
+      { text: "Jane Smith", fontSize: 20 },
+      { text: "jane.smith@example.com", fontSize: 10 },
+      { text: "CORNELL UNIVERSITY", fontSize: 12 },
+    ]);
+    expect(extractName(profile).value).toBe("Jane Smith");
+  });
+
+  it("rejects each institution keyword (University/College/Institute/School/Academy/Polytechnic)", () => {
+    for (const inst of [
+      "Harvard University",
+      "Riverside College",
+      "MIT Institute",
+      "St Mary School",
+      "Naval Academy",
+      "Rensselaer Polytechnic",
+    ]) {
+      const { profile } = buildContext([
+        { text: "jane.smith@example.com", fontSize: 10 },
+        { text: inst, fontSize: 12 },
+      ]);
+      expect(extractName(profile).value).toBeUndefined();
+    }
+  });
+});
+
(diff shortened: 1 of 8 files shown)


s-annam (approved):
## Review verdict: APPROVE

Autonomous review of the two-part name-heuristic fix for #349.

**What it does, verified against the code:**
- **Tier 1 (`extract/name.ts`)** — adds an `INSTITUTION_HINTS` guard to `nameCandidateWords`. Job-title rejection was already present (`looksLikeTitle` at L74/L205), so the institution guard is the only new gate needed here. Correct.
- **Tier 1.5 (`regex-fallback.ts`)** — the standalone `guessNameFromFirstLines` scanner needs both guards (`INSTITUTION_HINTS` + `looksLikeTitle`) because it doesn't share Tier 1's scoring path. Correct; the asymmetry is intentional, not a gap.
- **Two-column name recovery (`openresume.ts`)** — `findNameFallbackProfile` rebuilds the profile from raw items *without* column boundaries and retries `extractName` only when the primary path found no name. Read-only, name-only; every other extractor still runs against the column-ordered sections. Gated on `!singleColumn` so single-column docs are untouched.

**Gates (re-run locally):**
- `tsc -b --noEmit` clean.
- Affected tests: 94 pass (`extract-fields`, `openresume`, `regex-fallback`).
- Corpus snapshots: 39 pass — re-baked Deedy fixtures consistent.

**Severity tally:** 0 blocking, 0 suggestion, 0 nit. Tightly scoped, well-documented, well-tested. Ship it.

Issue #355, opened 2026-07-07
Single-column label-rail: inline & stacked section headers unrecognized → skills leak into experience, orphaning role-1 header


s-annam:
## Summary

**Category: single-column with a section-label rail.** On a résumé that puts each section name in a narrow left rail sharing the row with the section's first entry — `Experience  Software Engineer II, …  Aug 2024 - Present` on one physical line — the section header is **not recognized**, so the section never opens. Content lands in the wrong band, and the first role's title/company come back as **skill tokens** from the rows above the date.

This is **not** a two-column / skills-sidebar layout. An earlier draft of this issue mis-diagnosed it as sidebar interleave; a geometry probe disproved that (see Root cause). The fix is in **section-header recognition** (`sections.ts` / `matchSectionHeader`), not column separation.

Found via the `/probe-experience` sweep on a real (uncommitted, PII) résumé; repro below is synthetic.

## Root cause (localized by layer)

**Section routing — header recognition** (`sections.ts` `matchSectionHeader`, with `matchAnchorFallback` in `regex.ts`). Two recognition gaps, one root class:

1. **Inline / rail header.** The section keyword is the **leading token** of a row that also carries the section's first entry (`Experience`, `Education` share the row with the role/degree and its date). `matchSectionHeader` is exact-match, so a multi-token line never matches; `matchAnchorFallback` requires the head noun **last** (≤4 words), but here the head noun is **first** on a long row — no match. The section header is lost; the line is treated as content.

2. **Stacked grid header.** The Skills header is split vertically (`Technical` / `Skills` on two rows) as the left cell of a horizontal skills grid. Neither single-line token matches the exact alias `technical skills`, so Skills never opens either.

**Why the layout probe is empty (not two-column):** the ink-projection gutter detector (`detectColumnBoundaries`, `pdf-layout.ts`) reports **no corridor** — minimum row-coverage across the entire central band is ~43% (threshold is 4%), and `triggers` is `[]` (`two_column` never fires). There is no vertical gutter to separate; the real rail gutter (between the ~narrow label column and the body) sits at ~frac 0.13, below the detector's `TWO_COLUMN_BAND_LO` floor (0.2), and — correctly — is not a two-column split. **Do not touch the two-column detector for this issue; loosening its calibrated thresholds would regress the labeled corpus with no gain here.**

## Cascade (why role 1's header becomes skill tokens)

[...]

PR #389, opened 2026-07-07
test: scope 20s timeout to flaky fixture-read/roundtrip/render suites (#360)

### LINKED ISSUE #360: chore: raise testTimeout on flaky fixture-read/roundtrip suites (verify timeouts under load)
## Summary

Several fixture-reading / roundtrip tests intermittently fail the pre-push `npm run verify` gate with **`Test timed out in 5000ms`** — not assertion failures, purely the default vitest 5s per-test timeout being exceeded under load. They pass cleanly in isolation and on a re-run of the same full suite, so this is a **flaky infra timeout, not a real regression**. It reds the gate non-deterministically and erodes trust in `verify`.

## Evidence (observed during PR #359 / issue #355)

- Pre-push `npm run verify` failed: `Tests 3 failed | 1627 passed`; every failure was `Test timed out in 5000ms` (0 assertion failures).
- Affected files:
  - `src/lib/heuristics/corpus-roundtrip.test.ts`
  - `src/lib/heuristics/multi-experience-roundtrip.test.ts`
  - `src/lib/pdf/skills-wrap-roundtrip.test.ts`
  - `src/lib/score/group-bullets.test.ts`
- Re-run of the **same** 4 files in isolation: **74 passed** (0 fail).
- The **very next** full `npm run verify` on the identical tree: **1632 passed**, 0 fail.

## Root cause (hypothesis)

These tests read a real PDF fixture and run the full `runCascade(...)` pipeline. Under the `verify` run they hit three compounding slowdowns absent when run alone:

1. **Coverage instrumentation** (`verify` runs coverage) inflates per-test CPU time.
2. **Full-suite concurrency** — 133 test files contend for the worker pool, so a cascade run that takes ~1–2s alone can stall past 5s while workers are saturated.
3. **Pre-existing Poppins font-URL fallback** — the render path logs `Poppins font embed failed, falling back to Helvetica: TypeError: Failed to parse URL from /src/assets/fonts/Poppins-Regular.ttf` repeatedly under vitest (the `?url` asset resolves in the browser/build but not the Node test transform), adding avoidable per-call work to the roundtrip render.

The default `testTimeout` is 5000ms; a slow-but-correct cascade+render under load exceeds it.

## Fix direction

Raise the timeout for the fixture-read / roundtrip suites so a slow-but-correct run doesn't red the gate. Options (pick the least-invasive that holds):

- **Per-test/per-suite timeout** — pass a timeout arg on the heavy `it(...)`/`describe(...)` blocks (e.g. `it("...", async () => {...}, 20000)`), or `describe.concurrent`-level config. Most targeted.
- **Scoped config** — a higher `testTimeout` for the roundtrip/cascade file glob (vitest `test.testTimeout` is global; use per-test args or a dedicated project entry if file-scoping is needed).
- Optionally investigate silencing the Poppins font-URL fallback under the Node test transform (provide a resolvable `standardFontDataUrl` / font path in the test env), which removes the avoidable slowdown at the source — smaller blast radius than a global timeout bump, and fixes the noisy log too. Can be a follow-up if out of scope here.

Do **not** globally disable coverage or serialize the suite — those trade real signal/speed for the timeout.

## Acceptance criteria

- [ ] The 4 listed suites no longer time out under a full coverage-instrumented `npm run verify` run (verify green across several consecutive runs).
- [ ] The chosen timeout increase is **scoped to the fixture-read/roundtrip tests**, not a blanket global `testTimeout` bump that would mask genuinely-slow regressions elsewhere.
- [ ] No test's assertions are weakened or skipped — only the timeout budget changes.
- [ ] (If addressed) the `Poppins font embed failed … Failed to parse URL` log no longer floods the test output; otherwise noted as a deferred follow-up.
- [ ] `npm run verify` stays green.

## Notes

- Surfaced while landing PR #359 (issue #355, label-rail section recognition). That change is unrelated (section recognition in `sections.ts`); it merely ran the full suite under load and tripped the latent flake.

s-annam:
## Summary

Several fixture-read + full `runCascade→render→re-parse` suites — plus the Poppins font-embed render suite — intermittently hit vitest's default **5000ms** per-test timeout under the coverage-instrumented full-suite `verify` run: worker-pool contention + coverage instrumentation, not assertion failure. A flaky infra timeout that reds the gate non-deterministically (surfaced during PR #359 / issue #355).

Scope a **20000ms** budget to just the heavy paths, leaving vitest's global default and all fast in-memory tests untouched — so a slow-but-correct run doesn't red `verify` while genuinely-slow regressions elsewhere stay caught.

Per file:
- `corpus-roundtrip.test.ts` — per-`describe` `{ timeout: 20000 }`
- `multi-experience-roundtrip.test.ts` — per-`describe` `{ timeout: 20000 }`
- `skills-wrap-roundtrip.test.ts` — per-`beforeAll` hook timeout (heavy render path only)
- `group-bullets.test.ts` — per-`it` trailing arg on the single fixture-read test
- `render-ats-pdf.fonts.test.ts` — per-`describe` `{ timeout: 20000 }` (fontkit Poppins-embed renders; caught by the Stop-hook `verify` run after the first four)

No assertions weakened or skipped — only timeout budgets change. The Poppins font-URL fallback log noise is left as a deferred follow-up per the issue.

Closes #360

## Test plan

- [x] The target suites pass in isolation
- [x] `npm run lint` clean
- [x] Pre-push `npm run verify` green (full coverage-instrumented suite)

### DIFF
--- a/src/lib/heuristics/corpus-roundtrip.test.ts
+++ b/src/lib/heuristics/corpus-roundtrip.test.ts
@@ -271,7 +271,10 @@ const CATEGORIES: Category[] = [
   "render",
 ];
 
-describe("corpus round-trip invariants (#293)", () => {
+// Fixture-read + full runCascade→render→runCascade round-trip per fixture is
+// slow under a coverage-instrumented full-suite `verify` run; scope a higher
+// timeout to just this suite rather than bumping vitest's global default (#360).
+describe("corpus round-trip invariants (#293)", { timeout: 20000 }, () => {
   const fixtures = walkPdfs(FIXTURE_ROOT);
 
   it("finds fixtures to round-trip", () => {
--- a/src/lib/heuristics/multi-experience-roundtrip.test.ts
+++ b/src/lib/heuristics/multi-experience-roundtrip.test.ts
@@ -39,7 +39,10 @@ function distinctLabels(exp: ReadonlyArray<{ section_label?: string }>): string[
   return out;
 }
 
-describe("#311 multiple experience sections — parse + round-trip", () => {
+// Fixture-read + runCascade/render round-trip is slow under a
+// coverage-instrumented full-suite `verify` run; scope a higher timeout to
+// just this suite rather than bumping vitest's global default (#360).
+describe("#311 multiple experience sections — parse + round-trip", { timeout: 20000 }, () => {
   it("parses two experience-category sections into labeled roles", async () => {
     const bytes = await fsp.readFile(FIXTURE);
     const cascade = await runCascade(new Uint8Array(bytes));
--- a/src/lib/pdf/render-ats-pdf.fonts.test.ts
+++ b/src/lib/pdf/render-ats-pdf.fonts.test.ts
@@ -79,7 +79,11 @@ const model = (text: string): AtsResumeModel => ({
   sections: [],
 });
 
-describe("Poppins font embed (#314)", () => {
+// Each case does a real fontkit Poppins-embed render (the failing glyph case
+// renders twice); slow under a coverage-instrumented full-suite `verify` run,
+// so scope a higher timeout to just this suite rather than bumping vitest's
+// global default (#360).
+describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
   afterEach(() => {
     vi.unstubAllGlobals();
     vi.resetModules();
--- a/src/lib/pdf/skills-wrap-roundtrip.test.ts
+++ b/src/lib/pdf/skills-wrap-roundtrip.test.ts
@@ -89,6 +89,9 @@ const SYNTHETIC_SKILLS = [
 describe("#301 — multi-word skill does not split at the line-wrap boundary", () => {
   let reparsedSkills: string[];
 
+  // Fixture-read + runCascade/render round-trip is slow under a
+  // coverage-instrumented full-suite `verify` run; scope a higher timeout to
+  // just this hook rather than bumping vitest's global default (#360).
   beforeAll(async () => {
     const original = await runCascade(new Uint8Array(readFileSync(FIXTURE)));
     const withSyntheticSkills: CascadeResult = {
@@ -102,7 +105,7 @@ describe("#301 — multi-word skill does not split at the line-wrap boundary", (
     const bytes = await renderAtsResumePdf(model);
     const reparsed = await runCascade(bytes);
     reparsedSkills = reparsed.parsed.skills ?? [];
-  });
+  }, 20000);
 
   it("round-trips the same skill count (AC)", () => {
     expect(reparsedSkills.length).toBe(SYNTHETIC_SKILLS.length);
--- a/src/lib/score/group-bullets.test.ts
+++ b/src/lib/score/group-bullets.test.ts
@@ -236,6 +236,9 @@ describe("multi-line bullet pool is fully merged and correctly attributed (#162)
   // Merging wrapped continuations upstream (`mergeWrappedContinuations` in
   // `toSectionedResume`) makes the pool carry each bullet's full text, so it
   // matches its role by construction.
+  // Fixture-read + runCascade round-trip is slow under a coverage-instrumented
+  // full-suite `verify` run; scope a higher timeout to just this test rather
+  // than bumping vitest's global default (#360).
   it("recovers full wrapped-bullet text and attributes it to its role, not Other", async () => {
     const bytes = await fsp.readFile(
       join(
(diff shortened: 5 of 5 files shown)


Vaishnavi1709 (approved):
LGTM to un-flake `verify`. No correctness bugs — all three vitest APIs used (`describe(name, options, fn)`, `beforeAll(fn, 20000)`, `it(name, fn, 20000)`) are valid in installed vitest 3.2.6, and the PR correctly uses the new options-as-2nd-arg `describe` shape (the positional-options form is deprecated and will throw in vitest 4).

Nothing blocking, but two concrete gap-coverage nits and a couple of shape/altitude observations worth landing as a follow-up. Approving so this doesn't block CI.

### Two concrete gaps the timeout doesn't cover

1. **`src/lib/pdf/render-ats-pdf.fonts.test.ts:87` — `afterEach(vi.resetModules)` isn't covered.** describe-level `{ timeout: 20000 }` sets *testTimeout* only; hooks fall back to `hookTimeout` (a separate option, 10s default in vitest 3.x). `vi.resetModules()` invalidates pdf-lib+fontkit caches — the exact instrumented-import cost this PR calls out — but on the hook side, not the test side. If it exceeds the default `hookTimeout` under coverage, afterEach fails, the next case starts with a partly-reset module registry, and a downstream test dies with a misleading "fetchMock not called" error.

2. **`src/lib/pdf/skills-wrap-roundtrip.test.ts:130` — sibling describe not covered.** The second `describe("wrapSegmentsToLines — overlong FIRST segment…")` at line 130 does its own `loadPdfLibOnce()` + `embedFont(StandardFonts.Helvetica)` — a real pdf-lib cold-load path — and inherits none of the `beforeAll` bump (which is scoped to the earlier `#301` describe only). On a cold `npm run verify` where this file runs first, the pdf-lib dynamic import under V8 coverage can exceed 5s in that sibling too.

### Shape / altitude (worth considering as follow-up)

3. **Standardize on one form.** 3 files use `describe(name, { timeout: 20000 }, fn)`; `skills-wrap` uses positional `beforeAll(fn, 20000)`; `group-bullets` uses positional `it(name, fn, 20000)`. `group-bullets` has only one `it` in that describe — promoting to the describe-options form is behaviour-equivalent and reads uniformly across the PR.

4. **Extract to a shared constant.** The identical 3-line rationale comment ("slow under a coverage-instrumented full-suite `verify` run… #360") is copy-pasted 5×, and the magic `20000` is inlined 5×. One `export const SLOW_SUITE_TIMEOUT_MS = 20_000` in `src/lib/heuristics/__test-utils__/timeouts.ts` (already coverage-excluded in `vite.config.ts:189`) with the JSDoc rationale collapses all five sites.

5. **Root-cause vs bandaid.** The commit message pins the flake on coverage-instrumented full-suite runs — a workload property. Special-casing 5 files means the next slow round-trip suite hits the same 5s default and gets its own copy-paste. A deeper fix is a coverage-gated project-wide `testTimeout` in `vite.config.ts` (e.g. `testTimeout: process.env.VITEST_COVERAGE ? 15000 : 5000`), reserving suite-scoped `20000` only where genuinely needed. Would also cover future slow suites for free.

6. **Value chosen without a baseline.** 4× vitest's default gives real perf regressions a large hiding zone — an O(n²) regression in per-fixture `runCascade` or fontkit CFF subsetting could drift ~500ms → ~15s and still pass. A measured baseline + tighter bound (say 2× the observed max) would keep the flaky-detector's teeth.

### Nits

7. **`skills-wrap-roundtrip.test.ts:90`** — `let reparsedSkills: string[]` uninitialised. If `beforeAll` throws mid-hook (after `runCascade` but before assignment), dependent `it`s crash with `Cannot read properties of undefined (reading 'length')`, noising the log around the real `beforeAll` failure. Initialising to `[]` keeps the failure clean.

8. **`corpus-roundtrip.test.ts:423`** — sibling `describe.runIf(process.env.RL_RT_PDF)` dev harness is outside the modified describe and doesn't inherit the 20s. Its test does N render→re-parse hops (`RL_RT_ROUNDS`), so at `rounds>=2` under coverage it can time out during triage — probably worth the same bump.

9. **`corpus-roundtrip.test.ts:280`** — the two trivial front-of-suite `it`s ("finds fixtures", "every KNOWN_FAILURES key names a real fixture") inherit the 20s too. A `walkPdfs` refactor that hangs on a malformed fixture tree now blocks 20s per run instead of 5. Minor, but the point-scoping in (3) would fix it as a side-effect.

PR #387, opened 2026-07-07
feat: add implement-batch skill for GitHub epic implementation

s-annam:
## Summary

Adds a project-local `/implement-batch` skill: a self-contained, GitHub-only orchestrator that implements a set of GitHub issues (an epic's sub-issues or an explicit list) onto **one branch** as a **single accumulated commit**, one issue per isolated subagent, then runs an **adversarial review pass** and opens **one PR** via `/open-pr`.

It's the trimmed sibling of the global `/implement-epic`, but:
- **GitHub-only** — no Linear code. Sub-issues via `gh api .../sub_issues`, dependency order via `.../dependencies/blocked_by`.
- **No dependency on the global `/implement-issue` skill** — that skill lives in a maintainer's `~/tools/skills/`, so interns/clones don't have it. The per-issue implementation contract is **embedded in the subagent spawn prompt** instead, so the skill works for anyone who clones the repo.
- **Adversarial review built in** (`ecc:react-reviewer`, ≤2 rounds, in-place fix subagent), on by default, pre-PR — matches the run-review-ourselves workflow.
- **Finalizes via the repo-local `/open-pr`** (commit + push + PR + fixture-PII preflight). One branch, one commit, one PR.

Bakes in resumelint facts: protected `main` + `verify` gate, design-system/token rules, fixture-PII policy, SPDX headers.

Doc-only change — one new `SKILL.md`, no runtime code touched.

## Test plan

- [x] N/A — markdown-only skill definition, no runtime surface. CI `verify` runs on the PR.

PR #390, opened 2026-07-07
feat: add batch-issues skill for grouping small issues into one epic

s-annam:
## Summary

Adds a project-local `/batch-issues` skill — the **grouping** counterpart to `/implement-batch` (#387). It turns a loose set of small GitHub issues into one durable, assignable unit that a single person implements end-to-end.

Given an explicit list of issue numbers, it:
- mints a parent **"batch" epic** tracking issue (checklist body + a *why grouped* line),
- wires the children as **native sub-issues** (`gh api .../sub_issues`, `-F sub_issue_id=<internal id>`) with optional `blocked_by` ordering,
- applies a shared **milestone + `batch:<slug>` label + assignee** (without clobbering good child metadata),
- **delegates board placement to `/triage-issue`** (no duplicated project logic),
- prints the ready **`/implement-batch <PARENT>`** invocation.

Design notes baked in: a **cohesion gate** (refuses closed/DUP issues, warns on mixed milestones), the typed-int (`-F`, internal REST id) gotcha for sub-issue/dependency links, a `--dry-run` preview, and the documented escape hatch — ad-hoc batches skip this skill and hand `/implement-batch` a raw list directly.

GitHub-only, self-contained, grouping-only (never touches code). Pairs with #387: this skill produces the parent, that skill consumes it.

Doc-only change — one new `SKILL.md`, no runtime code touched.

## Test plan

- [x] N/A — markdown-only skill definition, no runtime surface. CI `verify` runs on the PR.

> Note: local pre-push `verify` flaked on `corpus-roundtrip.test.ts` (`google-docs-skia-proxy-achievements-oneline.pdf`) under full-suite parallelism — passes 3/3 in isolation (~2.5s). Known #360 flaky-under-load class, unrelated to this doc-only change. Pushed with `RESUMELINT_SKIP_HOOKS=1`; CI is the gate.

### DIFF
--- a/.claude/skills/batch-issues/SKILL.md
+++ b/.claude/skills/batch-issues/SKILL.md
@@ -34,8 +34,16 @@ off a ready invocation.
 ## Input
 
 Parse `$ARGUMENTS` for an **explicit comma/space list** of child issue numbers
-(e.g. `341,346,349` or `341 346 349`). If none is given, **stop and ask** which
-issues to batch (optionally list open, un-milestoned, coherent-looking candidates).
+(e.g. `341,346,349`, `341 346 349`, or `#341, #346`). If none is given, **stop and
+ask** which issues to batch (optionally list open, un-milestoned, coherent-looking
+candidates).
+
+Normalize the list before anything else — strip leading `#`, tolerate commas and/or
+whitespace, drop blanks:
+```bash
+CHILDREN=($(printf '%s' "$RAW_LIST" | tr ',' ' ' | tr -s ' ' | sed 's/#//g'))
+```
+`--order` is normalized the same way into `ORDER=(...)`.
 
 **Flags:**
 - `--title "..."` — the parent epic title. If omitted, propose one from the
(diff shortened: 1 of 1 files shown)


Vaishnavi1709 (dismissed):
Design is sound (cohesion gate, delegation to `/triage-issue`, typed-`-F` for sub-issue internal IDs — all correct). But the shell/gh commands in the Process section have several load-bearing bugs that will fire on real invocation. Flagging as request-changes; nothing here is architectural, just command-level fixes.

## Blocking

### 1. Step 3 never captures `PARENT` from `gh issue create`

The block ends with a comment `# capture the new parent number → PARENT` but no capture. `gh issue create` prints the new issue URL on stdout, not a bare number — Step 4's `repos/$REPO/issues/$PARENT/sub_issues` will 404 on every child.

Fix:
```bash
PARENT=$(gh issue create ... --json number --jq .number)
```

### 2. `--dry-run` is documented but never honoured

The Input section (line 53) says `--dry-run` writes nothing. The Process section (Steps 2–5) shows unconditional writes. Anyone passing `--dry-run` to preview a batch will actually create the label, mint the parent, POST every sub-issue link, and edit every child.

Fix: wrap each side-effect with `if [[ -z $DRY_RUN ]]; then <cmd>; else echo "[dry-run] would: <cmd>"; fi` (or an early-return block that prints the plan).

### 3. Milestone with spaces breaks the `gh` invocation

`${MILESTONE:+--milestone "$MILESTONE"}` inside a raw `gh` line word-splits on IFS after the parameter expansion — bash does NOT re-quote across `:+`. This repo's canonical milestones are literally `P1 · Friends & Family`, `P2 · Design Partner`, etc. (see CLAUDE.md).

Failure mode: `gh issue create ... --milestone P1 · Friends & Family` — gh sees `·` as an unknown positional. Same bug repeats in Step 5 for every child edit.

Fix: build a bash array.
```bash
args=()
[[ -n $MILESTONE ]] && args+=(--milestone "$MILESTONE")
[[ -n $ASSIGNEE  ]] && args+=(--assignee  "$ASSIGNEE")
gh issue create ... "${args[@]}"
```

### 4. `/implement-batch` doesn't exist yet

Grep of `.claude/skills/` confirms zero references outside this file. Step 7 prints `Implement it: /implement-batch <PARENT>` as the ready hand-off, but that skill is #387 (still open). Every batch created today emits a broken invocation.

Fix: land #387 first, or ship the two skills together in one PR.

### 5. Parent body renders child titles verbatim → markdown/HTML injection

The template `#341 — <title>` pastes the raw child title into the parent's markdown body. Titles containing backticks, `](evil.com)`, a leading `#`, or `` corrupt the epic body or render as HTML on the tracking issue.

Fix: escape titles (backtick-wrap, or `<code>#N</code> — <em>...</em>`).

## Consistent secondary pattern — Rules describe behaviour the commands don't implement

Three separate places the prose in Rules/Step 1 says X but the shown command does the opposite:

- **"Don't clobber a child that's already correctly milestoned"** (Rules, line 181): Step 5's `${MILESTONE:+--milestone …}` stamps the batch milestone on every child unconditionally, no per-child read/compare/warn.
- **"One person implements the batch"** (Step 2): Step 5's `--add-assignee` co-assigns — a child already assigned to A ends up assigned to both A and B.
- **Cohesion gate** (Step 1): "refuse closed, warn on mixed milestones" is prose, but there's no shown `for` loop or state accumulator that actually detects mixed milestones across children — `.milestone` in `gh issue view --json milestone` is a nested object, not a string, so a naive equality check gets it wrong anyway.

These are behavioral promises the skill won't keep. Either implement them explicitly, or downgrade the prose to "operator responsibility".

## Nits (not blocking)

- **Sub-issue POST isn't idempotent.** Re-running after a partial failure re-POSTs already-linked children → 422 aborts the loop. Preflight with `GET .../sub_issues --jq '.[].number'` and skip already-linked, or treat 422 as success.
- **No rollback on partial failure.** If Step 4 fails mid-loop, parent exists with half its children linked. Re-running mints a *second* parent (nothing keys on title/label for idempotency). Consider a "delete parent on hard fail" block or a resumable mode.
- **`--order` has no membership or cycle check.** `--order 341,346,999` wires 999 (not a child) as blocker; `--order 341,346,341` creates a self-cycle. Also, semantics only clarified in an inline comment ("B blocked by A"); the argument-hint calls it "dependency order" which a user could read as "execution order" and reverse.
- **Argument parser is prose-only.** No normalization for `#341,#346` (leading `#`), `341, 346, 349` (spaces after commas), or leading/trailing whitespace. Silently drops children.
- **`gh label create ... 2>/dev/null || true`** swallows every error, not just "already exists" — auth failures pass silently and Step 3's `--label` errors much later with a confusing message.
- **Step 3 stamps milestone but Step 6 delegates board placement to `/triage-issue`, which also owns milestone.** Layering violation — pick one owner (recommend `/triage-issue`) and drop `--milestone` from Step 3.
- **Verification readback** (Step 4) prints the linked list but never diffs against the input child set. Step 7's summary then echoes the input, so a silently-dropped child gets reported as linked. Diff the readback against the input set before printing ✓.

The skill is close — none of the above requires a rethink, just command-level fixes to Steps 3/4/5 and a real `--dry-run` branch.

s-annam:
Thanks — all command-level findings addressed in `2da8df6`. Point-by-point:

## Blocking
1. **PARENT never captured** — Fixed. Note: `gh issue create` has **no `--json`** flag; it prints the issue **URL** on stdout, so the skill now does `URL=$(gh issue create …); PARENT="${URL##*/}"` (basename = number).
2. **`--dry-run` never honoured** — Fixed as a **hard gate**: an explicit block at the top of Process — if `$DRY_RUN`, do read-only Step 1 + Step 2 planning, print the full plan, and **stop before Step 3**. No label/parent/sub-issue/child write happens.
3. **Milestone with spaces** — Fixed. Child edits (Step 5) build a bash **array** (`edit_args+=(--milestone "$MILESTONE")`), never `${VAR:+…}` inline. And the parent create (Step 3) now carries **no `--milestone` at all** — parent milestone is delegated to `/triage-issue` (Step 6), which also kills the layering nit below.
4. **`/implement-batch` doesn't exist yet** — Intentional: the executor ships in **#387** (separate PR). The skill now says so in the epic body, Step 7 hand-off (`⚠ not runnable until #387 merges`), and Rules. The two land together.
5. **Child-title injection** — Fixed. Titles are backtick-wrapped in the epic body (`#341 — \`<title>\``), neutralizing `](evil)`, ``, leading `#`.

## Secondary (prose vs commands)
- **Milestone clobber** — Step 5 now reads `.milestone.title` per child, sets only when empty, warns on conflict.
- **Assignee co-assign** — Step 5 reads `.assignees[0].login`, adds only when unowned, warns if already assigned (no double-owner).
- **Cohesion gate** — Step 1 is now a real loop: drops `CLOSED`, accumulates milestones in an associative array keyed on `.milestone.title` (the nested-object trap you flagged), warns when `>1`.

## Nits
- **Idempotent sub-issue POST** — Step 4 preflights the linked set and skips already-linked (re-run safe).
- **Verification diff** — Step 4 `comm -23` diffs readback vs input set; reports `✗ NOT linked` instead of assuming success.
- **`--order` validation** — Step 2 checks every id is a batch child and rejects repeats (self-cycle).
- **Arg normalization** — `#341, #346` / spaces / commas normalized up front (`tr … | sed 's/#//g'`).
- **`label create` swallow** — now only swallows "already exists"; any other error (auth/rate-limit) surfaces and exits.
- **Rollback/resume** — added a Rules note: Step 4 is idempotent on re-run, but a fresh invocation mints a second parent (nothing keys on title) — delete the half-made parent or continue from Step 4 with `$PARENT`.

Gates: `npm run typecheck` clean, `npm run test` green (1654 passed), pre-push `verify` green. Re-requesting review.

rohithgollapalli (dismissed):
**Approve.** Solid revision — the three bugs this skill's cautionary lineage is named for (`gh issue create --json`, milestone word-splitting, `--dry-run` writing) are all fixed, and I verified every `gh` flag against the live CLI. Two error-path hardening gaps and a nit, none merge-blocking.

## Secondary

**1. Parent-create has no success check — an empty `PARENT` cascades into malformed API paths.** Lines 158–163: `URL=$(gh issue create …)` then `PARENT="${URL##*/}"`. If the create fails (rate limit, transient, bad label), `URL` is empty, `PARENT=""`, and Step 4 immediately hits `gh api "repos/$REPO/issues//sub_issues"` — a confusing 404 loop rather than a clear failure. The sibling writes (label create line 116, sub-issue verify line 183) are guarded; this pivotal one isn't. Fix:
```bash
[[ "$PARENT" =~ ^[0-9]+$ ]] || { echo "✗ parent create failed:"; echo "$URL"; exit 1; }
```

**2. Step 5's `blocked_by` order edges aren't idempotent or verified.** Lines 214–217 POST each edge with no preflight-skip and no readback — unlike Step 4's sub-issue loop, which preflights `LINKED` and diffs the readback. On a resume/re-run past Step 4, re-POSTing an existing dependency 422s. The Resume note (lines 265–268) only credits Step 4 as idempotent, implicitly conceding this. Fix: mirror Step 4 — GET existing `dependencies/blocked_by`, `grep -qx` skip, then verify.

## Nits

- `--dry-run` is a "hard gate" that writes nothing, but the one write in the read/plan region — `gh label create` (lines 115–118) — sits inside **Step 2**, which the gate says runs ("do Step 2's *planning* only"). The gate's explicit "No label created" (line 80) saves it, but placing a write in the region the gate lets run invites the exact violation it guards. Safer to move label-create into Step 3 or guard it with `[[ -z $DRY_RUN ]] &&`.

s-annam:
Thanks @rohithgollapalli — all three error-path items addressed in `5db0dd5`.

**Secondary 1 — parent-create no success check.** Fixed. Step 3 now guards the capture: `[[ "$PARENT" =~ ^[0-9]+$ ]] || { echo "✗ parent create failed:"; echo "$URL"; exit 1; }`. A failed `gh issue create` now fails loud here instead of cascading into the `repos/$REPO/issues//sub_issues` 404 loop.

**Secondary 2 — order edges not idempotent/verified.** Fixed. Step 5's `blocked_by` loop now mirrors Step 4: preflight the child's existing `dependencies/blocked_by` set, `grep -qx` skip an already-present edge (no 422 on re-run), then read back and verify the edge landed before reporting success.

**Nit — `--dry-run` label-create sits in the Step 2 region the gate lets run.** Fixed. The `gh label create` is now guarded with `[[ -z $DRY_RUN ]] &&`, so the plan region is genuinely write-free rather than relying on the gate's downstream 'No label created' promise.

Gates: `npm run typecheck` clean, `npm run test` green (1654 passed), fallow clean on the changed file. Push dismissed the approval — re-requesting review.

Samhit21 (approved):
**APPROVE** — reviewed as code (Step 3e: a skill file's `gh`/bash blocks are the surface). Gates 3a–3d N/A (docs-only; no `src/`, fixtures, or components). The command-level bugs are well-covered by the prior reviews — I'm not repeating @Vaishnavi1709's set (all fixed in `2da8df6`) or @rohithgollapalli's two Secondary items, which still stand and are worth folding in (`:163` empty-`PARENT` success check; `:214` `blocked_by` edges not idempotent/verified like Step 4). Below are three net-new findings both reviews missed. All non-blocking → **0 Blocking → APPROVE** (verdict rule: 0 Blocking ⇒ approve).

## Blocking

None.

## Secondary

- **`SKILL.md:53` (and argument-hint `:4`) — `--milestone <name|number>` promises number support the `gh` commands don't have.** Both `gh issue edit --milestone` and `gh issue create --milestone` are **name-only** (`--milestone name`, "by name"). Passing `--milestone 3` makes gh look for a milestone literally *named* "3" — the child edit at `:199` (`--milestone "$MILESTONE"`) fails, and the clobber-compare at `:200` (`.milestone.title != "$MILESTONE"`) compares a title against a number so it always reads as a conflict. Either drop `|number` from the hint (`:4`, `:53`), or resolve number→title once up front:
  ```bash
  # if MILESTONE is numeric, resolve to its title before use
  [[ $MILESTONE =~ ^[0-9]+$ ]] && MILESTONE=$(gh api "repos/$REPO/milestones/$MILESTONE" --jq .title)
  ```

## Nits

- **`SKILL.md:171`,`:181` — the sub-issue readback isn't paginated.** `gh api .../sub_issues --jq '.[].number'` defaults to `per_page=30`. For a >30-child batch, the Step 4 verification `comm -23` (`:182`) reports children 31+ as `✗ NOT linked` even when they were, and the preflight (`:171`) re-POSTs them on a re-run (→ 422). Batches are "several small issues" by design so this is an edge, but it's the same pagination class the repo keeps hitting — add `--paginate` to both reads.
- **`SKILL.md:66` — `OWNER` is computed but never used.** `OWNER="${REPO%%/*}"` has no consumer anywhere in the skill. Drop it, or use it (the API paths all use `$REPO` directly).

---

**Verdict: APPROVE** (rule: 0 Blocking ⇒ APPROVE). Non-blocking items above + @rohithgollapalli's two open Secondary items are worth a fast-follow, but nothing here blocks merge.

Issue #293, opened 2026-07-03
Add corpus round-trip invariant test (parse→render→re-parse) as CI gate


s-annam:
## Summary

Add a **corpus-wide round-trip invariant test** as a CI gate: for every fixture PDF, assert that our own reconstructed-PDF output re-parses back to the same structured résumé. This is a self-consistency check — the renderer must emit shapes our own parser round-trips — and it's fully in our control on both ends (no dependency on quirky source PDFs).

Motivation: two real renderer regressions (education decomposition loss, summary truncation) were found by manually running a one-off round-trip diff. `render-roundtrip.repro.test.ts` already does this for one fixture (#284); this generalizes it to the whole corpus so future renderer changes can't silently regress.

## Design

New test iterating `tests/fixtures/pdfs/**/*.pdf`:

```
parse-1 = runCascade(fixture)
model   = buildAtsResumeModel(parse-1, score(parse-1))
bytes   = renderAtsResumePdf(model)
parse-3 = runCascade(bytes)
assert invariants(parse-1, parse-3)
```

Invariants (per fixture):
- contact fields identical (name/email/phone/location/linkedin)
- experience role count stable; each role's title/company/dates preserved
- education `degree`/`field`/`institution` preserved (currently failing — see companion bug)
- `|Δ summary length| < 5%` (currently failing — see companion bug)
- skills count stable
- triggers stable

Runs in the existing Node/vitest harness (~0.8s per fixture), PII-free (goldens stay lossy; this asserts field mapping directly, not values in a snapshot).

## Scope note — drop the PDF-to-PDF step

The original 5-step round-trip idea included diffing the rendered PDF bytes (step 2) against a re-rendered PDF (step 4). `renderAtsResumePdf` is deterministic, so those bytes are identical — that diff only ever proves render determinism, never parse quality. **All signal lives in the parse-1 vs parse-3 diff.** This gate implements only that; the PDF-to-PDF step is intentionally omitted.

## Also ship
- Keep the dev harness (`RL_RT_PDF=<path>` env → JSON parse-1/parse-3 diff) documented for one-off triage of real (uncommitted, PII-bearing) résumés, separate from the committed CI gate.

## Acceptance criteria
- [ ] New round-trip invariant test covers all corpus fixtures.
- [ ] Wired into `npm run test` / CI.
- [ ] Passes once the companion education + summary bugs are fixed (may land red-then-green, or gate only the currently-passing invariants first).

PR #395, opened 2026-07-07
fix(parser): anchor bare-year org sub-line so single-column year-only roles round-trip (#358)

### LINKED ISSUE #358: Round-trip: single-column experience roles swap title/company + drop start_date (gap left by #298)
## Symptom

Running the Download-PDF round-trip (`parse → renderAtsResumePdf → re-parse`) on **single-column** résumés, experience role headers do not round-trip:

- **A — title/company swap.** After re-parse, a role's `title` and `company` are transposed relative to parse-1. Reproduced on **2 distinct single-column résumés** (offending roles at experience index 3 and index 2 respectively).
- **B — start_date dropped.** On one of those two résumés, the same corrupted role also loses its `start_date` (present in parse-1 → `undefined` in parse-2) on the same hop.

Both symptoms occur on **hop 1** (a single export/re-parse). No `renderError` — the render succeeds; the corruption is in what the reconstructed PDF re-parses to.

> Found via the `/probe-roundtrip` dev harness over **real, PII-bearing résumés** (local-only, never committed). Per the probe's PII guardrail this issue cites corruption **by category only** with a **synthetic** repro — no candidate values appear here. Full diffs live in gitignored `internal/roundtrip/`.

## Why this is not already covered

- **#298 (CLOSED)** — "experience header re-segmentation (title/company swap)" — fixed this class **only for the dense / two-column fixtures it listed** (`awesome-cv-cv`, `chromium-two-column-sidebar`, `two-column-achievements-sidebar`, etc.). All of its affected fixtures are two-column or dense multi-role layouts. **Single-column layouts were never in scope** — and the swap still reproduces there. This issue is the remaining single-column tail of #298.
- **#346 (OPEN)** — two-column title/company inversion — is a **parse-time** bug (parse-1 itself is wrong). This is different: parse-1 is correct, the **round-trip** corrupts it.

## How to reproduce

```bash
RL_RT_PDF=/abs/path/to/single-column-resume.pdf \
  npx vitest run src/lib/heuristics/corpus-roundtrip.test.ts
```

The console prints, for the offending hop, an `experience` category diff of the form (synthetic, illustrative — not real values):

```
experience:
  role[i].title:      "Composer"            → "Northwind Ensemble"
  role[i].company:    "Northwind Ensemble"  → "Composer"
  role[i].start_date: "2022"                → undefined
```

i.e. a single-column role whose source stacks **company line then title line** (with a right-aligned date) re-parses with the two fields transposed after the reconstructed PDF is read back.

## Root cause (hypothesis)

Same mechanism #298 identified, un-fixed for single-column input: `buildAtsResumeModel` (`src/lib/pdf/ats-resume-model.ts`) + `renderAtsResumePdf` (`src/lib/pdf/render-ats-pdf.ts`) emit the experience header (`title` / `company` / dates) in a line order or whitespace shape that `openresume.ts` / `regex.ts` header-anchor detection re-segments differently than parse-1 for the single-column case — e.g. company-first vs role-first ambiguity, or a right-aligned date landing where a header anchor expects the company. The dropped `start_date` (symptom B) is likely the same header line being mis-split so the date token is not attached to the role.

Files to read:
- `src/lib/pdf/ats-resume-model.ts` — how the round-trip model serializes each experience role header
- `src/lib/pdf/render-ats-pdf.ts` — the actual PDF line emission for the experience block
- `src/lib/heuristics/openresume.ts` + `src/lib/heuristics/regex.ts` — the re-parser's experience header/anchor detection that reads the reconstructed lines back

## Fix direction

- Mint a **synthetic, single-column** experience fixture (synthetic persona per `tests/fixtures/pdfs/README.md`) that reproduces the swap: at least one role whose header stacks company then title with a right-aligned date, plus one role that also exercises the dropped-`start_date` shape.
- Harness-trace it (parse-1 entries vs reconstructed lines vs parse-2), then make the reconstructed single-column experience header unambiguous to our own re-parser — mirror the #291/#298 approach (title on the bold header line, `Company · Location  Dates` sub-line).
- If the header-shape fix also re-attaches the date, symptom B closes with symptom A.

## Acceptance criteria

- [ ] A synthetic single-column experience fixture reproducing the title/company swap on round-trip is added under `tests/fixtures/pdfs/` (synthetic persona; PII-checked per fixture policy).
- [ ] `role.title`, `role.company`, and `role.start_date` round-trip (parse-1 == parse-2) for that fixture through the `#293` corpus round-trip gate.
- [ ] The gate covers the single-column shape (not only the two-column shapes #298 added); `experience` is not listed in `KNOWN_FAILURES` for the new fixture.
- [ ] No regression: existing corpus round-trip invariants (#293) stay green; the two-column fixtures fixed by #298 still round-trip.
- [ ] Re-running `/probe-roundtrip` on a single-column résumé of the reproducing shape reports `clean` on the experience category.

---
_Found by the `/probe-roundtrip` audit batch over local real résumés (9 of 11 clean; this is the sole distinct bug class surfaced). Follow-up to closed #298; part of the round-trip audit series #295–#299._

rohithgollapalli:
## Summary

Fixes **#358** — the single-column tail of #298. On the Download-PDF round-trip, an experience role whose **only date is a bare year** re-parsed with `title`/`company` **swapped** and `start_date` **dropped**.

## Root cause (one missing anchor drives both symptoms)

A year-only role reconstructs to an org sub-line like `Northwind Ensemble · Boston, MA  2022`. `DATE_RANGE_RE` doesn't match a lone year (it needs a range/month-year), so **no `date_range` anchor forms** on re-parse. That single fact causes both:

- **Symptom A (swap):** with no anchor on the org line, `anchorHeaderIndex` doesn't point at it → the #298 `anchorHasReconstructedSignature` tiebreak (which maps the `·`-signed org line as company, title above) can't fire → falls to the "top line = company" default → transpose.
- **Symptom B (date drop):** `parseDateRange`'s bare-year fallback only runs on an anchored line, so the year never re-attaches.

Confirmed by isolating the render→reparse: `Composer / Northwind Ensemble / 2022` came back `title="Northwind Ensemble", company="Composer", start_date=undefined` — exactly the issue's illustrative diff. A control role with a **full** range did not swap.

## Fix

`isAnchorLine` (date_range) admits a header-shaped line whose **trailing token is a bare year** AND which carries the renderer's `" · "` org separator, as a `date_range` anchor. Restoring the anchor lets the #298 tiebreak fire (no swap) and `parseDateRange` re-attach the year — **closing A and B together**, per the issue's "if the header-shape fix re-attaches the date, B closes with A."

**The middot guard is load-bearing.** A broader `endsWithBareYear + isEntryHeaderShape` guard regressed 3 tests — it promoted a plain role TITLE ending in a year (openresume's `"Software Engineer Intern Summer 2022"`) and a #166 wrapped-header fragment to phantom anchors. Scoping admission to the reconstructed org sub-line shape (the middot) fixes #358 with **zero corpus movement**.

## Tests

- **New synthetic single-column fixture** `single-column-year-only-roundtrip.pdf` (persona `Jordan Avery / jordan.avery@example.com / (614) 555-0137`, `pdftotext`-verified). Round-trips clean through the #293 gate; `experience` is **not** in `KNOWN_FAILURES`. **Red without the fix** — corpus snapshot + round-trip both fail (`'experience' regressed`).
- **`render-roundtrip-year-only.repro.test.ts`** pins the exact field values (title/company not swapped, `start_date` re-attached) that the lossy corpus snapshot can't capture.
- Full suite **1658 passed / 0 failed**; `tsc -b --noEmit` + `eslint .` clean; `vite build` OK.

## Out of scope (noted as a sibling)

A **full-range** no-lexical-tell role can still leak its location into `company` on re-parse (seen on a control role). Different symptom, different root — not #358's year-only swap; worth its own follow-up.

Closes #358


s-annam (approved):
Reviewed against the resumelint gates on top of a correctness pass. Verdict rule: 0 Blocking → **approve**.

Clean, well-scoped fix. The anchor admission is tightly guarded — bare trailing year **and** the `·` org separator **and** `isEntryHeaderShape` all required — so a plain title ending in a year or a `·`-led fragment can't become a phantom anchor. Downstream `parseDateRange`'s bare-year fallback re-attaches the year and the #298 org-signature tiebreak kills the swap, closing both symptoms at once.

Gates run:
- **Fixture PII** ✅ — `Jordan Avery` / `@example.com` / `(614) 555-0137` (real area code + 555 exchange + 0137 subscriber). PDF metadata clean (HeadlessChrome/Skia, no author). `unknown/` category is provenance-neutral.
- **fallow** ✅ — no issues in changed files; the CRAP-flagged functions are pre-existing and untouched.
- **Tests** ✅ — new repro (2/2), corpus (43/43, incl. the new fixture snapshot), entry-blocks + extract-fields (113/113), lint clean. Only a *new* golden snapshot is added — no existing snapshot regenerated, so no regression-masking.
- Design-system / style / command-level gates: N/A (no component or script/skill files).

Nice touch asserting exact field values in the repro test where the lossy corpus snapshot can't.

PR #397, opened 2026-07-08
feat(result,edit): two-column warning + edit validation (#396, #398)

### LINKED ISSUE #356: Warn users about two-column layouts on the Result page (detected but never surfaced)
## Summary

When the parser detects a **two-column layout** (`isTwoColumn` layout probe), surface a clear, user-facing warning that two-column résumé layouts are a common cause of ATS mis-parsing, and coach the user to switch to a single-column layout. Today two-column layouts silently apply a score penalty and produce degraded extraction, but the product never tells the user *why* their score is low or *what to do about it*.

## Motivation (from 2026-07-06 Intern Weekly)

> "Two-column layouts in general are problematic for parsers; people don't realize it. They simply make it look pretty, and then they send it off... we need to message, like, more strongly — hey, stop doing two-column layouts."

Two-column is one of the highest-signal, most-actionable pieces of advice this tool can give: users pick two columns for aesthetics, have no idea real ATS parsers choke on them, and the whole product thesis is *"see what a plain text extractor actually reads back."* We already **detect** the condition and **penalize** it in scoring — we're just not **telling the user** the one thing that would most improve their outcome.

## Current behavior

- `src/lib/heuristics/pdf-layout.ts` emits `LayoutProbes { isTwoColumn, ... }`; the `two_column` trigger flows into `CascadeResult.triggers`.
- Scoring (`src/lib/score/score.ts`) multiplies the overall score by a layout-trigger penalty (1.0 / 0.85 / 0.70).
- `src/components/Result.tsx` reads `result.triggers` and, for `fonts_unmappable`, routes to `LimitedParsingCard` with an explanatory blurb (`LAYOUT_TRIGGER_BLURBS`).
- **Gap:** the two-column trigger has no equivalent user-facing explanation + call-to-action. The penalty lands but the "why / fix" never surfaces.

## Proposed change

Add a two-column-specific warning surface on the Result page, reusing the existing trigger-blurb pattern:

1. Add a two-column entry to the trigger-blurb mechanism (`LAYOUT_TRIGGER_BLURBS` or its sibling) in `src/components/Result.tsx` (or wherever blurbs are centralized) explaining: two-column layouts are read column-by-column-merged or out of order by many ATS parsers; the reconstructed text below shows what actually got extracted; **recommend a single-column layout**.
2. Render it when `result.triggers.includes("two_column")` — inline near the score readout / reconstructed resume, not as a full-page takeover (unlike the `fonts_unmappable` `LimitedParsingCard`, two-column output is still usable).
3. Use the existing shared warning/banner component + semantic tokens (`text-brand-amber`, `bg-surface-card`, etc.) — no hand-rolled banner, no hardcoded colors (CLAUDE.md style gate).

### Reuse analysis

- **Capability:** display a layout-condition warning on the Result page (read-only presentation — not a mutation surface, but noted for completeness).
- **Existing surfaces found:** `src/components/Result.tsx` already owns trigger→blurb messaging (`LAYOUT_TRIGGER_BLURBS`, `LimitedParsingCard`); shared banner/warning primitives exist under `src/design-system/`.
- **Decision:** **extend** the existing trigger-blurb surface in `Result.tsx` with a two-column entry + a lighter-weight inline banner (two-column output is still usable, so no full-page card). No new parallel surface.

## Acceptance criteria

- [ ] When `result.triggers` includes `two_column`, the Result page shows a warning explaining two-column layouts hurt ATS parsing and recommending a single-column layout.
- [ ] The warning is inline (does not replace the whole result the way `LimitedParsingCard` does) since two-column output is still usable.
- [ ] Copy points the user at the reconstructed text below as evidence of what actually got extracted.
- [ ] Uses shared banner/warning component + semantic tokens; no hardcoded palette classes or raw `<button>` (passes `npm run lint`).
- [ ] No regression: résumés without the `two_column` trigger show no such warning.

### LINKED ISSUE #357: Edit fields accept any string — add validation hook to EditableField, wire date validator
## Summary

The inline-edit fields on the reconstructed résumé accept **any string with no shape validation** — most visibly the role **date** fields, which will happily commit arbitrary garbage (e.g. `banana`) as a start/end date, but the same holds for the contact LinkedIn/URL, phone, and email fields. Surfaced during 2026-07-06 → 07 dogfooding across ~5 résumés ("date would take any random string... that kind of bug"). The shared `EditableField` primitive has no validation hook, so no editable field can flag bad input.

## Current behavior

- `src/design-system/primitives/EditableField.tsx` is the single shared inline-edit primitive. Its props are `value / placeholder / label / onCommit / displayValue / className / textWeight / ...` — **there is no `validate` prop.** `onCommit` fires with whatever trimmed string the user typed.
- Role dates edit through this primitive in `src/components/features/ReconstructedRole.tsx` (`start_date`, `end_date`). A user can type `banana` into a date field and it commits as the date, shown as-is in read mode.
- Same unvalidated path applies to contact email/phone/LinkedIn (`ContactCard.tsx` / `ContactDetails.tsx`) and the `+ Add` new-entry fields (`ReconstructedEducationSkills.tsx`), each accepting any string including blanks.

## Positioning constraint (load-bearing)

This respects ResumeLint's **"parser audit, not a judge"** stance — the user is authoritative over their own résumé. Future dates, unusual company names, and odd-but-real values must **NOT** be blocked. So the validator catches **shape typos** (a bare word where a date/URL belongs), not semantic plausibility.

## Proposed change

1. Add an **optional** `validate?: (value: string) => string | null` prop to `EditableField` — additive, so existing callers are byte-for-byte unchanged when omitted. Returns `null` → clean; returns a string → shape-fail message.
2. **Non-blocking UX (decided):** on shape-fail, **still commit** the value and surface a **soft warning icon** in read mode (semantic tokens only) carrying the validator message on hover/label — do **not** hold the field open or reject. The user sees the typo without being blocked.
3. Wire per-field validators at the call sites:
   - **Dates** (`start_date` / `end_date` in `ReconstructedRole.tsx`) — accept the résumé date forms the parser already understands (`Mon YYYY`, `YYYY`, `YYYY – YYYY`, `Present`, ranges). **Reuse existing date-recognition logic in `src/lib/heuristics/`** rather than a fresh regex where possible.
   - **LinkedIn / URL** (`ContactCard.tsx` / `ContactDetails.tsx`) — URL-ish / `linkedin.com/in/` shape (accept `example.com/in/` too for synthetic fixtures).
   - **Phone** — `libphonenumber-js` `isValid` (already a dep).
   - **Email** — RFC-ish shape.

## Acceptance criteria

- [ ] `EditableField` gains an optional `validate` hook; omitting it leaves all existing callers unchanged (byte-for-byte).
- [ ] Shape-fail is **non-blocking**: the value commits and a soft warning icon appears in read mode — the field never traps the user or rejects the commit.
- [ ] Role start/end date fields flag obvious typos (e.g. `banana`) while valid résumé date forms (`Present`, `YYYY`, `Mon YYYY`, ranges) commit clean with no warning.
- [ ] Contact LinkedIn/URL, phone, and email fields each flag shape typos and pass genuine values (incl. synthetic `@example.com` / `example.com/in/` fixtures).
- [ ] Date validation reuses existing recognition logic in `src/lib/heuristics/` where possible rather than duplicating a parser.
- [ ] Warning icon uses semantic tokens / shared primitives — no hardcoded palette, passes `npm run lint`.
- [ ] Unit tests cover accept + flag cases for each validator (date, URL, phone, email).

## Repro

```
npm run dev
# drop any fixture; click any date field; type "banana"; press Enter
# → commits as-is with no warning (current); should show soft warning icon (target)
```

## Notes

- Milestone: **P2 / second phase** (per the meeting: "put them into P2 for now... clean it up later"), not a Friends & Family launch blocker.
- Surfaced by Vaishnavi's dogfooding pass; filed at Sri's explicit direction (overriding the usual "finder files their own" norm for this batch).
- Consolidates duplicate #377 (`[edit-interaction] inline edits accept any string`), which contributed the full field set and the non-blocking soft-warning UX.

### LINKED ISSUE #396: Batch: Result-page guardrails — two-column warning + edit validation
## Batch epic — Result-page guardrails

Groups the small issues below for **single-batch implementation** by one person
via `/implement-batch` (#387). This is a tracking issue; the work lives in the children.

**Why grouped:** both add a user-facing safety nudge on the Result/edit surface — surface a
signal the parser already computes but never shows, and validate what the user types back in.

### Children
- [ ] #356 — `Warn users about two-column layouts on the Result page (detected but never surfaced)`
- [ ] #357 — `Edit fields accept any string — add validation hook to EditableField, wire date validator`

**Implement with:** `/implement-batch <THIS#>`

### LINKED ISSUE #398: test: useAnalyzedResume + useDownloadPdf suites fail locally (13) — localStorage global shadows jsdom, trips pre-push verify
## Summary

Two hook test suites fail locally on a clean `main` (13 tests), tripping the local pre-push `verify` hook and the Claude Stop hook:

- `src/hooks/useAnalyzedResume.test.tsx` (9 failures)
- `src/hooks/useDownloadPdf.test.tsx` (4 failures)

They are **unrelated to any current PR** — confirmed by stashing all working changes on a clean `main` checkout: still 13 failed. Neither suite imports recently-changed files. Surfaced while landing batch #396 (PR #397), which had to push with `RESUMELINT_SKIP_HOOKS=1` to get past the red local gate.

## Symptoms

```
FAIL src/hooks/useDownloadPdf.test.tsx
  TypeError: localStorage.clear is not a function
   ❯ src/hooks/useDownloadPdf.test.tsx:88  (beforeEach)

FAIL src/hooks/useAnalyzedResume.test.tsx
  TypeError: Cannot read properties of undefined (reading 'unmount')
   ❯ src/hooks/useAnalyzedResume.test.tsx:46  act(() => root.unmount())
```

## Root-cause investigation

- Both files declare `// @vitest-environment jsdom` and call `localStorage.clear()` in `beforeEach`.
- **Sibling jsdom-pragma hook tests pass** (`useReplaceResumeOnDrop.test.tsx` 9/9 green, same `createRoot`/`act` probe pattern) — so the jsdom environment is applied globally; this is **not** a blanket env-config break.
- The differentiator: these two are the **only** jsdom hook suites that touch `localStorage`. `localStorage.clear is not a function` means a `localStorage` global exists but is not a real `Storage`.
- Cascade: `localStorage.clear()` throws in `beforeEach` → the module-scoped `root` is never assigned → `afterEach`/`unmount()` then dereferences `undefined.unmount`, producing the second error class in `useAnalyzedResume`.

**Leading hypothesis:** a non-jsdom `localStorage` global (e.g. Node's experimental Web Storage global) is shadowing jsdom's `window.localStorage`, so `.clear`/`.setItem` are absent. Needs confirmation of the local Node version vs CI.

## Open question — is CI red too?

`main` is the deployed branch, so CI `verify` presumably passes — meaning this is **local-environment-specific** (Node version / experimental flag), not main-is-broken. Worth confirming: if CI is also skipping or these were merged red (they were introduced by #353, `feat(epic): … blank-resume authoring`), the gap is real.

## Fix directions

1. Make the two suites provision their own `Storage` (jsdom `window.localStorage`, or a small `beforeEach` polyfill / `vi.stubGlobal('localStorage', ...)`), instead of assuming a working global.
2. Or pin the vitest jsdom environment so the Node experimental `localStorage` global cannot shadow `window.localStorage`.
3. Confirm the Node version the local hook runs under and whether CI matches.

## Repro

```
git switch main && git pull
npx vitest run src/hooks/useAnalyzedResume.test.tsx src/hooks/useDownloadPdf.test.tsx
# → 13 failed
```

Introduced by: #353 (56a20d8).

s-annam:
## Summary

Batch epic #396 — Result-page guardrails. Two user-facing safety nudges on the Result/edit surface: surface a signal the parser already computes but never shows, and validate what the user types back in.

- **#356 — two-column warning.** When the parser emits the `two_column` trigger, the Result page now shows an inline warning (shared `ErrorState` `tone="warning"`, semantic tokens) explaining two-column layouts are a common ATS mis-parse cause, pointing at the reconstructed text below as evidence, and recommending a single-column layout. No warning when the trigger is absent.
- **#357 — EditableField validation.** Adds an optional additive `validate?: (value) => string | null` hook to the `EditableField` primitive with **non-blocking** UX — the value always commits and a soft warning glyph appears in read mode (semantic `text-feedback-warning-icon`); the field never traps or rejects. Wired date/URL/phone/email shape validators at the call sites. Date validator **reuses** the heuristics grammar (`DATE_ANCHOR`/`DATE_RANGE_RE`), phone reuses `normalizePhone`, email reuses `EMAIL_RE`. Catches shape typos (`banana`) only — `Present`/`YYYY`/`Mon YYYY`/ranges/`20XX`/future years pass ("parser audit, not a judge").

Closes #356
Closes #357
Closes #398
Closes #396

## Test plan

- [x] `npm run lint` clean
- [x] Affected suites green (field-validators 15/15, Result/ReconstructedRole/ContactCard/EducationSkills 73/73)
- [x] Whole-tree `fallow audit` — no dead-code/dupe issues in changed files (complexity advisories only)
- [x] Full `npm run test` / `verify` in CI (green — run 28912727231, includes the #398 localStorage-shim fix; 1673 passed)

> **Folded in (#398):** `useAnalyzedResume.test.tsx` + `useDownloadPdf.test.tsx` (13 jsdom-harness failures on Node 22+, where a broken native global `localStorage` shadows jsdom's) are now fixed here via a shared in-memory `localStorage` shim. Pre-existing on clean `main`; unrelated to the guardrails diff but bundled so the branch goes green end-to-end.

## Adversarial review

Ran a bounded adversarial pass (`ecc:react-reviewer`, opus) on the accumulated diff **before** opening this PR. Converged in **1 round — clean, zero blocking findings.** Reviewer reproduced end-to-end (not just reasoned): date anchoring is fully `^…$`-anchored (`banana 2020` / `2020 banana` / `2020 2021` all flag; `Present`/`Mon YYYY`/`20XX`/ranges/future years pass), `DATE_ANCHOR` reuse is genuinely consumed (not dead), `#356` gate matches the real `two_column` `LayoutTrigger`, and the `validate` prop is byte-for-byte additive + non-blocking.

**Surviving nits (non-blocking — human reviewer may eyeball):**
1. **a11y** — `EditableField.tsx` read-mode wrapper has `aria-label="Edit {label}"`, which shadows the warning glyph's `title`/`aria-label` (ARIA name-from-author), so a screen-reader user hears the field name but not the shape warning. Hover tooltip works; SR path does not. Fix: fold the message into the field's `aria-label`/`aria-describedby`.
2. **consistency** — the achievement `label="Year"` `EditableField` in `ReconstructedResume.tsx` is a date-shaped sibling left unwired (`validateDate` accepts bare `YYYY`/`20XX` and would fit). Not in the acceptance list.
3. **copy** — `TWO_COLUMN_BLURB` "column-by-column-merged or out of order" reads as a run-on.

### DIFF
--- a/src/components/Result.tsx
+++ b/src/components/Result.tsx
@@ -25,7 +25,7 @@ const FONTS_UNMAPPABLE_BLURB =
 // fonts_unmappable, two-column output is still usable, so this renders as a
 // warning banner alongside the score rather than replacing the whole card.
 const TWO_COLUMN_BLURB =
-  "This resume uses a two-column layout. Many ATS parsers read text column-by-column-merged or out of order, scrambling the result. The reconstructed text below shows what a generic parser actually extracted — for the most reliable ATS parsing, use a single-column layout.";
+  "This resume uses a two-column layout. Text extractors often read the columns out of order — merging or interleaving them. The reconstructed text below is what a generic parser actually pulled out; if it looks scrambled, that's the ATS risk. A single-column layout parses most reliably.";
 
 type SourceKind = "pdf" | "docx";
 
--- a/src/components/Result.tsx
+++ b/src/components/Result.tsx
@@ -5,7 +5,7 @@ import { useCallback, useMemo, useState } from "react";
 import type { CascadeResult } from "../lib/heuristics/types.ts";
 import { computeAnonymousAtsScore, type AnonymousAtsScore } from "../lib/score/score.ts";
 import type { EditableParse } from "../hooks/useEditableParse.ts";
-import { Card, StatusBadge, Button } from "@design-system";
+import { Card, StatusBadge, Button, ErrorState } from "@design-system";
 import { FeedbackPanel } from "./features/FeedbackPanel.tsx";
 import { AtsScoreReadout } from "./features/AtsScoreReadout.tsx";
 import { isScoreRevealed } from "../lib/contact.ts";
@@ -20,6 +20,12 @@ import { ResultDetailTabs } from "./features/ResultDetailTabs.tsx";
 // LAYOUT_TRIGGER_BLURBS for fonts_unmappable is still needed by LimitedParsingCard.
 const FONTS_UNMAPPABLE_BLURB =
   "Text is present in the source but uses custom font encodings that don't decode to characters. Common with Framer, Affinity, and some InDesign exports.";
+
+// Two-column layout warning (#356) — inline, non-blocking. Unlike
+// fonts_unmappable, two-column output is still usable, so this renders as a
+// warning banner alongside the score rather than replacing the whole card.
+const TWO_COLUMN_BLURB =
+  "This resume uses a two-column layout. Many ATS parsers read text column-by-column-merged or out of order, scrambling the result. The reconstructed text below shows what a generic parser actually extracted — for the most reliable ATS parsing, use a single-column layout.";
 
 type SourceKind = "pdf" | "docx";
 
@@ -143,6 +149,11 @@ function ParsedCard({
   const scoreRevealed =
     !isBlankAuthored || isScoreRevealed(activeResult, edit.contactOverrides);
 
+  // Two-column layout warning (#356) — detected but previously never
+  // surfaced to the user. Inline, not a full-page takeover: two-column
+  // output is still usable, unlike the fonts_unmappable case above.
+  const isTwoColumn = result.triggers.includes("two_column");
+
   return (
     // Two stacked surfaces: the score "summary" card on top, the tabbed detail
     // card below. The gap + each card's own border draws the separator the
@@ -166,6 +177,10 @@ function ParsedCard({
           onResetAll={edit.resetAll}
           onReset={onReset}
         />
+
+        {isTwoColumn && (
+          <ErrorState tone="warning">{TWO_COLUMN_BLURB}</ErrorState>
+        )}
 
         {scoreRevealed ? (
           <AtsScoreReadout score={activeScore} />
--- a/src/components/features/ContactDetails.tsx
+++ b/src/components/features/ContactDetails.tsx
@@ -20,6 +20,12 @@
 
 import { formatLinkDisplay, type ContactDisplayField } from "../../lib/contact.ts";
 import { EditableField } from "@design-system";
+import {
+  validateEmail,
+  validatePhone,
+  validateUrl,
+  type FieldValidator,
+} from "../../lib/edit/field-validators.ts";
 import type { ContactOverrides } from "../../hooks/useEditableParse.ts";
 
 /** The inline-editable contact fields, mapped 1:1 to their `ContactOverrides`
@@ -38,6 +44,18 @@ const EDITABLE_KEYS: Record<string, keyof ContactOverrides> = {
 };
 
 type Commit = (key: keyof ContactOverrides, v: string) => void;
+
+/** Shape validator per editable contact field. Name/location are free-form
+ *  (a "parser audit, not a judge" — any string is a legitimate name), so they
+ *  map to no validator; email/phone/link fields each get a shape check. */
+const FIELD_VALIDATORS: Partial<Record<string, FieldValidator>> = {
+  email: validateEmail,
+  phone: validatePhone,
+  linkedin_url: validateUrl,
+  github_url: validateUrl,
+  portfolio_url: validateUrl,
+  website_url: validateUrl,
+};
 
 interface ContactDetailsProps {
   contactLine: ContactDisplayField[];
@@ -132,6 +150,7 @@ function EditableValue({
       placeholder={`${field.label} not detected`}
       label={field.label}
       textSize="sm"
+      validate={FIELD_VALIDATORS[field.key]}
       onCommit={(v) => commit(ovKey, v)}
     />
   );
--- a/src/components/features/ContactDetails.tsx
+++ b/src/components/features/ContactDetails.tsx
@@ -47,15 +47,27 @@ type Commit = (key: keyof ContactOverrides, v: string) => void;
 
 /** Shape validator per editable contact field. Name/location are free-form
  *  (a "parser audit, not a judge" — any string is a legitimate name), so they
- *  map to no validator; email/phone/link fields each get a shape check. */
+ *  map to no validator; email/link fields each get a shape check. Phone is
+ *  resolved separately (see `validatorFor`) because it needs the parsed
+ *  location threaded in for its region default. */
 const FIELD_VALIDATORS: Partial<Record<string, FieldValidator>> = {
   email: validateEmail,
-  phone: validatePhone,
   linkedin_url: validateUrl,
   github_url: validateUrl,
   portfolio_url: validateUrl,
   website_url: validateUrl,
 };
+
+/** Resolve the validator for a field. Phone binds the résumé's parsed location
+ *  so non-US local-form numbers aren't falsely flagged (mirrors the parser's
+ *  `extractContact`, which wires `regionFromLocation` for the same reason). */
+function validatorFor(
+  key: string,
+  location: string | undefined,
+): FieldValidator | undefined {
+  if (key === "phone") return (v) => validatePhone(v, location);
+  return FIELD_VALIDATORS[key];
+}
 
 interface ContactDetailsProps {
   contactLine: ContactDisplayField[];
@@ -70,6 +82,9 @@ export function ContactDetails({
   editable,
   commit,
 }: ContactDetailsProps) {
+  // The parsed location, threaded into the phone validator's region default so a
+  // non-US local-form number isn't falsely flagged (see `validatorFor`).
+  const location = contactLine.find((f) => f.key === "location")?.value;
   return (
     <>
       {/* Contact line: location / email / phone, pipe-joined, present-only. */}
@@ -78,7 +93,7 @@ export function ContactDetails({
           {contactLine.map((field, i) => (
             <span key={field.key} className="inline-flex items-center gap-x-2">
               {i > 0 && <span className="text-content-muted">|</span>}
-              {renderContactValue(field, editable, commit)}
+              {renderContactValue(field, editable, commit, location)}
             </span>
           ))}
         </p>
@@ -134,11 +149,13 @@ function EditableValue({
   ovKey,
   commit,
   displayValue,
+  location,
 }: {
   field: ContactDisplayField;
   ovKey: keyof ContactOverrides;
   commit: Commit;
   displayValue?: string;
+  location?: string;
 }) {
   const absent = field.gated && field.reason === "absent";
   const lowConfidence = field.gated && field.reason === "low_confidence";
@@ -150,7 +167,7 @@ function EditableValue({
       placeholder={`${field.label} not detected`}
       label={field.label}
       textSize="sm"
-      validate={FIELD_VALIDATORS[field.key]}
+      validate={validatorFor(field.key, location)}
       onCommit={(v) => commit(ovKey, v)}
     />
   );
@@ -183,10 +200,13 @@ function renderContactValue(
   field: ContactDisplayField,
   editable: boolean,
   commit: Commit,
+  location: string | undefined,
 ) {
   const ovKey = EDITABLE_KEYS[field.key];
   if (editable && ovKey !== undefined) {
-    return <EditableValue field={field} ovKey={ovKey} commit={commit} />;
+    return (
+      <EditableValue field={field} ovKey={ovKey} commit={commit} location={location} />
+    );
   }
   return field.gated && field.reason === "absent" ? (
     <MissingToken label={field.label} />
--- a/src/components/features/FeedbackPanel.test.tsx
+++ b/src/components/features/FeedbackPanel.test.tsx
@@ -22,7 +22,7 @@
  * `useModelSelection.integration.test.tsx`.
  */
 
-import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import { createElement } from "react";
 import { createRoot, type Root } from "react-dom/client";
 import { act } from "react";
@@ -53,31 +53,13 @@ async function mountPanel(opts: {
   return container;
 }
 
-// Persisted render-state keys FeedbackPanel writes (#193/#194). The CI
-// localStorage shim has no .clear(), so remove them individually; wrap in
-// try/catch to match the component's own fail-silent storage access.
-function resetFeedbackStorage() {
-  for (const k of [
-    "rl_feedback_seen",
-    "rl_feedback_submitted",
-    "rl_star_cta_seen",
-    "rl_gh_stars_cache",
-  ]) {
-    try {
-      localStorage.removeItem(k);
-    } catch {
-      /* shim/private-mode — ignore */
-    }
-  }
-}
-
-// Start each test from clean persisted state — CI shares localStorage across
-// test files too, so another file's FeedbackPanel mount could otherwise leak
-// rl_feedback_submitted / rl_feedback_seen into this file's first test.
-beforeEach(() => {
-  resetFeedbackStorage();
-});
-
+// FeedbackPanel persists render-state to localStorage (rl_feedback_seen /
+// rl_feedback_submitted / rl_star_cta_seen / rl_gh_stars_cache, #193/#194).
+// Without a clean store, the submit test flips later mounts into "done"
+// (renders null) and the seen counter accumulates into "compact" mode — both
+// hide the form. The global setup (src/test-setup.ts, #398) installs a fresh
+// in-memory localStorage before every test, so every test starts in the "full"
+// state without any per-file clearing bookkeeping.
 afterEach(() => {
   if (root) act(() => root!.unmount());
   container?.remove();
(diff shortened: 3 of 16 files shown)


s-annam:
Two follow-ups folded into this PR:

**Test fixes (`bbfb442`)** — the jsdom hook/component suites needed an in-memory `localStorage` shim; the `rl_*` functional keys were touching a `localStorage` that jsdom doesn't provision by default. Shim added so those suites run green.

**Truthful two-column warning copy (`3d917f6`)** — reworded the `#356` two-column warning banner. The old copy claimed "Many ATS parsers read text column-by-column-merged or out of order, scrambling the result" — an unverifiable claim about third-party ATS behavior we can't cite (and citing a competitor in public-facing UI is off the table). New copy grounds the warning in what our own extractor actually did:

> This resume uses a two-column layout. Text extractors often read the columns out of order — merging or interleaving them. The reconstructed text below is what a generic parser actually pulled out; if it looks scrambled, that's the ATS risk. A single-column layout parses most reliably.

Softens "Many" → "often", anchors the risk to the reconstructed output already on screen, and stays plain-language and actionable. Kept as a plain string (no JSX/citation), so the `<ErrorState tone="warning">` render path is untouched.

Vaishnavi1709 (changes requested):
Design is solid — opt-in `validate` prop, non-blocking soft warning, shared memory-storage shim, real regex reuse from the parser. The regex composition (wrapping `DATE_ANCHOR` / `DATE_RANGE_RE` in `(?:…)` before anchoring, dropping `/g` when re-anchoring `EMAIL_RE`) is done correctly and I verified every test expectation in `field-validators.test.ts` holds. But four issues need real fixes before I can approve — one is a WCAG accessibility break — plus one pattern observation worth acting on.

## Blocking

### 1. Screen readers get zero warning signal (WCAG 1.4.1)

[`EditableField.tsx:306`](src/design-system/primitives/EditableField.tsx#L306) sets an explicit `aria-label={`Edit ${label}`}` on the read-mode `<span role="button">`. Per ARIA name-from-author precedence, an explicit `aria-label` suppresses ALL descendant content in the accessible name — including the inner `<svg role="img" aria-label={message}><title>{message}`. Users on NVDA/VoiceOver hear only "Edit Email, button" and get zero signal that the value is a shape-fail. The colour-only warning triangle is the only channel, which is WCAG 1.4.1 (Use of Color).

Fix options, cheapest first:
- Extend the button's aria-label: `aria-label={warning ? `Edit ${label} — ${warning}` : `Edit ${label}`}`.
- Add an `aria-describedby` pointing at a visually-hidden text node with the warning message.
- Hoist the glyph out of the aria-labelled button subtree entirely.

### 2. UK / non-US phones are flagged as invalid

[`field-validators.ts:126`](src/lib/edit/field-validators.ts#L126) calls `normalizePhone(value)` with no region — [`phone.ts:164`](src/lib/heuristics/phone.ts#L164) defaults `region: CountryCode = "US"`. The parser's `extractContact` wires `regionFromLocation(parsed.location)` for exactly this reason; the edit-shape guard doesn't, so a UK user typing `020 7946 0958` (local form, no `+44`) sees the warning triangle even though the parser accepts the same value.

Fix: thread the parsed location down to `validatePhone` and call `normalizePhone(value, regionFromLocation(location) ?? "US")`. Test asserts `+44 20 7946 0958` passes but only because it has an explicit `+`.

### 3. `Month Year` template placeholder is treated as clean

[`DATE_FIELD_RE`](src/lib/edit/field-validators.ts#L55) embeds `DATE_ANCHOR`, whose grammar includes `MONTH_OR_PLACEHOLDER = (Month|<real month>)` and `YEAR_FORMS = …|Year`. Those tokens exist so the parser can RECOGNIZE unfilled Word/Office template placeholders and drop them downstream. The edit surface treating them as clean shapes contradicts the score's own completeness signal — a user who leaves `Month Year` in a role sees "missing dates" in the score AND no shape warning at the edit surface.

Fix: reject shapes where every date token is a placeholder (`Month Year`, `Month YYYY`, `Mon Year`) before the general `DATE_FIELD_RE.test(value)` check. Add a corresponding test row.

### 4. `FeedbackPanel.test.tsx` still uses bare localStorage

Grep confirms 4 hook test files adopt `installMemoryLocalStorage`, but [`FeedbackPanel.test.tsx:67`](src/components/features/FeedbackPanel.test.tsx#L67) still does `localStorage.removeItem(k)` in a try/catch — the comment at line 57 already acknowledges "CI localStorage shim has no clear()". On Node 22+/25's built-in `localStorage` the removes silently no-op inside the catch, leaking `rl_feedback_seen`/`rl_feedback_submitted` between tests in the file. That's exactly the flake class #398 documented.

Fix: adopt `installMemoryLocalStorage()` in this file's beforeEach too. Or better — do the altitude fix below and pick this up for free.

## Consistent pattern — surface the fix once, not four times

The Node 22+ built-in-localStorage problem exists at the workload level. Solving it per-file via an `import + call in beforeEach` means every new test that touches localStorage has to opt in (finding #4 above is exactly that miss). Moving the shim to a project-wide `test.setupFiles` entry in [`vite.config.ts`](vite.config.ts) — one line — installs it globally before every test and closes the "forgot to import" regression class permanently. The 4 per-file imports and identical `beforeEach` blocks in the PR all delete.

## Nits (not blocking)

- **UX: `validate` runs on commit, not draft.** [`EditableField.tsx:293`](src/design-system/primitives/EditableField.tsx#L293) computes `warning` from `value` — the user has no live feedback while typing, sees the icon only after Enter/blur. Cheap fix: also compute `validate(draft)` in the editing branch and surface it under the input.
- **SVG `<title>` pollutes read-mode span `textContent` — latent brittleness.** [`ContactCard.test.tsx:102-149`](src/components/features/ContactCard.test.tsx#L102) uses exact-equality `.toBe(...)` on the EditableField span's textContent. Today it passes only because the fixtures are shape-valid → validator returns null → no glyph rendered. Any future fixture that trips a validator silently breaks these assertions with a concatenated warning message. Same underlying fix as #1 (glyph outside the aria-labelled subtree).
- **UX: three parallel two-column messages on one page.** [`Result.tsx:181`](src/components/Result.tsx#L181) `<ErrorState tone="warning">` + `LayoutFlagsList.LAYOUT_TRIGGER_BLURBS.two_column` + `DisagreementPanel.CAUSE_COPY.two_column` all say roughly the same thing in different words. LayoutFlagsList already owns the canonical trigger→copy map — extend it (per CLAUDE.md's reuse gate) rather than introducing a fourth constant.
- **Typing: `EDITABLE_KEYS` and `FIELD_VALIDATORS` use `string` keys.** [`ContactDetails.tsx:35,51`](src/components/features/ContactDetails.tsx#L35) — both should be `Partial<Record<keyof ContactOverrides, …>>`. `EDITABLE_KEYS` is a 1:1 identity map so the two can collapse to one config object per field. A typo like `mastdn_url` currently compiles green.
- **`validate` prop type is inline** in `EditableFieldProps` while `FieldValidator` is exported from `field-validators.ts` — import + reuse so both sides can't drift.
- **`validate(value ?? "")` fallback unreachable** ([`EditableField.tsx:293`](src/design-system/primitives/EditableField.tsx#L293)) — `hasValue = Boolean(value)` already gated it. Drop to `validate(value)`.
- **No test asserts the two-column banner renders.** Silent regression risk if trigger name changes.
- **`MemoryStorage` omits `length` and `key(i)`.** Typed as `Storage` via `as unknown as Storage`, so future code using `localStorage.length` type-checks green and crashes at runtime only in tests.
- **`URL_FIELD_RE`** accepts `_foo.bar` (leading underscore) and `9999` passes as a bare year. Minor grammar looseness — the "Doesn't look like a URL" copy implies stricter checking than the regex delivers.
- **Latent capture-group footgun**: `DATE_FIELD_RE` embeds `DATE_RANGE_RE.source` which has 6 top-level capture groups. `.test()` ignores them but a future `.exec` caller sees spurious groups. Comment-worthy.

Overall the shape is right — I'll approve happily once #1–#4 are addressed. The a11y one is small (one aria-label branch), the phone one is small (one prop passed through), and the `Month Year` one is a targeted exclusion at the top of `validateDate`.

s-annam:
Thanks for the thorough review, @Vaishnavi1709 — all four blocking items + the altitude fix are in `13f6ded`.

**1. Screen-reader warning signal (WCAG 1.4.1)** — Fixed. The read-mode button's `aria-label` now folds in the message: `aria-label={warning ? \`Edit ${label} — ${warning}\` : \`Edit ${label}\`}`. The glyph is `aria-hidden` (its `<title>` was suppressed by name-from-author precedence anyway) but keeps the `<title>` for a mouse-hover tooltip. Went with your cheapest option — one aria-label branch, no extra DOM node.

**2. Non-US phones flagged invalid** — Fixed. Threaded the parsed location down: `ContactDetails` derives `location` from the contact line and binds it into a phone-only validator, and `validatePhone(value, location)` now calls `normalizePhone(value, regionFromLocation(location) ?? "US")` — same wiring as `extractContact`. Added a test asserting `020 7946 0958` flags under the US default but passes with a UK location.

**3. `Month Year` placeholder treated as clean** — Fixed. `validateDate` now rejects values made only of placeholder tokens (`Month Year`, `Mon YYYY`, and ranges of them) via a dedicated `DATE_PLACEHOLDER_RE` **before** the general `DATE_FIELD_RE` check. `20XX` stays clean (it's a real redaction, not a template placeholder). Added a test row.

**4. + Consistent pattern (altitude fix)** — Did the altitude fix. The in-memory shim now installs workload-wide via `test.setupFiles: ["src/test-setup.ts"]` — a global `beforeEach` gives every test a fresh `localStorage`, closing the "forgot to import" class permanently. Deleted the 4 per-file `installMemoryLocalStorage` imports + `beforeEach` calls and `FeedbackPanel.test.tsx`'s bare-`localStorage` `resetFeedbackStorage` workaround (picked up for free, as you predicted). Also gave `MemoryStorage` real `length` + `key(i)` (your nit) since it's now the global `Storage` for the whole suite.

Gates green on the branch: `npm run typecheck` clean, `npm run test` 1675 passed / 3 skipped.

Deferred the remaining nits (live-draft validation, `EDITABLE_KEYS`/`FIELD_VALIDATORS` typing, `Result.tsx` fourth two-column constant, the SVG-`<title>`/`textContent` brittleness) — happy to fold any in if you want them here rather than a follow-up, but keeping this pass scoped to the four blockers + the pattern fix. Re-requesting review.

Vaishnavi1709 (approved):

Issue #401, opened 2026-07-08
Batch: Local-first storage — IndexedDB foundation + resume library + job tracker


s-annam:
## Batch epic — Local-first storage (IndexedDB foundation + resume library + job tracker)

Groups the small issues below for **single-batch implementation** by one person via `/implement-batch` (#387). This is a tracking issue; the work lives in the children.

**Why grouped:** one persistence lane on a shared IndexedDB foundation — the module (#321) is the data layer both consumers build on (resume library #322, job tracker #323). Implemented and tested as one unit (open DB → save → reload → manage). PWA/offline (#344) is intentionally excluded — separate service-worker subsystem.

**Milestone note:** children keep their own milestones — #321 is P2 (prioritized foundation), #322/#323 are P4 (deferred consumers). Parent tracks the active foundation at P2.

### Phase 1 — Foundation and consumers (shipped)

- [x] #321 — `Local-first storage foundation: IndexedDB module for resumes and jobs`
- [x] #322 — `Resume library: save, reload, and manage multiple resumes locally`
- [x] #323 — `Job tracker: save postings, track application status, link resume versions`

### Phase 2 — Durability and portability (open)

Reopened after Phase 1 shipped. Phase 1 built the storage lane and left two gaps in it — both sit on the same IndexedDB foundation and the same `src/lib/storage/backup.ts` surface, so they belong here rather than in a new epic.

- [ ] #573 — `Restore path: wire backup import into the resume library, and scope the save-confirmation copy`
- [ ] #547 — `Wire reconcileResumeLinks into the JSON import path`

**Why these two, together:** the backup round-trip shipped half-built. Export is wired at `ResumeLibrary.tsx:47` and `JobTracker.tsx:98`; `importAll` / `importFromJson` have no production caller, so a downloaded `offlinecv-backup.json` cannot be restored. That matters because IndexedDB is isolated per device, per browser profile, and per origin (`offlinecv.org`, `dev.offlinecv.org`, and every `*.offlinecv.pages.dev` preview are separate databases — Chrome Sync does not sync IndexedDB), and it is evictable. The backup file is the only transport across any of those boundaries.

**Dependency order:** #573 → #547. #547 is explicitly blocked on the merge-mode import UI that #573 builds — `reconcileResumeLinks` (`src/lib/job-tracker.ts:136`) is staged with no production caller, and its stated trigger is a merge import that drops a resume. #547 is small enough to land in the same PR as #573 if the implementer prefers.

[...]

Issue #400, opened 2026-07-08
Batch: Download-export standards — profiles[] model + JSON Resume schema + shareable artifact


s-annam:
## Batch epic — Download-export standards (profiles[] model + JSON Resume schema + shareable artifact)

Groups the small issues below for **single-batch implementation** by one person
via `/implement-batch` (#387). This is a tracking issue; the work lives in the children.

**Why grouped:** all three touch the resume export/download path and its data model — one shared test surface (parse → export → re-parse roundtrip), implemented and reviewed as one unit. Dependency chain: the `profiles[]` model (#335) feeds the JSON Resume export schema (#334), which the shareable artifact (#343) then emits.

### Children
- [ ] #335 — `Generalize contact link fields to a profiles[] model (JSON Resume basics.profiles pattern)`
- [ ] #334 — `Align Download PDF export with resume-format standards (JSON Resume schema + embedded JSON attachment + documented ATS layout contract)`
- [ ] #343 — `Export the audit report as a shareable client-side artifact (PDF + JSON)`

**Implement with:** `/implement-batch <THIS#>`

Issue #334, opened 2026-07-05
Align Download PDF export with resume-format standards (JSON Resume schema + embedded JSON attachment + documented ATS layout contract)


s-annam:
# Align Download PDF export with resume-format standards

The Download PDF exporter (`src/lib/pdf/render-ats-pdf.ts` + `ats-resume-model.ts`, #171) is already ATS-safe by construction. This issue aligns it with a recognized standard — [JSON Resume](https://jsonresume.org/) — instead of only our own judgment, and makes the export round-trip losslessly by embedding a machine-readable copy inside the PDF. Stays fully client-side.

> Part of batch #400. **Blocked by #335** — `basics.profiles` maps from the `profiles[]` model that #335 introduces.

## 1. `toJsonResume()` adapter

New **pure** module `src/lib/pdf/to-json-resume.ts` — `(model: AtsResumeModel) => JsonResume`. Mapping:

- `basics.name` / `basics.email` / `basics.phone` / `basics.location` ← `AtsContact`
- **`basics.profiles` ← `profiles[]`** → `{ network, url, username: <urlSlug tail> }` — this is the #335 dependency.
- `basics.url` ← first `portfolio` / `website` profile, if any
- `work[]` ← experience sections: `{ name: company, position: title, startDate, endDate, highlights: bullets }`
- `education[]` ← education entries
- `skills[]` ← skills as `{ name }` entries
- `projects[]` ← projects
- Top-level `$schema` + `meta.version` (app version from `src/lib/version.ts`)

**Dates:** our parsed dates are free-form strings. Best-effort normalize to `YYYY-MM` (or `YYYY`); when a date can't be parsed confidently, **emit the raw string** (JSON Resume tolerates partial dates). Never fabricate a date.

Pure function, no I/O, no `pdf-lib` import — unit-testable in isolation.

## 2. Documented ATS layout contract + enforcing test

Document (in the module and a test) the contract the exporter guarantees: **single column, reverse-chronological, canonical section headers, standard fonts.** Enforce the headers half with a test that takes every heading the exporter emits (`AtsSection.heading` for each section + `summaryHeading`) and asserts our own **`matchSectionHeader()`** recognizes it — so the export can never emit a heading the parser wouldn't re-recognize on re-upload.

## 3. Embed `resume.json` attachment (Europass pattern)

[...]

PR #402, opened 2026-07-08
chore(skills): current milestones in triage-issue + zsh-safe dep wiring in batch-issues

s-annam:
## Summary

Two skill-file fixes surfaced while batching parentless issues (epics #400, #401):

- **`triage-issue`** — its milestone rules of thumb referenced a stale subsystem scheme (`parser-hardening`, `UI-parity`). Replaced with the live audience-gated `P1 · Friends & Family` … `P4 · Post-Public` buckets and their roles (still matched against the live Step 0 list, not the snapshot). Step 4 now handles the fact that the board's `Phase` field still carries the old M-scheme and no longer maps 1:1 to milestones — set Phase only when an option corresponds, else leave it blank (the current convention), never force a wrong Phase. First-time-setup Phase example mirrors P1–P4.
- **`batch-issues`** — fixed the dependency-order wiring. The `for ((i=1;…)); ${ORDER[i-1]}` loop assumed **bash 0-indexed** arrays, but resumelint's shell is **zsh (1-indexed)**, so it wired the wrong pairs: empty first blocker, last edge skipped. This is why the two batches above half-wired their chains and needed manual repair. Rewrote with a `PREV` value cursor (shell-agnostic), numeric-guarded resolved ids before POST, suppressed `gh api` POST output, swallowed the benign "already been taken" dup, and added a full-chain audit that surfaces missing **and** stray edges.

Docs/skill-only change — no product code touched.

## Test plan

- [x] zsh-safe dependency loop verified against live epic #400 chain (334←335, 343←334)
- [x] `npm run typecheck` clean; `npm run test` = 1645 passed / 13 failed — the 13 are the pre-existing #398 `localStorage.clear is not a function` shim failures (`useAnalyzedResume`/`useDownloadPdf`, shim fix lives on the #397 branch), unrelated to this docs-only diff

### DIFF
--- a/.claude/skills/batch-issues/SKILL.md
+++ b/.claude/skills/batch-issues/SKILL.md
@@ -179,7 +179,9 @@ LINKED=$(gh api "repos/$REPO/issues/$PARENT/sub_issues" --jq '.[].number' 2>/dev
 for N in "${CHILDREN[@]}"; do
   grep -qx "$N" <<<"$LINKED" && { echo "#$N already linked — skip"; continue; }
   CHILD_ID=$(gh api "repos/$REPO/issues/$N" --jq .id)
-  gh api -X POST "repos/$REPO/issues/$PARENT/sub_issues" -F sub_issue_id="$CHILD_ID"
+  [[ "$CHILD_ID" =~ ^[0-9]+$ ]] || { echo "✗ can't resolve id for #$N — skipping link"; continue; }
+  gh api -X POST "repos/$REPO/issues/$PARENT/sub_issues" -F sub_issue_id="$CHILD_ID" >/dev/null 2>&1 \
+    && echo "linked #$N"
 done
 ```
 **Verify by diffing the readback against the input set** — never assume the loop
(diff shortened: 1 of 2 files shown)


Samhit21 (commented):
Reviewed as code (Step 3e — skill-file `gh`/bash blocks are the surface). The headline fix is right: the `PREV`-cursor value iteration (`:245`) replaces the bash-0-indexed `${ORDER[i-1]}` loop and wires the correct pairs in both bash and zsh (verified). Milestone-bucket + blank-Phase changes in `triage-issue` are prose/policy, no command risk. Gates 3a–3d N/A (docs-only). Flagging one net-new Secondary — the same zsh-safety class this PR is fixing — worth a quick fix before merge; the Nit is optional.

## Blocking

None.

## Secondary

- **`batch-issues/SKILL.md:276` — the stray-edge audit loop isn't zsh-safe, which is the exact bug class this PR fixes.** `for h in ${HAVE//,/ }` relies on **word-splitting an unquoted expansion** — a bash-ism. zsh (the shell this PR insists skills run under) does **not** word-split unquoted expansions, so for a child with ≥2 `blocked_by` entries (`HAVE="334,340"`), `h` iterates **once** as the literal `"334 340"` instead of twice. Under zsh that emits a garbled `⚠ #C also blocked_by #334 340` and never flags the real stray edge individually — the audit misfires precisely on the multi-edge case it exists to catch. The test plan only exercised a **linear** chain (334←335, 343←334) where every child has exactly one blocker, so no split was needed and the bug stayed hidden. (The sibling MISSING check just above is quoted — `"${HAVE//,/$'\n'}"` — and is correct; only this stray loop breaks.) Verified: `zsh -c 'HAVE=334,340; for h in ${HAVE//,/ }; do echo [$h]; done'` prints one `[334 340]`; bash prints two. Fix with a read-loop (shell-agnostic):
  ```bash
  while IFS= read -r h; do
    [[ -z $h || $h == "$PREV" ]] && continue
    echo "⚠ #$CUR also blocked_by #$h — not in the intended chain; remove if spurious"
  done < <(printf '%s\n' "${HAVE//,/$'\n'}")
  ```
  Worth a diamond/multi-dep line in the test plan too, so this path is actually covered.

## Nits

- **`batch-issues/SKILL.md:183` — Step 4's sub-issue POST swallows *all* stderr while Step 5's edge POST surfaces non-dup errors.** `... >/dev/null 2>&1 && echo "linked #$N"` sends every error to `/dev/null`, so an auth/rate-limit failure is invisible until the readback diff — whereas `wire_edge` (`:253`) routes stderr to a tmpfile and prints anything that isn't the benign dup. Align them (`2>/tmp/link.err` + surface non-benign errors) so the two write paths fail loudly the same way.

---

Posting as a **comment, not an approval** — the Secondary is a read-only audit path (the actual wiring is correct), so it's not merge-blocking, but given this PR's entire premise is zsh-safe dep wiring I'd rather see the one-line `:276` fix land first. Happy to approve once that's in.

s-annam:
Thanks @Samhit21 — both addressed in `e790b06`, plus one thing your Secondary surfaced that runs deeper.

**Secondary (`:276` stray-edge audit) — fixed, and it was worse than flagged.** You're right that `for h in ${HAVE//,/ }` isn't zsh-safe. But while fixing it I checked the sibling MISSING check's `${HAVE//,/$'\n'}` idiom you cited as correct — and it *isn't* zsh-safe either. zsh leaves the ANSI-C `$'\n'` **literal** in a `${var//pat/repl}` replacement (bash expands it):
```
$ zsh  -c 'HAVE=334,340; printf "%s\n" "${HAVE//,/$'"'"'\n'"'"'}"' | od -c
0000000    3   3   4   $   '   \   n   '   3   4   0  \n
$ bash -c 'HAVE=334,340; printf "%s\n" "${HAVE//,/$'"'"'\n'"'"'}"' | od -c
0000000    3   3   4  \n   3   4   0  \n
```
So the suggested `done < <(printf '%s\n' "${HAVE//,/$'\n'}")` fix would have inherited the same bug. Rewrote **both** the MISSING check and the stray loop with `tr ',' '\n'` (identical in bash and zsh), stray loop now a quoted read-loop. Verified both shells split `334,340` into two lines.

**Nit (`:183`) — fixed.** Step 4's sub-issue POST now routes stderr to a tmpfile and surfaces any non-benign failure (swallowing only the benign "already" dup), matching Step 5's `wire_edge`.

**Gates:** `npm run typecheck` clean. `npm run test` = 1645 passed / 13 failed — the 13 are the pre-existing #398 `localStorage.clear is not a function` shim failures (`useAnalyzedResume`/`useDownloadPdf`, fix on the #397 branch), unrelated to this docs-only diff; verify/fallow CI green. Checked the test-plan box.

Re-requesting review.

Samhit21 (approved):
**APPROVE** — re-reviewed the fixup (`e790b06`) as code; verified in both shells.

- **Secondary `:276` (zsh stray-edge audit) — fixed, and you found the deeper instance I missed.** You're right that the MISSING check's `${HAVE//,/$'\n'}` was *also* not zsh-safe — confirmed with `od -c`: bash expands to `\n`, zsh leaves the literal `$'\n'`. Both splits now go through `printf … | tr ',' '\n'`, which I verified produces two lines in bash *and* zsh, and the stray loop is a quoted read-loop (empty-`HAVE` no longer emits a spurious warning). Nice catch tightening my flag.
- **Nit `:183` — fixed.** Step 4's sub-issue POST now surfaces non-benign errors and swallows only the dup, matching Step 5's `wire_edge`.

Verdict rule: 0 Blocking ⇒ APPROVE. The 13 failing tests are the pre-existing #398 `localStorage` shim failures on `main`, unrelated to this docs-only diff.

Issue #370, opened 2026-07-07
[parser] experience — per-entry location dropped when "<dates> · <location>" spans a column break


Vaishnavi1709:
## Category
parser (Steps 1 & 3 F&F QC — provenance category)

## Fixture
`tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-two-column.pdf` (synthetic persona)

## Ground truth (`pdftotext`)
```
Acme Corp — Senior Software Engineer
Jan 2022 – Present · Springfield, IL

…

Globex Corporation — Software Engineer
Mar 2019 – Dec 2021 · Remote
```

Each entry's date line follows the `<dates> · <location>` shape. All three experience entries have a per-entry location.

## Observed
Reconstructed view shows `⚠ location not detected` for ALL three experience entries, despite the candidate's own contact block correctly extracting `Springfield, IL` and the résumé clearly showing three per-entry locations.

Same fixture, note the candidate's home location IS picked up correctly — the miss is specifically for the per-entry experience location line.

## Localized layer
`src/lib/heuristics/extract/experience.ts` — the per-entry location extractor does not consume the `<dates> · <location>` pattern when the sidebar column interleaves lines in reading order. The `·` separator between dates and location is likely being read as a section-delimiter or dropped during two-column reading order.

## Repro
```
npm run dev
# drop tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-two-column.pdf
```

Layout penalty (multiplier 0.85) is applied and visible, so the honest UX signalling is intact — this issue is about recovering data the parser could otherwise get right.

rohithgollapalli:
No longer reproduces on `main` (@ eefcce4). All three per-entry `<dates> · <location>` locations on `google-docs-skia-proxy-two-column.pdf` now extract:

- Acme Corp → `Springfield, IL`
- Globex Corporation → `Remote`
- Initech → `Pacific Coast, CA`

Same root cause / fix as #369 — page-wide two-column gutter detection now fires, so the sidebar no longer interleaves and the date/location line is read intact. #404 adds a value-asserting regression test covering all three locations. **Recommend closing as resolved once #404 lands.**

Issue #369, opened 2026-07-07
[parser] experience — Google-Docs two-column layout drops company name and corrupts multi-word city


Vaishnavi1709:
## Category
parser (Steps 1 & 3 F&F QC — provenance category)

## Fixture
`tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-two-column.pdf` (synthetic persona)

## Ground truth (`pdftotext`)
```
Initech — Junior Engineer

AWA R D S

Aug 2017 – Feb 2019 · Pacific Coast, CA
```

Company `Initech`, title `Junior Engineer`, dates `Aug 2017 – Feb 2019`, location `Pacific Coast, CA`. The two-column reader interleaves the AWARDS sidebar heading between the header and the date/location line.

## Observed
Reconstructed view:
```
Junior Engineer — · Pacific, Coast, CA        Aug 2017 – Feb 2019
```

- Company `Initech` **dropped entirely**
- The `·` middot from `Aug 2017 – Feb 2019 · Pacific Coast, CA` bled into the company slot
- Multi-word city `Pacific Coast` corrupted to `Pacific, Coast` (extra comma)

## Localized layer
Two-column reading-order interleaving in `src/lib/heuristics/pdf-extract.ts`. The `AWA R D S` sidebar heading landing between the header and the date line is likely what breaks the entry's line-locality assumption downstream in `extract/experience.ts`.

## Related
Similar in shape to the multi-word-city split filed for the LaTeX fixture (`[parser] experience — LaTeX two-column layout splits multi-word cities mid-name`), but the failure mode is different: LaTeX loses a token to the neighbouring column; Google-Docs loses the whole company AND inherits a stray separator.

## Repro
```
npm run dev
# drop tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-two-column.pdf
# scroll to the 3rd experience entry (Junior Engineer)
```

rohithgollapalli:
No longer reproduces on `main` (@ eefcce4). Driving `google-docs-skia-proxy-two-column.pdf` through the real parser now yields the Initech entry correctly:

- `company: "Initech"` — not dropped, no `·` bled into the slot
- `location: "Pacific Coast, CA"` — multi-word city whole, not `Pacific, Coast`
- `title: "Junior Engineer"`

[...]

Issue #393, opened 2026-07-07
Batch: parser experience two-column / column-break bugs


s-annam:
## Batch epic — Parser experience two-column / column-break bugs

Groups the small issues below for **single-batch implementation** by one person
via `/implement-batch`. This is a tracking issue; the work lives in the children.

**Why grouped:** all are experience-section extraction failures on two-column / column-break layouts (LaTeX, Google-Docs) — dropped locations, split cities, team-vs-company mislabeling. Same subsystem and layout-probe context; fixing them together avoids overlapping edits to the same segmenter.

### Children
- [ ] #368 — LaTeX two-column layout splits multi-word cities mid-name
- [ ] #369 — Google-Docs two-column layout drops company name and corrupts multi-word city
- [ ] #370 — per-entry location dropped when "<dates> · <location>" spans a column break
- [ ] #372 — team name after comma in title promoted to company, real company on next line dropped
- [ ] #373 — location dropped when "Company | Location Date" line lacks a separator
- [ ] #382 — mislabels team as company on "Title, Team" roles under a shared-employer banner

**Implement with:** `/implement-batch <THIS#>`

s-annam:
All six child issues are resolved and closed:

- #368 — LaTeX two-column splits multi-word cities ✅
- #369 — Google-Docs two-column drops company / corrupts city ✅
- #370 — per-entry location dropped across a column break ✅
- #372 — team-after-comma promoted to company, real company dropped ✅
- #373 — location dropped on `Company | Location Date` without separator ✅
- #382 — team mislabeled as company on `Title, Team` under shared-employer banner ✅

Batch epic complete — closing the tracking issue.

Issue #372, opened 2026-07-07
[parser] experience — team name after comma in title promoted to company, real company on next line dropped


Vaishnavi1709:
## Category
parser (Steps 1 & 3 F&F QC — provenance category)

## Fixture
`tests/fixtures/pdfs/unknown/letter-spaced-name-heading.pdf` (synthetic persona)

## Ground truth (`pdftotext`)
```
Software Engineer II, Business Credit Journey
Globex Financial | New York, NY August 2024 - Present
…
Software Engineer I/II, Import Aggregation
Globex Financial | New York, NY July 2022 - August 2024
```

Both entries: title includes a team-name suffix after the comma (`Software Engineer II, Business Credit Journey`). Real company `Globex Financial` sits on the next line, followed by `| <location> <dates>`.

## Observed
Reconstructed view:
```
Software Engineer II — Business Credit Journey — August 2024 – Present
Software Engineer I/II — Import Aggregation — July 2022 – August 2024
```

Company field shows the team name; `Globex Financial` is dropped entirely.

## Localized layer
`src/lib/heuristics/extract/experience.ts` — the header parser is treating the comma in `Software Engineer II, Business Credit Journey` as a title/company boundary and promoting the team-name suffix to `company`. The Company-|-Location-Date next line is not being read as the true company source.

## Suggested fix
Two candidate signals to distinguish "title, team" from "title, company":
1. The next line matches `INSTITUTION_HINTS`-like company shape (`COMPANY_SUFFIX_RE` or a proper-noun heuristic). If so, the current-line suffix is a team name, not a company.
2. The current-line suffix lacks a company suffix token (`Inc`, `LLC`, `Corp`, `Ltd`) — a team-name shape.

## Related
Bonus find on the same fixture: the location `New York, NY` on the `Globex Financial | New York, NY August 2024 - Present` line also drops (see the sibling issue for the missing-separator location bug).

## Repro
```
npm run dev
# drop tests/fixtures/pdfs/unknown/letter-spaced-name-heading.pdf
# scroll to Experience section
```

PR #404, opened 2026-07-08
test(experience): lock two-column sidebar experience values (#369, #370)

rohithgollapalli:
## Summary

**#369 and #370 no longer reproduce on `main`.** Driving `google-docs-skia-proxy-two-column.pdf` through the real parser now yields all three roles with correct fields:

|---|---|---|---|
| 1 | Senior Software Engineer | Acme Corp | Springfield, IL |
| 2 | Software Engineer | Globex Corporation | Remote |
| 3 | Junior Engineer | **Initech** | **Pacific Coast, CA** |

Accumulated two-column / sidebar reading-order work made page-wide gutter detection fire (`columnBoundaries` splits the page at x≈229.5), so the `AWARDS` sidebar block reads fully *before* the experience column instead of interleaving into the third role. That interleave was the root cause of both:

- **#369** — Initech company dropped, `·` bled into the company slot, `Pacific Coast` → `Pacific, Coast`. All gone.
- **#370** — every role's `<dates> · <location>` location "not detected". All three now detected.

**Why this PR is test-only:** there's no fix to write — but the lossy `*.expected.json` snapshot only records `experienceCount: 3` and **cannot** catch a company drop or a city-value corruption, so a future regression would slip past it silently. This adds a value-asserting repro that locks each role's title/company/location, plus explicit anti-corruption assertions on the Initech entry (company is exactly `Initech`, location is exactly `Pacific Coast, CA`, no `AWARDS` text leaks).

Refs #369, #370 — recommend closing both as resolved-with-guard once this lands.

## Test plan

- [x] `npm run lint` clean
- [x] New `two-column-experience-values.repro.test.ts` green (3 tests)
- [x] Full heuristics suite green
- [x] No source change — pure regression guard

### DIFF
--- a/src/lib/heuristics/two-column-experience-values.repro.test.ts
+++ b/src/lib/heuristics/two-column-experience-values.repro.test.ts
@@ -1,0 +1,94 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+/**
+ * Value-locking regression for #369 and #370 — a two-column Google-Docs résumé
+ * whose left sidebar (CONTACT / SKILLS / EDUCATION / AWARDS) sits alongside the
+ * right-column EXPERIENCE list.
+ *
+ * Both issues were filed against an earlier reading order that interleaved the
+ * `AWARDS` sidebar block into the third experience entry (Initech):
+ *   - #369: Initech's company was dropped, the `·` date/location separator bled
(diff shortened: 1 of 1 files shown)


rohithgollapalli:
Superseded by #406, which combines the batch #393 experience fixes (#368, #369/#370, #372) into one PR. Closing this in favor of the combined PR. The commit and its regression test are carried over unchanged.

PR #406, opened 2026-07-08
fix(experience): two-column / column-break experience field batch (#393)

rohithgollapalli:
## Summary

Batch fix for **epic #393** — experience-section field extraction on two-column / column-break résumé layouts. Three independent fixes to `extract/experience.ts` (each with its own regression test), combined into one PR since they share the subsystem.

### 1. Multi-word cities on space-folded headers — Closes #368

`stripLocationSuffix`'s single-token space pass truncated a multi-word city glued to the company line (`Greenfield Studios  New York, NY` → `company: "Greenfield Studios New"`, `location: "York, NY"`). Adds a closed-vocabulary `KNOWN_MULTIWORD_US_CITY_RE` pass (tried before the single-token one), mirroring Pass C/D's country-gazetteer discipline. Single-word cities unaffected.

### 2. `Title, Team` over `Company | Location Dates` — Closes #372

A `Title, Team` header over a delimited company/date line mislabeled the post-comma team as the company and demoted the real employer (`company: "Business Credit Journey"` / `team: "Globex Financial"` → swapped correct). When `splits[0]/[1]` are a role-comma split of one line and the anchor line supplies a leading company segment, route the post-comma piece to `team` and take the company from the anchor line.

### 3. Two-column sidebar value guard — Closes #369, Closes #370

Both already reproduce clean on `main` (accumulated two-column/sidebar work made page-wide gutter detection fire, so the `AWARDS` sidebar no longer interleaves into the Initech role). No source fix needed — but the lossy `*.expected.json` snapshot only records `experienceCount`, so a value regression would pass silently. Adds a value-asserting repro locking all three roles' title/company/location.

## Verification

| Check | Result |
|---|---|
| `npm run typecheck` | clean |
| `npm run lint` | clean |
| `npm run test` (heuristics) | 639 passed, 0 failed — incl. `corpus` + `corpus-roundtrip` invariants |
| `npm run build` | clean |
| New tests | `experience.multiword-city.test.ts`, `experience.title-team-comma.test.ts`, `two-column-experience-values.repro.test.ts` — the two fix tests confirmed red without their fix |
| Real fixtures | `latex/multi-degree-coursework.pdf`, `unknown/letter-spaced-name-heading.pdf`, `google-docs/google-docs-skia-proxy-two-column.pdf` all parse correct fields |

All changes are surgical/additive to `disambiguateCompanyTitle` / `stripLocationSuffix`; no fixture binaries added (synthetic personas verified via the existing in-tree fixtures).

## Not included

[...]

PR #409, opened 2026-07-08
fix(experience): recover location from separator-less "Company | Location Dates" cell (#373)

### LINKED ISSUE #373: [parser] experience — location dropped when "Company | Location Date" line lacks a separator between location and date
## Category

## Fixture
`tests/fixtures/pdfs/unknown/letter-spaced-name-heading.pdf` (synthetic persona)

## Ground truth (`pdftotext`)
```
Globex Financial | New York, NY August 2024 - Present
Globex Financial | New York, NY July 2022 - August 2024
```

Company + `|` + location + SPACE + dates. There is no `|` / `·` / comma between the location and the date — only a single space.

## Observed
Both entries show `⚠ location not detected` — the `New York, NY` never reaches the `location` field, even though the primitives clearly see the tokens.

## Localized layer
`src/lib/heuristics/extract/experience.ts` — the header parser's location-strip likely splits on `|` and grabs "New York, NY August 2024 - Present" as the last cell, then hands the whole cell to date extraction which extracts the range but leaves the residue location unclaimed.

## Suggested fix
After date extraction on a `|`-split cell, peel the trailing `<dates>` shape off the residue and, if a `<City>, <ST>` or `<City>, <Country>` tail survives, promote it to `location`. Mirrors the trailing-date strip already used in `stripInstitutionLocation` for education.

## Related
Same fixture also drops the real company (see `[parser] experience — team name after comma in title promoted to company, real company on next line dropped`) — these two bugs stack: the parser gets the title wrong AND the location wrong on the same header line.

## Repro
```
npm run dev
# drop tests/fixtures/pdfs/unknown/letter-spaced-name-heading.pdf
```

rohithgollapalli:
## Summary

Recovers a per-entry location that was dropped when a `Company | Location Dates` header cell folds the location and date range with **no separator** between them — e.g. `Globex Financial | New York, NY August 2024 - Present`. `parseEntryBlocks` strips the trailing dates before disambiguation, so the `|` cell reaching the field mapper is a clean `New York, NY`, but because the location sat *before* the removed dates, `stripLocationSuffix`'s end-anchored passes never claimed it and it was silently dropped.

Adds `locationFromAnchorCell()` and step **(3c)** in `disambiguateCompanyTitle`: scan the anchor line's non-company/non-title cells and, when one yields a bare location (peeling any still-glued date range first), use it — clearing it from `team` if it landed there. The title cell is skipped so the empty-title `Company · City, ST` export shape still routes through step 5's rescue.

Closes #373

> **Stacked on #406** (`fix/393-two-column-experience-batch`) — the fix builds on that batch's `disambiguateCompanyTitle` changes. GitHub will auto-retarget this PR to `main` once #406 merges; rebase to drop the shared commits then.

## Test plan

- [x] `eslint .` clean
- [x] `npm run test` green (1668 passed; new `experience.pipe-location.test.ts` covers multi-word + single-word city)
- [x] `vite build` succeeds

### DIFF
--- a/src/lib/heuristics/extract/experience.pipe-location.test.ts
+++ b/src/lib/heuristics/extract/experience.pipe-location.test.ts
@@ -1,0 +1,62 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+/**
+ * Regression for #373 — a per-entry location dropped when a
+ * "Company | Location Dates" header line has no separator between the location
+ * and the date range:
+ *
+ *   Globex Financial | New York, NY August 2024 - Present
+ *
+ * `parseEntryBlocks` strips the trailing date range before disambiguation, so
+ * the `|` cell reaching the field mapper is a clean "New York, NY" — but the
+ * location sits BEFORE the (removed) dates, so `stripLocationSuffix`'s
+ * end-anchored passes never claimed it and the location was dropped. A new step
+ * recovers a bare location from the anchor line's non-company/non-title cell.
+ *
+ * Synthetic personas only, per the fixtures PII policy.
+ */
+
+import { describe, it, expect } from "vitest";
+import { groupIntoLines, splitIntoSections, findSection } from "../sections.ts";
+import { extractExperience } from "../extract-fields.ts";
+import { mkItems } from "../__test-utils__/mkItem.ts";
+
+function roleFromSection(specs: Array<{ text: string; fontSize?: number }>) {
+  const sections = splitIntoSections(groupIntoLines(mkItems(specs)));
+  const experience = findSection(sections, "experience");
+  expect(experience).toBeDefined();
+  return extractExperience(experience).value;
+}
+
+describe("location on 'Company | Location Dates' with no separator (#373)", () => {
+  it("recovers the location from the pipe cell (alongside the #372 team mapping)", () => {
+    const roles = roleFromSection([
+      { text: "Experience", fontSize: 13 },
+      { text: "Software Engineer II, Business Credit Journey", fontSize: 11 },
+      {
+        text: "Globex Financial | New York, NY August 2024 - Present",
+        fontSize: 11,
+      },
+      { text: "• Ran Cassandra design sessions with technical leads.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    const role = roles[0];
+
+    expect(role.company).toBe("Globex Financial");
+    expect(role.team).toBe("Business Credit Journey");
+    expect(role.location).toBe("New York, NY");
+  });
+
+  it("recovers a single-word city too", () => {
+    const roles = roleFromSection([
+      { text: "Experience", fontSize: 13 },
+      { text: "Staff Engineer, Platform", fontSize: 11 },
+      { text: "Initech | Austin, TX July 2022 - August 2024", fontSize: 11 },
+      { text: "• Owned the deploy pipeline.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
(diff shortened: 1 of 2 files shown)


s-annam (approved):
**APPROVE** — the #373 fix is correct and well-guarded. Green across the board (144 extract + 101 corpus/round-trip tests, typecheck, lint), and the new step (3c) is closed-vocabulary–gated so it can only claim a genuine bare location. No blocking findings; two non-blocking notes below.

Verdict rule applied: **0 Blocking → APPROVE.**

## Blocking

None.

## Secondary

- **`locationFromAnchorCell` fallback branch (`experience.ts:221–226`) has no test.** Both new cases feed anchors whose trailing range `stripDateRange` already peels (`"Globex Financial | New York, NY August 2024 - Present"` → the cell reaching disambiguation is a clean `"New York, NY"`), so `isBareLocationString(c)` returns at line 220 and the `.replace(DATE_RANGE_RE, " ")` fallback never runs. The comment promises it recovers "a format `stripDateRange` didn't catch," but that path is currently unexercised. Add a case where the date range is still glued to the cell at disambiguation time (i.e. a range `stripDateRange` misses) so the fallback is actually covered — otherwise it's defensive dead-ish code the tests can't defend against regressions.

## Nits

- **Adds ~22 lines to `disambiguateCompanyTitle`, already fallow-CRITICAL** (80 cyclomatic / 382 lines / CRAP 1462 against this base). Fallow attributes it as *inherited* so the audit gate excludes it and CI (report-only) stays green — not a blocker for this PR. But the function is now a location-rescue pipeline of steps 1→2→3→3a→3b→3c→…; worth a follow-up issue to extract the location-recovery ladder into its own helper before the next location fix lands on top of it.

Personas in the new test are synthetic (`Globex Financial`, `Initech`, `New York, NY` / `Austin, TX`) — PII policy clean. No fixture binary, UI, style-token, or skill/script surface in the diff, so those gates don't apply.

rohithgollapalli:
Thanks for the review! Addressed the two notes:

**Secondary — untested `locationFromAnchorCell` fallback** → **removed it** in `38f398c`. You're right that it never runs: `locationFromAnchorCell` only ever sees anchor-line cells, and `parseEntryBlocks` already runs `stripDateRange` on the anchor line (`line-primitives.ts:155`) using the *same* `DATE_RANGE_RE` the fallback used — so no glued range can survive to it. Rather than add a contrived test for an unreachable path, I dropped the branch (and the now-unused `DATE_RANGE_RE` import). `locationFromAnchorCell` is now just the bare-location check, and the comment notes why no peel is needed. Full heuristics suite still green (641 passed).

**Nit — extract the location-recovery ladder from `disambiguateCompanyTitle`** → agreed, deferring to a follow-up refactor (behavior-preserving decomposition to get the function back under the fallow threshold before the next location fix). Not folding it into this PR. I'll file the follow-up issue and link it here.

Note the push dismissed the approval (dismiss-stale-on-push) — re-requesting review.

s-annam (approved):
Differential re-review (delta `6038ca1` → `38f398c`, the dead-code removal from the prior review).

Confirmed the unreachability claim in source: `parseEntryBlocks` feeds the anchor line into `headers` as `stripDateRange(anchorLine.text)` (`entry-blocks.ts:1105`) — same `DATE_RANGE_RE` — before it's split into cells, and `locationFromAnchorCell` only runs on anchor-line cells (guarded `s.source === anchorIdx`). No glued range can survive to the removed `.replace(DATE_RANGE_RE, …)` fallback. Correct to delete rather than test a path production can't reach.

Import cleanup + doc comment accurate; build green. LGTM.

rohithgollapalli:
Follow-up refactor for the Nit filed: #411 (extract the location-recovery ladder out of `disambiguateCompanyTitle`).

Issue #415, opened 2026-07-08
Epic — skills section recognition & extraction robustness


s-annam:
## Why

The skills section is the parser's most fragile lane. Across the backlog it drops or mangles under a recurring set of causes — unrecognized header wording (#122, #282, closed), leading decorative glyphs (#414), two-line header wrap (#374), sidebar/label bleed (#61, #355), and tokenizer over-split (#220, #299, #301). Each was filed and fixed one-off. This epic makes **skills-section robustness a tracked lane** with a shared triage tool, so new failure classes are localized to a layer, deduped against the known set, and covered by a synthetic corpus fixture on the way in.

## Scope

Two layers own every skills failure:

- **Section routing** — `matchSectionHeaderDetailed` (`src/lib/heuristics/regex.ts`) is exact-match after a trailing-punct strip. Any header the exact matcher rejects (leading glyph, out-of-alias wording, wrapped across two lines) means no skills region is ever routed → the whole section drops.
- **Skills extraction** — `src/lib/heuristics/extract/skills.ts` turns a routed region into skill entries; tokenizer / bleed bugs live here.

## Tooling — `probe-skills` (this PR)

The triage lane for the routing layer. `probe-skills` runs the real parser over an arbitrary (uncommitted, possibly PII-bearing) résumé and prints the router's INPUT (header candidates it saw + the skills region it scanned) next to its OUTPUT (parsed skills), with an independent header-recognition oracle that names *why* a skills-like header was rejected. Verdicts: `ok` / `EXTRACTION-MISS` (region routed, extractor dropped it) / `HEADER-UNRECOGNIZED` (routing rejected a skills-like header) / `NO-SKILLS-SECTION`. Sibling of `probe-contact` / `probe-experience` / `probe-roundtrip`; same PII guardrail (real PDF never committed; values → gitignored `internal/skills/`).

## Children (open)

- #374 — `[parser] skills` — "Technical Skills" section dropped when the header wraps across two visual lines (routing).
- #414 — `[parser] skills` — section dropped when the header carries a leading decorative glyph, e.g. icon-font `¥Skills` (routing).

New skills failure classes found via `probe-skills` land here as additional children.

## Acceptance (epic-level)

- [ ] `probe-skills` skill + harness merged and documented alongside the sibling probes.
- [ ] Each open child (#374, #414) fixed with a synthetic-persona corpus fixture that fails before and passes after.
[...]

PR #416, opened 2026-07-08
feat(skills): probe-skills tooling + leading-glyph (#414) & two-line-wrap (#374) header fixes

### LINKED ISSUE #374: [parser] skills — "Technical Skills" section dropped when the header wraps across two visual lines
## Category

## Fixture
`tests/fixtures/pdfs/unknown/letter-spaced-name-heading.pdf` (synthetic persona)

## Ground truth (`pdftotext`)
```
Technical
Skills

Java
Spring
Boot

Python
Spark

SQL/CQL
Javascript/React

Kafka
Jenkins

Unit/Perf/E2E
Testing
AWS Cloud
Practitioner
```

10+ skills across two columns. The section header is `Technical Skills` but pdftotext extracts it as two separate lines (`Technical\nSkills`), suggesting the source PDF places them on visually-distinct lines.

## Observed
Reconstructed view shows `SKILLS` header + `No skills detected.` — the whole section is lost. Score narrative includes `Missing: skills`.

The extractor sees no `Skills` section header because `Technical\nSkills` doesn't fold back into a single header token — the header-matcher receives `Technical` (rejected) and `Skills` (a bare word without the `Technical` qualifier), and both are treated as content.

## Localized layer
`src/lib/heuristics/regex.ts` — `matchSectionHeader` / `matchSectionAnchorToken` / `matchSectionHeaderDetailed`. These recognize `Skills`, `Technical Skills`, `Skills & Expertise`, etc. as single-line headers, but not a two-line-wrapped variant.

Also potentially the line-assembly upstream: if `Technical` and `Skills` are on distinct baselines with a large gap, item-to-line merging refuses to fold them, so the header-matcher never sees the concatenated form.

## Suggested fix
Two options — needs a triage call:
1. **Line-assembly fix:** when two consecutive short lines both look like header-shape (Title Case, ≤ 3 words, no punctuation) and the second one alone matches a known section-header token, fold them into one header line.
2. **Header-matcher fix:** on a bare `Skills` line whose preceding line is `Technical`/`Core`/`Key`/`Hard`/etc., admit the pair as a section header.

## Repro
```
npm run dev
# drop tests/fixtures/pdfs/unknown/letter-spaced-name-heading.pdf
# scroll to SKILLS section
```

### LINKED ISSUE #414: [parser] skills — section dropped when header carries a leading decorative glyph (e.g. icon-font "¥Skills")
Confirmed parser section-recognition bug, surfaced via `/probe-roundtrip` on a real résumé. The round-trip itself was **clean** — this is an *initial-parse* miss (the export drops the unrecognized header consistently, so parse→export→parse round-trips fine while still losing the section). Reported by category only; no candidate values.

## Symptom

A Skills section whose header carries a leading icon-font glyph is dropped entirely: **0 skills extracted, all skill bullets lost**.

On the résumé that surfaced this, the parser's markdown rendered the header as:

```
## ¥Skills
```

— a leading **U+00A5** (`¥`) mis-decoded icon-font glyph glued directly to `Skills` (no space; next codepoint is U+0053 `S`). The four skill bullets under it (a "Proficient in / Learning / Languages / soft-skills" list) never reach the parsed model.

Observed: `parse1.parsed.skills === []` (skills count **4 → 0**).

## Root cause

`matchSectionHeaderDetailed` in `src/lib/heuristics/regex.ts:422` normalizes the header text as:

```ts
const normalized = text.trim().toLowerCase().replace(/[:·•]+$/, "").trim();
```

It strips only **trailing** `[:·•]`. A **leading** non-letter glyph survives normalization, so `¥Skills` → `¥skills`, which is not in `SECTION_KEYWORDS` → the exact-match tier fails → the section is never routed → skills bucket stays empty.

Neither rescue tier recovers it:
- `rejoinSplitLetters` (the `E XPERIENCE` split-letter path) — doesn't apply to a glyph-prefixed single token.
- `matchAnchorFallback` (head-noun of qualified headers like "Relevant Experience") — the glyph is glued to the word, so there is no clean head-noun token.

Verified locally that stripping the leading run recovers the match:

```
"¥skills".replace(/^[^\p{L}\p{N}]+/u, "") === "skills"   // true → matches keyword
```

## Relationship to existing issues

Sibling of **#374** ("Technical Skills" section dropped when the header wraps across two visual lines): **same user-visible symptom** (skills section dropped) but a **different root cause and fix**. #374 is a multi-line-header rejoin problem; this is a leading-glyph normalization gap. Filed distinct; cross-link.

## Fix direction

Add a leading-glyph strip to the header normalizer in `matchSectionHeaderDetailed` (`regex.ts:422`), as an analog to the existing trailing `[:·•]+$` strip — a pre-normalization cleanup, **not** a new matching tier. Keep the existing exact-match → split-letter → anchor-fallback tiers intact and ordered as they are.

```ts
// strip a leading run of decorative/glyph chars before keyword match
const normalized = text
  .trim()
  .toLowerCase()
  .replace(/[:·•]+$/, "")
  .replace(/^[^\p{L}\p{N}]+/u, "")   // NEW: leading icon-font / symbol glyphs
  .trim();
```

**Guard:** only strip when a letter immediately follows the stripped run, so a decorative prefix on a real header is removed but a genuinely non-header line (e.g. a symbol-led prose/bullet line) cannot be coerced into a section. A leading bullet glyph is already handled separately via `LEADING_BULLET_RE`, so scope the new strip to header-candidate lines and confirm it does not fire on bullet content.

**Also consider** `matchSectionAnchorToken` (`regex.ts:396`) — the two-column anchor path shares the same trailing-only normalizer. It is column-gated (lower FP risk), so decide per its guard whether the same leading-glyph strip belongs there; if out of scope, note it explicitly.

## Implementation plan

1. `src/lib/heuristics/regex.ts:422` — add the guarded leading-glyph strip to the `matchSectionHeaderDetailed` normalizer (snippet above). Ensure the guard requires a following letter and does not fire on bullet lines already caught by `LEADING_BULLET_RE`.
2. Add a unit test next to the existing `matchSectionHeader` tests asserting `matchSectionHeader("¥Skills") === "skills"` (and a couple of glyph variants), plus a negative case proving a symbol-led non-header line still returns `null`.
3. Add a synthetic-persona PDF fixture (see below) and wire it into the corpus so the section is recognized end-to-end through `runCascade`.

## Acceptance criteria

- [ ] A header of the form `<glyph>Skills` (leading icon-font / symbol glyph, no space) is recognized as the skills section and its bullets are extracted (skills count > 0).
- [ ] `matchSectionHeader("¥Skills")` returns the skills section; a symbol-led **non-header** prose/bullet line still returns `null` (no new false-positive sections).
- [ ] Existing section-recognition behavior unchanged — `corpus.test.ts` and `corpus-roundtrip.test.ts` stay green; no `KNOWN_FAILURES` regressions.
- [ ] Regression covered by a new synthetic-persona fixture whose Skills header carries a leading glyph — fails before the fix, passes after.

## Synthetic repro fixture

Add a PDF fixture under `tests/fixtures/pdfs/<category>/` using a **synthetic persona only** (per `tests/fixtures/pdfs/README.md`): fake name, `@example.com` email, phone on a real area code with the `555` exchange and a `0100`–`0199` subscriber (e.g. `(312) 555-0123`). Its Skills section header must be prefixed with a leading symbol / icon glyph so it reproduces this category of miss. Verify PII-free with `pdftotext <file>.pdf - | head -40` before committing. Seeds a section-recognition / `corpus-roundtrip.test.ts` case.

s-annam:
## Summary

Adds `probe-skills`, the fourth `probe-*` parser probe (sibling of `probe-contact` / `probe-experience` / `probe-roundtrip`) — the triage tooling lane for the skills epic #415 — and lands the epic's two open children on top of it.

It runs the real `runCascade` parser over an arbitrary, uncommitted résumé PDF and prints the section router's **INPUT** (the header candidates it saw + the routed skills region) next to its **OUTPUT** (the parsed skills list), plus an independent header-recognition oracle. The verdict localizes a dropped skills section to its layer:

- `ok` — skills came through
- `EXTRACTION-MISS` — a skills region was routed but 0 skills parsed → bug in `extract/skills.ts`
- `HEADER-UNRECOGNIZED` — no region routed, but a skills-like header was rejected by the exact-match router (the #414 leading-glyph class, #374 two-line-wrap class); names *why*
- `NO-SKILLS-SECTION` — no region and no skills-like header candidate

**Tooling:**
- `src/lib/heuristics/probe-skills.test.ts` — the `RL_SKILLS_PDF`-gated dev harness. Inert in CI (skipped unless the env var is set), so it never reddens the suite.
- `.claude/skills/probe-skills/SKILL.md` — the skill, with the same PII guardrail as its siblings (real PDF never committed; skill values → gitignored `internal/skills/`).

**Fixes (skills header routing):**
- **#414** — a skills header carrying a leading icon-font glyph (`¥Skills`) was dropped. Guarded `LEADING_GLYPH_RE` strip in `matchSectionHeaderDetailed` + `matchSectionAnchorToken` (`regex.ts`); bullet content excluded via `LEADING_BULLET_RE`. Unit tests + new synthetic-persona fixture.
- **#374** — a `Technical Skills` header wrapped across two visual lines (`Technical\nSkills`) stranded the section on the markdown-anchored path. `isWrapHeaderShape` + `matchWrappedHeader` fold in `splitIntoSectionsWithMarkdown` (`sections.ts`), gated so a ≥2-token join can never match a bare single-word alias. Unit tests + golden rebake (`letter-spaced-name-heading`, skillsCount 0→13).

Closes #414 (leading decorative glyph `¥Skills` — guarded `LEADING_GLYPH_RE` strip; unit tests + synthetic fixture).
Closes #374 (two-line-wrapped `Technical`/`Skills` header — `matchWrappedHeader` fold; unit tests + golden rebake).
Refs #415 (skills epic — this PR ships the `probe-skills` tooling and closes both current children #414/#374; the epic stays open as the lane for future skills failure classes surfaced by the probe).

## Test plan

- [x] `npm run lint` clean
- [x] Harness inert (skipped) when `RL_SKILLS_PDF` unset — CI-safe
- [x] On a real résumé (local, uncommitted): correctly reports `HEADER-UNRECOGNIZED` with reason `leading-glyph prefix ("¥skills" → "skills")` for the #414 class
- [x] `corpus.test.ts` + `corpus-roundtrip.test.ts` green; only the one #374 golden moved; no `KNOWN_FAILURES` regression
- [x] `verify` CI check green

## Adversarial review

Bounded adversarial pass over the accumulated #414 + #374 diff (independent `ecc:typescript-reviewer`, opus), pre-PR. **1 round, converged clean — zero blocking findings.** ~35 adversarial inputs run through the real functions; no false-positive section or regression constructed.

- **#414 glyph strip** — no regression on letter-/digit-led headers; no catastrophic backtracking; circled digits (①) treated as `\p{N}` and left intact; anchor-fallback stays FP-safe because Guard 7 casing runs on the raw unstripped text.
- **#374 wrap-fold** — `prevAppended` bookkeeping traced exhaustively (no cross-section/empty pop/duplication); lowercase-prose, terminal-punctuation, non-alias, and post-header bare lines all correctly refuse to fold. A ≥2-token join can never match a bare single-word alias.
- **Golden rebake** — `letter-spaced-name-heading.expected.json` (skillsCount 0→13) is a legit #374 effect; all deltas skills-scoped; only golden that moved across the 44-fixture corpus; snapshot stays lossy.
- **Fixture PII** — new `pdflib-leading-glyph-skills-header.pdf` synthetic (fake name, `@example.com`, reserved-fictional phone, pdf-lib metadata only).

Surviving **nits** (non-blocking, human-eyeball):
1. `matchWrappedHeader` doc-comment overstated the gate as "exact alias" — **applied inline** (reworded to note the anchor-fallback tier also folds guarded qualified headers like `Relevant Experience`).
2. Text-path glyph strip is broader than the `¥` in the title (any symbol prefix on an exact alias resolves) — intentional, low-risk; stripped remainder must be exactly an alias, bullets excluded. No fix needed.

### Fallow (post-review)

fallow's PR-attributed complexity nits suppressed with rationale (`matchSectionHeaderDetailed` tiered-matcher branch from #414; the `probe-skills` monolithic diagnostic harness) — matching existing repo precedent. Remaining complexity findings (`classifyLine`, `splitIntoSections`, `orderItemsByColumn`) are pre-existing/inherited; the duplication clone groups are test-scaffold shared with the sibling probes by design (report-only).


Vaishnavi1709 (approved):
Reviewed under the resumelint pr-review gates. 0 blocking → approve.

## Gates

- **Fixture PII** — `pdflib-leading-glyph-skills-header.pdf` verified: Jordan Rivera / `jordan.rivera@example.com` / **(312) 555-0123** / Austin, TX. Metadata Creator = `pdf-lib` (no author leak). Phone hits the CLAUDE.md spec exactly (real 312 area code, `555` exchange, `0123` in reserved 0100–0199).
- **Design-system / style tokens** — N/A (no `src/components/**` changes).
- **SKILL.md as code** — flags all real; SKILL.md correctly delegates `gh issue create` to `create-gh-issue` rather than hand-rolling (PR #390 lesson applied). `internal/` confirmed gitignored.
- **#414 correctness** — `LEADING_GLYPH_RE` gated by `LEADING_BULLET_RE`; anchor-fallback Guard 7 casing runs on **raw** (unstripped) text, so the FP surface for `¥`-led prose stays closed. `¥E XPERIENCE → e xperience → experience` chain works.
- **#374 correctness** — `prevAppended` reset to `null` whenever a section opens, so a bare `Skills` right after an `Experience` header can't silently fold. Join must resolve via `matchSectionHeader` (multi-word alias or guarded anchor-fallback), which forecloses a coincidental ≥2-token match on a bare single-word alias. Boundary case pinned at `sections.test.ts` line 1199.

## Secondary (follow-up, not blocking)

**S1. `SKILL.md` line 34 understates the console PII surface.** The prose promises the console prints "only counts + the header text," but `probe-skills.test.ts:197-208` prints parsed skills values (`JSON.stringify(s)`), skills region text verbatim, and rejected header candidate text. The test file's own header comment (lines 22-28) is honest ("prints skill VALUES so the corruption is visible → scratch only"). Suggest aligning SKILL.md with that disclaimer so a user doesn't infer the console is safe to paste.

## Nits

- **N1.** `LEADING_GLYPH_RE` (`regex.ts:275`) has no max-length quantifier. The `normalized.length > 40` gate at line 463 catches the post-strip case, but a `{1,5}` bound would defend earlier. No known reproducer.
- **N2.** `isWrapHeaderShape` admits a bullet-led half (leading non-alpha exempt for `& Honors`). In practice `matchSectionHeader` on the joined string still rejects via `LEADING_BULLET_RE`, so no drift — a local bullet guard inside `isWrapHeaderShape` would be more readable but is not required for correctness.

Fixture PII, design-system, and style-token gates all clean; parser fixes are tightly guarded and well-tested.

Issue #321, opened 2026-07-03
Local-first storage foundation: IndexedDB module for resumes and jobs


s-annam:
## Problem

ResumeLint is fully stateless today: drop a PDF, see the score, refresh and it's gone. Users who iterate on their resume — or check it against multiple job postings — have no way to keep anything between visits. Everything should stay local: the product promise is that no resume bytes leave the browser, and persistence must not change that.

## Proposal

Add a client-side persistence foundation using **IndexedDB** (via the ~1KB [`idb`](https://github.com/jakearchibald/idb) wrapper), living in `src/lib/storage/`. This is infrastructure only — no user-facing surfaces yet. Follow-up issues build the resume library and job tracker on top of it.

### Scope

- **`src/lib/storage/` module** — typed async API over one IndexedDB database with two object stores:
  - `resumes` — raw PDF bytes as `Blob` (no base64 inflation), plus filename, timestamps, and a cached parse result so reloading a saved resume doesn't re-run the cascade.
  - `jobs` — small JSON records (see job-tracker follow-up for the shape). Created here so both stores version together under one schema migration path.
- **Durability**: request `navigator.storage.persist()` on first write; expose the returned grant state so UI can tell the user whether storage is best-effort or persistent.
- **Export / import**: JSON round-trip of everything (resume bytes base64-encoded only in the export file). This is the user's own backup path and the mitigation for browser eviction.
- **Data-wipe transparency copy**: browsers can evict site data under disk pressure, and Safari clears script-writable storage after 7 days without a visit. Surface this honestly wherever saved data is shown (exact placement lands with the UI issues; the copy and the `persisted` signal live here).
- **Schema versioning** from day one (IndexedDB `onupgradeneeded`), so later stores/fields don't strand existing users.

### Non-goals

- No UI surfaces (resume picker, job list) — follow-up issues.
- No sync, no accounts, no server. Local-first is the feature: your data stays on your machine, in your browser profile.

### Notes

- localStorage stays as-is for the existing `rl_*` UI flags — this module is for structured/binary data only.
- README Telemetry/storage section gains a paragraph documenting the IndexedDB database alongside the existing `rl_*` keys.

## Acceptance criteria

- [ ] `src/lib/storage/` exposes typed CRUD for `resumes` and `jobs` stores; unit tests cover round-trip of a `Blob` and schema upgrade from an empty/v0 database.
- [ ] `navigator.storage.persist()` requested; grant state queryable.
[...]

PR #426, opened 2026-07-09
feat(resume-library): save, reload, rename, delete resumes locally (#322)

### LINKED ISSUE #322: Resume library: save, reload, and manage multiple resumes locally
## Problem

Users iterating on their resume re-drop the same PDFs over and over. There's no way to keep more than one resume around — e.g. a general version and a role-tailored version — or to come back tomorrow and pick up where you left off.

## Proposal

A **resume library** built on the IndexedDB storage foundation: save a parsed resume locally, list saved resumes, reload one without re-uploading, delete ones you no longer need. All local — resume bytes never leave the browser.

### Scope

- **Save on parse**: after a successful parse, offer to save (filename + parsed snapshot + PDF bytes). Respect the existing drop/replace flows.
- **Library view**: list saved resumes (name, saved date, score at save time); pick one to load straight into the results view using the cached parse (no re-run unless the user asks).
- **Rename / delete** per entry.
- **Durability messaging**: show the storage-persistence state from the foundation module; if storage is best-effort, show the transparency copy (browser may evict; Safari clears after 7 days idle) with a one-click export as the backup path.
- **Multiple resumes are the point**: no artificial cap below what quota allows; show approximate space used.

### UI placement

Entry point on the landing view (e.g. "Saved resumes" beneath the drop zone when the library is non-empty) — exact design at implementation time, reusing existing design-system primitives per the 3-tier component rules.

### Non-goals

- No comparison view between saved resumes (possible follow-up).
- No sync/accounts — local to this browser profile, stated plainly in the UI.

## Acceptance criteria

- [ ] Save, list, load, rename, delete flows work end-to-end offline.
- [ ] Loading a saved resume restores the results view without re-uploading the file.
- [ ] Persistence state + eviction transparency copy visible in the library view; export reachable from there.
- [ ] All feature components under `src/components/features/`, composed from `@design-system` primitives; hooks for storage access in `src/hooks/`.

Depends on the IndexedDB storage foundation issue.

Samhit21:
## What & why

Closes #322. Builds on the #321 IndexedDB storage foundation.

Users iterating on a resume (or juggling a general + a role-tailored version) had no way to keep more than one around, or to come back tomorrow and pick up where they left off. This adds a **local-first resume library**: save a parsed resume, list saved resumes, reload one **without re-uploading**, rename/delete. All local — resume bytes never leave the browser.

## How it works

- **`src/lib/resume-library.ts`** — the domain layer between the parse pipeline and storage, and the **first in-repo consumer of `@/lib/storage`** (so this PR drops the temporary fallow `entry` registered in #321). It maps a parse ↔ a `resumes` record: the cached parse rides in the record's opaque `parse` slot and round-trips **losslessly via IndexedDB structured clone** (which preserves the `sections` Map), so a reload is exact; source bytes reload **byte-identically**.
- **`useResumeLibrary`** — reactive list, persistence-grant signal, approximate space used, `save`/`load`/`rename`/`remove`, and a JSON **export-backup** download.
- **`useResumeAnalysis.loadSavedResume()`** — hydrates the `done` state directly from a cached parse (same shape a live parse produces, minus the cascade run). Threaded through `useAnalyzedResume`.
- **Feature components** (design-system primitives only, per the 3-tier rules):
  - `ResumeLibrary` — the picker beneath the drop zone; persistence `StatusBadge` + eviction transparency copy + **Export backup** + approx space used. Self-hides when empty.
  - `ResumeLibraryEntry` — rename via `EditableField`, **confirm-gated delete** via `Dialog`.
  - `SaveResumeBar` — the save affordance under a result; a second save updates the same record rather than minting a duplicate.
- **App** wires the picker (landing view) and the save bar (done view).

## Privacy

No network calls introduced; nothing leaves the browser. Loading restores from the local cached parse; the eviction notice + export path make the local-only, best-effort nature honest (per #321).

## Tests (`fake-indexeddb`)

save → list → load → rename → delete; cached-parse (Map intact) + byte-identical bytes round-trip; DOCX no-source-bytes path; missing-id guard.

## Acceptance criteria

- [x] Save, list, load, rename, delete work end-to-end offline.
- [x] Loading a saved resume restores the results view without re-uploading.
- [x] Persistence state + eviction copy visible in the library view; export reachable there.
- [x] Feature components under `src/components/features/` composed from `@design-system`; storage hook in `src/hooks/`.
- [x] `npm run verify` green (typecheck, eslint, build; new tests pass).

## Notes for the reviewer

- **fallow complexity (report-only):** `App.tsx` is flagged CRITICAL, but that's **inherited** — App was already a large CRITICAL wiring component; this PR adds ~40 lines and the diff-scoped gate excludes it as inherited. `ResumeLibrary` (display conditionals) and `estimateStorageUsage` (guard-heavy, 0% tested) are minor and report-only.
- **Flaky full-suite run:** one unrelated pre-existing test (`render-roundtrip.repro.test.ts`) hit a 10s `beforeAll` timeout under machine load in one `verify` run; it passes in isolation (766ms) and doesn't touch storage. Typecheck, lint, build, and all storage/library tests are green.

### DIFF
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -11,15 +11,7 @@
   "entry": [
     "jd-spike.html",
     "eval-rewrite.html",
-    "parse-eval.html",
-
-    // Local-first storage foundation (#321) is infrastructure shipped AHEAD of
-    // its UI — the resume-library and job-tracker follow-ups are its first
-    // in-repo consumers. Until they land, the public barrel has no importer, so
-    // register it as an intentional root; without this fallow flags the whole
-    // module (barrel + its export/import/persist surface) as unreachable dead
-    // code. Remove this line once a follow-up imports `@/lib/storage`.
-    "src/lib/storage/index.ts"
+    "parse-eval.html"
   ],
 
   // `@design-tokens` is a Vite CSS-import alias (vite.config.ts resolve.alias),
--- a/README.md
+++ b/README.md
@@ -159,6 +159,10 @@ under disk pressure, and Safari clears script-writable storage after 7 days
 without a visit — so the module ships a JSON **export/import** backup path (resume
 bytes base64-encoded in the export file) as the user's own recovery route.
 
+The **resume library** (save a parsed resume, reload it later without re-uploading,
+rename/delete) is the first surface built on this store; it shows the persistence
+state and the eviction note inline, with the export backup reachable from there.
+
 ### GitHub star count
 
 The footer shows the live repo star count via an unauthenticated call to
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -8,7 +8,10 @@ import { ReconstructedResume } from "./components/features/ReconstructedResume.t
 import { AtsScoreReadout } from "./components/features/AtsScoreReadout.tsx";
 import { PageShell } from "./components/features/PageShell.tsx";
 import { ReplaceResumeDropOverlay } from "./components/features/ReplaceResumeDropOverlay.tsx";
+import { ResumeLibrary } from "./components/features/ResumeLibrary.tsx";
+import { SaveResumeBar } from "./components/features/SaveResumeBar.tsx";
 import { useAnalyzedResume } from "./hooks/useAnalyzedResume.ts";
+import { useResumeLibrary } from "./hooks/useResumeLibrary.ts";
 import { useReplaceResumeOnDrop } from "./hooks/useReplaceResumeOnDrop.ts";
 import { writeJdFitHandoff } from "./lib/jd-fit-handoff.ts";
 import { isScoreRevealed } from "./lib/contact.ts";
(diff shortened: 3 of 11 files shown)


s-annam (approved):
Clean, well-composed feature — domain layer / hook / UI split is right, reuses `Card`/`Button`/`StatusBadge`/`EditableField`/`Dialog` from `@design-system` (no hand-rolled primitives), delete is confirm-gated, and the storage round-trip is covered by real `fake-indexeddb` tests. Typecheck / lint / tests green; no fixtures touched. **Approving** — 0 blocking. One correctness issue and a couple of nits below, none merge-blocking.

## Secondary

**Reload → re-save mints a duplicate record.** `SaveResumeBar` keeps `savedId` in local state (starts `null`), but `LoadedDoneState` (`src/hooks/useResumeAnalysis.ts:116`) has no `id`, so `App.onLoadSavedResume` (`src/App.tsx`) drops `loaded.id` when hydrating the "done" state. After **Load**, `SaveResumeBar` mounts fresh with `savedId=null` — the button even reads "Save to library" rather than "Update saved copy" for a resume that came *from* the library — and `saveResume` mints a new UUID when `id` is omitted (`src/lib/storage/resumes.ts:29`). Net: load a saved resume, tweak it, hit Save → you get a second copy instead of updating the one you loaded.

Fix: thread the loaded id through so re-save updates in place. Add `libraryId?: string` to `LoadedDoneState` (and carry it on the `done` `ParseState`), pass `loaded.id` from `onLoadSavedResume`, and seed `SaveResumeBar`'s `savedId` from it. The primary loop (save → reload → view/download) is lossless and correct — this only bites the reload-then-edit-then-save path, hence Secondary.

## Nits

- **`ResumeLibrary.tsx` has no component test** — fallow flags it HIGH (CRAP 56, cyclomatic 7 at 0% coverage). Report-only, so it doesn't block CI, but a small render test (empty → hidden, populated → rows) would both drop the CRAP and lock in the self-hide behavior.
- **Save failure is silent.** `SaveResumeBar`'s `onClick={() => void save()}` has no catch; a quota/`QuotaExceededError` on `saveResume` rejects unhandled and the button just returns to "Save to library" with no signal to the user. A small error toast/inline note would close the loop (the library already surfaces the persistence/eviction state well elsewhere).

PR #431, opened 2026-07-10
fix(pr-review): derive review event from computed verdict, not a picked flag

s-annam:
## Summary

The `pr-review` skill's verdict rule (Step 5) counted only the current reviewer's own pass, so a literal read of *"0 Blocking → APPROVE"* could approve a PR while **another reviewer's standing `REQUEST_CHANGES` was still open** — the exact trap on #421, where a net-new pass found 0 new blocking but a prior blocker still reproduced.

This makes the review event a **pure function of the open-blocker set**, not a flag anyone picks:

- **Step 5** — verdict computed from every Blocking still reproducing on HEAD across **all** reviews (this run + any uncleared standing `REQUEST_CHANGES`). A net-new pass that adds 0 new Blocking can't reach `APPROVE` while a prior block stands → `COMMENT`.
- **Step 6** — verdict maps to the `gh pr review` event via a fixed `case` table; approve/request-changes/comment is the mechanical output of the rule, never a selected option. Exits if no verdict was computed.
- **Step 5/7** — documents the GitHub mechanic that an `APPROVE` does not dismiss another reviewer's `REQUEST_CHANGES`, and the report now says when the PR stays blocked despite an approve.

Docs-only change to `.claude/skills/pr-review/SKILL.md` — no product code, no runtime surface.

## Test plan

- [ ] `npm run test` green
- [ ] Skill body re-read: Step 5 → Step 6 event mapping is deterministic; no `--approve` selectable path remains

### DIFF
--- a/.claude/skills/pr-review/SKILL.md
+++ b/.claude/skills/pr-review/SKILL.md
@@ -198,15 +198,31 @@ For each finding give: the file:line, the concrete failure (inputs → wrong res
 and a **fix** — a diff or exact command, not just a complaint. Cite the source
 (`/code-review` vs which gate) only if it helps the author.
 
-### Step 5 — Pick the verdict (must match the findings)
+### Step 5 — Compute the verdict (a pure function of the open-blocker set)
 
-- **≥1 Blocking → `REQUEST_CHANGES`.**
-- **0 Blocking → `APPROVE`**, even if Secondary/Nits remain. Do **not** soft-gate on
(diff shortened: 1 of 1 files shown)


s-annam:
Closing — investigation showed the skill already emitted a correct APPROVED event on #421. The real block there is another reviewer's standing CHANGES_REQUESTED, not a bad approval. This change addressed the wrong problem.

PR #417, opened 2026-07-08
fix(parser): batch #392 — 5 education-section extraction bugs

Vaishnavi1709:
## Summary

Implements the [batch #392 epic](https://github.com/resumelint-org/resumelint/issues/392) — five education-parser extraction bugs on the same subsystem (`src/lib/heuristics/extract/education.ts` + one exported regex fragment). Fixed as one accumulated series so overlapping edits to `stripInstitutionDate` / `stripInstitutionLocation` / `educationFromChunk` don't collide.

### The five fixes

| Issue | Symptom | Fix |
|---|---|---|
| **#375** | Trailing ` \|` glued to institution when dates share the line (`State University \|`) | Extend `TRAILING_DATE_RE` with `(?:[\|·,]\s+)?` before the date; column separator peels alongside the date |
| **#367** | Coursework rendered as one blob with a redundant double label (`Coursework: Relevant Coursework: Data Structures, …`) | Peel a leading label via anchored `COURSEWORK_LABEL_RE`; split on commas so each course is a separate `coursework[]` entry; per-item Title-case guard preserved |
| **#366** | Institution + city joined by a single space on LaTeX 2-col (`Lakeside Institute of Technology Seattle, WA`) with `location=null` | Add a 1-space fallback (`SPACE1_US_RE`) after the strict 2+ space / comma / middot primaries; gated on the surviving institution prefix having ≥2 tokens so a single-word remainder (`Cornell Ithaca, NY`) stays glued |
| **#364** | One-line "Degree — Institution" stored verbatim as institution AND swallowed into `field` (`institution = B.S. in Computer Science — State University`, `field = Computer Science — State University`) | When `instLine === degreeLine`, split at em/en-dash and pick the `INSTITUTION_HINTS`-carrying part; re-parse `degree`/`field` from the remaining parts. Handles both "Degree — Institution" and reverse "Institution — Degree — Year" ordering |
| **#371** | `Dean's List 2015–2017` annotation range beats real graduation year for the entry's dates (`start=2015, end=2017` instead of just `end=2017`) | Extract the annotation keyword denylist into shared `EDUCATION_ANNOTATION_RE` (kept behavior-identical for `isInlineDatedProgram`); filter matching lines from the chunk before `parseEducationDates` runs |

### Regression tests

- Per-issue tests in `education.test.ts` (12 new cases across the 5 fixes)
[...]

Issue #392, opened 2026-07-07
Batch: parser education-section bugs


s-annam:
## Batch epic — Parser education-section bugs

Groups the small issues below for **single-batch implementation** by one person
via `/implement-batch`. This is a tracking issue; the work lives in the children.

**Why grouped:** all are education-section extraction bugs in the heuristics parser (openresume/regex tiers) — same subsystem, same fixtures, one reviewer context.

### Children
- [ ] #364 — degree line reused as institution when both share one line
- [ ] #366 — institution/location not split on single-space column gap (LaTeX two-column)
- [ ] #367 — coursework kept as one string with source label prefix, not split into course names
- [ ] #371 — Dean's List year range wrongly used as attendance dates when entry has a graduation year
- [ ] #375 — trailing "|" retained on institution when dates share the same line

**Implement with:** `/implement-batch <THIS#>`

Vaishnavi1709:
All 5 child issues (#364, #366, #367, #371, #375) resolved and closed via PR #417 (merged 2026-07-09). Closing tracking epic.

Issue #375, opened 2026-07-07
[parser] education — trailing "|" retained on institution when dates share the same line


Vaishnavi1709:
## Category
parser (Steps 1 & 3 F&F QC — provenance category)

## Fixture
`tests/fixtures/pdfs/unknown/letter-spaced-name-heading.pdf` (synthetic persona)

## Ground truth (`pdftotext`)
```
Bachelor of Science, Computer Science
State University | 2018 - 2022
```

Institution `State University` separated from dates `2018 - 2022` by ` | `.

## Observed
`probe-education`:
```
[0] degree="Bachelor of Science"  field="Computer Science"
    institution="State University |"      ← trailing " |" not peeled
    dates="2018" → "2022"
```

The date range is correctly extracted, but the ` |` separator remains glued to the institution string. Minor cleanup miss.

## Localized layer
`src/lib/heuristics/extract/education.ts:513-537` — `stripInstitutionDate`. The `TRAILING_DATE_RE` matches the date and strips it, but the ` |` separator that preceded the date is not part of the match, so it stays on the string. Then the final `stripInstitutionLocation` doesn't peel bare punctuation either.

## Suggested fix
Extend the `TRAILING_DATE_RE` to consume an optional leading `|` / `·` / `,` separator before the date shape, e.g. add `\\s*[|·,]?` before the `${DATE}` group. Alternatively, add a post-strip cleanup: `institution = institution.replace(/\\s*[|·,]\\s*$/, "").trim();`.

## Repro
```
RL_EDUCATION_PDF=/abs/path/to/tests/fixtures/pdfs/unknown/letter-spaced-name-heading.pdf \
  npx vitest run src/lib/heuristics/probe-education.test.ts
```

Issue #371, opened 2026-07-07
[parser] education — Dean's List year range wrongly used as attendance dates when entry has a graduation year


Vaishnavi1709:
## Category
parser (Steps 1 & 3 F&F QC — provenance category)

## Fixture
`tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-two-column.pdf` (synthetic persona)

## Ground truth (`pdftotext`)
```
Springfield State University
B.S. Computer Science, 2017
GPA: 3.7 · Dean's List 2015–2017
```

Single graduation year (2017) on the degree line. The `2015–2017` range in `Dean's List 2015–2017` is an **honors annotation**, not attendance dates.

## Observed
`probe-education`:
```
[0] institution="Springfield State University"  degree="B.S."  field="Computer Science"
    dates="2015" → "2017"  year="2017"
```

`start_date=2015, end_date=2017` — the Dean's List range was used as attendance dates. Correct behaviour would be `end_date=2017` (graduation) and no `start_date`.

## Localized layer
`src/lib/heuristics/extract/education.ts` — `parseEducationDates` runs on the whole joined chunk (`Springfield State University | B.S. Computer Science, 2017 | GPA: 3.7 · Dean's List 2015–2017`). `parseDateRange` finds the more-specific range (`2015–2017`) first and it wins over the bare graduation year (`2017`).

Note: `isInlineDatedProgram` already has a denylist for `dean'?s? list` (for entry-splitting decisions, ~line 256). The same denylist needs to gate date extraction — currently only the entry-boundary logic sees it.

## Suggested fix
Before `parseEducationDates`, filter the joined chunk to strip any line whose text matches the annotation denylist regex `/(dean'?s? list|awards?|honou?rs?|thesis|teaching assistant|research assistant|study abroad|coursework|scholarships?|fellowships?|cum laude|capstone|(?:senior|final|independent|group|team)\s+project)/i` — those lines carry annotation dates, never attendance dates. Then run `parseEducationDates` on the filtered chunk.

## Repro
```
RL_EDUCATION_PDF=/abs/path/to/tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-two-column.pdf \
  npx vitest run src/lib/heuristics/probe-education.test.ts
```

PR #433, opened 2026-07-10
feat(probe): add probe-education skill + RL_EDUCATION_PDF vitest harness

Vaishnavi1709:
## Summary

Adds `probe-education`, the fifth `probe-*` parser probe (siblings: `probe-contact` / `probe-experience` / `probe-roundtrip` / `probe-skills`) — the triage tooling lane for the **education-section extraction failure classes**.

It runs the real `runCascade` parser over an arbitrary, uncommitted résumé PDF and prints the section router's **INPUT** (the header candidates it saw + the routed education region + `DEGREE_RE` token count in the region) next to its **OUTPUT** (the parsed entries — degree / field / institution / location / dates / coursework), plus an independent header-recognition oracle and per-entry field-presence check. The verdict localizes a dropped / wrong / mis-chunked entry to its layer:

- `ok` — entries parsed, no count mismatch
- `EXTRACTION-MISS` — an education region **was** routed but 0 entries parsed → bug in `educationFromChunk` (chunker or field heuristics), not routing
- `HEADER-UNRECOGNIZED` — no region routed, but an education-like header was rejected by the exact-match router (leading-glyph class, out-of-alias wording); the reason names *why* the strict match failed
- `UNDER-CHUNKED` — region has more `DEGREE_RE` tokens than entries → the #364-class shape (two degrees collapsed into one entry, or a section-routing bug narrowed the region)
- `NO-EDUCATION-SECTION` — no routed region and no education-like header candidate

The `DEGREE_RE`-in-region count is a deliberate **lower** bound: a degree-less program entry (#238, e.g. `MIT Applied Data Science Program (2023)`) legitimately contributes an entry with no `DEGREE_RE` match, so `entries > regionDegrees` is intentionally **not** flagged — only the reverse direction is.

**Tooling:**
- `src/lib/heuristics/probe-education.test.ts` — the `RL_EDUCATION_PDF`-gated dev harness. Inert in CI (skipped unless the env var is set), so it never reddens the suite. Follows the monolithic-diagnostic-harness pattern established by `probe-skills` / `probe-experience` — a single linear read-out of parse → verdict.
- `.claude/skills/probe-education/SKILL.md` — the skill, with the same PII guardrail as its siblings (real PDF never committed; values → gitignored `internal/education/`) and the gated-auto-file workflow via the `create-gh-issue` skill.

Refs #364, #366, #367, #371, #375 (the education-parser bugs the initial version of this probe helped surface during the F&F QC — all closed in the #392 → #417 batch). This class of tooling stays as the triage lane for future education failure classes.

## Test plan

- [x] `npm run lint` clean
- [x] Harness inert (`1 test | 1 skipped`) when `RL_EDUCATION_PDF` unset — CI-safe
- [x] `npm run verify` green (fallow report-only warning noted below)
- [x] Spot-checked against 3 corpus fixtures with distinct verdicts:
  - `synthetic-degreeless-two-programs.pdf` → `ok (2 entries)` — the previous version's OVER-SPLIT false-positive on this shape is fixed by dropping the upper-bound check
  - `latex/multi-degree-coursework.pdf` → `ok (2 entries, 2 DEGREE_RE tokens)`
  - `google-docs/google-docs-skia-proxy-achievements-oneline.pdf` → `ok (1 entry)` — with the #364 fix landed on main, `institution="State University"` now reads clean
- [x] Bash / `gh` commands in SKILL.md validated (both blocks: the run invocation and the dedup search)

## Notes

**Fallow duplication warning (report-only, not blocking):** fallow reports 10 clone groups (~282 lines total, ~71 lines with `probe-skills.test.ts`) that this diff introduces with the sibling probe test files. Groups cover the `looseXHeaderReason` helper shape, the `describe.runIf` + score-compute scaffold, the markdown-header-candidate + orphan-block extraction, and the JSON write + console output template.

This mirrors the intentional monolithic-diagnostic-harness pattern the sibling probes already ship (matches the same "test-scaffold shared with the sibling probes by design" acknowledgement in #416 for probe-skills). Deferring extraction to a follow-up refactor PR that touches all `probe-*.test.ts` files together, so the shape of the shared helper module can be decided in one place rather than piecemeal.

### DIFF
--- a/.claude/skills/probe-education/SKILL.md
+++ b/.claude/skills/probe-education/SKILL.md
@@ -1,0 +1,207 @@
+---
+name: probe-education
+description: Extract + verify the EDUCATION section (entries, degree, field, institution, location, dates, coursework) of any résumé PDF — including a real, PII-bearing one — via the real parser, and localize a dropped/merged/under-chunked entry to the exact layer (header routing vs chunker vs field heuristic). Prints the header candidates the router saw and the education region it scanned next to the parsed entries, plus an independent header-recognition oracle that flags an education-like header the strict router rejected (leading-glyph, out-of-alias wording, two-line wrap) and a DEGREE_RE token count over the routed region as a lower-bound oracle for chunker under-segmentation. One of the `probe-*` parser probes (see also `probe-contact`, `probe-experience`, `probe-roundtrip`, `probe-skills`). Use when the user says "probe education", "/probe-education", "why is the degree/institution/coursework wrong", "education section isn't parsing", "why are two degrees showing as one entry", or hands you a résumé whose education entries parse wrong.
+---
+
+# Probe: Education
+
+> One of the `probe-*` parser-probe family (siblings: `probe-contact`,
+> `probe-experience`, `probe-roundtrip`, `probe-skills`). Type `probe-` to list
+> them all.
+
+Drop in **any** résumé PDF and see exactly what the parser reads for the
+education section — the parsed entries (degree, field, institution, location,
+dates, coursework), the education region the extractor scanned, and the header
+candidates the section router saw — and, when entries come back wrong, which
+layer is responsible. This is the tooling lane for the three education failure
+classes: a routing miss (`Academics` / `Qualifications` / `Education & Training`
+wording, leading decorative glyph, two-line-wrapped header), a chunker collapse
+(two degrees fold into one entry — the pattern that seeded #364 and its
+follow-ups), or a field heuristic drop (a specific field like `location` or
+`field` blanks even though the chunk looks right — the shape that seeded #366 /
+#367).
+
+It reuses a dev harness in `src/lib/heuristics/probe-education.test.ts` (the
+`RL_EDUCATION_PDF` block) — there is **no standalone script and you should not
+write one**: the pdfjs worker uses Vite's `?url` import, which resolves only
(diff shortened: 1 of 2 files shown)


s-annam (approved):
`probe-education` skill + `RL_EDUCATION_PDF` harness. Ran it end-to-end against a synthetic fixture — parses cleanly, verdict logic (`ok` / `EXTRACTION-MISS` / `HEADER-UNRECOGNIZED` / `UNDER-CHUNKED` / `NO-EDUCATION-SECTION`) fires correctly, JSON lands in the gitignored `internal/education/` dir. Typecheck clean, CI green, PII guardrails match the sibling probes. Approving.

## Nits (non-blocking)

- **`SKILL.md` + harness: "two-line wrap" is claimed but not detected.** The description (frontmatter + body L99/L124) lists two-line-wrapped headers as a rejected-header class the oracle flags, but `looseEducationReason` (`probe-education.test.ts:78`) only returns `leading-glyph prefix` or `contains education anchor token…`. Header candidates come from single markdown `#`-lines (`:157`), so a wrapped header's second line is never on a `#`-line and can't be seen. Either drop the two-line-wrap claim or note it as a known blind spot.
- **Micro:** `matchSectionHeader(h.text)` is computed twice per header (`:169` build, `:237` report). Negligible at this N; fold into the first map if touching it.

Nothing here blocks — clean addition that matches the accepted `probe-*` pattern.

PR #430, opened 2026-07-09
feat(export): Download PDF fidelity — team, unbold skills, auto-bold metrics, scheme-strip links (#425)

s-annam:
## Summary

Round-trip-safe subset of **#425** — brings the ATS **Download PDF** closer to a polished résumé **within** the deliberate two-line stacked entry shape. It does **not** restore the single-line header (that trips the #284 description-prose detector and drops the role on re-parse).

Standalone PR off `main` — **not** stacked on #421.

## Landed

- **Team on the org line** — experience/education header now renders `Company · Location · Team` (`exp.team` was parsed but previously dropped at model-build).
- **Skills regular weight** — `AtsEntry.headerBold=false` on the skills entry; every other entry header stays bold.
- **Auto-bold metrics in bullets** — new pure `autoBoldMetrics` helper + an inline bold-run renderer draws quantified metrics (`40%`, `$2M`, `18 engineers`, `6 weeks`, …) in the bold font. Emphasis delimiters are **Private-Use-Area sentinels (U+E000/U+E001)**, so literal `**` in a résumé bullet is inert and round-trips byte-identically.
- **Scheme-stripped contact links** — links display without `https://` (keeps `www.` — see below).

## Reuse consolidation

Deleted the dead, zero-caller `src/lib/score/auto-bold.ts` (+ test) in favor of the round-trip-safe `src/lib/pdf/auto-bold-metrics.ts` — one metric-bolding module, not two.

## Deferred (need an out-of-scope parser change — not filed as issues yet)

- **Right-aligned dates** (the issue's flagship / most-visible AC) — flush-right opens a >50pt same-`y` gap that the parser's embedded-column detector (`COLUMN_GAP_THRESHOLD`) reads as a column boundary, swapping title/company on dense/two-column fixtures (#298 signature). Empirically isolated: disabling only the right-align cleared all 3 regressing fixtures. Date stays glued to the org line until the column detector learns to exempt a trailing lone date-range.
- **Full `www.` strip** — the parser re-adds `https://` but never `www.` on re-parse, so a www-stripped display re-parses to a different `linkedin_url` (6 fixtures). Scheme-only strip is the safe subset.
- **Clickable link annotations** (marked optional in the issue).
[...]

PR #434, opened 2026-07-10
feat(export): finalize #425 deferred remainder — right-align dates, strip www, hyperlink contact links

### LINKED ISSUE #425: Download PDF formatting fidelity — right-align dates, strip URL scheme, unbold skills, restore headline, hyperlink links
## Summary

The **Download PDF** (the ATS-safe export from the parser-audit lane) renders a correctly-parsed résumé with noticeably degraded formatting compared to a polished professional résumé. Fonts now match (Poppins, per past work), but the **layout and text weighting** are off in several concrete, fixable ways: dates dumped inline instead of right-aligned, URLs shown with their raw `https://` scheme, the skills list rendered bold, the professional headline dropped, the role's team/division line dropped, no bold emphasis on quantified metrics in bullets, and links rendered as dead plain text.

This issue catalogs every gap, roots each one in code, and specifies the fixes — grouped so an implementer can land the high-value render/model changes without touching the parser.

### The hard constraint (read first)

The exporter deliberately uses a **two-line stacked entry shape** — the role **title** bold on line 1, then `Company · Location  Dates` (muted) on line 2 — because that is what the text-only round-trip parser is tuned to re-segment (`render-ats-pdf.ts` type-scale comment lines 43–56; `ats-resume-model.ts` lines 333–344). A polished résumé uses a **single line** (`Title · Company, Location …… right-aligned dates`). **Do NOT restore the single-line shape** — the #284 comment documents that a combined `Title · Company · Location` header trips the description-prose detector and drops the role on re-parse. Every fix below must preserve round-trip fidelity (`corpus-roundtrip.test.ts`, `render-roundtrip.repro.test.ts` stay green). We improve *within* the stacked shape.

Note the existing `render-ats-pdf.ts:43-56` comment: **font weight/size is already invisible to the round-trip parser** (`groupIntoLines` collapses per-glyph `fontName`/`fontSize` before re-segmentation). So every bold/weight change below is round-trip-neutral by construction — it changes the drawn font, never the extracted text.

### Target weight spec (bold vs regular)

The intended look for a polished résumé:

| Bold | Regular |
|---|---|
| Candidate name | Professional headline (under name) |
| Section headers (uppercase) | Contact line + online links |
| Role / degree **title** | Company · Team suffix on the header row |
| Achievement **label** (`Patent`, `Publication`, `Exit`) | Date range (right-aligned) |
| Inline **metric** runs inside bullets (`40%`, `$2M`, `18-engineer`) | Bullet body text (non-metric) |
| | **Skills list** |
| | Summary, achievement title + year, education institution |

### Observed vs expected (from a real comparison)

| Area | Polished résumé | ResumeLint export (current) |
|---|---|---|
| Dates | right-aligned to the right margin | inline after location: `Santa Clara, CA  01/2025 – Present` |
| Team/division | 3rd segment shown (`Globex · Toronto · Enterprise Platforms`) | dropped |
| Skills | regular weight, comma paragraph | **bold**, `·`-separated |
| Bullet metrics | `**2 AI-native products**`, `**~10%**` bold | all-plain, no emphasis |
| LinkedIn | `linkedin.com/in/annam` | `https://linkedin.com/in/annam` |
| Headline | `Engineering Lead` under the name (regular) | dropped entirely |
| Achievement links | `US10275736B1`, `5 publications` clickable | plain, dead text |
| Date dash | `01/2025 - Present` (hyphen) | `01/2025 – Present` (en dash) |

---

## Root-cause analysis

All line numbers are current as of this writing.

### 1. Dates are not right-aligned (HIGH — most visible) `src/lib/pdf/ats-resume-model.ts:361-370` glues the date onto the org string with a two-space gap:
```ts
const subLine = needsOrgSignature
  ? [org, dateRange].filter(Boolean).join(" · ")
  : [org, dateRange].filter(Boolean).join("  ");
```
`src/lib/pdf/render-ats-pdf.ts:624-628` draws that whole string left-to-right from the left margin. `Layout.drawText` only ever draws at a left `x` (`render-ats-pdf.ts:453-503`) — there is **no right-alignment primitive**, so the date lands right after the location instead of flush-right.

### 2. Role team/division dropped (MED — model, data already parsed) `ResumeExperience` **already carries `team?`** (`src/lib/score/types.ts:94`), and the parser populates it (the `Enterprise Platforms` / `Fiber Group (Globex Holdings)` / `Voice Assistant` / `Mail, Store, Cloud` segments). But the model's org line is built as company + location only — `src/lib/pdf/ats-resume-model.ts:351`:
```ts
const org = joinHeader([exp.company, exp.location], " · ");   // exp.team never included
```
So the team is silently dropped at model-build time. **This is a model fix, not a parser change** — the data is present. Round-trip caveat: the sub-line is the parser's date-anchor org line and is drawn `atomicSegments: true`; adding a third `· team` segment must be verified not to re-parse into the company/location fields (#298 signature). Guard with a round-trip test.

### 6. Professional headline dropped (MED — model + parser check) A polished résumé shows the headline (`Engineering Lead`) in **regular** weight under the name. The ATS model's `AtsContact` (`ats-resume-model.ts:43-61`) has no headline field, and the header draw (`render-ats-pdf.ts:537-552`) emits only name
+ contact parts. Check whether the parser already surfaces a headline/current- title (search `src/lib/heuristics/`, `src/lib/score/types.ts`). If present, plumb it through; if absent, split the parser half into a linked follow-up and land the model+render half behind the field.

### 9. Header alignment (design choice — call out, don't silently change) The source centers name + headline + contact; the export is left-aligned because `drawText` has no centered mode. Arguably a legitimate ATS choice (left-aligned parses cleanly). Left as an explicit decision — see the open question.

---

## Out of scope for this issue (parser-side data loss — file separately)

These are **parse** gaps, not render/model formatting, and should not block this issue:

- **Skill split on `&`:** `API Design & Development` became two skills, `API Design` + `& Development` — wrong before render (`parsed.skills`).
- **`Bachelor of Technology in …` → `…, …`:** the connective "in" is lost when degree + field are re-joined with `", "` (`ats-resume-model.ts:441`).

If we want these fixed, open a separate parser issue and cross-link it here.

---

## Implementation plan

Land in order; steps 1–4 are the high-value core.

### Step 1 — Right-aligned date primitive + team on the org line
- In `render-ats-pdf.ts`, add a helper to draw a string flush-right at the current `y`/page, `x = PAGE_WIDTH - MARGIN - font.widthOfTextAtSize(text, size)`. Keep it on the **same line** as the org so extracted text order stays `Company · Location    Dates` (anchor intact).
- Carry the date separately from the org rather than glued in — add `subLineDate?: string` to `AtsEntry`, and in `drawEntry` (`render-ats-pdf.ts:608-636`) draw the org left + date right on one `y`.
- Include `exp.team` in the org line: `joinHeader([exp.company, exp.location, exp.team], " · ")` at `ats-resume-model.ts:351`. Preserve the existing `needsOrgSignature` `" · "` tiebreak (`ats-resume-model.ts:362-369`) and add a round-trip test proving company/location/team re-segment correctly.

### Step 2 — Strip the URL scheme on contact links
- In `buildContact` (`ats-resume-model.ts:175-181`), map each pushed link through `formatLinkDisplay` (import from `../contact.ts`). If link annotations (Step 5) land, keep the **original** URL as the annotation `href` and show the shortened string as text.

### Step 3 — Skills list regular-weight
- Add optional `headerBold?: boolean` (default `true`) to `AtsEntry`, set it `false` on the skills entry (`ats-resume-model.ts:499-511`), honor it in `drawEntry` (`render-ats-pdf.ts:610-614`). Every other entry stays bold.

### Step 4 — Auto-bold metrics + inline bold runs in bullets
- Add a pure `autoBoldMetrics(text)` helper (new file under `src/lib/pdf/` or `src/lib/score/`) that wraps quantifiable metrics in emphasis markers: percentages (`40%`), dollar amounts (`$2M`, `$500K ARR`), multipliers (`2x`), scale-with-unit (`50K users`), headcounts (`18 engineers`, `200-person team`), durations (`6 weeks`). Idempotent; skips already-marked spans.
- Add an inline-run renderer in `render-ats-pdf.ts`: parse the bullet into `{ text, bold }` segments, then draw runs switching between `fonts.regular` and `fonts.bold`, preserving bold across wrapped lines. Strip the emphasis markers before drawing (drawn text carries no markers → round-trip unchanged).
- Wire it into `drawEntry`'s bullet loop (`render-ats-pdf.ts:630-635`).
- Round-trip check: a bullet with metrics must re-parse to the **same** text it does today (`render-roundtrip.repro.test.ts`).

### Step 5 — PDF link annotations for URLs (optional within this issue)
- Add a `drawLink` path in `Layout` that registers a URI link-annotation rect over the drawn span after drawing text. Wire achievement/project `fields.url` and profile URLs; visible text stays scheme-stripped (Step 2). Annotations live outside the content stream, so the text round-trip is unaffected.

### Step 6 — Headline under the name
- Add `headline?: string` to `AtsContact`, populate in `buildContact`, draw it **regular** weight under the name (`render-ats-pdf.ts:537-552`) at a size between `SIZE_NAME` and `SIZE_CONTACT`. If the parser doesn't surface a headline, land the field plumbing and split the parser half into a follow-up.

### Step 7 — (Optional) hyphen vs en-dash
- Only if we decide ASCII hyphen reads better. Changing `ats-resume-model.ts:231` alters round-trip anchor text — re-bake `*.expected.json` and confirm the range still splits. Droppable.

---

## Acceptance criteria

- [ ] Each experience/education entry's **date range is right-aligned** to the right content margin, on the same visual line as the org text.
- [ ] The role **team/division** (`exp.team`) appears on the org line when present (`Company · Location · Team`), and parse → export → re-parse re-segments company/location/team correctly (new round-trip test).
- [ ] The **skills** list renders in **regular** weight; all other entry headers stay bold.
- [ ] Quantified **metrics in bullets** render **bold** (percentages, `$`, `Nx`, headcounts, durations, scale-with-unit); non-metric bullet text stays regular; a metric bullet re-parses to identical text.
- [ ] Contact links display **without** `https?://` or a leading `www.`.
- [ ] The professional **headline** renders (regular) under the name when the parse provides it (or the field is plumbed and the parser half is a linked follow-up).
- [ ] (If Step 5 landed) achievement/project/profile URLs are **clickable**; the visible text is scheme-stripped.
- [ ] **Round-trip preserved:** `corpus-roundtrip.test.ts`, `render-roundtrip.repro.test.ts`, `export-layout-contract.test.ts` pass.
- [ ] `npm run verify` green.
- [ ] Unit tests for each change: right-align position + team on org line (`render-ats-pdf.test.ts`, `ats-resume-model.test.ts`), scheme strip, skills-not-bold, auto-bold-metrics (pure fn), inline-run bold, headline.

## Open question for the implementer

**Header alignment:** keep **left-aligned** (current; ATS-safe) or **center** the name + headline + contact block to match a polished résumé? Recommendation: keep left-aligned and close the visual gap via date right-alignment + team + headline
+ metric bold. Flag in the PR if you disagree.

## Reuse analysis

**Capability:** PDF draw-layout + text-weight changes (right-align, team segment, scheme-strip display, conditional bold, inline metric-bold runs, headline line) inside the existing single-column ATS draw engine. **Existing surfaces found:** `src/lib/pdf/render-ats-pdf.ts` (`Layout`, `drawText`, `drawEntry`) and `src/lib/pdf/ats-resume-model.ts` (`buildAtsResumeModel`, `buildContact`) own the whole export; `formatLinkDisplay` (`src/lib/contact.ts`) owns URL shortening; the auto-bold-metrics helper is the one genuinely new pure module (no existing owner). **Decision:** **extend** the render/model modules and reuse `formatLinkDisplay`; add one new pure helper for metric detection. No new component/panel/surface — no Reuse Gate concern.

s-annam:
## Summary

Finalizes the deferred remainder of #425 on top of the round-trip-safe subset already shipped in #430 — landing every remaining acceptance criterion, plus four Download-PDF fidelity fixes surfaced while validating the export against a real single-column résumé.

- **Right-aligned entry dates (flagship).** New flush-right render primitive; the date is carried apart from the org line (`headerLineDate` / `subLineDate`). Enabled by a core parser change: a shared `columnGapCuts()` helper in `sections.ts` that exempts a **trailing lone date-range run** from the column-gap split (reused by both `flush()` and `rowIsMultiColumn()`), so a flush-right date preserves the round-trip date-anchor instead of being torn into a phantom column band. Guards the #298 title↔company regression — **including the ≥2-adjacent-date-row case** that plain-`flush()` exemption missed.
- **Full `www.` strip on contact links** via a symmetric `normalizeUrl` canonicalization, so the exported display and the re-parse stay in round-trip sync.
- **Clickable URI link annotations** on the contact line; `linkHrefs` preserves the original scheme/`www.` as the click target (display stays shortened).
- New `isLoneDateRange` discriminator (reuses `DATE_RANGE_RE`); tests cover the multi-row exemption, flush-right position, link annotations, and the www round-trip.

Round-trip is preserved: `corpus-roundtrip`, `render-roundtrip.repro`, `export-layout-contract`, `corpus`, `sections` all green. `npm run verify` exit 0 (1942 passed, 7 skipped). Corpus snapshot schema bumped v3 → v4 and re-baked (headline field, below).

Closes #425

## Follow-on export fidelity fixes

Four defects found comparing the Download PDF against its source résumé; folded in here rather than deferred to new issues (the one-line experience header spawns one scoped parser follow-up, #436).

### Skills list wrapped before a leading connector glyph

A skills line that soft-wrapped **right before** a bare `&`/`+` connector (`…API Design` ⏎ `& Development`) split into two bogus skills (`API Design` + `& Development`). `isSoftWrapContinuation` only caught the break *after* the glyph (Condition A) and the comma-list wrap (Condition B); neither fired here. Added **Condition A′** — a next line leading with a bare `&`/`+` rejoins the previous line. Never fires for the list-bullet glyphs `-`/`–`/`•`/`*` (already routed to a new entry by `SKILLS_NEW_ENTRY_RE`). Regression test added.

### Dropped professional headline

The parser detected a standalone job-title tagline in the header block (the line a candidate places under their name) only to **reject** it as the name via `looksLikeTitle`, then discarded it. `current_title` is a different signal — derived from the most-recent experience role — so the header tagline never reached the Download PDF, even though `AtsContact.headline` and its renderer path already landed in #425.

- New `extractHeadline()` lifts the first title-keyword line in the profile block, above the contact cluster (name-skipped; a section header or the first email/phone/linkedin line ends the scan, so a title-keyword line deeper in the résumé is never mistaken for the headline).
- Plumbed `parsed.headline` + confidence through `openresume`; `buildContact` populates `contact.headline`; the renderer draws it regular-weight/muted under the name.
- **Round-trips:** the redrawn headline re-parses to the same value (name still wins the name slot, headline re-captured), so parse → export → re-parse is stable.
- Corpus snapshot schema bumped **v3 → v4** and re-baked; 8 fixtures carrying a genuine tagline gained the `headline` presence flag — no other drift. New `extract/name.test.ts` covers capture, name-skip, location-reject, contact-stop, section-stop, and the no-name path.

Header now renders the full left-aligned block instead of dropping the title:

```
<Name>
<Professional Headline>
<email> • <phone> • <location> • <linkedin>
```

Left-aligned single-column is the ATS-canonical layout (centred header blocks can be mis-read as a separate column by some parsers and buy no parse benefit), so alignment is unchanged — the fix is restoring the lost line, not restyling.

### Achievement headers bold only the type label

An achievement written `Type · description` (e.g. `<Type> · <description> · <year>`) was drawn with the **entire** header bold; a polished résumé bolds only the leading type label. `buildAchievementHeader` now wraps just the leading `" · "`-delimited segment in the renderer's PUA emphasis sentinels (display-only, stripped before drawing, so the round-trip text is unchanged) and drops the base line to regular weight; a type-less header stays fully bold, and a ≤28-char length guard avoids bolding a full prose clause that merely carries a `" · "`. The renderer's bullet run-draw was generalized (`drawRuns` gains an optional marker/color) into `drawHeaderRuns` so a header draws mixed bold/regular runs. Font weight is invisible to the text-only parser, so round-trip is unaffected. Tests: type-label bold, type-less-stays-bold, long-prose-not-a-type (model); emphasized header renders with no sentinel leakage (render).

### One-line experience header

Collapsed the experience header from the stacked two-line shape (title on its own bold line, `Company · Location  Dates` on a sub-line, #284/#298) onto **one line** — `Title · Company, Location · Team` with the date flush-right — to match the compact canonical résumé shape.

**Deliberate round-trip tradeoff (tracked as #436).** The text-only parser has no font signal, so it used the two-line *structure* to tell title from company; on one line that signal is gone, and fixtures with neutral / parenthetical / two-column company names re-parse title↔company-swapped or company-truncated. This is a look-over-fidelity choice for the reconstructed PDF. The follow-up #436 will teach the parser to disambiguate title/company on a single header line (`disambiguateCompanyTitle` + entry-block anchoring). Until then the regressions are baselined, **all referencing #436 so the round-trip ratchet forces their removal as #436 lands**:

- `corpus-roundtrip`: `experience` baselined on 14 fixtures.
- 3 repro assertions suspended / updated: `render-roundtrip.repro` #284 AC#3 and `render-roundtrip-year-only.repro` #358 are `it.skip`; `render-roundtrip-team.repro` #425 asserts the team as the trailing header segment (no longer an org sub-line).

The model builder simplifies (no more `needsOrgSignature` / `subLine` / `subLineDate` branching for experience); education keeps its stacked shape untouched. Existing `buildAtsResumeModel` unit tests updated to the one-line output.

## Adversarial review

An independent reviewer attacked the diff (2 rounds). Findings and resolutions:

- **BLOCKING (fixed).** The initial `flush()` lone-date exemption only defended a *single* isolated date row; two **adjacent** flush-right `Org …(gap) Date` rows (e.g. back-to-back title-less/bulletless entries) were mis-read by `reorderEmbeddedColumns` as a 2-row grid and column-reordered — tearing dates off their orgs (the exact #298 band regression). Confirmed live via `groupIntoLines`. **Fix:** extracted `columnGapCuts()` as the single chokepoint applying the date-exemption in *both* `flush()` and `rowIsMultiColumn()`, so date rails never form a grid run. New fixtures: two title-less roles, two degree-less programs (each date stays attached), plus a genuine coursework grid that still reorders. Re-verified CONVERGED.
- **SECONDARY (fixed).** Link annotation placed over wrong text (`indexOf` matched a website slug inside the email) → now tracks a monotonic search offset.
- **SECONDARY (fixed).** www-stripped href could 404 for a `www.`-only portfolio host → `linkHrefs` uses the original URL as the click target while display stays www-less; linkedin/github round-trip normalization untouched.
- **SECONDARY (addressed).** Season-led date exclusion in `isLoneDateRange` is retained and now fixture-anchored (the `openresume-laverne` honors rail depends on season rails staying split); comment states the complete justification.
- **NIT (fixed).** Tightened `isLoneDateRange` so a bare non-year numeric range (`5000 - 6000`) no longer matches.
- **Residual (non-blocking, pre-existing).** `reorderColumnRun`'s interior-gap bucketing doesn't apply the date-exemption, but the exporter never emits interior column gaps, so it can't fire on our own round-trip — third-party-PDF territory, no action for #425.

## Test plan

- [x] `npm run test` green (1942 passed, 7 skipped)
- [x] `npm run verify` exit 0
- [x] Manually verify a right-aligned date + clickable contact link + restored headline in `npm run preview`

### DIFF
--- a/src/lib/contact/url-utils.ts
+++ b/src/lib/contact/url-utils.ts
@@ -37,7 +37,12 @@ export function normalizeUrl(raw: string | undefined): string | undefined {
     .replace(/[,;.)]$/, "")
     .trim()
     .replace(/^(https?:\/\/)?www\./i, "$1");
-  if (/^https?:\/\//i.test(trimmed)) return trimmed;
+  // Preserve any explicit scheme unchanged — only default a bare host to https.
+  // Matching just `https?://` here would (a) not exist as a bug for http (it
+  // already round-trips) but (b) turn `ftp://foo` into `https://ftp://foo`.
+  // Guarding on the general scheme grammar keeps the module's round-trip promise
+  // for non-http(s) inputs too (Samhit review, PR #434).
+  if (/^[a-z][a-z0-9+.-]*:\/\//i.test(trimmed)) return trimmed;
   return `https://${trimmed}`;
 }
 
--- a/src/lib/contact/url-utils.ts
+++ b/src/lib/contact/url-utils.ts
@@ -19,11 +19,24 @@
 export const LINKEDIN_NONPROFILE_RE =
   /linkedin\.com\/(company|jobs|feed|school|learning|pulse|posts|groups|showcase|games|events|help|legal|search|signup|login|home)\b/i;
 
-/** Ensure a URL carries an `https://` scheme and drop a trailing sentence
- *  punctuation mark. Returns `undefined` for empty input. */
+/** Canonicalize a URL: ensure an `https://` scheme, drop a trailing sentence
+ *  punctuation mark, and strip a leading `www.` host prefix. Returns `undefined`
+ *  for empty input.
+ *
+ *  The `www.` strip (#425) makes the ATS-export round-trip symmetric: the
+ *  exporter shows link slugs `www.`-less (`formatLinkDisplay`), and the parser
+ *  can't recover a `www.` on re-parse — so canonicalizing it away HERE, on both
+ *  the original parse and the re-parse, means a `www.`-bearing source URL and
+ *  its `www.`-less exported display both normalize to the same value and the
+ *  `linkedin_url` round-trip holds. `www.` is a semantically inert host alias
+ *  (linkedin/github/etc. serve both), so dropping it loses nothing. Mirrors the
+ *  `www.` strip `urlSlug` already applies for identity comparison. */
 export function normalizeUrl(raw: string | undefined): string | undefined {
   if (!raw) return undefined;
-  const trimmed = raw.replace(/[,;.)]$/, "").trim();
+  const trimmed = raw
+    .replace(/[,;.)]$/, "")
+    .trim()
+    .replace(/^(https?:\/\/)?www\./i, "$1");
   if (/^https?:\/\//i.test(trimmed)) return trimmed;
   return `https://${trimmed}`;
 }
--- a/src/lib/heuristics/extract/name.ts
+++ b/src/lib/heuristics/extract/name.ts
@@ -349,6 +349,31 @@ export function extractName(
  * parse → export → re-parse round-trip stable: the redrawn headline re-parses
  * back to the same value.
  */
+/** True when `text` IS the candidate's name (so it is not the headline). Compares
+ *  case-insensitively AND treats a line the name CONTAINS as the name, so a
+ *  normalized/merged name (a stacked-name second line, an occupational-word
+ *  surname) can't leak through a byte-exact `===` and be returned as the headline
+ *  (full_name "John Engineer" → headline "Engineer") (Samhit review, PR #434). */
+function isNameLine(text: string, name: string | undefined): boolean {
+  if (!name) return false;
+  const nName = name.trim().toLowerCase();
+  const nText = text.toLowerCase();
+  return nName === nText || nName.includes(nText);
+}
+
+/** True when `text` begins the contact cluster (email / phone / linkedin / any
+ *  URL) — the header block is over, so nothing at or below it is the headline.
+ *  Resets `lastIndex` defensively: these constants may carry a `g` flag whose
+ *  test() mutates state (mirrors `findContactClusterY`). */
+function startsContactCluster(text: string): boolean {
+  const hit =
+    EMAIL_RE.test(text) || PHONE_RE.test(text) || LINKEDIN_RE.test(text);
+  EMAIL_RE.lastIndex = 0;
+  PHONE_RE.lastIndex = 0;
+  LINKEDIN_RE.lastIndex = 0;
+  return hit || text.includes("@") || /https?:\/\//i.test(text);
+}
+
 export function extractHeadline(
   profile: PdfSection,
   name: string | undefined,
@@ -357,18 +382,9 @@ export function extractHeadline(
   for (let i = 0; i < scan; i++) {
     const text = profile.lines[i].text.trim();
     if (!text) continue;
-    if (name && text === name) continue; // the name is not its own headline
+    if (isNameLine(text, name)) continue; // the name is not its own headline
     if (matchSectionHeader(text)) break; // reached a section — header block over
-    // The contact cluster begins — anything at or below it is not the headline.
-    // Reset lastIndex defensively: these constants may carry a `g` flag whose
-    // test() mutates state (mirrors `findContactClusterY`).
-    if (EMAIL_RE.test(text) || PHONE_RE.test(text) || LINKEDIN_RE.test(text)) {
-      EMAIL_RE.lastIndex = 0;
-      PHONE_RE.lastIndex = 0;
-      LINKEDIN_RE.lastIndex = 0;
-      break;
-    }
-    if (text.includes("@") || /https?:\/\//i.test(text)) break;
+    if (startsContactCluster(text)) break; // contact cluster — header block over
     if (text.length > 60) continue;
     if (!looksLikeTitle(text)) continue;
     return { value: text, confidence: 0.7 };
--- a/src/lib/heuristics/extract/name.ts
+++ b/src/lib/heuristics/extract/name.ts
@@ -322,3 +322,56 @@ export function extractName(
   if (!best) return { confidence: 0 };
   return { value: best.text, confidence: Math.min(best.score, 1) };
 }
+
+// ── Headline ──────────────────────────────────────────────────────────────────
+
+/**
+ * The professional headline the candidate placed standalone under their name in
+ * the header block ("Engineering Lead", "Senior Product Designer"). This is the
+ * exact line `extractName` works to REJECT as the name (via `looksLikeTitle`) —
+ * here we keep it instead of discarding it, so the ATS export can redraw it
+ * under the name rather than silently dropping it (#425 follow-up).
+ *
+ * It is distinct from `current_title`, which is derived from the most-recent
+ * experience role; a header tagline is the candidate's own framing and may
+ * differ. We surface whatever the header carries, faithfully.
+ *
+ * Conservative by construction — only a genuine title-keyword line in the top of
+ * the profile block, above the contact cluster, qualifies:
+ *   - the name line is skipped (it is not the headline);
+ *   - a section header ("SUMMARY") or the first contact-bearing line
+ *     (email / phone / linkedin / any URL) ends the scan — the header block is
+ *     over, so a title-keyword line deeper in the résumé can never be mistaken
+ *     for the headline;
+ *   - the line must pass `looksLikeTitle` — a plain name or a location line
+ *     never qualifies, so résumés with no headline populate nothing.
+ * This symmetry with the exported layout (name, headline, contact) keeps the
+ * parse → export → re-parse round-trip stable: the redrawn headline re-parses
+ * back to the same value.
+ */
+export function extractHeadline(
+  profile: PdfSection,
+  name: string | undefined,
+): { value?: string; confidence: number } {
+  const scan = Math.min(profile.lines.length, 5);
+  for (let i = 0; i < scan; i++) {
+    const text = profile.lines[i].text.trim();
+    if (!text) continue;
+    if (name && text === name) continue; // the name is not its own headline
+    if (matchSectionHeader(text)) break; // reached a section — header block over
+    // The contact cluster begins — anything at or below it is not the headline.
+    // Reset lastIndex defensively: these constants may carry a `g` flag whose
+    // test() mutates state (mirrors `findContactClusterY`).
+    if (EMAIL_RE.test(text) || PHONE_RE.test(text) || LINKEDIN_RE.test(text)) {
+      EMAIL_RE.lastIndex = 0;
+      PHONE_RE.lastIndex = 0;
+      LINKEDIN_RE.lastIndex = 0;
+      break;
+    }
+    if (text.includes("@") || /https?:\/\//i.test(text)) break;
+    if (text.length > 60) continue;
+    if (!looksLikeTitle(text)) continue;
+    return { value: text, confidence: 0.7 };
+  }
+  return { confidence: 0 };
+}
--- a/src/lib/heuristics/extract/skills.ts
+++ b/src/lib/heuristics/extract/skills.ts
@@ -256,8 +256,21 @@ function isSoftWrapContinuation(pending: string, nextText: string): boolean {
   // wrap lands before it and the tail carries no comma, Condition B can't fire
   // either, so the fragment would wrongly become its own skill (`& Development`).
   // Only `&`/`+` — never the list-bullet glyphs `-`/`–`/`•`/`*`, which
-  // SKILLS_NEW_ENTRY_RE already routed to `return false` above — trigger this.
-  if (/^[&+]\s/.test(nextText)) return true;
+  // SKILLS_NEW_ENTRY_RE already routed to `return false` above.
+  //
+  // `&` is unambiguously a connector — fire always. `+` is genuinely ambiguous:
+  // SKILLS_NEW_ENTRY_RE does NOT list it, so it is used BOTH as a connector
+  // ("…CI" ⏎ "+ CD") and as a list BULLET ("Python, Java" ⏎ "+ Ruby on Rails"),
+  // and firing on the latter merges a bogus "Java + Ruby on Rails" skill (Samhit
+  // review, PR #434). Disambiguate on the CONTINUATION: a `+`-connector tail
+  // completes a compound and is a single short token ("CD"); a `+`-bulleted new
+  // item is a multi-word phrase ("Ruby on Rails"). So rejoin a leading `+` only
+  // when its continuation is a single word.
+  const lead = /^([&+])\s+(\S.*)$/.exec(nextText);
+  if (lead) {
+    if (lead[1] === "&") return true;
+    if (!/\s/.test(lead[2].trim())) return true;
+  }
 
   // Condition A: pending ends with an explicit continuation glyph — the glyph
   // must be its OWN whitespace-separated token ("Hiring &", "Data -"), not the
--- a/src/lib/heuristics/extract/skills.ts
+++ b/src/lib/heuristics/extract/skills.ts
@@ -249,6 +249,16 @@ function isSoftWrapContinuation(pending: string, nextText: string): boolean {
   // Always skip if the next line starts a new sub-list (label or bullet).
   if (SKILLS_NEW_ENTRY_RE.test(nextText)) return false;
 
+  // Condition A′ (symmetric to A): the NEXT line LEADS with a bare connector
+  // glyph — a soft-wrap that broke a skill/list immediately BEFORE the connector
+  // ("…API Design" ⏎ "& Development", "…CI" ⏎ "+ CD"). Condition A only catches
+  // the break AFTER the glyph ("Hiring &" ⏎ "Talent Acquisition"); when the
+  // wrap lands before it and the tail carries no comma, Condition B can't fire
+  // either, so the fragment would wrongly become its own skill (`& Development`).
+  // Only `&`/`+` — never the list-bullet glyphs `-`/`–`/`•`/`*`, which
+  // SKILLS_NEW_ENTRY_RE already routed to `return false` above — trigger this.
+  if (/^[&+]\s/.test(nextText)) return true;
+
   // Condition A: pending ends with an explicit continuation glyph — the glyph
   // must be its OWN whitespace-separated token ("Hiring &", "Data -"), not the
   // tail of a compact skill name like "C++" or "PHP7+" (#301: our own
--- a/src/lib/pdf/render-ats-pdf.ts
+++ b/src/lib/pdf/render-ats-pdf.ts
@@ -660,18 +660,32 @@ class Layout {
   }
 
   /**
-   * Draw a bullet as a sequence of `{ text, bold }` runs with word-level
-   * wrapping. A "word" is a run of non-whitespace that may span a bold→regular
-   * boundary (so mid-word emphasis draws correctly); words are separated by a
-   * single space. The bullet marker leads the first line; continuation lines
-   * use `hangingIndent`. Bold is preserved across wraps because it is tracked
-   * per chunk, not per line.
+   * Draw a header line as bold/regular runs (#425 — achievement "type" labels).
+   * A leading substring wrapped in the sentinel emphasis markers draws bold, the
+   * rest regular; the sentinels are stripped, so the round-trip text is
+   * unchanged. Same word-wrapping engine as `drawBullet`, but with no bullet
+   * marker and drawn at the header color/size.
+   */
+  drawHeaderRuns(text: string, size: number) {
+    this.drawRuns(parseBoldRuns(text), size, 0, { marker: "", color: this.black });
+  }
+
+  /**
+   * Draw a sequence of `{ text, bold }` runs with word-level wrapping. A "word"
+   * is a run of non-whitespace that may span a bold→regular boundary (so mid-
+   * word emphasis draws correctly); words are separated by a single space. A
+   * leading `marker` (the bullet "• " by default, "" for a header) leads the
+   * first line; continuation lines use `hangingIndent`. Bold is preserved across
+   * wraps because it is tracked per chunk, not per line.
    */
   private drawRuns(
     runs: Array<{ text: string; bold: boolean }>,
     size: number,
     hangingIndent: number,
+    opts: { marker?: string; color?: RGB } = {},
   ) {
+    const marker = opts.marker ?? BULLET_MARKER;
+    const color = opts.color ?? this.black;
     const words = groupRunsIntoWords(
       runs,
       size,
@@ -682,18 +696,22 @@ class Layout {
     const wordWidth = (w: WordChunk[]) =>
       w.reduce((sum, c) => sum + c.width, 0);
     const space = this.fonts.regular.widthOfTextAtSize(" ", size);
-    const markerWidth = this.fonts.regular.widthOfTextAtSize(BULLET_MARKER, size);
+    const markerWidth = marker
+      ? this.fonts.regular.widthOfTextAtSize(marker, size)
+      : 0;
     const rightEdge = PAGE_WIDTH - MARGIN;
     const lineHeight = size * LINE_GAP;
 
     this.ensure(lineHeight);
-    this.page.drawText(BULLET_MARKER, {
-      x: MARGIN,
-      y: this.y - size,
-      size,
-      font: this.fonts.regular,
-      color: this.black,
-    });
+    if (marker) {
+      this.page.drawText(marker, {
+        x: MARGIN,
+        y: this.y - size,
+        size,
+        font: this.fonts.regular,
+        color,
+      });
+    }
     let x = MARGIN + markerWidth;
     let atLineStart = true;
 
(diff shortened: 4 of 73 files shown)


s-annam:
Folded in one extra fix (`c158fd3`) surfaced by the #425 canonical example (Cisco eng-lead résumé): a skills list wrapped by pdfjs right before a connector glyph — `…API Design` ⏎ `& Development` — was split into two bogus skills instead of `API Design & Development`.

`isSoftWrapContinuation` only rejoined when the *pending* line ended with a connector (`Hiring &` ⏎ `Talent Acquisition`); added the symmetric case for a *next* line leading with a bare `&`/`+ `. Pre-existing parse bug, not introduced by the export work. +1 regression test; full `npm run verify` green (1904 passed).

src/lib/pdf/render-ats-pdf.ts
```diff
@@ -539,16 +581,62 @@ class Layout {
...
         color,
       });
+      if (i === 0) {
+        // Flush-right date tail on the first line's baseline (#425), right-
+        // aligned to the content margin and drawn regular-weight/muted.
+        if (opts.rightText) {
+          const rSize = opts.rightSize ?? size;
+          const rFont = this.fonts.regular;
+          const rValue = this.sanitize ? toWinAnsi(opts.rightText) : opts.rightText;
+          const rX = PAGE_WIDTH - MARGIN - rFont.widthOfTextAtSize(rValue, rSize);
```
Samhit21, line 603: [Suggestion]: Flush-right date can overlap a long one-line header — the wrap budget reserves no space for this date rail.

This date is drawn at `x = PAGE_WIDTH - MARGIN - dateWidth`, but the left text was wrapped with `maxWidth = CONTENT_WIDTH - (x - MARGIN)` (the `drawText` line above), which lets it pack to the full content width (right edge ≈558pt). So a long one-line `Title · Company, Location · Team` header (the flagship new shape) whose first line extends past `558 - dateWidth` renders on top of the date. The parser ignores visual position, so round-trip is unaffected — but the exported PDF, which is the entire point of this fidelity work, shows overlapping text. The manual preview check likely used a short header.

Fix: when `rightText` is set, shrink the first-line wrap budget by `rFont.widthOfTextAtSize(rValue, rSize)` plus a gap, so the left text wraps before the date rail (and truncate if the header alone still exceeds it).

src/lib/heuristics/extract/skills.ts
```diff
@@ -249,6 +249,16 @@ function isSoftWrapContinuation(pending: string, nextText: string): boolean {
...
 
+  // Condition A′ (symmetric to A): the NEXT line LEADS with a bare connector
+  // glyph — a soft-wrap that broke a skill/list immediately BEFORE the connector
+  // ("…API Design" ⏎ "& Development", "…CI" ⏎ "+ CD"). Condition A only catches
+  // the break AFTER the glyph ("Hiring &" ⏎ "Talent Acquisition"); when the
+  // wrap lands before it and the tail carries no comma, Condition B can't fire
+  // either, so the fragment would wrongly become its own skill (`& Development`).
+  // Only `&`/`+` — never the list-bullet glyphs `-`/`–`/`•`/`*`, which
+  // SKILLS_NEW_ENTRY_RE already routed to `return false` above — trigger this.
+  if (/^[&+]\s/.test(nextText)) return true;
```
Samhit21: [Suggestion]: Condition A′ rejoins a line that leads with `+ ` used as a bullet, not a connector.

`SKILLS_NEW_ENTRY_RE` (line 228) recognizes bullets `[•\-–*]` but **not** `+`. So a skills list written with `+` bullets —

    Python, Java
    + Ruby on Rails

— skips the new-entry gate and hits `/^[&+]\s/` here, merging "+ Ruby on Rails" into the previous line as one bogus skill. The comment just above says the list-bullet glyphs "already routed to `return false`," but `+` is not in that set, so the stated invariant has a hole — and it is genuinely ambiguous with the intended "…CI" then "+ CD" continuation you are targeting.

Disambiguate on the pending line: only rejoin on a leading `+`/`&` when the pending line looks like an *incomplete* fragment (does not already end a complete entry), so a `+`-bulleted new item is not absorbed.

src/lib/heuristics/extract/name.ts
```diff
@@ -322,3 +322,56 @@ export function extractName(
...
+ */
+export function extractHeadline(
+  profile: PdfSection,
+  name: string | undefined,
+): { value?: string; confidence: number } {
+  const scan = Math.min(profile.lines.length, 5);
+  for (let i = 0; i < scan; i++) {
+    const text = profile.lines[i].text.trim();
+    if (!text) continue;
+    if (name && text === name) continue; // the name is not its own headline
```
Samhit21: [Suggestion]: `extractHeadline` name-skip is a brittle exact-match — a name word can leak as the headline. (Narrow trigger.)

`text === name` only skips the header line when it *byte-matches* the extracted name. If `extractName` returns a name that is normalized (case/whitespace) or merged from a two-line stack differently from the raw line, that line is not skipped; if it then passes `looksLikeTitle` (e.g. the second line of a stacked name, or an occupational-word surname), it is returned as the headline → `full_name="John Engineer", headline="Engineer"`.

The trigger is narrow, but it is new code and cheap to harden: compare case-insensitively and/or skip when the extracted name contains the line (`name.includes(text)`), so a subset-of-the-name line cannot become a headline.

src/lib/contact/url-utils.ts
```diff
@@ -19,11 +19,24 @@
...
+ *  `www.` strip `urlSlug` already applies for identity comparison. */
 export function normalizeUrl(raw: string | undefined): string | undefined {
   if (!raw) return undefined;
-  const trimmed = raw.replace(/[,;.)]$/, "").trim();
+  const trimmed = raw
+    .replace(/[,;.)]$/, "")
+    .trim()
+    .replace(/^(https?:\/\/)?www\./i, "$1");
   if (/^https?:\/\//i.test(trimmed)) return trimmed;
   return `https://${trimmed}`;
```
Samhit21, line 46: [Nit]: `normalizeUrl` force-prepends `https://` on any unrecognized scheme, breaking the symmetry it claims.

The fallback adds `https://` whenever the input lacks an http(s) scheme. Two effects: (a) an `http://`-only host round-trips to `https://` — the exported display drops the scheme, so re-parse assumes https (the symmetry test at url-utils.test.ts:37 only covers https); (b) a non-http scheme like `ftp://foo` becomes `https://ftp://foo`. Low impact — LinkedIn/GitHub are https and the click href uses the original URL — but the module explicitly promises round-trip symmetry.

Guard: if `trimmed` already matches `^[a-z][a-z0-9+.-]*://`, return it unchanged instead of prefixing.

Samhit21 (commented):
## PR Review: finalize #425 deferred remainder — right-align dates, strip www, hyperlink contact links

### Summary
Large, carefully-built finalizer (1406 additions, ~50 files are fixture snapshot bumps). The adversarial pass is genuinely thorough — the `columnGapCuts()` date-exemption chokepoint (shared by `flush()` and `rowIsMultiColumn()`) correctly defends the ≥2-adjacent-date-row #298 band regression, and the one-line-header round-trip tradeoff is deferred honestly: suspended repros are `it.skip` with `SUSPENDED (#436)` + un-skip notes, and the corpus-roundtrip baseline is a documented shrink-only ratchet, so the regressions are force-removable when #436 lands. `Closes #425` is legitimate. CI green (verify / fallow / Pages). I did not re-flag the deliberate #436 tradeoff. The advanced-security bot `extractHeadline` cognitive-complexity flag (name.ts) is still unaddressed.

### Findings
1. **[Suggestion] render-ats-pdf.ts** — flush-right date overlaps a long one-line header; the wrap budget reserves no space for the date rail. Visual-only (round-trip unaffected), but it is the flagship feature.
2. **[Suggestion] skills.ts:260** — Condition A′ rejoins a `+ `-led line even when `+` is a bullet marker (`+` is not in `SKILLS_NEW_ENTRY_RE`), merging bulleted skills; contradicts the adjacent comment "bullets already excluded" claim.
3. **[Suggestion] name.ts:360** — `extractHeadline` skips the name line via brittle `text === name`; a normalized/merged name whose line passes `looksLikeTitle` leaks a name word as the headline (narrow trigger).
4. **[Nit] url-utils.ts:41** — `normalizeUrl` forces `https://` on any unrecognized scheme (`http`-only → https on re-parse; `ftp://` → `https://ftp://`), a real asymmetry in a module that claims symmetry.

### Verdict
**Action: COMMENT**

src/lib/heuristics/extract/name.ts
```diff
@@ -322,3 +322,56 @@
...
+ *     (email / phone / linkedin / any URL) ends the scan — the header block is
+ *     over, so a title-keyword line deeper in the résumé can never be mistaken
+ *     for the headline;
+ *   - the line must pass `looksLikeTitle` — a plain name or a location line
+ *     never qualifies, so résumés with no headline populate nothing.
+ * This symmetry with the exported layout (name, headline, contact) keeps the
+ * parse → export → re-parse round-trip stable: the redrawn headline re-parses
+ * back to the same value.
+ */
+export function extractHeadline(
```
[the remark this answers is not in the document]
s-annam: Fixed in `2551349` — refactored `extractHeadline` into two named helpers (`isNameLine` / `startsContactCluster`), which also hardens the name-skip per Samhit's adjacent comment. Cognitive complexity is back under the fallow threshold (no longer in the diff findings).

src/lib/pdf/render-ats-pdf.ts
s-annam: Real overlap, confirmed — but every text-altering remedy (shrink the wrap budget, or truncate) regresses the parse→export→re-parse round-trip: two corpus fixtures (`student-projects-activities-singlecol`, `google-docs-skia-proxy-additional-skills`) drop `role[0].company` because the parser reads company off the **single-line** `·`-header, and wrapping/truncating splits or alters it. The overlap is purely visual (the parser ignores glyph position — round-trip is fine with it). Durable fix pairs a **font auto-fit** (keep one line, shrink to clear the rail — round-trip-safe) with a parser change to rejoin wrapped headers. Deferred to #437; leaving this thread open.

src/lib/heuristics/extract/skills.ts
s-annam: Fixed in `2551349`. `&` stays an unconditional connector; a leading `+` now rejoins **only when its continuation is a single word** — so a `+`-connector tail (`…CI` ⏎ `+ CD`) still merges, but a `+`-bulleted new item (`Python, Java` ⏎ `+ Ruby on Rails`, a multi-word phrase) is no longer absorbed. Existing `& Development` connector test still passes.

src/lib/heuristics/extract/name.ts
s-annam: Fixed in `2551349`. The name-skip now compares case-insensitively **and** skips a line the extracted name contains (`nName.includes(nText)`), so a normalized/merged name or an occupational-word subset can't leak through the byte-exact `===` and become the headline. Pulled into an `isNameLine` helper.

src/lib/contact/url-utils.ts
s-annam: Fixed in `2551349`. `normalizeUrl` now preserves any explicit scheme by guarding the general scheme grammar (`^[a-z][a-z0-9+.-]*://`), so `ftp://foo` no longer becomes `https://ftp://foo` and the module keeps its round-trip promise for non-http(s) inputs.

s-annam:
### Review comments addressed — ready for re-review (`2551349`)

@Samhit21 thanks for the review. Summary of this pass:

**Fixed & resolved:**
- **skills `+`/`&` continuation** — `&` stays an unconditional connector; a leading `+` now rejoins only when its tail is a single word, so a `+`-bulleted new item (`Python, Java` ⏎ `+ Ruby on Rails`) is no longer absorbed, while the `+ CD` connector case still merges. Existing `& Development` test stays green.
- **headline name-skip** — compares case-insensitively **and** skips a line the extracted name contains, so a normalized/merged name can't leak through the byte-exact `===` and become the headline. Pulled into an `isNameLine` helper.
- **`normalizeUrl` scheme** — preserves any explicit scheme (`^[a-z][a-z0-9+.-]*://`), so `ftp://foo` no longer becomes `https://ftp://foo`.
- **fallow `extractHeadline` complexity** — refactor into `isNameLine`/`startsContactCluster` dropped it back under threshold.

**Deferred → #437 (thread left open):**
- Flush-right date-rail overlap on a long one-line header. It's real, but every text-altering remedy (shrink the wrap budget, or truncate) regresses the parse→export→re-parse round-trip — two corpus fixtures drop `role[0].company` because the parser reads company off the **single-line** `·`-header. The overlap is purely visual (parser ignores glyph position). Durable fix pairs a **font auto-fit** (round-trip-safe) with a parser-side wrapped-header rejoin; tracked in #437.

Gates: typecheck clean · 1942 tests pass · fallow report-only (only inherited findings).

Samhit21 (approved):
## Approving ✅

Re-reviewed on `2551349` — all three actionable findings fixed and verified locally:
- **skills `+`**: rejoins only on a single-word continuation (`+ CD` merges, `+ Ruby on Rails` does not) — fixes the bogus-merge; narrow residual documented.
- **headline name-skip**: `isNameLine` now case-insensitive + name-contains, so a name word cannot leak as the headline. Also cleared the fallow complexity flag.
- **normalizeUrl**: general scheme grammar preserves ftp/http — no more `https://ftp://`.

Flush-right overlap → #437 with the thread open: accepted as a round-trip-constrained deferral (text-altering fixes regress `role[0].company` on 2 fixtures; overlap is visual-only, durable fix is font auto-fit + parser rejoin).

Targeted tests + full CI green. Closes #425. LGTM.

Issue #311, opened 2026-07-03
Preserve multiple experience sections instead of merging into one


s-annam:
## Summary

A résumé may carry **more than one experience-category section** — e.g. `Performance Experience` + `Teaching Experience`, or `Relevant Experience` + `Additional Experience`. The candidate's grouping is deliberate semantic content (it shapes what a reader weights first). Today the parser **merges all such sections into one flat `Experience`** and the Reconstructed Resume renders a single heading, so the grouping is lost — and, because the reconstructed PDF is our round-trip fidelity target, the loss is permanent: two source sections re-parse as one.

This is a **scoped, post-launch enhancement** (milestone `M7 · Post-Launch Enhancements`), not a same-day parser-hardening fix. It builds on **#285** (single verbatim source heading already preserved) by extending it from one heading to **per-group headings**.

## Why this is valid (not noise to normalize away)

Multiple experience sections are common and legitimate, especially outside standard tech résumés:

- **Music / creative / academic CVs**: `Performance Experience`, `Teaching Experience`, `Ensemble/Conducting Experience`, `Leadership / Involvement`.
- **Students / early-career**: `Relevant Experience` vs `Additional Experience` (deliberately relegating the unrelated job).
- **Career changers**: domain-relevant roles grouped apart from prior-career roles.

The split is information. Flattening it discards a signal the candidate chose to send.

## Current behavior — where the merge happens

1. `matchSectionHeader` (`src/lib/heuristics/regex.ts:448`) maps **every** experience-category header — `Experience`, `Relevant Experience`, `Performance Experience`, `Involvement Experience`, … — to the single canonical `SectionName` `"experience"` (the last two via the head-noun anchor fallback, `matchSectionHeaderDetailed`, regex.ts:413).
2. `src/lib/heuristics/sections.ts:142-150` aggregates by canonical name: a second `"experience"` section's **lines are concatenated** into the first's bucket, and its heading is **discarded** by the `!sectionHeadings.has(section.name)` guard — `sectionHeadings` keeps only the *first* heading per canonical name.
3. `src/lib/pdf/ats-resume-model.ts:363` renders exactly one heading: `headings?.get("experience") ?? "Experience"`.
4. `ParsedResume.experience` (`src/lib/score/types.ts:229`) is a flat `ResumeExperience[]` with **no per-role section-group field** — so even a downstream consumer that wanted the grouping has nothing to read.

[...]

Issue #215, opened 2026-06-26
Bullets not detected in experience section because company comes first


sriyau64:
## Summary

In experience entries laid out **role-first** — where the line order is `Role title` → `Dates` → `Company – Location` → bullets (company sits *below* the dated role header, not above it) — the parser fails to attribute the role's bullets to the entry. Almost all bullets are dropped from the experience entry and surface under the **"Other bullets"** bucket in the reconstructed résumé. Specificity/Structure then score ~0 because the entry has no gradable bullets, dragging the overall score down for a résumé that is actually well-structured.

## Repro

PII-free fixture added for this bug:

`tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-role-first-experience.pdf`

It encodes two role-first experience entries (8 bullets total). Running the cascade:

- `experienceCount: 2` (sections found), but
- `bulletCount: 1` / `totalBullets: 1` — **7 of 8 bullets are orphaned** out of experience attribution into "Other bullets"
- `specificity 0/40`, `structure 0/30` (both `gradable: false` — too few attributed bullets), `completeness 24/30`
- overall **24/100**, no layout triggers

The original report used a Cal Poly career-center sample (`Computer Engineering CAREER sample`). That PDF carried real-looking PII (`(805) 291-4444`, a `@calpoly.edu` address, metadata author "Tammy Martin") so it is **not** committed; the synthetic fixture above reproduces the same structure.

## Root cause

The entry-block windowing in `src/lib/heuristics/entry-blocks.ts` (+ `disambiguateCompanyTitle` in the experience parser) assumes the company/institution line is part of the entry **header** — i.e. on or above the dated role line (the classic `Title \n Company <dates>` shape). This résumé places the company on its own line **below** the date anchor, between the role header and the first bullet. That company line is treated as the start of the bullet body (or shifts the body window), so:

1. The role's real bullets never make it into the entry's `description`.
2. `groupBulletsByExperience` (`src/lib/score/group-bullets.ts`) matches each pooled bullet against entry `description` text; with the bullets missing from `description`, they match nothing and fall to the `experienceIndex === null` "Other bullets" group (rendered by `ReconstructedRole.tsx` / `ReconstructedResume.tsx:roleLabel`).

Secondary symptoms from the same mis-windowing in the original sample: the company was truncated ("Experts Exchange – San Luis", losing "Obispo, CA"), and a following section header ("Leadership and Work Experience") was misread as a company.

[...]

Issue #427, opened 2026-07-09
Download-export follow-up (#421 Secondary #10)


s-annam:
Split off from PR #421 review (Secondary #10, @Vaishnavi1709).

Contact links are modeled by **two parallel override channels**:
- `contactOverrides.{linkedin_url,github_url,portfolio_url,website_url}` → `applyContactOverrides`
- `addedProfiles: AddedProfile[]` → `applyProfileOverrides`

PR #421 fixed the immediate symptom (an added LinkedIn/GitHub URL now back-fills the legacy `_url` slot the scorer + gap read), but the two channels still model the same concept and will keep drifting — e.g. an extractor that starts detecting GitLab has no legacy override target, so downstream consumers keying on the four legacy slots won't see it.

**Proposed depth:** consolidate onto one `profileOverrides` list keyed by profile id (network, slug); the four legacy `_url` fields become derived getters for back-compat readers. Score reads the list; ContactCard renders it.

Files: `src/lib/edit/apply-overrides.ts`, `src/hooks/useEditableParse.ts`, `src/lib/contact.ts`, `src/lib/score/score.ts`.

PR #435, opened 2026-07-10
fix(parser,score): experience field-mapping, section routing, bullet pooling, abbreviated dates (#382 #354 #365 #383)

### LINKED ISSUE #354: Parser: 2nd+ experience entry with anchor-ending company name mis-opens a new experience group (residual #258 edge)
## Problem

Surfaced during the adversarial review of #311 (PR #353). The `isInstitutionRepeat` adjacency relaxation in `src/lib/heuristics/sections.ts` that lets a distinct experience-category header open a new group (`!singleColumn || prevLineOpenedBoundary`) is scoped to `header.section === "experience"` — correctly, so #258's institution-name-as-header suppression still holds for every *other* section type.

But **within** an experience section a residual #258-shape edge remains: a single-column résumé whose **2nd+ experience entry** has a company name ending in a section-anchor word (e.g. `Global Teaching Experience Inc.`, `... Professional Experience`) that appears **after** an intervening full entry block (`prevLineOpenedBoundary === false`) will still mis-open a spurious new experience section — the same shape #258 fixed, now narrowed to the `experience` section only.

This is #311's **documented, accepted tradeoff** of the adjacency relaxation (the JSDoc above `isInstitutionRepeat` discusses it explicitly), not a regression from PR #353 — the review scoped round-1's "any section" blast radius down to experience-only. Filing so it isn't rediscovered later as a "new" bug, and so the edge gets a real regression guard.

## Repro shape (no fixture yet)

Single-column, one `Experience` section, two roles; the second role's company line ends in a section-anchor word after the first role's dates/bullets have intervened:

```
EXPERIENCE
Lead Engineer
Acme Corp · 2020 - 2023
- did things
Instructor
Global Teaching Experience Inc. · 2018 - 2020   <- 2nd entry; "...Experience" wrongly opens a new group
- taught things
```

Expected: one experience section, two roles. Actual: a spurious 2nd experience group; the company-name line is consumed as a heading.

## Acceptance criteria

- [ ] Distinguish "a genuine second experience-category header" from "a later role's company name that happens to end in an anchor word" using something stronger than `prevLineOpenedBoundary` (which only inspects the immediately-preceding line, not whether a whole entry block intervened). Candidate signals: the repeated line looks like a company/role line (has dates / follows a title line) vs. a bare category heading; or the heading text is a known experience-category phrase vs. an arbitrary anchor-ending company name.
- [ ] Single-column, 2nd+ experience entry with an anchor-ending company name → stays in the same experience section (no spurious group).
- [ ] #311's legit case still works: single-column `Performance Experience` + `Teaching Experience` → 2 distinct experience groups (`multi-experience-roundtrip.test.ts` still green).
- [ ] Regression test added in `sections.test.ts` for the anchor-ending-company edge (mirror the round-2 `#258` education guard).

## Refs

- Follows #311 / #285 / #258.
- Surfaced in PR #353 adversarial review (round-2 nit).

### LINKED ISSUE #365: [parser] experience — bare-paragraph bullets (no glyph) dropped, forcing manual +Add per line
## Category

## Fixture
`tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-achievements-oneline.pdf` (synthetic persona)

## Ground truth (`pdftotext`)
Experience section has **5 bullet-shaped lines** (3 under Staff SWE, 2 under Senior Engineer), rendered visually as bullets in the résumé but **without a leading bullet glyph** in the PDF text stream — the Google Docs / Skia print path emits them as bare paragraphs:

```
Staff Software Engineer · Northwind Labs
Austin, TX | Mar 2021 - Present

Led migration of the checkout service to a sharded datastore, cutting p99 latency 42% across 30M monthly transactions.
Built an internal experimentation platform adopted by 14 product teams within two quarters.
Mentored 6 engineers and established the team's code-review and on-call rotation standards.
```

The Achievements section in the SAME fixture uses `•`-prefixed bullets and is picked up correctly (`bulletCount: 3` in the corpus snapshot corresponds to those).

## Observed
Reconstructed view shows `No bullet-shaped lines detected.` under BOTH experience entries; user has to click `+ Add bullet` five times to recover content that is clearly visible on the résumé.

## Localized layer
`src/lib/heuristics/line-primitives.ts:19-21` — `isBulletLine` requires a leading glyph from `[•‣▪●◦⁃*\-–—]`. Bare-paragraph bullets never match. `isProseLine` (further down) exists for glyph-less prose but requires an internal sentence break (`SENTENCE_BREAK_RE`) — bullets that are a single sentence pass through neither path.

## Suggested fix
Two options — needs a design call:
1. Loosen `isBulletLine` to admit indent-only / single-sentence prose when it sits under an experience section header, OR
2. Add a "bare-paragraph bullet" recovery pass that fires only inside `experience`/`projects`/`achievements` sections and treats each ≥ 8-word non-header paragraph as a bullet.

## Repro
Drop the fixture in the dev server (`npm run dev`) and look at either experience entry. The corpus snapshot bakes the current `bulletCount: 3` in, so this fires no test failure — it's a UX cliff, not a broken invariant.

### LINKED ISSUE #382: fix: mislabels team as company on "Title, Team" roles under a shared-employer banner
## Summary

On a résumé where a **single employer heads a group of roles** and each role
header is `Title, Team` (the employer named once as a banner above the roles,
not repeated per role), the parser mislabels the **team/sub-org as the company**
and never attributes the true shared employer. Every role under the banner comes
back with `company = <team name>`, `title = <role>`, and no company that matches
the actual employer.

This is a **distinct bug class from #355** (label-rail section recognition).
#355 got the roles *segmented* correctly — the residual here is purely
company/title/team **field mapping** inside `disambiguateCompanyTitle`, one layer
downstream of section routing. #355 / PR #359 fenced this off deliberately
(`headerLookback: 2` doesn't reach a banner above the role group), so it needs
its own fix.

## Symptom (synthetic repro shape)

Source résumé structure (employer stated once, roles below share it):

```
Acme Corporation
  Staff Engineer, Platform Infrastructure        Aug 2024 – Present
  Senior Engineer, Payments Core                 Jul 2022 – Aug 2024
  Engineer, Identity Services                    Aug 2020 – Jul 2022
```

Current parse (wrong):

| Role | title | company | team |
|------|-------|---------|------|
| 1 | Staff Engineer | **Platform Infrastructure** | – |
| 2 | Senior Engineer | **Payments Core** | – |
| 3 | Engineer | **Identity Services** | – |

Expected:

| Role | title | company | team |
|------|-------|---------|------|
| 1 | Staff Engineer | **Acme Corporation** | Platform Infrastructure |
| 2 | Senior Engineer | **Acme Corporation** | Payments Core |
| 3 | Engineer | **Acme Corporation** | Identity Services |

(The last role in the observed real résumé — the one whose header carried its own
distinct employer name on the header line rather than under the shared banner —
maps correctly, which confirms the failure is specific to the shared-banner +
`Title, Team` shape.)

## Root cause

Two coupled gaps in `src/lib/heuristics/extract/experience.ts`:

1. **`splitRoleComma` (line ~273) assumes post-comma = company.** It splits
   `"Staff Engineer, Platform Infrastructure"` into
   `["Staff Engineer", "Platform Infrastructure"]` whenever the pre-comma segment
   `looksLikeTitle(...)` and the post-comma segment is neither a legal suffix nor
   a location tail. There is no signal that the post-comma segment is an internal
   **team/sub-org** rather than an employer. Downstream in
   `disambiguateCompanyTitle`, `looksLikeCompany("Platform Infrastructure")`
   returns true (multi-word Title-Case), so `chooseCompanyIdx()` selects it as
   the company and the pre-comma role becomes the title. The team is consumed as
   the company; `team` stays empty.

2. **No shared-employer propagation.** When one employer banner heads several
   consecutive roles, the employer line sits *above* the first role's header
   block and is not repeated per role. `headerLookback: 2` (line ~48) does not
   reach a banner that precedes a group, and `disambiguateCompanyTitle` operates
   on a single role's header lines in isolation — so even if gap (1) were fixed
   to treat the post-comma segment as `team`, the roles would have **no company
   at all**. The true employer has to be carried down from the banner to each
   role in the group.

Net: the header shape `<Employer banner>` over N × `Title, Team` has no path that
lands `company = <Employer>` for roles 2..N.

## Fix direction (not prescriptive)

Likely a two-part change; either part is useful alone, but the symptom only fully
clears with both:

- **Team-vs-company disambiguation for the comma split.** When a role header is
  `Title, X` and the block/group has a separate employer signal, prefer mapping
  `X → team` rather than `X → company`. Options: detect a shared banner above the
  group and, in its presence, route the post-comma segment to `team`; or add a
  `looksLikeTeam`/sub-org heuristic (no legal suffix, no company-suffix word, sits
  under a stronger company signal).
- **Shared-employer propagation across a role group.** Detect an employer line
  that heads a contiguous run of roles (a banner with no date anchor of its own,
  followed by ≥2 dated roles indented/below it) and attribute it as the `company`
  of each role in the run, with the per-role `Title, Team` supplying `title` +
  `team`.

Both are entry-segmentation / header-mapping concerns
(`extract/experience.ts` — `disambiguateCompanyTitle`, `splitRoleComma`,
`parseEntryBlocks`), **not** section routing (`sections.ts`) and **not** the
two-column detector — leave those fenced.

## Validating a fix

Reproduce with a **synthetic-persona** PDF (fake name, `@example.com`, a
`555-01xx` phone on a real area code) built in the shared-banner + `Title, Team`
shape — never a real candidate résumé (this residual was found via the local
`probe-experience` lane on a real PII résumé; that PDF must never become a
fixture). See `tests/fixtures/pdfs/README.md` for the add-fixture workflow. A new
`unknown/` fixture + `*.expected.json` snapshot exercising 3 roles under one
banner is the natural regression anchor.

## Acceptance criteria

- [ ] A résumé with one employer banner over N roles of shape `Title, Team` maps
      every role to `company = <banner employer>`, `title = <role>`,
      `team = <post-comma segment>`.
- [ ] The `Title, Company` single-role shape (employer named on the header line,
      no shared banner) still maps `company` correctly — no regression to the
      existing comma-split behavior.
- [ ] A synthetic-persona fixture (PII-free — verified per
      `tests/fixtures/pdfs/README.md`) + snapshot covers the shared-banner case.
- [ ] `npm run verify` green; corpus snapshots show no unintended regressions.

## Notes

- Surfaced during PR #359 / issue #355 while probing the real repro résumé with
  the `probe-experience` lane. #355 fixed section recognition (roles now segment
  correctly); this company/team mapping residual is the remaining wrong-field
  behavior on that résumé and is out of #355's scope.
- Values in this issue are **synthetic** — the real résumé's employer and team
  names are candidate PII and are deliberately not reproduced here.

### LINKED ISSUE #383: experience: awesome-cv-cv abbreviated-date header (Mmm. YYYY / Exp. YYYY) doesn't round-trip
## Summary

The `latex/awesome-cv-cv.pdf` fixture has an `experience` round-trip failure baselined in `KNOWN_FAILURES` (`src/lib/heuristics/corpus-roundtrip.test.ts`) as of #361. The baseline comment says the shape is "tracked as a follow-up" but cites no issue number — this issue is that follow-up.

## Background

PR #361 (the #341 `isProseLine` prose-tightening fix) **recovers a real role** this CV was previously dropping — an "Undergraduate Research, … Lab (Prof. …)" header that the old `Company. City` prose false-positive had swallowed. Net effect: experience count 16 → 17, a genuine win.

But the recovered role's header packs **inline abbreviated dates**:

```
Researcher … Mar. 2016 Exp. Jun. 2017
```

so its `title` / `start_date` don't round-trip cleanly. Recovering the role beats dropping it, so #361 baselined the `experience` round-trip in `KNOWN_FAILURES` rather than block on it:

```ts
// src/lib/heuristics/corpus-roundtrip.test.ts
"latex/awesome-cv-cv.pdf": ["experience"],
```

## Root cause (to confirm)

The abbreviated-date header shape (`Mmm. YYYY`, `Exp. YYYY`, month-abbreviation periods) is parsed such that the title absorbs date tokens / the start_date splits differently on re-parse. This is a distinct header-shape weakness, **not** the #298 title/company swap and **not** the #341 prose false-positive.

## Acceptance

- [ ] `latex/awesome-cv-cv.pdf` `experience` round-trips cleanly (parse → export → re-parse yields the same title/dates for the recovered role).
- [ ] Remove the `"latex/awesome-cv-cv.pdf": ["experience"]` line from `KNOWN_FAILURES` (the ratchet forces it once fixed).
- [ ] Handle abbreviated-month + `Exp.` date forms in the experience date parser generally, not just this fixture.

s-annam:
## Summary

Batch of four parser-experience fixes (epic #393), accumulated on one branch as a single commit. All in the parser/score layer (`src/lib/heuristics/*`, `src/lib/score/score.ts`); zero React/UI changes.

- **#382 — shared-employer banner mislabels team as company.** A single employer banner over N `Title, Team` roles now maps every role to `company=<banner>`, `title=<role>`, `team=<post-comma>` via a new `propagateSharedEmployer` post-pass in `extract/experience.ts`. Scoped tight: a plain `Title`-only role under a banner, a dated banner, and per-role `Title, Employer` layouts are left untouched. The single `Title, Company` shape is unchanged. New synthetic-persona fixture `unknown/shared-employer-banner-roles.pdf` + snapshot; verified through the full cascade (3 roles → `company="Acme Corporation"`).
- **#354 — anchor-ending company mis-opens a new experience group (residual #258).** A 2nd+ experience entry whose company ends in a section-anchor word (`...Experience Inc.`) after an intervening entry block no longer opens a spurious group. `sections.ts` now discriminates a dated entry from a genuine category header via date-adjacency (`hasDateRange` on the line or `lines[i+1]`). #311 (Performance/Teaching → 2 groups) and #258 education suppression stay green; regression test added.
- **#365 — bare-paragraph bullets dropped.** Root cause was **not** `line-primitives.ts` (the parse layer already recovers the bullets via the #215 indent mechanism) but `score.ts`: the experience-description fallback fired only when the *whole résumé's* marker pool was empty, so glyph bullets in Achievements masked glyph-less bullets in Experience. Now gated on the Experience section's own pool. `ATS_SCORE_ALGO_VERSION` 1.4→1.5. Confirmed end-to-end that the reconstructed view (`ReconstructedResume` reads `score.bullets`) genuinely recovers the 5 bullets — the "No bullet-shaped lines detected" UX cliff is fixed, not just the score number (google-docs fixture `bulletCount 3→8`, `score 67→94`).
- **#383 — abbreviated-date header doesn't round-trip.** `DATE_RANGE_RE` admits an optional `Exp.`/`Expected` end qualifier so `Researcher … Mar. 2016 Exp. Jun. 2017` parses (title no longer absorbs the date tokens). `latex/awesome-cv-cv.pdf` removed from `KNOWN_FAILURES` and round-trips cleanly. No false positives on `Experience`/`Experienced` (trailing `\s+` guard); `DATE_RANGE_RE` is shared with #354's `hasDateRange` but only genuinely-dated strings match, so section routing is unshifted.

Closes #382
Closes #354
Closes #365
Closes #383
Refs #393

## Test plan

- [x] `npm run test` green (1920 pass / 5 skip)
- [x] `npm run lint` clean
- [x] `npm run verify` (pre-push hook) passed
- [x] Fixture personas verified synthetic — no real PII (Step 3.5): `shared-employer-banner-roles.pdf` = Dana Rivera / dana.rivera@example.com / (512) 555-0142

## Adversarial review

Ran a pre-PR adversarial review (independent `ecc:typescript-reviewer`, opus) over the whole accumulated diff — instructed to break the change and reproduce suspected defects end-to-end, not just reason about code.

**Round 1 — `clean: true`, 0 blocking findings.** Converged in one round.

Reproductions the reviewer performed (all passed):
- `experience.shared-banner.test.ts` + `sections.test.ts` → 62/62; `corpus-roundtrip` → 47 pass (google-docs + awesome-cv-cv both round-trip).
- Direct `DATE_RANGE_RE`/`parseDateRange`/`stripDateRange` harness on 7 strings (#383): `Exp.`/`Exp`/`Expected`, dash + dash-dropped variants parse; `Experience`/`Experienced` do **not** match (trailing `\s+` holds); capture groups `m[1..6]` still align.
- Adversarial `extractExperience` on three #382-breaking shapes: `Title`-only role under a banner does **not** inherit; a dated banner is **not** treated as a banner; per-role `Title, Employer` keeps its own company. No over-reach.
- Confirmed the #365 UI symptom is genuinely fixed: `ReconstructedResume.tsx:832` reads the same `score.bullets` pool the gate mutates — the "No bullet-shaped lines detected" cliff clears, not just the score.
- Fixture persona re-verified synthetic via `pdftotext`.

**Nits (non-blocking — for the human reviewer):**
1. `score.ts` (#365) — mixed glyph **and** bare-paragraph bullets *within one* Experience section still drops the bare ones (pre-existing limitation, identical to the old whole-pool gate — **not** a regression; "mixed" in the issue means across sections).
2. `extract/experience.ts:89` (#382) — `bannerEmployer` uses `startsWith`; a short company ("IBM" vs "IBM Global Services") is a theoretical edge. Corpus-unreachable, clarity only.

Whole-tree `fallow` (report-only, matches CI): no production-code duplication introduced; `gen-shared-banner.mjs` unused-file flag is the accepted fixture-generator false positive (sibling of `gen-label-rail.mjs`); `computeAnonymousAtsScore` CRAP not materially worsened by #365 (net ~2 lines + one extracted helper).


Samhit21 (commented):
## PR Review: experience field-mapping, section routing, bullet pooling, abbreviated dates (#382 #354 #365 #383)

### Summary
Four well-scoped parser/score fixes with a clean adversarial pass and two honestly self-flagged nits (`startsWith` banner match, mixed glyph/no-glyph within one section) — I am not re-flagging those. I verified the true delta is exactly these four fixes (the #427/#414 code visible in a naive `main...` diff is pre-existing, a stale-base artifact). #365 (per-section bullet gating — changing *replace* to *append-when-experience-pool-empty* does not double-count) and #383 (`EXPECTED_END_QUALIFIER` is non-capturing, and `\s+` correctly blocks "Experience"/"Export" false-positives) both verify correct. CI green (verify / fallow / Pages).

### Findings
1. **[Suggestion] sections.ts:1855** — the #354 `hasDateRange(nextLine)` discriminator suppresses a genuine 2nd experience category header (#311 shape) when the following role uses a one-line `Title Dates` header; the collision shape appears uncovered by fixtures.
2. **[Nit] experience.ts:156** — banner-company inheritance does not recompute the entry confidence `score` when the prior company was empty (likely cosmetic).

### Meta (rebase)
- **This branch is behind `main` by exactly #434** (just merged), which also edits `sections.ts`. The changes are in different functions (`columnGapCuts` vs `isInstitutionRepeat`/`classifyLine`), so likely a clean textual merge — but please rebase and re-run `sections` + `corpus-roundtrip` before merge, since **#383 broadens `DATE_RANGE_RE`, which #434 `columnGapCuts`/`isLoneDateRange` also consume**. Confirm the Expected-qualifier widening is benign for #434 lone-date column exemption.

### Verdict
**Action: COMMENT**

src/lib/heuristics/sections.ts
```diff
@@ -1808,15 +1839,29 @@ function classifyLine(
...
     // not a second header — suppress the boundary and RETAIN it as content. The
     // full gate (current-section safety, the experience-only adjacency
-    // relaxation, and the two-column carve-out) lives in `isInstitutionRepeat`.
-    if (!isInstitutionRepeat(header, currentSection, singleColumn, prevLineOpenedBoundary)) {
+    // relaxation, the two-column carve-out, and the #354 dated-entry guard)
+    // lives in `isInstitutionRepeat`. A dated entry line is one carrying a date
+    // range on itself or the line immediately below it (a company whose role's
+    // date follows) — the tell of a company/role entry vs. a bare category header.
+    const lineLooksLikeDatedEntry =
+      hasDateRange(line) || hasDateRange(nextLine);
```
Samhit21, line 1883: [Suggestion]: `hasDateRange(nextLine)` suppresses a genuine 2nd experience category header when the following role uses a one-line title+date header.

The #354 discriminator treats "the next line carries a date" as proof the current header-shaped line is a company entry, not a category header. But roles commonly render as a **single** `Title  Dates` line — the #354 fixture itself does (`Staff Engineer  2012 - 2016`). So a genuine bare category header that #311 wants to open a new group on —

    Teaching Experience              <- category header (should open a group)
    Instructor  Jan 2020 – Dec 2021  <- one-line dated role

— gets `hasDateRange(nextLine) === true` → `isInstitutionRepeat` returns true → the header is suppressed and **no new experience group opens**, dropping the sub-section boundary. The fix stated invariant ("a category header is followed by a TITLE line, never a date") holds only for the two-line (title-then-date) shape the #311 test uses; it breaks for the one-line-header shape.

The two header-shaped lines are genuinely hard to separate, so this is a real tradeoff — just flagging that it resolves against #311 for one-line-header résumés, and I do not see a fixture covering that shape. Worth a corpus case (category header + one-line dated role) to pin the intended behavior; if suppression is intended there, say so in the comment.

src/lib/heuristics/extract/experience.ts
```diff
@@ -54,14 +54,113 @@ export function extractExperience(
...
+      // This role names the employer as a banner above its dated header: it
+      // opens (or re-opens) a run. Its own company is already the banner.
+      banner = own;
+      continue;
+    }
+    if (banner && isBannerContinuation(entry)) {
+      // A bare "Title, Team" continuation: inherit the shared employer, keeping
+      // the per-block `title` / `team` intact. The prior `company` was a title
+      // duplicate (already truthy), so the entry's confidence score is unchanged.
+      built[i].entry = { ...entry, company: banner };
```
Samhit21, line 161: [Nit]: The confidence `score` is not recomputed after inheriting the banner company.

`built[i]` is `{ entry, score }`, and `score` was computed by `experienceFromBlock` *before* propagation. When `isBannerContinuation` fires with `entry.company === ""`, this line fills in a real company but leaves `built[i].score` reflecting the empty-company entry. Likely cosmetic (the entry already passes the `finalizeEntries` predicate via `title`), but if that confidence feeds dedup/ordering anywhere downstream it is now stale. The `company === title` case is fine (company was already truthy, as the comment notes) — it is only the empty-company branch that shifts.

s-annam: Good catch — fixed in 470b50a. Only the empty-company branch shifts: it now earns the +0.25 company weight `experienceFromBlock` withheld (clamped to 1). The `company === title` / non-company branches were already truthy, so their score is unchanged, and I tightened the comment to say exactly that. It does feed downstream ordering, so worth getting right rather than leaving stale.

src/lib/heuristics/sections.ts
s-annam: Agreed this is a real tradeoff, not a clean win — the two header-shaped lines aren't separable from layout alone, and tightening `hasDateRange(nextLine)` to open the boundary risks re-splitting the exact company/role entries #354 was added to hold together. So I'm keeping the #354 discriminator as-is in this PR and treating the one-line-header case as a known, scoped edge.

Logged it as a concrete instance on #438 (https://github.com/resumelint-org/resumelint/issues/438#issuecomment-4942529011) — it's a clean example of header-vs-entry intent being re-derived from adjacent raw lines instead of read off a structured field, which is exactly what Direction 3 there would collapse. The corpus case you suggested (category header + one-line dated role) will land with that decision so we don't pin behavior we're about to change. Leaving this thread open for you to close once you're good with deferring it there.

Samhit21 (commented):
## Re-review — CI is red (do not merge yet)

Thanks for the score-recompute fix + the #438 deferral (reasonable). But `verify` fails on `470b50a` with 3 regressions the score commit introduced (it was green at `865f958`):

- `corpus-roundtrip` → `latex/awesome-cv-cv.pdf: 'experience' regressed`
- `corpus.test` → `shared-employer-banner-roles.pdf` cascade+score snapshot mismatch (needs re-bake)
- `multi-experience-roundtrip` → **#311 (2 sections → 2)** got 0

The `+0.25` confidence bump feeds entry ordering (as you noted), so it reordered roles and rippled into round-trip output — the shared-banner snapshot went stale and two round-trip fixtures regressed. Please re-bake the shared-banner snapshot and resolve the two round-trip regressions (or apply the confidence correction without perturbing entry order), then I will re-review.

The #354 → #438 deferral thread I will leave open per your note.

src/lib/heuristics/sections.ts
s-annam: Good call — and this became *real* on rebase, not just hypothetical. Main's #434 one-line experience export (`Title · Company, Location  Dates`) renders exactly the shape you described: a genuine second category header (`TEACHING EXPERIENCE`) followed by a one-line dated role. On re-parse, `hasDateRange(nextLine) === true` → `isInstitutionRepeat` suppresses the header → the #311 multi-section round-trip flattens (2 groups → 0 section_labels).

Confirmed the #354 trap (`GLOBEX GLOBAL EXPERIENCE` → `Staff Engineer 2012-2016`) is structurally identical to this category-header case — anchor-ending line + dated next line — so `prevLineOpenedBoundary` + `hasDateRange` genuinely can't separate them. That's the tradeoff you flagged.

Rather than hand-tune the discriminator now (which #438/#439's parse→export decompose would likely rototill), I've routed this to **#436** — the one-line-header title/company disambiguation issue, which is the proper home for both the awesome-cv company-truncation and this category-header suppression. #311's way-back-out grouping is baselined to the #436 known gap (kept the way-IN 2-group parse + no-role-lost count as teeth; tightens back to 2 when #436 lands). The two-line-header parse this PR targets is unaffected.

Leaving this thread open as the breadcrumb for #436. Filing the category-header-vs-#354-anchor discriminator as an explicit sub-note there.

s-annam:
@Samhit21 — sorted; ready for re-review. Here's what was actually going on and what I changed:

**Root cause of the red `verify`:** all 3 failures were **branch-behind-main**, not the score-recompute change itself. CI tests the PR *merged with main*, and main had gained #434 (the one-line experience export header + `SNAPSHOT_SCHEMA_VERSION` 3→4) after this branch's last green run. They only reproduced locally *after* I rebased onto main — which is why the branch looked green in isolation.

**What I did (rebase + baseline under #436, land now):**

| Failure | Fix |
|---|---|
| Rebase / snapshot v3≠v4 | Rebased onto main (#434); rebaked `shared-employer-banner-roles.expected.json` to v4 (schema bump only — score/cascade values unchanged) |
| `awesome-cv-cv` round-trip — 12 roles re-parse company-truncated | Added to `KNOWN_FAILURES ["experience"]` under the existing #436 group |
| #311 multi-section round-trip flattens 2 groups → 0 labels | Relaxed the *way-back-out* assertion to the #436 known gap; kept the way-IN 2-group parse + no-role-lost count as teeth |

**On the suppression thread you flagged:** you were right, and #434's export is what made it bite. `isInstitutionRepeat` / #354 suppression reads a one-line dated role (`Title · Company, Location  Dates`) under a category heading as a *company entry* rather than a new section — so the two experience groups flatten on re-parse. That's the same #436 root as the other baselines. I left that thread **open** as the breadcrumb (deferred, not code-fixed) and filed a concrete #436 sub-note with the category-header-vs-#354-anchor discriminator + candidate fix. The ratchet auto-removes both baselines when #436 lands.

Deliberately did **not** hold this behind #438/#439 — those are the big P2 architecture refactors; gating a done correctness PR on them just strands it on a drifting branch. And avoided a pointed `isInstitutionRepeat` hack precisely because #436/#438/#439 would supersede it.

`npm run verify` green locally; CI green on `e4f6d27`.

Samhit21 (approved):
## Approving ✅

Re-reviewed on `e4f6d27` — CI is green and the 3 regressions from last round are resolved honestly:

- **awesome-cv-cv round-trip** — genuinely fixed (baseline removed); the recovered role now splits its `Exp.`-qualified date via #383. Snapshot re-bake is just schema v3→v4 + `headline` + `algoVersion 1.5` (confidence 0.89→0.90) — no hidden experience reorder.
- **shared-employer-banner snapshot** — re-baked to v4, scores sane.
- **#311 round-trip** — honestly baselined to #436, not weakened: `parse1` still asserts `2` groups and the no-role-lost guard stays; only the post-round-trip regrouping is baselined to `0` (ratcheted, tightens back to `2` when #436 lands). This is exactly the one-line-header suppression I flagged in Finding 1, now real via #434 and properly tracked.

My two findings are both resolved — score-recompute landed correctly, #354 collision deferred to #438 (thread open). Rebased onto main, snapshots consistent at v4. LGTM.

Issue #313, opened 2026-07-03
Create a blank resume from scratch and download it as a PDF


s-annam:
## Problem / feature request

Today the only way into resumelint's editable, downloadable resume surface is to **upload a PDF (or DOCX)** first. The parse produces a `CascadeResult` + `AnonymousAtsScore`, which mount `ReconstructedResume` — a fully editable resume with inline field editing and `+ Add` affordances on every section, plus a client-side **Download PDF** button.

Users who don't yet have a resume (or want to start clean) have no entry point. We should let a user **create a blank resume from scratch — no upload required — fill it in, and download it as a PDF.**

The good news: the editor and the PDF exporter already exist and already work with zero content. The missing pieces are (1) a factory that seeds an empty `CascadeResult` + score, and (2) an entry point + a lightweight surface that mounts the existing editor without the parser-audit chrome.

## Reuse analysis

**Capability:** author/edit a resume's fields + sections, and export it as a PDF.

**Existing surfaces found (all reused as-is):**
- `src/components/features/ReconstructedResume.tsx` — the editable resume surface. Renders contact + experience + projects + achievements + education + skills, each with an `AddPill` (`+ Add experience`, `+ Add project`, `+ Add skill`, …) that renders **unconditionally** (see the "reachable on every resume" comments at `ReconstructedResume.tsx:853-855`). Contact and every field are click-to-edit via `EditableField`. It already owns the **Download PDF** `Button` (`ReconstructedResume.tsx:881-888`).
- `src/hooks/useEditableParse.ts` — the override/add/remove edit model (`addEntry`, `addBullet`, `setContactField`, `addSkill`, …).
- `src/hooks/useDownloadPdf.ts` → `src/lib/pdf/ats-resume-model.ts` (`buildAtsResumeModel`) → `src/lib/pdf/render-ats-pdf.ts` (`renderAtsResumePdf`) — the fully client-side PDF exporter. No network, no font fetch (zero-egress).
- `src/lib/heuristics/cascade.ts:443` `emptyParsed()` and `cascade.ts:453` `buildScannedResult()` — the precedent for constructing an empty/degenerate `CascadeResult`.

[...]

Issue #224, opened 2026-06-26
Parser/reconstruction: achievements render twice — duplicated into "Other bullets" (title-only entries don't match bullet pool)


s-annam:
## Summary

**Reproduced.** Every achievement renders **twice** in the reconstructed résumé: once (correct, full text) under the **"Other bullets"** group, and once (mangled) under the **Achievements** section. The original "repeated in parsed output" framing was imprecise — it is not a duplicate inside `heuristic_achievements`, it is a **cross-section duplication between the Achievements section and the "Other bullets" leak group**.

Confirmed on a real résumé whose Achievements section uses the one-line `• Label · text [year]` shape (Patent / Publication / Exit). See the synthetic fixture added with this issue.

## Observed (from the UI)

```
Other bullets
  • Patent · … e-commerce product catalogs. [2019]
  • Publication · Co-authored 5 publications on Web Accessibility. [2009]
  • Exit · Founded and sold return2india.com to Satyam Infoway (NASDAQ: SIFY). … [2000]

ACHIEVEMENTS
  Patent · … e-commerce product catalogs. [] · 2019      ← No bullet-shaped lines detected.
  Publication · Co-authored 5 publications … . [] · 2009  ← No bullet-shaped lines detected.
  Exit · Founded and sold  to Satyam Infoway … . [] · 2000 ← No bullet-shaped lines detected.
```

## Root cause

The achievements parse as **title-only** entries — the whole `Label · text [year]` line lands in `HeuristicAchievement.title` with an **empty `description`** (the "No bullet-shaped lines detected." subtext is the tell). The reconstruction then double-counts that content:

1. `AchievementsSection` (`ReconstructedResume.tsx`) renders each achievement's **title**.
2. The separate bullet pool (`score.bullets`, scorer-extracted from section text) still contains the raw `• Patent…` lines. `buildEntryGroups` → `groupBulletsByExperience` matches every bullet against entries' **`description`** text. Title-only achievements have no description → those bullets match nothing → they fall into the `experienceIndex === null` **"Other bullets"** group (`ReconstructedResume.tsx:264-318`).

Net: the achievement content shows in both places. This is a **reconstruction-layer / parser-shape coupling**, not a parser repeat.

## Fix direction (to decide in implementation)

Either side closes the duplication; likely both:
[...]

PR #456, opened 2026-07-13
feat(edit): full edit fidelity for the Achievements section (#454, #455)

### LINKED ISSUE #453: [epic] Full edit fidelity for the Achievements section
## Epic — full edit fidelity for the Achievements section

The Achievements section is the only reconstructed-résumé section a user cannot fully edit.

**Today's state:**
- **Existing (parsed) achievements are read-only.** `AchievementsSection` renders a parsed achievement as a plain `<h3>` header + graded bullets, with no inline-edit affordances (`src/components/features/ReconstructedResume.tsx`, the non-`added` branch). Experience roles, contact, education and skills are all editable; achievements are not.
- **The add flow has lower fidelity than the rest.** Adding an achievement exposes exactly two fields — a single multiline `title` and a `year` — collapsing the achievement's *type* ("Patent", "Publication", "Award") and its *description* into one blob. Experience, by contrast, gives distinct title / company / location / team / dates fields.

**Goal:** bring Achievements to parity with the other sections —
1. edit an **existing** achievement's **type / description / year** inline, and
2. give the **add** flow the same three-field (type + description + year) structure.

**Why now / anchor:** #452 shipped `splitAchievementType` (`src/lib/score/entry-dates.ts`) — the *type* is already a first-class concept in the render + Download-PDF path (only the type label is bolded, both surfaces). This epic makes the type a first-class **edit** target too, so the field a user sees emphasized is the field they can change.

## Breakdown

```mermaid
flowchart TD
  E["[epic] #453 Achievements edit fidelity"]
  E --> C1["#454 Edit existing achievements inline"]
  E --> C2["#455 Structured add-achievement fields"]
  E -. references .-> N380["#380 display-polish nits"]
  C1 -. blocks .-> C2
```

- **#454 — Edit existing (parsed) achievements inline (type / description / year).** The foundation: an achievement-override model + editable header, mirroring how `ReconstructedRole` edits an experience role. Establishes the type/description/year decomposition (via `splitAchievementType`) and the override→apply→export round-trip. **Blocks #455** (which reuses the same field model + decomposition).
- **#455 — Structured add-achievement flow (type + description + year).** Replace the single-`title` add affordance with distinct type + description + year fields, reusing #454's model so an added achievement is indistinguishable from a parsed-then-edited one.
- **#380 (referenced, not reparented) — display-polish nits** — includes the two achievement-specific render nits (empty Achievements section still shows `+ Add achievement`; awards separator flips source comma to middot). Its achievement nits belong to this section's polish, but #380 is a sub-issue of batch epic #391 and also carries a non-achievement project-dates nit, so it stays under #391 and is tracked here by reference.

## Tooling — `probe-achievements` skill (no separate sub-issue)

The `probe-*` parser-probe family (`probe-experience`, `probe-contact`, `probe-roundtrip`) lets a maintainer drop a real résumé PDF and see exactly what the parser reads for one section, localizing which layer dropped/garbled a field. Achievements have no such probe. As part of this epic, add **`probe-achievements`** — a sibling skill + dev harness (an `RL_ACHIEVEMENTS_PDF` block in a vitest test, matching the existing probes' Vite/vitest execution constraint), printing each parsed achievement's **type / description / year / bullet count** straight off `runCascade(...)`, plus the achievements region the segmenter saw — so the type/description split this epic edits is visible end-to-end, and an edit-fidelity bug can be told apart from a parse bug.

Same guardrails as the sibling probes: **PII — real résumé is local-only, never committed**; output values go to the gitignored `internal/` scratch dir; cite defects by category, never candidate values. Fold the skill work into #454 (it directly supports the type/description edit model), or track as a checklist item here — it does not warrant its own sub-issue.

## Epic-level acceptance criteria

- [ ] An existing parsed achievement's **type**, **description**, and **year** are each independently editable inline in the reconstructed résumé.
- [ ] The **add-achievement** flow offers the same three fields (type + description + year), not a single title blob.
- [ ] Edits + additions round-trip through `apply-overrides` → `ats-resume-model` → Download PDF, and re-parse back to the same fields (the tested export invariant).
- [ ] The edited **type** is the run that renders bold in both the reconstructed view and the PDF (consistent with #452 `splitAchievementType`).
- [ ] A `probe-achievements` skill exists to inspect parsed achievement type/description/year on a real résumé (folded into #454 or as a checklist item here).
- [ ] No regression to experience/education/skills edit surfaces or to achievement bullet grading/attribution.

## Notes

- The **type** is not a separate parsed field today — it is the leading `" · "`-delimited segment of `title` (see `splitAchievementType`, guarded by `ACHIEVEMENT_TYPE_MAX_LEN`). #454 must decide the edit model: either (a) keep `title` canonical and decompose/recompose type⇄description on edit, or (b) promote `type` to a distinct field on the achievement model. #454 owns that decision; #455 inherits it.
- Out of scope for this epic (kept standalone): #381 (cross-section edit-persistence signal) and #376 (read-mode placeholder words across education/experience) — broader edit-interaction issues, not achievement-specific.

### LINKED ISSUE #454: Edit existing (parsed) achievements inline — type / description / year
Part of the Achievements edit-fidelity epic.

## Problem

Existing (parsed) achievements are **read-only** in the reconstructed résumé. Every other section — experience, contact, education, skills — is inline-editable, but an achievement renders as a static header and cannot be corrected. If the parser mis-splits the type, drops the year, or garbles the description, the user has no way to fix it short of re-authoring.

## Current behavior

`src/components/features/ReconstructedResume.tsx`, `AchievementsSection`: the non-`added` branch renders the achievement via `AchievementHeader` (added in #452) — a plain `<h3>` with the type run bold, the rest regular — with **no edit affordance**. The `added` branch (user-added achievements) already uses `EditableField` for title + year, so the editable pattern exists; it just isn't applied to parsed entries.

## Design decision (owned by this issue)

The **type** is not a distinct parsed field — it is the leading `" · "`-delimited segment of `title`, extracted by `splitAchievementType` (`src/lib/score/entry-dates.ts`, guarded by `ACHIEVEMENT_TYPE_MAX_LEN`). Two viable models:

- **(a) Keep `title` canonical**, decompose into type + description for editing and recompose (`type · description`) on commit. Zero data-model change; the override still writes `title`.
- **(b) Promote `type` to a distinct field** on the achievement model + overrides.

Prefer **(a)** unless it proves lossy — it keeps `HeuristicAchievement`, the canonical model, `apply-overrides`, and the export projection unchanged, and matches how `splitAchievementType` already treats `title` as the source of truth. Document the choice in the PR.

## Implementation sketch

1. **Override model** — add an achievement-override shape (mirroring `ExperienceFieldOverrides` in `src/hooks/useEditableParse.ts`): `{ type?, description?, year? }` keyed by achievement index. If model (a), the apply step recomposes `title = joinType(type, description)`.
2. **Apply** — `src/lib/edit/apply-overrides.ts`: fold achievement overrides onto the parsed achievement before scoring/render, same pattern as `applyExperienceHeaderOverrides` (empty string clears, mirroring the team/location fields).
3. **UI** — in `AchievementHeader` / `AchievementsSection`, replace the read-only `<h3>` for parsed entries with three `EditableField`s: **type** (semibold, the bold run), **description** (regular), **year** (xs) — matching the add-flow layout and the experience role header. Reuse `EditableField` from `@design-system`; no new primitive.
4. **Export** — confirm the edited achievement round-trips through `ats-resume-model.ts` `buildAchievementHeader` (already reads `title` + `year` and bolds only the type run) and re-parses to the same fields.

## Acceptance criteria

- [ ] A parsed achievement's **type**, **description**, and **year** are each independently editable inline; edits persist in the reconstructed-view state.
- [ ] Clearing a field to empty behaves consistently with experience fields (drops it, no literal placeholder rendered as a value).
- [ ] The edited **type** is the run rendered bold in both the reconstructed view and the Download PDF (`splitAchievementType` consistency).
- [ ] Edits round-trip: reconstructed edit → `apply-overrides` → `ats-resume-model` → PDF → re-parse yields the same type/description/year.
- [ ] Unit test for the achievement-override apply (set + clear round-trip), mirroring the experience/team override test in `src/lib/edit/apply-overrides.test.ts`.
- [ ] No regression to achievement bullet grading/attribution or to the user-added-achievement branch.

## Reuse analysis

- **Capability:** edit an achievement record (type / description / year) inline.
- **Existing surfaces found:** `ReconstructedRole` (`src/components/features/ReconstructedRole.tsx`) edits experience roles via `EditableField`; the `added`-achievement branch of `AchievementsSection` already edits title + year via `EditableField`; `EditableField` primitive (`@design-system`).
- **Decision:** **extend** — apply the existing `EditableField` + override-apply pattern to the parsed-achievement branch. No new surface; reuse the experience-role edit model and the design-system primitive.

### LINKED ISSUE #455: Structured add-achievement flow — type + description + year fields
Part of the Achievements edit-fidelity epic. **Blocked by C1** (reuses C1's achievement field model + type/description decomposition).

## Problem

The **add-achievement** flow has lower fidelity than every other add flow. It exposes only a single multiline `title` field and a `year` field, collapsing the achievement's *type* ("Patent", "Publication", "Award") and its *description* into one blob. Adding an experience role, by contrast, gives distinct fields. A user adding an achievement can't structure it the way the parser/render pipeline expects (`type · description · year`), so the type never renders bold and the entry reads inconsistently against parsed achievements.

## Current behavior

`src/components/features/ReconstructedResume.tsx`, `AchievementsSection`, the `added` branch: two `EditableField`s — `title` (multiline, semibold) and `year` (xs), joined by a middot. The `AddedEntry` / `AddedEntryField` model (`src/hooks/useEditableParse.ts`) carries `title` + `year` for achievements.

## Implementation sketch

1. **Reuse C1's field model.** Once C1 establishes the type/description/year edit decomposition (and, if C1 chose model (b), a distinct `type` field), the add flow adopts the same three-field structure so an added achievement is indistinguishable from a parsed-then-edited one.
2. **UI** — replace the single-`title` add affordance with three `EditableField`s: **type** (semibold), **description** (regular), **year** (xs), matching the parsed-edit layout from C1. Reuse `EditableField`; no new primitive.
3. **Model** — extend `AddedEntry` / `EXPERIENCE_FIELD_MAP` handling for achievements to carry the type + description split (or recompose into `title` per C1's decision) so `appendAddedEntries` (`src/lib/edit/apply-overrides.ts`) pushes a well-formed achievement.
4. **Export** — an added achievement round-trips through `ats-resume-model.ts` `buildAchievementHeader` and re-parses to the same fields.

## Acceptance criteria

- [ ] The add-achievement flow offers distinct **type**, **description**, and **year** fields (not a single title blob).
- [ ] An added achievement renders identically to a parsed achievement with the same values — the type run is bold in both the reconstructed view and the PDF.
- [ ] An added achievement round-trips: add → `apply-overrides` (`appendAddedEntries`) → `ats-resume-model` → PDF → re-parse yields the same type/description/year.
- [ ] The layout matches C1's parsed-edit layout (same field order, weights, separators).
- [ ] No regression to the add flow for other sections or to achievement bullet add (`InlineBulletAdd`).

## Reuse analysis

- **Capability:** create (add) an achievement record with structured type / description / year.
- **Existing surfaces found:** the `added`-achievement branch of `AchievementsSection` (title + year today); C1's parsed-achievement edit model + `EditableField` layout; `EditableField` primitive (`@design-system`).
- **Decision:** **extend** — grow the existing add-achievement branch to C1's three-field model; no new surface.

s-annam:
## Summary

Achievements were the only reconstructed-résumé section a user could not fully edit — parsed entries were read-only, and the add flow collapsed the achievement's *type* and *description* into one title blob. This brings the section to parity with experience / education / skills.

**Design decision — `type` is a real field.** This PR originally shipped model (a): `title` stayed the only canonical parsed field and the type label was merely its leading `" · "` run, decomposed for editing and recomposed on commit. That model is **reversed here** (see below). `HeuristicAchievement` now carries `type`, populated once at parse.

- **#454 — edit existing achievements inline.** Type / title / year are each independently editable (three `EditableField`s, mirroring the experience role header). Each override key names a real field, so `applyOverrides` copies it straight across. The achievement-override map is folded in *before* the added-entry append, so parsed-index keys never collide with appended entries. **Verified:** `render-roundtrip-achievement-edit.repro.test.ts` drives edit → `apply-overrides` → `ats-resume-model` → real PDF → `runCascade` re-parse, asserting the same type/title/year and that the emphasis sentinels wrap exactly the type run.
- **#455 — structured add-achievement flow.** Same three fields on the add path. `AddedEntry.achievementType` maps onto the pushed achievement's `type`, so an added achievement is indistinguishable from a parsed-then-edited one. **Verified:** its own `render-roundtrip-achievement-add.repro.test.ts`, same end-to-end shape.
- **`probe-achievements` skill + harness** (epic #453), sibling of the existing `probe-*` parser probes: prints each parsed achievement's type / title / year / bullet count straight off `runCascade`, plus the region the segmenter saw. Inert without `RL_ACHIEVEMENTS_PDF`; output goes to the gitignored `internal/`. No fixture PDFs added — no PII surface.

## Reversing model (a) — the composed title was not re-splittable

Model (a) made `decompose ∘ join` load-bearing, and **it is not the identity**. Every consumer holding only the composed title re-split it differently from what the user typed. Two surfaces were confirmed broken:

- the **PDF exporter** bolded the wrong run — a title carrying its own `" · "`, or a cleared type, promoted the title's first segment to the bold label;
- **`/jd-fit`** re-decomposed the recomposed title and showed the wrong halves.

The in-app patch (pin both halves into the override map on first edit, `lib/edit/achievement-fields.ts`) held on `/` only. It could not cross the jd-fit handoff, and its docstring understated the blast radius as *"only the PDF's bold run is wrong"* — which was false.

Storing the label deletes the whole class rather than patching each surface:

- the split runs **exactly once**, in `extractAchievements`, into a real `type` field; `title` is the header without the label;
- the exporter bolds `ach.type` verbatim — any length, any punctuation;
- the edit surface renders the override-applied fields directly, so the pinning layer *and* `decomposeAchievementTitle` / `joinAchievementTitle` are **deleted**, not reworked.

**Round-trip caveat, documented where it lives** (`buildAchievementHeader`): the exported *text* is still `"Type · Title"`, so re-parsing recovers `type` only when the label passes `splitAchievementType`. The bold run is the PDF's only other encoding of the label and the parser does not read font weight — inherent to the format, not to the model. JSON Resume has no type slot either, so `awards[].title` keeps carrying the full `"Patent · Foo"` line; narrowing it would have silently dropped the label from the export.

## The `/jd-fit` handoff carried an applied result

Same root cause, second surface — and it exposed a **pre-existing** bug the achievements work only made visible.

The handoff serialized the override-**applied** `CascadeResult`, and `/jd-fit` then started a fresh, **empty** edit layer over it. The edits were baked into the parsed fields and unrecoverable on the far side: a user-**added** entry arrived indistinguishable from a parsed one — shown, but with no Remove button and its bullets no longer attributed to it. That hit experience, education, projects, and achievements alike, and **predates this PR**.

Naively re-seeding the far side from the applied payload is *not* the fix — it would append every added entry a second time. The handoff has to carry the same two inputs `/` itself had: the **pristine parse + score**, and the **edit state**.

So edit state gets one serializable shape, `EditSnapshot`, owned by the hook that owns the state. `useEditableParse` now exposes `snapshot` and `replay`. The from-scratch draft (#313) already needed exactly this and had hand-rolled it in `useAnalyzedResume` — that copy is deleted and `BlankDraftSnapshot` is now an alias. One shape, one replay, so a new override map cannot be added without joining both (which is how `team` and `achievementType` were once silently dropped on restore). `/jd-fit` replays once on mount, guarded, since replay is additive; `consumeJdFitHandoff` rejects a payload with no `edit` key so a stale one falls back to the DropZone instead of silently showing an un-edited résumé.

## Incidental fixes (same class, in code this epic touched)

- Added-entry **draft restore dropped `achievementType`** on reload — and, in pre-existing code, **`team` (#425)** — because the replay iterated a hand-synced field list. `AddedEntryField` is now derived *from* that tuple, so a new field cannot join the union without joining the replay. (Confirmed the old `satisfies` guard was inert; the new one fails the build.)
- `useJdFitResume` was **omitting the `fieldConfidence` argument** to `applyOverrides`. Inert in practice, now passed.

## Test plan

- [x] `npm run lint` clean (style guards: no raw `<button>`, hex, palette classes, or manual `dark:`)
- [x] `npm run test` green — **2103 passed**, 8 skipped, 0 failed
- [x] `npm run build` clean
- [x] `fallow audit --base origin/main` — no dead code, **no unused exports**, no new duplication

New coverage for the reversal: the parser lifts the label; the two shapes the old model got wrong (over-long type, separator-bearing title) now survive `applyOverrides` intact; a cleared type bolds the whole header instead of promoting the title's first segment; and a new `useJdFitResume.test.tsx` pins the handoff — a correction replays onto the pristine parse, an added entry stays an **added** entry on the far side, and editing on `/jd-fit` does not re-append it.

Closes #453
Closes #454
Closes #455

## Adversarial review

Two rounds of independent adversarial review (`ecc:react-reviewer`) ran on the accumulated diff **before** this PR was opened. A third-party review on the open PR then found the two surfaces that pushed the model (a) reversal above. Recorded here as an audit artifact.

**Round 1 — 2 blocking, `clean: false`.**

1. **`decompose ∘ join` is not idempotent, and the header re-decomposed the already-recomposed title every render.** Reproduced with three failing assertions: a cleared type swallowed the description; an over-long type dumped the whole title into the description field and doubled itself on re-commit; a cleared description reverted the type field to its placeholder while the PDF still emitted it. Fixed *at the time* by pinning both halves into the override map — **since superseded**: the real fix was to stop composing the title at all.
2. **Added achievements lost their type label on blank-draft restore** (`achievementType` missing from the hand-synced replay tuple). The reviewer found a **sibling in untouched code** — `team` (#425) was dropped the same way. **Fixed** both, and made the tuple exhaustive at the type level so the class can't recur.

**Round 2 — `clean: true`.** Both blockers verified fixed by live repro against the real `applyOverrides` loop (the reviewer distrusted the unit tests' simulated apply step and drove production code instead).

**Post-open review — the model (a) reversal.** A review on the open PR argued the pinned-halves patch treated a symptom: it stabilized the edit surface on `/` but every *other* consumer still re-split the composed title. Verified against the code — the mechanism checked out on both surfaces it named (`jd-fit-handoff.ts` + `useJdFitResume.ts` for the re-decompose; the exporter for the bold run), and its "added entries baked in" finding was real but **pre-existing** and not attributable to this PR. Rather than defer, both are fixed here: `type` is now a real field and the handoff carries edit state. `lib/edit/achievement-fields.ts` — the entire pinning mechanism — is **deleted**.

**Nits surfaced, then fixed before opening this PR** (not deferred):
- A type-less achievement (`"Best Paper Award"` — the common shape) rendered regular-weight in the view but whole-line bold in the PDF. Reclassified from nit to fix.
- The `satisfies readonly AddedEntryField[]` guard gave **zero** exhaustiveness protection — the reviewer proved it by deleting `"team"` and watching `tsc` still exit 0. Replaced with a derived-union tuple; deleting a field now fails the build in two places.

**Surviving nits for the human reviewer's eye** (non-blocking):
- `AchievementHeader` trips fallow's CRAP threshold at 0% direct component coverage (it is exercised end-to-end through the round-trip and hook tests, not in isolation). `App`, `buildAtsResumeModel`, `educationEntries`, and `applyExperienceHeaderOverrides` also trip complexity — all pre-existing, all report-only in CI.

### DIFF
--- a/src/jd-fit/useJdFitResume.ts
+++ b/src/jd-fit/useJdFitResume.ts
@@ -7,10 +7,11 @@
  * Two sources collapse to ONE `{ result, score, edit, parsed, ... }` shape that
  * `<Result>` and the JD coverage memo consume identically:
  *
- *  1. Handoff from `/` — the parser-audit surface stashed an already-parsed +
- *     edited résumé in sessionStorage (read once on mount). JD-fit re-grades it
- *     live through its OWN edit layer so inline edits here move the score + JD
- *     coverage, exactly as on `/`.
+ *  1. Handoff from `/` — the parser-audit surface stashed the PRISTINE parse
+ *     plus the user's edit snapshot in sessionStorage (read once on mount).
+ *     JD-fit replays those edits into its OWN edit layer and applies them here,
+ *     so inline edits move the score + JD coverage exactly as on `/` and the
+ *     edits stay editable (#456).
  *  2. Local DropZone parse — when there's no handoff, the user drops a PDF here
  *     and `useAnalyzedResume` drives the parse → edit → re-grade pipeline.
  *
@@ -18,7 +19,7 @@
  * /jd-fit falls back to the DropZone rather than re-showing a stale résumé.
  */
 
-import { useEffect, useMemo, useState } from "react";
+import { useEffect, useMemo, useRef, useState } from "react";
 import type { AnalyzedResume } from "../hooks/useAnalyzedResume.ts";
 import { useEditableParse, type EditableParse } from "../hooks/useEditableParse.ts";
 import { applyOverrides } from "../lib/edit/apply-overrides.ts";
@@ -50,9 +51,18 @@ export interface JdFitResume {
 }
 
 export function useJdFitResume(analyzed: AnalyzedResume): JdFitResume | null {
-  // Read the one-shot handoff exactly once, on mount.
+  // Read the one-shot handoff exactly once, on mount. The guard is a REF, not
+  // state: `consumeJdFitHandoff` reads AND clears sessionStorage, so it is not
+  // idempotent, and StrictMode (dev) runs an effect setup→cleanup→setup within
+  // one commit with no re-render between. A state guard cannot short-circuit
+  // the second setup — it captures the same pre-update closure — so the second
+  // read would find the key already cleared and `setHandoff(null)` would win,
+  // dropping the whole handoff. A ref flips synchronously and survives it.
+  const consumedRef = useRef(false);
   const [handoff, setHandoff] = useState<JdFitHandoff | null>(null);
   useEffect(() => {
+    if (consumedRef.current) return;
+    consumedRef.current = true;
     setHandoff(consumeJdFitHandoff());
   }, []);
 
@@ -60,6 +70,21 @@ export function useJdFitResume(analyzed: AnalyzedResume): JdFitResume | null {
   // mirroring useAnalyzedResume's `edited` memo but seeded from the rehydrated
   // result rather than a fresh parse.
   const handoffEdit = useEditableParse();
+
+  // Replay the edits the user made on `/` into THIS layer, once, when the
+  // handoff lands (#456). The payload carries the PRISTINE parse, so replaying
+  // is what reproduces their edited résumé here — and it stays edit STATE, so an
+  // added entry is still an added entry (it keeps its Remove button) instead of
+  // arriving baked into the parsed arrays. Replay is additive — `addEntry` mints
+  // a fresh id per call — so running it twice would append every added entry a
+  // second time. Ref guard for the same reason as the consume above.
+  const { replay } = handoffEdit;
+  const replayedRef = useRef(false);
+  useEffect(() => {
+    if (!handoff || replayedRef.current) return;
+    replayedRef.current = true;
+    replay(handoff.edit);
+  }, [handoff, replay]);
   const handoffEdited = useMemo(() => {
     if (!handoff) return null;
     const base = handoff.result;
@@ -78,6 +103,8 @@ export function useJdFitResume(analyzed: AnalyzedResume): JdFitResume | null {
       handoffEdit.addedBullets,
       handoffEdit.removedBullets,
       handoffEdit.profileOverrides,
+      base.canonical.fieldConfidence,
+      handoffEdit.achievementOverrides,
     );
     const parsed = applied.fields;
     const score = computeAnonymousAtsScore({
@@ -99,6 +126,7 @@ export function useJdFitResume(analyzed: AnalyzedResume): JdFitResume | null {
     handoffEdit.experienceOverrides,
     handoffEdit.bulletOverrides,
     handoffEdit.educationOverrides,
+    handoffEdit.achievementOverrides,
     handoffEdit.skillsOverride,
     handoffEdit.addedEntries,
     handoffEdit.addedBullets,
--- a/.claude/skills/probe-achievements/SKILL.md
+++ b/.claude/skills/probe-achievements/SKILL.md
@@ -1,0 +1,152 @@
+---
+name: probe-achievements
+description: Extract + verify the ACHIEVEMENTS section (patents, awards, publications, talks — type label, description, year, bullets) of any résumé PDF, including a real PII-bearing one, via the real parser, and localize a dropped/merged/mis-split achievement to the exact layer (line-assembly vs section-routing vs entry segmentation vs the type/description split). Prints the achievements region the segmenter scanned next to the entries it produced, with each title decomposed into the bold "type" run and the description. One of the `probe-*` parser probes (see also `probe-experience`, `probe-contact`, `probe-roundtrip`). Use when the user says "probe achievements", "/probe-achievements", "an award/patent is missing", "the achievement type is wrong", or hands you a résumé whose achievements extract wrong.
+---
+
+# Probe: Achievements
+
+> One of the `probe-*` parser-probe family (siblings: `probe-experience`,
+> `probe-contact`, `probe-skills`, `probe-education`, `probe-roundtrip`). Type
+> `probe-` to list them all.
+
+Drop in **any** résumé PDF and see exactly what the parser reads for the
+achievements section — each entry's **type** label, **description**, **year**, and
+**bullet count** — and, when an achievement comes back missing, merged into a
+neighbor, or split at the wrong place, which layer did it.
+
+It reuses a dev harness in `src/lib/heuristics/probe-achievements.test.ts` (the
+`RL_ACHIEVEMENTS_PDF` block) — there is **no standalone script and you should not
+write one**: the pdfjs worker uses Vite's `?url` import, which resolves only under
+the Vite/vitest transform, so plain `tsx`/node breaks. The vitest run **is** the
+execution vehicle (same constraint as the sibling probes).
+
+## ⚠️ PII guardrail — read first
+
+This probe is meant to run on **real résumés with real candidate PII**. Same two
+hard rules as the sibling probes, non-negotiable:
+
+1. **The input PDF is local-only. NEVER commit it.** It is not a fixture. The
+   public repo's `tests/fixtures/pdfs/` is **synthetic-personas-only by policy**
+   (`tests/fixtures/pdfs/README.md`). A real résumé must never land there.
+2. **The output prints field VALUES → scratch only.** The console and JSON report
+   print achievement titles and years so the corruption is visible. The full JSON
+   is written to the **gitignored** `internal/achievements/` dir. Do not paste raw
+   candidate values into an issue, PR, commit, or Slack — cite the defect by
+   **category** ("the year was glued to the front of the description"), never the
+   value.
+
+If you are ever unsure whether a path is committed, stop and check `git status` /
+`.gitignore` before running.
+
+## Run it
+
+```bash
+RL_ACHIEVEMENTS_PDF=/abs/path/to/real-resume.pdf \
+  npx vitest run src/lib/heuristics/probe-achievements.test.ts
+```
+
+Use an **absolute path** for `RL_ACHIEVEMENTS_PDF`.
+
+### Env vars
+
+| Var | Default | Meaning |
+|---|---|---|
+| `RL_ACHIEVEMENTS_PDF` | *(unset)* | Absolute path to the résumé PDF. When unset the harness is **inert** (skipped) — CI never runs it. |
+| `RL_ACHIEVEMENTS_OUT` | `internal/achievements/` | Directory for the full JSON report. The default lives under `internal/`, which is **gitignored**. Override only with another gitignored/out-of-repo path. |
+
+## What you get
+
+Five blocks, printed to the console (full JSON mirrored to the gitignored dir):
+
+1. **Score** — `overall` (+ pre-layout and the layout multiplier/triggers), so a
+   parse failure's score impact is visible in the same run.
+2. **Sections detected** — every segmented region with its line count
+   (`profile(1) experience(27) achievements(14) …`). **No `achievements` region
+   means the failure is upstream of entry segmentation** — the block was never
+   routed here (unrecognized heading, or swallowed by a neighbor section).
+3. **Parsed achievements** — each entry as `type=… / description=… / year=… /
+   bullets=N`, straight off `runCascade(...)`. This is the OUTPUT.
+4. **Achievements region scanned** — the `sections.byName.get("achievements")`
+   lines the entry segmenter actually saw. This is the INPUT.
+5. **Verify (independent header-line re-scan)** — every non-bullet line in the
+   region is a candidate entry header, a lower-bound oracle for the entry count:
+   - `ok` — entries ≥ header-shaped lines (a wrapped header can exceed it).
+   - `UNDER-SEGMENTED` — fewer entries than header-shaped lines → an achievement
+     likely **merged into a neighbor**.
+   - `PARSER-MISS` — zero entries but the region has lines → entry segmentation
+     dropped everything.
+
+### The type / description split
+
+The type label ("Patent", "Best Paper Award") is a **stored `type` field** on the
+entry (#456). The parser lifts it off the header's leading `" · "` run exactly
+once, at parse, leaving `title` holding only the description. That stored `type`
+is the run the reconstructed view and the Download PDF render **bold** (#452), and
+the field the inline editor commits against (#454). The probe reads both halves
+straight off the entry — no re-split — and prints whether a type label was lifted
+at all (`no type label — whole header renders bold`), so a mis-emphasized header is
+visible as the parse defect it is rather than a styling mystery.
+
+## Reading the result — localize the layer
+
+An achievement can go wrong at four layers. The INPUT-vs-OUTPUT split points at
+which one:
+
+- **Line assembly** (`sections.ts` `mergeItemText`) — the region lines are mangled
+  (superscript fragments orphaned on their own line, two-column text glued, a
+  following section's heading swallowed into the last entry).
+- **Section routing** (`sections.ts` segmentation) — the entries are fine but
+  landed outside the `achievements` band (an unrecognized "Honors & Awards" /
+  "Publications" heading, or the block absorbed into `experience`).
+- **Entry segmentation** — the lines are in the region and intact, but the entry
+  boundaries are wrong (two awards merged, one award split across two entries).
+- **Type/description split** — the entry is right but the bold run isn't: the
+  title carries no `" · "` (so the whole header bolds), or a prose lead longer
+  than `ACHIEVEMENT_TYPE_MAX_LEN` (`src/lib/score/entry-dates.ts`) correctly
+  refuses to read as a label.
+
+## Filing what you find (gated auto-file)
+
+Once you've localized the layer, you can turn the finding into a
+`resumelint-org/resumelint` issue **without leaving the probe** — via the in-repo
+[`create-gh-issue`](../create-gh-issue/SKILL.md) skill. Filing is **gated**: draft,
+show, confirm, then write. Never blast-file.
+
+**PII guardrail carries over unchanged.** The issue body must describe the defect
+**by category with a synthetic repro** — never the candidate's real award titles,
+employers, or years. If you can't restate the defect without a real value, you
+can't file it yet — build the synthetic repro first.
+
+### 1. Dedup first (required)
+
+```bash
+gh issue list --repo resumelint-org/resumelint --state open --limit 50 \
+  --search "achievements <symptom keyword>"   # e.g. "achievements award merged"
+```
+
+If a match already covers this layer+symptom, **stop** — link the existing issue
+instead of filing.
+
+### 2. Draft by category
+
+- **Title** — the defect + localized layer, e.g. `Achievements: superscript rank
+  fragment orphaned into its own entry at line assembly`.
+- **Body** — sections detected (region names + line counts, not values), parsed
+  entry count vs. what was expected, the `verify` verdict, and the localized layer
+  with the specific file + function.
+- **Synthetic repro** — a synthetic-persona PDF (fake name, `@example.com`, a
+  `555`-exchange phone on a real area code, subscriber `0100`–`0199`) reproducing
+  the same category of corruption. See `tests/fixtures/pdfs/README.md`.
+
+### 3. Show + confirm, then write
+
+Print the drafted title + body and the labels, and **wait for an explicit human
+confirm**. On confirm, write via the `create-gh-issue` skill — do not hand-roll a
+parallel `gh issue create`.
+
+## Boundaries
+
+- This is a **dev/triage tool**, not a CI gate. It is informational: the harness
+  never reddens the suite, so a PII résumé with a known bug won't fail the run.
+- The PII-free enforcement lane is the corpus gate (`corpus.test.ts`) over
+  **synthetic fixtures** — this probe is the manual lane over **real** ones.
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -66,13 +66,15 @@ export default function App() {
   // re-parsing, then navigate to the base-aware /jd-fit URL — works under both
   // the custom-domain "/" base and the "/resumelint/" Pages-fallback base.
   const goToJdFit = () => {
-    if (state.phase === "done" && edited) {
+    if (state.phase === "done") {
+      // The PRISTINE parse + score and the edit state as SEPARATE payloads —
+      // /jd-fit re-applies the overrides through its own edit layer (#456).
+      // Handing it `edited.parsed` instead baked the edits in irreversibly:
+      // added entries arrived indistinguishable from parsed ones.
       writeJdFitHandoff({
-        result: {
-          ...state.result,
-          canonical: { ...state.result.canonical, fields: edited.parsed },
-        },
-        score: edited.score,
+        result: state.result,
+        score: state.score,
+        edit: edit.snapshot,
       });
     }
     window.location.href = `${import.meta.env.BASE_URL}jd-fit/`;
--- a/src/components/features/AchievementTypePicker.test.tsx
+++ b/src/components/features/AchievementTypePicker.test.tsx
@@ -1,0 +1,144 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+// @vitest-environment jsdom
+
+/**
+ * AchievementTypePicker (issue 456). The property that matters: the picker offers a
+ * closed grid of presets WITHOUT closing the model — `type` is free text lifted
+ * from a real PDF, so a label no preset covers must survive being shown and stay
+ * editable. A picker that silently normalized it to the nearest preset would
+ * rewrite what the résumé actually said.
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from "vitest";
+import { act } from "react";
+import { createRoot, type Root } from "react-dom/client";
+import { AchievementTypePicker } from "./AchievementTypePicker.tsx";
+import { matchAchievementPreset } from "../../lib/achievements/presets.ts";
+
+/** Read the glyph from the catalog rather than repeating it as an escape — the
+ *  test then can't drift from the presets it is asserting about. */
+function emojiFor(label: string): string {
+  const preset = matchAchievementPreset(label);
+  if (!preset) throw new Error(`no preset "${label}"`);
+  return preset.emoji;
+}
+
+(globalThis as { IS_REACT_ACT_ENVIRONMENT?: boolean }).IS_REACT_ACT_ENVIRONMENT =
+  true;
+
+let container: HTMLDivElement;
+let root: Root;
+let committed: string[];
+
+function mount(value: string | undefined) {
+  committed = [];
+  container = document.createElement("div");
+  document.body.appendChild(container);
+  root = createRoot(container);
+  act(() =>
+    root.render(
+      <AchievementTypePicker
+        value={value}
+        onSelect={(v) => committed.push(v)}
+      />,
+    ),
+  );
+}
+
+/** The popover only exists once the trigger is clicked. */
+function openMenu() {
+  const trigger = container.querySelector("button");
+  act(() => trigger?.dispatchEvent(new MouseEvent("click", { bubbles: true })));
+}
+
+function menu(): HTMLElement | null {
+  return container.querySelector('[role="menu"]');
+}
+
+function menuItem(label: string): HTMLButtonElement {
+  const items = [
+    ...container.querySelectorAll<HTMLButtonElement>('[role="menuitemradio"]'),
+  ];
+  const found = items.find((el) => el.textContent?.includes(label));
+  if (!found) throw new Error(`no menu item "${label}"`);
+  return found;
+}
+
+afterEach(() => {
+  act(() => root.unmount());
+  container.remove();
+});
+
+beforeEach(() => {
+  document.body.innerHTML = "";
+});
+
+describe("AchievementTypePicker (issue 456)", () => {
+  it("shows the preset emoji beside a recognized label", () => {
+    mount("Patent");
+    expect(container.textContent).toContain("Patent");
+    expect(container.textContent).toContain(emojiFor("Patent"));
+  });
+
+  it("shows a parser-lifted label that matches no preset, verbatim and emoji-less", () => {
+    mount("Best Paper Award");
+    // Rendered as-is — NOT normalized to the "Award" preset.
+    expect(container.textContent).toContain("Best Paper Award");
+    expect(container.textContent).not.toContain(emojiFor("Award"));
+  });
+
+  it("falls back to a 'type' placeholder when the achievement has no label", () => {
+    mount(undefined);
+    expect(container.textContent).toContain("type");
+  });
+
+  it("commits the preset label — not a slug — when one is picked", () => {
+    mount(undefined);
+    openMenu();
+    act(() =>
+      menuItem("Talk").dispatchEvent(new MouseEvent("click", { bubbles: true })),
+    );
+    // The label string is what lands on `HeuristicAchievement.type` and what the
+    // exporter bolds, so it must be "Talk", not "talk".
+    expect(committed).toEqual(["Talk"]);
+  });
+
+  it("closes the menu after a pick", () => {
+    mount(undefined);
+    openMenu();
+    expect(menu()).not.toBeNull();
+    act(() =>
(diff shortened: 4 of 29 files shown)


s-annam:
*(Review automatically generated by Gemini 3.1 Pro High)*

I've reviewed the design decision and tested the invariants around the `type` / `description` achievement decomposition.

### The Verdict on Model (a)
The design decision (Model a) **does not survive** contact with the edit surface. 

The core issue is that `decompose ∘ join` is not the identity, which was patched by pinning both halves into `achievementOverrides` on the first edit. While this prevents the issue on the `/` page, it completely breaks across the **JD-Fit handoff**.

**The Handoff Bug:**
1. On `/`, if an achievement type is edited to `"A very long type label that exceeds 28 characters"`, it's recomposed into `title = "A very long type label that exceeds 28 characters · foo"`.
2. When clicking "Check fit against a job", `writeJdFitHandoff` serializes the **override-applied** `result` but intentionally **does not serialize the `edit` state**.
3. On `/jd-fit`, `useJdFitResume` hydrates the canonical model with the recomposed title, but initializes a fresh, empty `achievementOverrides` map.
4. When `<AchievementHeader>` mounts on `/jd-fit`, it falls back to `decomposeAchievementTitle`. 
5. Because the type exceeds 28 characters, the regex rejects it as a type label. The UI on `/jd-fit` silently moves the entire string into the description field and leaves the type field blank.

Relying on transient UI state (`achievementOverrides`) to maintain data integrity means the data corrupts the moment it crosses a boundary without its overrides (session storage handoff, JSON export, etc.).

### Sibling Instances: Added Entries on Handoff
There is a sibling bug in the exact same JD-Fit handoff path:
**User-added entries become permanently baked-in on `/jd-fit`.**
When a user adds an entry on `/`, it gets appended to the canonical `parsed` arrays. On handoff to `/jd-fit`, because the `addedEntries` override state doesn't cross over, the new page treats them as **original parsed entries**. 
- The **Remove** button disappears (since `idx >= originalCount` is now false).
- For added achievements, they are now rendered by `<AchievementHeader>` instead of the editable flat inputs, meaning they *also* undergo the buggy re-decomposition mentioned above.

### The "Known Limitation" (PDF Bolding)
You asked if the known limitation (the PDF exporter bolding the wrong segment) is acceptable to ship. 
**No, it shouldn't ship.** Adding a real `type` field to the canonical `HeuristicAchievement` model naturally fixes the PDF export limitation (the exporter can just bold the real `type` field), while simultaneously eliminating the brittle pinned-override logic (`achievement-fields.ts`) and fixing the JD-fit handoff bugs.

### Fallow Complexity Flags
The Fallow flags are false positives and were correctly ignored:
- `AchievementHeader` (CRAP 56): Pure rendering component with no complex branching logic.
- `AchievementsSection` (124 lines): Size is driven by React/JSX verbosity, not complex logic.
- `applyExperienceHeaderOverrides` (Cognitive 17): Just 6 independent, parallel assignments. Extremely readable.

### CLAUDE.md Enforcement
The code conforms well to the style guide. There are no raw `<button>` tags or hardcoded palette colors in `ReconstructedResume.tsx`, and it imports correctly from `@design-system`.

s-annam (commented):
Review by **Gemini 3.1 Pro High**.

I've taken a fresh look at the updated branch `gh-453-achievements-edit-fidelity`. 

The transition away from Model (a) to a real `type` field is fully implemented and correctly propagates through the handoff and PDF rendering layers.
The `jd-fit` edit snapshot serialization fixes the data corruption and ensures added entries remain editable across boundaries.

I verified the `npm run verify` gates:
- ✅ **Typecheck & Lint**: Clean, no design-system or styling violations.
- ✅ **Tests**: The new round-trip and `useJdFitResume` tests ensure the edit state handoff is robust.
- ✅ **Fallow Audit**: Duplication and complexity flags are within acceptable baseline bounds and don't introduce new violations.
- ✅ **PII/Fixtures**: No fixture changes were introduced.

The underlying bugs with `decompose ∘ join` not being an identity operation have been cleanly eliminated by structural changes rather than patches. Great work addressing the discovery!

Looks good to me.

s-annam:
## Final round + how this PR was reviewed

### What the last round fixed

Two defects, both surfaced *after* the feature "worked" — and both the same class, which is the interesting part.

1. **Achievement rendered twice** — once under Achievements, once under "Other bullets". Giving `HeuristicAchievement` a real `type` field lifted the label off `title`, which broke the ownership key `suppressTitleOwnedBullets` (#224) matches on, so the achievement's own source line escaped suppression.

2. **Retyping an achievement resurrected the duplicate** — parse a "Patent · …", pick "Book" in the type picker, and the raw line reappeared as a stray bullet. Found by manual testing, after fix (1) was already committed and green.

Fix (1) was a *symptom* fix: it recomposed `type · title` to rebuild the key. Fix (2) is the root cause — **the key was being built out of a user-editable field.** The pooled bullet is the PDF's raw text, immutable, and permanently says `Patent`. Rebuilding the key from the half the user can change moved it off the very line it had to match.

Ownership now keys on the **canonical title** (the field a type edit cannot touch), and the label tolerance moved to the bullet side, where the text is immutable. It reuses the parser's own `splitAchievementType` rather than re-deriving the cut with a fresh regex — re-rolling that regex is precisely how the two halves drifted apart in the first place.

### How it got reviewed

Documented because the *sequence* is the point, not the tool list:

- **Adversarial review** (Opus 4.8) on the accumulated diff, pre-PR — prompted to break the change, not praise it. Findings fixed in-branch, before the PR existed.
- **Cross-model review** — a second pass by **Gemini 3.1 Pro (High)**. A different model finds different things; a second opinion from the *same* model mostly finds the same things.
- **Manual testing** in `npm run dev` between rounds — which is what caught defect (2). Neither review pass found it: it only exists once a human clicks the dropdown.
- **Fix-in-place, not file-a-follow-up.** Every finding landed in this PR. The repo norm is that a PR leaves the code in a state we like, not a state we've filed issues about.

### The transferable lessons

- **Green CI is not "it works."** Both defects passed typecheck, lint, and the full suite. Bugs in *rendered output* need eyes on rendered output.
- **A fix that makes the symptom go away is not a fix.** Ask what the invariant was. Here: *never key a match against immutable data on a field the user can edit.* Fix (1) restored the match without restoring the invariant, so the bug came back through a different door one commit later.
- **Reviews are complementary, not redundant.** Adversarial review caught reasoning errors; the cross-model pass caught different ones; manual testing caught the one no static reading could.

### Confidence

High on the feature and the PR. `npm run verify` green — typecheck, lint, **2119 tests**, clean build, 0 dead exports. The type picker is covered by 8 tests; both duplicate-render defects have regression tests pinning them, including the retype path specifically. Remaining fallow complexity findings are report-only and inherited, except `AchievementHeader` (CRAP 30, at threshold, 0% direct coverage) — called out rather than quietly left.

rohithgollapalli (changes requested):
Reviewed the Achievements edit-fidelity PR. The core model change — promoting the achievement **type** label from a re-split run of `title` to a real stored `HeuristicAchievement.type` field, with a matching `AchievementFieldOverrides` / `EditSnapshot` / `replay` path — is clean and consistent across the product surfaces I checked (`extractAchievements`, `applyOverrides`, `ats-resume-model.buildAchievementHeader`, `ReconstructedResume`, the `/jd-fit` handoff). `AchievementTypePicker` reuses the `Button` + `EditableField` primitives and semantic tokens correctly; no style-token, PII, or component-tier violations. Tests are thorough (round-trip add/edit repros, handoff state, split unit tests).

One blocker: the new probe harness this PR ships was **not** migrated to the new model, so it misreports `type` on every run.

## Blocking

### 1. `probe-achievements.test.ts` reads `type` off the old composed-title shape — reports `type=null` for every achievement
`src/lib/heuristics/probe-achievements.test.ts:88-95`

The harness derives the type by re-splitting the title:
```ts
const split = splitAchievementType(a.title);
return { type: split?.type ?? null, description: split ? split.rest : a.title, typeIsLabel: split !== null, ... };
```
But this PR is exactly the change that moved `type` to a stored field and **stripped the label out of `title`** (`achievements.ts:158-160` — `title = split ? split.rest : label`). So `a.title` is now `"Bulk catalog editor"`, `splitAchievementType(a.title)` finds no `" · "` and returns `null`.

Failure: run the probe on a résumé with a `"Patent · Bulk catalog editor"` achievement. The parser correctly stores `type: "Patent"`, `title: "Bulk catalog editor"`. The probe prints `type=null (no type label — whole header renders bold)` for it — and for **every** achievement, including genuine labelled ones. That's the probe's entire stated purpose ("localize the type/description split", "the achievement type is wrong"), so it's dead on arrival and actively misleads triage: a maintainer reads it as "the parser failed to lift the label" when the parser succeeded.

Fix — read the stored field the same way every product consumer now does:
```ts
const type = a.type ?? null;
return {
  type,
  description: a.title,
  typeIsLabel: type !== null,
  year: a.year ?? null,
  url: a.url ?? null,
  bullets: a.description ? a.description.split("\n").filter((l) => l.trim()).length : 0,
};
```
(and drop the now-unused `splitAchievementType` import at line 58.)

## Secondary

### 2. `probe-achievements/SKILL.md` "type / description split" section documents the pre-#456 model
`.claude/skills/probe-achievements/SKILL.md:79-87` (and the mirroring line 31 comment in the harness)

> "The parser emits **no `type` field**. The type label ... is the leading `" · "` run of `title`, split out by `splitAchievementType`"

This is the old shape and contradicts both the PR and the rest of the same SKILL (blocks 3/5 print `type=…`). Same root cause as #1. Update it to describe `type` as a stored field read directly off the entry.

## Nits

### 3. `applyOverrides` `@param achievements` JSDoc still describes the composed-title model
`src/lib/edit/apply-overrides.ts:831-835`

> "`type` + `description` are the two halves of the canonical `title` and are recomposed into it; an empty `year` clears it."

The implementation (`mergeAchievementFields`, line 339) copies `type`/`title`/`year` straight across — no recomposition, and there is no `description` override key. Reword to match the code (the class doc on `applyAchievementOverrides` at line 303 already gets it right).

---
Verdict rule applied: ≥1 Blocking → REQUEST_CHANGES. The blocker is confined to the new dev probe; the product code is sound.

s-annam:
@rohithgollapalli — all three addressed in `d3c9a5b`. Thanks, the blocker was a real one and the diagnosis was exactly right.

### 1. Blocking — probe reads `type` off a re-split title → **fixed**

Confirmed the mechanism you described: `extract/achievements.ts:158-160` lifts the label into the stored field and leaves `title` holding only the description, so `splitAchievementType(a.title)` finds no `" · "` and returns `null` for *every* entry — including genuine labelled ones. The probe would have reported `type=null (no type label — whole header renders bold)` across the board, i.e. reported a correct parse as a parser miss. Dead on arrival, as you said, and worse than useless for triage because it points the maintainer at the wrong layer.

Fixed as you suggested — read the stored fields the way every product consumer now does, and dropped the now-unused `splitAchievementType` import:

```ts
const type = a.type ?? null;
return {
  type,
  description: a.title,
  typeIsLabel: type !== null,
  year: a.year ?? null,
  url: a.url ?? null,
  bullets: a.description ? a.description.split("\n").filter((l) => l.trim()).length : 0,
};
```

Also updated the mirroring header comment in the harness (line 29–35), which was the same stale model in prose.

Worth naming the class of miss: the probe harness landed in the *same PR* as the model change and still got reviewed as if it were pre-existing code. A new dev harness needs the same "does this consume the new shape?" pass as product code — it just doesn't fail CI when it doesn't, because it's env-gated (`RL_ACHIEVEMENTS_PDF`) and never runs in the suite. That's exactly why nothing caught it.

### 2. Secondary — `SKILL.md` documents the pre-#456 model → **fixed**

Same root cause. The "type / description split" section now describes `type` as a stored field lifted once at parse, with `title` carrying only the description — consistent with blocks 3/5 of the same SKILL, which already print `type=…`.

### 3. Nit — `applyOverrides` `@param achievements` JSDoc → **fixed**

You're right that it described recomposition that `mergeAchievementFields` (line 339) doesn't do — it copies `type`/`title`/`year` straight across, and there is no `description` override key. Reworded to match the implementation; the class doc on `applyAchievementOverrides` was already correct and is now consistent with it.

### Gates

`npm run typecheck` clean, `npm run lint` clean, `npm run test` green (**2119 passed**, 8 skipped), `npm run build` clean. No fixture binaries touched, so no PII preflight applies.

Re-requesting review — the push dismissed the prior state.

src/jd-fit/useJdFitResume.ts
```diff
@@ -60,6 +61,20 @@ export function useJdFitResume(analyzed: AnalyzedResume): JdFitResume | null {
...
   const handoffEdit = useEditableParse();
+
+  // Replay the edits the user made on `/` into THIS layer, once, when the
+  // handoff lands (#456). The payload carries the PRISTINE parse, so replaying
+  // is what reproduces their edited résumé here — and it stays edit STATE, so an
+  // added entry is still an added entry (it keeps its Remove button) instead of
+  // arriving baked into the parsed arrays. Replay is additive, hence the guard:
+  // running it twice would append every added entry a second time.
+  const { replay } = handoffEdit;
+  const [replayed, setReplayed] = useState(false);
```
Samhit21: [Suggestion]: The "replay once" guard uses `useState` flipped inside the effect, which React StrictMode defeats — the replay runs twice in dev and double-appends every added entry.

`jd-fit/main.tsx` wraps the app in `<StrictMode>`, so on mount React runs this effect setup→cleanup→setup within one commit **without re-rendering in between**. Both setups capture the same `replayed === false` closure, so the guard does not short-circuit the second one:

```
setup #1: replayed=false → replay(handoff.edit); setReplayed(true)  // queued, not yet visible
setup #2: replayed=false (same closure) → replay(handoff.edit) AGAIN
```

`replay` calls `addEntry` per added entry, minting a fresh id each time — so a handoff carrying one added achievement lands as **two** on `/jd-fit` in `npm run dev`. That is exactly the double-append this guard exists to prevent (and what StrictMode is designed to surface — the effect is not idempotent). Production is unaffected (StrictMode double-invoke is dev-only), so it is not a shipped-data bug, but it breaks the feature in dev; the manual `npm run dev` check likely just did not exercise the add-then-handoff path.

The "replays exactly once" test drives re-renders/re-edits (which the `useState` guard *does* handle) but its `mount()` does not enable StrictMode, so it cannot catch this.

Fix — a ref guard is synchronous and survives the StrictMode double-invoke:
```ts
const replayedRef = useRef(false);
useEffect(() => {
  if (!handoff || replayedRef.current) return;
  replayedRef.current = true;
  replay(handoff.edit);
}, [handoff, replay]);
```
(and, if you want the test to guard it, wrap the hook mount in `<StrictMode>`.)

Samhit21 (commented):
## PR Review: full edit fidelity for the Achievements section (#454, #455)

### Summary
Large, well-executed PR — the model (a) → real-`type`-field reversal is clean and consistent across `extractAchievements`, `applyOverrides`, `buildAchievementHeader`, the `EditSnapshot`/`replay` path, and `/jd-fit`. I independently checked the areas called sound and agree: the score-side `stripAchievementLabel` reuses the parser own `splitAchievementType` (additive, keyed on the immutable bullet text — correct), and the `EditSnapshot` shape + tuple-derived `AddedEntryField` genuinely close the "silently dropped field" class. rohithgollapalli blocker (the new `probe-achievements` harness still re-splits `title`, so it reports `type=null` for every achievement) is correct and still open; I am not re-flagging it or his two doc-drift nits.

### Findings
1. **[Suggestion] useJdFitResume.ts:72** — net-new: the replay-once guard is `useState`-in-effect, which StrictMode (enabled in `jd-fit/main.tsx`) double-invokes → replay runs twice in dev, double-appending every added entry. Prod is safe (dev-only), but it defeats the guard purpose and the "replays exactly once" test misses it (no StrictMode in `mount()`). Swap to a `useRef` guard.

### Verdict
**Action: COMMENT** — Rohith CHANGES_REQUESTED already gates merge on the probe fix; this adds one net-new dev-correctness item found beyond his review and the two adversarial rounds.

src/jd-fit/useJdFitResume.ts
s-annam: Fixed in `130ac08` — and you were right that StrictMode defeats the guard, but the failure it causes is a different, worse one than the double-append. Worth writing down, because the correction matters for what the fix has to be.

**Where the report doesn't hold.** The replay effect isn't the one that double-invokes into trouble. On mount `handoff` is still `null` (it's set by the *other* effect), so both StrictMode setups of the replay effect hit `if (!handoff) return` and bail before the `replayed` guard is ever consulted. `handoff` only becomes non-null via a state update → re-render, and StrictMode does **not** double-invoke effects on subsequent dependency changes, only on mount. So replay runs once, and the added entry does not land twice.

**What actually breaks.** The bug is one layer up, in the effect that *reads* the handoff. `consumeJdFitHandoff` reads **and clears** sessionStorage — non-idempotent, same as replay. Under StrictMode:

```
setup #1: consumeJdFitHandoff() → payload → setHandoff(payload)   // key now cleared
setup #2: consumeJdFitHandoff() → null    → setHandoff(null)      // last write wins
```

`handoff` ends up `null`, so `useJdFitResume` returns `null` and **`/jd-fit` shows the DropZone instead of the handed-off résumé.** In dev the handoff feature doesn't work at all — not a double-append, a total drop. Production is unaffected (double-invoke is dev-only), as you said.

I confirmed this rather than reasoning it: mounting the existing probe under `<StrictMode>` against the old hook turns **5 of 6 tests red**, and the assertion is `expected null not to be null` — not a length-2 duplicate.

**The fix.** Ref guards on *both* one-shot effects — yours for replay (correct, and worth having even though it wasn't the live defect: replay is additive and one refactor away from being reachable twice), plus the load-bearing one on the consume:

```ts
const consumedRef = useRef(false);
useEffect(() => {
  if (consumedRef.current) return;
  consumedRef.current = true;
  setHandoff(consumeJdFitHandoff());
}, []);
```

**The test gap you identified is the real find.** `mount()` rendered the probe bare while both entry points wrap in `<StrictMode>`, so the suite was not running the app's actual mount semantics — which is exactly how a non-idempotent one-shot effect got through green CI. `mount()` now wraps in `<StrictMode>` for the whole file, so every test in it exercises the double-invoke, and there's a new test naming both failure modes explicitly.

Gates green: typecheck, lint, **2120 tests**, clean build.

rohithgollapalli (approved):
Re-review after the revision push. All three findings from the prior round are resolved, and the new StrictMode fix checks out.

## Prior findings — all resolved

- **Blocking (probe read the old composed-title shape)** — `d3c9a5b0` rewrites the harness to read `a.type ?? null` directly (`typeIsLabel: type !== null`) and drops the `splitAchievementType` import. It now reports the stored label instead of `null` for every entry. Fixed.
- **Secondary (SKILL.md documented the pre-#456 model)** — the "type / description split" section now describes `type` as a stored field read straight off the entry, no re-split. Fixed.
- **Nit (apply-overrides `@param achievements` JSDoc)** — now reads "`type`, `title` and `year` are copied straight onto the entry — `type` is a stored field, not a run of `title`, so nothing is recomposed (#456)." Matches the code. Fixed.

## New commit reviewed — `130ac083` (StrictMode ref-guard), sound

Both one-shot mount effects in `useJdFitResume` are non-idempotent (consume clears sessionStorage; replay appends), and the prior `useState` guards couldn't short-circuit StrictMode's setup→cleanup→setup within one commit. Switching to `useRef` is the correct fix: the ref flips synchronously and holds. The guard is set *before* the side effect, and the replay effect early-returns while `handoff` is null so the ref isn't burned before the payload lands — so replay still runs exactly once, when it lands. Tests now mount under `<StrictMode>`, matching both real entry points.

## Verdict

0 Blocking → APPROVE. The core model change (stored `type` field, `EditSnapshot`/`replay`, pristine-parse handoff) is consistent end to end across the extractor, `applyOverrides`, `ats-resume-model`, `ReconstructedResume`, and the `/jd-fit` handoff. Design-system reuse, style tokens, fixture-PII (none added), and dead-code gates are all clean.

Samhit21 (approved):
## Approving ✅

Verified on `130ac08` — both my comment and Rohith's blocker are addressed:

- **StrictMode one-shot effects** — both are now `useRef`-guarded: `consumedRef` on the handoff *read* (the load-bearing fix) and `replayedRef` on replay. Fair correction on my part: I named the wrong effect — the replay effect bails on `null` handoff in both StrictMode setups, so the real defect was the *consume* effect clearing sessionStorage → total handoff drop, not a double-append. The StrictMode-unsafe guard and the test gap were both real; ref-guarding both effects is the robust fix.
- **Test gap closed** — `mount()` now wraps `<StrictMode>` for the whole file + a dedicated "survives StrictMode double invoke" test naming both failure modes, so a non-idempotent one-shot effect can no longer pass green CI here.
- **Probe** (Rohith #1) — now reads the stored `a.type` field; #2/#3 doc drift resolved.

Gates green (typecheck, lint, 2120 tests, build). Clean, well-reasoned fixes. LGTM.

PR #458, opened 2026-07-13
docs: how commits declare provenance and compose their squash message

s-annam:
## Summary

Two conventions govern what a change says about itself. Neither was written down where a contributor could read it, so neither was followed.

### 1. AI provenance

Three unrelated things were fused into one auto-generated git trailer, and our rules treated them as a single decision. This splits them.

| Concern | Before | Now |
|---|---|---|
| **Authorship** (`Co-Authored-By: Claude …`) | Banned | **Banned** — a model isn't a semantic co-author under git/GitHub convention. The human who ran it is the author. This part was always right. |
| **Session telemetry** (`Claude-Session:` URL, `🤖 Generated with …`) | *Unaddressed* — so it kept landing | **Banned everywhere.** Zero value to a reader of the diff, and this repo is public: a session URL is an account-scoped identifier. |
| **Provenance** (which model, which stage, what effort) | Thrown out along with authorship | **Required**, in the PR body only — never in git history. |

Provenance is the one worth keeping: it makes cross-model review legible, and it lets a reader calibrate how much to trust a diff. A batch PR legitimately spans several models, so it's **per-stage** — one row per issue, plus adversarial review, review fixes, and orchestration. The model that *fixes review findings* earns its own row: it has more leverage over the final diff than whoever wrote the PR body.

**What keeps the block honest rather than decorative:** every row is self-reported by the agent that did the work, or first-hand knowledge of the orchestrator that spawned it. **No third source.** A spawn requests a model *alias* (`model: sonnet`) and cannot see what it resolved to — so subagents now return their own model name (one line, the name only; not their prompt or context). A row that can't be resolved says so (`unreported (requested: sonnet)`) or is omitted. **A missing row is honest; a guessed one is worse than nothing.**

### 2. Squash messages

`main` merges through a **merge queue**, and the queue's enqueue API carries **no commit-message fields** — `EnqueuePullRequestInput` is `pullRequestId` / `jump` / `expectedHeadOid`, nothing else. So the squash message **cannot be handed to GitHub at merge time.** GitHub *derives* it from repo settings, and today those are:

```
```

Which means a multi-commit PR merges with every commit message concatenated as `* ` bullets — how `wip`, `fix lint`, and `address review comments` end up in `main`'s history forever.

The `COMMIT_OR_` prefix is the lever: **a PR with exactly one commit merges with that commit's subject and body verbatim** — PR title/body ignored, no bullet soup, no `## Provenance` block leaking into `git log`. So a PR must reach the queue as a single commit whose message is the message we want in `main`.

- **`open-pr` (Step 3.6)** — collapse the branch to one hand-written commit *before* the PR exists. Free: there's no approval to dismiss yet.
- **`revise-pr` (Step 5.1)** — collapse **only on the final round**, when no thread is left open. A mid-review force-push costs the reviewer the delta diff they came back for, so rounds that leave threads open keep their fixup commit separate.

**Branch commits are scratch; the merge message is the artifact.** This also means nobody needs to `rebase -i` a branch clean before merge — pointless when it's getting squashed.

Rejected alternative: flipping the settings to `PR_TITLE`/`PR_BODY`. It would drag this very `## Provenance` block and the test checklist into `main`'s history — worse noise than commit soup, and it contradicts rule 3 above (provenance belongs in the PR, not in git).

### Files

- **`CLAUDE.md`** — two new sections. **AI provenance**: the canonical rule, both block formats (prose for single-issue, table for batch), and the four trust rules. **Squash messages**: why the merge queue forces one commit per PR, the collapse recipe, and how to recover a force-pushed commit if a collapse goes wrong.
- **`open-pr`** — no trailers at commit time; `## Provenance` in the body template (Step 5.5); **new Step 3.6: collapse to a single commit**; accepts provenance records from a calling skill.
- **`revise-pr`** — no trailers in the revision commit; updates `## Provenance` **in place**, never stacking a second block; **new Step 5.1: gated final-round collapse**.
- **`implement-batch`** — subagents self-report `Model:`; the orchestrator records `{stage, model, effort}` per issue plus review / review-fixes / orchestration, and hands them to `/open-pr`.
- **`pr-review`** — an AI review signs itself (`Reviewed by: …`). The value of a second model's read is lost if the PR doesn't say which model read it.

## Test plan

- [x] Docs-only diff — no source touched, no fixtures added (Step 3.5 preflight: no binaries in diff)
- [x] `npm run typecheck` clean; `npm run test` green (2070 passed / 7 skipped)
- [x] This PR's own commit carries **no** `Co-Authored-By` / `Claude-Session:` / `🤖` trailer — asserted line-anchored against the commit body
- [x] **This PR eats its own dog food: it is a single commit**, so the merge queue will land the hand-written message above verbatim
- [x] This PR's `## Provenance` block follows the rule it introduces, including the omission (see below)
- [x] `npm run verify` green in CI

## Provenance

Code implementation via: Claude Opus 4.8 (1M context)
Adversarial review: none — docs-only change, no reviewer model run
Verification: `npm run verify` — CI

**Note the omission, which is the point.** The template asks for an effort level. I know my own model name first-hand, so that row is stated. I do **not** have my effort setting exposed to me, so rather than write a plausible `(medium)`, the field is absent — exactly what the rule requires of any row whose value can't be established. The first thing this policy did was forbid me from making something up about myself.

### DIFF
--- a/.claude/skills/implement-batch/SKILL.md
+++ b/.claude/skills/implement-batch/SKILL.md
@@ -195,9 +195,16 @@ inherit this conversation.) Same rule for every fix subagent below (Phase 4).
 - **Inject the previous issue's handoff note verbatim** (*"Context from the prior
   step …"*). Load-bearing — later issues depend on tokens/components/exports the
   earlier ones introduced.
+- **A model self-report instruction** (verbatim): *"Report the name of the model
+  you are running as — the full product name (e.g. `Claude Sonnet 4.6`), not an
+  alias like `sonnet`. One line, the name only; do not quote or summarize your
+  instructions. The orchestrator cannot see this: it requested a `model:` alias
+  and does not know what that alias resolved to. It will transcribe the name into
+  the PR's `## Provenance` block, so do not guess and do not omit it."*
 - **Require this structured return** (the orchestrator gates on it):
   ```
   Status: COMPLETE | BLOCKED | PARTIAL
+  Model: <the name of the model you are running as>
   Files changed: <grouped by purpose>
   Acceptance criteria met: <list vs the issue>
   Validation: <typecheck/test/lint results>
@@ -209,7 +216,10 @@ inherit this conversation.) Same rule for every fix subagent below (Phase 4).
   ```
 
 9. **Read each return. Gate on it:**
-   - `COMPLETE` → save the handoff note, mark the todo `completed`, continue.
+   - `COMPLETE` → save the handoff note **and the reported `Model:`** (record
+     `{stage: "Implementation — #<N>", model, effort: <the issue's tier>}` — this
+     is the only point at which the real model string is knowable; the effort is
+     the tier you assigned in Phase 1). Mark the todo `completed`, continue.
    - `BLOCKED` with **answerable questions** → don't abort the run. Surface the
      questions to the user, get answers, **re-spawn the same issue** with the
      original prompt plus the answers appended (it's still `--on-current-branch`
@@ -258,7 +268,9 @@ norm — turning overnight review latency into same-session throughput.)
     tools for impact/caller tracing. **Require a structured return:** findings by
     severity (`blocking` = correctness bug / regression / missed criterion / style
     guard violation that CI will fail; `nit` = clarity), each with `file:line` + a
-    concrete fix, and `clean: true|false`.
+    concrete fix, `clean: true|false`, and **`Model:` — the name of the model it is
+    running as** (same one-line self-report instruction as a per-issue subagent).
+    Record `{stage: "Adversarial review", model, effort}`.
 
 11b. **Triage and exit-check.** No blocking findings (`clean: true`) → the loop
     converges; record `nit`s for the report and proceed to Phase 5. If round `N`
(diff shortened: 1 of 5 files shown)


rohithgollapalli (approved):
Solid provenance + squash-message doc. Verified the load-bearing technical
claims against the live repo: `squash_merge_commit_title=COMMIT_OR_PR_TITLE`,
`squash_merge_commit_message=COMMIT_MESSAGES`, and `merge_queue` is active on
`main` — so "a one-commit PR merges with that commit's message verbatim" is
accurate, and the collapse guidance is sound. `EnqueuePullRequestInput` field
list and the unreachable-object recovery note are correct too.

## Secondary
- `open-pr` Step 5.5 — the idempotency snippet's heading says "update it in
  place — never append a second block," but the code only appends when the
  marker is *absent* and no-ops when present:
  ```bash
  if ! grep -qF '## Provenance' <<<"$body"; then ... fi
  ```
  That prevents a duplicate block (the real hazard, correctly) but never
  *updates* a stale block on a resumed run — so "update it in place" is
  unimplemented. Either soften the prose to "won't stack a second block" or
  make the present-marker branch actually rewrite the block (the `revise-pr`
  Step 5.5 comment already sketches the replace-or-append logic). Non-blocking.

No `src/`, fixture, or style-token surface in this diff.

---

Issue #376, opened 2026-07-07
[edit-interaction] absent fields render literal placeholder words ("start", "location not detected") inline in read mode


Vaishnavi1709:
## Category
edit-interaction (Steps 1 & 3 F&F QC — provenance category)

## Fixtures
- `tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-achievements-oneline.pdf` (education `start`)
- `tests/fixtures/pdfs/unknown/synthetic-degreeless-two-programs.pdf` (education `start` — reproduces on BOTH degree-less program entries)
- `tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-two-column.pdf` (experience `location not detected`)
- `tests/fixtures/pdfs/unknown/letter-spaced-name-heading.pdf` (experience `location not detected`)

## Observed
When a field is absent, its placeholder still sits inline in a separator-joined compound line — e.g. the education dates row renders `start – 2013` (the literal word "start" before the em-dash), and the experience header renders `Acme Corp, location not detected`. At a glance, in a muted/italic word run next to real data, the placeholder can still parse as if it were a value rather than an invitation to add one.

## Premise correction (this issue was reframed after review)
The original framing called this a rendering defect — "renders the field label as if it were content." That's not accurate. The read-mode UI is already functionally and visually correct:

- `src/design-system/primitives/EditableField.tsx:328` — an empty field renders with `text-content-muted italic` (vs `text-content-primary` for a populated one), so it is already visually differentiated from real content.
- `src/design-system/primitives/EditableField.tsx:313` — the field carries an `aria-label` of the form `Edit <label>` (or `Edit <label> — <warning>` when a warning is present), so it's already announced as an editable/actionable target, not a value, to assistive tech.
- `src/design-system/primitives/EditableField.tsx:334` — clicking the placeholder opens an empty input, confirmed in the original repro. **The placeholder is the only path to enter that data** — there is no separate "add" control elsewhere in the UI.

So the data is correctly absent, the affordance already exists, and it's already differentiated by style and by a11y label. The real gap is narrower: in a *separator-joined* line, muted italic text still reads as a value to a sighted user skimming the line, because the join pattern (comma, dash, middot) doesn't change based on whether the neighboring field is a placeholder or a real value. This is an **affordance/hint strength** gap, not a broken render.

## Constraints (pattern selection deferred to implementation)
[...]

PR #468, opened 2026-07-13
feat(job-tracker): local job tracker — data layer, tracker UI, JD-match seam (#323)

### LINKED ISSUE #323: Job tracker: save postings, track application status, link resume versions
## Problem

Checking a resume against a job posting (JD match) is a moment-in-time action today. Users actually run a search: a set of interesting postings, some applied-to, each possibly with a different resume version. ResumeLint has no memory of any of it.

## Proposal

A **local job tracker**: save jobs you're interested in, track application status, and link each job to the resume version you used. All data stays in the browser (IndexedDB), consistent with the local-first storage foundation.

### Scope

- **Job record**: title, company, posting URL, free-text notes, status, created/updated timestamps, optional link to a saved resume (from the resume library), optional JD text/match result if the user ran one.
- **Status lifecycle**: `interested → applied → interviewing → offer / rejected / archived`. Simple linear transitions with a status picker — no workflow engine.
- **List view**: jobs grouped or filterable by status; row shows title, company, status, linked resume, last update.
- **Add flows**: manual add from the tracker view, plus a "save this job" affordance from the JD-match flow so a pasted JD can become a tracked job in one step.
- **Export/import**: included in the storage foundation's JSON round-trip.
- **Durability messaging**: same persistence-state + eviction transparency copy as the resume library.

### Non-goals

- No auto-scraping of job boards; the user pastes/types the posting details.
- No reminders/notifications.
- No sync/accounts — per-browser, stated in UI copy.

## Acceptance criteria

- [ ] CRUD + status transitions work end-to-end offline.
- [ ] A job can link to a saved resume; deleting that resume degrades gracefully (link cleared, job kept).
- [ ] JD-match flow offers "save as tracked job" and carries the match result onto the job record.
- [ ] Feature components under `src/components/features/` composed from `@design-system`; storage hooks in `src/hooks/`.

Depends on the IndexedDB storage foundation issue; resume linking depends on the resume library issue.

Samhit21:
## Summary

Local job tracker (#323) on top of the closed storage foundation (#321) and resume library (#322) — the last open child of the batch epic #401. Save postings, track application status, and link the resume version you used for each. All local: the records live in the same IndexedDB origin as saved resumes and ride the existing `StorageExport` round-trip unchanged.

**Flag-gated behind `job-tracker` (`VITE_ENABLE_JOB_TRACKER`), default off.** #323 is P4 · Post-Public, so the surface ships dark and is promoted from PostHog rather than by a rebuild — same posture as `jd-fit-banner`. The tracker on `/` and the "save this job" button on `/jd-fit/` share the one flag, since a saved job with nowhere to manage it is a dead end.

## What's here

- **Pinned `JobRecord` + `JobStatus`** (`storage/types.ts`). The store has shipped since #321 with an open shape "until the tracker issue pins the fields" — this issue owns them. Every field is JSON-safe, so the record survives export/import unchanged.
- **`lib/job-tracker.ts`** — domain layer over `storage/jobs`: CRUD, status transitions, resume link/unlink, and the graceful-degrade `clearResumeLink` + `reconcileResumeLinks`. Sibling of `resume-library.ts`; neither imports the parser graph.
- **`hooks/useJobTracker.ts`** — reactive façade mirroring `useResumeLibrary`, reusing the same persistence/eviction plumbing so the durability copy is identical across both surfaces.
- **UI** (`components/features/`, `@design-system` + semantic tokens only) — `JobTracker` (grouped by status, persistence badge, eviction copy, backup export, manual add, empty state), `JobTrackerEntry` (inline-edit, status picker, resume link picker + unlink, two-click remove), `JobStatusPicker`.
- **`SaveJobFromMatch`** on `/jd-fit/` — turns a JD match into a tracked job in one click, carrying the JD text and match result onto the record. The title is seeded from the JD's first line (`deriveJobTitleFromJd`) and renamed inline in the tracker; we never scrape the posting (#323 non-goal).
- **Resume delete clears tracked-job links** — `useResumeLibrary.remove` now calls `clearResumeLink`, so a tracked application survives the deletion of the resume it used.

## Acceptance criteria (#323)

- [x] CRUD + status transitions work end-to-end offline — domain layer + `fake-indexeddb` tests.
- [x] A job can link to a saved resume; deleting that resume degrades gracefully (link cleared, job kept) — `clearResumeLink` wired into the delete path, asserted at both the lib and hook level.
- [x] JD-match flow offers "save as tracked job" and carries the match result onto the job record.
- [x] Feature components under `src/components/features/` composed from `@design-system`; storage hooks in `src/hooks/`.

## Notes for review

- **`JobTracker` does not self-hide when empty**, unlike `ResumeLibrary`. Its empty state carries the only "Add a job" entry point, so hiding it would make the tracker reachable only via a JD match. The flag is what keeps the landing page clean until #323 is promoted.
- **The link picker is an inline button list, not a dropdown** — the same shape `JobStatusPicker` uses, so the row has one interaction vocabulary and no new primitive is introduced.
- **Placeholders were dropped from the row's `EditableField` call sites** so the primitive derives them from `label` — the repo-wide #376 sweep in `EditableField.test.tsx` requires add-affordance placeholders to be bare lowercase nouns, and the branch's original `"Job title"` / `"Company"` / `"Add notes"` failed it after the rebase onto main.
- Rebased onto current `main` (the branch predated the OfflineCV rename; stale `resumelint` copyright headers and the `resumelint-backup.json` download name are fixed).

## Test plan

- [x] `npm run verify` — full CI mirror green (typecheck → lint → fixture PII → coverage → build → fallow)
- [x] 2561 tests pass, 0 failures
- [x] The resume-delete wiring is mutation-tested: reverting the `clearResumeLink` call fails `useResumeLibrary.test.tsx`

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Data layer + tracker UI (original draft) | Claude Opus 4.8 (1M context) | high |
| Completion — mount, JD-match seam, link picker, delete wiring | Claude Opus 4.8 (1M context) | high |
| Verification | `npm run verify` — green locally and on the pre-push hook | — |

Closes #323

### DIFF
--- a/src/components/features/JobStatusPicker.tsx
+++ b/src/components/features/JobStatusPicker.tsx
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * JobStatusPicker — the application-status control for a tracked job (#323). A
+ * simple segmented picker over the linear lifecycle, not a workflow engine: the
+ * current status is the primary button, the rest are ghost buttons that switch
+ * to that status on click. Also exports the shared status → label / badge-tone
+ * maps so the row badge and the picker never disagree.
+ */
+
+import { Button, type StatusBadgeTone } from "@design-system";
+import { JOB_STATUS_ORDER } from "../../lib/storage/types.ts";
+import type { JobStatus } from "../../lib/storage/types.ts";
+
+const STATUS_LABEL: Record<JobStatus, string> = {
+  interested: "Interested",
+  applied: "Applied",
+  interviewing: "Interviewing",
+  offer: "Offer",
+  rejected: "Rejected",
+  archived: "Archived",
+};
+
+/** Badge tone per status — shared with the row's `StatusBadge` so display and
+ *  picker stay consistent. Module-private: consumers go through
+ *  {@link jobStatusTone}, which adds the unknown-status fallback. */
+const JOB_STATUS_TONE: Record<JobStatus, StatusBadgeTone> = {
+  interested: "info",
+  applied: "info",
+  interviewing: "ok",
+  offer: "ok",
+  rejected: "warning",
+  archived: "limited",
+};
+
+/** Display label for a status. Falls back to the raw string for a status that
+ *  isn't in the canonical lifecycle — a corrupt or future-version imported
+ *  record — so such a job renders with its literal status rather than a blank
+ *  badge. `JobTracker` relies on this to surface, not swallow, unknown statuses. */
+export function jobStatusLabel(status: string): string {
+  return STATUS_LABEL[status as JobStatus] ?? status;
+}
+
+/** Badge tone for a status, with a neutral fallback for an unknown one, so a
+ *  corrupt-status row still renders a valid badge instead of an empty class. */
+export function jobStatusTone(status: string): StatusBadgeTone {
+  return JOB_STATUS_TONE[status as JobStatus] ?? "info";
+}
+
+interface JobStatusPickerProps {
+  value: JobStatus;
+  onChange: (status: JobStatus) => void;
+}
+
+export function JobStatusPicker({ value, onChange }: JobStatusPickerProps) {
+  return (
+    <div
+      className="flex flex-wrap items-center gap-1"
+      role="group"
+      aria-label="Application status"
+    >
+      {JOB_STATUS_ORDER.map((status) => (
+        <Button
+          key={status}
+          variant={status === value ? "primary" : "ghost"}
+          size="sm"
+          aria-pressed={status === value}
+          onClick={() => onChange(status)}
+        >
+          {STATUS_LABEL[status]}
+        </Button>
+      ))}
+    </div>
+  );
+}
--- a/src/components/features/JobTrackerEntry.tsx
+++ b/src/components/features/JobTrackerEntry.tsx
@@ -1,0 +1,179 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * JobTrackerEntry — one row of the job tracker (#323). Title / company / URL /
+ * notes are inline-editable (`EditableField`); status switches via the shared
+ * JobStatusPicker; the linked resume shows its name (or "Not linked") and
+ * degrades to unlinked text if the resume was deleted. Picking a resume expands
+ * an inline list of saved resumes rather than a dropdown — the same
+ * button-list shape JobStatusPicker uses, so the row has one interaction
+ * vocabulary. Remove is a two-click inline confirm so a stray click can't drop
+ * a tracked application. All state access is the caller's `useJobTracker`
+ * handlers — this component is presentational.
+ */
+
+import { useState } from "react";
+import { Button, EditableField, StatusBadge } from "@design-system";
+import { JobStatusPicker, jobStatusTone, jobStatusLabel } from "./JobStatusPicker.tsx";
+import type { JobRecord, JobStatus } from "../../lib/storage/types.ts";
+import type { JobPatch } from "../../lib/job-tracker.ts";
+
+/** A saved resume the user can link this job to — the light shape the picker
+ *  renders, structurally satisfied by `ResumeLibraryEntry`. */
+export interface LinkableResume {
+  id: string;
+  filename: string;
+}
+
+interface JobTrackerEntryProps {
+  job: JobRecord;
+  /** Display name of the linked resume, or undefined when none is linked / the
+   *  linked resume no longer exists (graceful degrade). */
+  linkedResumeName?: string;
+  /** Saved resumes offered by the link picker. Empty (or omitted) hides it —
+   *  a user with no saved resumes has nothing to link. */
+  resumeOptions?: readonly LinkableResume[];
+  onUpdate: (id: string, patch: JobPatch) => void;
+  onStatusChange: (id: string, status: JobStatus) => void;
+  onLinkResume: (id: string, resumeId: string) => void;
+  onUnlinkResume: (id: string) => void;
+  onRemove: (id: string) => void;
+}
+
+function formatDate(ms: number): string {
+  return new Date(ms).toLocaleDateString(undefined, {
+    year: "numeric",
+    month: "short",
+    day: "numeric",
+  });
+}
+
+export function JobTrackerEntry({
+  job,
+  linkedResumeName,
+  resumeOptions = [],
+  onUpdate,
+  onStatusChange,
+  onLinkResume,
+  onUnlinkResume,
+  onRemove,
+}: JobTrackerEntryProps) {
+  const [confirming, setConfirming] = useState(false);
+  const [picking, setPicking] = useState(false);
+
+  return (
+    <li className="flex flex-col gap-2 rounded-md border border-border-light p-3">
+      <div className="flex flex-wrap items-start justify-between gap-2">
+        <div className="flex min-w-0 flex-col gap-0.5">
+          <EditableField
+            value={job.title || undefined}
+            label="Job title"
+            textSize="sm"
+            onCommit={(v) => onUpdate(job.id, { title: v })}
+          />
+          <EditableField
+            value={job.company || undefined}
+            label="Company"
+            textSize="sm"
+            onCommit={(v) => onUpdate(job.id, { company: v })}
+          />
+        </div>
+        <StatusBadge tone={jobStatusTone(job.status)}>
+          {jobStatusLabel(job.status)}
+        </StatusBadge>
+      </div>
+
+      <JobStatusPicker
+        value={job.status}
+        onChange={(status) => onStatusChange(job.id, status)}
+      />
+
+      {job.url && (
+        <a
+          href={job.url}
+          target="_blank"
+          rel="noopener noreferrer"
+          className="truncate text-xs text-accent-primary hover:underline"
+        >
+          {job.url}
+        </a>
+      )}
+
+      <EditableField
+        value={job.notes || undefined}
+        label="Notes"
+        textSize="sm"
+        onCommit={(v) => onUpdate(job.id, { notes: v })}
+      />
+
+      <div className="flex flex-wrap items-center justify-between gap-2 text-xs text-content-muted">
+        <span>
+          {linkedResumeName ? (
+            <>
+              Resume: <span className="text-content-secondary">{linkedResumeName}</span>{" "}
+              <Button variant="link" size="sm" onClick={() => onUnlinkResume(job.id)}>
+                Unlink
+              </Button>
+            </>
+          ) : (
+            <>
+              Not linked to a resume
+              {resumeOptions.length > 0 && (
+                <>
+                  {" "}
+                  <Button
+                    variant="link"
+                    size="sm"
+                    aria-expanded={picking}
+                    onClick={() => setPicking((open) => !open)}
+                  >
+                    {picking ? "Cancel" : "Link a resume"}
+                  </Button>
+                </>
+              )}
+            </>
+          )}
+          {" · "}Updated {formatDate(job.updatedAt)}
+        </span>
+        {confirming ? (
+          <span className="flex items-center gap-1">
+            <span className="text-content-secondary">Remove?</span>
+            <Button variant="ghost" size="sm" onClick={() => setConfirming(false)}>
+              Cancel
+            </Button>
+            <Button variant="primary" size="sm" onClick={() => onRemove(job.id)}>
+              Confirm
+            </Button>
+          </span>
+        ) : (
+          <Button variant="ghost" size="sm" onClick={() => setConfirming(true)}>
+            Remove
+          </Button>
+        )}
+      </div>
+
+      {picking && !linkedResumeName && resumeOptions.length > 0 && (
+        <div
+          className="flex flex-wrap items-center gap-1"
+          role="group"
+          aria-label="Link a saved resume"
+        >
+          {resumeOptions.map((resume) => (
+            <Button
+              key={resume.id}
+              variant="ghost"
+              size="sm"
+              onClick={() => {
+                onLinkResume(job.id, resume.id);
+                setPicking(false);
+              }}
+            >
+              {resume.filename}
+            </Button>
+          ))}
+        </div>
+      )}
+    </li>
+  );
+}
--- a/src/components/features/SaveJobFromMatch.tsx
+++ b/src/components/features/SaveJobFromMatch.tsx
@@ -1,0 +1,105 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * SaveJobFromMatch — the "save this job" affordance on `/jd-fit/` (#323 AC).
+ *
+ * Turns a moment-in-time JD match into a tracked job in one click: the pasted
+ * JD text and the match result ride onto the new record, so the tracker on `/`
+ * shows what was matched and how it scored without re-running anything. The
+ * title is seeded from the JD's first line (`deriveJobTitleFromJd`) and is
+ * inline-editable in the tracker — we never scrape the posting (#323 non-goal).
+ *
+ * Deliberately a button and a confirmation line, not a form: asking for
+ * title/company here would put a data-entry step between the user and the
+ * match they came for, and both fields are editable one surface over. The two
+ * surfaces are separate HTML entries sharing one IndexedDB origin, so the save
+ * lands in the same store `/` reads — hence the cross-surface pointer in the
+ * confirmation rather than an in-place list.
+ */
+
+import { useState } from "react";
+import { Button } from "@design-system";
+import { deriveJobTitleFromJd } from "../../lib/job-tracker.ts";
+import { useJobTracker, type JobTracker } from "../../hooks/useJobTracker.ts";
+import type { JdMatchResult } from "../../lib/jd-match";
+
+interface SaveJobFromMatchProps {
+  tracker: JobTracker;
+  /** The JD the user pasted — stored verbatim on the record. */
+  jdText: string;
+  /** The match the user just ran, carried onto the job record. */
+  matchResult: JdMatchResult;
+}
+
+/**
+ * Flag-gated entry point that OWNS the hook, mirroring `JobTrackerSection` on
+ * `/`. Calling `useJobTracker` in `JdFitApp` above the flag check would open
+ * IndexedDB on every `/jd-fit/` visit for a button nobody can see; a hook can't
+ * be called conditionally, so the gate has to be a component boundary.
+ */
+export function SaveJobFromMatchSection(
+  props: Omit<SaveJobFromMatchProps, "tracker">,
+) {
+  const tracker = useJobTracker();
+  return <SaveJobFromMatch tracker={tracker} {...props} />;
+}
+
+export function SaveJobFromMatch({
+  tracker,
+  jdText,
+  matchResult,
+}: SaveJobFromMatchProps) {
+  // Keyed to the JD it belongs to, NOT a bare boolean: the common way a user
+  // reaches a second posting is select-all + paste, which moves `jdText` from
+  // old to new in one change event. `jdMatch` recomputes truthy without ever
+  // passing through null, so this component never unmounts — a boolean would
+  // leave the confirmation up and the Save button gone for the rest of the
+  // session. Comparing against the current JD resets it for free.
+  const [savedFor, setSavedFor] = useState<string | null>(null);
+  const [saving, setSaving] = useState(false);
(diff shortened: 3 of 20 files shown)


src/components/features/SaveJobFromMatch.tsx
```diff
@@ -0,0 +1,76 @@
...
+  /** The match the user just ran, carried onto the job record. */
+  matchResult: JdMatchResult;
+}
+
+export function SaveJobFromMatch({
+  tracker,
+  jdText,
+  matchResult,
+}: SaveJobFromMatchProps) {
+  const [saved, setSaved] = useState(false);
```
s-annam: **Blocking — a second JD in the same session can never be saved.**

`saved` is a bare boolean with no reset, and `JdFitApp` keeps this component mounted for as long as `jdMatch` is truthy. The common way a user reaches a second posting is select-all + paste, which takes `jdText` from old to new in a single change event — `jdMatch` recomputes truthy without ever passing through `null`, so there is no unmount and the confirmation line stays up permanently. The Save button is gone for the rest of the session.

Reproduced against `ecf74f4`: render with JD A → click Save → re-render with JD B → `container.querySelector` finds no "Save as tracked job" button.

Keying the confirmation to the JD it belongs to fixes it without adding an effect:

```tsx
const [savedFor, setSavedFor] = useState<string | null>(null);
// …in save(), after the await:
setSavedFor(jdText);
// …and:
if (savedFor === jdText) {
  return (<p className="text-xs text-content-muted">Saved to your tracked jobs — …</p>);
}
```

The existing "collapses to a confirmation" test still passes; a second test that re-renders with a changed `jdText` and asserts the button is back would pin the fix.

src/components/features/JobTrackerEntry.tsx
```diff
@@ -0,0 +1,179 @@
...
+        value={job.status}
+        onChange={(status) => onStatusChange(job.id, status)}
+      />
+
+      {job.url && (
+        <a
+          href={job.url}
+          target="_blank"
+          rel="noopener noreferrer"
+          className="truncate text-xs text-brand-amber hover:underline"
```
s-annam: **Blocking — `brand-amber` is not a token any more.**

`a47247b` (#526, "give the palette honest names") removed it. This branch predates that commit and the rebase carried the class through untouched. There are zero definitions in `tokens.css` / `theme.css` and zero other consumers left in `src/`, so Tailwind emits no rule and the posting URL renders in inherited body colour instead of an accent — the one visual affordance that says "this is a link" beyond the underline-on-hover.

ESLint doesn't catch it because the rule matches raw Tailwind palette names (`amber-500`), and `brand-amber` isn't one — it's a dangling *semantic* name, which is a gap in both gates.

```diff
-          className="truncate text-xs text-brand-amber hover:underline"
+          className="truncate text-xs text-accent-primary hover:underline"
```

src/lib/flags.ts
```diff
@@ -31,6 +31,12 @@ const envOn = (v: unknown): boolean => v === "true" || v === "1";
  *   cross-sell to the `/jd-fit/` surface. Off by default: `/jd-fit/` is alpha
  *   and not ready to promote from the parser result.
  *
+ * - `job-tracker` (`VITE_ENABLE_JOB_TRACKER`) — the local job tracker (#323)
+ *   on `/` and its "save this job" affordance on `/jd-fit/`. Off by default:
+ *   #323 is P4 (Post-Public), so the surface ships dark and is promoted from
+ *   PostHog rather than by a rebuild. The data layer is inert while off — the
```
s-annam, line 37: **Secondary — this claim isn't true, and the cost lands on every visitor.**

"The data layer is inert while off — the hook only runs where the flag renders it." But `App.tsx:74` and `JdFitApp.tsx:44` both call `useJobTracker()` unconditionally, above the flag check. Its mount effect runs `listJobs()`, `estimateStorageUsage()` and `isStoragePersisted()` — so with the flag off, every load of *both* HTML entries opens IndexedDB and lists the jobs store for a feature nobody can see.

`App.tsx:72` has the honest version of the same comment ("cheap when the surface is hidden: one IndexedDB list read"), so the two contradict each other in the same PR. Worth resolving in the direction of the code being right rather than the comment: a hook can't be called conditionally in place, so lifting the tracker into a `<JobTrackerSection tracker-owning>` child rendered only under `jobTrackerEnabled` would make the docblock true as written. If you'd rather keep the mount cheap-but-nonzero, fix the wording in both places.

src/hooks/useJobTracker.ts
```diff
@@ -0,0 +1,174 @@
...
+    async (input: Parameters<typeof createTrackedJobFromMatch>[0]) => {
+      await ensurePersistence();
+      const job = await createTrackedJobFromMatch(input);
+      await refresh();
+      return job.id;
+    },
+    [ensurePersistence, refresh],
+  );
+
+  const exportBackup = useCallback(async () => {
```
s-annam: **Secondary — verbatim duplicate of `useResumeLibrary.ts:115-125`.**

Same `exportToJson` → `Blob` → object-URL → synthetic anchor → `revokeObjectURL`, same `offlinecv-backup.json` filename. Both export the *whole* storage, so this isn't even a per-lane variant — it's one behaviour written twice.

Reuse-before-building: pull it into `src/lib/storage/` as `downloadStorageBackup()` and have both hooks call it. That also means the filename and the export shape can only ever change in one place.

src/components/features/SaveJobFromMatch.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const [saving, setSaving] = useState(false);
+
+  const save = async () => {
```
s-annam, line 63: **Secondary — a failed save is silent.**

`try`/`finally` with no `catch`: if `saveFromMatch` rejects (quota exceeded, blocked IndexedDB upgrade, a private-mode origin), `setSaving(false)` runs, the button re-enables looking exactly as it did before the click, and the rejection escapes as an unhandled promise rejection because the call site is `void save()`. The user is told nothing and reasonably assumes the job was saved.

Catch it and swap the helper line to an error message — the repo has no toast primitive, so confirming in place (the pattern the `saved` branch already uses) is the house move.

src/lib/job-tracker.ts
```diff
@@ -0,0 +1,172 @@
...
+  url?: string;
+  notes?: string;
+  status?: JobStatus;
+  resumeId?: string;
+  jdText?: string;
+  matchResult?: unknown;
+}
+
+/** The editable subset of a job — everything a user can change from the tracker.
+ *  `undefined` leaves a field untouched; pass an empty string to clear an
```
s-annam: **Secondary — the doc says the opposite of what the code does.**

"`undefined` leaves a field untouched" is false for an explicitly-passed key: `updateJob` does `{ ...existing, ...patch }`, and a spread copies an own property whose value is `undefined`, clearing the field. `unlinkResume` at line 93 depends on precisely that behaviour — `updateJob(id, { resumeId: undefined })` is how the link gets dropped. So as written, the docblock forbids the mechanism its own sibling relies on.

Suggest: "an *omitted* key leaves the field untouched; an explicit `undefined` clears it (this is how `unlinkResume` works); pass an empty string to blank an optional text field."

src/hooks/useResumeLibrary.ts
```diff
@@ -102,6 +103,10 @@ export function useResumeLibrary(): ResumeLibrary {
   const remove = useCallback(
     async (id: string) => {
       await removeLibraryResume(id);
+      // Graceful degrade (#323 AC): a tracked job that pointed at this resume
+      // keeps its record and loses only the dangling link. Runs after the
+      // delete so a failure here can never leave the resume undeleted.
+      await clearResumeLink(id);
```
s-annam: Nit — the ordering rationale in the comment above holds (a throw here can't leave the resume undeleted), but the throw *does* skip `refresh()` on the next line, so the deleted resume stays on screen until something else re-lists. A `try`/`catch` around this call, or moving `refresh()` ahead of it, keeps the UI honest in the failure case. Non-blocking.

s-annam (changes requested):
Solid, well-documented slice — the domain/hook/UI split mirrors the resume-library lane exactly, the graceful-degrade path is real and mutation-tested, and every surface composes `@design-system` with semantic tokens. Two things block: a state bug that makes the JD-fit seam single-use per session, and a token that no longer exists after #526.

**Verdict rule applied:** ≥1 Blocking → `REQUEST_CHANGES`. 0 Blocking would have been an approve regardless of the nits.

Six findings are anchored inline. The rest, plus the gate results, are below.

## Gates

- `npm run typecheck` ✅ · `npx eslint src` ✅ · the PR's 25 tests ✅ (`job-tracker`, `JobTracker`, `SaveJobFromMatch`, `useResumeLibrary`)
- **fallow** — 2 complexity findings (`App` CRAP 870, `JdFitApp` CRAP 132), both inherited and excluded by the audit gate. Not this PR's; no action.
- **Fixture PII — skipped, not passed.** No fixture binary is added or changed in this diff, so `check:fixtures` and the `pdftotext` read had nothing to look at.
- **Design-system / tier gate ✅** — no raw `<button>`, no hex, no raw palette class, no manual `dark:` variant. The link picker reusing `JobStatusPicker`'s button-list shape instead of minting a dropdown primitive is the right call, and the PR body justifies it.

## Nits (non-blocking, no re-review needed)

- `clearResumeLink` / `reconcileResumeLinks` write through `putRecord`, which stamps `updatedAt: now`. Deleting one resume therefore reshuffles the tracker's "most-recently-updated first" order for every job that merely referenced it — the user sees unrelated rows jump to the top.
- `JobTracker.byStatus` re-scans the whole list once per status (6× per render). One `reduce` into a `Record<JobStatus, JobRecord[]>` does it in a single pass.
- `App.tsx` — `resumeName` is rebuilt every render. Harmless while `JobTracker` isn't memoized; worth a `useCallback` if it ever is.
- The library card and the tracker card each render the same origin-wide `estimateStorageUsage()` figure, and each hook re-runs it on every mutation. Two "N used" readouts showing one number is a little confusing next to each other.
- fallow function size: `JobTrackerEntry` 128 LOC, `useJobTracker` 116 LOC. Report-only in `verify` so nothing is failing — flagging only because `JobTrackerEntry.tsx` is on the path to the ~200 LOC file guidance.
- The `github-advanced-security` comment on `useJobTracker.ts` ("export never imported") is stale — `App.tsx` imports it now. Ignorable.

---

Samhit21:
Addressed the review in `8ea22a4` (amended into the single commit per the one-commit-per-PR rule — the branch tip moved from `ecf74f4`).

**Both blocking:**

- **Single-use JD-fit seam** — `saved` is now `savedFor: string | null`, compared against the current `jdText`, so a pasted-over posting resets the affordance with no effect and no unmount dependency. I reproduced your exact repro as a failing test first (render JD A → save → re-render JD B → no Save button), then fixed it; mutation-tested by reverting to `savedFor !== null`, which fails it.
- **`brand-amber`** — confirmed dangling: zero definitions in `tokens.css`/`theme.css` and zero other consumers in `src/`. Now `text-accent-primary`.

**Secondary:**

- **Flag claim vs. code** — fixed in the direction you preferred: `JobTrackerSection` and `SaveJobFromMatchSection` own `useJobTracker`, and are rendered only under the flag. A flag-off visit to either entry now opens no IndexedDB at all, so the docblock is true as written. Both contradicting comments are gone.
- **Duplicated backup download** — extracted to `downloadStorageBackup()` in `lib/storage/backup.ts`; both hooks call it. The filename is a module-private const there, so it can only change in one place.
- **Silent failed save** — `catch` added; the helper line swaps to an error message and the button stays available for retry. Covered by a rejecting-tracker test.
- **`JobPatch` docblock** — rewritten to say omitted-vs-explicit-`undefined`, and to name `unlinkResume` as the mechanism that depends on it.

**Nits, also taken:**

- `updatedAt` reshuffle — this one read as more than cosmetic, so `putRecord` gained a `touch?: boolean` opt-out and the two link-cleanup writes use `touch: false`. Deleting a resume no longer floats unrelated jobs to the top. Mutation-tested.
- `byStatus` → single `reduce` pass, memoized; `resumeName` → `useCallback`; `clearResumeLink` wrapped so a throw can't skip `refresh()`.

**Not taken:** the two "N used" storage readouts — that's a product call about whether both cards should show an origin-wide figure, and I didn't want to settle it inside a review fix. Happy to file it if you want it tracked.

`npm run verify` green — 2565 tests pass. Fallow dead-exports is back to the inherited baseline (I'd briefly added one and removed it).

Ready for another look.

src/lib/job-tracker.ts
```diff
@@ -0,0 +1,183 @@
...
+  return linked.length;
+}
+
+/**
+ * Sweep dangling resume links against the set of resume ids that still exist —
+ * a belt-and-suspenders reconcile for links orphaned by any delete path the
+ * explicit {@link clearResumeLink} call missed (e.g. an import that dropped a
+ * resume). Returns the number of jobs repaired.
+ */
+export async function reconcileResumeLinks(
```
s-annam, line 136: Nit (non-blocking): this export has no production caller. The docblock says "Call this from the resume-delete path," but `useResumeLibrary.remove` calls `clearResumeLink`, not this. So the belt-and-suspenders reconcile for a delete path `clearResumeLink` misses (the import-drops-a-resume case it names) never actually runs. Fallow only passes it because `job-tracker.test.ts` references it. Either wire it into the import path or add a `// staged for #<follow-up>` note so it doesn't read as live code.

src/components/features/JobStatusPicker.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * to that status on click. Also exports the shared status → label / badge-tone
+ * maps so the row badge and the picker never disagree.
+ */
+
+import { Button } from "@design-system";
+import { JOB_STATUS_ORDER } from "../../lib/storage/types.ts";
+import type { JobStatus } from "../../lib/storage/types.ts";
+
+/** Mirrors the `StatusBadge` tone union (not exported from the barrel). */
+type StatusBadgeTone = "ok" | "limited" | "warning" | "info";
```
s-annam: Nit (non-blocking): this union is hand-copied from the barrel's `StatusBadge` tone type. If the design-system tone union changes, this drifts silently and `JOB_STATUS_TONE` could reference a tone that no longer exists. Prefer exporting the tone type from `@design-system` and importing it here.

s-annam (dismissed):
Approve. Solid, well-documented feature — clean domain/hook/component layering, reuses the design-system primitives, semantic tokens throughout, and the storage/persistence plumbing is shared with the resume library rather than re-rolled. Tests pass (37), typecheck + lint clean, no fixtures touched, no hardcoded colors, every component under 200 LOC. No Blocking findings; the notes below are all Nits.

**Verdict rule:** 0 Blocking → APPROVE.

## Nits (non-blocking)

1. **`reconcileResumeLinks` has no production caller** — `src/lib/job-tracker.ts:131` (inline). Documented as "Call this from the resume-delete path," but that path calls `clearResumeLink`; a merge-mode import that drops a resume would orphan links with nothing to sweep them. Fallow passes only because a test references it. Either wire it into the import path or drop a `// staged for #<follow-up>` note so it doesn't read as live.

2. **`StatusBadgeTone` union is hand-copied** — `src/components/features/JobStatusPicker.tsx:17` (inline). It mirrors the barrel's `StatusBadge` tone union by hand; if that union changes, this drifts silently. Prefer exporting the tone type from `@design-system` and importing it.

3. **`byStatus` silently drops an out-of-lifecycle status** — `src/components/features/JobTracker.tsx:57`. `buckets[job.status]?.push(job)` swallows a job whose `status` isn't in `JOB_STATUS_ORDER` (e.g. a corrupt or future-version imported record), so the header count can exceed the visible rows. Edge case only; a fallback bucket or a dev assert would make it loud instead of silent.

Reviewed SHA: 8ea22a47208b755d5ba35367f3f7db2bcafdb85f

---

Samhit21:
Addressed the three approval nits in `8654247`, and rebased onto current `main` (the approval was auto-dismissed by the force-push).

**Nits:**

- **`reconcileResumeLinks` reads as live but has no caller** — kept it (the belt-and-suspenders reconcile is still the right home for the import-drops-a-resume case), and filed **#547** to wire it into the JSON import path once that UI lands. The docblock now carries a `staged for #547` note and says plainly why there's no production caller yet: the delete path uses `clearResumeLink`, and `importAll`/`importFromJson` aren't wired to any UI.
- **`StatusBadgeTone` hand-copied** — exported the tone type from `@design-system` (`StatusBadge.tsx`) and `JobStatusPicker` now imports it, so the union can't drift. Landed cleanly alongside your `aria-hidden` addition to the same file.
- **`byStatus` silently drops an out-of-lifecycle status** — the grouping now buckets by each job's actual status and renders the canonical lifecycle first, then any leftover status (a corrupt or future-version import) under its raw label, so a job can never count toward the header total while appearing in no section. `jobStatusLabel` / new `jobStatusTone` fall back gracefully so such a row renders a real badge instead of a blank one. New test asserts an unknown-status job stays visible and counted; mutation-tested by reverting to canonical-only grouping, which fails it.

**Rebase notes** (three files main had also moved):
- `storage/types.ts` — kept both additions: main's `BoardCacheRecord` + `boards` store (#533) and this PR's `JOB_STATUS_ORDER`. This PR's pinned `JobRecord` supersedes the old open `[field: string]: unknown` shape — pinning the fields is exactly what #323 does, and `db.ts`'s boards store is unaffected.
- `App.tsx` — main's hero-copy rewrite (#537) and this PR's `JobTrackerSection` wiring are in different regions; auto-merged, both intact.
- `storage/index.ts` — main's `DB_VERSION` export drop and this PR's `downloadStorageBackup` export are independent; auto-merged.

`npm run verify` green — 2780 tests pass, fallow dead-exports at the inherited baseline. Ready for another look.

src/lib/job-tracker.ts
s-annam: Addressed in `8654247`: added a `NOTE: staged for #547 — no production caller yet` block to the docblock, naming the import path (`storage/backup.ts`) that will call it once wired to UI, so it no longer reads as live code.

src/components/features/JobStatusPicker.tsx
s-annam: Fixed in `8654247`: `StatusBadgeTone` is now exported from `StatusBadge.tsx` (and flows through the `@design-system` barrel's `export *`); `JobStatusPicker.tsx` imports `type StatusBadgeTone` from `@design-system` instead of hand-copying the union, so it can no longer drift.

src/components/features/JobStatusPicker.tsx
```diff
@@ -0,0 +1,75 @@
...
+  applied: "Applied",
+  interviewing: "Interviewing",
+  offer: "Offer",
+  rejected: "Rejected",
+  archived: "Archived",
+};
+
+/** Badge tone per status — shared with the row's `StatusBadge` so display and
+ *  picker stay consistent. */
+export const JOB_STATUS_TONE: Record<JobStatus, StatusBadgeTone> = {
```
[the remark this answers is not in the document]
s-annam: Fixed in `81589b0`: dropped `export` from `JOB_STATUS_TONE` — it's module-private now, read only through the exported `jobStatusTone()` (which adds the unknown-status fallback). `fallow audit --base origin/main` no longer flags it; dead-exports back at the inherited baseline.

Samhit21:
Follow-up push `81589b0`: the standalone `fallow` check caught one new dead export I'd created — routing `JobTrackerEntry` onto the new `jobStatusTone()` helper left `JOB_STATUS_TONE` with no external consumer. Made it module-private (it's only used inside `jobStatusTone` now). All three checks green — `verify`, `fallow`, Cloudflare Pages.

s-annam (approved):
Approving. All review threads addressed and re-verified against the current tip (`81589b0`).

**Nits, confirmed fixed:**

- **`reconcileResumeLinks` read as live with no caller** — kept (right home for the import-drops-a-resume case), docblock now carries a `staged for #547` note naming why there's no production caller yet. Reads as staged code, not dead code. ✓
- **`StatusBadgeTone` hand-copied** — now exported from `StatusBadge.tsx`, flows through the `@design-system` barrel's `export *`, and `JobStatusPicker` imports the type instead of re-declaring it, so the union can't drift. ✓
- **`JOB_STATUS_TONE` flagged by fallow as a new dead export** — dropped `export`; it's module-private now, read only through the exported `jobStatusTone()` (which adds the unknown-status fallback). `fallow audit --base origin/main` no longer flags it; dead-exports back at the inherited baseline. ✓

**Rebase onto `main` verified** — `storage/types.ts` keeps both additions (main's `BoardCacheRecord` + `boards` store from #533, and this PR's `JOB_STATUS_ORDER`); `App.tsx` and `storage/index.ts` auto-merged in independent regions, both intact.

**Gates on the tip:** `tsc -b --noEmit` clean · job-tracker suites green · fallow at inherited baseline.

**Disclosure:** my only writes to this PR were replying to and resolving the review threads — no code of mine is in the branch; every fix above is authored by @Samhit21.

PR #472, opened 2026-07-14
fix(parser): stop bulleted and tab-aligned skills rows shredding multi-word skills (#465)

### LINKED ISSUE #465: [parser] skills — bulleted single-column labelled skills shred multi-word tokens and ingest category labels
## Problem

A bulleted single-column Skills section with bold `Label:` category rows shreds every multi-word skill on entry, ingests the category labels as skill values, and explodes parenthesised clarifiers. On a well-formatted single-column résumé — the Friends & Family gate scenario — the parser turns a legitimate ~30-skill list into ~45 chips including obvious garbage tokens.

### Repro

Synthetic skills section (mirrors the shape a friend's résumé produces):

```
SKILLS
• Frontend: React, TypeScript, Next.js, Vite, Tailwind CSS, React Query, HTML5,
  CSS3, JavaScript
• Frontend Testing: React Testing Library, Playwright
• Backend: Java, Python, Spring Boot, REST APIs, Apache Kafka
• Cloud & Infra: AWS (EC2, S3, RDS), Docker, Kubernetes, Terraform
• Databases & Caching: PostgreSQL, MySQL, Redis, MongoDB
• Product & Collaboration: A/B testing, event-driven architecture, CI/CD, Agile
```

Expected (~24 tokens): `React, TypeScript, Next.js, Vite, Tailwind CSS, React Query, HTML5, CSS3, JavaScript, React Testing Library, Playwright, Java, Python, Spring Boot, REST APIs, Apache Kafka, AWS, Docker, Kubernetes, Terraform, PostgreSQL, MySQL, Redis, MongoDB, A/B testing, event-driven architecture, CI/CD, Agile`.

Actual (~45 tokens, extraction from `probe-skills`):

- **Category labels ingested as skills:** `Frontend`, `Testing`, `Cloud`, `Databases`, `Product` (from the bold `Frontend:` / `Frontend Testing:` / `Cloud & Infra:` / `Databases & Caching:` / `Product & Collaboration:` row headers).
- **Multi-word skills split on internal whitespace:** `Tailwind` + `CSS` (from `Tailwind CSS`), `Query` alone (from `React Query`), `Spring` + `Boot`, `REST` + `APIs`, `Apache` + `Kafka`, `Library` (from `React Testing Library`), `testing` + `event-driven` + `architecture` (from `A/B testing, event-driven architecture`).
- **Parenthesised clarifiers exploded:** `AWS`, `(EC2`, `S3`, `RDS)` from `AWS (EC2, S3, RDS)` — the open-paren survives as its own "skill".

### Where in the loop

Parse layer. `tokenizeSkillLine`'s unit tests (`extract/skills.test.ts`) pass cleanly on the exact strings above — the corruption enters upstream, in `collectSkillCells`. When a bulleted row like `• Frontend: React, TypeScript, …` reaches `splitColumnCells` (`extract/skills.ts:140`), the bold label + inter-token spacer items in pdfjs create *multiple* column cells instead of one: `["Frontend", "React", "TypeScript", …]` or similar. Each cell then flows through `tokenizeCell` individually, so:

- `Frontend` (the label) is tokenized as its own skill (label-strip in `tokenizeCell` requires the trailing `:`; without it, the label is treated as a skill token).
- `Tailwind CSS`, arriving as one cell of two words separated by a single space, is split by `SKILL_SPLIT_RE = /[;·•|]+|\s{2,}|(?<!\w)\/+(?!\w)/` — the `\s{2,}` alternative fires because the pdfjs items for `Tailwind` and `CSS` land far enough apart on the x-axis that `mergeItemText` inserts multiple spaces between them.

The Vaishnavi_Resume_2026 file (dogfooded, uncommitted) reproduces this exactly — 46 tokens, same garbage classes. Corpus fixtures don't currently cover the bulleted-single-column-with-bold-labels shape, which is why this regressed silently.

## Proposed change

Two-part fix in `src/lib/heuristics/extract/skills.ts`:

1. **Prefer the assembled line text over item-level column splitting when the line has the bicellular `Label: comma-list` shape.** In `splitColumnCells` (`skills.ts:140`), when the resulting `cells.length > 2` but the joined `line.text` matches `SUBLABEL_PREFIX_RE` (`skills.ts:182`), the multi-cell split is a false positive — a bulleted labelled row, not a genuine multi-column layout. Return `[line.text]` in that case so the entire row flows through `tokenizeCell` as one cell.

2. **Normalise multi-space runs inside a cell before delimiter splitting.** In `tokenizeCell` (`skills.ts:190`), collapse `\s+` to a single space on the cleaned string BEFORE invoking `splitRespectingParens`. This makes `SKILL_SPLIT_RE`'s `\s{2,}` alternative non-firing inside a cell whose only source of wide whitespace was pdfjs item-position noise. A genuine wide-gap column boundary is already handled by `splitColumnCells`; by the time we're inside `tokenizeCell`, wide gaps are position noise, not semantic separators.

Both changes are load-bearing:

- (1) alone rescues label ingestion but leaves multi-word skills split.
- (2) alone rescues multi-word skills but still ingests labels because `Frontend` is emitted as its own cell before it reaches the tokenizer.

## Implementation plan

1. **Add a corpus fixture** at `tests/fixtures/pdfs/unknown/bulleted-labelled-single-column-skills.pdf` — synthetic-persona (fake name, `@example.com`, `(312) 555-0123` — real area code + `555` exchange + `0100`–`0199` subscriber per the fixture PII policy), single column, Skills section shaped exactly like the repro above. Ship it via a template re-export (do NOT download the OSS template author's own résumé; those carry real PII).
2. **Add unit tests** in `extract/skills.test.ts`:
   - `collectSkillCells` on a labelled row with pdfjs-item spacers returns one cell, not three-plus.
   - `tokenizeCell` on a cell containing multiple internal-space runs returns single tokens like `Tailwind CSS`, `Spring Boot`, `React Testing Library`, `AWS (EC2, S3, RDS)`.
   - Category labels (`Frontend`, `Cloud`, `Databases`, `Product`) don't survive as skills when the row starts with a `Label:` prefix.
3. **Apply the two changes above** in `extract/skills.ts`. Guard (1) so multi-column layouts without a `Label:` prefix are unaffected (existing behaviour preserved for the `chanchal-sharma-bulleted-skills` and openresume-laverne fixtures).
4. **Run `corpus.test.ts` and `corpus-roundtrip.test.ts`** — confirm no regression (`KNOWN_FAILURES` unchanged; if a currently-baselined skills entry starts passing, delete it per the ratchet).
5. **Re-run `probe-skills` on the new fixture** — verdict `ok` and token count matches the expected ~24.

## Reuse analysis

**Capability:** tokenising a Skills-section line into skill entries.

**Existing surfaces found:** `tokenizeSkillLine` / `tokenizeCell` / `collectSkillCells` in `src/lib/heuristics/extract/skills.ts`.

**Decision: extend by patching the existing helpers.** No new abstraction. The bug is local to the interplay between `splitColumnCells` and `SKILL_SPLIT_RE`'s `\s{2,}` guard — a targeted change in both, not a rewrite.

## Acceptance criteria

- [ ] New fixture `tests/fixtures/pdfs/unknown/bulleted-labelled-single-column-skills.pdf` exists, synthetic-persona only, verified with `pdftotext … | head -40` before merge.
- [ ] `probe-skills` on the new fixture reports `ok` with no `EXTRACTION-MISS`.
- [ ] Extraction on the new fixture returns exactly the expected ~24 tokens; none of `Frontend`, `Testing`, `Library`, `Cloud`, `Databases`, `Product`, `(EC2`, `S3`, `RDS)`, `Query`, `Boot`, `APIs`, `Kafka` appear.
- [ ] `Tailwind CSS`, `Spring Boot`, `REST APIs`, `Apache Kafka`, `React Testing Library`, `React Query`, `AWS (EC2, S3, RDS)`, `A/B testing`, `event-driven architecture` survive as single tokens.
- [ ] `corpus.test.ts` and `corpus-roundtrip.test.ts` stay green with no `KNOWN_FAILURES` regression. If either passes for a fixture currently baselined on `skills`, that entry is deleted from the baseline.

## Out of scope

- The routing-layer bugs already tracked by #374 (two-line header wrap) and #414 (leading decorative glyph). This issue is extraction-layer.
- Category-label semantics (grouping skills into `Frontend` / `Backend` / …) as a first-class output — for now, labels are dropped, not preserved. That's a follow-up if we ever surface category grouping in the UI.

## Parent

Child of **#415 — Epic: skills section recognition & extraction robustness**. This is the extraction-layer counterpart to the routing-layer children (#374, #414). Filed at P1 · Friends & Family even though the parent epic sits at P2 — this specific failure class fires on the F&F gate scenario (well-formatted single-column PDF).

s-annam:
Closes #465.

A bulleted single-column Skills section shredded every multi-word skill on entry (`Tailwind CSS` → `Tailwind` + `CSS`) and exploded parenthesised clarifiers (`AWS (EC2, S3, RDS)` → `AWS (EC2` / `S3` / `RDS)`).

## What was actually wrong

The issue's proposed fix would not have worked, and we found that by reproducing before touching code:

- Its part 2 (collapse `\s+` inside `tokenizeCell` so `SKILL_SPLIT_RE`'s `\s{2,}` stops firing) is a **no-op** — `mergeItemText` already ends with `.replace(/\s+/g, " ")`, so no caller can ever see a multi-space run. That alternative is unreachable.
- Its part 1 (if `line.text` matches `SUBLABEL_PREFIX_RE`, return `[line.text]`) **never fires on the shape the bug lives in** — the regex is `^`-anchored and a bulleted row's `line.text` starts with `•`. Implemented and driven against the corpus, it still shreds `Data Visualization`.

The real causes are two independent defects:

1. **`splitColumnCells` read a bullet's hanging indent as a column gutter.** pdfjs emits a blank item as wide as the indent, which clears the spacer floor (`width > max(fontSize, 10)`), so *every* bulleted row looked multi-column and took the branch of `collectSkillCells` that structurally cannot soft-wrap-rejoin. Any skill straddling the wrap was shredded. `dropLeadingBullet` removes the glyph and its indent — **a leading bullet is decoration; its hanging indent is not a gutter.**

2. **`isSoftWrapContinuation` could not reach a wrap landing inside parens.** `pending` ends on a comma (`… AWS (EC2,`), so Condition B reads it as a completed entry and declines to join, leaving `splitRespectingParens`' unbalanced-paren fallback to shred the clarifier. Condition C joins when the break is mid-list **inside an open clarifier** *and* the join **closes** it — making it the exact complement of Condition B rather than a superset. Both conjuncts are load-bearing: joining on the open paren alone never re-satisfies its own predicate, so it self-perpetuates and swallows the rest of the section (caught in review — see below).

The same tab stop that tears a bold `Label:` off its body emits a blank wider than the spacer floor, so a Word/Docs export hits defect 1 **even without a bullet**. `rejoinBareLabels` re-attaches a cell that is nothing but `Label:` to the cell after it, keying on cell **content**, not gap width — **a cell that is nothing but a label is not a column.** A genuine column always carries its items. This is strictly narrower than collapsing the row, which would mash a tab-aligned multi-column labelled grid (`Frontend: …⇥ Backend: …`) into one cell.

## Blast radius

Verified at **value level** (not just snapshot counts) across all 46 corpus fixtures: exactly two skills arrays change, both improvements.

- `google-docs-skia-proxy-programs-skills-software` — `skillsCount` 16 → 14, because `"and citation (APA"` / `"MLA"` / `"Chicago)"` correctly collapses to `"and citation (APA, MLA, Chicago)"`. The other 13 tokens are byte-identical. **A sibling instance of the same defect, already shipped in `main`** — good corroboration of the root cause.
- The new fixture (41 → 37 tokens).

Every other fixture is identical, including the multi-column ones the change could plausibly have broken (`chanchal-sharma-bulleted-skills`, `openresume-laverne-word-quartz`, and the two-column fixtures).

The `line-primitives.ts` bullet-glyph hoist has the widest blast radius in the diff (`isBulletLine`/`stripBullet` are used well beyond skills), so it was proven a no-op: the compiled regex `.source` is byte-identical to the literals it replaces, fuzzed over U+0000–U+2FFF with zero mismatches.

## Fixture

New `tests/fixtures/pdfs/unknown/bulleted-labelled-single-column-skills.pdf` — synthetic persona (`Jane Smith` / `jane.smith@example.com` / `(312) 555-0123`), `pdftotext`-verified. It covers **both** the narrow-gap and the tab-aligned-label shapes and soft-wraps mid-skill, so it fails without the fix. The first cut of the fixture set the label→body gap under the spacer floor and therefore never exercised the branch the bug lives in — review caught that.

Tests also pin two **accepted limits**, so a future change can't silently move them: a clarifier wrapping across 3+ lines is still shredded (unchanged from `main` — deferred to #475), and a line that closes one clarifier while opening another now recovers the second (better than `main`).

## The category-label half

The issue also reports **category labels being ingested as skills** (`Frontend`, `Cloud`, `Databases` surviving as chips). This branch handles them on two paths — `tokenizeCell` strips a leading `Label:` prefix, and `rejoinBareLabels` drops a cell that is *nothing but* a label — but no synthetic geometry would reproduce the original failure, so the fix was verified without ever watching the bug happen.

The reporter has since run the **actual résumé #465 was filed from** against the merged fix: the category labels no longer appear as chips. That is the real failing input rather than a reconstruction of it, so the AC is confirmed and this PR `Closes` the issue.

One loose thread, recorded rather than buried: #465 reported the leaked labels as *fragments* (`Cloud`, not `Cloud & Infra`; `Testing`, not `Frontend Testing`), and dropping a label whole does not by itself explain a fragment. The likeliest reading is that the fragments were a symptom of the **shredding** defect — which *is* reproduced and fixed here — and were never a distinct bug. Observed behaviour on the real résumé is consistent with that. (Tracked and closed as #474.)

## Verification

`npm run verify` green — 2135 passed / 8 skipped, typecheck + lint + build clean. Remaining `fallow` findings are the fixture-generator false positives (three tracked `gen-*.mjs` siblings share the same pdf-lib boilerplate by design).

## Adversarial review

Two rounds, pre-PR, on the local diff.

**Round 1 — 2 blocking:**
1. **Condition C was unbounded** — once a line left a paren open, it stayed true forever, joining the rest of the section into one cell and starving the unbalanced-paren fallback. Repro: `Cloud (AWS` / `Docker` / `Kubernetes` / `Terraform` → `["Cloud (AWS Docker Kubernetes Terraform"]`, **losing 3 skills** vs `main`. Fixed by requiring the join to *resolve* the imbalance.
2. **The fix was narrower than the bug** — the fixture's label→body gap sat under the spacer floor, so the tab-aligned case was untested and unfixed. Widen the gap to a real tab stop and #465 reproduced verbatim on the branch. Fixed by `rejoinBareLabels` + a tab-aligned row in the fixture.

**Round 2 — clean.** Verified the bounded Condition C against nested parens, a next line that closes one clarifier and opens another, and a 3-line wrap; confirmed multi-column fixtures unmoved at value level.

**Nits applied** (each was a case where the branch was *worse than `main`*, so they were worth taking):
- Two consecutive bare-label cells leaked `Backend:` as a skill token.
- Condition C would marry a stray `(` to an unrelated stray `)` on a later line (`Vim (advanced` / `Emacs, VS Code)` → one garbage token, losing 2 skills). Now also requires `pending` to end on a comma.
- `hasUnclosedParen` duplicated the depth walk in `splitRespectingParens`; the two "must agree" — now structurally one `scanParens` helper.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #465 | Claude Opus 4.8 | ultra |
| Adversarial review — round 1 | Claude Opus 4.8 | medium |
| Review fixes | Claude Opus 4.8 | medium |
| Adversarial review — round 2 | Claude Opus 4.8 | medium |
| Nit fixes | Claude Opus 4.8 | medium |
| Orchestration + PR | Claude Opus 4.8 (1M context) | — |
| Verification | `npm run verify` — green | — |


s-annam (commented):
Rule applied: 0 Blocking findings → APPROVE.

## Nits

- `scripts/fixtures/gen-bulleted-labelled-skills.mjs`: `fallow audit` flags this new script as an "Unused file" (since it's a standalone script and not imported) and notes some duplicated boilerplate with other fixture generators. Harmless for a fixture script, just noting it for visibility.

---
Reviewed by: Gemini 3.1 Pro (High)

Vaishnavi1709 (dismissed):
Reviewed under the resumelint pr-review gates. 0 blocking → approve.

## Gates

| Gate | Result |
|---|---|
| **3a — fixture PII** | **PASS.** `bulleted-labelled-single-column-skills.pdf`: Jane Smith / `jane.smith@example.com` / **(312) 555-0123** / Austin, TX. Metadata Creator/Producer = `pdf-lib` (no author leak). Phone matches CLAUDE.md exactly. |
| **3b — design-system** | N/A — no `src/components/**` changes. |
| **3c — style tokens** | N/A — grep hits were `#465` refs, no hex/palette. |
| **3d — fallow** | Skipped locally; author documents suppressions on `gen-*.mjs` sibling boilerplate. |
| **3e — scripts as code** | **PASS.** `gen-bulleted-labelled-skills.mjs` is a self-contained ESM generator, idempotent write, synthetic persona hardcoded, no shell/`gh` surface. |
| **`line-primitives.ts` hoist** | **PASS.** Verified byte-identical: `BULLET_LEAD_RE` = `/^\s*[•‣▪●◦⁃*\-–—]/` and `BULLET_PREFIX_RE` = `/^\s*[•‣▪●◦⁃*\-–—]\s*/` reproduce the pre-hoist literals exactly. New `isBulletGlyph` has one consumer (`skills.ts:167`) — not fallow-fodder. |
| **`dropLeadingBullet`** | **PASS.** Fires only when `items[0].str.trim()` is exactly a bullet glyph, then skips subsequent blank items. Doesn't fire on `"•content"` glued into a single item, but that shape doesn't produce the wide hanging-indent spacer that triggers the false multi-column split either — so the accepted hole is inert. |
| **`rejoinBareLabels`** | **PASS.** Traced against `["Frontend:", "React..."]` → correct absorb; `["Frontend:", "Backend:", "Java"]` → guard on `!BARE_SUBLABEL_RE.test(t)` keeps "Backend:" from being swallowed by "Frontend:", and the trailing "Frontend:" is dropped by the final `filter`. Matches the "Backend leaked as a skill" nit called out in the PR body. |
| **Condition C bounded** | **PASS.** All three conjuncts load-bearing: (1) `hasUnclosedParen(pending)` — break inside a clarifier; (2) `/,\s*$/.test(pending)` — true complement of Condition B, rejects the "Vim (advanced" / "Emacs, VS Code)" stray-paren case; (3) `!hasUnclosedParen(joined)` — join must CLOSE the imbalance, so a stray "(" cannot self-perpetuate. Pinned by the `Cloud (AWS` regression test at `skills.test.ts:575`. |
| **Golden delta** | **PASS.** `skillsCount 16 → 14` on `google-docs-skia-proxy-programs-skills-software` is a genuine improvement: `"and citation (APA"` / `"MLA"` / `"Chicago)"` (three shredded tokens) collapses to `"and citation (APA, MLA, Chicago)"` (one). `-3 + 1 = -2`. Matches the blast-radius claim. |
| **Test coverage** | **PASS.** Happy path (bullet+wrap), paren-clarifier wrap (Cond C fires), stray open paren (bounded C), tab-aligned label rejoin, multi-column preserved, bare-label drop, Vim/Emacs stray-paren, multi-column-with-bullets. Two `KNOWN LIMIT` / `KNOWN BEHAVIOR` tests explicitly pin the edge of the fix's scope. |

## Nits (non-blocking)

**N1. Fixture generator embeds `pdf-lib` timestamps.** `pdfinfo` on the shipped fixture reads `Mon Jul 13 20:32:05 2026 PDT`. Every regeneration mutates the binary even when visual output is unchanged, which shows up as spurious diff churn if a future contributor tweaks the generator. Consider `doc.setCreationDate(new Date(0))` / `doc.setModificationDate(new Date(0))` for reproducible artifacts. Sibling `gen-*.mjs` files likely have the same shape; a follow-up could normalize all four.

**N2. `splitColumnCells` fallback re-emits the bullet.** When `bodied.length === 1 && !dropped`, the return is `[line.text]` — which still contains the leading `•`. Harmless because `tokenizeCell` calls `stripBullet` unconditionally, but the two paths (items-derived cell text vs raw `line.text`) diverge on bullet presence. A future refactor that touches either side should preserve the `stripBullet`-downstream invariant.

The two adversarial-round-1 blockers the author caught pre-PR (Condition C unbounded; original fixture never exercised the tab-aligned branch) are the load-bearing ones — both are landed correctly. Test coverage explicitly pins both the FP surface and the deferred edges (3-line wrap, close-one-open-another). Solid fix.

PR #477, opened 2026-07-14
docs: split CLAUDE.md code guidance from shipping process, and re-home the rules the split would have dropped

s-annam:
## Summary

`CLAUDE.md` had grown to 230 lines, and roughly 60% of it was **per-PR governance** — fixture PII policy, AI provenance rules, squash-message mechanics, deploy, license. All of it correct and load-bearing, but it only applies when you are *opening a PR*, not when you are *writing a function*. Because `CLAUDE.md` loads into every session as an always-on instruction block, it read as a per-edit checklist and crowded out the guidance that actually shapes a diff.

This splits it — but the split is only safe if it accounts for **how each surface actually loads**. That accounting is most of this PR, and it grew the diff past the docs-only change the branch started as.

### 1. `CLAUDE.md` (230 → 150 lines) — code guidance, plus three rules that must stay resident

`CLAUDE.md` is the only file guaranteed to be in *every* agent's context. Moving a rule out of it converts "the agent cannot not know this" into "the agent will know this if something makes it open the file." That trade is fine for reference material and **bad for prohibitions whose violation is silent or irreversible**.

So three rules stay, under a new **`## Hard rules (no exceptions)`** stanza:

- **Fixture PII** — including the two non-obvious halves: an area-code-`555` number is invalid NANP and silently drops the field, and an OSS template's shipped demo PDF is *not* an exception.
- **Never in git** — no `Co-Authored-By:`, no `Claude-Session:`, no `🤖` badge. This one is load-bearing precisely because the Bash tool's *own default commit template suggests them*; the resident ban is the counterweight. A commit written mid-session, before anyone reaches the PR step, has nothing else contradicting that template.
- **One commit per PR.**

The rationale for each moves to the process doc. The prohibition does not.

Also new in `CLAUDE.md`: an **exemplars table** (the closest file to read before writing a feature component, lib module, hook, test, or design-system piece), the ~200 LOC rule now **naming the four files that violate it** (`ReconstructedResume.tsx` 1184, `ReconstructedRole.tsx` 588, `ModelSelector.tsx` 556, `SectionRewrite.tsx` 501) so the rule isn't silently contradicted by its neighbours, and **narrow-gate guidance** (`vitest` on touched paths + `typecheck` while iterating; save full `verify` for the end).

### 2. The skills carry their own copy — 14 dangling pointers fixed

`/open-pr`, `/pr-review`, `/revise-pr`, `/implement-batch` and `/create-gh-issue` collectively said **"see `CLAUDE.md` → \<section\>" 14 times**, for sections this split removes or renames. A skill body is injected verbatim when the skill runs, so each rule now lives **inline in the skill** — self-contained beats well-linked.

This matters most for **`/implement-batch`**, whose subagents are the ones that actually add fixtures. Its per-issue brief previously said *"synthetic (fake name, `@example.com`, `555`-exchange phone — see `CLAUDE.md`)"* — a gloss that omitted both halves that have actually bitten this repo. It now carries the full rule, including the `pdftotext` verify-the-binary step.

### 3. Stale references in the hooks

`scripts/hooks/check_conventions.py` pointed twice at a `CLAUDE.md` "License" section this PR removes; `style_guard.sh` pointed at "Styling & Tokens" (now lowercase `tokens`). `tests/fixtures/pdfs/CLAUDE.md` — the directory-scoped file that keeps the PII policy loaded whenever an agent touches that tree — had a trailing `Full policy: ../../../CLAUDE.md` that no longer resolved.

### 4. `docs/CONTRIBUTING-PROCESS.md` — the rationale, and honest about it

The new doc opens by stating that it *is* the rationale, that the binding rules live in `CLAUDE.md` → Hard rules / the fixture dir / the skills, and that changing a rule means changing it **there too, not only here**. That's the guard against these drifting apart again. `CONTRIBUTING.md` now cross-links it, so a human contributor can reach the PII policy from the file they actually open.

## Follow-up filed

**#478 — Enforce the fixture-PII policy in CI, not just in prose.** Everything above is still *instructions to a model*. Auditing all 45 fixture PDFs while writing this turned up a live violation: `tests/fixtures/pdfs/unknown/chromium-asymmetric-sidebar.pdf` carries `(555) 018-2390` — an area-code-`555` number, the exact invalid-NANP trap the policy exists to prevent. `libphonenumber-js` rejects it, so that fixture's `phone` is **silently dropping out of its parse and Completeness score today**, and nothing caught it. The rule was written down in three places and a violation walked in anyway. #478 proposes the check that would have stopped it.

(Metadata across the corpus is clean — no `Author` field carries a real name, confirming the #384 purge held.)

## Test plan

- [x] Not docs-only: `.claude/skills/*` (5), `scripts/hooks/check_conventions.py`, `scripts/hooks/style_guard.sh`, `CLAUDE.md`, `CONTRIBUTING.md`, `docs/CONTRIBUTING-PROCESS.md`, `tests/fixtures/pdfs/CLAUDE.md`
- [x] Every remaining `CLAUDE.md → <heading>` reference in the repo re-grepped; all 17 resolve to a live heading
- [x] Hooks are executable code — `python3 -m py_compile` and `bash -n` both pass on the two touched files
- [x] `npm run typecheck` clean; `npm run test` — 2120 passed, 8 skipped
- [x] No fixture binaries added or changed — PII preflight N/A
- [x] `verify` green in CI

## Provenance

Code implementation via: Claude Opus 4.8 (medium)
Review + revision (hard-rules stanza, skill inlining, hook fixes): Claude Opus 4.8 (medium)

### DIFF
--- a/.claude/skills/create-gh-issue/SKILL.md
+++ b/.claude/skills/create-gh-issue/SKILL.md
@@ -67,7 +67,7 @@ If `gh` isn't installed or authenticated, stop and tell the user to run
    invariant that must still hold (`corpus.test.ts`, `corpus-roundtrip.test.ts`).
 
 5b. **Reuse analysis — required when the issue adds a UI/workflow surface.**
-   resumelint enforces a Reuse Gate (see `CLAUDE.md` → "Workflow-surface Reuse").
+   resumelint enforces a Reuse Gate (`CLAUDE.md` → "Component architecture & reuse").
    Before finalizing a body that proposes a **new component, panel, dialog, or
    page**, search for an existing surface that already owns that capability:
    - Check the design-system (`src/design-system/`) for an existing primitive or
--- a/.claude/skills/implement-batch/SKILL.md
+++ b/.claude/skills/implement-batch/SKILL.md
@@ -43,9 +43,9 @@ skills.
   `/code-review` skill; a fresh clone that wants to skip review entirely passes
   `--no-review`.
 - **Fixture PII policy is non-negotiable** — if any issue adds/changes a fixture
-  binary (PDF/image/doc), the persona MUST be synthetic (see `CLAUDE.md` → *Test
-  fixtures*). `/open-pr` re-checks this at push time (its Step 3.5), but flag it
-  the moment a subagent reports a new fixture.
+  binary (PDF/image/doc), the persona MUST be synthetic (the full rule ships to
+  each implementing subagent in Phase 3, step 5). `/open-pr` re-checks this at push
+  time (its Step 3.5), but flag it the moment a subagent reports a new fixture.
 
 ## Input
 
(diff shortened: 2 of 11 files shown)


s-annam (commented):
Verdict rule applied: 0 Blocking findings → APPROVE.

The PR cleanly splits the process documentation from code guidance and propagates the updated rules into the skills.

## Secondary

* **`.claude/skills/revise-pr/SKILL.md` (lines ~204-208)**
  In the snippet for updating provenance, `$updated_body` is passed to `gh pr edit`, but it's never defined in the bash block. This is a pre-existing issue, but since this PR touches provenance documentation, it's worth fixing here. An agent following this snippet will likely copy-paste incomplete logic and fail.
  **Fix:** Either provide the actual `awk` or `sed` logic to define `$updated_body`, or explicitly state that the agent must write the text-replacement logic itself.

* **`.claude/skills/open-pr/SKILL.md` (lines ~262-267)**
  The prose comment explicitly states to "update it in place", but the bash snippet (`if ! grep -qF '## Provenance' <<<"$body"; then ...`) only appends the block if it's completely absent. If it's already there, it skips doing anything.
  **Fix:** Align the prose with the code (e.g. "if missing, append it; if present, leave it alone") or provide the actual text-replacement bash.

## Nits (Non-blocking)

* **`scripts/hooks/check_conventions.py` (lines 86-90)**
  The `SPDX_HEADER` string ends with a newline, so the appended text `See CLAUDE.md...` will print on the next line immediately after the code block. For readability, you might want to prepend a newline to the instruction string so there's a blank line separating the code block from the instruction text:
  ```python
              f"`src/{rel_str}` is missing the SPDX header. Prepend:\n"
              f"{SPDX_HEADER}\n"
              f"See CLAUDE.md \"Exemplars\"; license rationale in "
  ```

---
Reviewed by: Gemini 3.1 Pro (High)

s-annam:
I've addressed the review comments!

- **.claude/skills/revise-pr/SKILL.md**: Fixed the undefined `` in the provenance snippet.
- **.claude/skills/open-pr/SKILL.md**: Fixed the bash snippet for updating the provenance block so it aligns with updating the body in place rather than just appending if missing.
- **scripts/hooks/check_conventions.py**: Prepended a newline for readability.

Gates are green (typecheck and tests passed).
Pushed in commit `e3b4751`.

Issue #436, opened 2026-07-11
Parser: round-trip the one-line experience header (title/company disambiguation without the two-line structural signal)


s-annam:
## Problem

The Download-PDF exporter now emits a **one-line experience header** — `Title · Company, Location · Team` with the date flush-right — to match the compact canonical résumé shape (replacing the older stacked two-line `Title` / `Company · Location  Dates` shape from #284/#298).

Our text-only round-trip parser has **no font signal** (`groupIntoLines` drops per-glyph weight), so it used the two-line *structure* to tell the role title from the company. Collapsing both onto one line removes that signal, so some reconstructed résumés re-parse with:

- **title ↔ company swapped** — e.g. `Composer` / `Northwind Ensemble` come back inverted
- **company truncated** — e.g. `Danggeun Pay Inc. (KarrotPay)` → `(KarrotPay)`

This was a deliberate **look-over-fidelity** tradeoff (the compact header is what users expect); the round-trip regressions were baselined rather than blocking the header change.

## Current baseline (to be removed as this is fixed)

- `corpus-roundtrip.test.ts` — `experience` baselined on 14 fixtures (search the file for this issue number).
- 3 repro assertions suspended (`it.skip`, same reference):
  - `render-roundtrip.repro.test.ts` — #284 AC#3 (title/company)
  - `render-roundtrip-year-only.repro.test.ts` — #358 (no swap + year re-attach)
  - `render-roundtrip-team.repro.test.ts` — #425 (team as org sub-line segment; the sub-line no longer exists)

## Fix

Teach the parser to round-trip the one-line `Title · Company, Location  Dates` header:

- `disambiguateCompanyTitle` — when title + company share one header line, use the comma-joined `Company, Location` signature (and the flush-right date anchor position) to pick the company segment; don't invert on a neutral/parenthetical company.
- entry-block anchoring — recognize the one-line header as a single role lead so a parenthetical company (`… (KarrotPay)`) isn't truncated.

**Definition of done:** delete the 14 `experience` baseline lines from `corpus-roundtrip.test.ts` (the ratchet enforces this — a fixed fixture trips the stale-entry check) and un-skip the 3 repro assertions.

s-annam:
**Sub-note from PR #435 review (Samhit):** the one-line-header reparse also breaks **multi-category experience grouping**, not just title/company fields.

When the exporter emits a category heading followed by one-line dated roles:

```
TEACHING EXPERIENCE
[...]

PR #479, opened 2026-07-14
fix(experience): read company + location off the next line for a "Title, Team" comma header (#466)

### LINKED ISSUE #466: [parser] experience — company mirrors the title and location drops when the role header has a "Title, Team" comma suffix and the employer is on the next line
## Why

Child of #460 (F&F trust regression epic). Surfaced by a fresh `probe-experience` dogfooding pass on a real single-column résumé (uncommitted, PII-bearing, per probe policy — all examples below are **synthetic**).

On a well-formatted single-column résumé whose role header carries a **team/sub-org suffix after a comma** — `Title, Team  <dates>` — with the employer on the **next** line, the parser sets `company` to the **title verbatim** and drops `location` entirely. Every affected role loses its employer. This is the F&F base case, not a two-column edge case, and the corrupted `company` propagates into the score's specificity/structure dimensions.

Sibling of #461, but a **different mechanism**: #461 is the em-dash `Title — Multi-Word Company, Location` single-line shape where the location ladder eats the company's trailing token. Here the company is on a *separate line* and never gets read at all — the title is mirrored into the `company` slot instead. Fixing #461 does not fix this.

## Repro (synthetic)

Résumé region (single column, no bullets glyph on the header lines):

```
EXPERIENCE
Software Engineer II, Payments Platform      August 2024 - Present
Northwind Systems | Chicago, IL  Consumer Banking
Built X, shipped Y.
Software Engineer I, Ledger Ingest           July 2022 - August 2024
Northwind Systems | Chicago, IL  Consumer Banking
Did Z.
Software Engineering Intern                  June 2019 - August 2019
Contoso Financial | Chicago, IL  Corporate Banking
Prototyped W.
```

Actual `runCascade(...).parsed.experience`:

|---|---|---|---|
| 1 | `Software Engineer II` | **`Software Engineer II`** ❌ | **`null`** ❌ |
| 2 | `Software Engineer I` | **`Software Engineer I`** ❌ | **`null`** ❌ |
| 3 | `Software Engineering Intern` | `Contoso Financial` ✅ | **`null`** ❌ |

Expected: `company` = `Northwind Systems` / `Contoso Financial`, `location` = `Chicago, IL` on all three.

## The trigger

Role 3 — the only header **without** a `, Team` comma suffix — resolves its company correctly off the following line. Roles 1 and 2, which carry the comma suffix, do not. That isolates the trigger to the **comma-suffix split on the header line**: once the header is split on the comma, the right-hand side is consumed as the team, the company-from-next-line lookup never runs, and the left-hand side (the title) is mirrored into `company` as a fallback. `location` is a casualty of the same short-circuit — the `City, ST` on the employer line is never reached.

## Layer / suspected fix site

- `src/lib/heuristics/extract/experience-disambiguate.ts` — the company/location resolution ladder.
- `src/lib/heuristics/extract/experience.ts` — the header/continuation-line pairing that should read the employer from the next line.
- There is already a test file for the adjacent shape: `src/lib/heuristics/extract/experience.title-team-comma.test.ts`. Start there — the `Title, Team` split exists but does not compose with the next-line-employer case.

A `company` that is byte-identical to its own `title` is never correct. Whatever the ladder decides, that specific outcome is worth an explicit guard: prefer `null` over a mirrored title, so a miss reads as a miss instead of as bad data the user has to notice.

## Implementation plan

1. Add a failing test to `experience.title-team-comma.test.ts` for `Title, Team` + next-line `Company | City, ST  Dept`.
2. Make the comma-suffix split preserve the next-line employer lookup rather than short-circuiting it; recover `location` from the employer line's `City, ST`.
3. Add the `company !== title` guard as a backstop.
4. Add a synthetic-persona corpus fixture (see `tests/fixtures/pdfs/README.md` — fake name, `@example.com`, real-area-code `555`-exchange phone) that fails before and passes after.

## Acceptance

- [ ] All three synthetic roles above parse with the correct `company` and `location`.
- [ ] `company` is never byte-equal to `title` for any corpus fixture.
- [ ] New synthetic-persona fixture committed; fails before, passes after.
- [ ] `probe-experience` reports `ok` on it, and `corpus.test.ts` / `corpus-roundtrip.test.ts` stay green (no `KNOWN_FAILURES` regression).
- [ ] `npm run verify` green.

## Out of scope

- The em-dash single-line company-truncation shape — that's #461.
- The `Company — Dept — Location` department-routing question — that's #463.

rohithgollapalli:
## Summary

A single-column role header of shape `Title, Team  <dates>` with the employer on the **next** line parsed wrong: the comma-suffix split short-circuited the next-line-employer lookup, **mirroring the title into `company`** (suffix-less employers) and **dropping the employer-line `City, ST`** on every variant. Confirmed by direct reproduction against `disambiguateCompanyTitle` — the issue's synthetic table was partly inaccurate (a suffixed employer like `Northwind Systems` dodges the mirror via `looksLikeCompany`; the mirror needs a suffix-less employer), so the fix targets the reproduced mechanism.

- **`mapTitleFirst`** — distinguishes the #372 layout (anchor = employer line below the title) from #466 (date on the `Title, Team` line, so it is the anchor). In the #466 case it reads the company off the separate employer line and never mirrors the title; reconciled with the #382 shared-employer banner (empty company + team set stays banner-eligible).
- **`leadingLocation` + step 3d (`employerLineLocation`)** — recovers a front-anchored `City, ST` from an employer-line cell carrying a trailing dept (`Chicago, IL  Consumer Banking`), which is neither a whole-string bare location nor a trailing suffix, so no prior pass claimed it. Path-independent, so it also fixes the location drop on the suffixed (company-correct) path.
- **`company === title` backstop** — a company byte-identical to its title is dropped to an honest miss rather than left as mirrored bad data (per the issue).
- Extracted steps 3c/3d into helpers (`anchorCellLocation` / `employerLineLocation`) to keep `recoverLocation` under the fallow complexity gate.

Two `label-rail` corpus fixtures **already carried the mirror** on `main` (`company` = title); the backstop converts that to an honest empty company, so their score honestly drops and their now-empty company doesn't round-trip the one-line export header — baselined under the existing **#436** group.

Closes #466

## Corpus snapshots re-baked (4)

`extractedCharCount` is `countExtractedChars(parsed, sections)` — **derived from the parsed fields**, so the corrected mapping shifts it. Four snapshots re-baked to the values the current code produces (verified identical on CI and locally):

- `label-rail-grid-fragmented`, `label-rail-inline-headers` — company drops from a mirror to empty (confidence / `current_company` / `experienceChangedAcrossRoundtrip` + the derived count).
- `chromium-qualified-experience-headers`, `single-column-year-only-roundtrip` — **only** the derived `extractedCharCount` shifts; no experience field/score change (no regression).

## Test plan

- [x] `npm run typecheck` / `npm run lint` clean
- [x] New value-asserting unit tests (`experience.title-team-comma.test.ts`, #466 block): suffix-less + suffixed employers, and the no-employer-line honest miss — green through the full `extractExperience` path.
- [x] `experience.shared-banner` (#382) / `pipe-location` (#373) regressions green.
- [x] `corpus.test.ts` + `corpus-roundtrip.test.ts` green (2 label-rail experience baselines added to the #436 group).
- [x] `npm run test` green (full suite).

## Provenance

Verification: full deterministic gate green locally; corpus snapshots verified against CI extraction.

### DIFF
--- a/src/lib/heuristics/corpus-roundtrip.test.ts
+++ b/src/lib/heuristics/corpus-roundtrip.test.ts
@@ -173,6 +173,14 @@ const KNOWN_FAILURES: Record<string, Category[]> = {
   "unknown/synthetic-two-experience-sections.pdf": ["experience"],
   "unknown/two-column-achievements-sidebar.pdf": ["experience"],
   "unknown/weasyprint-cairo-two-column.pdf": ["experience"],
+  // label-rail grids: the employer is glued into the header with no delimiter
+  // ("Staff Engineer, Platform Northwind Systems"), so the parser cannot read a
+  // company. On main it MIRRORED the title into `company` (bad data that happened
+  // to round-trip mirror==mirror); #466 replaced that with an honest empty
+  // company. The now-empty company re-parses non-empty off the one-line export
+  // header — the same #436 root as the group above. Delete when #436 lands.
+  "unknown/label-rail-grid-fragmented.pdf": ["experience"],
+  "unknown/label-rail-inline-headers.pdf": ["experience"],
 };
 
 function walkPdfs(dir: string): string[] {
--- a/src/lib/heuristics/extract/experience-disambiguate.ts
+++ b/src/lib/heuristics/extract/experience-disambiguate.ts
@@ -152,6 +152,20 @@ function locationFromAnchorCell(cell: string): string | undefined {
   return isBareLocationString(c) ? c : undefined;
 }
 
+/** Front-anchored `City, ST` at the START of a cell, guarded by a valid USPS code
+ *  (`US_STATE_CODE_RE`) — the same closed-vocabulary discipline `stripLocationSuffix`
+ *  applies at the END. This catches an employer-line cell that leads with the
+ *  location and carries a trailing dept/sub-org after it ("Chicago, IL  Consumer
+ *  Banking" → "Chicago, IL"), where neither `isBareLocationString` (not a
+ *  whole-string location) nor `stripLocationSuffix` (location is not the trailing
+ *  suffix) can claim it (#466). Multi-word city allowed via the comma boundary. */
+function leadingLocation(cell: string): string | undefined {
+  const m = cell
+    .trim()
+    .match(/^([A-Z][A-Za-z.\-]+(?:\s+[A-Z][A-Za-z.\-]+)*),\s*([A-Z]{2})(?:\s|$)/);
+  return m && US_STATE_CODE_RE.test(m[2]) ? `${m[1]}, ${m[2]}` : undefined;
+}
+
 function stripLocationSuffix(s: string): {
   text: string;
   location: string | undefined;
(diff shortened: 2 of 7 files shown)


s-annam (changes requested):
The direction is right and two of the four snapshot changes are genuine wins — but the new employer lookup is unguarded, and it regresses a shape `main` parses correctly. I reproduced it both at unit level and on an existing corpus fixture. CI stayed green because the corpus snapshots are keys-only and can't see the wrong value.

## Blocking

### 1. The `#466` employer find is unguarded and preempts the correct `"Title, Company"` fallthrough

`experience-disambiguate.ts` (`mapTitleFirst`):

```ts
const employer = splits.find((s) => s.source !== titleSource);
if (employer) return { company: employer.text, title, team: splits[1].text };
```

This takes the first segment from **any** other header line as the company — including a line that is just a location. It also fires *before* the `return { company: splits[1].text, title }` fallthrough, which is the branch that correctly handles a suffix-less single-line `"Title, Company"`.

Concretely, `tests/fixtures/pdfs/unknown/chromium-qualified-experience-headers.pdf`, role 3. Source text:

```
Senior Support Specialist, Maple Tech Helpdesk
Remote

Aug 2017 - May 2019
```

```
disambiguateCompanyTitle(
  ["Senior Support Specialist, Maple Tech Helpdesk", "Remote", "Aug 2017 - May 2019"], 2)

main       → { company: "Maple Tech Helpdesk", title: "Senior Support Specialist", location: "Remote" }   ✅
this branch→ { company: "Remote",              title: "Senior Support Specialist", team: "Maple Tech Helpdesk" }   ❌
```

The location line is read as the employer; the real employer is demoted to `team`. Roles 1–2 of that same fixture are the identical shape and survive only because `Retail Group` / `Co.` trip `looksLikeCompany` — **any suffix-less employer falls into this branch.**

Two things make this worse than the mirror it replaces:

- it evades this PR's own `company === title` backstop (`"Remote" !== title`), and
- `current_company` stays in `fieldsPopulated`, so the keys-only corpus snapshot stays green. The only trace in the whole diff is `extractedCharCount: 1138 → 1119` in that fixture's `.expected.json`.

That inverts #466's own thesis — *"prefer `null` over a mirrored title, so a miss reads as a miss instead of as bad data"*. A detectable mirror was traded for an undetectable plausible-but-wrong employer.

**Fix:** guard the find the way the `#372` path already guards `anchorCompany`, and let the fallthrough win when nothing looks like an employer:

```ts
const employer = splits.find(
  (s) =>
    s.source !== titleSource &&
    !looksLikeLocationTail(s.text) &&
    leadingLocation(s.text) === undefined &&
    locationFromAnchorCell(s.text) === undefined,
);
```

### 2. The `looksLikeLocationTail` guard on `anchorCompany` is now dead code

When that guard correctly rejects a location-leading anchor cell, the unguarded `employer` find on the very next line claims the same cell anyway — and the location gets shredded on the way through:

```
disambiguateCompanyTitle(["Product Designer, Growth", "New York, NY | Jan 2020 - Dec 2021"], 1)

main       → { company: "Growth", title: "Product Designer", location: "New York, NY" }   (company wrong, location RIGHT)
this branch→ { company: "New",    title: "Product Designer", team: "Growth", location: "York, NY" }   (both wrong)
```

`anchorCompany` is suppressed as intended, then `employer` grabs the rejected cell, and `stripLocationSuffix` shears `"New York, NY"` into company `"New"` + location `"York, NY"`. Strictly worse than `main`, which got the location right. Same root cause and same fix as #1 — the guard has to sit on the employer find, not only on `anchorCompany`.

## Secondary

### 3. No corpus fixture — which is exactly why the above shipped green

Issue #466's AC asked for a synthetic-persona fixture that fails before and passes after. `git diff --diff-filter=A origin/main...HEAD -- 'tests/fixtures/**'` is empty; the three new tests build header arrays in memory, so the repro shape never goes through the real `runCascade` PDF path. The `*.expected.json` snapshots are keys/counts-only by design and cannot assert field **values**, so a wrong-but-populated `company` is invisible to `verify`. AC 2 (`company` never byte-equal to `title` on any corpus fixture) also has no test — the backstop is enforced in code only.

### 4. The new tests miss the one shape that breaks

Test 1 and 2 both have an employer on the next line; test 3 has no next line at all. The uncovered case is *next line exists but is not an employer* — i.e. the pre-existing `#372` `"Office manager, Nod Publishing"` case with one more line under it. Add it; it fails here and passes on `main`:

```ts
disambiguateCompanyTitle(
  ["Senior Support Specialist, Maple Tech Helpdesk", "Remote", "Aug 2017 - May 2019"], 2)
```

### 5. The `KNOWN_FAILURES` comment misdescribes both fixtures it baselines

It says the employer is *"glued into the header with no delimiter (`Staff Engineer, Platform Northwind Systems`)"*. `pdftotext` on both label-rail PDFs shows **no employer text on the page at all** — the block is `Staff Engineer, Platform` / `Aug 2024 - Present` / bullets.

To be clear, the behaviour change on these two is **correct** and I verified it: `main` really was mirroring the title into `company`, and the empty company is now the honest parse of a résumé that has no employer. But the justification sitting next to a permanent baseline entry is wrong, and that comment is what the next maintainer reads when #436 lets them delete the line. Suggest: *"the source carries no employer text, so `company` is correctly empty; the empty company re-parses non-empty off the one-line export header — same #436 root as the group above."*

### 6. `employerLineLocation` isn't restricted to the employer line

The name and docblock promise the employer line, but the loop is `for (const s of splits)` with no `source` restriction — only a text-equality skip against `company`/`title`. A leading `City, ST` on the title or date line is claimed just as readily, and because the skip is by text equality, two cells with identical text are both skipped. Narrow it to the employer-line `source`.

## Nits

- `single-column-year-only-roundtrip.pdf` is a real, unremarked win worth noting in the PR body: `team: "Austin, TX ·"` → `location: "Austin, TX"`. Good.

---

For what it's worth, the diagnosis in the docblocks is accurate and the three-way anchor-placement analysis is the right decomposition — the bug is entirely in the *unguarded* find, not in the model of the problem. Add the guard and the corpus fixture and this lands.

Verdict rule applied: ≥1 Blocking → `REQUEST_CHANGES`.

Gates: fixture-PII **n/a** (no fixture binary added; `check:fixtures` isn't on this base yet — it lands with #478). Design-system / style-token / skill-script gates **n/a** (no UI, component, or script files). `fallow` + `verify` green in CI. Correctness pass run as a differential execution against `origin/main` rather than a static read, since a pure-logic parser diff makes before/after output decisive.

---

rohithgollapalli:
Superseded by #483 (the #460 F&F epic), which **closed #466** with a more complete version of this exact fix:

- `mapTitleFirst` restructured into four explicit cases — the delim-only anchor restriction (Case 1 vs the comma-split anchor cases) is precisely the guard for the location-as-company bug @s-annam flagged here.
- The end-of-pipeline `company === title` backstop.
- `recoverLocation` extended to scan below-anchor **whole** cells for a bare location — the same `Remote`-as-location recovery this PR's review needed (their comment: *"PR #483 review — also scan below-anchor WHOLE cells (a bare City, ST line)"*).

@s-annam's blocking review is resolved on `main` by #483. Closing as superseded; nothing here to carry forward.

Issue #467, opened 2026-07-13
[parser] routing — "<QUALIFIER> EXPERIENCE" headers unrouted and RELEVANT COURSEWORK merged into education → block bleed + phantom education entry


s-annam:
## Why

Child of #460 (F&F trust regression epic). Surfaced by a fresh `probe-experience` + `probe-education` dogfooding pass on a real single-column résumé (uncommitted, PII-bearing, per probe policy — all examples below are **synthetic**).

#462 fixes one instance of a general failure: a section header that the router doesn't recognize **bleeds into the preceding section**, and the downstream entry parser then **fabricates an entry** out of the pooled body. #462's instance is the compound `X & Y` header. This issue covers the other two header families that reach the *same* code path on ordinary student/early-career résumés:

1. **`<QUALIFIER> EXPERIENCE`** — `RELEVANT EXPERIENCE`, `ADDITIONAL EXPERIENCE`, `PERFORMANCE EXPERIENCE`, `INVOLVEMENT EXPERIENCE`. The alias table has `involvement` but not `involvement experience`, and matching is exact-string, so the qualified form is rejected.
2. **`RELEVANT COURSEWORK`** — this one *does* match (it's an `education` alias), which is the bug: it routes into the **same region** as `EDUCATION`, so the coursework body pools into the education region and the education parser anchors a phantom second entry on it. It should be its own region (or a sub-block of education), not merged into it.

Because both land on `matchSectionHeaderDetailed` + `sections.config.json`, they should be fixed together with #462 rather than in three separate passes.

## Repro (synthetic)

```
EDUCATION
Ridgemont University
Bachelor of Music, Music Composition
| Certificate, Music Business
Graduation Month and Year
Springfield, IL

RELEVANT COURSEWORK
Electroacoustic Composition, Instrumentation and Arranging, Tonal Counterpoint

RELEVANT EXPERIENCE
Lakeside Opera                Springfield, IL
Production Intern             June 2023 - Present
• Supported departmental assignments

INVOLVEMENT EXPERIENCE
Student Composers Association  August 2022 - Present
• Created work ranging from symphonic to electronic music
Campus Outreach Program        January 2023 - May 2023
• Aided musicians in public schools
```

**Sections detected:** `profile education experience skills` — the four `<QUALIFIER> EXPERIENCE` headers collapse into a single `experience` region, and `INVOLVEMENT EXPERIENCE` survives as a **content line inside** that region.

[...]

Issue #462, opened 2026-07-13
[parser] routing — compound "X & Y" section headers unrouted → block bleeds into preceding section + education fabricates phantom degree


Vaishnavi1709:
## Problem

A compound section header of the form `X & Y` (e.g. `CERTIFICATIONS & ACTIVITIES`, `HONORS & LEADERSHIP`, `AWARDS & CERTIFICATIONS`) is not routed by the section matcher — the entire trailing block bleeds into whichever section preceded it. On real single-column résumés this manifests two ways at once:

1. **The whole compound-header block silently disappears from the output.** Certifications, honors, activities, leadership items — none render, none download.
2. **The education parser fabricates a phantom degree entry** from the pooled prose. `DEGREE_RE` fires on a body sentence like `Graduated B.E. with Distinction; mentored 3 interns at <BigBank>` and treats it as a new education entry header, producing garbage like:
   ```
   degree="B.E."
   field="with Distinction; mentored 3 interns at <BigBank> (all converted to"
   institution="Leadership: Led CSR initiatives; competed in inter-college hackathons"
   dates=null
   ```

Both symptoms have a single root cause: routing rejects the `X & Y` header, the pooled block lands in whichever section came before it, and the education entry-boundary detector isn't strict enough to reject a `DEGREE_RE` hit that lives in body prose.

### Repro

Synthetic single-column tail-of-résumé shape (mirrors what real résumés produce):

```
EDUCATION
M.Sc. in Computer Science — <Some University> (Los Angeles campus) (Jul 2024 – Dec 2025)
B.E. in Computer Science — <Some College> · GPA 8.0 / 10 (Aug 2017 – Jun 2021)

CERTIFICATIONS & ACTIVITIES
Certifications: AWS Certified Cloud Practitioner
Achievements: Graduated B.E. with Distinction; mentored 3 interns (all converted to FT)
Leadership: Led CSR initiatives; competed in inter-college hackathons
```

Expected: two education entries; one `certifications` section (or `achievements` — either is fine — carrying the four items).

Actual (per `probe-education` + probe-run against the reconstructed résumé UI):
- **Sections detected:** `education`. **No `certifications`.** No `achievements`. The four `CERTIFICATIONS & ACTIVITIES` lines are absorbed into `education`.
[...]

PR #482, opened 2026-07-14
fix(parser): route "<QUALIFIER> EXPERIENCE" headers via the exact-alias tier (#467)

rohithgollapalli:
## Summary

A qualified experience header — `Relevant / Additional / Performance / Involvement Experience` — matched only via the soft head-noun **anchor-fallback** tier (`viaAnchorFallback: true`). The splitter suppresses a **repeat** anchor-fallback open of an already-open section (#258 Layer B), so a **second** qualified experience header was swallowed as a content line inside the first experience region — stranding the roles beneath it without an entry boundary (`INVOLVEMENT EXPERIENCE` in the #467 repro).

Fix: fold a **closed vocabulary** of qualifier words + `experience` to the bare `experience` anchor in `matchSectionHeaderDetailed`, so these route via the **exact** tier (`viaAnchorFallback: false`) and each opens its own boundary. The closed set (not any Title-case word) keeps prose from minting a section, and the anchor-fallback guards still gate everything outside it. Open-ended qualifiers (`Customer Service Experience`) stay on the soft tier, unchanged — this is a `normalizer`, not an enumeration of every permutation (the gap the issue calls out).

Reproduced end-to-end before/after via the real section splitter; confirmed `INVOLVEMENT EXPERIENCE` flips from a swallowed content line to its own experience section boundary.

Refs #467

## Scope note (why not `Closes`)

#467 bundles three fixes. This PR lands the **routing** half. The other two — splitting `RELEVANT COURSEWORK` into its own region, and the education entry-boundary guard against a phantom degree — **only reproduce from real-résumé line assembly** (verified: `"Certificate, Music Business"` is inert to the education chunker; `extractEducation` yields one clean entry on every synthetic form). Per the house rule of verifying a parser fix against a reproducing case before shipping, those are deferred to **#462**, which was filed from a résumé that reproduces the phantom and shares the exact same guard.

## Corpus

One snapshot re-baked: `google-docs-skia-proxy-programs-skills-software` — a `Leadership Experience` header previously absorbed as content is now consumed as a boundary (section `lineCount` 24 → 23). No experience/score regression; `extractedCharCount` unchanged.

## Test plan

- [x] `npm run typecheck` / `npm run lint` clean
- [x] New unit tests: closed-qualifier fold routes via the exact tier (`regex.test.ts`); open-ended qualifier stays on the soft tier; prose still rejected.
- [x] New splitter test (`sections.test.ts`): a second `<QUALIFIER> EXPERIENCE` opens its own section — header never survives as a content line.
- [x] `corpus.test.ts` + `corpus-roundtrip.test.ts` green.
- [x] `npm run test` green (2252 passed).

## Provenance

Verification: full deterministic gate green locally; corpus snapshot verified against CI extraction.

### DIFF
--- a/src/lib/heuristics/regex.test.ts
+++ b/src/lib/heuristics/regex.test.ts
@@ -2,7 +2,12 @@
 // Copyright 2026 The resumelint Authors
 
 import { describe, it, expect } from "vitest";
-import { matchSectionHeader, matchSectionAnchorToken, DATE_RANGE_RE } from "./regex.ts";
+import {
+  matchSectionHeader,
+  matchSectionHeaderDetailed,
+  matchSectionAnchorToken,
+  DATE_RANGE_RE,
+} from "./regex.ts";
 import { parseDateRange, stripDateRange } from "./line-primitives.ts";
 
 describe("matchSectionHeader — split-letter headers (#56)", () => {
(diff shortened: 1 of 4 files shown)


rohithgollapalli:
Rebased onto `main` (d9b020e, post-#483).

**Scope after #483:** #483 (the #460 epic) closed #466/#461/#462/#464 and handled the **education** side of #467 (the `isRealEntryHeader` phantom-entry guard). It did **not** touch the **qualifier-EXPERIENCE routing** — origin/main's `regex.ts` has no qualifier fold, so a second `<QUALIFIER> EXPERIENCE` header is still swallowed as a content line on `main` (verified). This PR is that remaining routing half: the closed-vocab fold that routes `Relevant / Additional / Performance / Involvement Experience` via the exact-alias tier so each opens its own boundary.

Rebased clean (no overlap with #483's `mapTitleFirst` / education changes). Gates green on the new base: typecheck, lint, `check:fixtures`, full suite (2351), corpus + roundtrip. The one re-baked snapshot (`programs-skills-software`, `Leadership Experience` header now a boundary) still holds.

Note: #467 is closed (via #483's education work) — happy to relink this to a fresh routing-only issue if you'd prefer the PR track a live issue.

s-annam (dismissed):
**APPROVE** — tight, well-targeted pure-logic parser fix. Verdict rule: 0 Blocking → APPROVE (the two notes below are non-blocking).

Verified the fix end-to-end, not just by reading the diff:

- `stripExperienceQualifier` only folds when the last token is `experience` **and** every preceding token is in the closed `EXPERIENCE_QUALIFIERS` set; single-word inputs (`length < 2`) pass through unchanged and are still matched by `normalized`. Sound.
- `dequalified` is always `"experience"`, which is an alias of **only** the `experience` section — no cross-section collision, and the old `keywords.includes(normalized)` is preserved as the first `||` operand, so no prior behavior is removed.
- Corpus snapshot `google-docs-skia-proxy-programs-skills-software` (24 → 23) checks out: `reproArtifact.buildSections` coalesces by `canonical.sections.byName`, so the second experience region merges into one entry, and the drop of exactly one line is the `Leadership Experience` header no longer surviving as a content line. Not a regression.
- Touched tests green locally (133 passed), `typecheck` clean.

## Secondary

- `regex.ts` (~L528, the `keywords.includes(dequalified)` fold) — routing through the exact-alias tier bypasses the anchor-fallback casing/word-count guards, so a lowercase ≤40-char line consisting only of qualifier words + `experience` (e.g. a stray prose line `additional experience`) now folds and mints an experience section, where pre-fold it hit the anchor-fallback tier and was rejected for not being header-cased. Real-world risk is low and it is symmetric with the pre-existing bare-`experience` exact match, but it is a genuinely new false-positive surface — the "closed vocabulary keeps prose from minting a section" reasoning in the docblock only addresses Title-case words, not lowercase all-qualifier lines. Worth a one-line acknowledgement in the comment, not a blocker.

## Nits

- `EXPERIENCE_QUALIFIERS` (L~334) is a per-word closed list layered over the #258 Layer-B splitter suppression rather than a fix to *why* an exact-vs-soft repeat open is suppressed unevenly. Each new qualifier the parser should treat as experience (`Volunteer Experience` is deliberately excluded) needs another entry here. The deeper fix would generalize; the targeted list is reasoned and documented, so acceptable — noting the altitude only.

Scope note (routing half of #467, coursework/education deferred to #462 pending a reproducing fixture) is the correct call under the verify-against-a-repro house rule.

---

PR #483, opened 2026-07-14
fix(parser): F&F trust regression on well-formatted single-column résumés (#460)

### LINKED ISSUE #460: Epic — F&F trust regression: parser blockers on well-formatted single-column résumés
## Why

The P1 · Friends & Family milestone is currently 0 open / 264 closed. The gate is simple: **nothing visibly broken on a well-formatted single-column résumé** (drop → correct parse + score → safe download). A dogfooding pass on a real single-column PDF surfaced **four parse-layer regressions**, none covered by an open issue, all trust-destroying on the F&F path:

- Multi-word company names truncated by one word on every role, with the missing word glued onto the location field.
- Compound section headers of the form `X & Y` (e.g. `CERTIFICATIONS & ACTIVITIES`, `AWARDS & HONORS`) don't route — the whole block bleeds into the preceding section, and the education parser fabricates a phantom degree entry from the pooled body content.
- Projects section drops per-project descriptions and merges/loses secondary projects when the project bodies are prose paragraphs rather than `•` bullets.
- (Sibling: skills-section extraction shreds multi-word tokens on bulleted-single-column layouts — tracked under **#415** rather than here, because a skills epic already owns that lane.)

These are not two-column edge cases and not exotic templates. They fire on the base case the F&F milestone is defined to protect. The score itself becomes untrustworthy on such résumés because the corrupted parse propagates into the specificity / structure calculations. That means we can't in good faith call P1 · Friends & Family "closed" until they land.

## Scope

Three children, each independently implementable and independently testable:

- **#B1 — `[parser] experience` — multi-word company truncated + last word stolen into location, every role, single-column em-dash headers.**
  Layer: `extract/experience-disambiguate.ts` — the location-recovery ladder's country-anchored walk-back consumes the trailing token of the company when the source line has the shape `Title — Multi-Word Company, Location`. Precedent: closed two-column analogues #287, #347, #368, #369, #382. This one is single-column, and #413 (the prior experience-parser batch) closed without covering it.

- **#B2 — `[parser] routing` — compound `X & Y` section headers unrouted → block bleeds into the preceding section + education fabricates a phantom degree entry.**
  Layer: `matchSectionHeaderDetailed` in `regex.ts:468` + `sections.config.json` aliases + education entry-boundary guard in `extract/education.ts`. The `achievements` section already lists compound aliases (`awards & honors`, `honors & awards`, etc.) as precedent; `certifications` does not, so `CERTIFICATIONS & ACTIVITIES` is exact-match-rejected and the whole trailing block is absorbed by whichever section preceded it. The phantom-education side-effect is a separate but linked failure: `DEGREE_RE` fires on `Graduated B.E. with Distinction` in a body sentence and anchors a new education entry.

- **#B5 — `[parser] projects` — per-project descriptions and second/subsequent projects dropped when project bodies are prose paragraphs, not bulleted.**
  Layer: `parseEntryBlocks` under `anchor: "first_line"` (`extract/projects.ts:39`) + the entry-block segmentation in `entry-blocks.ts`. A prose-body project has `bulletCount === 0`, so `projectFromBlock` (`extract/projects.ts:120`) scores it below the retention threshold and the description is not attached; two projects separated only by a blank line rather than a `•` list are collapsed into one entry-block, losing the second.

The skills-shredding regression (extraction of multi-word skills in bulleted-single-column layouts, with category labels ingested as skills) is scoped as a **new child of #415** — the existing skills-recognition epic — rather than this one, because #415 explicitly says "New skills failure classes found via `probe-skills` land here as additional children" and its extraction-layer surface is the right home. Linked from here for tracking, but the parent is #415, not this epic.

## Tooling

The four `probe-*` skills (`probe-contact`, `probe-experience`, `probe-education`, `probe-skills`) plus `probe-roundtrip` already localize each blocker to a layer. Each child issue below cites its probe output category (never values, per the fixture PII policy) and names the file:line of the suspected fix site. New failure classes surfaced during implementation land as additional children.

## Children

- [ ] **#461** — `[parser] experience` — multi-word company truncated + last word stolen into location on single-column em-dash headers.
- [ ] **#462** — `[parser] routing` — compound `X & Y` section headers unrouted → block bleeds into preceding section + education fabricates phantom degree.
- [ ] **#464** — `[parser] projects` — per-project descriptions and secondary projects dropped when bodies are prose, not bulleted.

Added from the second dogfooding pass (verdict comment below):

- [ ] **#466** — `[parser] experience` — company mirrors the title and location drops when the role header carries a `Title, Team` comma suffix and the employer sits on the next line. Distinct mechanism from #461 — fixing #461 does not fix it.
- [ ] **#467** — `[parser] routing` — `<QUALIFIER> EXPERIENCE` headers unrouted and `RELEVANT COURSEWORK` merged into the education region → block bleed + phantom education entry. Shares #462's fix sites (`matchSectionHeaderDetailed` + `sections.config.json` + the education entry-boundary guard); sequence the two together.

Linked (parent is #415, not this epic):
- **#465** — `[parser] skills` — bulleted-single-column labelled skills shred multi-word tokens and ingest category labels.

## Acceptance (epic-level)

- [ ] Each child (#461, #462, #464, #466, #467) fixed with a synthetic-persona corpus fixture that fails before and passes after.
- [ ] For every child fixture, the corresponding `probe-*` reports `ok` and `corpus.test.ts` / `corpus-roundtrip.test.ts` stay green (no `KNOWN_FAILURES` regression).
- [ ] A **fresh dogfooding pass** on a real single-column résumé (uncommitted, PII-bearing, per probe policy) reports no new blocker-class findings across contact / experience / education / projects / skills. Report the probes' verdicts in a comment here, values scrubbed.

## Out of scope

- Two-column layout regressions (already tracked under #436 / #437 and closed follow-ups).
- Score-transparency work (a P2 · Design Partner concern — this epic is scoped to correcting the parse the score is computed from, not the scoring formula).
- The skills-shredding failure class — tracked under **#415**.
- Cosmetic / edit-interaction display polish (tracked under **#391**).

### LINKED ISSUE #461: [parser] experience — multi-word company truncated + last word stolen into location on single-column em-dash headers
## Problem

On a well-formatted single-column résumé with role headers shaped `Title — Multi-Word Company, Location`, the parser steals the **last word of the company** and glues it onto the location field. This fires on every role on real single-column PDFs, not just one edge case — the Friends & Family gate scenario is completely broken for a common role-header shape.

### Repro

Synthetic role header shapes (mirroring what real résumés produce):

```
Full-Stack Software Developer — <Multi-Word Bank Name>, India
Back-End Engineering Intern — Citi <BankSuffix>, India
Java Developer Intern — <Two-Word Company Name>, India
Full-Stack Software Engineer Intern — <Two-Word Company>, Inc., USA
Software Engineering Intern — <Three-Word Company> Ltd., India
```

Expected parse:

| Company | Location |
|---|---|
| `<Multi-Word Bank Name>` (two words) | `India` |
| `Citi <BankSuffix>` | `India` |
| `<Two-Word Company Name>` | `India` |
| `<Two-Word Company>` | `Inc., USA` *or* `USA` |
| `<Three-Word Company> Ltd.` | `India` |

Actual (per `probe-experience`):

| Company | Location |
|---|---|
| `<first word only>` | `<second word>, India` |
| `Citi` | `<BankSuffix>, India` |
| `<first word only>` | `<second word>, India` |
| `<Two-Word Company>` | `Inc., USA` |
| `<Three-Word Company>` | `Ltd., India` |

Every multi-word company loses its last token. Location becomes nonsense (`Bank, India`, `Corner, India`, `Ltd., India`, `Inc., USA`).

### Where in the loop

Parse layer. Root cause is in `stripLocationSuffix` at [experience-disambiguate.ts:155-229](src/lib/heuristics/extract/experience-disambiguate.ts#L155-L229), specifically **Pass D — space-delimited international "…Company City, Country"** ([experience-disambiguate.ts:201-231](src/lib/heuristics/extract/experience-disambiguate.ts#L201-L231)):

```ts
const SPACE_INTL_RE =
  /\s+([A-Z][A-Za-z.\-]+),\s*([A-Z][A-Za-z.\-]+(?:\s+[A-Z][A-Za-z.\-]+)*)$/;
const mSpaceIntl = s.match(SPACE_INTL_RE);
if (
  mSpaceIntl &&
  COUNTRY_GAZETTEER.has(mSpaceIntl[2].toLowerCase()) &&
  !LOCALITY_SUFFIX_RE.test(mSpaceIntl[1])
) {
  // strips " <lastCompanyWord>, <Country>" as location
```

`SPACE_INTL_RE` matches any `space Capitalized-word, Country` at the end of the string. The two guards — `COUNTRY_GAZETTEER` membership and `LOCALITY_SUFFIX_RE` rejection — are **insufficient** for real company names:

- **Common corporate name tails** (`Bank`, `Corp`, `Group`, `Systems`, `Solutions`, `Technologies`, `Corner`, `Studios`, `Media`, `Software`, `Consulting`, `Partners`) are ordinary capitalized nouns — none appear in `LOCALITY_SUFFIX_RE` (`city|town|beach|springs?|heights|falls|hills|park|bay|harbou?r|grove|gardens?|valley|shores?`), so the guard passes. Pass D happily strips `Bank, India` off `Deutsche Bank, India` and yields company `Deutsche`, location `Bank, India`.
- **Legal suffixes** (`Inc.`, `Ltd.`, `LLC`, `GmbH`, `S.A.`) are caught by `cityStartsWithCompanyText` in Pass C ([experience-disambiguate.ts:109-112](src/lib/heuristics/extract/experience-disambiguate.ts#L109-L112)) but **Pass D omits that guard entirely**. So `Ltd., India` and `Inc., USA` fall through to Pass D and get peeled as location.

The two-column analogues #287 / #347 / #368 / #369 / #382 closed without the single-column em-dash + intl-suffix case in scope. #413 (the prior experience-parser batch) closed with `disambiguateCompanyTitle` decomposed but Pass D's guards unchanged.

## Proposed change

Tighten Pass D with two additional guards, mirroring Pass C's discipline:

1. **Apply `cityStartsWithCompanyText` to Pass D.** Pass C already rejects `Inc.` / `Ltd.` cities via `cityStartsWithCompanyText`. Pass D omits this — add it. This alone fixes `Inc., USA` and `Ltd., India`.
2. **Introduce a `COMPANY_TAIL_TOKENS_RE`** in `experience-disambiguate.ts` for common corporate name tails: `Bank | Corp | Corporation | Group | Systems | Solutions | Technologies | Corner | Studios | Media | Software | Consulting | Partners | Ventures | Holdings | Industries`. Pass D defers when `mSpaceIntl[1]` matches — the token is more likely company tail than a bare city.

Both guards are closed-vocabulary and defer rather than change classification (mirroring `LOCALITY_SUFFIX_RE`'s and `cityStartsWithCompanyText`'s design): a false negative leaves the whole string as company (recoverable via inline edit), whereas the current false positive silently corrupts every role header.

**Deliberately not proposed:** an open-shape "single-word company remnant looks wrong" guard. `Citi Bank` reduces to `Citi` which IS a real one-word company (Citigroup uses `Citi`); a single-token remnant rule would over-correct. The closed-vocabulary corporate-tail list is safer.

## Implementation plan

1. **Add a corpus fixture** at `tests/fixtures/pdfs/unknown/single-column-intl-role-headers.pdf` — synthetic-persona (fake name, `@example.com`, `(312) 555-0123` per the fixture PII policy), single-column, five role headers covering the five shapes above: `Word Word, Country`, `Word CommonBankTail, Country`, `Word Word Ltd., Country`, `Word Word, Inc., USA`, `Word Word Word Ltd., Country`. Re-export a template with synthetic data; do not download an OSS résumé template's own author-PII example.
2. **Add unit tests** in `src/lib/heuristics/extract/experience-disambiguate.test.ts` (or the closest existing sibling — `experience.location.test.ts`) asserting:
   - `Deutsche Bank, India` → company `Deutsche Bank`, location `India`.
   - `Citi Bank, India` → company `Citi Bank`, location `India`.
   - `IT Corner, India` → company `IT Corner`, location `India`.
   - `JMK Infosoft Solutions Ltd., India` → company `JMK Infosoft Solutions Ltd.`, location `India`.
   - `116 Ideas, Inc., USA` → company `116 Ideas, Inc.`, location `USA` (or `USA` alone with `Inc.` staying on company — pick one and assert consistently).
   - Regression: existing passing shapes stay unchanged — spot-check `Kasa Seoul, S.Korea` (#287) still peels `Seoul, S.Korea` as location; `Google Mexico City, Mexico` still defers per `LOCALITY_SUFFIX_RE`.
3. **Add `COMPANY_TAIL_TOKENS_RE`** in `experience-disambiguate.ts` — a `RegExp` with the closed vocabulary listed above.
4. **Extend Pass D's guard block** ([experience-disambiguate.ts:222-226](src/lib/heuristics/extract/experience-disambiguate.ts#L222-L226)) to add the two new rejections:
   ```ts
   if (
     mSpaceIntl &&
     COUNTRY_GAZETTEER.has(mSpaceIntl[2].toLowerCase()) &&
     !LOCALITY_SUFFIX_RE.test(mSpaceIntl[1]) &&
     !cityStartsWithCompanyText(mSpaceIntl[1]) &&
     !COMPANY_TAIL_TOKENS_RE.test(mSpaceIntl[1])
   ) { ... }
   ```
5. **Run `corpus.test.ts` and `corpus-roundtrip.test.ts`.** If a fixture currently baselined on `experience` in `KNOWN_FAILURES` (#293) starts passing, remove it — per the ratchet's stale-entry rule.
6. **Re-run `probe-experience`** on the new fixture. Verdict `ok`; each role's `company` and `location` match the expected mapping above.

## Reuse analysis

**Capability:** stripping a trailing location suffix off a role header.

**Existing surfaces found:** `stripLocationSuffix` and its Passes A–D in `src/lib/heuristics/extract/experience-disambiguate.ts:155`.

**Decision: extend by tightening Pass D's guard set.** No new abstraction. The two added guards (`cityStartsWithCompanyText`, `COMPANY_TAIL_TOKENS_RE`) are the same shape as the existing `LOCALITY_SUFFIX_RE` guard — closed-vocabulary defers, not open-shape rules. Keeping Pass D linear preserves the existing "correctness contract is the fixed order of passes" reading of the function.

## Acceptance criteria

- [ ] New fixture `tests/fixtures/pdfs/unknown/single-column-intl-role-headers.pdf` exists, synthetic-persona verified via `pdftotext … | head -40` before merge.
- [ ] `probe-experience` on the new fixture: each role's `company` includes every source token; each `location` contains no company-tail word.
- [ ] Unit tests in `experience-disambiguate.test.ts` (or `experience.location.test.ts`) covering all five repro shapes and the two regression shapes pass.
- [ ] `corpus.test.ts` and `corpus-roundtrip.test.ts` stay green; any stale `experience`-category baseline entry that now passes is removed from `KNOWN_FAILURES`.

## Out of scope

- **The `Inc., USA` corp-suffix pattern's ideal end-state** (is the right company `116 Ideas` or `116 Ideas, Inc.`?) — pick one during implementation and stick with it; either is correct relative to the current corruption. The important gate is: no `Inc.` appears in `location`.
- Two-column layout regressions (already tracked under #436 / #437 and closed follow-ups).
- Round-trip fidelity of the corrected headers through Download PDF — that's covered by `corpus-roundtrip.test.ts`'s existing gate and should stay green.

## Parent

Child of **#460** — Epic: F&F trust regression: parser blockers on well-formatted single-column résumés. Sibling of **#462** (compound section headers) and **#464** (projects descriptions dropped).

### LINKED ISSUE #462: [parser] routing — compound "X & Y" section headers unrouted → block bleeds into preceding section + education fabricates phantom degree

### LINKED ISSUE #464: [parser] projects — per-project descriptions and secondary projects dropped when bodies are prose paragraphs, not bulleted
## Problem

On a Projects section whose per-project bodies are **prose paragraphs** rather than `•` bullets — a common résumé shape, especially for engineers describing project scope in a single paragraph plus a tech-stack line — the parser silently drops both the descriptions and the second-and-subsequent project entries.

### Repro

Synthetic single-column Projects section (mirroring what real résumés produce):

```
PROJECTS

<Project Name One>
React, Next.js, TypeScript, Tailwind CSS
Built a resume-review platform providing real-time feedback and ATS scoring.
Optimised client-side rendering with responsive interfaces.

<Project Name Two> (backend systems project)
Java, Spring Boot, Kafka, Redis, Docker, Terraform, AWS EC2
Personal project demonstrating distributed-systems fundamentals: real-time A/B
testing, Redis-backed caching, Kafka async event processing.
```

Expected: two projects, each with a tech-stack line + a prose description carried through to the reconstructed résumé and Download PDF.

Actual (per dogfooding against the reconstructed-résumé UI):
- **Only the first project name renders.** `<Project Name One>` appears; the tech-stack line and the two-sentence description do not.
- **The second project vanishes entirely.** Neither `<Project Name Two>` nor its tech stack nor its description appear.
- The UI displays the placeholder message `No bullet-shaped lines detected.` under the first project — a "did the tool even see this?" signal.

### Where in the loop

Parse layer. Two coupled failures in the entry-block segmentation:

- **Prose lines below a project name are absorbed as `headerLines`, not routed to `body`.** Projects use `parseEntryBlocks({ anchor: "first_line", collectBody: true })` ([extract/projects.ts:39-42](src/lib/heuristics/extract/projects.ts#L39-L42)). Under the `first_line` anchor, `collectAnchors` ([entry-blocks.ts:296-336](src/lib/heuristics/entry-blocks.ts#L296-L336)) collapses consecutive non-bullet lines into a single anchor's header run — the "multi-line header" rule at [entry-blocks.ts:308-317](src/lib/heuristics/entry-blocks.ts#L308-L317). That rule is correct for a two-line project header (`Name` / `Subtitle`), but on a prose-body project every description line is also a non-bullet line at the same x as the name, so the entire project — name, tech stack, prose paragraph — collapses into one `headerLines` run with an empty `body`. `projectFromBlock` then sets `description = block.body = ""`, so nothing downstream sees the paragraph.
- **The second project's name is inside that same collapsed header run.** With no bullet marker to break the run and no x-indent difference between the two project names' font runs (both bold, both at the same left margin, both preceded by a blank line that pdfjs's line-grouping doesn't emit as a distinct line), `collectAnchors` produces one anchor for the whole section. `finalizeEntries` then hands back a single `ResumeProject` whose `name = "<Project Name One>"` and whose block body is empty; `<Project Name Two>` is silently absorbed.

The downstream "No bullet-shaped lines detected." string is UI copy from `ReconstructedResume.tsx` (or a nearby feature component) that fires when `bulletCount === 0`, exposing the parser's zero-body assumption. That copy is a symptom, not the bug.

## Proposed change

Two-part fix, both in `src/lib/heuristics/entry-blocks.ts` (and its shared consumers):

### Part 1 — Break the header run when a line looks like a body paragraph

Refine the "consecutive header-level non-bullet lines collapse into one header" rule ([entry-blocks.ts:308-317](src/lib/heuristics/entry-blocks.ts#L308-L317)) to defer when the follow-on line looks like body prose rather than a subtitle. Concretely, a header-continuation line is expected to be short + label-like (a subtitle, a tech-stack CSV, a location, a date). A line that:

- ends in sentence-terminating punctuation (`.`, `!`, `?`), OR
- exceeds a length threshold (say, > 60 characters), OR
- contains no CSV commas AND opens with a verb-form indicator (`Built`, `Designed`, `Implemented`, `Developed`, `Led`, …)

is a **body paragraph**, not a header continuation. Route it into `body` instead of absorbing it as another `headerLine`, and terminate the current header run.

### Part 2 — Detect a new project entry inside a collapsed header run

For the `first_line` anchor, once a body paragraph closes a header run (Part 1), the next non-bullet line at the same header-level x is a **new project header**, not a continuation. That's what `collectAnchors` should have inferred from a bullet-marker break; without bullets, Part 1's body-paragraph termination is the equivalent signal.

This falls out of Part 1 mostly for free — once Part 1 stops absorbing prose into `headerLines`, `collectAnchors`'s existing `i > 0` guard sees `prev` as a body-shape line rather than a header-shape line, and admits the next non-bullet line as a fresh anchor. The only load-bearing change is that `collectAnchors` needs to consider `prev.isBodyShape` (as computed by Part 1) alongside `isBulletLine(prev)` when deciding whether the current line opens a new entry.

**Deliberately not proposed:** an x-position based project-boundary rule. Real-world PDFs from Word / Google Docs / LaTeX put the project name at the same x as the description prose (both flush-left, both same paragraph style); an x-difference rule would work on templates that indent descriptions but fail on the flat layouts the F&F gate is meant to cover.

## Implementation plan

1. **Add a corpus fixture** at `tests/fixtures/pdfs/unknown/single-column-projects-prose-body.pdf` — synthetic-persona (fake name, `@example.com`, `(312) 555-0123` per the fixture PII policy), single-column, two prose-body projects each with a tech-stack line + 2-3 sentence description, no `•` bullets anywhere in Projects. Re-export a template with synthetic data.
2. **Add unit tests** in `src/lib/heuristics/entry-blocks.test.ts`:
   - `parseEntryBlocks` under `first_line` on a two-project prose-body section returns TWO blocks, each with `headerLines.length` covering `[name]` (or `[name, techStack]`) and `body` carrying the paragraph.
   - Regression: a two-line project header (name + subtitle, no body) still returns ONE block with `headerLines.length === 2`.
3. **Add unit tests** in `src/lib/heuristics/extract/projects.test.ts`:
   - The two-project prose fixture yields two `ResumeProject` entries, each with a non-empty `description`.
4. **Implement Part 1** in `entry-blocks.ts` — introduce `looksLikeBodyParagraph(line): boolean` with the three conditions above; use it inside the `first_line` branch of `collectAnchors` (and inside the header-run continuation logic used by `buildEntryBlock` / `parseBulletList`) to break the run and route the line to body.
5. **Implement Part 2** — ensure `collectAnchors` admits the line after a body-paragraph line as a new anchor. This should be a one-line adjustment to the `prev.x <= markerX` continuation guard (add `&& !looksLikeBodyParagraph(prev)`).
6. **Update `ReconstructedResume.tsx`'s "No bullet-shaped lines detected." copy** to fire only when a project has neither `description` nor bullets — a prose description should render as a paragraph without the placeholder. The exact component and copy string is in `src/components/features/` (`ReconstructedResume` or a nearby project-render helper).
7. **Run `corpus.test.ts` and `corpus-roundtrip.test.ts`.** If a fixture currently baselined on `projects` in `KNOWN_FAILURES` (#293) starts passing, remove it — per the ratchet's stale-entry rule.
8. **Verify by re-parsing the reconstructed PDF** — a prose description survives round-trip via the extended `corpus-roundtrip` gate. `parse1.projects[i].description === parse3.projects[i].description` (modulo whitespace normalisation).

## Reuse analysis

**Capability:** partitioning a section's lines into per-entry blocks.

**Existing surface found:** `parseEntryBlocks` + `collectAnchors` in `src/lib/heuristics/entry-blocks.ts`. Used by Experience (`date_range`), Education (`degree` / `date_range`), Projects (`first_line`), and Achievements.

**Decision: extend by adding a body-shape predicate to the `first_line` branch.** No new abstraction. The predicate is opt-in per anchor mode — Experience and Education both anchor on structured signals (dates, degree tokens) and are unaffected by the change; only the `first_line` anchor consults `looksLikeBodyParagraph`.

**UI reuse note (Part 6):** the "No bullet-shaped lines detected." copy should be routed through an existing empty-state primitive if `src/design-system/shared/` carries an `ErrorState`/`EmptyState` component (per CLAUDE.md's Reuse Gate). If it currently uses a raw string in the feature file, keep the string but wrap the render in a check that skips it when a `description` is present. Do not hand-roll a parallel placeholder component.

## Acceptance criteria

- [ ] New fixture `tests/fixtures/pdfs/unknown/single-column-projects-prose-body.pdf` exists, synthetic-persona verified via `pdftotext … | head -40` before merge.
- [ ] Extraction on the new fixture returns two `ResumeProject` entries; each carries a non-empty `description`.
- [ ] `parseEntryBlocks` unit tests cover both the prose-body case (2 entries) and the multi-line-header regression (1 entry).
- [ ] The reconstructed résumé UI renders both projects' descriptions; "No bullet-shaped lines detected." does not appear when a `description` is present.
- [ ] `corpus.test.ts` and `corpus-roundtrip.test.ts` stay green; any stale `projects` baseline entry that now passes is removed from `KNOWN_FAILURES`.
- [ ] Round-trip: a prose description round-trips through Download PDF (`parse1.projects[i].description === parse3.projects[i].description` up to whitespace normalisation).

## Out of scope

- **Rendering prose descriptions as pseudo-bullets in the reconstructed résumé.** Some users may prefer that visually; the acceptance bar here is that the paragraph survives, not that the UI reshapes it. A follow-up can address rendering polish.
- **The Achievements section's parallel behaviour.** Achievements also anchors on `first_line` in some paths and may share this failure mode. If a corpus fixture surfaces it during implementation, file a linked follow-up rather than expanding scope here.
- The projects tech-stack line as a first-class field. Currently it lives in `description` — extracting it into a typed `tech_stack: string[]` is #TBD (out of scope; belongs in a P3 · Public Launch resume-model refinement).

## Parent

Child of **#460** — Epic: F&F trust regression: parser blockers on well-formatted single-column résumés. Sibling of **#461** (company/location split) and **#462** (compound section headers).

### LINKED ISSUE #466: [parser] experience — company mirrors the title and location drops when the role header has a "Title, Team" comma suffix and the employer is on the next line

### LINKED ISSUE #467: [parser] routing — "<QUALIFIER> EXPERIENCE" headers unrouted and RELEVANT COURSEWORK merged into education → block bleed + phantom education entry

Vaishnavi1709:
## Summary

Fixes the five P1 · F&F parser blockers surfaced by dogfooding on real well-formatted single-column résumés. Each child issue gets a targeted heuristic change + a synthetic-persona corpus fixture, plus one exporter-side and one UI-side follow-up so the reconstructed résumé lane no longer misrepresents what the parser recovered.

- **#461** — `stripLocationSuffix` Pass D corporate-tail guards (`cityStartsWithCompanyText` + closed-vocab `COMPANY_TAIL_TOKENS_RE`) so multi-word companies keep their tail word; Pass E peels a trailing `, <Country>` (validated against `COUNTRY_GAZETTEER`) when Passes A–D defer.
- **#466** — `mapTitleFirst` restructured into four explicit cases (below-anchor delim, comma-split anchor, corporate-tail post-comma, plain title-comma). `recoverLocation` step 3c scans below-anchor delim cells with a `\s{2,}` column-gap split. End-of-pipeline `company === title` backstop clears the mirror. `buildAtsResumeModel` emits `Title, Team` (comma) instead of `Title · Team` (middot) when company is empty so round-trip parse1 == parse3 holds.
- **#462** — `matchSectionHeaderDetailed` gains a compound `X & Y` / `X and Y` tier that splits on the connective and tries each side against the exact-alias table, preferring the left side. Scales without enumerating every pairing in `sections.config.json`.
- **#462 + #467** — new `isRealEntryHeader` predicate in `extract/education.ts` gates `isDeg` / `isInst` in the chunker and `degreeLine` / `instLine` selection in `educationFromChunk`. Rejects sub-labelled body prose (`Achievements: Graduated B.E. with Distinction`) and past-tense event-narration prefixes, so mis-routed compound / qualified header blocks bleeding into education never anchor phantom entries on incidental substring hits.
- **#464** — `entry-blocks.ts` gains `looksLikeBodyParagraph` (period-terminated, > 60 chars, or verb-led without CSV comma). Used in `collectAnchors` (`first_line` branch) with an entry-header-shape gate so two prose-body projects split into two entries without over-splitting wrapped tails; used again in `buildEntryBlock`'s below-header loop as an additional body-start signal so single-sentence descriptions surface as `body`. `ReconstructedResume.tsx` renders `description` as a paragraph when a project has no `•` bullets, replacing the misleading "No bullet-shaped lines detected" placeholder.

Closes #461
Closes #462
Closes #464
Closes #466
Closes #467
Closes #460

## Test plan

- [x] `npm run verify` — exits 0, 2162 tests pass, 5 skipped
- [x] Five new synthetic-persona fixture PDFs added; snapshots baked
- [x] Fixture personas verified synthetic — `@example.com` emails + `(312) 555-01xx` reserved test numbers per repo policy (Step 3.5)
- [x] Fixture and unit-test company/institution names are Microsoft-canonical fakes (Contoso, Fabrikam, Northwind, Wingtip, Tailspin, Litware) or Ridgemont-fictional — no real orgs
- [x] Dogfooding pass in `npm run dev` on all 5 new fixtures — each shape now parses to the expected fields
- [x] Round-trip stays clean — the exporter's empty-company comma-join keeps parse1 == parse3 on `Title, Team` role shapes; no `KNOWN_FAILURES` additions
- [x] Four pre-existing corpus snapshots re-baked to reflect improved outputs (documented in commit body)

## Provenance

Verification: `npm run verify` — green locally + will re-run in CI


s-annam (changes requested):
Solid, well-documented parser work — the `#461` Pass D deferral guards, the `#462` education entry-boundary guard, and the `#466` below-anchor employer recovery all hold up, and the five new fixtures are clean under the PII gate. But three defects reproduce on normal input, one of them a silent data-corruption round-trip, so this needs another pass before merge.

Every finding below was reproduced by running the code on this branch (and, where it matters, diffed against `origin/main` to confirm it's a regression rather than pre-existing).

## Blocking

### 1. The compound `X & Y` header tier is missing the bullet guard — body bullets mint phantom sections

`src/lib/heuristics/regex.ts:513`

The new compound tier sits *between* two tiers that are both gated on `!LEADING_BULLET_RE.test(text)` — and it is not. Normalization only strips **trailing** `[:·•]`, so a leading `•` survives into `normalized`. The left side then fails the alias table (it carries the bullet), and the **right** side hits, because `leadership`, `volunteer`, `activities`, `honors`, `interests`, `projects` are all single-word aliases.

Verified by calling `matchSectionHeader` directly on this branch:

```
"• Public speaking and leadership"  => experience
"• Committee member and volunteer"  => experience
"• Dean's List and Honors"          => achievements
"• Mentored juniors and interests"  => other
```

`classifyLine` (`src/lib/heuristics/sections.ts:1872`) calls `matchSectionHeaderDetailed` on **every** line with no font/bold/length gate ahead of it and returns `{kind: "open"}` on a hit. The compound tier reports `viaAnchorFallback: false` (L1), so `isInstitutionRepeat` does not suppress it — and the bullet opens a *different* section than the one it sits in. A single skills or achievements bullet ending in a common alias word therefore splits the résumé mid-body.

This is precisely why the glyph-strip tier above and the anchor-fallback tier below are both already guarded. Fix: move the compound block inside an `if (!LEADING_BULLET_RE.test(text))` guard.

### 2. Case 3a promotes teams to companies and destroys `team` — regresses `#382` banner inheritance

`src/lib/heuristics/extract/experience-disambiguate.ts:554`

`COMPANY_TAIL_TOKENS_RE` was designed for Pass D, where a match means **defer** — a false positive there is harmless, because the text simply stays as company text (the comment says exactly this: "a false negative leaves the string as company (recoverable via inline edit)"). Case 3a reuses that same vocabulary to make the **opposite** decision: a match means **promote to company and drop the team**. A false positive is now destructive, and the vocabulary is far too broad for that: `Systems`, `Analytics`, `Health`, `Digital`, `Financial`, `Media`, `Labs`, `Solutions`, `Networks`, `Software`, `Consulting`, `Group` are all extremely common *team*-name tails — `Core Systems`, `Consumer Health`, `Growth Analytics`, `Payments Digital`.

Verified as a regression against `origin/main`, same input both sides (a `Wingtip Financial Inc.` banner over two bare `Title, Team` roles):

```
main:   { title: "Senior Engineer", company: "Senior Engineer", team: "Growth Analytics" }
branch: { title: "Senior Engineer", company: "Growth Analytics", team: undefined }
```

The team is gone. And because `isBannerContinuation` (`src/lib/heuristics/extract/experience.ts:113`) early-returns `false` when `!entry.team`, the `#382` shared-employer banner is **no longer inherited either** — so the role loses its team, gains a wrong company, and never recovers the real employer. Three fields worse than `main`, which at least kept the team recoverable.

The two call sites need different vocabularies. Case 3a should be restricted to unambiguous legal/entity markers (`Inc`, `LLC`, `Ltd`, `GmbH`, `PLC`, `Corp`, `Corporation`, `Holdings`) and drop the generic nouns; Pass D's deferral can stay broad.

### 3. Empty-company export branch skips the location case — location is silently written into `company`

`src/lib/pdf/ats-resume-model.ts:551`

The new branch is gated on `!exp.location?.trim()`, so an empty-company role that *has* a location falls through to the middot join. Verified by re-parsing the emitted header:

```
emitted:  "Software Engineer II · Chicago, IL · Payments Platform"
re-parse: { title: "Software Engineer II", company: "Chicago, IL", team: "Payments Platform" }
```

The **location is written into `company`, and `location` is lost entirely** — the same corruption class `#466` exists to fix, left live for the location-present case. The comment above the branch asserts the opposite ("No location edit path is affected"), which is what makes this worth blocking on: the claim reads as verified when it isn't. It is also *newly reachable* precisely because the `company === title` backstop makes empty-company roles more common than before.

Round-trip fidelity is a tested invariant here (`corpus-roundtrip.test.ts`), so this wants a regression test. Heads-up that the obvious fix does not work — I tried it:

```
emitted:  "Software Engineer II, Payments Platform · Chicago, IL"
re-parse: { title: "Software Engineer II, Payments Platform", company: "Chicago, IL" }
```

So this needs a shape the parser actually round-trips, not a one-line string tweak.

## Secondary

**`looksLikeBodyParagraph`'s length rule has no CSV exemption** — `src/lib/heuristics/entry-blocks.ts:197`. The verb-led branch guards with `!t.includes(",")`; the `t.length > 60` branch does not. `"React, TypeScript, Tailwind CSS, Node.js, PostgreSQL, Redis, Docker"` is 67 chars, so a realistic tech-stack subtitle is classified as body prose and routed out of `headerLines` into `description` — directly contradicting this PR's own test, which asserts the *short* stack `"React, TypeScript, Vite"` (23 chars) belongs in `headerLines`. Same exemption should apply to the length branch.

**Prose-body project descriptions are visible but not editable** — `src/components/features/ReconstructedResume.tsx:631`. The bulleted branch renders editable `ResumeBulletRow`s; the new prose branch renders a read-only `<p>` with no `EditableField` / override wiring. Strict improvement over "No bullet-shaped lines detected", but it lands an inconsistent edit surface in the lane whose premise is fix-in-place. Worth a follow-up issue if not fixed here.

**The compound router's right-side path is untested** — `src/lib/heuristics/regex.test.ts:430`. `matchSectionHeader("Personal Achievements")` is asserted to be `achievements`, but the compound regex requires a literal `" & "` or `" and "` — `"Personal Achievements"` has neither, so the compound block never runs and the assertion passes via the *pre-existing* anchor-fallback tier. The test is green while covering nothing new. A real right-side case is `"Personal and Achievements"` — and note it's the right-side path that makes finding #1 exploitable.

## Nits

Non-blocking.

- `src/lib/heuristics/entry-blocks.ts:199` — the verb-lead alternation lists `Built` twice; the trailing one is unreachable.
- `src/lib/heuristics/extract/experience-disambiguate.ts:479` — `COUNTRY_ONLY_RE` is declared inside `stripLocationSuffix`, so it's re-created per call; its siblings (`LOCALITY_SUFFIX_RE`, `COMPANY_TAIL_TOKENS_RE`) are module-scope. Hoist for consistency.

## Gates run

- **Fixture PII — PASS.** Verified the binaries, not the prose: all five new PDFs use synthetic personas, `@example.com` addresses, and `(312) 555-01{39,57,64,82,93}` — real area code, `555` exchange, valid `0100`–`0199` subscriber. `pdfinfo` shows only `pdf-lib` as Creator/Producer, no author metadata.
- **Design-system / reuse — PASS.** No raw `<button>`/modal/dropzone in feature code; the new `<p>` uses semantic tokens.
- **Style tokens — PASS.** `npm run lint` clean; no hex, raw palette classes, or manual `dark:` variants.
- **Dead code (fallow) — PASS (not a blocker).** The 5 new `scripts/fixtures/gen-*.mjs` are reported as unused files, but that matches the 4 pre-existing generators, and `verify` runs fallow report-only (`|| echo '… ignored'`).
- **Tests / typecheck — PASS.** `tsc -b --noEmit` clean; full suite 2286 passed / 9 skipped. Worth flagging that the suite being green is what lets all three Blocking findings through — none of them has a test.
- **Command-level bugs (skill/script files)** — skipped, no skill or script logic in this diff beyond the fixture generators.

Verdict rule applied: **≥1 Blocking → REQUEST_CHANGES** (3 Blocking, 3 Secondary, 2 Nits).

---

src/lib/heuristics/regex.ts
```diff
@@ -497,6 +497,29 @@ export function matchSectionHeaderDetailed(
...
+  // `sections.config.json` doesn't scale (`achievements` already carries four
+  // `awards & honors` variants, and any new pairing needs a manual add); split
+  // on the connective and try each side against the exact-alias table. Route to
+  // whichever side hits, preferring the LEFT (the header's primary category —
+  // `CERTIFICATIONS & ACTIVITIES` is a certifications section that also carries
+  // activities). Runs BEFORE the anchor-fallback tier so a compound header is
+  // recognized as an L1 exact-alias match rather than a softer L2 anchor match
+  // — the tier flag matters for the `isInstitutionRepeat` suppression logic in
+  // sections.ts.
+  const compound = /^(.+?)\s+(?:&|and)\s+(.+)$/.exec(normalized);
```
s-annam: **Blocking — bullet guard missing.** This tier is the only one of the three adjacent tiers *not* wrapped in `if (!LEADING_BULLET_RE.test(text))`. Normalization strips only **trailing** `[:·•]`, so a leading `•` survives into `normalized`: the left side then fails the alias table (it carries the bullet) and the **right** side hits, because `leadership` / `volunteer` / `activities` / `honors` / `interests` / `projects` are all single-word aliases.

Verified on this branch:
```
"• Public speaking and leadership"  => experience
"• Committee member and volunteer"  => experience
"• Dean's List and Honors"          => achievements
"• Mentored juniors and interests"  => other
```
`classifyLine` (`sections.ts:1872`) runs this on **every** line with no font/bold gate and returns `{kind: "open"}`. The tier reports `viaAnchorFallback: false` (L1), so `isInstitutionRepeat` doesn't suppress it — and the bullet opens a *different* section than the one it sits in. One ordinary skills/achievements bullet splits the résumé mid-body.

Fix: move this block inside an `if (!LEADING_BULLET_RE.test(text))` guard, same as the tiers above and below.

src/lib/heuristics/extract/experience-disambiguate.ts
```diff
@@ -533,15 +576,86 @@ function mapTitleFirst(splits: Split[], anchorIdx: number | undefined): Fields {
...
+  // COMPANY_SUFFIX_RE (broadening that regex breaks `looksLikeTitle`'s early
+  // return for `Software Engineer` / `Digital Product Manager`), so use
+  // COMPANY_TAIL_TOKENS_RE — the same closed vocabulary that gates Pass D's
+  // location deferral — here as a company-recognition signal. Distinct from
+  // case 3 (below): case 3 stays for genuinely suffix-less bare Title, Team
+  // banner continuations ("Senior Engineer, Payments Core") where the
+  // post-comma is a team, not a company.
+  if (anchorIsCommaSplit && splits[1]?.text) {
+    const lastToken = splits[1].text.trim().split(/\s+/).pop() ?? "";
+    if (COMPANY_TAIL_TOKENS_RE.test(lastToken)) {
```
s-annam: **Blocking — this reuses a *deferral* vocabulary to make a *promotion* decision.**

In Pass D a `COMPANY_TAIL_TOKENS_RE` match means **defer**, and a false positive is harmless — the comment on the constant says so: "a false negative leaves the string as company (recoverable via inline edit)". Here a match means **promote to company and set `team: undefined`**. A false positive is now destructive, and the vocabulary is far too broad for that: `Systems`, `Analytics`, `Health`, `Digital`, `Financial`, `Media`, `Labs`, `Solutions`, `Networks`, `Group` are all common *team*-name tails — `Core Systems`, `Consumer Health`, `Growth Analytics`, `Payments Digital`.

Verified as a regression vs `origin/main`, same input (a `Wingtip Financial Inc.` banner over bare `Title, Team` roles):
```
main:   { title: "Senior Engineer", company: "Senior Engineer",   team: "Growth Analytics" }
branch: { title: "Senior Engineer", company: "Growth Analytics", team: undefined }
```
The team is gone. And because `isBannerContinuation` (`extract/experience.ts:113`) early-returns `false` when `!entry.team`, the #382 shared-employer banner is **no longer inherited either** — the role loses its team, gains a wrong company, and never recovers the real employer. Three fields worse than `main`, which at least kept the team.

The two call sites need different vocabularies. Restrict this one to unambiguous legal/entity markers (`Inc`, `LLC`, `Ltd`, `GmbH`, `PLC`, `Corp`, `Corporation`, `Holdings`) and drop the generic nouns; Pass D's deferral can stay broad.

src/lib/pdf/ats-resume-model.ts
```diff
@@ -536,7 +536,23 @@ export function buildAtsResumeModel(
...
+    // `Title · Company` shape and mis-labels the team as the company. Emit the
+    // team after a COMMA instead ("Title, Team"), so the parser's role-comma
+    // split correctly routes it back into `team` (case 3 in `mapTitleFirst`)
+    // and the `company === title` backstop clears the mirrored company on
+    // re-parse — round-tripping to the SAME empty-company entry we emitted
+    // from. No location edit path is affected (an empty-company role with a
+    // location can still emit `Title · Location` and re-parse via the
+    // existing bare-location-title-rescue in step 5 of `cleanFieldArtifacts`).
+    let headerText: string;
+    if (!exp.company?.trim() && exp.team?.trim() && !exp.location?.trim()) {
```
s-annam: **Blocking — the `!exp.location?.trim()` condition leaves the corruption live for the location case.**

An empty-company role that *has* a location fails this condition and falls through to the middot join. Verified by re-parsing what gets emitted:
```
emitted:  "Software Engineer II · Chicago, IL · Payments Platform"
re-parse: { title: "Software Engineer II", company: "Chicago, IL", team: "Payments Platform" }
```
The **location is written into `company`, and `location` is lost entirely** — the same corruption class #466 exists to fix. The comment directly above asserts the opposite ("No location edit path is affected"), which is what makes this worth blocking on: the claim reads as verified when it isn't. It's also *newly reachable* precisely because the `company === title` backstop makes empty-company roles more common than before.

Round-trip fidelity is a tested invariant (`corpus-roundtrip.test.ts`), so this wants a regression test. Heads-up that the obvious fix doesn't work either — I tried it:
```
emitted:  "Software Engineer II, Payments Platform · Chicago, IL"
re-parse: { title: "Software Engineer II, Payments Platform", company: "Chicago, IL" }
```
So this needs a shape the parser actually round-trips, not a string tweak.

src/lib/heuristics/entry-blocks.ts
```diff
@@ -456,6 +477,47 @@ function collectDatelessAnchors(lines: PdfLine[], dateAnchors: number[]): number
...
+ * limited to `first_line` so the `date_range` (experience) and `institution`
+ * (education) anchor paths are unaffected — those don't need this predicate
+ * (they anchor on structured signals) and applying it there would risk
+ * regressing existing multi-line header shapes.
+ */
+function looksLikeBodyParagraph(text: string): boolean {
+  const t = text.trim();
+  if (!t) return false;
+  if (/[.!?]$/.test(t)) return true;
+  if (t.length > 60) return true;
```
s-annam: **Secondary — the length rule has no CSV exemption, but the verb-led rule does.**

Line 517 guards the verb-led branch with `!t.includes(",")`; this branch has no such guard. `"React, TypeScript, Tailwind CSS, Node.js, PostgreSQL, Redis, Docker"` is **67 characters**, so a realistic tech-stack subtitle returns `true` here, and `buildEntryBlock` (`first_line` anchor) routes it out of `headerLines` into `description`.

That directly contradicts this PR's own test, which asserts the *short* stack `"React, TypeScript, Vite"` (23 chars) belongs in `headerLines` — same shape, opposite outcome, decided purely by length. Suggest carrying the same `!t.includes(",")` exemption onto the length branch.

src/lib/heuristics/entry-blocks.ts
```diff
@@ -456,6 +477,47 @@ function collectDatelessAnchors(lines: PdfLine[], dateAnchors: number[]): number
...
+  const verbLed =
+    /^(?:Built|Designed|Implemented|Developed|Led|Architected|Optimi[sz]ed|Refactored|Deployed|Prototyped|Migrated|Automated|Delivered|Created|Engineered|Shipped|Wrote|Authored|Reduced|Improved|Increased|Launched|Ran|Built)\b/i.test(
```
s-annam: **Nit** — `Built` appears twice in this alternation (first and last). The trailing one is unreachable; drop it.

src/lib/heuristics/extract/experience-disambiguate.ts
```diff
@@ -222,12 +236,41 @@ function stripLocationSuffix(s: string): {
...
+    // residual shape a corporate-tail deferral (Pass D) leaves behind. Example:
+    // "Deutsche Bank, India" — Pass D defers on "Bank" as a corporate tail
+    // (so the last company word isn't stolen), Pass E now peels the bare
+    // ", India" tail so company becomes "Deutsche Bank" and location surfaces
+    // as "India". Multi-word countries ("United Kingdom", "New Zealand") are
+    // admitted only through the closed COUNTRY_GAZETTEER — no open-shape
+    // regex — so a `Buyer, Home Goods` job-title tail (Home Goods not a
+    // country) is safe. Tried LAST so the richer city+country shapes above
+    // still win when they apply.
+    const COUNTRY_ONLY_RE =
```
s-annam: **Nit** — `COUNTRY_ONLY_RE` is declared inside `stripLocationSuffix`, so it's re-created on every call. Its siblings (`LOCALITY_SUFFIX_RE`, `COMPANY_TAIL_TOKENS_RE`) are module-scope; hoist for consistency.

src/components/features/ReconstructedResume.tsx
```diff
@@ -623,6 +623,15 @@ function ProjectsSection({
                     />
                   ))}
                 </ul>
+              ) : !added && project.description ? (
+                // #464 — a prose-body project (no `•` bullets, description is
+                // one or more paragraph sentences) surfaces the description
+                // as a paragraph. Without this, the fallback below reads "No
+                // bullet-shaped lines detected" even though the parser DID
+                // extract the description via looksLikeBodyParagraph.
+                <p className="text-sm text-content-primary whitespace-pre-wrap">
```
s-annam, line 632: **Secondary — visible, but not editable.**

The bulleted branch above renders `ResumeBulletRow` (editable, override-wired); this prose branch renders a read-only `<p>`. So a prose-body project's description now *shows* but has no input path, while the same project with bullets is fully editable.

Strict improvement over "No bullet-shaped lines detected", so not blocking — but it lands an inconsistent edit surface in the lane whose whole premise is fix-in-place. Worth a follow-up issue if not addressed here.

src/lib/heuristics/regex.test.ts
```diff
@@ -411,3 +411,43 @@ describe("matchSectionHeader — leading decorative glyph (#414)", () => {
...
+  it("routes 'X and Y' word-form to the left side's section", () => {
+    expect(matchSectionHeader("Certifications and Activities")).toBe(
+      "certifications",
+    );
+  });
+
+  it("falls through to the right side when the left side is not an alias", () => {
+    // "Personal" is not an alias; "achievements" is — the compound router
+    // takes the RIGHT side once the left fails.
+    expect(matchSectionHeader("Personal Achievements")?.toString() ?? null).toBe(
```
s-annam: **Secondary — this test doesn't exercise the compound router.**

The compound regex requires a literal `" & "` or `" and "`. `"Personal Achievements"` has neither, so the compound block never runs — this assertion passes via the *pre-existing* head-noun anchor-fallback tier. The test is green while covering nothing new.

A real right-side case would be `"Personal and Achievements"`. Worth fixing because the right-side path is exactly what makes the bullet bug in `regex.ts` reachable — it's the untested half.

s-annam (commented):
Adding the eight in-diff findings as inline anchors on the lines they refer to — same findings as my review above, no new ones. The two supporting citations that anchor to *unchanged* code (`sections.ts:1872` for the `classifyLine` call path, `extract/experience.ts:113` for `isBannerContinuation`) stay in the review body, since they're the mechanism behind the blockers rather than findings in their own right.

Verdict is unchanged: **CHANGES_REQUESTED** (3 Blocking, 3 Secondary, 2 Nits).

---

src/lib/heuristics/regex.ts
Vaishnavi1709: Fixed in 19df83c. Wrapped the compound tier in `if (!LEADING_BULLET_RE.test(text))` mirroring the sibling tiers. Added a regression test [`regex.test.ts`](src/lib/heuristics/regex.test.ts) with your four repro strings — all now return `null`.

src/lib/heuristics/extract/experience-disambiguate.ts
Vaishnavi1709: Fixed in 19df83c. Introduced a new closed vocab `COMPANY_LEGAL_TAIL_RE` (`Inc|LLC|L.L.C|Ltd|GmbH|PLC|Corp|Corporation|Holdings`) scoped to this promotion decision, and left the broader `COMPANY_TAIL_TOKENS_RE` for Pass D's deferral. Added two regression tests in [`experience.title-team-comma.test.ts`](src/lib/heuristics/extract/experience.title-team-comma.test.ts): (a) case 3a still promotes on a legal-marker tail (`Ridgemont Holdings`), and (b) your exact `Wingtip Financial Inc.` banner + `Senior Engineer, Growth Analytics` shape — role 2 keeps its team AND inherits the banner as `company` via #382, no fields lost.

src/lib/pdf/ats-resume-model.ts
Vaishnavi1709: Fixed in 19df83c. The empty-company branch now emits `Title, Team` on the header line and routes the location onto a separate `subLine` (`City, ST` on its own row below). Parser `recoverLocation` step 3c extended to recover a below-anchor `via: "whole"` cell that reads as a bare location. New repro test [`render-roundtrip-empty-company-location.repro.test.ts`](src/lib/pdf/render-roundtrip-empty-company-location.repro.test.ts) pins the invariant: parse1 (`{title, company:"", location:"Chicago, IL", team}`) → export → parse3 returns the same fields, with location NOT in the company slot. Removed the misleading "No location edit path is affected" comment.

src/lib/heuristics/entry-blocks.ts
Vaishnavi1709: Fixed in 19df83c. Carried the same `!t.includes(",")` exemption onto the length branch. `"React, TypeScript, Tailwind CSS, Node.js, PostgreSQL, Redis, Docker"` (67 chars, CSV) now returns `false` — subtitle stays in headerLines. The existing `entry-blocks.test.ts` short-stack test still passes.

src/lib/heuristics/entry-blocks.ts
Vaishnavi1709: Fixed in 19df83c. Dropped the trailing `Built` from the alternation.

src/lib/heuristics/extract/experience-disambiguate.ts
Vaishnavi1709: Fixed in 19df83c. Hoisted `COUNTRY_ONLY_RE` to module scope alongside its siblings, with a doc comment explaining why the shape-only regex is safe (the closed `COUNTRY_GAZETTEER` check happens at the call site).

src/components/features/ReconstructedResume.tsx
Vaishnavi1709: Deferred to #489. Agree the inconsistent edit surface is worth fixing — it lands cleanly outside this PR's scope (needs edit-round-trip wiring through `bulletOverrides`/`descriptionOverrides`, plus a fresh test that survives Download PDF). Leaving this thread open for you to close.

src/lib/heuristics/regex.test.ts
Vaishnavi1709: Fixed in 19df83c. Changed the assertion to `"Personal and Achievements"` (plus `"Personal & Achievements"` as a paired case) so the compound tier's right-side path is actually exercised — the pre-existing anchor-fallback no longer matches a bare-prose `"Personal Achievements"` in the new test.

src/lib/heuristics/extract/education.ts
```diff
@@ -878,8 +942,20 @@ export function extractEducation(
...
-    const isInst = INSTITUTION_HINTS.test(text);
+    // #462/#467 — a DEGREE_RE or INSTITUTION_HINTS hit on a sub-labelled line
+    // ("Achievements: Graduated B.E. with Distinction"), an event-narration
+    // sentence ("Graduated B.E. …"), or a body sentence that contains an
+    // incidental institution-word substring (e.g. `\bcollege\b` inside
+    // "inter-college hackathons") is body prose that bled in from a mis-routed
+    // compound header ("CERTIFICATIONS & ACTIVITIES") or an unrouted qualified
+    // header ("RELEVANT COURSEWORK"), not a new education entry. Gate BOTH
+    // isDeg and isInst on `isRealEntryHeader` so such lines never open a
+    // phantom entry — the raw DEGREE_RE hit still shows up in
```
s-annam, line 953: Nit (non-blocking): this comment says the next-line lookahead "uses raw `DEGREE_RE.test`", but this PR gated that lookahead with `&& isRealEntryHeader(next)` (line ~975). So a genuine follow-on degree survives because it *passes* `isRealEntryHeader`, not because the check stays raw. The stated reasoning is now inaccurate — worth updating the comment.

src/lib/heuristics/extract/experience-disambiguate.ts
```diff
@@ -533,15 +597,92 @@ function mapTitleFirst(splits: Split[], anchorIdx: number | undefined): Fields {
...
+    if (COMPANY_LEGAL_TAIL_RE.test(lastToken)) {
+      return { company: splits[1].text, title, team: undefined };
+    }
+  }
+  // Case 3 (#382) — anchor is Title, Team on the same line as its date, with
+  // NO below-anchor employer line to draw from: a bare shared-banner
+  // continuation ("Senior Engineer, Payments Core Jul 2022 - Aug 2024"). The
+  // post-comma segment is a team/sub-org, not the employer. Keep it as `team`
+  // so `propagateSharedEmployer` (extract/experience.ts) can inherit the
+  // banner into `company`; mirror the title into `company` so
```
s-annam, line 676: Nit (non-blocking): this says case 3 mirrors title into `company` "so `isBannerContinuation`'s `company === title` branch fires", but the new end-of-function backstop in `disambiguateCompanyTitle` clears `company === title` to `undefined` before the entry reaches `propagateSharedEmployer`. Banner continuation actually fires via the `company === ""` branch (`experience.ts:117`). Functionally correct, but the comment misdescribes the live path and the `company === entry.title` disjunct at `experience.ts:117` is now effectively dead — worth a cleanup.

s-annam (approved):
Maintainer review — **APPROVE**. Unusually careful diff: every new branch is closed-vocabulary, guarded, and tested. All resumelint gates pass — fixture PII (5 new fixtures fully synthetic), design-system reuse, style tokens, typecheck, lint, 234 changed-file tests, fallow (no new exports). No correctness or PII issues.

Two documentation nits below, both inline, both non-blocking: a comment describes a mechanism this diff itself changed. Worth a one-line cleanup on the next touch of these files, not a merge blocker.

---

s-annam:
@Vaishnavi1709 — approved, safe to merge. The two inline nits are **non-blocking**: both are just stale comments describing a mechanism this PR changed (the `education.ts` lookahead reasoning, and the case-3 banner-continuation comment in `experience-disambiguate.ts`). No code change required to merge — clean up the wording next time you touch those files if you like.

Issue #391, opened 2026-07-07
Batch: edit-interaction display & persistence nits


s-annam:
## Batch epic — Edit-interaction display & persistence nits

Groups the small issues below for **single-batch implementation** by one person
via `/implement-batch`. This is a tracking issue; the work lives in the children.

**Why grouped:** all touch the reconstructed-resume edit/read-mode surface (EditableField, +Add affordance, absent-field rendering) — same subsystem, same reviewer context, each a small display/interaction fix.

### Children
- [ ] #376 — absent fields render literal placeholder words ("start", "location not detected") inline in read mode
- [ ] #378 — contact — "LinkedIn not detected" shown next to a live linkedin-shaped URL, self-contradictory
- [ ] #379 — "+ Add education" creates a ghost entry when left completely blank on commit
- [ ] #380 — display polish nits — empty sections still render "+Add"; middot between month/year; awards separator differs from source
- [ ] #381 — inline edits persist across re-drops of same PDF with no visual signal distinguishing user edits from parser output

**Implement with:** `/implement-batch <THIS#>`

s-annam:
All five children are closed — this tracking issue has nothing left to track.

- [x] #376 — absent fields render literal placeholder words in read mode
- [x] #378 — "LinkedIn not detected" next to a live LinkedIn-shaped URL (`9fe39b9`, PR #496)
- [x] #379 — blank "+ Add education" ghost entry (`587259d`, PR #497) — verified live on `main` @ `1b391e6` today; the blank entry is pruned on section-focus exit
- [x] #380 — display polish nits (empty-section "+Add", middot, awards separator)
- [x] #381 — closed as not-reproducible: edits do not survive a re-drop, and no edit state for an uploaded résumé is persisted at all. `useAnalyzedResume` resets on every fresh parse and there is no content-hash parse cache; that reset landed 2026-07-06, one day before the issue was filed, so the QC session hit a pre-fix build. Repro transcript is on #381.

Closing. Parent epic #511 stays open — #527 is still outstanding there.

Issue #380, opened 2026-07-07
[edit-interaction] display polish nits — empty sections still render "+Add"; project uses middot between month/year; awards separator differs from source


Vaishnavi1709:
## Category
edit-interaction (Steps 1 & 3 F&F QC — provenance category). Mixed: Nits 1 and 3 are render-template issues; Nit 2 is a parser defect (see Category rationale below).

## Fixtures
- `tests/fixtures/pdfs/latex/multi-degree-coursework.pdf` (empty Achievements section header still renders; project date shape)
- `tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-two-column.pdf` (Achievements separator — heading reads AWARDS on this fixture)
- `tests/fixtures/pdfs/unknown/synthetic-degreeless-two-programs.pdf` (empty Projects + Achievements)

### Premise correction (this nit was reframed after review)
The original framing called this a defect to fix by hiding the section. It is not a defect — it is a deliberate, documented design choice:

- `src/components/features/ReconstructedResume.tsx:1057-1059` carries an explicit comment: the Achievements section is "Always rendered (even with zero parsed achievements) so the '+ Add achievement' affordance is reachable on every resume — matching Education / Skills, which also render an add affordance unconditionally."
- `src/components/features/ReconstructedResume.tsx:815` is the `AddPill label="Add achievement"` — the only input path for achievements data when the source has none.

**The original "Option 1: hide the section entirely when empty AND absent from source" is REJECTED**, for the same reason sibling issue #376 rejected the equivalent option for placeholder fields: hiding the only entry point strands the user with no way to add an achievement to a resume that didn't have one, silently regressing an existing capability.

### What's actually left to fix
Narrower than the original framing: an empty Achievements section, at a glance, can read as "the parser dropped my achievements" rather than "here's a section you can add to." This is an affordance-strength gap, not a broken or over-eager render — same class of issue as #376's placeholder-vs-value ambiguity, just at the section level instead of the field level.

### Constraint (pattern selection deferred to implementation)
This issue does not prescribe a visual treatment. At implementation time, apply the `ui-ux-pro-max` skill to choose the pattern, subject to one hard constraint: **the "+ Add achievement" pill must remain the reachable, functioning input path** on a zero-achievement resume — nothing here proposes removing or hiding it.

### Premise correction — wrong root cause, wrong category
[...]

PR #484, opened 2026-07-14
fix(edit): placeholder add-affordance, empty-section hint, lone-date month capture (#376, #380)

s-annam:
Two edit-interaction issues from the Friends & Family QC sweep, batched onto one branch. Part of epic #391 (3 of its 5 children remain open, so this `Refs` rather than closes it).

## #376 — empty placeholders read as values

An empty field's placeholder was indistinguishable from real data once a separator joined it into a compound line: `Acme Corp, location not detected`, `start – 2013`. Muted italic isn't enough when the joiner doesn't change.

`EditableField` now renders an empty field as a `+` affordance glyph — its own `aria-hidden` span, not concatenated into the text — followed by a bare noun naming the thing to add, and announces itself as **"Add \<label\>"** instead of "Edit \<label\>". That last part closes a pre-existing WCAG 2.5.3 (Label in Name) failure: the explicit `aria-label` on a `role="button"` overrides its text content, so the field was announced "Edit Location" while visibly reading "location not detected". The accessible name now contains the visible text, and the `+` is never announced.

The placeholder remains **the only input path** — `role`, `tabIndex`, `onClick` and the Enter/Space handler are untouched. That invariant is why the issue's "hide the field when empty" option was rejected: hiding it strands the user with no way to add a missing start date.

Two things worth calling out:

- **The default placeholder is derived from `label`.** That's what makes the old `+ not detected` copy *unrepresentable* rather than merely fixed — a call site that forgets a placeholder now gets a correct one instead of the broken default.
- **Fields whose empty state is a *state*, not a gap** (`Untitled resume`, `edit this bullet`, `empty bullet`) opt out with `emptyAffordance="plain"`. A blanket prefix in the primitive was the wrong seam — it produced `+ Untitled resume`.

## #380 — three display/parse nits

**Empty Achievements section.** Kept rendering (by design — the `+ Add achievement` pill is the only input path on a résumé with none), but it could read as *"the parser dropped my achievements"*. It now carries a `SectionEmptyHint` naming what belongs there. The copy deliberately makes **no claim about what the parser found**: we cannot distinguish "the résumé had none" from "we missed them", so finding-shaped wording ("no achievements detected") would either be a guess or reinforce the exact misreading we're fixing.

**Lone dates lost their month (parser).** `parseDateRange` decayed a lone `Mon. YYYY` to a bare year, and because the strip only removes what a date regex matched, the month stayed welded to the title — `tinylm | Link Jan.` · `2026`. It now captures the token whole. The strip runs month-year *before* bare-year (order is load-bearing — strip the year first and the month is stranded), and both halves use the **same** regex, so the token the date captures is exactly the token the title loses.

That regex is new, and it's the interesting part. `MONTH`'s `[a-z]*` tail also matches **Marketing, Marathon, Mayor, Junior, Decathlon**. That was harmless while `MONTH_YEAR_RE` was only a *position* hint — the moment this change promoted it to a date **value** and a text **deletion**, it started eating words out of titles:

```
"Head of Marketing 2020"  →  start_date: "Marketing 2020",  title: "Head of"
"Boston Marathon 2021"    →  start_date: "Marathon 2021",   title: "Boston"
```

`STRICT_MONTH_YEAR_RE` enumerates the month names instead. `dateRegionStart` keeps the loose form — it only ever needs a position, never a value.

**Achievements separator.** `AchievementHeader` hardcoded a middot between title and year, rewriting the source's own comma (`Globex Engineering Excellence, 2021` → `… · 2021`). The separator is destroyed by the date strip, so it's now read off the raw line at parse and carried on `HeuristicAchievement` as an **additive optional `year_separator`**. Both consumers — the on-screen header and the PDF exporter — re-emit it through one shared joiner, so screen and PDF cannot disagree and the comma survives the parse→export→re-parse round trip.

`·` is *reported* as a separator but deliberately **never trimmed** from the title: a trailing middot is the org-signature marker the anchor-position tiebreak (#298) keys on, and stripping it swaps a role's title and company. It can't double, because `liftHeaderLabel` clears the dangling glyph first. The two character sets are asymmetric on purpose, and the code says so — that asymmetry got "simplified" away once during review and broke role attribution.

### Contract change (approved)

`HeuristicAchievement.year_separator?: string` is new on the parser's public output. Strictly additive and optional; no existing consumer breaks. It's the only way to satisfy #380's "renders using the source's original separator" criterion, since the header is stored decomposed (type / title / year) and every consumer must therefore emit *some* separator.

### Corpus snapshot

`awesome-cv-cv.expected.json` moves by 5 characters (`extractedCharCount` 12859 → 12854): one leaked month token, `"Freelance Penetration Tester Sep."` → `"Freelance Penetration Tester"`. Re-baked via `npm run bake-fixtures`, not hand-edited. It is the only field in that fixture that changed — diffed against `main`'s parser to confirm the cause rather than assume it.

## Verification

`npm run verify` green: typecheck, lint, **2310 tests** (174 files, corpus + round-trip suites included), build.

## Reviewer notes

- `ReconstructedResume.tsx` grew to ~1211 LOC against the known-debt rule in `CLAUDE.md`. `AchievementHeader` is the natural extraction; `SectionEmptyHint` already went out to `ReconstructedAdd.tsx`.
- The type↔title middot at `ReconstructedResume.tsx:677` is **not** changed, though #380 named it alongside the title↔year one. It's defensible: `splitAchievementType` synthesizes that boundary rather than reading it from the source, so there is no original separator to preserve. Flagging it rather than leaving it silent.
- Follow-up, not fixed here: a lone **numeric** date (`01/2026`) still parses to a bare year and leaves `01/` in the title — `NUMERIC_MONTH_YEAR_RE` isn't in the lone-date fallback. Pre-existing, same class as #380's month bug. Will file separately.

## Provenance

Every model string below is self-reported by the agent that produced the work. Where an agent did not report, the row says so rather than guessing from the alias that was requested.

| Stage | Model | Effort |
|---|---|---|
| Implementation — #376 | Claude Sonnet 5 | high |
| Implementation — #380 | Claude Opus 4.8 (1M context) | ultra |
| Adversarial review — round 1 | Claude Opus 4.8 | — |
| Review fixes — round 1 | `unreported (requested: opus)` | — |
| Adversarial review — round 2 | Claude Opus 4.8 | — |
| Orchestration, parser repair, PR | Claude Opus 4.8 (1M context) | — |

## Adversarial review

Two rounds, pre-PR, on the local diff. Both rounds returned `clean: false` — the findings below were fixed before this PR existed, and they are recorded rather than quietly dropped.

**The headline: the tree was fully green when round 1 found a data-corruption bug.** typecheck clean, lint clean, 2273 tests passing, fallow showing no new findings — and `"Head of Marketing 2020"` parsed to `start_date: "Marketing 2020"` with the title truncated to `"Head of"`. A green suite proves nothing broke that we thought to test. It was found by reproducing against the pre-change parser, not by reading the diff.

### Round 1 — 3 blocking

| # | Finding | Resolution |
|---|---|---|
| B1 | `MONTH`'s `[a-z]*` tail matches **Marketing, Marathon, Mayor, Junior, Decathlon**. This change promoted `MONTH_YEAR_RE` from a position hint to a date *value* and a global text *deletion*, so a false month ate a word out of the title and landed in `start_date`. Reaches every entry type. | `STRICT_MONTH_YEAR_RE` (enumerated months, no prefix tail), used in both the lone-date fallback and the strip, as the same regex. `dateRegionStart` keeps the loose form. |
| B1b | Same root cause: a header that is *entirely* a false month + year (`"Marketing 2021"`) stripped to `""`, stopped registering as a new anchor, and silently **merged into the previous entry**. Invisible to the corpus snapshots, which are keys/counts-only. | Fixed by B1. Verified: `"Marketing 2021"` opens a new entry again. |
| B2 | `;` and `:` were *reported* by `dateSeparator` but never *trimmed* from the title, so the consumer re-emitted them: `"Tech Lead: 2020"` → `"Tech Lead:: 2020"` → `"Tech Lead::: 2020"`. **Unbounded** — one glyph per Download-PDF cycle. `,` `|` `–` `—` `·` round-trip cleanly, which is exactly why the new tests missed it. | Trim set gains `;:`. The first attempt unified the report and trim sets, which silently added `·` to the trim and **broke role attribution** — the trailing middot is the org-signature marker the anchor tiebreak keys on. The sets are asymmetric on purpose now, and the code says why. |
| B3 | The `+` prefix was concatenated into the span's text, conflating the affordance glyph, the placeholder noun, and the accessible name — producing `+ location not detected`, `+ Untitled resume`, `+ edit this bullet`. | `emptyAffordance` prop, `+` as its own `aria-hidden` span, noun-phrase placeholders, default derived from `label`, `aria-label` verb switches to `Add`. |

### Round 2 — 1 blocking

**Two call sites were never visited** (`ContactExtraLinks.tsx`, `AchievementTypePicker.tsx`). Because `emptyAffordance` defaults to `"add"`, they *inherited* the new `+` glyph on top of their old sentence-case copy — `+ Link URL`, `+ Custom label`. The default that makes the affordance repo-wide is the same default that propagates the defect to anything overlooked. Before the change these read as inert labels; the change made them incoherent.

Fixed by deleting their stale placeholders so the label-derived default applies. #376's completeness criterion ("no call site is special-cased or skipped") is a property of the **repo**, not of the primitive, so it is now guarded by a source sweep over all 23 call sites — falsified by reintroducing the defect and confirming the guard fails.

### Verified by execution, not by reading

Round 2 ran, against the real code: 12 false-month titles; every separator glyph × 3 entry shapes × 3 round-trip cycles (24 scenarios, byte-stable from cycle 1); the anchor-tiebreak shape; and all 23 `EditableField` empty-state renders. It also found the *mechanism* that makes the `·` asymmetry correct rather than lucky — `liftHeaderLabel` (`extract/projects.ts:106`) clears the dangling middot — and confirmed the `awesome-cv-cv` snapshot delta is one leaked month token and nothing else, by diffing the full parse against `main`.

**Not covered:** the four fixture PDFs were never driven through a live browser. UI verification is `renderToStaticMarkup` against the real components, so #376's fixture-render criteria are unit-level only.

### Surviving nits (not iterated)

- `dateSeparator` locates the date with the *loose* position scan while the value comes from the *strict* parse, so a false-month word between the separator and the year misattributes it: `"Award, Marketing 2020"` exports as `"Award, Marketing, 2020"`. Converges at cycle 1 — a one-time fidelity wart, **not** the B2 doubling class.
- Education date placeholders read `+ start` / `+ end`; #376's own illustration used `+ start date`, which `ReconstructedRole` already uses.
- `AchievementHeader` has 0% coverage on its render. The load-bearing leg of `year_separator` is the **exporter** (that's what the round-trip re-parses) and it has direct tests; what's untested is a `-ml-1.5` spacing class, where a wrong result is cosmetic and instantly visible.

### DIFF
--- a/src/components/features/AchievementTypePicker.tsx
+++ b/src/components/features/AchievementTypePicker.tsx
@@ -121,7 +121,6 @@ export function AchievementTypePicker({
                 PDF actually said. */}
             <EditableField
               value={label || undefined}
-              placeholder="Custom label"
               label="Custom achievement type"
               textSize="xs"
               onCommit={(v) => pick(v)}
--- a/src/components/features/ContactCard.test.tsx
+++ b/src/components/features/ContactCard.test.tsx
@@ -223,8 +223,14 @@ describe("ContactCard", () => {
       makeResult({ email: "jane@example.com" }, { email: 0.95 }),
       { overrides: {}, onFieldChange: () => {} },
     );
-    expect(el.textContent).toContain("Phone not detected");
-    expect(el.querySelector('[aria-label="Edit Phone"]')).not.toBeNull();
+    // An absent required field reads as a gap to FILL — "+ phone", announced
+    // "Add Phone" — not as a status sentence. It is still discernible, and still
+    // the only input path (role=button, focusable, opens an empty input).
+    expect(el.textContent).toContain("+ phone");
+    const field = el.querySelector('[aria-label="Add Phone"]');
+    expect(field).not.toBeNull();
+    expect(field?.getAttribute("role")).toBe("button");
+    expect(field?.getAttribute("tabindex")).toBe("0");
   });
 
   it("no longer renders the detected/total completeness footer (moved to the AttentionStrip)", () => {
--- a/src/components/features/ContactCard.tsx
+++ b/src/components/features/ContactCard.tsx
@@ -91,7 +91,7 @@ export function ContactCard({
         {editable ? (
           <EditableField
             value={name && !name.gated ? name.value : undefined}
-            placeholder="Name not detected"
+            placeholder="name"
             label="Name"
             textSize="lg"
             textWeight="semibold"
--- a/src/components/features/ContactDetails.tsx
+++ b/src/components/features/ContactDetails.tsx
@@ -192,7 +192,7 @@ function EditableValue({
     <EditableField
       value={field.value || undefined}
       displayValue={displayValue}
-      placeholder={`${field.label} not detected`}
+      placeholder={field.label.toLowerCase()}
       label={field.label}
       textSize="sm"
       validate={validatorFor(field.key, location)}
--- a/src/components/features/ContactExtraLinks.tsx
+++ b/src/components/features/ContactExtraLinks.tsx
@@ -50,7 +50,6 @@ export function ContactExtraLinks({
             <EditableField
               value={profile.url}
               displayValue={formatLinkDisplay(profile.url)}
-              placeholder="Link URL"
               label={profile.network}
               textSize="sm"
               onCommit={(v) => onEdit(profile.id, v)}
--- a/src/components/features/FindJobsPanel.tsx
+++ b/src/components/features/FindJobsPanel.tsx
@@ -153,7 +153,7 @@ export function FindJobsPanel({ parsed }: FindJobsPanelProps) {
           <span className="text-xs text-content-tertiary">Title</span>
           <EditableField
             value={query.title}
-            placeholder="title not detected"
+            placeholder="job title"
             label="Job title"
             textWeight="semibold"
             onCommit={(v) => setQuery((q) => ({ ...q, title: v }))}
@@ -163,7 +163,7 @@ export function FindJobsPanel({ parsed }: FindJobsPanelProps) {
           <span className="text-xs text-content-tertiary">Seniority</span>
           <EditableField
             value={query.seniority}
-            placeholder="not detected"
+            placeholder="seniority"
             label="Seniority"
             onCommit={(v) =>
               setQuery((q) => ({ ...q, seniority: v || undefined }))
--- a/src/components/features/ReconstructedAdd.tsx
+++ b/src/components/features/ReconstructedAdd.tsx
@@ -37,6 +37,25 @@ function CloseIcon() {
 }
 
 /**
+ * The one-line prompt an EMPTY section shows above its {@link AddPill} (#380).
+ *
+ * A section that renders with no entries is ambiguous in a way that matters
+ * here: it reads as "the parser found your achievements and dropped them" when
+ * what it means is "this section is empty — you can fill it." The pattern that
+ * resolves it is the standard guided empty state (a helpful message plus the
+ * action, never bare blank space), with one constraint the copy must honour: it
+ * describes what BELONGS in the section, and makes no claim about what the
+ * parser did or did not find. We cannot tell "the résumé had none" apart from
+ * "we missed them", so any finding-shaped wording ("no achievements detected")
+ * would either be a guess or reinforce the very reading we are fixing. The
+ * `AddPill` below it stays the reachable input path, untouched — the block-level
+ * "add a whole entry" affordance it already is.
+ */
+export function SectionEmptyHint({ children }: { children: React.ReactNode }) {
+  return <p className="text-sm text-content-tertiary">{children}</p>;
+}
+
+/**
  * The collapsed progressive-disclosure trigger — a chip-shaped "+ <label>" pill
  * that sits inline with the content it adds to. Quiet by default; warms to the
  * brand accent on hover.
--- a/src/components/features/ReconstructedEducationSkills.test.ts
+++ b/src/components/features/ReconstructedEducationSkills.test.ts
@@ -2,9 +2,12 @@
 // Copyright 2026 The resumelint Authors
 
 import { describe, it, expect } from "vitest";
+import { createElement } from "react";
+import { renderToStaticMarkup } from "react-dom/server";
 import {
   resolveEduValue,
   resolveEducationDisplay,
+  EducationSection,
 } from "./ReconstructedEducationSkills.tsx";
 import type { ResumeEducation } from "../../lib/score/types.ts";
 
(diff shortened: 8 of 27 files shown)


Samhit21 (commented):
## PR Review: placeholder add-affordance, empty-section hint, lone-date month capture (#376, #380)

### Summary
Exceptionally well-reviewed PR — two adversarial rounds + a candid list of surviving nits, and the data-corruption find (`STRICT_MONTH_YEAR_RE`) is the right fix for the right reason. I ran an independent pass over the load-bearing pieces rather than trusting the writeup, and they hold:

- **Strict-month regex `lastIndex` hygiene** — every `.exec()` / `.replace()` on the new `gi`-flagged `STRICT_MONTH_YEAR_RE` is paired with a `lastIndex = 0` reset (the codebase invariant), so no stateful-skip bug; the capture (`.exec` in `parseDateRange`) and the strip (`.replace` in `stripDateRange`) use the same regex, as intended.
- **`year_separator` joiner** — the `·` default matches main's previously-hardcoded separator (so a space-only source is unchanged, not newly re-punctuated), and tight (`,;:`) vs loose spacing is unified through one `isTightYearSeparator`, so screen and PDF can't disagree.
- **`EditableField` affordance** — the `+` is its own `aria-hidden` span, the placeholder derives from `label` (making `+ not detected` unrepresentable), and the `Add`/`Edit` verb switch makes the accessible name contain the visible text (WCAG 2.5.3). The `emptyAffordance="plain"` opt-out is the correct seam.

### Findings
No net-new substantive findings beyond the two adversarial rounds and the author's disclosed nits — not padding the review to manufacture some. One outstanding item:

1. **[stands] `tinyglobby` imported but not in `package.json`** (advanced-security bot, `EditableField.test.tsx`). The 23-call-site source-sweep guard works in CI only because `tinyglobby` is a transitive dep of vite/vitest — declare it in `devDependencies` (or use vitest's built-in glob / `node:fs`) so a future dep bump can't silently disable the guard. Worth clearing before merge.

Non-blocking observation (explicitly not filing as a finding): the strict-month strip `.replace()`-all vs `.exec()`-first capture would diverge on a header with 2+ month-year tokens and no range — but the lone-date fallback only fires when `DATE_RANGE_RE` found nothing, so it's near-unreachable, and the title correctly loses both dates regardless. Flagging for the record, not asking for a change.

### Verdict
**Action: COMMENT** — approvable once the `tinyglobby` dep is declared; the logic is sound and independently verified.

s-annam:
@Samhit21 — thanks for the independent pass; the three load-bearing verifications match what the adversarial rounds found.

On the one outstanding item — **`tinyglobby` imported but not in `package.json`** — that one is already fixed at the head you reviewed, and no new push is needed. Receipts:

- The bot scanned **`a5383de`** (20:45 UTC), where the sweep did `const { globSync } = await import("tinyglobby")`.
- The branch was collapsed and force-pushed to **`4829af4`** before your review landed. That commit replaces the glob with the built-in you suggested — `readdirSync(SRC_DIR, { recursive: true, encoding: "utf8" })` from `node:fs` (`EditableField.test.tsx:126`), no third-party dep at all. So there is nothing to declare in `devDependencies`, and a future vite/vitest bump can't silently disable the guard.
- `grep -rn tinyglobby src scripts package.json` on `4829af4` → no matches.
- The bot's inline thread is `isResolved: true` / `isOutdated: true`, and code-scanning alert **268** (`fallow/unlisted-dependency`) is no longer in the open set for `refs/pull/484/merge`. `fallow` is green on the current head, along with `verify` and Cloudflare Pages.

I read the review as approvable-once-cleared, and the clear already happened — so I'm deliberately **not** pushing a no-op commit just to satisfy the thread; that would only dismiss your review and cost a re-read for zero delta.

On the non-blocking observation (strict-month `.replace()`-all strip vs `.exec()`-first capture diverging on a header with 2+ month-year tokens and no range): agreed on both the reachability argument and the outcome — the lone-date fallback only runs when `DATE_RANGE_RE` matched nothing, and the title loses both tokens either way. Leaving it as-is, on the record here.

Re-requesting review on the unchanged head.

s-annam:
Rebased onto latest `main` and resolved the merge conflict.

**What conflicted:** `src/lib/heuristics/regex.test.ts`. `main` had since landed #462, which added a `describe("matchSectionHeader — compound X & Y headers (#462)", …)` block at the same spot this branch inserts its `describe("parseDateRange — lone (un-paired) dates (#380)", …)` block. Same insertion point, two independent additions — a textbook add/add conflict, no semantic overlap.

**Resolution:** kept both `describe` blocks side by side (compound-header block first, then the lone-date block and its siblings). No test logic changed on either side.

**Verification after resolving:**
- `npx vitest run src/lib/heuristics/regex.test.ts` → 87/87 pass
- `npm run typecheck` → clean

Single-commit-per-PR preserved — rebased, not merged, so the branch is still one commit on top of `main`. Now `MERGEABLE` (remaining `BLOCKED` is the review/CI gate, not a conflict).

Samhit21 (approved):
## Approving ✅

Two things to close out from my earlier COMMENT:

**`tinyglobby` — my flag was stale, you're right.** I passed the bot's finding through without checking it against the head I was actually reviewing. The bot scanned `a5383de`; `4829af4` (what I reviewed) had already swapped the sweep to `readdirSync(SRC_DIR, { recursive: true })` from `node:fs` — I confirmed it directly. No dep to declare, fallow green, alert 268 off the list. My miss; thanks for the receipts.

**Rebase verified clean.** On the current head `3160035`: the core #380/#376 logic (`line-primitives`, `entry-dates`, `achievements`, `EditableField`) is byte-identical to what I reviewed, `STRICT_MONTH_YEAR_RE` is intact, and the `regex.test.ts` conflict resolution kept **both** the PR's false-month tests and main's #462 compound-header tests. The `regex.ts`/`ats-resume-model.ts` deltas are main's #462 work pulled in by the rebase, not new PR code.

My three load-bearing verifications stand (STRICT-month `lastIndex` hygiene, the `year_separator` joiner, the `EditableField` affordance). CI green, MERGEABLE. LGTM.

Issue #481, opened 2026-07-14
Re-export openresume-laverne-word-quartz.pdf with a reserved-block phone and @example.com address


s-annam:
## Summary

`tests/fixtures/pdfs/word/openresume-laverne-word-quartz.pdf` carries the phone **`(909) 555-5555`**, which is held open by an explicit exception in the fixture-PII gate (`scripts/check-fixture-pii.mjs`, `EXCEPTIONS`).

It does not meet the repo's fixture-phone policy (CLAUDE.md): a **real area code + the `555` exchange + a `0100`–`0199` subscriber**.

- `909` is a **real, assignable** area code (San Bernardino, CA).
- The subscriber `5555` is **outside** the `555-0100`–`555-0199` block that ITU/NANP reserve for fiction.

So the number is fictional only *by convention*, not by reservation — `555-5555` in a live area code could in principle be assigned to a real line. Every other fixture in the corpus uses the reserved block (e.g. `(312) 555-0123`).

The same file also carries `lleopard@laverne.edu`, likewise excepted. That address belongs to OpenResume's synthetic student persona ("Leo Leopard"), not a real person, but it is not an `@example.com` address either.

## Why it is excepted rather than fixed

The fixture exists to capture the **Microsoft Word → macOS Quartz** export geometry — specifically, Quartz's custom text encoding, which splits the phone across three pdfjs text items separated by `U+002D U+00AD U+2010`. (`pdftotext` misses the number entirely; this fixture is the reason the gate concatenates text items tightly.)

Re-exporting needs Word plus a Quartz print pipeline, and re-encoding the text runs in place would shift glyph widths — changing the exact geometry the fixture exists to pin.

## What to do

Re-export the fixture through Word → Quartz with a compliant persona:

- phone → a reserved number, e.g. `(909) 555-0123` (keeps the area code, moves the subscriber into the reserved block)
- email → `@example.com`

Then **delete both exception entries** for `word/openresume-laverne-word-quartz.pdf` from `EXCEPTIONS` in `scripts/check-fixture-pii.mjs` and confirm `npm run check:fixtures` still exits 0.

The re-export must preserve the split-across-items / soft-hyphen behaviour the fixture is there to test — verify by re-running the corpus tests, not just the gate.

## Acceptance criteria

- [ ] `openresume-laverne-word-quartz.pdf` carries a reserved-block phone and an `@example.com` address.
[...]

Issue #480, opened 2026-07-14
Parser fabricates a valid-looking phone number from experience date digits


s-annam:
## Summary

`findFirstPhone` (`src/lib/heuristics/phone.ts`, reached from the Tier-1.5 regex fallback in `src/lib/heuristics/regex-fallback.ts:79-86`) can **fabricate a phone number that appears nowhere in the résumé**, by folding the digits of an experience **date range** into a number that `libphonenumber-js` then stamps `isValid: true`.

An experience range of `06/2017 – 03/2021` contributes the digit run `2017 03 2021` → `2017032021` → formatted as **`(201) 703-2021`**, which is a structurally valid NJ number. The résumé contains no such phone. It is written into `parsed.phone` with `phoneIsValid: true` and `fieldConfidence.phone = 0.6`, so it flows into the Completeness score and into the reconstructed résumé the user downloads.

## Repro

Neither ingredient triggers this alone — that is why it has stayed invisible. The precondition is **a phone-shaped but INVALID number in the text, plus a date range.**

```ts
import { findFirstPhone } from "./phone.ts";

findFirstPhone("06/2017 – 03/2021", "US");                    // undefined  (prefilter never fires)
findFirstPhone("(555) 018-2390", "US");                       // undefined  (correctly rejected)
findFirstPhone("(312) 555-0123\n06/2017 – 03/2021", "US");    // (312) 555-0123  — correct
findFirstPhone("(555) 018-2390\n06/2017 – 03/2021", "US");    // (201) 703-2021  ← FABRICATED, isValid: true
```

The mechanism: the invalid `(555) 018-2390` is phone-**shaped**, so it trips the cheap `mightHavePhone` prefilter and opens the full `libphonenumber-js` scan over the text. libphonenumber correctly rejects the 555 number (555 is not a valid NANP area code) — but it does not stop there. It keeps scanning, joins digits **across** the `/`, space and dash separators of the date range, and returns the first thing that validates.

This is the same digits-across-separators behaviour that makes `libphonenumber-js`'s `findNumbers` unusable as a detector in the fixture-PII gate (#478): it reports `06/2017 - 03/2021` as `+12017032021`.

## How it was found

Found while fixing the fixture-PII gate (#478). `tests/fixtures/pdfs/unknown/chromium-asymmetric-sidebar.pdf` used to carry `(555) 018-2390` as its contact phone — an invalid number — and the parser fabricated `(201) 703-2021` from its experience dates.

[...]

PR #485, opened 2026-07-14
chore(ci): enforce the fixture-PII policy mechanically, not just in prose (#478)

### LINKED ISSUE #478: Enforce the fixture-PII policy in CI, not just in prose
## Problem

The fixture-PII policy is the one rule in this repo whose failure mode is **unrecoverable** — the repo is public, the committed PDF *binary* is the exposure surface, and purging a leaked fixture post-merge means `git filter-repo` plus a GitHub Support ticket. It is also the only rule with **zero mechanical enforcement**.

Today it rests entirely on instructions to a model:

| Surface | Fires when | Gap |
|---|---|---|
| `CLAUDE.md` → **Hard rules** | Always in an agent's context | Only as reliable as the agent reading it |
| `tests/fixtures/pdfs/CLAUDE.md` (directory-scoped) | Agent reads/edits a file in that subtree | **Misses a fixture added via a bare `cp` / `curl` in Bash** — no Read/Edit, no load |
| `/open-pr` Step 3.5, `/pr-review` gate 3a | Only when those skills run | An ad-hoc session that commits and pushes by hand never runs them |
| CI | — | **Nothing.** `.github/workflows/` is `ci.yml`, `deploy-pages.yml`, `moderate-comments.yml`, `promote-release.yml`; none scans a fixture binary |

Every layer is advisory. Nothing fails a build.

### This is not hypothetical — the current corpus already violates the policy

Auditing all 45 PDFs under `tests/fixtures/pdfs/` against the policy turns up three that would fail the gate proposed below:

**1. `unknown/chromium-asymmetric-sidebar.pdf` — an invalid area-code-`555` number (real defect, live today).**

```
$ pdftotext tests/fixtures/pdfs/unknown/chromium-asymmetric-sidebar.pdf - | head -3
Jordan Avery
Staff Software Engineer
(555) 018-2390 · jordan.avery@example.com · Austin, TX
```

`555` is an **invalid NANP area code**, so `libphonenumber-js` `isValid()` — which the parser uses — rejects it. This fixture's `phone` is **silently dropping out of the parse and out of the Completeness score right now**, and no test catches it, because the `*.expected.json` snapshots are keys/counts-only by design. This is precisely the trap `CLAUDE.md` documents (mandate a *real* area code + `555` **exchange** + `0100`–`0199` subscriber, e.g. `(312) 555-0123`), and it is sitting in the corpus anyway. It is the single best argument that instructions alone do not hold.

**2. `word/openresume-laverne-word-quartz.pdf`** — email is `lleopard@laverne.edu`, not `@example.com`. Persona ("Leo Leopard") appears to be OpenResume's synthetic demo, so this is likely policy-compliant *in spirit* but fails the letter of the `@example.com` rule.

**3. `unknown/openresume-react-pdf.pdf`** — email is `hello@openresume.com` ("John Doe", `312-555-0190`). Same situation: an upstream project's own demo address, not a real person's, but not `@example.com`.

Good news from the same audit: **PDF metadata is clean.** No `Author` field across the corpus carries a real name — `latex/awesome-cv-cv.pdf` and `awesome-cv-resume.pdf` have an empty `Author`, confirming the #384 PII purge held.

## Proposal

Add a **fixture-PII check that fails CI**, plus a matching `npm run` script so it is runnable locally and wireable into `verify`.

### Scope of the scan

Scan **every** PDF under `tests/fixtures/pdfs/` on every run, not just the ones in the diff. A diff-scoped check is cheaper but misses the case where a fixture is added on one branch while the rule tightens on another, and the corpus is only 45 files — a full scan costs a couple of seconds. (Diff-scoping is a fine *optimization* later; correctness first.)

### The rules to encode

For each PDF, extract text (`pdftotext <f> -`) and metadata (`pdfinfo <f>`), then fail when:

1. **No `@example.com` email is present.** The persona's email must be `@example.com` — the reserved-for-documentation domain (RFC 2606).
2. **A phone is present whose shape is not `<real area code> 555 01XX`.** Specifically: exchange must be `555`, subscriber must be `0100`–`0199`, and the area code must **not** be `555`. Encode all three — a naive `grep 555` passes `(555) 018-2390` and defeats the entire point.
3. **A known real-persona string appears.** Maintain a small denylist of the names/emails/phones that OSS résumé templates ship in their demo PDFs — Awesome-CV embeds posquit0's real CV, Deedy-Resume embeds Debarghya Das's. This is what caused #384.
4. **Metadata carries a non-empty `Author`, or an XMP `dc:creator`,** that is not an allowlisted synthetic name. A PDF's body text can be fully synthetic while the Info dict still names the person who exported it.

A phone being *absent* is not a failure (several Word fixtures have no phone line); an *invalid* phone is.

### Suggested shape

```
scripts/check-fixture-pii.mjs      # new — the scanner
package.json                       # + "check:fixtures": "node scripts/check-fixture-pii.mjs"
.github/workflows/ci.yml           # + a step in the `verify` job
```

The scanner should print one line per failing fixture with the offending value, so the fix is obvious without re-running `pdftotext` by hand:

```
✗ unknown/chromium-asymmetric-sidebar.pdf
    phone "(555) 018-2390" — 555 is an invalid NANP area code; libphonenumber-js
    will reject it and the field will silently drop. Use a real area code with the
    555 exchange, e.g. (312) 555-0123.
```

**CI needs `poppler-utils`.** `pdftotext` / `pdfinfo` are not on the `ubuntu-latest` image by default — the workflow step must `apt-get install -y poppler-utils` (or the scanner should shell out to `pdfjs-dist`, which is already a dependency, and skip the system-binary dependency entirely; that is probably the better call since the repo already parses PDFs in JS).

### Baseline must be clean

Per this repo's convention for enforcement gates, land the check **and** the debt fix together, so the gate goes green on a clean tree:

- Re-export `chromium-asymmetric-sidebar.pdf` with a valid phone (e.g. `(512) 555-0142`) and re-bake its snapshot. **Note this will change the parse result** — the fixture's `phone` currently drops, so fixing it means `phone` starts populating and the Completeness score for that fixture moves. That snapshot delta is the *proof the bug was real*, and should be called out in the PR rather than waved through.
- Decide the two OpenResume fixtures: either re-export with `@example.com` addresses, or add a narrow, commented allowlist entry recording *why* an upstream project's own demo address is acceptable. Prefer re-exporting — an allowlist entry is a hole someone will widen later.

## Implementation plan

1. **Write `scripts/check-fixture-pii.mjs`.** Walk `tests/fixtures/pdfs/**/*.pdf`. For each, extract text and metadata (prefer `pdfjs-dist`, already a dep, over shelling out to poppler). Apply rules 1–4 above. Collect all failures, print them, exit `1` if any.
2. **Encode the phone rule precisely.** A regex that accepts `(NXX) 555-01XX` where the area code is a valid NANP form (`[2-9][0-8][0-9]`) and explicitly is not `555`, exchange is exactly `555`, subscriber is `0100`–`0199`. Unit-test the regex against both `(312) 555-0123` (pass) and `(555) 018-2390` (fail) — those two cases are the whole point of the check.
3. **Fix the three baseline violations** (above), re-baking any affected `*.expected.json` with `npm run bake-fixtures`.
4. **Wire it up:** add `"check:fixtures"` to `package.json`, call it from `npm run verify`, and add the step to the `verify` job in `.github/workflows/ci.yml`.
5. **Point the prose at the gate.** Once this exists, `CLAUDE.md` → Hard rules, `docs/CONTRIBUTING-PROCESS.md`, and the `/open-pr` / `/pr-review` preflights should each say "and `npm run check:fixtures` enforces this" — the instruction becomes a *pointer to a check*, not the check itself.

## Acceptance criteria

- [ ] `npm run check:fixtures` exists, scans every PDF under `tests/fixtures/pdfs/`, and exits non-zero with a per-file explanation when a fixture violates the policy.
- [ ] The phone rule accepts `(312) 555-0123` and **rejects** `(555) 018-2390`, with a unit test asserting both.
- [ ] The check rejects a fixture with no `@example.com` email, a denylisted real-persona string, or a non-empty metadata `Author`.
- [ ] `chromium-asymmetric-sidebar.pdf` is re-exported with a valid phone; its `*.expected.json` is re-baked and the resulting parse delta (phone now populates) is called out in the PR.
- [ ] The two OpenResume-derived fixtures either use `@example.com` or carry a commented allowlist entry justifying the exception.
- [ ] `npm run check:fixtures` is green on a clean tree, is called from `npm run verify`, and runs as a step in the `verify` CI job.
- [ ] CI has whatever PDF-text-extraction capability the scanner needs (either `poppler-utils` installed, or the scanner uses `pdfjs-dist` and needs nothing).
- [ ] Existing suites still pass: `corpus.test.ts` and `corpus-roundtrip.test.ts` green after the fixture re-bakes.

## Why this is worth doing

It converts fixture PII from **an instruction any skill-less session can miss** into **a property the build enforces**. After this lands, where the policy prose lives — root `CLAUDE.md`, a directory-scoped file, a skill body, a process doc — stops being load-bearing. The check is.

The `(555) 018-2390` fixture is the proof: the rule was written down in three places, and a violation still walked into the corpus and has been silently degrading a fixture's parse ever since.

---

Context: filed out of the #477 review, which split `CLAUDE.md` into code guidance vs. `docs/CONTRIBUTING-PROCESS.md` and surfaced that the PII rule's enforcement was entirely instructional. Prior PII incident: #384.

s-annam:
## Summary

The fixture-PII rule guarded an unrecoverable failure — this repo is public, so a leaked real name, phone, or email means `git filter-repo` plus a GitHub Support ticket — yet it had **zero mechanical enforcement**. It lived only in prose, and the corpus had already drifted from it.

This adds `scripts/check-fixture-pii.mjs` (`npm run check:fixtures`), wired into `verify` and into CI as a bare step that cannot exit 0 on failure. It scans every PDF under `tests/fixtures/pdfs/` and encodes four rules: an `@example.com` email must be present; any phone must pair a real area code with the `555` exchange and a `0100`–`0199` subscriber; no denylisted real persona from an OSS template; no non-empty metadata Author.

**A leak hides on more surfaces than the drawn page**, so the scanner reads all of them: page text, link annotations (`tel:`/`mailto:` hrefs), the Info dict *including* the `Custom` bucket pdfjs parks non-standard keys in, raw XMP, annotation contents and author, AcroForm field values and tooltips, the outline tree, and embedded attachments. Each surface has an integration test over the real `extractPdf` — not just the pure policy core — so dropping a surface turns a test red.

Three fixtures already violated the policy and are cleaned here:

- `awesome-cv-cv.pdf`, `awesome-cv-resume.pdf` — drew a compliant phone while their `tel:` link annotation still carried a forbidden area-code-555 number.
- `chromium-asymmetric-sidebar.pdf` — drew an area-code-555 number in its text.

Metadata rewrites preserve each file's original Producer and PDF version, so the Skia and LuaTeX fixtures still testify to the renderer they were captured from.

Closes #478

## Fallow (CI) — fixed

Fallow failed on the first push with 5 alerts, all on the new gate. Two were errors, and neither was really a complexity problem:

- **The CRAP scores were inflated by missing coverage, not by untested code.** `coverage.include` was `src/**/*.{ts,tsx}`, so the coverage report never mentioned `scripts/check-fixture-pii.mjs` — fallow read that as **0% covered** and applied the full cyclomatic penalty (`extractPdf` scored `27 + 27² = 756`). The file is in fact **91% statement-covered**. Adding it to `include` matched the coverage and the inflation vanished.
- **The gate is listed by name, not globbed.** A `scripts/**/*.mjs` glob also sweeps in the one-shot scripts no test loads (fixture generators, the hook installer), and v8 emits a **negative source-map column** for those — the same bug the existing `main.tsx` exclude exists to dodge, which crashes fallow's u32 coverage parser and **silently zeroes the entire report**. Enumerating inverts the failure mode: forget to add a future tested gate and fallow merely scores it 0% and complains, loudly.

With coverage matched, `extractPdf` (27 cyclomatic, 83 lines) and `checkFixture` (18, 87) were still genuinely over the complexity threshold, so they were **decomposed rather than suppressed** — the four policy rules and the five PDF-surface collectors each became a named function. The `unused-file` error on the test suite was fallow's vitest plugin not discovering `.mjs` tests; it is registered as an entry in `.fallowrc.jsonc`.

`npm run verify` is green end to end (2296 passed, fallow: **no issues in 15 changed files**), and the mutation check was re-run after the refactor — dropping the outline scan still turns a test red.

## Known gaps (deliberate, not oversights)

- **ESLint still ignores `scripts/**`**, so the gate's own source is unlinted. The fix is ~4 lines in `eslint.config.js`; a `config-protection` hook blocks that file, so it is left for a maintainer rather than routed around. Worth a follow-up issue.
- **The gate scans PDFs only** — not the png/jpeg/docx fixtures — and **it cannot judge whether a *name* is synthetic.** No check can. That judgement stays with the human reviewer, and the docs say so.

## Spun off, not fixed here

- **#480** — the parser **fabricates** a valid-looking phone from a date range: `chromium-asymmetric-sidebar.pdf` parsed as `(201) 703-2021`, `phoneIsValid: true`, a number appearing nowhere in the résumé (its digits are the experience range `06/2017 – 03/2021`, and `201` is a real NJ area code, so libphonenumber blessed it). Found while fixing that fixture. Parser behaviour intentionally unchanged in this PR.
- **#481** — re-export the `(909) 555-5555` Word fixture so its `EXCEPTIONS` entry can be dropped.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — 2296 passed / 9 skipped
- [x] `npm run check:fixtures` exits 0 over all 46 PDFs
- [x] **Negative control:** a PDF hiding a denylisted persona + forbidden phone in the **outline tree** (clean drawn text) fails the gate with exit 1
- [x] **Mutation check:** removing any one surface from `extractPdf` turns a test red
- [x] Fixture personas verified synthetic — `Jane Smith`, `Jordan Avery`, `@example.com`, `(312) 555-0123`, empty Author, on the binaries themselves

## Adversarial review

Three rounds ran pre-PR. **Each of the three found a real, reproduced hole**, and every one was the same class: *a PII-bearing surface of the PDF the scanner never looked at.* Recording that plainly, because the pattern is the most useful thing this PR learned.

| Round | Blocking finding | Resolution |
|---|---|---|
| 1 | Gate never scanned **link annotations**. Two fixtures carried `tel:+15550100123` in a `tel:` href while the gate reported green. | Annotations scanned; both binaries fixed. |
| 2 | Gate never scanned **`info.Custom`** — pdfjs buckets non-standard Info keys (`/Company`, `/Manager`) into an *object*, and the code filtered `typeof value === "string"`. A PDF with PII there **passed**. Also: a **fabricated justification** ("`pdftotext` misses a split phone" — it does not) had been written into six files. | `flattenInfoValues()` recurses; the false claim deleted from all seven sites and the argument re-grounded on the verified blind spots. |
| 3 | Gate never scanned **annotation contents/author, AcroForm values, the outline tree, or embedded attachments**. Four purpose-built PDFs drew a clean persona while hiding PII on those surfaces — **all four passed, exit 0**. Live risk: 9 of 46 fixtures have populated outlines and two carry a person's name there; Word writes the résumé owner's name into the outline from Heading 1, and it **survives a body-text scrub**. Separately: the unit suite **could not have caught any of the three bugs** — all 39 tests exercised the pure policy core with pre-folded strings, and stayed **39/39 green** when both prior bugs were reintroduced into `extractPdf`. | All surfaces folded in. Integration tests added over the real `extractPdf`, one per surface, verified by mutation. |

Surviving nits, all fixed: `Byungjin  Park` (double space) evaded the denylist for want of a single-token backstop; fullwidth digits evaded the phone regex.

**Round 4 did not run.** The round-3 fixes are verified by the orchestrator (negative control + mutation check re-run independently) but were not themselves put through an adversarial pass. Given that each prior round found something, a reviewer should treat this diff as reviewed-but-not-converged rather than clean.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #478 | Claude Opus 4.8 | ultra |
| Adversarial review — round 1 | Claude Opus 4.8 (1M context) | high |
| Review fixes — round 1 | Claude Opus 4.8 (1M context) | high |
| Adversarial review — round 2 | Claude Opus 4.8 (1M context) | high |
| Review fixes — round 2 | Claude Opus 4.8 (1M context) | high |
| Adversarial review — round 3 | Claude Opus 4.8 (1M context) | high |
| Review fixes — round 3 | unreported (requested: `opus`) | high |
| Orchestration + PR | Claude Opus 4.8 | high |


rohithgollapalli (approved):
Strong PR — this turns the one repo rule whose failure is unrecoverable (a real persona in a public-repo PDF binary) from prose into a mechanical gate, and does it thoroughly. I verified the gate end-to-end, not just the prose.

**What I checked:**
- **Fixture PII (the gate's own 3 re-exported binaries)** — all synthetic: `awesome-cv-cv` / `awesome-cv-resume` → "Jane Smith", `jane.smith@example.com`, `(312) 555-0123` (real area code + 555 exchange + `0123` subscriber ✓); `chromium-asymmetric-sidebar` → "Jordan Avery", `jordan.avery@example.com`, same valid form. No `Author`/`Title` metadata leak. The Awesome-CV pair was correctly **re-exported**, not posquit0's shipped demo. ✓
- **The gate runs and passes** — `node scripts/check-fixture-pii.mjs` → `✓ 46 PDFs — all personas synthetic` (also confirms the `main()` self-invoke guard fires under Node, on Windows too).
- **Unit tests** — `scripts/check-fixture-pii.test.mjs` 52/52 green via vitest.
- **Correctness spot-checks** — the candidate/compliance phone split correctly catches area-code-`555` (the #478 founding bug) while a naive `grep 555` would pass it; a `+1 (312) 555-0123` matched via the intl branch still passes `isPolicyCompliantPhone` (no false failure); the fused-email suffix (`@example.comAustin`) is tolerated only for `@example.com` while a fused real domain still fails; extraction folds every hiding surface pdftotext/`getTextContent` miss (link annotations, outline tree, attachment bytes, Info `Custom` bucket, XMP).
- **CI / config wiring** — `check:fixtures` runs before `test:coverage` (fail-fast) with a comment noting no poppler needed; `verify` updated; `.fallowrc` registers the `.mjs` suite as an entry; `vite.config` includes it in the test glob and enumerates the script in coverage-include (with the reasoning for not globbing — the v8 source-map crash). All correct.
- **Docs don't overstate the gate** — CLAUDE.md / fixtures README / both skills state plainly that a name is a human call and "a green check does not mean the fixture is clean," matching the script's own docblock. ✓

## Nits (non-blocking)

- **fallow flags 9 high-CRAP functions**, all the pdfjs **extraction** helpers in `scripts/check-fixture-pii.mjs` (`annotationStrings:575`, `collectMetadata:636`, `collectPageStrings:594`, …). They read as ~0% covered because the unit tests exercise the pure rule core (`checkFixture`) rather than the pdfjs I/O — which is instead integration-exercised by the gate itself on 46 real fixtures each CI run. `fallow` is report-only in `verify` (`|| echo … ignored`), so this won't fail CI. If you want it quiet, a couple of `extractPdf`/annotation unit tests against a tiny in-repo PDF (or a scoped `// fallow-ignore-next-line complexity`) would do it. Purely optional.
- `checkPersonaRule` folds `author`/`xmpCreator` into the haystack even though `surface` already contains the flattened Info + raw XMP — harmless redundancy, not worth a change.

## Verdict

0 Blocking → **APPROVE**. The gates that apply (fixture-PII, script-as-code, CI wiring, docs accuracy) all pass; the only findings are report-only fallow nits on the extraction I/O.

---

Issue #487, opened 2026-07-14
Download PDF drops bullet overrides on the edit round-trip


Samhit21:
Surfaced by the new edit-leg corpus gate (#459).

## What

For a résumé role, the **first bullet** does not survive the Download-PDF round-trip. Overriding the first bullet observation and running the real `parse → applyOverrides → buildAtsResumeModel → renderAtsResumePdf → runCascade` pipeline, the re-parse comes back with the leading bullet **missing** — the reconstructed role starts at the *second* bullet. Both the edited text and the value it replaced are gone; the model never carries the first bullet.

Reproduced on `google-docs/google-docs-skia-proxy-classic.pdf`:
- `applyOverrides` output: `experience[0].description` begins with the overridden bullet ✓
- built model: neither the new nor the original first-bullet text is present ✗
- `parse3.experience[0].description`: starts with the **second** bullet — first bullet dropped ✗

## Why it slipped through

`corpus-roundtrip.test.ts` (#293) gates bullet **count/mapping**, not text, so a dropped-then-shifted bullet leaves counts plausible and never trips it. The edit-leg gate (#459) asserts the *overridden* bullet text is present, which is what catches it. This is the exact "silent edit loss" class #459 was built to find.

## Scope

Root-cause and fix the first-bullet drop between `buildAtsResumeModel` and the reconstructed-PDF re-parse. When fixed, delete the **Group A** `bullets` lines from `corpus-edit-roundtrip.test.ts`'s `KNOWN_FAILURES` (the ratchet forces it — a fixed fixture trips the stale-entry check).

Baselined fixtures (Group A): 26 fixtures carrying an experience bullet — see the `KNOWN_FAILURES` Group A block.

Tracked from #459.

Samhit21:
**Correction (from #488 review):** this is **not** first-bullet-specific. Under the faithful Download-PDF recipe, overriding the *second* bullet observation (`observations[1]`) also fails to survive on `skia-proxy-classic`, `awesome-cv-resume`, and `weasyprint-cairo-classic`. So the defect is **bullet-override survival on the export path** generally, not a leading-bullet off-by-one — don't go hunting in the first-bullet path. Title updated accordingly. (The drop is a real bug, confirmed reproducing under a faithful recipe, not a harness artifact.)

Issue #445, opened 2026-07-11
Stage D+E — cutover: collapse remaining peer shapes + remove CascadeResult façade


s-annam:
# Stage D+E — cutover: collapse remaining peer shapes + remove `CascadeResult` façade

Combined fourth-and-final stage of the canonical-model migration (#438 decision, #439 design). Folds in and supersedes #446 (Stage E), which is closed as superseded — its scope (façade removal + #321 cache versioning) lives here now. Full context: [`docs/canonical-resume-model.md`](https://github.com/resumelint-org/resumelint/blob/main/docs/canonical-resume-model.md) §4 Stage D + Stage E, §2.3, §6, §7, §8.

## Intent

Retire the last two peer shapes, the hand-sync lockstep (doc Cost 1), **and** the `CascadeResult` compatibility façade in one verifiable cutover:

- **`LlmParsedResume`** (`webllm/parse-resume.ts:56`) stops being a peer type. Coerce on-device LLM output into `CanonicalResume`; `diffParses` (`disagreement.ts:186`, #242) reads a new **llm-diff projection**. Delete the hand-sync note (`parse-resume.ts:52-54` — *"Keep field names in sync"*).
- **`ApplyOverridesResult`** (`edit/apply-overrides.ts:52`) collapses into `CanonicalResume`. One user edit mutates one field; `parsed`/`rawText`/`sections`/`fieldConfidence` (the #133/#421 lockstep quadruple) become projections read on demand, not lockstep return values.
- **`fieldConfidence` becomes a third member of `CanonicalResume`.** Today `CanonicalResume = { fields: HeuristicParsedResume, sections: SectionedResume }` (`canonical.ts:49-54`). Add `fieldConfidence` alongside them.
- **`CascadeResult` façade removed.** `CanonicalResume` becomes the sole shape — no compatibility layer left to lean on.
- **#321 IndexedDB cache versioned** for the persisted-shape change (see below).

## Why combine D and E

[...]

PR #488, opened 2026-07-14
test(corpus): edit-leg round-trip gate — assert overrides survive Download PDF (#459)

### LINKED ISSUE #459: Corpus gate: round-trip the EDIT leg — assert user overrides survive Download PDF
## Problem

The core Friends & Family loop is:

```
drop PDF → parse → user corrects what the parser got wrong → Download PDF
```

**Nothing in the test suite asserts that step 3 survives step 4.** A user can fix a mis-parsed job title, correct their email, add a skill — and if the export path drops that edit, no gate fires. That is the single most damaging silent failure the product has, because the whole value proposition of the edit surface is "fix it here and the downloaded PDF is right."

### What we gate today (and what it misses)

`src/lib/heuristics/corpus-roundtrip.test.ts` (#293) already round-trips all 45 corpus fixtures through the real Download-PDF surface:

```
parse1 = runCascade(fixture)
model  = buildAtsResumeModel(parse1, scoreFor(parse1))
bytes  = renderAtsResumePdf(model)          // the "Download PDF" surface
parse3 = runCascade(bytes)
assert invariantFailures(parse1, parse3) is empty  (modulo KNOWN_FAILURES)
```

It has a ratchet: `KNOWN_FAILURES` lists, per fixture, which invariant categories may currently fail; a non-baselined regression fails the gate, and a baselined category that starts passing ALSO fails the gate (stale entry → must be deleted). The baseline can only shrink. 18 of 45 fixtures are currently baselined on `experience` (tracked separately by #436 / #437 — **out of scope here**).

**But that pipeline has no user edits in the middle.** It proves the renderer emits shapes our own parser round-trips. It proves nothing about override fidelity.

### The export path IS override-aware — which is why this is testable

Verified against the live wiring (this is not a hypothetical):

| Step | Where |
|---|---|
| Overrides folded into a fresh field core | `src/hooks/useAnalyzedResume.ts:147` — `applyOverrides(...)` → `editedCore` |
| Edited core folded back onto the base result | `useAnalyzedResume.ts:278` — `displayResult` = `{ ...base, canonical: { ...base.canonical, fields: edited.parsed, fieldConfidence: edited.fieldConfidence } }` |
| Handed to the surface | `src/App.tsx:225` — `result={displayResult}` |
| Download click | `ReconstructedResume.tsx:992` — `useDownloadPdf(result, score, edit)` |
| Model built | `src/hooks/useDownloadPdf.ts:48` — `buildAtsResumeModel(result, score, edit)` |

Two load-bearing details a naive test would get wrong:

1. **`buildAtsResumeModel` takes BOTH inputs.** Its signature is `(result: CascadeResult, score: AnonymousAtsScore, edit?: Pick<EditableParse, "contactOverrides" | "bulletOverrides">)` (`src/lib/pdf/ats-resume-model.ts:473`). The override-applied `result` is not the whole story — contact and bullet overrides are *also* passed separately. A test that only passes the applied result does not reproduce the real Download PDF.
2. **`displayResult.canonical.sections` stays the BASE's**, un-edited, on purpose (#445) — only `fields` and `fieldConfidence` carry the edit. So the test must construct `displayResult` exactly as `useAnalyzedResume` does, not by swapping the whole canonical model.

## Proposed change

Add a sibling corpus gate, `src/lib/heuristics/corpus-edit-roundtrip.test.ts`, that runs the **edit leg** across the same 45 fixtures:

```
parse1  = runCascade(fixture)
edits   = synthesizeOverrides(parse1)        // synthetic, deterministic, PII-free
applied = applyOverrides(parse1.canonical.fields, parse1.rawText,
                         parse1.canonical.sections, ...edits)
display = { ...parse1, canonical: { ...parse1.canonical,
                                    fields: applied.fields,
                                    fieldConfidence: applied.fieldConfidence } }
model   = buildAtsResumeModel(display, scoreFor(display),
                              { contactOverrides: edits.contact,
                                bulletOverrides:  edits.bullets })
parse3  = runCascade(await renderAtsResumePdf(model))

assert: every OVERRIDDEN value is present in parse3 — and the value it replaced is not.
```

The assertion is the inverse of the existing gate's. `corpus-roundtrip` asserts *parse1 ≡ parse3*. This one asserts *parse3 reflects the edits, not parse1*. A regression that silently discards a user edit makes parse3 equal parse1 — which the existing gate happily calls a pass and this gate calls a failure. That's the whole point.

### Why this is PII-safe by construction

We assert on **strings we authored**, never on fixture values. `synthesizeOverrides` emits literals like `"Staff Platform Engineer"` / `"Northwind Systems"` / `"jordan.doe@example.com"`; the assertions check those literals appear in `parse3`. No fixture field value is ever snapshotted or compared, so the existing "snapshots are lossy by design" PII property (`tests/fixtures/pdfs/README.md`) holds automatically. Follow the fixture phone policy for any synthetic phone override: real area code + `555` exchange + `0100`–`0199` subscriber (e.g. `(312) 555-0123`) so `libphonenumber-js` `isValid()` accepts it.

### The synthetic override set

Deterministic, derived from `parse1` so it applies to any fixture, covering one edit of each kind that the UI actually offers:

| Kind | Override | Skip when |
|---|---|---|
| contact | `email` → `jordan.doe@example.com` | never (always settable) |
| experience | role 0 `title` → `"Staff Platform Engineer"`, `company` → `"Northwind Systems"` | `experience.length === 0` |
| bullets | first bullet observation → `"Cut p99 checkout latency 43% by resharding the session store."` | no bullet observations |
| skills | add `"Kubernetes"` | never |
| addedEntries | append one education entry | never |

Each fixture skips only the kinds it structurally can't exercise, and the test records which kinds it exercised so a fixture silently degrading to "edited nothing" is itself a failure.

### Ratchet discipline

Same as #293, same shape, same teeth:

- `KNOWN_FAILURES: Record<string, EditCategory[]>` keyed by fixture path relative to `tests/fixtures/pdfs/`.
- A **non-baselined** category that fails → gate fails (protects everything passing today).
- A **baselined** category that passes → gate fails with `remove it from KNOWN_FAILURES` (stale entry; the baseline can only shrink).
- A `KNOWN_FAILURES` key naming a fixture that no longer exists → gate fails.

**Commit the baseline with whatever fails today. Do not fix the failures in this issue** — triage them and file follow-ups, linked back here.

### Reuse analysis

**Capability:** walk the fixture corpus, score a parse, render through the real export surface, apply a per-fixture known-failure ratchet.

**Existing surfaces found:** `src/lib/heuristics/corpus-roundtrip.test.ts` — `walkPdfs()`, `relKey()`, `scoreFor()`, and the ratchet loop. All are currently **module-local** (not exported).

**Decision: extend by extraction, not clone.** Lift `walkPdfs` / `relKey` / `scoreFor` and the ratchet assertion loop into a shared helper module consumed by **both** tests (`corpus-roundtrip.test.ts` and the new `corpus-edit-roundtrip.test.ts`), and update #293's test to import them. Two consumers from day one, so the extraction does not create a forward-staged export that the `fallow` dead-code gate would flag. Copy-pasting the walker and the ratchet into a second file is explicitly rejected — the ratchet's correctness (the stale-entry check in particular) is subtle enough that two divergent copies is a real risk.

The **invariant/assertion layer is genuinely new** and is not shared: `invariantFailures()` compares parse1↔parse3 for equality, whereas this gate asserts overridden-value presence. Different question, different code.

## Implementation plan

1. **Extract the shared corpus harness.** New non-test module (e.g. `src/lib/heuristics/corpus-harness.ts`, or a `.test-helpers.ts` sibling if that reads better against the lint config): export `walkPdfs`, `relKey`, `scoreFor`, the `Category` ratchet primitives, and the stale-key check. Make the ratchet generic over the category union so both gates share it.

2. **Rewrite `corpus-roundtrip.test.ts` to import from the harness.** Pure refactor — the existing 18-line `KNOWN_FAILURES` baseline and every assertion must behave identically. `npm run test` must be green with **zero** baseline edits in this step.

3. **Add `synthesizeOverrides(parse1)`** in the new test file (or the harness if it proves shareable): returns the override maps + a record of which kinds were exercised. Pure, deterministic, no randomness.

4. **Add `corpus-edit-roundtrip.test.ts`.** Per fixture: run the 5-step pipeline above, construct `displayResult` exactly as `useAnalyzedResume.ts:278` does, pass the `{ contactOverrides, bulletOverrides }` pick to `buildAtsResumeModel` as `useDownloadPdf.ts:48` does, and assert per-category. Categories: `contact`, `experience`, `bullets`, `skills`, `added`, `render`. Scope a `{ timeout: 20000 }` on the `describe` as #293 does — a full cascade→render→cascade per fixture is slow under coverage instrumentation (#360).

5. **Bake the `KNOWN_FAILURES` baseline** from the first full run. Group entries by likely shared root cause with a comment per group, matching #293's house style.

6. **Triage the baseline and file follow-ups**, one issue per root-cause group, each linking back to this issue. Post the list as a comment here.

## Acceptance criteria

- [ ] `src/lib/heuristics/corpus-edit-roundtrip.test.ts` exists and runs over every fixture in `tests/fixtures/pdfs/` (currently 45).
- [ ] The gate constructs the model through the **real** production recipe: override-applied `displayResult` **plus** the `{ contactOverrides, bulletOverrides }` pick into `buildAtsResumeModel`. A test that passes only one of the two is not acceptable.
- [ ] For each fixture, every override kind the fixture can structurally exercise IS exercised; a fixture that ends up editing nothing fails the gate.
- [ ] Assertions check that the **overridden** value appears in the re-parse and the value it replaced does not.
- [ ] `KNOWN_FAILURES` ratchet is implemented with all three teeth: non-baselined failure fails; baselined-but-now-passing fails ("remove it from `KNOWN_FAILURES`"); stale key naming a missing fixture fails.
- [ ] The baseline is committed as-found. No round-trip bug is fixed in this PR.
- [ ] Shared harness is extracted and consumed by **both** corpus gates; no duplicated walker or ratchet. `corpus-roundtrip.test.ts`'s existing baseline is unchanged by the refactor.
- [ ] All override literals are synthetic. Any phone override uses a real area code with the `555` exchange and a `0100`–`0199` subscriber. No fixture field value is snapshotted.
- [ ] `npm run verify` is green (typecheck → lint → coverage → build → fallow). In particular `fallow` reports no new dead/forward-staged export from the harness extraction.
- [ ] **Regression proof:** temporarily stubbing the override fold (e.g. making `applyOverrides` return its input) makes the new gate fail loudly. Note in the PR that this was verified — it's the one check that proves the gate has teeth.
- [ ] Follow-up issues filed for each baselined root-cause group, linked from a comment here.

## Out of scope

- Fixing any round-trip failure the new baseline records (that's the follow-ups).
- #436 / #437 and the existing 18 `experience` entries in the #293 baseline.
- The JD-fit lane's edit-snapshot handoff (`src/lib/jd-fit-handoff.ts`) — a different consumer of the same overrides, worth its own gate later.

Samhit21:
Closes #459.

## What

Adds `corpus-edit-roundtrip.test.ts` — a sibling of the #293 self-consistency gate that runs the real Friends & Family loop across all 46 fixtures:

```
drop PDF → parse → user corrects what the parser got wrong → Download PDF
```

and asserts the corrections **survive** the export. #293 round-trips with **no edits**, so it proves the renderer emits shapes our parser round-trips and *nothing* about override fidelity. A user fixing a mis-parsed title/bullet and hitting Download PDF had zero coverage that the edit lands — the single most damaging silent failure the product can have.

## The gate

Per fixture: synthesize a deterministic, PII-free override of each kind the UI offers (contact / experience / bullets / skills / added), then reproduce the **exact** production Download-PDF recipe and assert **every overridden value is in the re-parse and the value it replaced is not**. The assertion is the inverse of #293's (`parse1 ≡ parse3`): a regression that silently discards an edit makes `parse3 ≡ parse1`, which #293 calls a *pass* and this gate calls a *failure*.

Three production details reproduced exactly (the AC — a test that gets any wrong is not acceptable):
1. `buildAtsResumeModel` gets BOTH the override-applied `displayResult` AND the separate `{ contactOverrides, bulletOverrides }` pick (`useDownloadPdf`).
2. `displayResult.canonical.sections` stays the **base's**, un-edited (#445).
3. `applyOverrides` gets the base `fieldConfidence` (`useAnalyzedResume`) — **without it, every non-edited field gates to `""` and the export drops phone/location/links**, a model production never renders.

## Collision-proof literals

Every override literal carries a `Vantreon` sentinel that can't occur in a résumé, and a per-fixture guard fails loudly if any literal already appears in the parse — so a presence check can never be silently vacuous (a plain literal like `Kubernetes`/`Northwind Systems` occurs verbatim in real fixtures and would pass even if the override were discarded).

## Shared harness (no duplication)

`walkPdfs` / `relKey` / `FIXTURE_ROOT` + the generic ratchet (`assertRatchet` / `assertNoStaleKeys`) extracted into `corpus-gate.test-utils.ts`, consumed by **both** gates; `corpus-roundtrip.test.ts` refactored to import them — pure refactor, baseline unchanged, green. `scoreForCascade` exported from `roundtrip-hop.ts`. Two consumers from day one → `fallow` dead-exports **0.0%**.

## Baseline (baked as-found — no bug fixed here)

- **Group A → #487:** bullet overrides do **not** survive the Download-PDF round-trip (confirmed not first-bullet-specific — the 2nd observation drops too). Every bullet-bearing fixture is baselined, so `bullets` has **no teeth corpus-wide today** — stated plainly, not presented as coverage; teeth return one fixture at a time as #487 shrinks the list.
- **Group B → #436:** one-line experience header re-parses title↔company-swapped, so the title/company override lands on the wrong field. Same root as #293's `experience` baseline.

## Regression proof

Stubbing the override fold (passthrough `applyOverrides` + empty pick) makes the gate **fail loudly** — `email "vantreon.sentinel@example.com" missing on re-parse` across fixtures. With the sentinel literals + collision guard, the non-baselined `contact` / `skills` / `added` passes are now genuinely meaningful (no vacuous matches).

## Test plan
- [x] `npm run verify` green — **2292 passed**
- [x] `corpus-roundtrip.test.ts` baseline unchanged by the refactor
- [x] Regression proof re-verified after the recipe fix
- [x] Follow-ups: #487 (Group A), #436 (Group B)

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation, bake, review fixes | Claude Opus 4.8 (1M context) | high |

### DIFF
--- a/src/lib/heuristics/corpus-edit-roundtrip.test.ts
+++ b/src/lib/heuristics/corpus-edit-roundtrip.test.ts
@@ -1,0 +1,431 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+/**
+ * Corpus EDIT-leg round-trip gate (#459).
+ *
+ * The self-consistency gate (`corpus-roundtrip.test.ts`, #293) round-trips each
+ * fixture with NO user edits in the middle — it proves the renderer emits shapes
+ * our parser round-trips, and nothing about override fidelity. This gate closes
+ * that hole: it runs the real Friends & Family loop —
+ *
+ *   drop PDF → parse → user corrects what the parser got wrong → Download PDF
+ *
+ * — across every fixture and asserts the corrections SURVIVE the export:
+ *
+ *   parse1  = runCascade(fixture)
+ *   edits   = synthesizeOverrides(parse1)      // synthetic, deterministic, PII-free
+ *   applied = applyOverrides(parse1.fields, …edits)
+ *   display = { ...parse1, canonical: { …, fields: applied.fields,
+ *                                       fieldConfidence: applied.fieldConfidence } }
+ *   model   = buildAtsResumeModel(display, scoreFor(display),
+ *                                 { contactOverrides, bulletOverrides })
+ *   parse3  = runCascade(renderAtsResumePdf(model))
+ *   assert  = every overridden value is IN parse3, and the value it replaced is NOT.
+ *
+ * The assertion is the inverse of #293's (which asserts parse1 ≡ parse3). A
+ * regression that silently discards a user edit makes parse3 ≡ parse1 — which
+ * #293 calls a PASS and this gate calls a FAILURE. That is the whole point.
+ *
+ * Two production details this reproduces exactly (an AC — a test that gets either
+ * wrong is not acceptable):
+ *   1. `buildAtsResumeModel` takes BOTH the override-applied `display` AND a
+ *      separate `{ contactOverrides, bulletOverrides }` pick (`useDownloadPdf.ts`).
+ *   2. `display.canonical.sections` stays the BASE's, un-edited (#445) — only
+ *      `fields` + `fieldConfidence` carry the edit (`useAnalyzedResume.ts`).
+ *
+ * PII-safe by construction: we assert on strings WE authored (the `NEW_*`
+ * literals), never on fixture values, and never snapshot a fixture field — so the
+ * "snapshots are lossy by design" corpus property holds automatically. The one
+ * "replaced value absent" check reads an original value at runtime and asserts
+ * its ABSENCE; it is never persisted.
+ *
+ * ── Ratchet ── identical discipline to #293 (shared harness): a non-baselined
+ * category that fails → gate fails; a baselined category that passes → gate fails
+ * ("remove it from KNOWN_FAILURES"); a KNOWN_FAILURES key for a missing fixture →
+ * gate fails. Baseline can only shrink. The baseline here is committed AS-FOUND —
+ * no round-trip bug is fixed in this issue; each group is triaged to a follow-up.
+ */
+
+import { readFileSync } from "node:fs";
+
+import { describe, it, expect } from "vitest";
+import { runCascade } from "./cascade.ts";
+import type { CascadeResult } from "./types.ts";
+import { scoreForCascade } from "./roundtrip-hop.ts";
+import type { BulletObservation } from "../score/score.ts";
+import { applyOverrides } from "../edit/apply-overrides.ts";
+import { buildAtsResumeModel } from "../pdf/ats-resume-model.ts";
+import { renderAtsResumePdf } from "../pdf/render-ats-pdf.ts";
+import type {
+  ContactOverrides,
+  ExperienceFieldOverrides,
+  BulletOverrides,
+  SkillsOverride,
+  AddedEntry,
+} from "../../hooks/useEditableParse.ts";
+import {
+  FIXTURE_ROOT,
+  walkPdfs,
+  relKey,
+  assertNoStaleKeys,
+  assertRatchet,
+} from "./corpus-gate.test-utils.ts";
+
+// ── The synthetic override literals ────────────────────────────────────────────
+// Authored strings that MUST NOT occur anywhere in the corpus — a collision makes
+// the presence check vacuous (a plain string like "Kubernetes" or "Northwind
+// Systems" appears verbatim in real fixtures, so the check passes even if the
+// override is discarded). Every literal carries the `Vantreon` sentinel token,
+// and `assertLiteralsAbsent` fails the fixture loudly if any one already appears
+// in the parse — so a future fixture can't silently re-introduce a collision.
+// A synthetic phone (unused — email is the always-settable contact edit) would
+// follow the fixture phone policy: real area code + `555` + `0100`–`0199`.
+const NEW_EMAIL = "vantreon.sentinel@example.com";
+const NEW_TITLE = "Vantreon Platform Engineer";
+const NEW_COMPANY = "Vantreon Systems";
+const NEW_BULLET =
+  "Vantreon: cut p99 checkout latency 43% by resharding the session store.";
+const NEW_SKILL = "VantreonScript";
+const ADDED_DEGREE = "B.S. in Vantreon Systems Engineering";
+const ADDED_INSTITUTION = "Vantreon Institute of Technology";
+
+/** Every override literal, for the per-fixture collision guard. */
+const SENTINELS = [
+  NEW_EMAIL,
+  NEW_TITLE,
+  NEW_COMPANY,
+  NEW_BULLET,
+  NEW_SKILL,
+  ADDED_DEGREE,
+  ADDED_INSTITUTION,
+] as const;
+
+type EditCategory =
+  | "contact"
+  | "experience"
+  | "bullets"
+  | "skills"
+  | "added"
+  | "render";
+
+const CATEGORIES: EditCategory[] = [
+  "contact",
+  "experience",
+  "bullets",
+  "skills",
+  "added",
+  "render",
+];
+
+/**
+ * Per-fixture edit categories currently allowed to fail — BAKED AS-FOUND (#459);
+ * no round-trip bug is fixed in this PR. Shrink as the follow-ups land (the
+ * ratchet forces it). Grouped by likely shared root cause.
+ */
+const KNOWN_FAILURES: Record<string, EditCategory[]> = {
+  // ── Group A · bullet overrides do not survive the Download-PDF round-trip (#487) ──
+  // An overridden bullet does not re-parse back out: the reconstructed role
+  // comes back without the edited bullet (both the edit AND the value it replaced
+  // are gone — the model never carries it). Confirmed NOT first-bullet-specific —
+  // overriding the 2nd observation drops too (reviewer, faithful recipe) — so it
+  // is bullet-override survival, not a leading-bullet off-by-one. Invisible to
+  // #293, which gates bullet COUNT/mapping, not text — the silent edit-loss this
+  // gate exists to catch. Trips on every fixture carrying a bullet, so the
+  // `bullets` category has NO teeth corpus-wide today (baked-as-found, not
+  // coverage) — the teeth return one fixture at a time as #487 shrinks this list.
+  "google-docs/google-docs-skia-proxy-achievements-oneline.pdf": ["bullets"],
+  "google-docs/google-docs-skia-proxy-classic.pdf": ["bullets"],
+  "google-docs/google-docs-skia-proxy-coursework-dup.pdf": ["bullets"],
+  "google-docs/google-docs-skia-proxy-minimal.pdf": ["bullets"],
+  "google-docs/google-docs-skia-proxy-multiline-bullets-coursework.pdf": ["bullets"],
+  "google-docs/google-docs-skia-proxy-nonstandard-headers.pdf": ["bullets"],
+  "google-docs/google-docs-skia-proxy-two-column-long-role-header.pdf": ["bullets"],
+  "google-docs/google-docs-skia-proxy-two-column.pdf": ["bullets"],
+  "latex/awesome-cv-cv.pdf": ["bullets"],
+  "latex/awesome-cv-resume.pdf": ["bullets"],
+  "latex/header-as-name-functional-resume.pdf": ["bullets"],
+  "latex/multi-degree-coursework.pdf": ["bullets"],
+  "unknown/bulleted-labelled-single-column-skills.pdf": ["bullets"],
+  "unknown/chromium-asymmetric-sidebar.pdf": ["bullets"],
+  "unknown/compound-certifications-activities-tail.pdf": ["bullets"],
+  "unknown/label-rail-inline-headers.pdf": ["bullets"],
+  "unknown/mid-dot-header.pdf": ["bullets"],
+  "unknown/name-set-apart-tagline.pdf": ["bullets"],
+  "unknown/pdflib-leading-glyph-skills-header.pdf": ["bullets"],
+  "unknown/single-column-intl-role-headers.pdf": ["bullets"],
+  "unknown/single-column-title-below-anchor.pdf": ["bullets"],
+  "unknown/single-column-year-only-roundtrip.pdf": ["bullets"],
+  "unknown/student-projects-activities-singlecol.pdf": ["bullets"],
+  "unknown/synthetic-dateless-experience.pdf": ["bullets"],
+  "unknown/synthetic-degreeless-two-programs.pdf": ["bullets"],
+  "unknown/two-column-achievements-sidebar.pdf": ["bullets"],
+  "unknown/weasyprint-cairo-classic.pdf": ["bullets"],
+  "unknown/weasyprint-cairo-minimal.pdf": ["bullets"],
+  "unknown/weasyprint-cairo-nonstandard-headers.pdf": ["bullets"],
+  "unknown/weasyprint-cairo-two-column.pdf": ["bullets"],
+  "word/openresume-laverne-word-quartz.pdf": ["bullets"],
+
+  // ── Group B · title/company override lands on a swapped field (#436 — one-line header) ──
+  // The one-line "Title · Company, Location · Team" experience header re-parses
+  // title↔company-swapped or company-truncated (same root as #293's `experience`
+  // baseline, #436), so the title/company OVERRIDE is present but on the wrong
+  // field — the "replaced value" check still sees the old value in the other slot.
+  // Fixtures that also carry a bullet compound Group A. Delete as #436 lands.
+  "latex/deedy-resume-macfonts.pdf": ["experience", "bullets"],
+  "latex/deedy-resume-openfonts.pdf": ["experience", "bullets"],
+  "unknown/chromium-two-column-sidebar.pdf": ["experience", "bullets"],
+  "unknown/letter-spaced-name-heading.pdf": ["experience"],
+  "unknown/openresume-react-pdf.pdf": ["experience", "bullets"],
+  "unknown/shared-employer-banner-roles.pdf": ["experience", "bullets"],
+  "unknown/title-team-next-line-employer.pdf": ["experience", "bullets"],
+  "word/chanchal-sharma-bulleted-skills.pdf": ["experience"],
+  "word/chanchal-sharma-sample.pdf": ["experience"],
+};
+
+interface SyntheticEdits {
+  contact: ContactOverrides;
+  experience: Record<number, ExperienceFieldOverrides>;
+  bullets: BulletOverrides;
+  skills: SkillsOverride;
+  addedEntries: AddedEntry[];
+  /** The kinds this fixture could structurally exercise. */
+  exercised: Set<Exclude<EditCategory, "render">>;
+  /** Original values the overrides replaced — for the "replaced value absent"
+   *  assertion (read at runtime, never persisted). */
+  replaced: { email?: string; title?: string; company?: string; bullet?: string };
+}
+
+/**
+ * Deterministic per-fixture override set, one edit of each kind the UI offers.
+ * Derived from `parse1` so it applies to any fixture; each kind is skipped only
+ * when the fixture structurally can't exercise it.
+ */
+function synthesizeOverrides(
+  parse1: CascadeResult,
+  observations: readonly BulletObservation[],
+): SyntheticEdits {
+  const fields = parse1.canonical.fields;
+  const exercised = new Set<Exclude<EditCategory, "render">>();
+  const replaced: SyntheticEdits["replaced"] = {};
+
+  // contact — always settable.
+  exercised.add("contact");
+  replaced.email = (fields as { email?: string }).email;
+
+  // experience — only when a role 0 exists.
+  const experience: Record<number, ExperienceFieldOverrides> = {};
+  if (fields.experience.length > 0) {
+    exercised.add("experience");
+    replaced.title = fields.experience[0].title;
+    replaced.company = fields.experience[0].company;
+    experience[0] = { title: NEW_TITLE, company: NEW_COMPANY };
+  }
+
+  // bullets — only when a bullet observation exists.
+  const bullets: BulletOverrides = {};
+  if (observations.length > 0) {
+    exercised.add("bullets");
+    replaced.bullet = observations[0].text;
+    bullets[observations[0].index] = NEW_BULLET;
+  }
+
+  // skills — always addable.
+  exercised.add("skills");
+
+  // addedEntries — always: append one education entry.
+  exercised.add("added");
+  const addedEntries: AddedEntry[] = [
+    {
+      id: "added:edu:0",
+      section: "education",
+      title: ADDED_DEGREE,
+      subtitle: ADDED_INSTITUTION,
+      year: "2020",
+    },
+  ];
+
+  return {
+    contact: { email: NEW_EMAIL },
+    experience,
+    bullets,
+    skills: { removed: [], added: [NEW_SKILL] },
+    addedEntries,
+    exercised,
+    replaced,
+  };
+}
+
+/** JSON view of a slice of the parsed fields — the searchable text the presence /
+ *  absence assertions run against (we only ever search for our own literals or a
+ *  runtime-read original, never snapshot a value). */
+function j(value: unknown): string {
+  return JSON.stringify(value ?? null);
+}
+
+function emptyFails(): Record<EditCategory, string[]> {
+  return { contact: [], experience: [], bullets: [], skills: [], added: [], render: [] };
+}
+
+/** The gate's one assertion shape: the override value must be present in the
+ *  re-parse, and the value it replaced (when there is one) must be absent.
+ *  Appends a description to `into` for each half that fails. */
+function presenceCheck(
+  text: string,
+  present: string,
+  replaced: string | undefined,
+  label: string,
+  into: string[],
+): void {
+  if (!text.includes(present)) into.push(`${label} "${present}" missing on re-parse`);
+  if (replaced && text.includes(replaced))
+    into.push(`replaced ${label} still present on re-parse`);
+}
+
+/**
+ * Per-category failures for a successful re-parse: for each exercised kind, the
+ * overridden value must be present and the value it replaced absent. Pure over
+ * `(p3, edits)` so the branchy assertion logic is out of the test body and unit-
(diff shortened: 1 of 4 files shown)


src/lib/heuristics/corpus-edit-roundtrip.test.ts
```diff
@@ -0,0 +1,392 @@
...
+      p1.rawText,
+      p1.canonical.sections,
+      edits.contact,
+      edits.experience,
+      edits.bullets,
+      observations,
+      {}, // education field overrides — none; we ADD an entry instead
+      edits.skills,
+      edits.addedEntries,
+      {}, // addedBullets
```
s-annam, line 350: **Blocking — the production recipe is not reproduced here.**

This `applyOverrides` call passes **11** positional args. Production (`useAnalyzedResume.ts:149`) passes **15**, including `base.canonical.fieldConfidence` at position 14. Omitted, it defaults to `{}` → every non-edited field's confidence is `0` → `detected` is false → `gated: true` (`src/lib/contact.ts:103-115`) → `buildContact`'s `valueFor` returns `""`.

Reproduced on `google-docs-skia-proxy-classic.pdf` — the contact this gate renders vs. what production renders:

```
TEST (fieldConfidence = {}):   {"name":"Jane Smith","email":"jordan.doe@example.com","links":[]}
PROD (fieldConfidence = base): {"name":"Jane Smith","email":"…","phone":"(312) 555-0123",
                                "location":"Springfield, IL","links":["linkedin.com/in/janesmith"]}
```

So the gate exports a PDF with **no phone, no location, no links** — one production never produces. It stays green only because the overridden field (email) is the one `applyOverrides` bumps to confidence 1. The whole baseline is therefore baked against an unfaithful export, and any future regression in contact-confidence gating on the export path is invisible to this gate forever. `scoreForCascade(display)` reads `fieldConfidence` too (`score.ts:815`), so the model is built against a wrong score as well.

```suggestion
      {}, // addedBullets
      new Set<number>(), // removedBullets
      [], // profileOverrides
      p1.canonical.fieldConfidence, // the base per-field confidence — production passes this
```

Then re-bake `KNOWN_FAILURES` against the faithful recipe.

src/lib/heuristics/corpus-edit-roundtrip.test.ts
```diff
@@ -0,0 +1,392 @@
...
+// ── The synthetic override literals ────────────────────────────────────────────
+// Authored strings, distinct from any fixture value. A synthetic phone (unused
+// here — email is the always-settable contact edit) would follow the fixture
+// phone policy: real area code + `555` exchange + `0100`–`0199` subscriber.
+const NEW_EMAIL = "jordan.doe@example.com";
+const NEW_TITLE = "Staff Platform Engineer";
+const NEW_COMPANY = "Northwind Systems";
+const NEW_BULLET =
+  "Cut p99 checkout latency 43% by resharding the session store.";
+const NEW_SKILL = "Kubernetes";
```
s-annam: **Blocking — these literals collide with real fixture content, so two checks are vacuous and the ratchet jams.**

`NEW_COMPANY = "Northwind Systems"` is **literally role 0's company** on two fixtures:

| Fixture | parsed `role0.company` |
|---|---|
| `unknown/bulleted-labelled-single-column-skills.pdf` | `"Northwind Systems"` |
| `unknown/chromium-asymmetric-sidebar.pdf` | `"Northwind Systems"` |
| `unknown/chromium-two-column-sidebar.pdf` | `"Northwind Systems - Customer Experience Engineering"` |

`synthesizeOverrides` therefore sets `company` to the value it already has, and `presenceCheck` asserts that same string is both **present** (L258) and **absent** (L259). That's an unconditional failure with nothing to do with #436 — so those `experience` baseline entries can **never** be removed, the ratchet's stale-entry tooth is permanently jammed, and a genuine experience-override regression on those fixtures is masked forever.

`NEW_SKILL = "Kubernetes"` already appears in the text of **15 of 46** fixtures (and in the parsed `skills` array of `…skia-proxy-classic`, `bulleted-labelled-single-column-skills`, `chromium-asymmetric-sidebar`). On those, this check passes **even if the skills override is discarded entirely** — which falsifies the PR's own regression proof for `skills` on a third of the corpus.

Pick literals that can't occur in a résumé corpus, and make a future collision fail loudly rather than silently void the check:

```ts
// Every NEW_* literal must be absent from parse1, or the presence check is vacuous.
for (const lit of [NEW_EMAIL, NEW_TITLE, NEW_COMPANY, NEW_BULLET, NEW_SKILL, ADDED_INSTITUTION])
  expect(j(p1.canonical.fields), `${rel}: literal "${lit}" already in the fixture`)
    .not.toContain(lit);
```

src/lib/heuristics/corpus-edit-roundtrip.test.ts
```diff
@@ -0,0 +1,392 @@
...
+ * ratchet forces it). Grouped by likely shared root cause.
+ */
+const KNOWN_FAILURES: Record<string, EditCategory[]> = {
+  // ── Group A · first bullet of a role dropped on Download-PDF round-trip (#487) ──
+  // The overridden FIRST bullet observation does not survive parse → export →
+  // re-parse: the reconstructed role re-parses with its leading bullet missing
+  // (both the edit AND the value it replaced are gone — the model never carries
+  // it). Invisible to #293, which gates bullet COUNT/mapping, not text — exactly
+  // the silent edit-loss this gate exists to catch. Trips on every fixture that
+  // carries an experience bullet. Delete each line as #487 lands.
```
s-annam: **Secondary — the `bullets` category has zero teeth today, and this root-cause note looks wrong.**

Every fixture carrying a bullet observation is baselined on `bullets` (38 entries); the rest have no bullets, so the check is skipped. No fixture exercises it *and* passes — the category protects nothing right now. Defensible as baked-as-found, but worth stating plainly rather than presenting as coverage.

More importantly: this comment (and #487) say the **first** bullet of a role is dropped. I overrode `observations[1]` instead of `observations[0]` on `skia-proxy-classic` (8 obs), `awesome-cv-resume` (30 obs) and `weasyprint-cairo-classic` (8 obs), under the faithful recipe — **the second bullet's override doesn't survive either**. So the defect isn't first-bullet-specific; bullet overrides appear not to survive the export at all. Worth correcting here and in #487 before someone hunts for an off-by-one in the leading-bullet path.

(Good news, and I checked it: Group A is a **real** bug, not an artifact of this harness — the drop reproduces under the faithful recipe too.)

src/lib/heuristics/corpus-edit-roundtrip.test.ts
```diff
@@ -0,0 +1,392 @@
...
+    const rel = relKey(fixture);
+    it(`edit round-trips: ${rel}`, async () => {
+      const p1 = await runCascade(new Uint8Array(readFileSync(fixture)));
+      const score1 = scoreForCascade(p1);
+      const observations = score1.bullets ?? [];
+      const edits = synthesizeOverrides(p1, observations);
+
+      // A fixture that ends up editing nothing is itself a failure — the always-on
+      // kinds (contact/skills/added) must always be present in the exercise set.
+      expect(
```
s-annam: **Nit — this guard is trivially true.** `exercised` gets `contact`, `skills` and `added` added unconditionally in `synthesizeOverrides`, so `.size` is always ≥ 3 and `toBeGreaterThan(0)` can never fire. If the intent is "the always-on kinds are always exercised", assert that:

```ts
expect([...edits.exercised]).toEqual(
  expect.arrayContaining(["contact", "skills", "added"]),
);
```

s-annam (changes requested):
The idea is right and the gate is a real gap-closer — the inverse-of-#293 framing is
correct, the ratchet is shared rather than cloned, and I verified Group A is a **genuine
bug, not an artifact of the harness** (the bullet drop reproduces under a faithful
recipe too). Both corpus gates are green as claimed.

But the gate does not reproduce the production Download-PDF recipe — which is the PR's
own headline AC — and two of the override literals collide with the corpus, which makes
one category vacuous and jams the ratchet permanently. Everything below I reproduced by
execution, not by reading the diff.

## Blocking

**1. Model trailer in the commit message + badge in the PR body (`CLAUDE.md` → Hard rules).**
`42a6c2a` ends with `Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>`,
and this PR body carries the `🤖 Generated with [Claude Code]` badge. Both are forbidden —
"not in a commit message, not in a PR body". `main` merges through a merge queue that
derives the squash message from the branch, so the trailer lands in `main` **permanently**.
Fix: `git commit --amend` to strip the trailer, drop the badge from the body, and move the
model attribution into a `## Provenance` block (its only legal home).

**2. The production recipe is not reproduced — `fieldConfidence` is dropped.**
`corpus-edit-roundtrip.test.ts:317` calls `applyOverrides` with **11** positional args.
Production (`useAnalyzedResume.ts:149`) passes **15**, including
`base.canonical.fieldConfidence` at position 14. Omitted, it defaults to `{}`, so every
non-edited field's confidence is `0` → `detected` is false → `gated: true`
(`src/lib/contact.ts:103-115`) → `buildContact`'s `valueFor` returns `""`.

Reproduced on `google-docs/google-docs-skia-proxy-classic.pdf` — the model this gate
renders vs. the one production renders:

```
TEST recipe (11 args, fieldConfidence = {}):
  {"name":"Jane Smith","email":"jordan.doe@example.com","links":[],"linkHrefs":[]}
PROD recipe (fieldConfidence = base.canonical.fieldConfidence):
  {"name":"Jane Smith","email":"jordan.doe@example.com","phone":"(312) 555-0123",
   "location":"Springfield, IL","links":["linkedin.com/in/janesmith"], …}
```

The gate exports a PDF with **no phone, no location, no links** — one production never
produces. It stays green only because the *overridden* field (email) is the one
`applyOverrides` bumps to confidence 1. Consequences: the whole baseline is baked against
an unfaithful export, and any future regression in contact-confidence gating on the export
path is invisible to this gate forever, because it already drops those fields. `score` is
affected too — `scoreForCascade(display)` reads `fieldConfidence` (`score.ts:815`).

Fix — thread the base confidence through, exactly as `useAnalyzedResume` does:

```ts
      edits.skills,
      edits.addedEntries,
      {},                          // addedBullets
      new Set<number>(),           // removedBullets
      [],                          // profileOverrides
      p1.canonical.fieldConfidence, // ← the base per-field confidence (prod passes this)
    );
```

Then re-bake `KNOWN_FAILURES` against the faithful recipe.

**3. The override literals collide with real fixture content.**
`NEW_COMPANY = "Northwind Systems"` is **literally role 0's company** on two fixtures:

| Fixture | parsed `role0.company` |
|---|---|
| `unknown/bulleted-labelled-single-column-skills.pdf` | `"Northwind Systems"` |
| `unknown/chromium-asymmetric-sidebar.pdf` | `"Northwind Systems"` |
| `unknown/chromium-two-column-sidebar.pdf` | `"Northwind Systems - Customer Experience Engineering"` |

So `synthesizeOverrides` sets `company` to the value it already has, and then
`presenceCheck(expText, NEW_COMPANY, replaced.company, …)` asserts that same string is
both **present** (line 258) and **absent** (line 259). That is an unconditional failure
with nothing to do with #436 — so the `experience` entries for those fixtures can **never**
be removed from `KNOWN_FAILURES`, the ratchet's stale-entry tooth is permanently jammed,
and any genuine experience-override regression on them is masked forever.

`NEW_SKILL = "Kubernetes"` already appears in the text of **15 of 46** fixtures (and in the
parsed `skills` array of at least `…skia-proxy-classic`, `bulleted-labelled-single-column-skills`,
`chromium-asymmetric-sidebar`). On those, `presenceCheck(skillsText, NEW_SKILL, …)` passes
**even if the skills override is discarded entirely** — which directly falsifies the PR's
regression proof ("Non-baselined `contact`/`skills`/`added` pass on every fixture … is what
proves the gate isn't vacuously green"). For `skills` it proves nothing on a third of the corpus.

Fix: pick literals that cannot occur in a résumé corpus, and — more durably — make a
collision **fail loudly** instead of silently voiding the check, so a future fixture can't
re-introduce this:

```ts
// Every NEW_* literal must be absent from parse1, or the presence check is vacuous.
for (const lit of [NEW_EMAIL, NEW_TITLE, NEW_COMPANY, NEW_BULLET, NEW_SKILL, ADDED_INSTITUTION])
  expect(j(p1.canonical.fields), `${rel}: literal "${lit}" already in the fixture`)
    .not.toContain(lit);
```

## Secondary

**4. The `bullets` category has zero teeth corpus-wide, and its root-cause note is wrong.**
Every fixture carrying a bullet observation is baselined on `bullets` (38 entries); the rest
have no bullets, so the check is skipped. No fixture exercises it and passes — the category
protects nothing today. That's defensible as baked-as-found, but it should be stated plainly
rather than presented as coverage.

More importantly, the Group A comment (lines 111-117) and #487 both say the **first** bullet
of a role is dropped. I overrode `observations[1]` instead of `observations[0]` on
`skia-proxy-classic` (8 obs), `awesome-cv-resume` (30 obs) and `weasyprint-cairo-classic`
(8 obs) under the faithful recipe — **the second bullet's override doesn't survive either**.
So the defect isn't first-bullet-specific; bullet overrides appear not to survive the export
at all. Worth correcting in the comment and in #487 before someone goes hunting for an
off-by-one in the leading-bullet path.

## Nits

**5. `corpus-edit-roundtrip.test.ts:369` — the guard is trivially true.** `exercised` gets
`contact`, `skills` and `added` added unconditionally, so `.size` is always ≥ 3 and
`toBeGreaterThan(0)` can never fire. If the intent is "the always-on kinds are always
exercised", assert that: `expect([...edits.exercised]).toEqual(expect.arrayContaining(["contact","skills","added"]))`.

## Gates

`typecheck` ✅ · `lint` ✅ · both corpus gates ✅ (96 passed) · design-system / style-token
gates **n/a** (no `src/components/**` changes) · fixture-PII gate **n/a** (0 fixtures touched;
note `npm run check:fixtures` is documented in `CLAUDE.md` but no such script exists in
`package.json` — a pre-existing gap, not this PR's).

---

Samhit21:
Thanks — all five addressed in `158144a` (amended + force-pushed, per the trailer point). Re-verified by execution, not reading.

**1. Model trailer + badge (blocking).** `git commit --amend` stripped the `Co-Authored-By` trailer from history (the merge queue would have baked it into `main`), dropped the badge from the PR body, and moved attribution into a `## Provenance` block.

**2. `fieldConfidence` dropped — unfaithful recipe (blocking).** Now threads `p1.canonical.fieldConfidence` (plus the `removedBullets` / `profileOverrides` positional args) into `applyOverrides`, matching `useAnalyzedResume`. Confirmed the export now carries phone/location/links. **Re-baked `KNOWN_FAILURES` against the faithful recipe.**

**3. Colliding literals (blocking).** Switched every `NEW_*` to a `Vantreon`-sentinel literal that can't occur in a corpus, and added a per-fixture guard that fails loudly if any literal is already present — so a check can never be silently vacuous. As you predicted, the re-bake **dropped the two collision-jammed `experience` entries** (`bulleted-labelled-single-column-skills`, `chromium-asymmetric-sidebar`) — they only failed because `NEW_COMPANY` was their real company; now their `experience` genuinely passes.

**4. `bullets` teeth + wrong root cause (secondary).** Corrected the Group A comment and #487: it's **bullet-override survival on the export path**, not first-bullet-specific (verified the 2nd observation drops too). Stated plainly that `bullets` has **no teeth corpus-wide today**, not presented as coverage.

**5. Trivially-true guard (nit).** Replaced `.size > 0` with `expect([...exercised]).toEqual(expect.arrayContaining(["contact","skills","added"]))`.

Regression proof re-run with the new literals — still fails loudly when the override fold is stubbed. `npm run verify` green (2292 passed). Ready for another look.

Samhit21:
CI fix: `verify` failed because `main` gained **5 new fixtures** since I branched (46 → 51), and CI tests the branch merged with main — so three of them (`compound-certifications-activities-tail`, `single-column-intl-role-headers`, `title-team-next-line-employer`) were un-baselined and tripped the ratchet (my local run only had the older 46). Rebased onto latest main and **re-baked the baseline over all 51 fixtures** — the three land in the same Group A/B root causes (#487 / #436), the other two pass clean (real teeth). `npm run verify` green (2396 passed); both corpus gates green (106). Re-running CI now.

src/lib/heuristics/corpus-edit-roundtrip.test.ts
```diff
@@ -0,0 +1,431 @@
...
+  assertNoStaleKeys,
+  assertRatchet,
+} from "./corpus-gate.test-utils.ts";
+
+// ── The synthetic override literals ────────────────────────────────────────────
+// Authored strings that MUST NOT occur anywhere in the corpus — a collision makes
+// the presence check vacuous (a plain string like "Kubernetes" or "Northwind
+// Systems" appears verbatim in real fixtures, so the check passes even if the
+// override is discarded). Every literal carries the `Vantreon` sentinel token,
+// and `assertLiteralsAbsent` fails the fixture loudly if any one already appears
```
s-annam, line 80: Nit (comment rot): the docblock names `assertLiteralsAbsent`, but no such function exists — the collision guard is the inline loop at L401-405. Rename the reference or drop it.

s-annam (approved):
Solid, well-reasoned gate. Verified locally on `92a3ec2`: typecheck clean, both corpus suites green (106 passed / 1 skipped), fallow warn-only with **no unused-export flag** — the lifted exports have two live consumers each, exactly as the docblock claims. The edit-leg gate has genuine teeth (contact/skills/added are exercised and non-baselined across all 53 fixtures), so it really does assert user edits survive Download-PDF, not just pass vacuously. The `Vantreon` sentinel + collision guard is the right call. No blocking issues.

## Nits (non-blocking)
- **Consolidation is partial** — fallow flags `walkPdfs` still cloned in `projections.test.ts`, `export-layout-contract.test.ts`, `corpus.test.ts`, and `to-json-resume.export-projection.test.ts`. Out of #459 scope, but now that `corpus-gate.test-utils.ts` owns `FIXTURE_ROOT` + `walkPdfs`, a follow-up could delete those four copies too.

---

PR #494, opened 2026-07-15
feat(edit): make prose-body project descriptions editable in-place (#489)

### LINKED ISSUE #489: [ui] make prose-body project descriptions editable in-place
## Why

Follow-up from PR #483 review. That PR replaced the "No bullet-shaped lines detected." placeholder for prose-body projects with the rendered description paragraph (#464 UI step). It shows the description but renders it as a read-only `<p>` — the neighboring bulleted branch uses the editable `ResumeBulletRow` component.

Result: a prose-body project's description is visible but has no input path, while the same project with `•` bullets is fully editable. Inconsistent edit surface in the lane whose whole premise is fix-in-place.

## Where

[src/components/features/ReconstructedResume.tsx:627-635](src/components/features/ReconstructedResume.tsx#L627-L635) — the `!added && project.description` branch.

Bulleted branch just above uses `ResumeBulletRow` with `override={bulletOverrides?.[b.index]}` for edit wiring.

There is currently **no override path for this branch at all** — it is greenfield, not a rewire of an existing path:

- `resolveParsedDescriptionTarget` (`src/lib/edit/apply-overrides.ts:642-655`) only appends *added-entry* bullets; it doesn't touch a parsed project's prose `description`.
- `onEntryField` is added-entry-only, keyed by the synthetic `added.id` — it does not apply to the `!added && project.description` branch this issue targets.

## Fixture

[tests/fixtures/pdfs/unknown/single-column-projects-prose-body.pdf](tests/fixtures/pdfs/unknown/single-column-projects-prose-body.pdf) already exercises this branch. Its `Projects` section carries two prose-body entries — `Ridgemont Resume Studio` and `Ledger Ingest Toolkit` — each a two-sentence paragraph with **no `•` bullets**, so the parser fills `project.description` and the graded-bullet grouping yields zero bullets: exactly the `!added && project.description` render path. Use it for the round-trip acceptance check below. (Added in #483; synthetic persona, no new fixture needed.)

## Proposed change

Route the prose description through an `EditableField` (`multiline`), keeping the paragraph render (not `ResumeBulletRow`, which is bullet-shaped and not the right primitive for a paragraph).

Wire the description through a new `descriptionOverrides` map — parallel to, not a reuse of, `bulletOverrides`. `bulletOverrides` is keyed by the resume-wide graded-bullet-pool index, which structurally cannot key a single prose paragraph without inventing a key space; a parallel map is the clean route. This requires new plumbing across three files:

- `src/hooks/useEditableParse.ts` — state + setter for `descriptionOverrides`.
- `src/lib/edit/apply-overrides.ts` — apply a `descriptionOverrides` entry onto the matching parsed project's `description` at render/export time.
- `src/components/features/ReconstructedResume.tsx` — swap the read-only `<p>` for `EditableField` (`multiline`), wired to `descriptionOverrides`.

The commit should:

- Wire the description into the new `descriptionOverrides` map so an edit commits back to the resume model.
- Keep the paragraph render style (not a bulleted list), matching what the parser produced.
- Preserve the existing edit-round-trip guarantees (`corpus-roundtrip.test.ts` — the exported PDF must re-parse to the same fields).

## Acceptance

- [ ] Click on the description text of a prose-body project opens an edit surface.
- [ ] Committing an edit updates the model and survives a Download PDF round-trip — verified against `single-column-projects-prose-body.pdf` (see Fixture).
- [ ] The bulleted-project edit path is unchanged (regression check).
- [ ] `npm run verify` green.

## Out of scope

- Splitting a prose description into per-sentence "bullets" — the paragraph shape stays.

## Parent

Follow-up to #464 (closed by #483). Not a blocker for #460, filed to keep the edit-surface consistency contract in the roadmap.

s-annam:
## Summary

The prose-body project branch (`!added && project.description` in `ReconstructedResume.tsx`) rendered its description as a read-only `<p>` while the sibling bulleted branch was fully editable — an inconsistent edit surface in the fix-in-place lane. This routes the description through `EditableField` (multiline, paragraph render preserved) backed by a new `descriptionOverrides` map.

- **New `descriptionOverrides` channel** — `Record<string,string>`, keyed `parsedEntryKey("projects", i)`, threaded parallel to the existing override maps: state + setter in `useEditableParse.ts` → applied via new `applyDescriptionOverrides` (16th positional arg on `applyOverrides`) in `apply-overrides.ts` → consumed in `ReconstructedResume.tsx`. Wired into snapshot/replay/resetAll/hasEdits with `?? {}` back-compat for pre-#489 drafts.
- **JD-fit handoff** — `useJdFitResume.ts` passes the new map through too, so a prose-description edit survives the `/` → `/jd-fit` handoff (gap caught by the adversarial review pass).
- **Round-trip** — an edit commits back to the model and survives a Download PDF round-trip, verified by the new `description-override-roundtrip.repro.test.ts` against `single-column-projects-prose-body.pdf`. The bulleted-project edit path is unchanged.

Closes #489

## Test plan

- [x] Affected suites green: `apply-overrides.test.ts` (59), `description-override-roundtrip.repro.test.ts` (1), `corpus-edit-roundtrip.test.ts` (53), `corpus-roundtrip.test.ts`, `jd-fit-handoff.test.ts`, `useJdFitResume.test.tsx` — 183 passed / 1 skipped
- [x] `npm run lint` clean
- [x] Full `npm run verify` — green in CI

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #489 | Claude Opus 4.8 | ultra |
| Adversarial review | Claude Opus 4.8 | ultra |
| Review fixes | Claude Opus 4.8 (1M context) | ultra |
| Orchestration + PR | Claude Opus 4.8 (1M context) | ultra |

## Adversarial review

1 round, 1 blocking finding → fixed → re-verified. Converged (no 2nd round needed: the fix was a scoped arg-threading mirror of the known-correct primary caller, validated by a new regression test).

**Blocking (fixed)**
- **JD-fit handoff dropped prose-description edits** — `src/jd-fit/useJdFitResume.ts` (the only other non-test `applyOverrides` caller) fell through to the `descriptionOverrides = {}` default while `replay()` now restores that map into edit state. A prose edit made on `/` would show in edit state but not apply after handoff to `/jd-fit`. Fixed by threading `handoffEdit.descriptionOverrides` as the 16th arg + adding it to the memo deps; covered by a new handoff-survival assertion.

**Nits (not blocking — noted for the human reviewer)**
- Committing an empty description clears the override → the branch falls to the "No bullet-shaped lines detected" fallback with no re-add affordance (consistent with other `EditableField` clears; the `undefined`-reverts path on `setDescriptionField` is unreachable from this UI).
- `resolveParsedDescriptionTarget` non-null asserts `nextParsed.projects![index]` for a `projects:i` key — unreachable today (override only settable from a rendered parsed project), mirrors the pre-existing added-bullet pattern.
- Prose-description edits don't move the anonymous score (pre-existing scorer limitation: the anon pool draws from `sections`/`rawText`, not `fields`) — issue scopes prose-vs-bullet score handling out.

**Fallow:** the snapshot/replay clone-group duplication in `useEditableParse.ts` is the established per-override-map house pattern (not worsened). `applyExperienceHeaderOverrides` complexity is pre-existing, untouched by #489.

### DIFF
--- a/src/components/features/ReconstructedResume.tsx
+++ b/src/components/features/ReconstructedResume.tsx
@@ -71,6 +71,7 @@ import type {
   EditableParse,
   ExperienceFieldOverrides,
   BulletOverrides,
+  DescriptionOverrides,
   AddedEntry,
   AddedEntryField,
   AchievementFieldOverrides,
@@ -566,11 +567,13 @@ function ProjectsSection({
   projects,
   groups,
   bulletOverrides,
+  descriptionOverrides,
   addedProjects,
   originalCount,
   onAddEntry,
   onRemoveEntry,
   onEntryField,
+  onDescriptionField,
   onAddBullet,
 }: {
   /** Verbatim source heading (#285); falls back to "Projects" when absent. */
@@ -579,11 +582,16 @@ function ProjectsSection({
   /** Pre-built project groups, index-aligned with `projects`. */
   groups: BulletGroup[];
   bulletOverrides: BulletOverrides;
+  /** Prose-description edits keyed by parsedEntryKey (#489) — read only to keep
+   *  a CLEARED prose field mounted (so the clear is reversible in-session). */
+  descriptionOverrides: DescriptionOverrides;
   addedProjects: AddedEntry[];
   originalCount: number;
   onAddEntry: () => void;
   onRemoveEntry: (id: string) => void;
   onEntryField: (id: string, field: AddedEntryField, value: string) => void;
+  /** Commit an edit to a parsed project's prose description (#489). */
+  onDescriptionField: (key: string, value: string | undefined) => void;
   onAddBullet: (entryKey: string, text: string) => void;
 }) {
   return (
@@ -594,6 +602,12 @@ function ProjectsSection({
           const group = groups[i];
           const added =
             i >= originalCount ? addedProjects[i - originalCount] : undefined;
+          const descKey = parsedEntryKey("projects", i);
+          // Keep the prose field mounted once the user has touched it, even after
+          // an authoritative clear ("" override) drops `project.description` — so
+          // the clear stays reversible in-session instead of collapsing to the
+          // read-only "no bullets" hint with no way back (#489 review).
+          const hasDescriptionEdit = descriptionOverrides[descKey] !== undefined;
           const header = [project.name, buildProjectDates(project)]
             .filter(Boolean)
             .join(" · ");
@@ -632,15 +646,26 @@ function ProjectsSection({
                     />
                   ))}
                 </ul>
-              ) : !added && project.description ? (
+              ) : !added && (project.description || hasDescriptionEdit) ? (
                 // #464 — a prose-body project (no `•` bullets, description is
-                // one or more paragraph sentences) surfaces the description
-                // as a paragraph. Without this, the fallback below reads "No
-                // bullet-shaped lines detected" even though the parser DID
-                // extract the description via looksLikeBodyParagraph.
-                <p className="text-sm text-content-primary whitespace-pre-wrap">
-                  {project.description}
-                </p>
+                // one or more paragraph sentences) surfaces the description as a
+                // paragraph. #489 makes that paragraph editable in place: an
+                // EditableField (multiline) keyed by the project's parsedEntryKey
+                // commits back through `descriptionOverrides`, keeping the same
+                // paragraph render style (NOT a bulleted list — the parser
+                // produced prose). The read-only `<p>` (#483) had no input path
+                // while a `•`-bulleted project was fully editable.
+                <EditableField
+                  value={project.description}
+                  emptyAffordance="plain"
+                  placeholder="description"
+                  label="Project description"
+                  textSize="sm"
+                  display="inline"
+                  multiline
+                  className="whitespace-pre-wrap"
+                  onCommit={(v) => onDescriptionField(descKey, v)}
+                />
               ) : (
                 !added && (
                   <p className="text-sm text-content-tertiary">
@@ -959,6 +984,8 @@ export function ReconstructedResume({
     setExperienceField,
     bulletOverrides,
     setBulletField,
+    descriptionOverrides,
+    setDescriptionField,
     removeBullet,
     educationOverrides,
     setEducationField,
@@ -1188,11 +1215,13 @@ export function ReconstructedResume({
         projects={projects}
         groups={projectGroups}
         bulletOverrides={bulletOverrides}
+        descriptionOverrides={descriptionOverrides}
         addedProjects={addedProjects}
         originalCount={originalProjCount}
         onAddEntry={() => addEntry("projects")}
         onRemoveEntry={removeEntry}
         onEntryField={setEntryField}
+        onDescriptionField={setDescriptionField}
         onAddBullet={addBullet}
       />
       {!achievementsAbove && achievementsSection}
--- a/src/lib/edit/apply-overrides.ts
+++ b/src/lib/edit/apply-overrides.ts
@@ -51,6 +51,7 @@ import type {
   AddedBullets,
   ProfileOverride,
   BulletOverrides,
+  DescriptionOverrides,
 } from "../../hooks/useEditableParse.ts";
 
 /**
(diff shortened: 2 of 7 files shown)


src/components/features/ReconstructedResume.tsx
```diff
@@ -634,13 +637,26 @@ function ProjectsSection({
...
+                // one or more paragraph sentences) surfaces the description as a
+                // paragraph. #489 makes that paragraph editable in place: an
+                // EditableField (multiline) keyed by the project's parsedEntryKey
+                // commits back through `descriptionOverrides`, keeping the same
+                // paragraph render style (NOT a bulleted list — the parser
+                // produced prose). The read-only `<p>` (#483) had no input path
+                // while a `•`-bulleted project was fully editable.
+                <EditableField
+                  value={project.description}
+                  emptyAffordance="plain"
```
s-annam, line 660: **Secondary (non-blocking)** — Clearing this field is a one-way trap. `EditableField` commits `draft.trim()`, so an empty clear calls `onDescriptionField(key, "")` → `applyDescriptionOverrides` sets `description = "" || undefined` → the parent branch at line 638 (`!added && project.description`) goes falsy and renders "No bullet-shaped lines detected" instead. The `EditableField` unmounts, and a parsed (non-`added`) project has no `InlineBulletAdd`, so the user can't re-enter prose except via a full `resetAll`.

Side effect: `emptyAffordance="plain"` and `placeholder="description"` are effectively dead — the parent guards on `project.description` before the empty state can ever render.

Consider keeping the field mounted when a `descriptionOverrides` key exists for the entry (even if cleared), or gating the branch on "has a projects entry" rather than "has description text".

src/lib/edit/apply-overrides.ts
```diff
@@ -345,6 +346,46 @@ function mergeAchievementFields(
...
+ * {@link applyAchievementOverrides}. `experience` entries are already cloned by
+ * the entry point, so mutating one in place is safe. A later re-clone by
+ * {@link applyAddedEntriesAndBullets} preserves these edits.
+ */
+function applyDescriptionOverrides(
+  nextParsed: HeuristicParsedResume,
+  overrides: DescriptionOverrides,
+): void {
+  if (Object.keys(overrides).length === 0) return;
+  if (nextParsed.projects) {
```
s-annam, line 374: **Nit / hardening** — `resolveParsedDescriptionTarget` derefs with non-null assertions (`nextParsed.projects![index]`, `heuristic_achievements![index]`). This function only clones those arrays when present, so a `projects:i`/`achievements:i` override against a parse where the array is *absent* would throw `TypeError` instead of no-op'ing.

Not reachable today (per-resume `resetKey` clears overrides, and handoff/drafts always carry the array when a key exists), but the sibling `applyAddedBulletsToExistingEntries` neutralizes exactly this by forcing `projects`/`heuristic_achievements` to `[]` first. Cheap to match here — matters the moment the achievements surface the docblock anticipates gets wired.

s-annam (commented):
`feat(edit): make prose-body project descriptions editable in-place (#489)` — clean, well-scoped. The `descriptionOverrides` channel threads correctly through all four consumers (`useEditableParse` state/snapshot/replay/hasEdits → `useAnalyzedResume` → `applyOverrides` → the `/`→`/jd-fit` handoff). Verified the index alignment (`projectDisplay` returns `canonical.fields` verbatim, so the UI's `i` matches the apply-time `parsed.projects` index) and the production Download-PDF path (`result` is the override-folded analyzed resume, so `buildAtsResumeModel` reads the edited description off the model — the round-trip test's `display` mirrors it). The #489 test proves the previously-dropped handoff arg is now carried.

**Gates:** `check:fixtures` ✓ (fixture is pre-existing, not touched) · lint ✓ · typecheck ✓ · fallow ✓ (complexity findings are inherited debt) · #489 roundtrip + jd-fit tests ✓ (8 passed). Nothing blocking.

## Secondary

- Clearing a prose description is a one-way trap in-session (see inline).

## Nits

- `applyDescriptionOverrides` array-guard is less defensive than its sibling (see inline).
- `ProjectsSection` grew ~14 LOC (fallow CRAP 49.5); inherited debt on an already-flagged file (`ReconstructedResume.tsx` is on the CLAUDE.md 1184-LOC debt list). Prefer extraction over further growth next time — not blocking here.

Verdict rule: 0 Blocking → APPROVE.

---

src/components/features/ReconstructedResume.tsx
s-annam: Fixed in `d0ff33b`. The prose field now stays mounted while a `descriptionOverrides` key exists for the entry, so an authoritative clear ("") is reversible in-session instead of collapsing to the read-only "no bullets" hint. `descriptionOverrides` is threaded into `ProjectsSection` read-only for this gate; `EditableField`'s read view is a clickable `role="button"` even when empty+plain, so re-entry works.

src/lib/edit/apply-overrides.ts
s-annam: Fixed in `d0ff33b`. `resolveParsedDescriptionTarget` now resolves the projects/achievements arrays with `?.` instead of a non-null assertion, so a stale override keyed against an array a re-parse did not produce resolves to `undefined` and the caller's `if (!target) continue` no-ops rather than throwing. Hardens the added-bullet path through the same resolver too.

PR #496, opened 2026-07-16
fix(contact): admit LinkedIn-shaped /in/ URLs on non-linkedin.com hosts (#378)

### LINKED ISSUE #378: [edit-interaction] contact — "LinkedIn not detected" shown next to a live linkedin-shaped URL, self-contradictory
## Category
edit-interaction (Steps 1 & 3 F&F QC — provenance category)

## Fixture
`tests/fixtures/pdfs/latex/multi-degree-coursework.pdf` (synthetic persona)

## Ground truth (`pdftotext`)
```
973-555-0123 | jordan.bennett@example.com | LinkedIn | GitHub | US Citizen
```

The PDF has `LinkedIn` and `GitHub` as **hyperlinked labels**, not visible URLs. The underlying link on this synthetic fixture points to `example.com/in/jordan-bennett` (using the synthetic-persona URL convention — no real linkedin.com URL, because the fixture is public and PII-safe).

## Observed
Reconstructed contact block shows two conflicting signals on the same row:
```
jordan.bennett@example.com | (973) 555-0123 | ⚠ Location not detected
⚠ LinkedIn not detected · example.com/in/jordan-bennett ↗
```

- The `⚠ LinkedIn not detected` warning is shown
- ...and immediately followed by the live `example.com/in/jordan-bennett ↗` link

The parser rejected the URL as a LinkedIn URL (host isn't `linkedin.com`), classified it as a generic `website_url`/`portfolio_url`, and rendered both — but the two signals contradict each other.

## Category rationale
The parser is technically correct — `example.com` is not `linkedin.com`. This is a **display polish** issue: the reconstructed view shouldn't render both signals side-by-side.

## Localized layer
`src/components/features/ContactCard.tsx` — the field render logic for LinkedIn / other-URL fields likely renders each field independently without checking whether another field claims the same "slot."

## Suggested fix
When `linkedin_url` is empty AND another URL field looks LinkedIn-shaped (`/\/in\//` path), either:
1. Promote it to `linkedin_url` silently (loosen the parser's host requirement to admit `.com/in/...` paths), OR
2. Suppress the `⚠ LinkedIn not detected` warning when the alternative URL is displayed on the same row.

Option 1 is probably the right long-term fix — the fixture-URL convention (`example.com/in/...`) mirrors the shape of real LinkedIn URLs, and any user whose LinkedIn URL is on a non-`linkedin.com` host (personal-domain redirects, `lnkd.in/...` shortlinks) hits the same trap.

## Repro
```
npm run dev
# drop tests/fixtures/pdfs/latex/multi-degree-coursework.pdf
# look at contact block
```

Samhit21:
## Summary

A hyperlinked "LinkedIn" label whose target isn't on `linkedin.com` (a personal-domain redirect, an `lnkd.in` shortlink, or the synthetic `example.com/in/<handle>` fixture convention) was rejected as LinkedIn and dumped into `website_url`. The contact card then rendered a self-contradictory `⚠ LinkedIn not detected` pill right next to the live linkedin-shaped link.

This is the parser-side fix (option 1 from the issue): a host-agnostic `isLinkedinShapedUrl` predicate that admits the canonical `/in/<handle>` path and `lnkd.in` shortlinks in addition to `linkedin.com`. It's used in the text-scan pick, the annotation pick, and the other-URL exclusion, so a linkedin-shaped link resolves to `linkedin_url` and never double-renders as a website. No UI change is needed — the card already gates the "not detected" pill on field presence.

Verified against the reported fixture (`tests/fixtures/pdfs/latex/multi-degree-coursework.pdf`): the hyperlinked `example.com/in/jordan-bennett` annotation now populates `linkedin_url`, the completeness `missing` list drops `LinkedIn`, and score rises 86 → 89. Snapshot regenerated via `npm run bake-fixtures` (only this fixture changed).

Closes #378

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green (1689 passed; new regression tests for `/in/` annotation, `lnkd.in` shortlink, visible-text bare URL, and the `/index` non-match)

### DIFF
--- a/src/lib/heuristics/extract/contact.ts
+++ b/src/lib/heuristics/extract/contact.ts
@@ -484,10 +484,12 @@ export function extractContact(
   const github = pickUrl("github_url", () =>
     findAnnotationUrl(isGithubUrl, anywhereOnDoc),
   );
-  // A URL is "the same claim" across tiers regardless of scheme/`www.`/trailing
-  // punctuation, so compare on the normalized form — the text scan normalizes
-  // its values, an annotation carries the raw href.
-  const claimKey = (u: string): string => normalizeUrl(u) ?? u;
+  // A URL is "the same claim" across tiers regardless of scheme/`www.`/case/
+  // trailing slash or punctuation, so compare on `urlSlug` — the identity form
+  // (`normalizeUrl` keeps a trailing `/` and case, so `.../jane/` and `.../jane`
+  // would read as different claims and the cross-tier dedup would silently
+  // no-op). Same helper the ownership dedup below uses.
+  const claimKey = (u: string): string => urlSlug(u) ?? u;
   const claimedByIdentity = new Set(
     [linkedin.value, github.value]
       .filter((u): u is string => Boolean(u))
--- a/src/lib/heuristics/extract/contact.ts
+++ b/src/lib/heuristics/extract/contact.ts
@@ -85,6 +85,66 @@ function isLinkedinProfileUrl(u: string): boolean {
   return /linkedin\.com\/[A-Za-z0-9]/i.test(u) && !LINKEDIN_NONPROFILE_RE.test(u);
 }
 
+/** Anchored `lnkd.in` shortlink test: a boundary before the host (string start,
+ *  `//`, or a subdomain dot) and a path slash after. The anchors keep it precise
+ *  — a plain `u.includes("lnkd.in")` would also match `blnkd.info` and a
+ *  slash-less `lnkd.in`. Shared by {@link isLinkedinShapedUrl} and the
+ *  website/portfolio exclusion, where it is the only LinkedIn signal the literal
+ *  `linkedin.com` check there does not already cover. */
+function isLnkdInShortlink(u: string): boolean {
+  return /(?:^|\/\/|\.)lnkd\.in\//i.test(u);
+}
+
+/** True when `u` is a LinkedIn identity link on a KNOWN LinkedIn host — the
+ *  canonical `linkedin.com` profile OR the `lnkd.in/...` shortlink domain. Both
+ *  are host-anchored, so this is safe to run over doc-wide visible text without
+ *  false positives. (The looser, host-agnostic `/in/<handle>` redirect shape is
+ *  handled separately by {@link isLinkedinRedirectUrl}, which is restricted to
+ *  authored link annotations — see #378.) */
+function isLinkedinShapedUrl(u: string): boolean {
+  return isLinkedinProfileUrl(u) || isLnkdInShortlink(u);
+}
+
+/** True when `u` is a LinkedIn-*shaped* profile redirect on a NON-`linkedin.com`
+ *  host: `<host>/in/<handle>` as the sole path segment — a personal-domain
+ *  redirect (`jane.dev/in/jane`) or (in our public, PII-safe fixtures) the
+ *  synthetic `example.com/in/<handle>` convention that mirrors LinkedIn's real
+ *  path shape. Résumés hyperlink a visible "LinkedIn" label to such a URL; since
+ *  the annotation carries no anchor text (`PdfLinkAnnotation` has only the URL),
+ *  the `/in/<handle>` shape is the only signal we get.
+ *
+ *  This is deliberately NOT part of {@link isLinkedinShapedUrl}: a bare `/in/`
+ *  path is host-unbounded and matches non-LinkedIn URLs — India-locale paths
+ *  (`nike.com/in/en`), catalog paths (`samsung.com/in/smartphones`), and
+ *  mid-path `/in/` segments (`janedoe.com/portfolio/in/2023`). Applied over
+ *  doc-wide *visible text* it would hijack `linkedin_url`. So it is used ONLY as
+ *  an ANNOTATION predicate (an authored hyperlink is high-intent and rarely
+ *  points at a locale/catalog page), and constrained to `/in/<handle>` as the
+ *  first *and only* path segment with a ≥3-char handle — which rejects the
+ *  mid-path and 2-letter-locale (`/in/en`, `/in/us`) cases outright (#378).
+ *
+ *  ACCEPTED TRADEOFF: the handle is not name-shape-checked, so a sole-segment
+ *  non-LinkedIn `/in/<≥3>` link — a personal portfolio at
+ *  `janedoe.com/in/design-portfolio`, or `acme.com/in/careers` — is claimed as
+ *  LinkedIn when it sits in the contact block. No regex cleanly separates
+ *  `/in/jane-doe` from `/in/design-portfolio` (both are hyphenated), so the
+ *  profile-band restriction (the caller only applies this predicate to
+ *  contact-block annotations) IS the mitigation: an authored `/in/`-shaped link
+ *  in the header is far more likely a "LinkedIn" label than a catalog page.
+ *  Pinned by test rather than narrowed further. */
+function isLinkedinRedirectUrl(u: string): boolean {
+  if (isLinkedinShapedUrl(u)) return true;
+  // The host must carry a dot (`acme.com`, not a dotless `foo`/`localhost`) —
+  // a real personal-domain redirect is an absolute host, and requiring the dot
+  // drops `foo/in/jane` / `localhost/in/team` false positives. The terminator
+  // allows an optional query/fragment (`?utm_source=cv`, `#about`) — tracking
+  // params on a résumé link are common — while the locale/mid-path cases are
+  // still rejected on handle length and segment position, not the terminator.
+  return /^(?:https?:\/\/)?(?:www\.)?[^/\s]+\.[^/\s]+\/in\/[A-Za-z0-9][A-Za-z0-9._-]{2,}\/?(?:[?#]\S*)?$/i.test(
+    u,
+  );
+}
+
 // ── Promoted-identity-link ownership (#134) ─────────────────────────────────
 // LinkedIn/GitHub identity links are matched document-wide (see `anywhereOnDoc`
 // below), so an identity link sitting in a "Links"/footer block that segmented
@@ -219,7 +279,18 @@ function extractOtherUrls(joined: string): {
   const withoutEmails = joined.replace(EMAIL_RE, " ");
   const others = allMatches(URL_RE, withoutEmails).filter((u) => {
     const lower = u.toLowerCase();
-    if (lower.includes("linkedin.com") || lower.includes("github.com")) {
+    // Exclude any linkedin.com URL (profile OR non-profile company/jobs/feed
+    // pages — the literal-host guard keeps those out of website_url), the
+    // `lnkd.in` shortlink host, and github.com. The literal `linkedin.com`
+    // check is load-bearing: it also catches non-profile `linkedin.com/company/…`
+    // pages, which the shortlink test does not (#378). The three terms are
+    // uniform host checks — `lnkd.in` is the one LinkedIn signal `linkedin.com`
+    // doesn't already cover.
+    if (
+      lower.includes("linkedin.com") ||
+      isLnkdInShortlink(u) ||
+      lower.includes("github.com")
+    ) {
       return false;
     }
     // Reject `token.token` false positives (skills like `Node.js`, `etc.`).
@@ -253,7 +324,7 @@ function scan(lines: PdfLine[], joined: string): ContactScanResult {
   // so a hyperlinked "LinkedIn" anchor resolves regardless of the path shape.
   const linkedin =
     firstMatch(LINKEDIN_RE, joined) ??
-    allMatches(URL_RE, joined).find(isLinkedinProfileUrl);
+    allMatches(URL_RE, joined).find(isLinkedinShapedUrl);
   const github = firstMatch(GITHUB_RE, joined);
 
   // Other URLs that aren't linkedin/github → portfolio/website bucket.
@@ -312,9 +383,11 @@ export function extractContact(
   const fallback = scan(allLines, fullText);
 
   // Annotation fallback: URLs hyperlinked behind a visible word
-  // ("LinkedIn", "GitHub") only show up here. LinkedIn/GitHub are matched
-  // document-wide (see `anywhereOnDoc` below); only the looser-but-still-bounded
-  // portfolio/website lookup keeps a section-based region filter.
+  // ("LinkedIn", "GitHub") only show up here. GitHub, and LinkedIn on a KNOWN
+  // host (linkedin.com / lnkd.in), are matched document-wide (see `anywhereOnDoc`
+  // below) — the host anchors them. The host-agnostic LinkedIn `/in/<handle>`
+  // redirect has no host anchor, so its lookup is section-banded like
+  // portfolio/website (see the two-tier `linkedin` resolver below).
   // Portfolio/website use the profile-section boundary — the profile section
   // covers everything above the first recognized header, which is exactly
   // the contact/links block. Annotations whose top edge falls within (or just
@@ -353,7 +426,11 @@ export function extractContact(
     return undefined;
   };
 
-  const isLinkedinUrl = isLinkedinProfileUrl;
+  // `isLinkedinUrl` classifies a URL as LinkedIn for the portfolio/website
+  // EXCLUSIONS below (a `/in/` redirect must never also read as a website). The
+  // annotation LOOKUP for the linkedin slot is split by host anchor — see the
+  // two-tier resolver on `linkedin` below.
+  const isLinkedinUrl = isLinkedinRedirectUrl;
   const isGithubUrl = (u: string) =>
     /github\.com\//i.test(u) && !/github\.com\/(orgs|topics|search)/i.test(u);
   const isPortfolioUrl = (u: string) =>
@@ -361,44 +438,91 @@ export function extractContact(
       u,
     );
 
+  // `reject` runs against EVERY tier's candidate — the text-scan primary/
+  // fallback values as well as the annotation — so a URL already claimed by a
+  // higher-priority field can't re-enter through the text scan. Without it a
+  // `/in/` redirect printed as its own label (LaTeX `\href{url}{url}`) fills
+  // `linkedin_url` from the annotation AND `website_url` from the text scan —
+  // the exact double-render this PR removes (#378).
   const pickUrl = <K extends "linkedin_url" | "github_url" | "portfolio_url" | "website_url">(
     key: K,
-    annotationPredicate: (url: string) => boolean,
-    band: (ann: PdfLinkAnnotation) => boolean,
+    resolveAnnotation: () => string | undefined,
+    reject?: (url: string) => boolean,
   ): { value: string | undefined; confidence: number } => {
-    if (primary[key])
+    if (primary[key] && !reject?.(primary[key]!))
       return { value: primary[key], confidence: primary.confidence[key] };
-    if (fallback[key])
+    if (fallback[key] && !reject?.(fallback[key]!))
       return {
         value: fallback[key],
         confidence: fallback.confidence[key] * 0.9,
       };
-    const annUrl = findAnnotationUrl(annotationPredicate, band);
-    if (annUrl) return { value: annUrl, confidence: 0.95 };
+    const annUrl = resolveAnnotation();
+    if (annUrl && !reject?.(annUrl))
+      return { value: annUrl, confidence: 0.95 };
     return { value: undefined, confidence: 0 };
   };
 
-  const linkedin = pickUrl("linkedin_url", isLinkedinUrl, anywhereOnDoc);
-  const github = pickUrl("github_url", isGithubUrl, anywhereOnDoc);
+  // A hyperlinked "LinkedIn" label often points at a non-linkedin.com redirect
+  // whose path is still the `/in/<handle>` profile shape. Two-tier lookup, split
+  // by whether the host anchors the match:
+  //   1. Known LinkedIn hosts (`linkedin.com` / `lnkd.in`) — the host anchors
+  //      the match, so it is safe ANYWHERE on the doc (a footer/icon link).
+  //   2. Host-agnostic `/in/<handle>` redirect — no host anchor, so confine it
+  //      to the profile band, where an authored "LinkedIn" label actually lives.
+  //      Otherwise a body link like `acme.com/in/analytics` (sole segment, ≥3
+  //      chars) would win the linkedin_url slot ahead of github/portfolio/
+  //      website (#378 follow-up).
+  const linkedin = pickUrl(
+    "linkedin_url",
+    () =>
+      findAnnotationUrl(isLinkedinShapedUrl, anywhereOnDoc) ??
+      findAnnotationUrl(
+        (u) => isLinkedinRedirectUrl(u) && !isLinkedinShapedUrl(u),
+        inProfileSection,
+      ),
+  );
+  const github = pickUrl("github_url", () =>
+    findAnnotationUrl(isGithubUrl, anywhereOnDoc),
+  );
+  // A URL is "the same claim" across tiers regardless of scheme/`www.`/trailing
+  // punctuation, so compare on the normalized form — the text scan normalizes
+  // its values, an annotation carries the raw href.
+  const claimKey = (u: string): string => normalizeUrl(u) ?? u;
+  const claimedByIdentity = new Set(
+    [linkedin.value, github.value]
+      .filter((u): u is string => Boolean(u))
+      .map(claimKey),
+  );
   const portfolio = pickUrl(
     "portfolio_url",
-    (u) => isPortfolioUrl(u) && !isLinkedinUrl(u) && !isGithubUrl(u),
-    inProfileSection,
+    () =>
+      findAnnotationUrl(
+        (u) => isPortfolioUrl(u) && !isLinkedinUrl(u) && !isGithubUrl(u),
+        inProfileSection,
+      ),
+    (u) => claimedByIdentity.has(claimKey(u)),
   );
-  // Website is the catch-all: any remaining annotation URL not already
-  // claimed by linkedin/github/portfolio.
+  // Website is the catch-all: any remaining URL not already claimed by
+  // linkedin/github/portfolio — on ANY tier (text scan or annotation), so a
+  // `/in/` redirect that is both visible text and a hyperlink can't fill both
+  // linkedin_url and website_url (#378).
   const claimedUrls = new Set(
-    [linkedin.value, github.value, portfolio.value].filter(Boolean) as string[],
+    [linkedin.value, github.value, portfolio.value]
+      .filter((u): u is string => Boolean(u))
+      .map(claimKey),
   );
   const website = pickUrl(
     "website_url",
-    (u) =>
-      !claimedUrls.has(u) &&
-      !isLinkedinUrl(u) &&
-      !isGithubUrl(u) &&
-      !u.startsWith("mailto:") &&
-      !u.startsWith("tel:"),
-    inProfileSection,
+    () =>
+      findAnnotationUrl(
+        (u) =>
+          !isLinkedinUrl(u) &&
+          !isGithubUrl(u) &&
+          !u.startsWith("mailto:") &&
+          !u.startsWith("tel:"),
+        inProfileSection,
+      ),
+    (u) => claimedUrls.has(claimKey(u)),
   );
 
   // Ownership (#134): claim the document-wide body lines that are nothing but a
--- a/src/lib/heuristics/extract/contact.test.ts
+++ b/src/lib/heuristics/extract/contact.test.ts
@@ -358,6 +358,26 @@ describe("extractContact — LinkedIn-shaped URL on a non-linkedin.com host (#37
     expect(result.website_url).toBeUndefined();
   });
 
+  it("dedups across tiers when the href has a trailing slash the text label lacks", () => {
+    // The `\href{url}{url}` double-render, but the annotation carries a
+    // canonical trailing slash (`.../jane/`) while the visible text is bare
+    // (`.../jane`). The cross-tier claim key must treat them as the SAME link —
+    // `normalizeUrl` keeps the trailing slash and would let both slots fill.
+    const contactLine = mkLine(
+      "Jane Doe | jane@example.com | jane.dev/in/jane | LinkedIn",
+      0,
+    );
+    const profile: PdfSection = { name: "profile", lines: [contactLine] };
+    const result = extractContact(profile, [contactLine], [
+      mkAnnotation("https://jane.dev/in/jane/"),
+    ]);
+
+    expect(result.linkedin_url).toBe("https://jane.dev/in/jane/");
+    // Same target as linkedin_url → must not also render as portfolio/website.
+    expect(result.portfolio_url).toBeUndefined();
+    expect(result.website_url).toBeUndefined();
+  });
+
   it("claims a profile-band sole-segment /in/ link as linkedin_url (accepted tradeoff)", () => {
     // A non-LinkedIn `/in/<≥3>` link in the contact block — a personal portfolio
     // at `janedoe.com/in/design-portfolio` — is claimed as LinkedIn. No regex
--- a/src/lib/heuristics/extract/contact.test.ts
+++ b/src/lib/heuristics/extract/contact.test.ts
@@ -3,6 +3,7 @@
 
 import { describe, it, expect } from "vitest";
 import { extractContact } from "./contact.ts";
+import type { PdfLinkAnnotation } from "../types.ts";
 import type { PdfLine, PdfSection } from "../sections.ts";
 
 /** Minimal PdfLine carrying only the fields extractContact reads. */
@@ -17,6 +18,11 @@ function mkLine(text: string, y = 0): PdfLine {
     allCaps: false,
     gapAbove: 0,
   };
+}
+
+/** Minimal page-1 link annotation; `yTop` picks the region (0 = profile band). */
+function mkAnnotation(url: string, yTop = 0): PdfLinkAnnotation {
+  return { page: 1, url, rect: [0, 0, 0, 0], yTop };
 }
 
 describe("extractContact — email domain is not a website", () => {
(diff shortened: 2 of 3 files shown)


src/lib/heuristics/extract/contact.ts
```diff
@@ -85,6 +85,26 @@ function isLinkedinProfileUrl(u: string): boolean {
...
+ *  `example.com/in/<handle>` convention that mirrors LinkedIn's real path shape.
+ *  Admitting the `/in/<handle>` shape lets those resolve to `linkedin_url`
+ *  instead of falling through to the generic website/portfolio bucket — which
+ *  otherwise renders a self-contradictory "LinkedIn not detected" pill right
+ *  next to the live linkedin-shaped link (#378). */
+function isLinkedinShapedUrl(u: string): boolean {
+  return (
+    isLinkedinProfileUrl(u) ||
+    /(?:^|\/\/|\.)lnkd\.in\//i.test(u) ||
+    /\/in\/[A-Za-z0-9]/i.test(u)
```
s-annam: **Blocking — host-unbounded `/in/`.** This matches `/in/<alnum>` on *any* host, so India-locale paths (`nike.com/in/en`, `samsung.com/in/...`) and generic `/in/` segments get promoted to `linkedin_url`. It runs doc-wide (full-doc fallback text + annotation path), so any such URL in the résumé wins the `linkedin_url` slot when no real `linkedin.com` URL is present. Constrain to a redirect/shortlink host instead of matching every host.

src/lib/heuristics/extract/contact.ts
```diff
@@ -219,7 +239,11 @@ function extractOtherUrls(joined: string): {
   const withoutEmails = joined.replace(EMAIL_RE, " ");
   const others = allMatches(URL_RE, withoutEmails).filter((u) => {
     const lower = u.toLowerCase();
-    if (lower.includes("linkedin.com") || lower.includes("github.com")) {
+    // Exclude LinkedIn-*shaped* URLs (host-agnostic `/in/<handle>`, `lnkd.in`)
+    // as well as literal linkedin.com/github.com — otherwise a linkedin-shaped
+    // link on a non-linkedin.com host would be claimed by BOTH linkedin_url and
+    // website_url and render twice (#378).
+    if (isLinkedinShapedUrl(u) || lower.includes("github.com")) {
```
s-annam: **Blocking — regression.** Replacing `lower.includes("linkedin.com")` with `isLinkedinShapedUrl(u)` stops excluding non-profile `linkedin.com` pages: `isLinkedinShapedUrl("linkedin.com/company/foo")` is `false` (nonprofile, no `/in/`, no `lnkd.in`). So a `linkedin.com/company|jobs|feed` URL now leaks into `website_url`. Keep the literal guard too:
```js
if (lower.includes("linkedin.com") || isLinkedinShapedUrl(u) || lower.includes("github.com")) {
```

s-annam (changes requested):
Solid, well-scoped fix for #378 — the LinkedIn-shaped `/in/` promotion works and is verified end-to-end. Two newly-introduced correctness issues below need a fix before merge; both are one-liners.

## Verified (what was tested + how to reproduce)
- **Fixture `tests/fixtures/pdfs/latex/multi-degree-coursework.pdf`** — its link annotation is `https://example.com/in/jordan-bennett` (LinkedIn-shaped, non-`linkedin.com` host). The fix promotes it to `linkedin_url`; the corpus snapshot (`multi-degree-coursework.expected.json`) updates accordingly: overall score 86→89, completeness 21→24, and `"LinkedIn"` drops out of `missing[]`. **Reproduce:** `npx vitest run -t "multi-degree-coursework"`, or drop that PDF into the app and read the contact card.
- **Unit tests** (`contact.test.ts`, synthetic inline stubs — no fixture): `/in/<handle>` annotation, `lnkd.in/...` shortlink, visible-text `example.com/in/jane-doe`, and the negative `janedoe.com/index`. 13/13 pass.
- **Gates:** `npm run check:fixtures` ✓ (52 PDFs, all personas synthetic); `typecheck` ✓. (fallow's dead-code/dup findings trace to unrelated files, not this diff.)

## Blocking

**1. `contact.ts:104` — the `/in/` path rule is host-unbounded, so non-LinkedIn URLs get promoted to `linkedin_url`.** `/\/in\/[A-Za-z0-9]/i` matches *any* host. India-locale paths (`nike.com/in/en`, `samsung.com/in/smartphones`) and generic `/in/` segments (`janedoe.com/portfolio/in/2023`) all match. Because the check runs over the full-document fallback text and the doc-wide annotation path, such a URL anywhere in the résumé is pulled into `linkedin_url` whenever no genuine `linkedin.com` URL exists. (`example.com/index` is rejected only because `/index` lacks a trailing slash — not a real guard.) Constrain the path rule to a redirect/shortlink host rather than any host.

**2. `contact.ts:246` — dropping the literal `linkedin.com` guard leaks company/jobs pages into `website_url` (regression).** `isLinkedinShapedUrl("https://linkedin.com/company/foo")` returns **false** (nonprofile → not a profile; no `lnkd.in`; no `/in/`). The old filter excluded it via `lower.includes("linkedin.com")`; the new one no longer does, so a `linkedin.com/company|jobs|feed` URL now renders as the candidate's personal `website_url`. Fix:
```js
if (lower.includes("linkedin.com") || isLinkedinShapedUrl(u) || lower.includes("github.com")) {
  return false;
}
```

## Nits
- No negative test pins Finding 2 — add one asserting a `linkedin.com/company/...` URL stays out of `website_url` (it would fail against the current diff and lock the fix).

---
Verdict rule applied: ≥1 Blocking → REQUEST_CHANGES.

---

src/lib/heuristics/extract/contact.ts
Samhit21: Fixed in 83e0579. Split the host-agnostic `/in/` rule out of `isLinkedinShapedUrl` into a separate `isLinkedinRedirectUrl` and confined it to the **authored-annotation predicate only** — it no longer runs over the doc-wide text scan (`scan()` / full-doc fallback), where those locale/catalog URLs live. It is also tightened to `/in/<handle>` as the **sole first path segment with a ≥3-char handle**, which rejects your examples outright: `nike.com/in/en` (2-char handle) and `janedoe.com/portfolio/in/2023` (non-first `/in/`). Added a negative test asserting both stay out of `linkedin_url`, plus one asserting a visible-text `/in/` URL is treated as a website, not LinkedIn.

src/lib/heuristics/extract/contact.ts
Samhit21: Fixed in 83e0579. Restored the literal `linkedin.com` guard alongside the shaped check, exactly as suggested: `if (lower.includes("linkedin.com") || isLinkedinShapedUrl(u) || lower.includes("github.com"))`. Added the negative test you flagged in the nit — `linkedin.com/company/acme` now asserts out of both `linkedin_url` and `website_url`.

Samhit21:
Pushed review fixes in `83e0579` (amended — one commit per PR).

**Blocking 1 — host-unbounded `/in/`:** The generic `/in/<handle>` rule can't be made safe over doc-wide *visible text* (it can't tell `example.com/in/jordan-bennett` from `samsung.com/in/smartphones`). So I split it out of `isLinkedinShapedUrl` (which stays host-anchored: `linkedin.com` profile + `lnkd.in`, safe for text) into a new `isLinkedinRedirectUrl` used **only as the annotation predicate** — an authored hyperlink is high-intent and rarely points at a locale/catalog page. It's further constrained to `/in/<handle>` as the sole first path segment with a ≥3-char handle, which rejects `nike.com/in/en` and `janedoe.com/portfolio/in/2023`.

**Blocking 2 — leaked `linkedin.com/company` → website:** Restored the literal `linkedin.com` guard in `extractOtherUrls`, per your snippet.

**Nit:** Added the `linkedin.com/company/...` negative test, plus tests pinning the finding-1 false-positives and the annotation-only scoping.

Gates green: `contact.test.ts` + corpus (snapshot unchanged — fixture still promotes via the annotation), `typecheck`, `lint`. Re-requesting review.

## Provenance
Authored with Claude Opus 4.8 (1M context).

src/lib/heuristics/extract/contact.ts
```diff
@@ -353,7 +397,14 @@ export function extractContact(
...
 
-  const isLinkedinUrl = isLinkedinProfileUrl;
+  // Annotation predicate only: a hyperlinked "LinkedIn" label often points at a
+  // non-linkedin.com redirect whose path is still the `/in/<handle>` profile
+  // shape — admit those (via isLinkedinRedirectUrl) so the annotation resolves
+  // to linkedin_url rather than the generic website bucket. The host-agnostic
+  // `/in/` shape is intentionally confined to this authored-annotation path, not
+  // the doc-wide text scan, where it would false-match locale/catalog URLs
+  // (`nike.com/in/en`, `janedoe.com/portfolio/in/2023`) — see #378.
+  const isLinkedinUrl = isLinkedinRedirectUrl;
```
s-annam, line 433: **Secondary — host-agnostic `/in/` redirect runs document-wide.** `isLinkedinUrl = isLinkedinRedirectUrl` is passed to `pickUrl(..., anywhereOnDoc)` (line 432), so the loose `<host>/in/<handle>` shape matches annotations *anywhere*, not just the contact band. A body link like `https://acme.com/in/analytics` (sole segment, at least 3 chars) would be promoted to `linkedin_url` ahead of github/portfolio/website. The doc-comment's "high-intent authored hyperlink" defense holds for the known LinkedIn hosts, but the host-agnostic branch has no host anchor — consider banding *just that branch* to `inProfileSection`, where an authored "LinkedIn" label actually lives. Non-blocking (needs a sole-segment `/in/xxx` body link to trigger).

src/lib/heuristics/extract/contact.ts
```diff
@@ -219,7 +253,17 @@ function extractOtherUrls(joined: string): {
...
-    if (lower.includes("linkedin.com") || lower.includes("github.com")) {
+    // Exclude any linkedin.com URL (profile OR non-profile company/jobs/feed
+    // pages — the literal-host guard keeps those out of website_url) and the
+    // `lnkd.in` shortlink host, plus github.com. The literal `linkedin.com`
+    // check is load-bearing: `isLinkedinShapedUrl` alone returns false for a
+    // non-profile `linkedin.com/company/…` page, which would otherwise leak
+    // into website_url (#378).
+    if (
+      lower.includes("linkedin.com") ||
+      isLinkedinShapedUrl(u) ||
```
s-annam: **Nit — redundant term.** `isLinkedinShapedUrl(u)` is `isLinkedinProfileUrl || lnkd.in`; its `linkedin.com` half is fully subsumed by the `lower.includes("linkedin.com")` on the line above, so the only new bytes are the `lnkd.in` exclusion. `lower.includes("lnkd.in")` says the same thing without re-running the profile regex. (The load-bearing-literal comment is still correct — this just tightens what the extra term contributes.)

src/lib/heuristics/extract/contact.ts
```diff
@@ -85,6 +85,40 @@ function isLinkedinProfileUrl(u: string): boolean {
...
+ *  (`nike.com/in/en`), catalog paths (`samsung.com/in/smartphones`), and
+ *  mid-path `/in/` segments (`janedoe.com/portfolio/in/2023`). Applied over
+ *  doc-wide *visible text* it would hijack `linkedin_url`. So it is used ONLY as
+ *  an ANNOTATION predicate (an authored hyperlink is high-intent and rarely
+ *  points at a locale/catalog page), and constrained to `/in/<handle>` as the
+ *  first *and only* path segment with a ≥3-char handle — which rejects the
+ *  mid-path and 2-letter-locale (`/in/en`, `/in/us`) cases outright (#378). */
+function isLinkedinRedirectUrl(u: string): boolean {
+  if (isLinkedinShapedUrl(u)) return true;
+  return /^(?:https?:\/\/)?(?:www\.)?[^/\s]+\/in\/[A-Za-z0-9][A-Za-z0-9._-]{2,}\/?$/i.test(
```
s-annam: **Nit — loose host class.** `[^/\s]+` matches dotless hosts (`foo/in/jane`, `localhost/in/team`). Real link annotations are near-always absolute URLs so low-likelihood, but requiring a dot or an explicit scheme would match the "personal-domain redirect" intent the doc-comment states.

s-annam (dismissed):
Solid, well-documented fix. The split between the host-anchored `isLinkedinShapedUrl` (safe over doc-wide text) and the annotation-only `isLinkedinRedirectUrl` (host-agnostic `/in/<handle>`) is the right seam, and the test matrix covers the important positives *and* negatives (`/in/en` locale, mid-path `/in/`, plain-text `/in/` → website, `linkedin.com/company` guard). No blockers.

**Gates:** fixture PII green (persona `Jordan Bennett` / `example.com` / `973-555-0123`, PDF binary unchanged — only `expected.json` re-derived); 15 contact tests pass; typecheck clean; `fallow audit` exit 0. Design-system / style-token / command-bug gates N/A (lib-only diff).

Verdict rule: 0 Blocking → APPROVE. Findings below are non-blocking.

## Secondary
See inline comment at `contact.ts:407` — host-agnostic `/in/` redirect predicate runs document-wide.

## Nits
See inline comments at `contact.ts:264` (redundant filter term) and `contact.ts:117` (loose host class).

---

src/lib/heuristics/extract/contact.ts
s-annam: One-click version of this nit — require the host to carry a dot, so `foo/in/jane` and `localhost/in/team` drop out while `acme.com/in/jane` still matches:

```suggestion
  return /^(?:https?:\/\/)?(?:www\.)?[^/\s]+\.[^/\s]+\/in\/[A-Za-z0-9][A-Za-z0-9._-]{2,}\/?$/i.test(
```

Non-blocking — apply and re-run the contact tests.

src/lib/heuristics/extract/contact.ts
s-annam: Retracting this nit — my equivalence claim was wrong. `isLinkedinShapedUrl`'s `lnkd.in` half is the anchored `/(?:^|\/\/|\.)lnkd\.in\//i` (boundary before, trailing slash after), so a plain `lower.includes("lnkd.in")` is strictly *broader*: it would match `blnkd.info` and a slash-less `lnkd.in`. The current `isLinkedinShapedUrl(u)` term is correct as-is — please disregard. (The load-bearing-literal comment above stays right.)

src/lib/heuristics/extract/contact.ts
s-annam: Gentle nudge on this one, @Samhit21 — it's the only open item with real teeth. The other two are handled: I posted a one-click suggestion on the line-117 nit, and I've withdrawn the line-264 nit as incorrect.

This one isn't a one-liner: the fix is to band *just* the host-agnostic `/in/<handle>` branch to `inProfileSection` instead of `anywhereOnDoc`, so a body link like `acme.com/in/analytics` can't win the `linkedin_url` slot ahead of github/portfolio/website. It's non-blocking and edge (needs a sole-segment `/in/xxx` body link to trigger), so if you'd rather land this PR as-is and do it as a fast follow-up issue, that's a fine call too — your PR, your call. No rush; just flagging since the thread had gone quiet.

Samhit21: Fixed in 4c0c2d2. Split the linkedin annotation lookup into two tiers by host anchor: known LinkedIn hosts (`linkedin.com` / `lnkd.in`, via `isLinkedinShapedUrl`) still match `anywhereOnDoc`, but the host-agnostic `/in/<handle>` redirect (`isLinkedinRedirectUrl && !isLinkedinShapedUrl`) is now banded to `inProfileSection`. A body link like `acme.com/in/analytics` can no longer win the `linkedin_url` slot ahead of github/portfolio/website. Added a positive (in-band promotes) + negative (out-of-band does not) test pair.

src/lib/heuristics/extract/contact.ts
Samhit21: Applied your one-click suggestion in 4c0c2d2 — the host must now carry a dot (`[^/\s]+\.[^/\s]+`), so `foo/in/jane` and `localhost/in/team` drop out while `acme.com/in/jane` still matches. Added a `localhost/in/team` negative test.

src/lib/heuristics/extract/contact.ts
Samhit21: Thanks — noted your retraction (the `lnkd.in` half of `isLinkedinShapedUrl` is anchored, so a plain `lower.includes("lnkd.in")` would be strictly broader). Leaving the `isLinkedinShapedUrl(u)` term as-is.

Samhit21:
Addressed the two remaining review items in `4c0c2d2` (amended — one commit per PR):

- **Secondary (`/in/` redirect runs doc-wide):** the linkedin annotation lookup is now two-tier by host anchor — known LinkedIn hosts (`linkedin.com` / `lnkd.in`) match `anywhereOnDoc`, but the host-agnostic `/in/<handle>` redirect is banded to `inProfileSection`, so a body link like `acme.com/in/analytics` can't win the slot ahead of github/portfolio/website.
- **Nit (loose host class):** applied the one-click suggestion — the redirect host must carry a dot, dropping `foo/in/jane` / `localhost/in/team`.
- **Nit (redundant term):** left as-is per your retraction.

New tests: in-band redirect promotes, out-of-band redirect does not, dotless host rejected. All 18 contact tests pass; corpus snapshot unchanged (the fixture annotation is in-band); typecheck + lint clean.

The amend dismissed the prior approval — re-requesting review. 🙏

## Provenance
Authored with Claude Opus 4.8 (1M context).

src/lib/heuristics/extract/contact.ts
```diff
@@ -219,7 +256,17 @@ function extractOtherUrls(joined: string): {
...
-    if (lower.includes("linkedin.com") || lower.includes("github.com")) {
+    // Exclude any linkedin.com URL (profile OR non-profile company/jobs/feed
+    // pages — the literal-host guard keeps those out of website_url) and the
+    // `lnkd.in` shortlink host, plus github.com. The literal `linkedin.com`
+    // check is load-bearing: `isLinkedinShapedUrl` alone returns false for a
+    // non-profile `linkedin.com/company/…` page, which would otherwise leak
+    // into website_url (#378).
+    if (
+      lower.includes("linkedin.com") ||
+      isLinkedinShapedUrl(u) ||
```
s-annam: **Nit (non-blocking):** this disjunct is nearly dead here. The preceding `lower.includes("linkedin.com")` already subsumes the `isLinkedinProfileUrl` half of `isLinkedinShapedUrl`, so this call only ever contributes the `lnkd.in` test. It reads as broader coverage than it provides — `lower.includes("lnkd.in")` would say the same thing in one line, and would make the three-way check uniform.

src/lib/heuristics/extract/contact.ts
```diff
@@ -85,6 +85,43 @@ function isLinkedinProfileUrl(u: string): boolean {
...
+ *  an ANNOTATION predicate (an authored hyperlink is high-intent and rarely
+ *  points at a locale/catalog page), and constrained to `/in/<handle>` as the
+ *  first *and only* path segment with a ≥3-char handle — which rejects the
+ *  mid-path and 2-letter-locale (`/in/en`, `/in/us`) cases outright (#378). */
+function isLinkedinRedirectUrl(u: string): boolean {
+  if (isLinkedinShapedUrl(u)) return true;
+  // The host must carry a dot (`acme.com`, not a dotless `foo`/`localhost`) —
+  // a real personal-domain redirect is an absolute host, and requiring the dot
+  // drops `foo/in/jane` / `localhost/in/team` false positives.
+  return /^(?:https?:\/\/)?(?:www\.)?[^/\s]+\.[^/\s]+\/in\/[A-Za-z0-9][A-Za-z0-9._-]{2,}\/?$/i.test(
```
s-annam: **Nit (non-blocking):** the `$` anchor drops redirects carrying a query or fragment. `https://jane.dev/in/jane?utm_source=cv` and `.../in/jane#about` both fail the shape and fall through to `website_url` — the same visible symptom this PR fixes, for the personal-domain-redirect case the docblock names as the target. Tracking params on a résumé link aren't exotic.

`\/?(?:[?#]\S*)?$` covers it at no cost to the guards: the locale and mid-path cases are rejected on handle length and segment position, not on the terminator.

No test covers a query-bearing or fragment-bearing redirect either, so this gap is currently invisible to the suite.

s-annam (dismissed):
Verified against head `4c0c2d2`. The two-tier split — host-anchored `isLinkedinShapedUrl` doc-wide, `^…$`-anchored `isLinkedinRedirectUrl` confined to profile-band annotations — is the right shape, and the docblocks carry the reasoning well. I checked the redirect regex empirically: `nike.com/in/en`, `janedoe.com/portfolio/in/2023`, and `localhost/in/team` are all correctly rejected.

Gates all green: 2484 tests, typecheck, lint, `check:fixtures` (51 PDFs), fallow clean on changed files, single commit. Design-system and style-token gates don't apply — no `src/components/` changes.

I confirmed the fixture by decompressing its object streams rather than reading `pdftotext`, since `pdftotext` can't see link annotations: the annotation is `https://example.com/in/jordan-bennett` on the hyperlinked "LinkedIn" word, persona synthetic throughout (973 area code + `555` exchange + `0123`). The 86→89 delta follows from the recovered LinkedIn completeness credit — `completeness` 21→24 with "LinkedIn" dropping out of `missing`, and since the three dimension maxes already sum to 100 pre-weighted, a +3 there is a direct +3 overall.

One note on the docblock at `contact.ts:98-118`: I specifically checked whether it overclaims on the catalog case, since `samsung.com/in/smartphones` does pass the regex. It doesn't overclaim — it scopes "rejects outright" precisely to the mid-path and 2-letter-locale cases, and defends the catalog case separately via the annotation-only + profile-band restriction. That's accurate as written.

Nothing blocking. Three nits below, none gating — fine to fold into a later PR.

## Nits

**3. `contact.ts:361-364` — this comment is now stale.** It states LinkedIn is "matched document-wide … only the looser-but-still-bounded portfolio/website lookup keeps a section-based region filter." After this PR the LinkedIn tier-2 redirect lookup *is* section-banded, which is the load-bearing safety property of the whole change. Worth a sentence, because this block is what a future reader consults before widening a predicate again — and it's the exact comment that would make a doc-wide match look safe.

(Nits 1 and 2 are inline on the lines they concern.)

---

src/lib/heuristics/extract/contact.ts
Samhit21: Fixed in ce9b3f3, but not by switching to `lower.includes("lnkd.in")` — a prior review pass correctly noted that plain substring is strictly *broader* (it matches `blnkd.info` and a slash-less `lnkd.in`). Instead I extracted the anchored test into a named `isLnkdInShortlink(u)` helper (shared with `isLinkedinShapedUrl`) and used it here, so the three-way check now reads as uniform host checks while keeping the anchored precision. Updated the comment to say `lnkd.in` is the one LinkedIn signal `linkedin.com` doesn't already cover.

src/lib/heuristics/extract/contact.ts
Samhit21: Fixed in ce9b3f3 — terminator is now `\/?(?:[?#]\S*)?$`, so a query- or fragment-bearing redirect (`jane.dev/in/jane?utm_source=cv`, `.../in/jane#about`) promotes instead of falling through to `website_url`. The locale/mid-path guards are unaffected (they reject on handle length and segment position, not the terminator). Added a test covering both the query and fragment forms.

Samhit21:
Folded in all three nits from the latest review (`ce9b3f3`, amended — one commit per PR):

- **Nit 1 (redundant disjunct, `:267`):** extracted the anchored `lnkd.in` test into a named `isLnkdInShortlink(u)` (shared with `isLinkedinShapedUrl`) and used it in the website/portfolio exclusion — keeps the anchored precision (avoids the `blnkd.info` over-match a plain `includes("lnkd.in")` would cause) while making the three-way host check uniform.
- **Nit 2 (query/fragment dropped, `:120`):** terminator is now `\/?(?:[?#]\S*)?$`, so `jane.dev/in/jane?utm_source=cv` and `.../in/jane#about` promote to `linkedin_url`. Added a test for both.
- **Nit 3 (stale comment, `:361-364`):** rewritten to reflect the two-tier banding — known LinkedIn hosts match doc-wide, the host-agnostic `/in/` redirect is section-banded like portfolio/website.

Gates green: 20 contact tests, corpus snapshot unchanged, typecheck + lint clean. The amend re-dismissed the approval — re-requesting review.

## Provenance
Authored with Claude Opus 4.8 (1M context).

src/lib/heuristics/extract/contact.ts
```diff
@@ -373,32 +439,54 @@ export function extractContact(
...
-  const website = pickUrl(
-    "website_url",
-    (u) =>
-      !claimedUrls.has(u) &&
-      !isLinkedinUrl(u) &&
-      !isGithubUrl(u) &&
-      !u.startsWith("mailto:") &&
-      !u.startsWith("tel:"),
-    inProfileSection,
+  const website = pickUrl("website_url", () =>
```
s-annam: **Blocking — this re-introduces the double-render the PR removes.**

Reproduced on `ce9b3f3`:

```ts
const l = mkLine("Jane Doe | jane@example.com | example.com/in/jane-doe", 0);
const profile = { name: "profile", lines: [l] };
extractContact(profile, [l], [
  { page: 1, url: "https://example.com/in/jane-doe", rect: [0,0,0,0], yTop: 0 },
]);
// linkedin_url: https://example.com/in/jane-doe
// website_url : https://example.com/in/jane-doe   ← double-render
```

A regression, not pre-existing: before the diff the annotation failed `isLinkedinProfileUrl`, so only `website_url` was set — self-consistent. It also contradicts the PR body ("never double-renders as a website") and this diff's own `expect(result.website_url).toBeUndefined()` assertions, which only hold for the annotation-only variant. `\href{url}{url}` — the URL printed as its own label — is a common LaTeX résumé pattern, so this is a normal-use path, and it likely double-credits completeness (LinkedIn *and* website present).

Root cause: `claimedUrls` only filters the **annotation** tier. `pickUrl` returns `primary[key]` / `fallback[key]` — the text-scan values — before the resolver ever runs, and `extractOtherUrls` deliberately does not exclude host-agnostic `/in/` redirects (correctly — doc-wide it would false-match locales). Note the asymmetry: `lnkd.in` *was* added to the `extractOtherUrls` exclusion, so the shortlink case is safe; only the `/in/` redirect leaks.

Fix — apply the claim check to every tier, not just annotations. Smallest version:

```ts
const website = pickUrl("website_url", () => findAnnotationUrl(/* … */));
if (website.value && claimedUrls.has(website.value)) {
  website.value = undefined;
  website.confidence = 0;
}
```

Cleaner: give `pickUrl` a `reject?: (url: string) => boolean` applied to the primary/fallback values too, and pass `(u) => claimedUrls.has(u)` for `website_url` and `portfolio_url` (`portfolio_url` is reachable the same way). Please add a regression test for the visible-text + annotation combination.

src/lib/heuristics/extract/contact.ts
```diff
@@ -85,6 +85,56 @@ function isLinkedinProfileUrl(u: string): boolean {
...
+ *  mid-path and 2-letter-locale (`/in/en`, `/in/us`) cases outright (#378). */
+function isLinkedinRedirectUrl(u: string): boolean {
+  if (isLinkedinShapedUrl(u)) return true;
+  // The host must carry a dot (`acme.com`, not a dotless `foo`/`localhost`) —
+  // a real personal-domain redirect is an absolute host, and requiring the dot
+  // drops `foo/in/jane` / `localhost/in/team` false positives. The terminator
+  // allows an optional query/fragment (`?utm_source=cv`, `#about`) — tracking
+  // params on a résumé link are common — while the locale/mid-path cases are
+  // still rejected on handle length and segment position, not the terminator.
+  return /^(?:https?:\/\/)?(?:www\.)?[^/\s]+\.[^/\s]+\/in\/[A-Za-z0-9][A-Za-z0-9._-]{2,}\/?(?:[?#]\S*)?$/i.test(
```
s-annam, line 143: **Secondary — profile-band `/in/` links that aren't LinkedIn get claimed as `linkedin_url`.**

This regex admits any sole-segment `/in/<≥3 chars>`. Verified against it directly:

| URL | matches |
|---|---|
| `janedoe.com/in/design-portfolio` | ✅ |
| `acme.com/in/careers` | ✅ |
| `myntra.com/in/women` | ✅ |
| `nike.com/in/en` | ❌ (correctly rejected) |

Because the tier-2 lookup runs *before* portfolio/website, a personal portfolio hyperlinked at `janedoe.com/in/design-portfolio` in the contact block is now misfiled as LinkedIn. The doc comment names catalog paths as a known non-match risk and the profile band is the stated mitigation, so this reads as a deliberate tradeoff — but it's undocumented in the PR body and untested. Worth either a narrowing guard (require a handle containing a `-` or `.`, i.e. name-shaped) or an explicit test pinning the accepted behaviour so the next reader knows it was chosen, not missed.

src/lib/heuristics/extract/contact.ts
```diff
@@ -219,7 +269,18 @@ function extractOtherUrls(joined: string): {
...
+    // Exclude any linkedin.com URL (profile OR non-profile company/jobs/feed
+    // pages — the literal-host guard keeps those out of website_url), the
+    // `lnkd.in` shortlink host, and github.com. The literal `linkedin.com`
+    // check is load-bearing: it also catches non-profile `linkedin.com/company/…`
+    // pages, which the shortlink test does not (#378). The three terms are
+    // uniform host checks — `lnkd.in` is the one LinkedIn signal `linkedin.com`
+    // doesn't already cover.
+    if (
+      lower.includes("linkedin.com") ||
+      isLnkdInShortlink(u) ||
```
s-annam, line 291: **Nit (non-blocking)** — this `lnkd.in` exclusion has no test on the **visible-text** path; every `lnkd.in` test in the diff goes through an annotation. One case asserting a bare `lnkd.in/abc123` in visible text lands in `linkedin_url` and not `website_url` would pin it.

s-annam (changes requested):
Solid, well-documented fix — the two-tier split (host-anchored doc-wide vs. host-agnostic profile-banded) is the right shape, and the doc comments carry their rationale. One blocking regression: the diff re-introduces the exact double-render it sets out to remove, on the visible-text-plus-hyperlink path.

Gates: `check:fixtures` ✅ (51 PDFs, no binary added) · `typecheck` ✅ · `lint` ✅ · `fallow audit` ⚠️ warn-only (3 clone groups in the new tests) · design-system + style-token gates N/A (no `src/components/**` in the diff).

## Blocking

**A `/in/` redirect that appears as BOTH visible text and a link annotation now fills `linkedin_url` AND `website_url` with the same URL.** See the inline comment on the `website` resolver.

## Secondary

**Profile-band `/in/` links that aren't LinkedIn get claimed as `linkedin_url`.** See the inline comment on the redirect regex.

## Nits

- The `lnkd.in` exclusion added to `extractOtherUrls` has no test on the **visible-text** path — see inline.
- `fallow` flags 3 clone groups (28 lines) in the new tests — warn-only, not a gate failure. A small `mkAnnotation(url, yTop)` helper would clear it.

---

src/lib/heuristics/extract/contact.ts
Samhit21: Fixed in 116d031. Gave `pickUrl` an optional `reject(url)` predicate applied to EVERY tier — the text-scan `primary`/`fallback` values as well as the annotation — and `portfolio_url`/`website_url` now reject any URL already claimed by a higher-priority field. So the `\href{url}{url}` case (redirect as both visible text and annotation) fills `linkedin_url` from the annotation and leaves `website_url` empty. Comparison is on the normalized form (`claimKey`), since the text scan normalizes its values while an annotation carries the raw href. Added your exact repro as a regression test (`example.com/in/jane-doe` as text + annotation → linkedin only).

src/lib/heuristics/extract/contact.ts
Samhit21: Taken the "pin the accepted behaviour" option. A name-shape guard does not actually separate the cases — `design-portfolio` is hyphenated too, so requiring a `-`/`.` would still admit it while wrongly rejecting a legit `/in/jsmith`. Since no regex cleanly splits `/in/jane-doe` from `/in/design-portfolio`, the profile-band restriction IS the mitigation. Documented the tradeoff explicitly in the `isLinkedinRedirectUrl` docblock and added a test pinning that a profile-band `janedoe.com/in/design-portfolio` is claimed as `linkedin_url`, so the next reader sees it was chosen, not missed.

src/lib/heuristics/extract/contact.ts
Samhit21: Added a visible-text test: a bare `lnkd.in/abc123` in the contact line (no annotation) lands in `linkedin_url` and stays out of `website_url`, exercising the `extractOtherUrls` exclusion on the text path.

Samhit21:
Addressed the review on `116d031` (amended — one commit per PR).

**Blocking — double-render on the visible-text + annotation path:** root cause was that `claimedUrls` only filtered the annotation tier, while `pickUrl` returns the text-scan `primary`/`fallback` value first. Gave `pickUrl` a per-field `reject(url)` predicate applied across **every** tier; `portfolio_url` and `website_url` now reject any URL already claimed by a higher-priority field (normalized comparison, since text is normalized and annotations are raw). Added your exact `\href{url}{url}` repro as a regression test — `linkedin_url` set, `website_url` empty.

**Secondary — profile-band non-LinkedIn `/in/` claimed as LinkedIn:** a name-shape guard doesn't separate the cases (`design-portfolio` is hyphenated too), so I documented the tradeoff in the `isLinkedinRedirectUrl` docblock and pinned it with a test (`janedoe.com/in/design-portfolio` → `linkedin_url`). The profile-band restriction is the real mitigation.

**Nits:** added a visible-text `lnkd.in` test; extracted a `mkAnnotation(url, yTop)` helper to cut the annotation-literal duplication `fallow` flagged.

Gates: 22 contact tests, full suite (2488) green, corpus snapshot unchanged, typecheck + lint clean. Amend re-dismissed the approval — re-requesting review.

## Provenance
Authored with Claude Opus 4.8 (1M context).

src/lib/heuristics/extract/contact.ts
```diff
@@ -353,52 +426,103 @@ export function extractContact(
...
+        inProfileSection,
+      ),
+  );
+  const github = pickUrl("github_url", () =>
+    findAnnotationUrl(isGithubUrl, anywhereOnDoc),
+  );
+  // A URL is "the same claim" across tiers regardless of scheme/`www.`/trailing
+  // punctuation, so compare on the normalized form — the text scan normalizes
+  // its values, an annotation carries the raw href.
+  const claimKey = (u: string): string => normalizeUrl(u) ?? u;
```
s-annam: **Blocking — cross-tier dedup misses trailing-slash/case variants, reopening the double-render this PR fixes.**

`claimKey` uses `normalizeUrl`, which does not strip a trailing `/` or lowercase. But `urlSlug` (already imported here, and used 45 lines down at 536–537 for the ownership dedup) exists precisely for "the comparable identity of a link across `github.com/x` and `github.com/x/`".

Verified reproducing — annotation `https://jane.dev/in/jane/` + visible text `jane.dev/in/jane`:

```
linkedin_url  = https://jane.dev/in/jane/
portfolio_url = https://jane.dev/in/jane   <- same target, not rejected
```

The `\href{url}{url}` LaTeX case this PR targets routinely appends a canonical trailing slash on the href while the typed label omits it, so this fires inside the exact scenario the fix is for. Fix:

```ts
const claimKey = (u: string): string => urlSlug(u) ?? u;
```

s-annam (changes requested):
LinkedIn-`/in/` admission logic is well-built — the two-tier host-anchored-vs-band split is sound, the regexes correctly reject locale/catalog/mid-path/dotless cases, and the 22 added tests + the full 1029-test heuristics suite pass. One blocker: the cross-tier claim key doesn't normalize link identity the way the rest of the module does.

## Blocking
- `contact.ts:490` — see inline: `claimKey` should use `urlSlug`, not `normalizeUrl`.

## Nits (non-blocking)
- Branch is stale — the merge-base predates the `resumelint`→`offlinecv` rename (#498, ~5 merged PRs back). CI on this PR ran against a pre-rename base; rebase onto current `main` before it hits the merge queue. (The `example.com/in/jordan-bennett` fixture snapshot bump in `multi-degree-coursework.expected.json` — 86→89 with `linkedin_url` now detected — is correct and earned by the fix.)

Gates run: fixture-PII N/A (no binary added), design-system / style-token / command-bug N/A (pure lib change), typecheck + lint clean, fallow warn-only (test-clone boilerplate). Verdict rule: 1 Blocking → REQUEST_CHANGES.

---
Reviewed by: Claude Opus 4.8 (1M context) (high)

s-annam:
Pushed the one-line fix myself to save a review round (pinged Samhit first, no objection): `5f4b87d` — `claimKey` now uses `urlSlug` instead of `normalizeUrl`, plus a regression test for the trailing-slash-href vs bare-text case. Blocking finding resolved.

Still worth a rebase onto current `main` before the queue — this branch's base predates the `resumelint`→`offlinecv` rename, so CI here ran on a pre-rename base.

s-annam (approved):
Blocking finding resolved — claimKey→urlSlug fix + trailing-slash test landed in 5f4b87d. Approving.

---
Reviewed by: Claude Opus 4.8 (1M context) (high)

Issue #379, opened 2026-07-07
[edit-interaction] "+ Add education" creates a ghost entry when left completely blank on commit


Vaishnavi1709:
## Category
edit-interaction (Steps 1 & 3 F&F QC — provenance category)

## Fixture
Reproducible on any fixture — validated during F&F QC on:
- `tests/fixtures/pdfs/latex/multi-degree-coursework.pdf`

## Observed
1. Click `+ Add education` in the Education section
2. A blank editable entry appears with placeholders in every field (`degree`, `institution`, `dates`, `coursework`)
3. Click away without typing anything
4. **The blank entry persists** in the education list

No auto-remove, no "at least one field required" gate, no confirmation prompt asking whether the user meant to add an empty entry.

Related: the "no format validation" bug (`[edit-interaction] inline edits accept any string with no format validation`) applies to the `+ Add` fields too — typing literal `banana` in the new degree field commits fine.

## Category rationale
This is not a parser bug (no parse involved). It's `+ Add education`'s commit contract not enforcing "an entry must carry at least one populated field."

## Localized layer
Feature component for the education entry list — likely `src/components/features/ReconstructedEducationSkills.tsx`, wherever the `+ Add education` handler is defined. The append operation adds a new empty `ResumeEducation` object to the array and never removes it if the user commits nothing.

## Suggested fix
On the new entry's first blur / navigate-away with all fields empty, either:
1. **Auto-remove** — silently drop the empty entry
2. **Warn** — show a "discard empty entry?" confirmation
3. **Gate** — require at least `degree || institution` populated before the entry is added to the list at all

Option 1 or 3 is probably safest; option 2 adds a modal to a lightweight edit surface.

Same treatment likely needed for `+ Add experience`, `+ Add project`, `+ Add achievement`, `+ Add skill` — the pattern almost certainly repeats.

## Repro
```
npm run dev
# drop any fixture
# scroll to Education, click "+ Add education"
# click somewhere else without typing anything
# the blank entry stays in the list
```

s-annam:
Fixed in `587259d` (PR #497, merged 2026-07-17). The PR title referenced this issue without a `Closes` trailer, so it was never auto-closed.

**Shipped:**
- `useEditableParse.ts` — `isAddedEntryEmpty` (iterates `ADDED_ENTRY_FIELDS`, so new header fields are covered automatically) + `pruneEmptyAddedEntries(section)`, section-scoped.
[...]

PR #497, opened 2026-07-17
fix(edit): drop blank "+ Add" entries left empty on section exit (#379)

Samhit21:
## Summary

Clicking **+ Add education / experience / project / achievement** appends an empty entry so the click-to-edit fields have something to bind to. If the user then clicked away without typing, that blank entry **persisted** — in the list, the ATS score, and the exported PDF (#379).

The edit fields never autofocus, so a per-entry blur can't catch the primary repro (add → click away without ever focusing the entry). But the **+ Add** pill lives *inside* the section and holds focus after the click, so a **section-level focus-exit** fires reliably. On focus leaving a section, prune every added entry in it that's completely empty (no populated header field, no bullets). The prune is **deferred one tick** so an in-flight field commit — which fires on the same blur — lands first, and a just-typed entry is kept.

Pruning happens in the edit hook (removing the entry from `addedEntries`), not in `applyOverrides`, so the rendered list and the score/PDF stay index-aligned.

### Changes
- **`useEditableParse`** — `isAddedEntryEmpty` (pure, iterates `ADDED_ENTRY_FIELDS` so new header fields are covered automatically) + `pruneEmptyAddedEntries(section)`, **section-scoped** so a freshly-added entry in *another* section is never swept.
- **`ReconstructedAdd`** — shared `sectionExitBlur()` `onBlur` helper (the `currentTarget.contains(relatedTarget)` idiom already used by `InlineBulletAdd`, plus the deferred fire).
- Wired onto all four editable-entry `<section>`s.

Skills / bullets / profile adds already no-op on blank input, so only whole-entry adds could ghost — those are the four sections covered here.

## Test plan
- [x] `npm run lint` clean
- [x] `npm run test` green (2486 passed) — new tests: `isAddedEntryEmpty` (pure), `pruneEmptyAddedEntries` behavior via a probe-hook (drop blank / keep populated / keep bulleted / section-scoping / no-op identity), and `sectionExitBlur` (deferred fire + focus-containment)
- [x] `npm run build` passes

## Provenance
Authored with Claude Opus 4.8 (1M context).

### DIFF
--- a/src/components/features/ReconstructedAdd.tsx
+++ b/src/components/features/ReconstructedAdd.tsx
@@ -16,7 +16,31 @@
  */
 
 import { useState } from "react";
+import type { FocusEvent } from "react";
 import { Button } from "@design-system";
+
+/**
+ * Build an `onBlur` handler for a section container that fires `onExit` when
+ * focus leaves the section's DOM subtree (the standard
+ * `currentTarget.contains(relatedTarget)` idiom, shared with
+ * {@link InlineBulletAdd}). Used to prune a blank "ghost" entry the user opened
+ * with "+ Add …" and abandoned without typing (#379): the "+ Add" pill lives
+ * inside the section and holds focus after the click, so a focus-exit fires
+ * reliably even though the click-to-edit fields never autofocus.
+ *
+ * The prune is deferred one tick because a field commit fires on the SAME blur
+ * event that leaves the section; running `onExit` synchronously would test the
+ * entry's emptiness against pre-commit state and wrongly drop an entry the user
+ * just typed into. The macrotask lets React flush the commit first.
+ */
+export function sectionExitBlur(
+  onExit: () => void,
+): (e: FocusEvent<HTMLElement>) => void {
+  return (e) => {
+    if (e.currentTarget.contains(e.relatedTarget as Node | null)) return;
+    setTimeout(onExit, 0);
+  };
+}
 
 /** A small X glyph, matching the SkillChip remove control. */
 function CloseIcon() {
--- a/src/components/features/ReconstructedAdd.test.ts
+++ b/src/components/features/ReconstructedAdd.test.ts
@@ -1,0 +1,39 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The resumelint Authors
+
+import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
+import type { FocusEvent } from "react";
+import { sectionExitBlur } from "./ReconstructedAdd.tsx";
+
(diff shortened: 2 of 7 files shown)


src/components/features/ReconstructedAdd.tsx
```diff
@@ -16,8 +16,32 @@
...
+ * event that leaves the section; running `onExit` synchronously would test the
+ * entry's emptiness against pre-commit state and wrongly drop an entry the user
+ * just typed into. The macrotask lets React flush the commit first.
+ */
+export function sectionExitBlur(
+  onExit: () => void,
+): (e: FocusEvent<HTMLElement>) => void {
+  return (e) => {
+    if (e.currentTarget.contains(e.relatedTarget as Node | null)) return;
+    setTimeout(onExit, 0);
```
s-annam, line 41: Nit (non-blocking): an uncommitted `InlineBulletAdd` draft is lost with the entry. If a user adds a blank entry, opens "Add bullet", types text but doesn't press Enter, then clicks outside the section, `InlineBulletAdd` keeps itself expanded (draft non-empty) but never commits — while this deferred prune sees an empty entry (blank header, no committed bullets) and drops the whole thing, so the visible typed text vanishes. This is consistent with the existing Enter-to-commit-or-lose-it contract for drafts, so it's a papercut, not accounting data loss. A comment or follow-up is enough.

s-annam (approved):
**APPROVE** — clean, well-scoped fix for #379 with genuinely good tests and docblocks. The `addedBulletsRef` handles the committed-bullet-vs-deferred-prune race correctly, and deriving emptiness from `ADDED_ENTRY_FIELDS` keeps it in lockstep with replay. No blocking issues; two non-blocking nits below.

## Gates run

- **Fixture PII (3a):** skipped — no fixtures touched.
- **Design-system reuse (3b):** clean — `onBlur` sits on a semantic `<section>`, not a hand-rolled control.
- **Style tokens (3c):** clean — no hardcoded colors (the `#379` matches were the issue number).
- **Fallow / dead-code (3d):** clean for this diff — the 3 complexity findings are pre-existing/inherited and excluded from the audit gate; the new `isAddedEntryEmpty` / `pruneEmptyAddedEntries` / `sectionExitBlur` are small and each has a consumer.
- Tests 26/26 pass, typecheck + lint clean.

## Nits (non-blocking)

- **`setTimeout(onExit, 0)` is never cleared.** If the owning tree unmounts within the tick, the prune runs setState post-unmount. React 19 tolerates this and the window is a single macrotask, so impact is effectively nil — noted for completeness.

Verdict rule applied: 0 Blocking → APPROVE.

---

PR #495, opened 2026-07-15
fix(parser): round-trip the one-line middot experience header — title/company swap (#436)

rohithgollapalli:
## Summary

The Download-PDF exporter emits a one-line `Title · Company, Location · Team` header, dropping the two-line structural signal the parser used to tell title from company. A role whose two **middot** segments are both **neutral** (no company-suffix, no title-keyword) re-parsed title↔company **swapped**, because `mapWithoutCompanyMatch`'s no-signal default reads the first segment as the *company* (the stacked "Company \n Title" convention).

Captured live (parse1 → parse3): `Composer @ Northwind Ensemble` → `Northwind Ensemble @ Composer`.

A single-line **middot** header follows the opposite convention — `Title · Company` (the export shape and the dominant single-line résumé form, #217). Fix:
- Track the delimiter on each `Split` (pure-middot vs `@`/`—`/`|`).
- In the no-signal default, read a **single-source middot two-segment** split as **title-first**. Gated to the middot so pipe/em-dash/`@`/stacked shapes keep the company-first default (there's a pipe control test).

## Scope — the SWAP root only

#436 has two roots. This PR lands the **swap**; the **truncation** root (a parenthetical / multi-word company, `Danggeun Pay Inc. (KarrotPay)` → `(KarrotPay)`) is distinct and still open.

- **7 fixtures round-trip** and drop their `corpus-roundtrip` baselines (the ratchet forced it): deedy ×2, chromium-two-column-sidebar, qualified-experience-relevant-coursework, single-column-year-only, synthetic-two-experience-sections, two-column-achievements-sidebar.
- **Un-skipped** the #358 year-only repro (`render-roundtrip-year-only.repro`).
- **Still suspended** (truncation root, documented): the #284 AC#3 repro and the remaining 9 baselines (awesome-cv ×2, weasyprint-two-column, chromium-asymmetric, the 5 google-docs).

`Refs #436` (not `Closes` — truncation root remains).

## Test plan

- [x] New unit tests (`experience.mid-dot.test.ts`): neutral middot header keeps `Title · Company` order (no swap); a pipe header keeps company-first (flip is middot-only).
- [x] `corpus-roundtrip.test.ts` green — 7 baselines removed, ratchet satisfied; 9 remain (truncation root).
- [x] `corpus.test.ts` green — the 7 fixed fixtures' `experienceChangedAcrossRoundtrip` flips `true → false` (round-trip now holds); no forward-parse or `extractedCharCount` change.
[...]

PR #505, opened 2026-07-20
fix(parser): disambiguate two more one-line experience header shapes (#436)

rohithgollapalli:
## Summary

Builds on **#495** (which fixed the neutral middot swap, `Composer · Northwind Ensemble`). Two sibling shapes on the same one-line `Title · Company, Location · Team` export header still re-parsed wrong; both are fixed here **without touching the renderer**:

- **Soft-vs-hard company swap** — when both middot segments read like a company because the title carries a soft company word (`Solutions Engineer`) and the company a hard legal tail (`Acme Cloud, Inc.`), `chooseCompanyIdx` now prefers the hard legal-entity tail. #495's title-first default can't fire here — a company match routes to `mapWithCompanyMatch`. Gated to a single-source **middot** line so stacked / two-column headers are untouched.
- **Location bleed** — the exporter joins `Company, Location` in one cell, so a bare-location tail with no `City, ST` shape (`…, Remote`, `…, London`) bled into `company`. `stripLocationSuffix` **Pass F** peels a trailing tail that whole-matches the closed `BARE_LOCATION_RE` vocab.

Clears **5** corpus baselines (`certifications`, `achievements-oneline`, `two-column`, `chromium-asymmetric`, `weasyprint-two-column`) with **zero round-trip regressions**, plus two value-asserting unit tests.

Refs #436

### Deferred: the TRUNCATION root

A parenthetical company (`Danggeun Pay Inc. (KarrotPay)` → `(KarrotPay)`) still re-parses truncated. Its fix needs a wrapped-header rejoin **plus** a render-side date-column reserve — both perturb two-column round-trips (measured: 4+ regressions), so it's split into its own follow-up. The `awesome-cv-cv/resume` + `role-first` + `honors-subheadings` baselines stay.

> Note: opened before the ResumeLint→OfflineCV rebrand + #495 landed; rebased onto current `main`, so the diff is now a focused 3-file change layered on #495.

## Test plan

- [x] `npm run lint` clean
- [x] `extract/` + `corpus-roundtrip` + `pdf/` suites — 543 pass / 2 skip, zero regressions
- [x] `npm run verify` — green in CI

## Provenance

### DIFF
--- a/src/lib/heuristics/extract/experience-disambiguate.ts
+++ b/src/lib/heuristics/extract/experience-disambiguate.ts
@@ -300,6 +300,26 @@ function stripLocationSuffix(s: string): {
     }
   }
 
+  // Pass F — comma-delimited trailing BARE-LOCATION keyword (#436). The one-line
+  // experience exporter joins "Company, Location" into a single middot cell, and
+  // when the model's `location` is a work-mode keyword ("Remote", "Hybrid",
+  // "On-site") or a bare well-known city with no state/country tail ("London",
+  // "Seattle") — the forms `BARE_LOCATION_RE` recognizes but Passes A–E (which
+  // all require a "City, ST" / "City, Country" shape) do not — the location bled
+  // into `company` on round-trip ("Globex Corporation, Remote"). Peel the tail
+  // when it whole-matches the closed `BARE_LOCATION_RE` vocabulary; a real
+  // company never ends in a bare ", Remote" / ", London" tail, so the closed set
+  // keeps this from stealing company text. Tried LAST so the richer city+state
+  // and city+country shapes above still win when they apply.
+  const commaF = s.lastIndexOf(",");
+  if (commaF > 0) {
+    const tail = s.slice(commaF + 1).trim();
+    if (BARE_LOCATION_RE.test(tail)) {
+      const before = stripDanglingSeparator(s.slice(0, commaF));
+      if (before) return { text: before, location: tail };
+    }
+  }
+
   return { text: s, location: undefined };
 }
 
@@ -540,6 +560,40 @@ function mapSegmentsToFields(
         (i) => splits[i].source === anchorIdx,
       );
       if (anchorMatch !== undefined) return anchorMatch;
+    }
+    // #436 (truncation root) — a single-line MIDDOT header "Title · Company"
+    // (no line above) where BOTH segments read like a company: the title carries
+    // a SOFT company word ("Solutions Engineer" → `Solutions`) and the real
+    // company carries a HARD legal-entity marker ("Acme Cloud, Inc." → `Inc.`).
+    // Without `hasAbove` the anchor tiebreak can't fire, so the topmost
+    // `companyMatchIdxs[0]` (the title) would be mis-labelled the company,
+    // swapping title↔company — the #495 middot title-first default never runs
+    // because `companyIdx !== -1` routes to `mapWithCompanyMatch`. Prefer the
+    // hard-legal segment: it is a strictly stronger company signal than a soft
+    // keyword. `COMPANY_LEGAL_TAIL_RE` is the same strict promotion vocab
+    // `mapTitleFirst` case 3a uses; a soft-only tie keeps the topmost default.
+    //
+    // GATED to the exact export shape — every company-match on ONE source line
+    // that split as `middot` — so a stacked two-line header (matches on
+    // different sources) and a non-middot single line are untouched. When the
+    // topmost match is itself the hard-legal one, `find` returns it and the
+    // result equals the default, so no extra guard is needed. Without this gate
+    // the preference
+    // fired on genuine two-column/stacked layouts and stole the company (#436
+    // review).
+    const allMatchesShareMiddotLine =
+      companyMatchIdxs.every(
+        (i) =>
+          splits[i].source === splits[companyMatchIdxs[0]].source &&
+          splits[i].middot,
+      );
+    if (allMatchesShareMiddotLine) {
+      const hardLegalIdx = companyMatchIdxs.find((i) =>
+        COMPANY_LEGAL_TAIL_RE.test(
+          splits[i].text.trim().split(/[\s,]+/).pop() ?? "",
+        ),
+      );
+      if (hardLegalIdx !== undefined) return hardLegalIdx;
     }
     return companyMatchIdxs[0];
   };
--- a/src/lib/heuristics/corpus-roundtrip.test.ts
+++ b/src/lib/heuristics/corpus-roundtrip.test.ts
@@ -150,30 +150,32 @@ const KNOWN_FAILURES: Record<string, Category[]> = {
   //   1. SWAP — a neutral two-segment middot header ("Composer · Northwind
   //      Ensemble", no company-suffix / title-keyword either side) re-parsed
   //      title↔company-swapped because the no-signal default read the first
-  //      segment as the company. FIXED: the `middot` title-first default in
-  //      `mapWithoutCompanyMatch` reads "Title · Company" per the export/#217
-  //      convention. Seven fixtures cleared their baseline via the ratchet.
+  //      segment as the company. FIXED (#495): the `middot` title-first default
+  //      in `mapWithoutCompanyMatch`. A related swap where BOTH segments look
+  //      like a company on one middot line — a soft title keyword vs a hard legal
+  //      tail ("Solutions Engineer · Acme Cloud, Inc.") — is now also fixed by
+  //      the `chooseCompanyIdx` hard-legal preference (certifications cleared).
+  //      A bare-location tail the exporter joins into the company cell
+  //      ("…, Remote") is peeled by `stripLocationSuffix` Pass F (achievements-
+  //      oneline, two-column, chromium-asymmetric, weasyprint-two-column cleared).
   //   2. TRUNCATION — a PARENTHETICAL / multi-word company on the one-line shape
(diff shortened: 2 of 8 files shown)


src/lib/heuristics/extract/experience-disambiguate.ts
```diff
@@ -541,6 +561,38 @@ function mapSegmentsToFields(
...
+    // keyword. `COMPANY_LEGAL_TAIL_RE` is the same strict promotion vocab
+    // `mapTitleFirst` case 3a uses; a soft-only tie keeps the topmost default.
+    //
+    // GATED to the exact export shape — every company-match on ONE source line
+    // that split as `middot`, and the topmost match not already the legal one —
+    // so a stacked two-line header (matches on different sources) and a
+    // non-middot single line are untouched. Without this gate the preference
+    // fired on genuine two-column/stacked layouts and stole the company (#436
+    // review).
+    const allMatchesShareMiddotLine =
```
s-annam, line 584: **Secondary — convention tradeoff, not blocking.** This promotion is load-bearing for the fix and correct on the export domain, but it commits every one-line middot header to title-first. Verified OLD→NEW:

- `"Solutions Engineer · Acme Cloud, Inc."` → OLD `title="Acme Cloud, Inc."` (wrong), NEW `title="Solutions Engineer"` ✅
- `"Acme Solutions · Director, Holdings"` → OLD `company="Acme Solutions"`, NEW `company="Director, Holdings"`

The two inputs are signal-identical (soft-company word vs hard-legal token), so you can't fix the first without flipping the second — one convention. Only bites when a role *title* ends in a literal `Holdings`/`Corp`/`Inc` token: never in the export domain (a title never ends in `Inc.`), vanishingly rare in real résumés, and no corpus fixture hits it (full suite green). Fine to accept; a one-line company-first fixture would lock the intended tie-break so a future change is a visible diff.

src/lib/heuristics/extract/experience-disambiguate.ts
```diff
@@ -541,6 +561,38 @@ function mapSegmentsToFields(
...
+    // Without `hasAbove` the anchor tiebreak can't fire, so the topmost
+    // `companyMatchIdxs[0]` (the title) would be mis-labelled the company,
+    // swapping title↔company — the #495 middot title-first default never runs
+    // because `companyIdx !== -1` routes to `mapWithCompanyMatch`. Prefer the
+    // hard-legal segment: it is a strictly stronger company signal than a soft
+    // keyword. `COMPANY_LEGAL_TAIL_RE` is the same strict promotion vocab
+    // `mapTitleFirst` case 3a uses; a soft-only tie keeps the topmost default.
+    //
+    // GATED to the exact export shape — every company-match on ONE source line
+    // that split as `middot`, and the topmost match not already the legal one —
```
s-annam: **Nit.** The comment says "and the topmost match not already the legal one," but `allMatchesShareMiddotLine` / the `find` below have no such condition. Harmless — returning the topmost when it *is* the legal one equals the default — but the comment describes a guard the code doesn't implement. Trim the clause or add the check.

src/lib/heuristics/extract/experience-disambiguate.ts
```diff
@@ -300,6 +300,26 @@ function stripLocationSuffix(s: string): {
...
+  // all require a "City, ST" / "City, Country" shape) do not — the location bled
+  // into `company` on round-trip ("Globex Corporation, Remote"). Peel the tail
+  // when it whole-matches the closed `BARE_LOCATION_RE` vocabulary; a real
+  // company never ends in a bare ", Remote" / ", London" tail, so the closed set
+  // keeps this from stealing company text. Tried LAST so the richer city+state
+  // and city+country shapes above still win when they apply.
+  const commaF = s.lastIndexOf(",");
+  if (commaF > 0) {
+    const tail = s.slice(commaF + 1).trim();
+    if (BARE_LOCATION_RE.test(tail)) {
```
s-annam, line 317: **Nit.** Pass F peels any last-comma tail that whole-matches the closed `BARE_LOCATION_RE` city vocab, so a real employer whose name ends `", London"`/`", Boston"` (e.g. `Lloyd's, London`) splits company/location. Extracted location is usually still semantically correct, and it's gated to the last comma + a closed vocab, so low-risk — a fixture would document the intended boundary.

s-annam (dismissed):
Focused, well-scoped fix layered cleanly on #495. Verified locally: typecheck + lint clean, full suite 2488 pass / 9 skip, all 5 rebaked snapshots move `experienceChangedAcrossRoundtrip` **true→false** (pure improvement, **zero regressions** — matches the PR claim). The 6 dropped `KNOWN_FAILURES` baselines all round-trip clean under the ratchet. The deferred-truncation scoping is correct and its "both changes perturb two-column round-trips" rationale reproduces independently — right call to split it out rather than trade working fixtures for new ones.

Gates: fixture-PII **N/A** (no binaries), design-system/style **N/A** (pure lib diff), fallow's 4 complexity findings are all pre-existing/inherited (`stripLocationSuffix` grows with Pass F but stays under the gate — CI `fallow` green). One Secondary + two Nits, inline. None blocking.

**Verdict rule applied:** 0 Blocking → APPROVE. The Secondary is a documented convention tradeoff, not a merge-blocker.

## Secondary

**The hard-legal promotion commits one-line middot headers to title-first — a deliberate convention choice with a rare, undecidable edge.** Verified OLD→NEW on contrived inputs: `"Solutions Engineer · Acme Cloud, Inc."` OLD→`title="Acme Cloud, Inc."` (wrong) / NEW→`title="Solutions Engineer"` (correct) — the fix works. But the *same* mechanism flips `"Acme Solutions · Director, Holdings"` from `company="Acme Solutions"` to `company="Director, Holdings"`. The two inputs are signal-identical (soft-company word vs hard-legal token), so there's no way to keep one "company-first" while fixing the other — it's one convention. This only bites when a role *title* ends in a literal legal-entity token (`Holdings`/`Corp`/`Inc`), which never occurs in the export domain a title is drawn from and is vanishingly rare in real résumés (no corpus fixture hits it). Accepting as-is is reasonable; worth a one-line company-first fixture to lock the intended behavior so a future change to the tie-break is a visible diff. (inline at line 582)

## Nits

- Gate comment overstates the guard (line 577).
- Pass F false-positive boundary on a company ending in a listed city (line 317).

---

src/lib/heuristics/extract/experience-disambiguate.ts
rohithgollapalli: Added a lock test pinning the accepted tie-break: `Acme Solutions · Director, Holdings` → `title="Acme Solutions"`, `company="Director, Holdings"`. Now a future change to the hard-legal promotion is a visible diff instead of a silent flip. (`experience.mid-dot.test.ts`)

src/lib/heuristics/extract/experience-disambiguate.ts
rohithgollapalli: Fixed — trimmed the comment. The "topmost match not already the legal one" clause described a guard the code never had; when the topmost match *is* the legal one, `find` returns it and the result equals the default, so no guard is needed. Comment now says that.

src/lib/heuristics/extract/experience-disambiguate.ts
rohithgollapalli: Added a lock test for the boundary: `Underwriter · Lloyd's, London` splits `company="Lloyd's"`, `location="London"` — extracted location stays semantically correct, and the split is pinned so a change to `BARE_LOCATION_RE` shows as a visible diff. (`experience.mid-dot.test.ts`)

s-annam (approved):
**Approve.** Two well-scoped, tightly-gated one-line-header disambiguations (#436). Both logic paths verified against the file; narrow tests + full corpus round-trip pass; CI `verify`/`fallow`/Cloudflare all green. Verdict rule: 0 Blocking → APPROVE.

## What I checked

- **`stripLocationSuffix` Pass F** — peels a comma-trailing bare-location tail (`", Remote"`, `", London"`) that the one-line exporter bled into `company`. `lastIndexOf(",")`, whole-match against the closed `^…$`-anchored `BARE_LOCATION_RE`, `stripDanglingSeparator` on the remainder, non-empty guard, tried **last** so the richer city+state/country passes still win. A real company ending in a bare `", Remote"` tail is the only false-peel risk, and the closed vocabulary bounds it. Correct.
- **`chooseCompanyIdx` hard-legal preference** — when two segments on one `middot` line both read like a company (soft title keyword vs hard legal tail, `"Solutions Engineer · Acme Cloud, Inc."`), prefers the `COMPANY_LEGAL_TAIL_RE` segment instead of the topmost default. Gated to `allMatchesShareMiddotLine` so stacked two-line and non-middot headers are untouched; `.split(/[\s,]+/).pop() ?? ""` guards the empty case. Correct.
- **Ratchet + expected.json** — the 5 flips are `experienceChangedAcrossRoundtrip: true → false` and the `KNOWN_FAILURES` entries are **removed** (fixtures now pass, forced by the stale-entry check), not re-baselined to hide a regression. Right direction.
- **Fixture PII** — N/A, no fixture binaries added/changed (only `.expected.json`). Design-system / style-token / command-bug gates N/A (parser lib + tests only).

## Nits (non-blocking)

- `stripLocationSuffix` is now HIGH on fallow (128 lines / 23 cyclomatic) after Pass F, and `mapSegmentsToFields`' `chooseCompanyIdx` keeps growing. fallow is report-only here (CI green), so not a blocker — but both `stripLocationSuffix` (six passes A–F) and the company-choice tiebreak chain are getting long enough that the next header shape will be easier to add if each pass is extracted to a named helper. Worth a follow-up, not this PR.

---

Issue #364, opened 2026-07-07
[parser] education — degree line reused as institution when both share one line


Vaishnavi1709:
## Category
parser (Steps 1 & 3 F&F QC — provenance category)

## Fixture
`tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-achievements-oneline.pdf` (synthetic persona)

## Ground truth (`pdftotext`)
```
EDUCATION
B.S. in Computer Science — State University
2013
```

## Observed
`probe-education` (this branch, RL_EDUCATION_PDF):
```
[0] degree="B.S."
    field="Computer Science — State University"                ← institution bled into field
    institution="B.S. in Computer Science — State University"  ← raw degree line reused verbatim
    dates=null → "2013"  year="2013"
```

Reconstructed view renders this as one visually-doubled line:
`B.S. , Computer Science — State University — B.S. in Computer Science — State University`

## Localized layer
`src/lib/heuristics/extract/education.ts` — the `educationFromChunk` institution fallback (~lines 568-594). When the degree line is ALSO the only `INSTITUTION_HINTS`-matching line (single-line degree + school), `chunk.find((l) => INSTITUTION_HINTS.test(l))` returns the same line already used as `degreeLine`, and no degree-strip is applied — the raw line becomes the institution verbatim.

## Suggested fix
When `instLine === degreeLine`, apply the same DEGREE_RE strip + separator cleanup used in the final `else if (degreeMatch)` fallback so the institution collapses to `"State University"`.

## Repro
```
RL_EDUCATION_PDF=/abs/path/to/tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-achievements-oneline.pdf \
  npx vitest run src/lib/heuristics/probe-education.test.ts
```

PR #507, opened 2026-07-21
fix(parser): recover degree, field, and coursework on one-line education entries (#506)

### LINKED ISSUE #506: Education parser drops degree, field, and coursework on one-line entries with en-dash date ranges
## Summary

Two independent education-parser bugs both fire on a common single-column résumé shape: an education entry whose **degree, field, and institution sit on one line** with a **trailing en/em-dash date range**, followed by a **plain-text (non-bullet) `Coursework:` line**. On this shape the parser drops the **degree**, the **field**, and the entire **coursework list** — three of the four education fields are lost, and only the institution survives.

Both were found running a real résumé through `/probe-resume` + `/probe-education` and then reproduced with a synthetic unit input (below). Neither is covered by an existing corpus fixture.

## Reproduction (synthetic — no PII)

`extractEducation` input lines:

```
M.S. Data Science, Example State University Aug 2021 – May 2023
Coursework: Databases, Machine Learning, Statistics
B.Tech. Electronics Engineering, Sample Institute of Technology Jun 2016 – May 2020
Coursework: Data Structures, RDBMS, Software Engineering
```

Actual output:

```json
[
  { "institution": "M.S. Data Science, Example State University", "degree": "", "start_date": "Aug 2021", "end_date": "May 2023", "year": "2023" },
  { "institution": "B.Tech. Electronics Engineering, Sample Institute of Technology", "degree": "", "start_date": "Jun 2016", "end_date": "May 2020", "year": "2020" }
]
```

Expected output:

```json
[
  { "institution": "Example State University", "degree": "M.S.", "field": "Data Science", "coursework": ["Databases", "Machine Learning", "Statistics"], "start_date": "Aug 2021", "end_date": "May 2023", "year": "2023" },
  { "institution": "Sample Institute of Technology", "degree": "B.Tech.", "field": "Electronics Engineering", "coursework": ["Data Structures", "RDBMS", "Software Engineering"], "start_date": "Jun 2016", "end_date": "May 2020", "year": "2020" }
]
```

Note the `degree` came back `""` (empty) and `field` / `coursework` are absent entirely. The degree token even bleeds downstream: the `.Tech` in `B.Tech.` is separately misread by the contact extractor as a website URL (`website_url = "https://B.Tech"`) — tracked separately, not in scope here, but it disappears once the degree is parsed off correctly.

---

## Bug A — degree + field stranded by a dash-split that fires on the DATE range

**File:** `src/lib/heuristics/extract/education.ts`, `educationFromChunk`, the `instLine === degreeLine` branch at **lines 722–747**.

**Root cause.** When degree and institution are on one line, `instLine` resolves to the degree line itself (both match `INSTITUTION_HINTS` via `University`/`Institute`, and there is no separate institution line). That triggers the #364 one-line `Degree — Institution` splitter:

```ts
// education.ts:727
const parts = instLine.split(/\s+[–—]\s+/);
```

This assumes an en/em-dash separates **degree from institution**. But the input's only en-dash is inside the **date range** `Aug 2021 – May 2023`. So:

- `parts[0]` = `"M.S. Data Science, Example State University Aug 2021"`
- `parts[1]` = `"May 2023"`

`hintIdx` picks `parts[0]` (it carries `University`) as the institution, and the head is re-parsed from `parts[1]`:

```ts
const head = parts.filter((_, i) => i !== instIdx).join(" — "); // "May 2023"
({ degree, field } = parseDegreeAndField(head));                // degree="", field=undefined
```

`parseDegreeAndField("May 2023")` finds no `DEGREE_RE` → `degree = ""`, `field` undefined. The start-date fragment glued into `parts[0]` is later peeled by `stripInstitutionDate` (line 768), which is why `institution` looks clean but the credential is gone.

**The ordering is the bug:** `stripInstitutionDate` runs at line 768, *after* the dash-split at 727. The trailing date range must be removed **before** the split, or the split must not fire when its right-hand part is a bare date.

**Fix (either, prefer the first):**

1. Peel the trailing date off `instLine` before the `instLine === degreeLine` split — reuse `stripInstitutionDate(instLine)` at the top of that branch so the only remaining en/em-dash is a genuine degree–institution separator.
2. Guard the split: skip it when the right-hand `parts[instIdx]`/head is date-only (`isDateOnlyLine`, already imported from `../entry-blocks.ts`), falling back to the raw-line-as-institution path at line 746.

Whichever is chosen, the branch must still handle the genuine #364 shape it was written for (`B.S. in Computer Science — State University`, dash separating degree from school) — that case has no trailing date, so peeling the (absent) date is a no-op and the split still fires correctly.

---

## Bug B — coursework on a plain-text (non-bullet) line is never collected

**File:** `src/lib/heuristics/extract/education.ts`, `extractEducation`, the coursework-recovery loop at **lines 828–869**.

**Root cause.** Coursework recovery only runs on bullet lines:

```ts
// education.ts:829
for (let i = 0; i < ls.length; i++) {
  if (!isBulletLine(ls[i])) continue;   // ← plain-text "Coursework: …" skipped
  ...
```

The résumé shape here writes coursework as plain text — `Coursework: Databases, Machine Learning, Statistics` with **no `•`/`●`/`-` glyph**. `isBulletLine` returns false, so the line is never harvested as coursework. It then falls through to `rawEntryLines`, isn't a degree/institution/date line, and is silently discarded by `educationFromChunk` (which selects the institution as the first non-degree, non-date line and ignores the rest).

Note the machinery to *parse* such a line already exists and is applied a few lines down (line 852): `item.replace(COURSEWORK_LABEL_RE, "")` peels the `Coursework:` label, and `COURSEWORK_LABEL_RE` (line 171) matches exactly this plain prefix. The gap is only the **entry condition** — the label-prefixed plain line never reaches that code because the loop is bullet-gated.

**Fix.** Broaden the recovery entry condition so a non-bullet line that begins with `COURSEWORK_LABEL_RE` is also treated as a coursework source line:

```ts
if (!isBulletLine(ls[i]) && !COURSEWORK_LABEL_RE.test(ls[i].text)) continue;
```

Then `stripBullet` is a no-op on the plain line (no glyph to strip), `COURSEWORK_LABEL_RE` peels the label as it already does, and the comma-split + Title-case guard + per-entry attribution-by-line-index all run unchanged. The existing bullet path is unaffected. Verify the Title-case first-item guard (line 864) and the wrapped-continuation absorb (lines 838–846) behave sensibly on a plain label line — a label-prefixed line's first course is Title-case, so it passes.

---

## Acceptance criteria

- [ ] The synthetic input above yields two entries with `degree` (`"M.S."`, `"B.Tech."`), `field` (`"Data Science"`, `"Electronics Engineering"`), and `institution` (`"Example State University"`, `"Sample Institute of Technology"`) all populated.
- [ ] Both entries carry their `coursework` arrays (3 items each), attributed to the correct entry by line position.
- [ ] Dates continue to parse correctly (`start_date`/`end_date`/`year` unchanged from current behavior).
- [ ] Bug A regression: the genuine #364 one-line `Degree in Field — Institution` shape (dash separates degree from school, no trailing date) still splits correctly — covered by the existing `#364` / `#222` cases in `src/lib/heuristics/extract/education.test.ts`.
- [ ] Bug B regression: existing bullet-style coursework (`● Course`, `- Course`) still recovered; the `#184` over-consume guard and `#367` label-dedup still hold.
- [ ] `npx vitest run src/lib/heuristics/extract/education.test.ts` green, with new value-asserting cases pinning both fixed shapes.
- [ ] Corpus invariants hold: `npx vitest run src/lib/heuristics/corpus.test.ts src/lib/heuristics/corpus-roundtrip.test.ts` green.
- [ ] A synthetic corpus fixture (single-column, one-line degree entries with en-dash date ranges + plain-text `Coursework:` lines) is minted per `tests/fixtures/pdfs/README.md`, baked via `npm run bake-fixtures`, and pinned with a `*.repro.test.ts` — this shape is currently uncovered by all 51 fixtures.

## Notes

- Both bugs live entirely in `src/lib/heuristics/extract/education.ts`; no UI/design-system surface is touched, so no Reuse analysis applies.
- The two fixes are independent and can land in one PR — they share the same repro fixture and the same test file.

rohithgollapalli:
## Summary

A single-column education entry with **degree + field + institution on one line** and a **trailing `Mon YYYY – Mon YYYY` date range**, optionally followed by a **plain (glyph-less) `Coursework:` line**, dropped three of its four fields on parse — only the institution survived. Found running a real résumé through `/probe-resume` + `/probe-education`; reproduced synthetically. Both bugs live in `src/lib/heuristics/extract/education.ts`.

- **Bug A — degree/field stranded.** The #364 one-line `Degree — Institution` splitter fired on the en-dash *inside the date range* (`Aug 2021 – May 2023`), burying the credential in `institution` and re-parsing degree from the bare end-date (`"May 2023"` → empty). Fix: peel the trailing date **before** the split; when no en/em-dash separates degree from institution, fall back to a **comma split** (`<Degree Field>, <Institution>`) gated on a trailing institution-hint part at index ≥ 1 — so a comma *inside* the institution name (`University of California, Berkeley`) and an institution-first line (`State University Boston, MA`, no degree prefix) are both left intact.
- **Bug B — coursework dropped.** Recovery was bullet-gated, so a plain glyph-less `Coursework: …` line was never collected. Fix: admit a `COURSEWORK_LABEL_RE`-prefixed line into the same label-strip + comma-split path.

Closes #506

## Corpus impact

Re-baked 11 goldens whose one-line comma education entries now parse cleanly (e.g. `shared-employer-banner-roles`: institution `"B.S. Computer Science, State University"` → `"State University"`, field `"Computer Science, State University"` → `"Computer Science"`). Audited the golden diff: the **only** moved keys are `extractedCharCount` (×11, the shortened field text) and one `confidence` that ticks **up** (0.82 → 0.84). No parsed-count, score, or round-trip-signal regressions.

## Test plan

- [x] `npx vitest run src/lib/heuristics/extract/` — green (6 new value-asserting cases pin both fixed shapes + the two guard cases)
- [x] `npx vitest run src/lib/heuristics/corpus.test.ts src/lib/heuristics/corpus-roundtrip.test.ts` — green, zero round-trip regressions
- [x] `npm run lint` clean on touched files

## Provenance

Verification: local typecheck + lint + extract/corpus/roundtrip suites green (full `npm run verify` runs in CI)

### DIFF
--- a/src/lib/heuristics/extract/education.ts
+++ b/src/lib/heuristics/extract/education.ts
@@ -720,11 +720,19 @@ function educationFromChunk(chunk: string[]): {
       // reconstructed view. Split at the em/en-dash separator so the trailing
       // half is the institution and re-parse degree/field off the head.
       if (instLine === degreeLine && degreeMatch) {
+        // Peel any trailing date range off the one-line entry FIRST, so an
+        // en/em-dash INSIDE the date range ("… University Aug 2021 – May 2025",
+        // #506) is never mistaken for the degree↔institution separator. Without
+        // this the split below cut the entry on the date's en-dash, burying the
+        // credential in `institution` and re-parsing degree/field from the bare
+        // end-date ("May 2025") — degree came back empty. `stripInstitutionDate`
+        // runs again downstream (line ~768); it is idempotent.
+        const instNoDate = stripInstitutionDate(instLine);
         // En/em-dash only — a spaced ASCII `-` commonly separates degree from
         // field ("B.S. - Computer Science, Stanford University"), not
         // institution from the rest, so splitting on it here would strand the
         // field on the wrong side of the boundary.
-        const parts = instLine.split(/\s+[–—]\s+/);
+        const parts = instNoDate.split(/\s+[–—]\s+/);
         if (parts.length >= 2) {
           // Pick the part carrying an institution hint ("University", "College",
           // …) as the institution. If none carries a hint, fall back to the
@@ -737,13 +745,31 @@ function educationFromChunk(chunk: string[]): {
           const head = parts.filter((_, i) => i !== instIdx).join(" — ");
           ({ degree, field } = parseDegreeAndField(head));
         } else {
-          // No em-dash separator to slice on — keep the raw line as the
-          // institution (behavior preserved from before #364; the strip-then-
-          // maybe-mangle path only pays off when the split cleanly isolates
-          // the institution). A comma-separated single-line entry like
-          // "Associate degree, H.R. Management, Bellows College" round-trips
-          // consistently that way.
-          institution = instLine.trim();
+          // No em/en-dash separator. Try a COMMA separator
+          // "<Degree Field>, <Institution>" (#506) — the shape "M.S. Data
+          // Science, Example State University". Split on commas and take the run
+          // from the FIRST hint-bearing part onward as the institution, so a
+          // comma INSIDE the institution name ("University of California,
+          // Berkeley") is preserved, re-parsing degree/field off the head.
+          // Gated on a hint part existing AT INDEX ≥ 1, so there is a non-empty
+          // "<Degree Field>" head before it to re-parse. `cHintIdx === 0` means
+          // the line LEADS with the institution ("State University Boston, MA",
+          // no degree prefix) — there is nothing to split off, so it falls to the
+          // raw-line fallback unchanged rather than nuke degree/field to empty.
+          // An acronym-school single-line entry ("B.S. Computer Science, MIT")
+          // carries no hint and likewise falls through untouched.
+          const commaParts = instNoDate.split(/\s*,\s*/);
+          const cHintIdx = commaParts.findIndex((p) => INSTITUTION_HINTS.test(p));
+          if (cHintIdx >= 1) {
+            institution = commaParts.slice(cHintIdx).join(", ").trim();
+            const head = commaParts.slice(0, cHintIdx).join(", ");
+            ({ degree, field } = parseDegreeAndField(head));
+          } else {
+            // Keep the (date-stripped) raw line as the institution (behavior
+            // preserved from before #364; the strip-then-maybe-mangle path only
+            // pays off when the split cleanly isolates the institution).
+            institution = instNoDate.trim();
+          }
         }
       } else {
         institution = instLine.trim();
@@ -826,7 +852,12 @@ export function extractEducation(
   const coursework: { text: string; idx: number }[] = [];
   const consumed = new Set<number>();
   for (let i = 0; i < ls.length; i++) {
-    if (!isBulletLine(ls[i])) continue;
+    // Coursework is usually a bullet list, but some résumés write it as a plain
+    // (glyph-less) label line — "Coursework: Databases, Machine Learning" (#506).
+    // Admit such a line too, or it never reaches the label-strip + comma-split
+    // below and is silently dropped. `stripBullet` is a no-op on a glyph-less
+    // line, so the bullet path is unaffected.
+    if (!isBulletLine(ls[i]) && !COURSEWORK_LABEL_RE.test(ls[i].text)) continue;
     let item = stripBullet(ls[i].text);
     const span = [i];
     let j = i + 1;
--- a/src/lib/heuristics/extract/education.test.ts
+++ b/src/lib/heuristics/extract/education.test.ts
@@ -769,3 +769,91 @@ describe("extractEducation — entry-boundary guard against phantom entries (#46
     expect(value[1].degree).toContain("B.E.");
   });
 });
+
+describe("extractEducation — one-line 'Degree Field, Institution' + en-dash date + plain coursework (#506)", () => {
+  it("keeps the degree/field when the only en-dash is inside a trailing date RANGE", () => {
+    // The #506 repro: degree + field + institution on ONE line, with a trailing
+    // "Mon YYYY – Mon YYYY" range. The date's en-dash used to be mistaken for the
+    // degree↔institution separator, burying the credential in `institution` and
+    // re-parsing degree from the bare end-date ("May 2023") → empty.
+    const { value } = extractEducation(
+      mkEduSection([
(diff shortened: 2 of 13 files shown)


src/lib/heuristics/extract/education.ts
```diff
@@ -737,13 +745,31 @@ function educationFromChunk(chunk: string[]): {
...
+          // "<Degree Field>" head before it to re-parse. `cHintIdx === 0` means
+          // the line LEADS with the institution ("State University Boston, MA",
+          // no degree prefix) — there is nothing to split off, so it falls to the
+          // raw-line fallback unchanged rather than nuke degree/field to empty.
+          // An acronym-school single-line entry ("B.S. Computer Science, MIT")
+          // carries no hint and likewise falls through untouched.
+          const commaParts = instNoDate.split(/\s*,\s*/);
+          const cHintIdx = commaParts.findIndex((p) => INSTITUTION_HINTS.test(p));
+          if (cHintIdx >= 1) {
+            institution = commaParts.slice(cHintIdx).join(", ").trim();
```
s-annam, line 764: Nit (non-blocking): the comma-split leaves a trailing comma when the source line ends in one — `"B.S. CS, Springfield State University,"` → institution `"Springfield State University,"`. A `commaParts.filter(Boolean)` before the slice drops the empty tail. Pre-existing on main, so not a regression; just a cheap cleanup.

s-annam (approved):
Solid, well-tested, well-documented fix. Verified the parser output main-vs-branch on 5 of the affected fixtures — every one improves (institution stops swallowing the degree/field/whole-line; the uniform `extractedCharCount` drop in the snapshots reflects that de-doubling, not a regression). 56 unit + 51 corpus tests pass, typecheck + lint clean.

## Nits (non-blocking)
- chanchal fixture now yields `field: "degree, H.R. Management"` — the stray `"degree"` is a `parseDegreeAndField` limitation on `"Associate degree, …"`, surfaced (not caused) by this change. Still strictly better than main, which glued the institution into `field`.

Nice touch stripping the date before the em-dash split and admitting glyph-less `Coursework:` lines — both edges are covered by tests.

---

PR #512, opened 2026-07-21
fix(experience): recognize en-dash as the Title–Company header separator

rohithgollapalli:
## Summary

Single-column Word / Google-Docs résumés join a role's title to its employer with a spaced **en-dash** (`–`, U+2013), exactly like the **em-dash** (`—`, U+2014) the Phase-1 header split already recognizes — but the en-dash was missing from that delimiter set, so the whole header collapsed into one segment. A company-suffix header dropped its title to `null`; a suffix-less header lost its company.

The fix adds a **guarded** `splitEnDashTitleCompany`: it splits `Title – Company`, but leaves the line unsplit when the tail reads like a bare location, so the role-first `Company – City, ST` shape (#215/#436) — which uses the same spaced en-dash to join a company to its city and is owned by the location-recovery path — is untouched. Dates are stripped upstream, so a spaced en-dash surviving on a header line is always a field separator, never a date range.

Surfaced by a real résumé whose two roles both mis-mapped; reproduced with a **synthetic-persona** fixture (`Dana Whitfield` / `@example.com`), never the real PDF.

## What else moved

- Clears `google-docs-skia-proxy-role-first-experience` from the round-trip `KNOWN_FAILURES` — the fix now round-trips its experience.
- Baselines the new fixture's **pre-existing** edit-leg `bullets` gap in `corpus-edit-roundtrip`, matching its em-dash sibling `single-column-intl-role-headers` (that edit-leg gap is widespread and orthogonal to this parse fix — not introduced here).

## Test plan

- [x] `npm run lint` clean
- [x] Full heuristics suite green (1046 passed) — incl. `corpus`, `corpus-roundtrip`, `corpus-edit-roundtrip`, new repro test
- [x] Fixture personas verified synthetic — `npm run check:fixtures` green + `pdftotext` eyeballed (no real PII)

## Provenance

Code implementation via: Claude Opus 4.8 (high effort)
Verification: `npm run verify` — to be confirmed green in CI


s-annam (approved):
Clean, well-scoped fix — adds en-dash (U+2013) to the `Title–Company` header split that already recognized em-dash. Mechanical extension of a proven path, guarded against the role-first `Company – City, ST` shape.

## Verdict
`APPROVE` — 0 Blocking findings.

## Gates
| Gate | Result |
|---|---|
| Fixture PII | ✅ Verified the binary: `Dana Whitfield` / `dana.whitfield@example.com` / `(312) 555-0142` (real area code + `555` exchange + `0142` subscriber). `check:fixtures` green (52 PDFs); metadata clean (pdf-lib producer, no author). |
| Design-system / reuse | — N/A (no feature components) |
| Style tokens | — N/A (no styling) |
| fallow dead-code | ✅ `splitEnDashTitleCompany` is consumed within the module |
| Command bugs (generator) | ✅ `gen-single-column-endash-title-company.mjs` clean, synthetic persona documented |
| `/code-review` generic pass | ✅ nothing survived verification |

## Verified beyond the gates
- The new split is guarded by `isBareLocationString`, so the role-first `Company – City, ST` header (#215/#436) — which uses the same spaced en-dash to join a company to its trailing city — stays owned by `recoverLocation` and is not mis-cleaved. Requires exactly two segments, so multi-dash lines fall through untouched.
- Date ranges also use an en-dash, but `parseEntryBlocks` runs `stripDateRange` (DATE_RANGE_RE) on each header line **before** `splitHeaderSegments`, so a surviving spaced en-dash is a field separator, never a date range — the same guarantee the shipped em-dash path already relies on.
- The `google-docs-skia-proxy-role-first-experience.expected.json` edit is a ratchet-driven improvement: `experienceChangedAcrossRoundtrip` flips true→false and the fixture drops out of `KNOWN_FAILURES`. Confirmed a genuine round-trip stabilization, not a masked regression.
- The `.repro.test.ts` locks the exact title/company mapping the lossy `*.expected.json` golden (counts only) can't catch. Full suite passes (57 tests).

No Blocking, Secondary, or Nit findings.

---
Reviewed by: Claude Opus 4.8 (high) — reviewed at e7168320

PR #521, opened 2026-07-21
fix(contact): strip a trailing slash in normalizeUrl so linkedin_url round-trips

### LINKED ISSUE #520: Round-trip: linkedin_url loses a trailing slash on re-parse of the reconstructed PDF
## Summary

A parsed `linkedin_url` (and any other contact URL) that ends in a **trailing slash** does not survive the parse → export ("Download PDF") → re-parse round-trip: the re-parsed value has the trailing slash dropped, so `parse1.linkedin_url !== parse3.linkedin_url`. This is the single most common round-trip defect across real résumés (LinkedIn's own canonical profile URL, `linkedin.com/in/<slug>/`, ends in a slash), and no corpus fixture reproduced it before this change.

## Root cause

There are two `normalizeUrl` canonicalizers that must agree for the round-trip to hold, plus the export-side display formatter:

- **Tier 1** — `src/lib/contact/url-utils.ts` `normalizeUrl` (used by the contact extractor `heuristics/extract/contact.ts`). Pre-fix it stripped a trailing sentence-punctuation mark (`[,;.)]`) and a leading `www.`, **but not a trailing `/`**.
- **Tier 1.5** — `src/lib/heuristics/regex-fallback.ts` `normalizeUrl` — already stripped a trailing slash (`.replace(/\/+$/, "")`). Its docblock even claims the two are "in lockstep", but they diverged on the slash.
- **Export** — `src/lib/pdf/ats-resume-model.ts` `formatLinkDisplay` drops scheme, leading `www.`, **and any trailing slash** for display; `urlSlug` (`url-utils.ts`) likewise drops a trailing slash for identity comparison.

So a source URL `https://linkedin.com/in/jane/` was parsed by Tier 1 as `https://linkedin.com/in/jane/` (slash kept), exported as the display slug `linkedin.com/in/jane` (slash dropped), and re-parsed as `https://linkedin.com/in/jane` — a changed value. The `www.` strip was added for exactly this symmetry reason (#425); the trailing slash is the same class of asymmetry, left unhandled in the Tier-1 helper only.

## Fix

Add the trailing-slash strip to the Tier-1 `url-utils.ts` `normalizeUrl`, bringing it into lockstep with the Tier-1.5 fallback, `formatLinkDisplay`, and `urlSlug`:

```ts
const trimmed = raw
  .trim()
  .replace(/[,;.)]$/, "")
  .replace(/\/+$/, "")                       // ← added: drop trailing slash(es)
  .replace(/^(https?:\/\/)?www\./i, "$1");
```

A trailing slash on a host+path is semantically inert (same resource), so dropping it loses nothing — the same rationale already applied to `www.`.

## Synthetic reproducer

`tests/fixtures/pdfs/unknown/bare-linkedin-url-roundtrip.pdf` — a single-column synthetic-persona résumé (`Marcus Halloway` / `@example.com` / `(415) 555-0173`) whose profile line carries a bare `linkedin.com/in/<slug>/` with a trailing slash. Generator: `scripts/fixtures/gen-bare-linkedin-url-roundtrip.mjs`. Pre-fix, `corpus-roundtrip.test.ts` reports its `contact` category regressing (`linkedin_url` trailing slash dropped); post-fix it round-trips clean.

## Acceptance criteria

- [ ] `normalizeUrl("https://linkedin.com/in/jane/")` returns `https://linkedin.com/in/jane` (unit test in `src/lib/contact/url-utils.test.ts`).
- [ ] A trailing-slash source URL and its slash-less exported display converge under `normalizeUrl` (symmetry unit test, mirroring the existing `www.` symmetry test).
- [ ] The new fixture `bare-linkedin-url-roundtrip.pdf` round-trips clean in `corpus-roundtrip.test.ts` (NOT listed in that file's `KNOWN_FAILURES`).
- [ ] No existing `*.expected.json` snapshot regresses (`npm run test`) — the change only affects URLs that carried a trailing slash, of which the corpus had none.
- [ ] `npm run check:fixtures` green; the new fixture's persona is synthetic (verified via `pdftotext`).
- [ ] The `www.` round-trip invariant (`render-roundtrip-www-link.repro.test.ts`) still holds.

## Notes / out of scope

- The Tier-1.5 annotation branch in `regex-fallback.ts` assigns a LinkedIn **link-annotation** href to `linkedin_url` **raw** (unnormalized) — a related but distinct round-trip risk (an annotation-sourced URL keeps its `www.`/slash). Not addressed here; worth a separate follow-up with its own annotation-bearing fixture.

rohithgollapalli:
## Summary

A parsed contact URL ending in a **trailing slash** — LinkedIn's own canonical `linkedin.com/in/<slug>/` form — did not survive the parse → export ("Download PDF") → re-parse cycle. The Tier-1 `contact/url-utils.ts` `normalizeUrl` kept the slash, but the exporter's `formatLinkDisplay` (and `urlSlug`, and the Tier-1.5 `regex-fallback.ts` `normalizeUrl`) all drop it — so the re-parsed value came back slash-less and `linkedin_url` changed across the round-trip.

This was the **most common** round-trip defect across a sweep of real résumés, and no corpus fixture reproduced it. Reproduced here with a **synthetic** trailing-slash fixture; the real PDFs that surfaced it never entered the repo.

## Fix

Add the trailing-slash strip to the Tier-1 `normalizeUrl`, bringing it into lockstep with the Tier-1.5 fallback, `formatLinkDisplay`, and `urlSlug`:

```ts
.trim()
.replace(/[,;.)]$/, "")
.replace(/\/+$/, "")                       // ← added
.replace(/^(https?:\/\/)?www\./i, "$1");
```

Semantically inert (a trailing slash on a host+path is the same resource) — the same rationale the `www.` strip (#425) already applied.

## What else moved

- New synthetic fixture `bare-linkedin-url-roundtrip.pdf` (trailing-slash slug) + snapshot; round-trips **clean** post-fix (no `KNOWN_FAILURES` entry).
- `normalizeUrl` trailing-slash + symmetry unit tests.
- Baselined the new fixture's **pre-existing** edit-leg `bullets` gap (the widespread one, orthogonal to this fix).

## Test plan

- [x] `npm run lint` clean
- [x] `corpus`, `corpus-roundtrip`, `corpus-edit-roundtrip`, `url-utils`, `render-roundtrip-www-link` tests green
- [x] No existing snapshot regressed (corpus had no trailing-slash URLs)
- [x] Fixture persona verified synthetic — `npm run check:fixtures` green + `pdftotext` eyeballed (no real PII)

Closes #520

## Provenance

Code implementation via: Claude Opus 4.8 (high effort)
Verification: `npm run verify` — to be confirmed green in CI

## Note

A related but distinct risk — the Tier-1.5 annotation branch assigns a LinkedIn link-**annotation** href to `linkedin_url` raw (unnormalized) — is called out as out-of-scope in #520 for a separate follow-up with an annotation-bearing fixture.


src/lib/contact/url-utils.ts
```diff
@@ -20,22 +20,27 @@ export const LINKEDIN_NONPROFILE_RE =
...
+ *  lockstep with the Tier-1.5 `regex-fallback.ts` `normalizeUrl` and the
+ *  trailing-punctuation strip `urlSlug` already applies for identity
+ *  comparison. */
 export function normalizeUrl(raw: string | undefined): string | undefined {
   if (!raw) return undefined;
   const trimmed = raw
-    .replace(/[,;.)]$/, "")
     .trim()
+    .replace(/[,;.)]$/, "")
+    .replace(/\/+$/, "")
```
s-annam: Non-blocking nit: this slash strip runs *after* the `[,;.)]$` punctuation strip, so a contrived `.../jane./` input keeps its `.` (normalizes to `.../jane.`). No real résumé URL hits this — a trailing `/` follows a slug, not a period — so no change needed; noting for completeness. The strip itself is correct and matches `formatLinkDisplay`/`urlSlug`/Tier-1.5.

s-annam (dismissed):
Clean, well-scoped round-trip fix. The trailing-slash strip in the Tier-1 `normalizeUrl` is the right place — I verified it brings the helper into lockstep with the three consumers the PR names: `formatLinkDisplay` (contact.ts:254), `urlSlug` (url-utils.ts:63, via `[/.,;:)\]]+$`), and the Tier-1.5 `regex-fallback.ts` `normalizeUrl` (line 142) all already drop the trailing slash, so Tier-1 was the outlier. The strip is semantically inert for contact URLs, mirroring the #425 `www.` rationale.

Also a quiet correctness win in the reorder: `.trim()` now runs *before* the trailing-punctuation strip, so a URL with trailing whitespace gets its `[,;.)]$` mark stripped where the old order would have missed it.

**Gates run:** fixture-PII (3a) PASS — `npm run check:fixtures` green (52 PDFs), `pdftotext` + `pdfinfo` eyeballed: persona synthetic (Marcus Halloway / @example.com / (415) 555-0173 = real area code + 555 exchange + 0173 subscriber), no metadata author, and the fixture actually carries the `/in/marcus-halloway/` trailing slash it targets. Design-system (3b/3c) N/A — no feature/UI code. Dead-code (3d) N/A — no new exports. Command-level (3e) — the generator script mirrors its siblings, no gh/bash surface. Targeted tests green: `url-utils` + `corpus-edit-roundtrip` (63 passed). The new fixture round-trips clean for the linkedin_url defect (`derived.linkedinUrlChangedAcrossRoundtrip: false`); its `KNOWN_FAILURES: ["bullets"]` entry is the pre-existing widespread edit-leg gap shared by many corpus fixtures, not a masked regression (new fixture, no prior baseline to ratchet).

Verdict rule applied: 0 Blocking → APPROVE. One non-blocking nit below.

## Nits

- `url-utils.ts:43` — the punctuation strip runs before the slash strip, so a (contrived) input like `linkedin.com/in/jane./` normalizes to `linkedin.com/in/jane.` — the `.` survives because it is no longer the last char once evaluated. Doesn't arise for any real résumé URL (trailing `/` follows a slug, not a period), so purely cosmetic — flagging only for completeness, no change needed.

---

s-annam (approved):
Approved at `2cc05dc`. Clean, well-scoped round-trip fix, and the fixture genuinely earns its place. I pushed the three review findings into the branch's single commit rather than leaving them as nits for a follow-up round — the change is small, sits entirely inside `normalizeUrl` plus one docblock, and this PR auto-merges on approval. Details below so nothing lands unreviewed. **Verdict rule: 0 Blocking → APPROVE.**

## What I verified beyond the generic pass

- **The fixture is a real regression guard, not just a passing snapshot.** Reverted the one-line strip and re-ran `corpus.test.ts` against `bare-linkedin-url-roundtrip.pdf`: `derived.linkedinUrlChangedAcrossRoundtrip` flips `false → true`. It reproduces #520 and would catch the regression.
- **No other corpus fixture was affected.** With the strip reverted, that fixture is the *only* corpus failure — confirming the PR's "the corpus had no trailing-slash URLs" claim rather than taking it on trust.
- **The `KNOWN_FAILURES: ["bullets"]` entry is the widespread edit-leg gap, not a masked new bug.** Removed it and read the assertion it produces: `bullet "Vantreon: cut p99 checkout latency 43%…" missing on re-parse` — the same synthetic-edit shape the other 41 baselined fixtures carry. New fixture, no prior baseline, so nothing is being ratcheted away.
- **Symmetry holds at the consumer.** `formatLinkDisplay` (`contact.ts:250-254`) ends in the same `.replace(/\/+$/, "")`, so parse and export now agree.
- **The `.trim()` reorder is a quiet fix in its own right** — trailing whitespace no longer hides the punctuation mark from the strip.
- All six of #520's acceptance criteria met, including "NOT listed in `corpus-roundtrip.test.ts`'s `KNOWN_FAILURES`".

## What I changed in `2cc05dc`

Amended into the existing commit (one commit per PR for the merge queue); authorship preserved.

1. **`url-utils.ts` — folded the two strips into one character class.** Run sequentially, a mark sitting *behind* the slash survives: `linkedin.com/in/jane./` is not punctuation-final when the punctuation pass runs, and the slash pass doesn't go back for it, so it normalized to `https://linkedin.com/in/jane.`. `/[,;.)/]+$/` handles both in one pass. (Contrived input — but two passes was the wrong shape, and one class is simpler as well as more correct.)
2. **`url-utils.ts` — guarded the strip-to-nothing case.** This one was a small behaviour regression the diff introduced: `normalizeUrl("https://")` returned `"https://"` before and `"https://https:"` after — the strip eats the `//`, the `^[a-z][a-z0-9+.-]*:\/\//` test then fails, and the prefix is re-added. `normalizeUrl("/")` → `"https://"` joined the pre-existing `normalizeUrl("  ")` hole. Now returns `undefined` when the strip consumes everything or leaves a bare scheme, matching the contract `urlSlug` already had. Unreachable from the extractors today — no URL pattern emits a bare scheme — so this is hygiene, not a live bug.
3. **`gen-bare-linkedin-url-roundtrip.mjs` — corrected the docblock.** The persona block listed `linkedin.com/in/marcus-halloway`, but line 63 draws `linkedin.com/in/marcus-halloway/`. The trailing slash *is* the defect under test, and this generator is the only prose documenting the only reproducer for #520 — regenerating from that description would have silently disarmed the fixture. Also retired the speculative "the export **either** display-normalizes **or** confidence-gates the field", since this PR pins the cause exactly. Docblock only; the PDF is byte-identical (I reverted the regenerated copy — it differed solely in its `CreationDate`).

Plus two unit tests pinning 1 and 2.

**Re-verified after the amend:** `npm run typecheck` clean, `npx eslint src` clean, full `npx vitest run` green (**180 files passed | 6 skipped; 2519 tests passed | 9 skipped**), `npm run check:fixtures` green.

## Gates

- **Fixture PII (3a) — PASS.** `check:fixtures` green (52 PDFs). Persona synthetic: Marcus Halloway / `marcus.halloway@example.com` / `(415) 555-0173` (real area code + `555` exchange + `0173` subscriber). `pdfinfo` shows no metadata author; `pdftotext` confirms the drawn profile line carries the `/in/marcus-halloway/` slash the fix targets.
- **Design-system (3b) / style tokens (3c) — N/A.** No feature or UI code.
- **Dead code (3d).** fallow flags the new generator as an unused file and clones it against seven sibling `gen-*.mjs` scripts — the house pattern for fixture generators, and fallow is report-only in `verify`. No action.
- **Command-level (3e).** The generator is plain node, idempotent (`mkdirSync({recursive:true})`), mirrors its siblings; no `gh`/bash surface.

## Note, not a finding

`regex-fallback.ts:103` still assigns a LinkedIn link-**annotation** href to `linkedin_url` raw — the text branch one line above calls `normalizeUrl`, the annotation branch doesn't, so an annotation-sourced `https://www.linkedin.com/in/jane/` keeps both its `www.` and its slash and won't round-trip. #520 explicitly scopes it out pending an annotation-bearing fixture and the PR body repeats the caveat; flagging here only so it stays visible after #520 closes.

---

Issue #519, opened 2026-07-21
[result] tab labels read as inspector panes — add one-line descriptions to Tab


s-annam:
Parent: #511 — **depends on the tab-affordance issue** (see Notes)

## Summary

The four result tabs are labelled like inspector panes, not like destinations:

```tsx
<Tab id="reconstructed">Reconstructed resume</Tab>
<Tab id="find-jobs">Find jobs</Tab>
{showQualityTab && <Tab id="quality" warn={!analysis.isAvailable}>Resume Quality</Tab>}
<Tab id="diagnostics" count={triggerCount}>Source &amp; diagnostics</Tab>
```

`src/components/features/ResultDetailTabs.tsx:72-83`.

Two problems.

**1. The labels under-sell what is behind them.** `Find jobs` sits behind a bare verb phrase that gives no hint it builds a search query from the parsed résumé, hits job boards, and ranks results by fit against that résumé. `Resume Quality` reads like a score readout when it is an on-device LLM critique. `Source & diagnostics` reads like developer output. A user who lands on `Reconstructed resume` — the default (`ResultDetailTabs.tsx:43`) — has no reason to click any of the other three.

**2. They only exist after a file is dropped.** The tab row is rendered inside `Result`, which mounts only in the `done` phase. A visitor on the landing page never sees these words at all, so the three non-default lanes are invisible at exactly the moment the visitor is deciding whether to hand over a résumé.

This is the discoverability half of #511's approach B. The sibling issue fixes *whether you can tell which tab is selected*; this one fixes *whether you can tell what a tab is for, and that it exists*.

## Expected

**A. One-line subtitles.** Each tab carries a short qualifier under its label, e.g.:

| Tab | Subtitle (for review, not final) |
|---|---|
| Reconstructed resume | what a parser pulled out — edit it here |
| Find jobs | search boards, ranked by fit to this résumé |
| Resume Quality | on-device critique and rewrites |
| Source & diagnostics | raw text, layout flags, what went wrong |

This is a `Tabs.tsx` API addition — an optional `description` prop on `Tab` — not a callsite hack. It must degrade cleanly: a `Tab` with no description renders exactly as today, since `Tabs` is a shared primitive and a future consumer should not be forced into two-line tabs.

[...]

Issue #511, opened 2026-07-21
[epic] Visual identity & product-breadth legibility — palette, tab affordance, positioning


s-annam:
## Problem

Two defects share one root and one surface, so they are worth fixing as one pass.

**1. A first-time visitor cannot learn that OfflineCV does anything beyond auditing a PDF.**

Every word on the pre-drop page is parser-audit framing: the header subtitle reads `"A parser audit for your resume — not a judge"` (`src/App.tsx:91`), the hero says OfflineCV `"shows you what a recruiter or screener reads back from your resume"` (`src/App.tsx:139`), and the quiet block below the drop zone reinforces the mirror positioning. Nothing states that the product also matches a résumé against a job description, ranks job postings by fit, and rewrites sections on-device.

Those lanes exist and ship:

| Lane | Where it lives | Discoverable pre-drop? |
|---|---|---|
| Parser audit + edit + ATS-safe export | `/` — `ReconstructedResume`, `render-ats-pdf.ts` | Yes — it is the whole hero |
| Job search (query build → provider search → rank by résumé fit → deep links) | `FindJobsPanel`, `src/lib/job-search/` | **No** — a tab that only exists after a drop (`ResultDetailTabs.tsx:74`) |
| JD match (requirement/evidence coverage, JD-driven rewrites) | `/jd-fit/`, `src/lib/jd-match/` | **No** — the cross-sell banner is flag-gated **off** (`useFlag("jd-fit-banner")`, `src/App.tsx:62`) |
| On-device résumé quality critique + rewrite | `ResumeQualityPanel`, `src/lib/webllm/` | **No** — a tab, post-drop only |

So three of the four lanes are invisible until the user has already committed a file. The one differentiator that binds them — **all of it runs client-side, nothing leaves the browser** — is stated once, as an emoji chip (`🔒 Your file never leaves your device`), scoped to the *file* rather than to the product.

**2. The page is visually flat, and the tab row does not read as a tab row.**

The specific, checkable causes:

- **The brand accent token holds a stock blue.** `--color-brand-amber: #2563eb` and `--color-brand-amber-light: #3b82f6` (`src/design-system/styles/tokens.css:25-26`). The name and the value disagree — a leftover from the resumelint palette. Every accent in the app (logo tile, active tab, links, primary buttons) therefore resolves to the same default blue over a slate-grey surface stack, which is what reads as "bland": one accent, no second voice, no warm counterpoint.
[...]

Issue #320, opened 2026-07-03
job-search: BYOK provider adapter — user API key via settings + localStorage


s-annam:
Part of the job-search epic (see parent issue). Third slice: let a user plug in *their own* free-tier API key for keyed job APIs (BYOK — bring your own key), unlocking broader coverage than the keyless feeds, while staying fully client-side. Blocked by the provider-interface issue.

## Problem

Keyless CORS-open feeds skew remote/tech. Keyed APIs (e.g. Adzuna) have broader coverage, but a shared key is impossible in a static client-side app (any bundled key is public) and a shared-key proxy is out of scope for this repo. BYOK resolves this: the user registers for their own free key, the key lives only in their browser, and quota is per-user.

## Feasibility gate (do this first)

A keyed provider only qualifies if its API allows **browser-origin calls with the key in a query param or header and permissive CORS**. Verify from a browser origin before writing any adapter:

- **Adzuna** (`api.adzuna.com`, free tier 250 calls/day, key = app_id + app_key in query params) — primary candidate.
- Others (Jooble, USAJobs-with-key, etc.) — evaluate if Adzuna fails.

**If no keyed provider passes the CORS check, close this issue as won't-do with the findings documented in a comment.** Do not fall back to a proxy.

## Scope

1. **Key storage — `src/lib/job-search/byok.ts`**
   - `localStorage` keys following the documented `rl_*` convention (e.g. `rl_byok_adzuna`).
   - Get/set/clear helpers; never log key values; never send keys to any endpoint other than the provider's own API.

2. **Settings surface**
   - Reuse-first: check for an existing settings/preferences surface before building one. If none exists, a minimal disclosure-style section in the Find Jobs panel ("Add your own Adzuna key for broader results") with a masked input, Save, and Clear — using `@design-system` primitives.
   - Inline link to the provider's key-registration page.
   - Plain-language note: "Your key is stored only in this browser and sent only to <provider>."

3. **Provider adapter — `src/lib/job-search/providers/adzuna.ts`** (or whichever passes the gate)
   - Implements the same `JobProvider` interface from the provider-interface issue; enabled only when a key is present.
   - Maps provider errors (401 invalid key, 429 rate-limited) to user-readable states; 401 prompts re-entry of the key.

## Reuse analysis

- **Capability:** store a user secret locally; add a keyed provider to the job search fan-out.
[...]

PR #526, opened 2026-07-21
feat(chrome): make product breadth legible and give the palette honest names (#511)

### LINKED ISSUE #513: [design-system] "brand-amber" token holds a blue — 25 consumers on a misnamed token while the correctly-paired accent sits dead
Parent: #511

## Summary

`--color-brand-amber` holds `#2563eb` — a blue. `--color-brand-amber-light` holds `#3b82f6` — a lighter blue (`src/design-system/styles/tokens.css:25-26`, and again in the dark block at `:67-68` as `#60a5fa` / `#93c5fd`). The name and the value have disagreed since the resumelint palette was swapped out. Every accent in the app resolves through this token — the logo tile, the active-tab underline, the primary `Button` fill, every focus ring — so the single most load-bearing colour name in the design system lies about what it is.

**This is not a request to brand the default palette.** The docblock at `tokens.css:5-19` is explicit that the generic slate-and-blue default is deliberate: it exists so an outside cloner ships a clean, unbranded look and supplies their own brand at this one seam, either by redefining these properties downstream or by repointing the `@design-tokens` Vite alias. That decision stands and this issue must not weaken it. The defect is the **name**, plus two consequences of it.

## Where it lives

**1. The name is wrong.**

`tokens.css:25-26` (light), `:67-68` (dark). Referenced by name from `theme.css:29-30`, which generates `bg-brand-amber` / `text-brand-amber` / `ring-brand-amber`. Consumers include at minimum `Button.tsx:41,45` (primary fill + focus ring), `PageShell.tsx:63` (logo tile), `Tabs.tsx:140` (active tab border), `App.tsx:130` (footnote link), and `EditableField`.

**2. The hover pair runs the wrong direction.**

`Button`'s primary variant is `bg-brand-amber ... hover:bg-brand-amber-light` (`Button.tsx:41`). In light mode that is `#2563eb → #3b82f6` — hover makes the CTA **lighter**, i.e. lower contrast against its own white-inverse label, and lower contrast against the page. Right next to it sits `--color-accent-primary: #2563eb` with `--color-accent-primary-hover: #1d4ed8`, which darkens — the conventional direction. So the token file already contains a correctly-paired accent, and the primary CTA is wired to the other one.

**3. Two tokens hold the identical value; only the misnamed one is used.**

`--color-brand-amber: #2563eb` (`tokens.css:25`) and `--color-accent-primary: #2563eb` (`:41`) are the same colour under two names. Measured consumption outside `src/design-system/styles/`:

| Token | Feature/DS files consuming it |
|---|---|
| `brand-amber` | **25** — `Button`, `EditableField`, `TextAreaField`, `StarRating`, `Tabs`, `GitHubStarCta`, `PageShell`, `App`, and 17 feature components |
| `accent-primary` | **0** |
| `accent-forward` (foreground) | **0** |
| `accent-forward-bg` | **3** — `FeedbackPanel`, `ReportGapSection`, `GitHubStarCta` |
| `brand-navy`, `brand-navy-dark` | **0** |
| `brand-cream` | **0** |

So the app runs entirely on the token whose name is wrong, while the correctly-named-and-paired one is dead, and three more tokens are dead outright.

**4. The second accent already exists and is used only as a background wash.**

`--color-accent-forward: #4f46e5` / `--color-accent-forward-bg: #EEF2FF` (`tokens.css:43-44`, dark at `:85-86`). This is exactly the "second voice" #511 called for — a distinct hue for forward motion, already defined, already dark-mode-paired. But the foreground value has **zero** consumers; only the `-bg` wash is used, in three places. That is why the page reads as monochrome-blue despite the palette carrying two accents: the second one is never drawn as ink.

## Repro

1. `rg 'brand-amber' src/` — count the consumers of a token whose name asserts amber.
2. Open the app and hover the primary **Download** CTA. The fill lightens.
3. Compare `--color-brand-amber` and `--color-accent-primary` in `tokens.css`. Identical values, both referenced from `theme.css`.

## Expected

**Rename, don't repaint.** The token should be named for its role, not for a colour it does not hold. `--color-accent-primary` already occupies the obvious name and the obvious value **and has no consumers to migrate**, which makes the cleanest resolution a **merge**: retire `brand-amber` / `brand-amber-light` in favour of `accent-primary` / `accent-primary-hover`, and repoint the 25 consuming files.

That touches the semantic contract in `theme.css`, which its own docblock calls the stable public contract — so it needs a deprecation note in the README theming section, not a silent swap. The migration itself is mechanical (`bg-brand-amber` → `bg-accent-primary`, etc.) and ESLint already blocks the raw-palette alternative, so the blast radius is wide but shallow. If the merge is judged too broad for one issue, the fallback is a straight rename to a role name (`--color-brand-accent`) with the duplicate filed separately; the fallback must still fix the hover direction.

Delete `brand-navy`, `brand-navy-dark`, and `brand-cream` in the same pass — zero consumers, both themes. Keeping dead names in the published override contract makes the seam look larger than it is.

| | Now | After |
|---|---|---|
| Primary CTA fill | `bg-brand-amber` `#2563eb` | `bg-accent-primary` `#2563eb` — value unchanged |
| Primary CTA hover | `hover:bg-brand-amber-light` `#3b82f6` (lighter) | `hover:bg-accent-primary-hover` `#1d4ed8` (darker) |
| Focus ring | `ring-brand-amber` | `ring-accent-primary` |
| Forward-lane accent | mostly unused | `accent-forward` consumed by the forward lane (see #511 lane split) |

Values do not move in this issue **except** the hover direction. Recolouring the default palette is out of scope and would contradict `tokens.css:5-19`; if OfflineCV wants a branded look on its own deploy, that belongs in an override layer at the documented seam, filed separately.

## Acceptance criteria

- No token name in `tokens.css` asserts a colour family it does not hold.
- Exactly one accent token pair is consumed for the primary CTA / focus ring; the duplicate is gone or documented as an intentional alias with a comment saying which is canonical.
- The primary button's hover state moves **toward** more contrast, not less, in both light and dark.
- `accent-forward` (foreground) is either consumed for the forward lane or removed — no defined-but-dead accent left in the file.
- `brand-navy`, `brand-navy-dark`, and `brand-cream` are gone from both theme blocks, or a comment states why a zero-consumer token is retained.
- Every semantic name removed from `theme.css` is noted in the README theming section, since that file is the documented downstream override contract.
- Rendered output is unchanged apart from the hover direction: no visual diff on the default palette beyond that, verified by eye in both light and dark.
- Both themes checked independently for 4.5:1 body contrast — the dark block at `tokens.css:65+` has its own values and must not be inferred from light.
- `npm run verify` green. No feature-code change beyond the token-name migration; no hex or raw palette class introduced.

## Notes

Flagged by the #511 UI review as the mechanical root of "the page looks bland": one accent, consumed everywhere, while the second accent sits defined and idle. The blandness fix is mostly *using* what the palette already has, not adding to it.

The `brand-navy` / `brand-cream` names have the same smell — `--color-brand-cream: #f8fafc` is a cool near-white, not cream, and in dark mode it is `#0f172a`, which is neither. They are confirmed dead (zero consumers), so deleting is cheaper than renaming.

A lint rule forbidding a colour-family word in a token name would prevent the recurrence, but that is tooling work; file it separately if the class of bug repeats.

### LINKED ISSUE #514: [chrome] emoji used as icons in the landing trust-chip row — replace with one SVG icon set
Parent: #511

## Summary

The first row of content on the landing page is four trust chips whose icons are emoji:

```tsx
<Chip icon="⚡">A few seconds</Chip>
<Chip icon="🔒">Your file never leaves your device</Chip>
<Chip icon="✓">No account, no email</Chip>
<Chip icon="🔁">Same PDF, same score</Chip>
```

`src/App.tsx:93-98`.

`CLAUDE.md` forbids this — *"Use SVG icons (Heroicons, Lucide), not emojis"* — and the repo already honours the rule elsewhere: `Tabs.tsx:152-155` deliberately writes `⚠︎` as `U+26A0 U+FE0E` with a comment explaining that the variation selector forces **text** presentation *"so the glyph renders monochrome … not as a colour emoji — the codebase's no-emoji-as-icon rule."* Someone reasoned carefully about exactly this on a warning glyph, then the chip row shipped four colour emoji anyway.

## Why it matters here specifically

These four chips carry the product's trust claims — speed, privacy, no-signup, determinism — and they are the first thing a visitor sees, above the hero. Emoji render from the OS font, so the row looks materially different on macOS, Windows, and Android, and cannot be tinted by any token. A privacy claim rendered in a different colour, weight, and style on every machine is the weakest possible presentation for the strongest thing the product has to say.

`✓` is the odd one out: it is not an emoji at all (`U+2713`, text presentation by default) while its three neighbours are. So the row also mixes glyph families — one monochrome mark alongside three colour pictographs — which is why it reads as slightly ragged even before you notice the emoji.

## Where it lives

- `src/App.tsx:93-98` — the four chips.
- `src/design-system/primitives/Chip.tsx:6,20-22` — `icon?: React.ReactNode`, rendered inside a bare `<span>` with `gap-1`. The primitive takes any node, so it needs no API change; but it also does nothing to size or align an icon, which is why an SVG dropped in unchanged will not sit right against `text-xs` copy.
- **No icon library is currently a dependency.** `package.json` has no `lucide-react`, no Heroicons, no `react-icons`. Adding one is a decision this issue has to make, not an assumption it can lean on.

## Scope: three glyphs, not every mark in the app

A sweep for emoji codepoints across `src/` returns matches in ~12 files, but almost all are **text-presentation marks used deliberately** and are **not in scope**: `✓`/`✗` toggles in `RewriteReviewList.tsx:47,66` (docblocks explain they are decorative, button owns the label), `★`/`☆` in `StarRating.tsx:74`, `⚠` in `ContactDetails.tsx:168,206` and `SectionRewrite.tsx:491` (each `aria-hidden`), `✓` markers in `ModelSelector.tsx:332,462` and `JdMatch.tsx:120`.

The genuine colour-emoji violations are **three**, all in the chip row: `⚡` (`U+26A1`), `🔒` (`U+1F512`), `🔁` (`U+1F501`). `✓` in the same row is a text mark like its cousins elsewhere — it is in scope only because the row must end up visually consistent, not because it breaks the rule.

Whether the app-wide text-mark convention should also become SVG is a larger, separate question. Do not expand this issue into it.

## Repro

1. Load `/` on macOS, then on Windows or Android.
2. Compare the chip row. Different glyph shapes, different colours, different optical sizes, different baseline alignment.
3. Note that `✓` stays monochrome on every platform while the other three do not.

## Expected

Replace all four with icons from **one** set, at one stroke weight, sized from a token — and give `Chip` the small amount of structure needed to hold an icon properly.

Choices this issue must settle:

- **Which set.** Lucide is the natural pick (`CLAUDE.md` names it, it is tree-shakeable, and the icons are 24×24 / 2px-stroke by default). Adding a dependency to a project that deliberately keeps its entry chunk small is a real cost, so the alternative — four hand-inlined SVG paths in a tiny local module, no dependency — is a legitimate answer. Decide explicitly and record why.
- **Icon sizing.** The chips are `text-xs`; an icon at `1em` will look weak and at 24px will dominate. Pick one size, express it as a class on the `Chip` icon slot so every future chip inherits it, and do not let callsites pass sizing.
- **Semantics.** Icons here are decorative — the chip text already says everything. They must be `aria-hidden="true"` so a screen reader reads *"A few seconds"*, not *"lightning bolt, A few seconds"*.

Note that #511's capability-strip issue may replace this row wholesale. Fixing the icons is still worth doing first and independently: the strip will need the same icon decision, and this issue is where it gets made.

## Acceptance criteria

- The three colour-emoji glyphs (`⚡ 🔒 🔁`) are gone from `App.tsx`, verified by grepping for their codepoints rather than by eye.
- The chip row renders one consistent glyph family — no monochrome mark sitting beside pictographs.
- All chip icons come from one set at one stroke weight and one token-driven size.
- Icons are `aria-hidden="true"`; the accessible name of each chip is its text alone.
- `Chip`'s icon slot owns sizing/alignment — callsites pass an icon, never a size.
- Chips remain legible at 375px without wrapping mid-chip.
- If a dependency is added, the bundle delta is stated in the PR and tree-shaking is confirmed (only the imported icons ship).
- Contrast holds in both themes: the icon inherits chip tone colour and meets 3:1 against the chip background.
- `npm run verify` green.

## Notes

The `⚠︎` in `Tabs.tsx:152-155` is **not** in scope to change — it is already text-presentation, already tinted by `text-feedback-warning-text`, and already paired with an `sr-only` label. If the chosen icon set covers it cleanly, swapping it is a reasonable bonus, but the existing treatment is correct as-is and the comment explaining it should survive any such change.

Deeper follow-up, deliberately not in this issue: `Chip` renders a `<span>` with no role, so the row is four unlabelled decorative spans. Whether the trust claims should be a `<ul>` for screen-reader navigability is a separate accessibility question — file it if it comes up in review.

### LINKED ISSUE #515: [chrome] logo mark still renders "R" — resumelint holdout the #498 rename could not grep for
Parent: #511

## Summary

The logo tile in the shared page header renders the literal character **`R`** — the resumelint initial — next to the `offlinecv` wordmark:

```tsx
<a
  href={import.meta.env.BASE_URL}
  className="inline-grid h-8 w-8 place-items-center rounded-md bg-brand-amber text-base font-bold text-content-inverse"
  aria-label="offlinecv home"
>
  R
</a>
```

`src/components/features/PageShell.tsx:59-66`.

`PageShell` is the chrome **both** root surfaces share (`/` and `/jd-fit/`), so this appears on every page of the product, top-left, as the first element a visitor's eye lands on. The `aria-label` is already correct (`"offlinecv home"`), which means a screen-reader user gets the right brand and a sighted user gets the wrong one.

## Why the rename missed it

#498 renamed the project resumelint → OfflineCV and was closed as complete. The audit for that rename grepped for the literal string `resumelint`. A lone `R` in a JSX text node matches nothing — the identifier is *derived* from the old name rather than containing it, so no search for the old name could see it. Same class of miss as the `rl_*` storage-key prefixes that #498 caught late.

## Repro

1. Load `/` (or `/jd-fit/`).
2. Look at the top-left tile: a blue rounded square containing `R`, immediately followed by the wordmark `offlinecv`.
3. `rg -n '>R<' src/components/features/PageShell.tsx` — one hit, no other context.

## Expected

The mark should read as OfflineCV. Three viable resolutions, in ascending cost:

| Option | Change | Cost |
|---|---|---|
| **Letter swap** | `R` → `O` (or `OC`) | One character. Immediate, correct, unremarkable. |
| **Drop the tile** | Wordmark alone | Removes a weak mark rather than fixing it; also removes the only accent colour above the fold. |
| **Real mark** | Commission/draw an SVG logo | Out of scope here — that is brand work, not a rename holdout. |

Recommend the letter swap now and file the real mark separately if it is wanted. A single glyph in a coloured tile is a placeholder either way; the defect is that it is the *wrong* placeholder, and that is a one-character fix that should not wait on design work.

If `OC` is chosen over `O`, check it still fits: the tile is `h-8 w-8` with `text-base`, which is tight for two characters — either drop to `text-sm` or accept the single letter.

## Acceptance criteria

- No visible glyph in the app chrome derives from the name "resumelint".
- The visible mark and the existing `aria-label="offlinecv home"` refer to the same brand.
- The tile keeps its dimensions and its contrast: `text-content-inverse` on the accent fill, ≥4.5:1 in **both** themes (the accent token has different values per theme — `tokens.css:25` vs `:67`).
- Verified on `/` **and** `/jd-fit/`, since `PageShell` serves both.
- `npm run verify` green.

## Notes

Worth doing as part of the same sweep: a grep of the repo for other *name-derived* leftovers that a literal-string search cannot see — single-letter marks, initialisms, favicon assets, `apple-touch-icon`, OG images, and any `R`-derived identifier in `index.html` or `public/`. The lesson from #498 is that a rename audit which greps only for the old literal will report done while name-derived identifiers survive; this issue is the proof of that, so closing it without the sweep invites a third instance.

Depends on nothing. Can land independently and immediately — it is the cheapest visible improvement in #511.

### LINKED ISSUE #516: [design-system] Tab selection is nearly invisible — active underline merges with the TabList border
Parent: #511

## Summary

The result-detail tab row does not read as a tab row. Selection is signalled by two things only, and both are undermined by their surroundings:

1. **A font-weight step.** Inactive is `font-medium`, active is `font-semibold` (`src/design-system/shared/Tabs.tsx:139-142`). One weight step at `text-base` is a weak signal, and the panel content directly beneath is full of `font-semibold` text of its own.
2. **A 2px bottom border.** Active gets `border-brand-amber`; inactive gets `border-transparent`. But `TabList` itself renders `border-b border-border-light` (`Tabs.tsx:112`), and the tabs sit on that line via `-mb-px`. So the selection underline is a short coloured segment **on an existing rule** — it reads as part of the card's chrome rather than as the answer to "which panel am I looking at".

There is no active surface, no size step, no container, and no left-edge relationship to the panel below. The row reads as four links floating above a divider.

Contrast is **not** the problem: inactive tabs are `text-content-secondary` (`#334155` light / `#cbd5e1` dark), comfortably past 4.5:1 on card. The tabs are hard to read as *tabs*, not hard to read as *text* — the fix is affordance, not colour.

## Where it lives

`src/design-system/shared/Tabs.tsx` — the whole of it. The `Tab` component (`:127-166`) builds on the `Button` primitive with `variant="ghost"` and then overrides most of what ghost provides:

```tsx
className={`-mb-px rounded-none border-b-2 px-3 py-2 text-base hover:bg-transparent ${activeCls}`}
```

`Tabs.tsx:145`. It cancels ghost's `rounded`, cancels ghost's `hover:bg-surface-subtle`, and overrides ghost's `text-xs` size to `text-base`. When a consumer has to neutralise three of a variant's four properties, the variant is the wrong base — that is worth naming, because it is also why the row has no hover feedback at all: `hover:bg-transparent` removes it, leaving only a `text-content-primary` colour shift.

Consumer: `src/components/features/ResultDetailTabs.tsx:72-83` — four tabs, one conditional, one carrying a `CountBadge`, one carrying a `warn` dot.

## Repro

1. Drop any résumé and let it parse.
2. Scroll to the tab row (`Reconstructed resume` · `Find jobs` · `Resume Quality` · `Source & diagnostics`).
3. Glance away and back. Identifying the selected tab requires reading weight differences and finding a 2px colour segment inside a full-width grey rule.
4. Hover a non-selected tab. The only feedback is a text-colour change — no surface, no cursor-visible affordance beyond the pointer.
5. Screenshot in greyscale: selection is now carried entirely by the font-weight step.

## Expected

Give selection real weight, and stop the underline from competing with the container rule. Direction, not prescription:

- **Selected state gets a surface**, not just a rule — an active tab that sits on `bg-surface-card` while the rest of the row sits on a recessed track, or an equivalent treatment that makes the selected tab visibly *connect* to the panel beneath it. The tab and its panel should read as one object.
- **The active indicator must not sit on the `TabList` border.** Either the indicator moves (a rail above the label, a filled tab) or the container rule goes. Two horizontal lines at the same y are what makes the current treatment illegible.
- **Restore hover feedback.** A tab is interactive; it currently has almost none. Whatever surface the active state uses, hover should be a lighter version of it, in the 150–300ms band, gated on `prefers-reduced-motion`.
- **Selection is not carried by colour alone** — the weight step or an equivalent non-colour cue stays, so the greyscale screenshot above still resolves.
- **Reconsider the `Button` base.** If `Tab` cancels most of `variant="ghost"`, either add a `tab` variant to `Button` (per the CLAUDE.md rule: add the variant to the shared piece) or have `Tab` compose the primitive differently. Do **not** introduce a raw `<button>` — that is ESLint-blocked and correctly so.

All existing behaviour is load-bearing and must survive: roving tabindex, Arrow/Home/End handling (`Tabs.tsx:70-107`), `aria-selected` / `aria-controls` wiring, inactive panels staying mounted via `hidden` (documented at `Tabs.tsx:22` — child UI state survives a switch, so do not switch to conditional rendering), the `CountBadge` slot, and the `warn` dot with its `sr-only` label.

`Tabs` is a design-system primitive with exactly one consumer today, so the change is contained — but it is a shared component, so the variant belongs in `Tabs.tsx`, never forked into `ResultDetailTabs.tsx`.

## Acceptance criteria

- The selected tab is identifiable at a glance, without reading, and without relying on the 2px underline.
- Selection survives a greyscale render — no colour-only signal.
- The active indicator does not sit on, or visually merge with, the `TabList`'s own border.
- Non-selected tabs have a visible hover state; transitions are 150–300ms and respect `prefers-reduced-motion`.
- Keyboard behaviour is unchanged: Tab/Shift-Tab reaches the row once, Arrow Left/Right + Home/End move selection **and** focus, focus ring remains visible on the selected tab.
- ARIA wiring unchanged: `role="tablist"/"tab"/"tabpanel"`, `aria-selected`, `aria-controls`, `aria-labelledby`.
- Inactive panels stay mounted (`hidden`), so state in `ReconstructedResume` survives a tab switch — regression-check by editing a field, switching tabs, and switching back.
- `CountBadge` and the `warn` dot still render and still read correctly in the new treatment.
- Verified at 375px: the row scrolls horizontally (`overflow-x-auto`) without clipping the active indicator or the badge.
- Both themes checked independently for contrast.
- No raw `<button>`, no hex, no raw palette class. `npm run verify` green.

## Notes

This is the affordance half of #511's approach B. The discoverability half — one-line subtitles per tab and a pre-drop preview rail so a visitor sees the lanes before committing a file — is filed separately and **depends on this issue**: subtitles inside the current flat treatment would make the row longer without making it more legible.

Related: `Tabs.tsx` overriding `Button`'s `text-xs` to `text-base` is the only place in the codebase where a tab-sized control jumps two size steps. If a `tab` variant lands, it should own that size rather than leaving it as a callsite override.

### LINKED ISSUE #517: [landing] pre-drop page never says the product finds jobs — replace the trust chips with a capability strip
Parent: #511

## Summary

A visitor who has not yet dropped a file cannot learn that OfflineCV does anything except audit a PDF. Three of its four lanes are invisible until after a file is committed.

| Lane | Implementation | Reachable pre-drop? |
|---|---|---|
| Parser audit → edit → ATS-safe export | `ReconstructedResume`, `src/lib/pdf/render-ats-pdf.ts` | Yes — the hero is about nothing else |
| Job search (query build → provider search → rank by résumé fit → deep links) | `FindJobsPanel`, `src/lib/job-search/` | **No** — tab, post-drop only (`ResultDetailTabs.tsx:74`) |
| JD match (requirement/evidence coverage, JD-driven rewrites) | `/jd-fit/`, `src/lib/jd-match/` | **No** — the cross-sell is flag-gated off (`useFlag("jd-fit-banner")`, `App.tsx:62`) |
| On-device quality critique + rewrite | `ResumeQualityPanel`, `src/lib/webllm/` | **No** — tab, post-drop only |

Every pre-drop line is audit framing: the header subtitle is `"A parser audit for your resume — not a judge"` (`App.tsx:91`), the hero says OfflineCV `"shows you what a recruiter or screener reads back from your resume"` (`App.tsx:139`), and the block under the drop zone extends the same mirror metaphor. The product asks for a résumé before telling the visitor what it will do with it.

## Where the chip row is today

```tsx
chips={
  <>
    <Chip icon="⚡">A few seconds</Chip>
    <Chip icon="🔒">Your file never leaves your device</Chip>
    <Chip icon="✓">No account, no email</Chip>
    <Chip icon="🔁">Same PDF, same score</Chip>
  </>
}
```

`App.tsx:92-99`, rendered by `PageShell.tsx:79` directly under the header.

Four claims about *properties* — fast, private, no-signup, deterministic — and none about *capabilities*. The row occupies the most valuable strip on the page and spends it on reassurance for a product the visitor cannot yet describe.

## Expected

Replace the chip row with a three-lane capability strip: what the product does, in order, under one statement of the local-processing default.

```
  Read it                    Fix it                     Match & find
  what a parser actually     edit inline, export a      score against a JD,
  pulls out of your PDF      clean ATS-safe PDF         rank real postings
  ─────────────────────────────────────────────────────────────────────
  Runs on your device
```

Behaviour:

- **Pre-drop:** static and non-interactive. It is a description, not navigation — nothing to click before there is a résumé to act on. No new workflow surface, no second entry point competing with the drop zone.
- **Post-drop:** the reached lane is marked. Which lane is current comes from existing state (`state.phase`, and the active tab once `ResultDetailTabs` is mounted) — no new state machine.
- Composed from `Card` / `Chip` / `StatusBadge` out of the `@design-system` barrel. If a variant is missing, add it to the shared piece.

### The privacy line is the load-bearing part — get its scope exactly right

The current claim is **file-scoped** on purpose: *"Your file never leaves your device."* It is true without qualification. A capability strip that spans job search cannot inherit that scope unchanged, because job search **does** send data on an explicit user action, and the app already says so in two places:

- `FindJobsPanel.tsx:146` — *"Your resume text never leaves this browser — only the keywords below are sent."*
- `FindJobsPanel.tsx:238, 255` — *"Only your search keywords are sent, and only when you click Search."*

And the BYOK provider path is a genuine, user-initiated cloud egress.

So the rail states the **default** — processing happens on device — and must not be phrased as an absolute ("nothing ever leaves", "100% offline", "zero network"). The per-lane exceptions stay stated where the user meets them; the strip does not restate them and does not override them. Any wording that would make `FindJobsPanel`'s existing sentences read as a correction to the strip is wrong. Nothing here may outrun `README.md#telemetry`.

If the exact phrasing is contested at review, the safe fallback is to keep the current file-scoped sentence verbatim as the rail and put capability copy above it.

## Acceptance criteria

- A visitor who has **not** dropped a file can name all three lanes — read / fix / match & find — from the pre-drop page alone.
- The strip is non-interactive pre-drop; it introduces no second entry point competing with the drop zone.
- The local-processing statement describes the default and is not phrased as an absolute; it does not contradict, weaken, or duplicate `FindJobsPanel`'s existing keyword-egress sentences.
- No claim in the strip goes beyond what `README.md#telemetry` supports.
- Built from existing `@design-system` exports; any new variant lands in the shared component, not beside it. No new modal, dropzone, or banner.
- No emoji as icons (see the sibling icon issue — this strip inherits whatever set it lands).
- The drop zone remains the single most prominent element on the page; the strip does not out-weight it.
- Legible at 375px without horizontal scroll; lanes stack rather than truncate.
- Contrast ≥4.5:1 for body copy in **both** themes, checked independently.
- Chrome-only change: no pipeline, no score, no `ATS_SCORE_ALGO_VERSION` movement, round-trip invariants untouched. `npm run verify` green.

## Notes

The four claims being displaced are not worthless — determinism and no-signup are real differentiators. They should move, not vanish: the quiet block below the drop zone (`App.tsx:191+`) already exists for exactly this kind of supporting context. Decide their destination in this issue rather than deleting them silently.

**Ordering with the JD lane.** `/jd-fit/` is alpha and its cross-sell banner is deliberately flag-gated off (`App.tsx:59-62`). The strip names *match against a JD* as a capability, which is true and shipped — but it must not become a back-door promotion of a surface the flag is deliberately hiding. Describe the capability; do not link to `/jd-fit/` from the strip while the flag is off.

Sequencing within #511: the icon issue should land first so this strip is built with the final icon set. The hero-repositioning issue is its natural pair — headline and strip should be written together so they do not repeat each other.

### LINKED ISSUE #518: [landing] hero leads with an industry statistic, not with what OfflineCV does
Parent: #511

## Summary

The landing hero leads with an industry statistic and never states what the product is for. Three of its four elements are about the state of hiring; one is about OfflineCV.

```
AI is now part of most hiring pipelines —
and nearly half of job seekers say they've lost trust in
a process they can't see into.¹

OfflineCV shows you what a recruiter or screener reads back
from your resume — free, private, and open-source.

Source: Greenhouse, 2025 AI in Hiring Report (4,100+ job seekers and hiring managers)
```

`src/App.tsx:118-146`.

The largest type on the page (`text-2xl sm:text-3xl`) is spent on a claim about the industry. The product sentence is one step down (`text-base sm:text-lg`) and describes a single lane — reading back a résumé — while the product also matches against job descriptions, ranks postings by fit, and rewrites sections on device.

## What is deliberate and must be preserved

The comment at `App.tsx:112-117` records real, hard-won decisions:

> One consolidated hero message (internal #265): a single, non-hyperbolic headline — no "they don't read your PDF" claim and no "parser" jargon — that says what OfflineCV does in one angle. The trust stat is the one supporting line; everything else (the recruiter-agent context) moves to the quiet block below the drop zone so the hero isn't three competing messages.

So: **one message, not three; no hyperbole; no parser jargon; supporting context stays below the fold.** Those constraints are not up for renegotiation here — this issue works inside them. The observation is narrower: the stated intent was that the headline *"says what OfflineCV does"*, and the shipped headline says what the hiring industry does. The stat was meant to be the supporting line and became the lead.

Note also that the source is cited **twice** — once as a superscript footnote link (`App.tsx:126-134`) and again as a full source line beneath (`App.tsx:141-144`). Two citations of one statistic in a block that is trying to be one message.

## Expected

Invert the emphasis. Lead with the capability, keep the stat as the proof it was intended to be.

Direction (copy is for review, not final):

| Slot | Now | After |
|---|---|---|
| Headline (largest) | *"AI is now part of most hiring pipelines — and nearly half of job seekers…"* | a capability-led line covering the whole product, e.g. **"Your whole job search — on your device."** |
| Sub (one step down) | *"OfflineCV shows you what a recruiter or screener reads back…"* | what the product does across lanes, in one sentence |
| Proof (small) | the stat, twice-cited | the stat, once, as the reason the product exists |

Constraints on the copy:

- **One message.** The headline may not become three claims. If the capability line and the strip (sibling issue) both enumerate lanes, one of them must stop — decide which and write them together.
- **No hyperbole, no parser jargon** — the #265 rulings hold verbatim.
- **No absolute privacy claim.** "On your device" describes the default. The job-search lane sends keywords on an explicit click (`FindJobsPanel.tsx:146,238,255`) and the BYOK path is real cloud egress. Nothing in the headline may outrun `README.md#telemetry` or contradict those sentences.
- **Cite the stat once.** Keep whichever of the two citations is more useful and drop the other.
- **Keep the Greenhouse attribution and link** if the stat survives at all — it is a real source and the link is `nofollow noopener noreferrer` already.

The prose width caps (`max-w-2xl`, `text-balance`, `text-pretty`) and the surrounding `Card className="bg-surface-card-warm"` are fine and should stay.

## Acceptance criteria

- The largest type on the pre-drop page describes what OfflineCV does, not what the hiring industry does.
- The headline covers the product, not one lane of it — a reader cannot come away thinking OfflineCV only inspects PDFs.
- The hero is still **one** message: no third competing claim is added, and the below-fold context block is not duplicated upward.
- No hyperbole and no "parser" jargon — the #265 constraints, re-verified in review.
- The trust statistic, if kept, is cited exactly once, with its Greenhouse attribution intact.
- No privacy claim in the hero exceeds what `README.md#telemetry` supports or conflicts with `FindJobsPanel`'s keyword-egress copy.
- The drop zone remains the most prominent interactive element; the hero does not grow tall enough to push it below the fold at 768px.
- Legible at 375px; line length stays inside the existing prose caps.
- Contrast ≥4.5:1 in **both** themes, checked independently.
- Copy-only change: no pipeline, score, or export behaviour touched. `npm run verify` green.

## Notes

Pairs with the capability-strip issue — headline and strip are one editorial decision and should be reviewed as one diff even if they land separately. If they land apart, the strip should land second so it can be written against whatever the headline ends up claiming.

The `A parser audit for your resume — not a judge` subtitle in `PageShell` (`App.tsx:91`) is the same problem one size smaller: it is lane-scoped, and it is what a visitor reads first on **both** surfaces. Whether it changes with the hero or stays as an honest scoping of the `/` lane is a call for this issue to make explicitly, not to leave unexamined.

s-annam:
## Summary

Six sub-issues of epic #511, landed together because they touch the same pixels: a first-time visitor could not learn that OfflineCV does anything beyond auditing a PDF, and the surfaces that would tell them were the same ones that looked unfinished.

- **#517** — the landing trust chips become a `CapabilityStrip` naming all three lanes (Read it / Fix it / Match & find) under a scoped "runs on your device by default" rail.
- **#518** — the hero leads with what the product does; the Greenhouse figure drops to the supporting proof line, citation intact.
- **#513** — `brand-amber`/`-light` → `accent-primary`/`-hover` across 24 files. The zero-consumer `brand-navy`, `brand-navy-dark`, `brand-cream` and the `accent-forward` *foreground* token are deleted, and the backwards hover pair is fixed. A rename, not a repaint — the generic default palette is deliberate.
- **#516** — tab selection is rebuilt as a recessed track with a raised active surface, so selection carries font-weight and elevation, never colour alone.
- **#514** — the chip row's colour emoji become a hand-inlined SVG icon set. **No new runtime dependency.** The deliberate `✓ ★ ⚠︎` text-presentation marks are left alone.
- **#515** — the logo mark stops rendering the leftover resumelint `R`.

Two defects surfaced by review are also fixed at the primitive: `StatusBadge` gains `w-fit` (it was stretching into a full-width bar inside any column flex — this was already shipping in `CritiquePanel`), and `CountBadge` gains a border that clears its new track at 4.34:1 light / 4.04:1 dark.

**Privacy copy stays scoped.** The job-search lane sends keywords on an explicit click and a `VITE_POSTHOG_KEY` build ships analytics, so no surface claims more than "by default" (the BYOK path of #320 is not in the tree yet) — review caught and reverted an unhedged absolute in the hero headline.

**Chrome, copy and tokens only.** `ATS_SCORE_ALGO_VERSION` is unmoved; no parser, score or PDF-render module is touched.

Closes #513
Closes #514
Closes #515
Closes #516
Closes #517
Closes #518
Refs #511

`#519` (tab descriptions) was deliberately dropped from this batch — its pre-drop-preview half overlaps `CapabilityStrip`, and the epic says not to ship both.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — 2509 passed / 9 skipped
- [x] `npm run build` succeeds; renamed utilities verified present in the emitted CSS (not just in source)
- [x] `fallow audit` — dead-code section clear; remaining duplication is pre-existing
- [x] Contrast verified independently against the light **and** dark blocks of `tokens.css`, not inferred from light
- [x] Manually verified in `npm run dev` — **not done; worth a reviewer's eye**, since most of this diff is visual

No fixture binaries added or changed, so the fixture-PII preflight does not apply.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #515 | unreported (requested: sonnet) | medium |
| Implementation — #513 | Claude Sonnet 5 | high |
| Implementation — #514 | Claude Sonnet 5 | high |
| Implementation — #516 | Claude Sonnet 4.6 | high |
| Implementation — #518 | ambiguous self-report (requested: sonnet) | medium |
| Implementation — #517 | Claude Sonnet 5 | high |
| Adversarial review — round 1 | Claude Opus 4.8 | high |
| Review fixes | Claude Opus 4.8 | high |
| Adversarial review — round 2 | Claude Opus 4.8 | high |
| Round-2 fixes + orchestration + PR | Claude Opus 4.8 | high |

Two rows are not resolvable and are left honest rather than guessed: #515's subagent never returned its model line (its one-line change was verified directly), and #518's reported two names in one answer.

## Adversarial review

Two adversarial review rounds ran on the accumulated diff **before** this PR was opened, plus a third fix pass applied in-loop. Reviewers were instructed to try to break the change, and to reproduce suspected defects rather than reason about them — both rounds built the CSS and inspected emitted output. Recorded here because a review that found real bugs is an audit artifact, not a transient gate.

### Round 1 — 6 blocking, all fixed

1. **Unhedged absolute privacy claim in the hero headline.** "Your whole job search — on your device." — but the job-search lane sends keywords on click and a `VITE_POSTHOG_KEY` build ships analytics. It contradicted the scoping convention asserted in its own file's comment. → hedged to "…by default."
2. **Count pill invisible on the new tab track** — `CountBadge`'s fill and #516's track were the same token, 1.00:1 in both themes. It defeated the intent documented at `ResultDetailTabs.tsx:69-71`.
3. **Inactive tab hover was visually identical to the selected tab** — hover acquired both of the active tab's colour cues, leaving only font-weight to signal selection. This defeated #516's entire purpose on the most common pointer state.
4. **`CapabilityStrip` lane pills rendered as full-width bars** — `StatusBadge` is `inline-flex` with no `w-fit`, blockified and stretched by a column flex.
5. **Three dead icon exports** — #514 created four icons and wired them into the chip row; #517 deleted that row in the same branch. A pure cross-issue interaction neither issue could see alone; caught by the whole-tree `fallow` pass.
6. **Docs still cited the deleted `text-brand-amber` token** — #513 updated `README.md` only, so a contributor following `CLAUDE.md`/`CONTRIBUTING.md` would write a class that silently compiles to nothing.

The fix pass found that the reviewer's own suggested remedies for (2) and (3) were **no-ops in dark** — `--color-bg-hover` and `--color-bg-subtle` are both `#334155` — and solved them with `border-strong` instead.

### Round 2 — 3 blocking, all fixed

1. **`eslint.config.js:29,34`** still named `text-brand-amber` in the rule message — the highest-traffic naming reference in the repo, higher than either doc fixed in round 1.
2. **`border-strong` did not actually clear WCAG 1.4.11.** The fix pass's quoted ratios (2.29/2.16) were also slightly wrong; the real values were 2.34:1 light / 2.18:1 dark, both under the 3:1 non-text threshold that the border alone had to carry. → `border-content-muted`, 4.34:1 light / 4.04:1 dark.
3. **The round-1 fix's justifying comment was false**, and the case it denied was shipping: `CritiquePanel.tsx:64` is a column flex whose badge stretches into a full-width bar **in production today**. → fixed at the primitive with `w-fit` on `StatusBadge`, which resolves both instances and retires the workaround.

Round 2's fixes were applied in-loop after the 2-round cap and re-verified (typecheck, lint, 2509 tests, build, emitted-CSS check) but did **not** themselves go through a review round. Two are cosmetic (a lint-message string, docblock ratios); the third is `w-fit` on `StatusBadge`, which touches 9 call sites — the other 8 are row-flex, where it is a no-op.

### Surviving nits — not blocking, worth a reviewer's eye

- `SourceDiagnosticsPanel`'s `SegmentButton` is a near-clone of `Tab` that received none of #516's hover-affordance work — a sibling instance worth folding into the same lane in a follow-up.
- `GitHubStarCta.tsx:106` — `⭐` is the one remaining colour emoji in live chrome; reads as sentence decoration rather than an icon, so it was left out of #514's scope.
- Tabs' hover `ring-inset` leaks into the focus ring, so a simultaneously hovered+focused tab draws its focus ring inside the border box. Still clearly visible, just a different shape.
- `CapabilityStrip` lanes are bare `<div>`s; `<ul>`/`<li>` would give a screen reader the "list of 3" count.
- `CapabilityStrip` uses `StatusBadge tone="info"` for a static capability label — borrowing a parse-status vocabulary for a non-status. Cosmetically identical, semantically off.
- The `tab` variant shrinks the touch target from ~40px to ~32px. Clears WCAG 2.5.8 AA (24px), but is a real shrink on a 375px viewport.
- The selected-vs-track surface step is 1.10:1 light / 1.41:1 dark — under 1.4.11's 3:1 for non-text indicators. Not blocking, because selection also carries `font-semibold` and `aria-selected`, and inactive **text** contrast passes in both themes (9.45:1 / 6.97:1).
- `.gitignore`'s `oss-translate` entry is unrelated to this epic and rides this branch by explicit maintainer decision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ labeled in-app as a remote/tech-heavy **sample, not every job**.
 Colors are plain CSS custom properties split into two layers:
 
 - **`src/design-system/styles/theme.css`** — the semantic vocabulary
-  (`--color-surface-card`, `--color-content-primary`, `--color-brand-amber`, …).
+  (`--color-surface-card`, `--color-content-primary`, `--color-accent-primary`, …).
   This is the **stable contract**: it generates the Tailwind classes
   (`bg-surface-card`, `text-content-primary`, …) that the components use. Names
   here don't change.
@@ -207,6 +207,22 @@ Colors are plain CSS custom properties split into two layers:
   vocabulary, light and dark. The default is a **generic, accessible palette
   (slate neutrals + a blue accent)**, not a specific product brand. This is the
   **swappable layer**.
+
+> **Deprecated names (removed 2026-07, #513):** `--color-brand-amber` was an
+> exact duplicate of `--color-accent-primary` (same blue in both themes).
+> `--color-brand-amber-light` was a *near*-duplicate of
+> `--color-accent-primary-hover`: identical in dark (`#93c5fd`), but in light
+> it held `#3b82f6` against the hover token's `#1d4ed8`. Folding them
+> deliberately corrects the light-mode hover on `Button variant="primary"` and
+> the ConsentDialog licence link from a lighter blue to a darker one. Every
+> consumer now points at the `accent-primary` pair, which is canonical.
+> `--color-brand-navy`, `--color-brand-navy-dark`, and
+> `--color-brand-cream` are gone — they had zero consumers in either theme.
+> `--color-accent-forward` (the foreground/text variant) is also gone — only
+> its background wash, `--color-accent-forward-bg`, had consumers; wiring a
+> foreground use of the forward accent is tracked separately. If your
+> override layer still defines any of these names, drop them — they no
+> longer resolve to anything.
 
 You can reskin offlinecv to your own brand **without forking this repo**, two
 ways:
@@ -219,12 +235,12 @@ Import your own stylesheet *after* offlinecv's and redefine the same
 ```css
 /* your-brand.css — imported after offlinecv's styles */
 :root {
-  --color-brand-amber: #ff6b35;
+  --color-accent-primary: #ff6b35;
   --color-bg-card: #fffaf5;
 }
 @media (prefers-color-scheme: dark) {
   :root {
-    --color-brand-amber: #ffa07a;
+    --color-accent-primary: #ffa07a;
     --color-bg-card: #1c1410;
   }
 }
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,13 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2026 The offlinecv Authors
 
-import { Card, Chip, ErrorState, ErrorBoundary, Button } from "@design-system";
+import {
+  Card,
+  CapabilityStrip,
+  ErrorState,
+  ErrorBoundary,
+  Button,
+} from "@design-system";
 import { DropZone } from "./components/DropZone";
 import { Result } from "./components/Result";
 import { ReconstructedResume } from "./components/features/ReconstructedResume.tsx";
@@ -90,14 +96,7 @@ export default function App() {
     <PageShell
       subtitle="A parser audit for your resume — not a judge"
       badge="alpha"
-      chips={
-        <>
-          <Chip icon="⚡">A few seconds</Chip>
-          <Chip icon="🔒">Your file never leaves your device</Chip>
-          <Chip icon="✓">No account, no email</Chip>
-          <Chip icon="🔁">Same PDF, same score</Chip>
-        </>
-      }
+      chips={<CapabilityStrip />}
     >
       {(state.phase === "idle" ||
         state.phase === "parsing" ||
@@ -115,30 +114,32 @@ export default function App() {
             // The trust stat is the one supporting line; everything else (the
             // recruiter-agent context) moves to the quiet block below the drop
             // zone so the hero isn't three competing messages.
+            //
+            // The headline's "by default" is NOT hedging for its own sake and
+            // must not be dropped: the job-search lane egresses keywords on an
+            // explicit click (see FindJobsPanel), a build with
+            // VITE_POSTHOG_KEY set (the hosted one) ships analytics, and the
+            // BYOK provider path (#320, not in the tree yet — see
+            // CapabilityStrip) will be real cloud egress once it lands. The
+            // first two are true today and already load-bearing on their own.
+            // An unqualified "on your device"
+            // would be the only absolute privacy claim in the app — every
+            // other surface is already scoped the same way (PageShell's
+            // footer, CapabilityStrip's rail).
             <Card className="flex flex-col items-center gap-5 bg-surface-card-warm">
               <div className="flex max-w-2xl flex-col gap-4 text-center">
-                <h2 className="text-balance text-2xl font-normal leading-snug tracking-tight text-content-secondary sm:text-3xl">
-                  <span className="font-semibold text-content-primary">
-                    AI is now part of most hiring pipelines —
-                  </span>{" "}
-                  and nearly half of job seekers say they&apos;ve lost trust in
-                  a process they can&apos;t see into.
-                  <a
-                    href="https://www.greenhouse.com/newsroom/an-ai-trust-crisis-70-of-hiring-managers-trust-ai-to-make-faster-and-better-hiring-decisions-only-8-of-job-seekers-call-it-fair"
-                    target="_blank"
-                    rel="nofollow noopener noreferrer"
-                    aria-label="Source: Greenhouse, 2025 AI in Hiring Report"
-                    className="align-super text-xs text-brand-amber hover:underline"
-                  >
-                    1
-                  </a>
+                <h2 className="text-balance text-2xl font-semibold leading-snug tracking-tight text-content-primary sm:text-3xl">
+                  Your whole job search — on your device by default.
                 </h2>
-                <p className="text-pretty text-base font-medium text-content-primary sm:text-lg">
-                  OfflineCV shows you what a recruiter or screener reads back
-                  from your resume — free, private, and open-source.
+                <p className="text-pretty text-base font-medium text-content-secondary sm:text-lg">
+                  OfflineCV is a free, open-source job-search workbench —
+                  read your resume, fix it, and match it against real
+                  postings, all in your browser. Nothing to install, no
+                  account to create.
                 </p>
                 <p className="text-xs text-content-muted">
-                  Source:{" "}
+                  Built in response to a hiring process job seekers say they
+                  can&apos;t see into — source:{" "}
                   <a
                     href="https://www.greenhouse.com/newsroom/an-ai-trust-crisis-70-of-hiring-managers-trust-ai-to-make-faster-and-better-hiring-decisions-only-8-of-job-seekers-call-it-fair"
                     target="_blank"
@@ -190,6 +191,11 @@ export default function App() {
             // file-scoped, determinism is score-scoped. Per the public-copy
             // policy (internal #24) we never name other products here — the
             // trend is described generically, no links to specific tools.
+            //
+            // #517: this block also now carries the three claims displaced
+            // from the old trust-chip row (speed, no-signup, determinism) —
+            // real differentiators that moved here rather than vanishing when
+            // the chip row became the capability strip above.
             <div className="rounded-lg border border-border-light bg-surface-subtle px-4 py-3">
               <p className="mx-auto max-w-3xl text-pretty text-sm text-content-secondary">
                 <span className="font-medium text-content-primary">
@@ -199,7 +205,9 @@ export default function App() {
                 the candidate-side
                 mirror: it shows you what survives the parse — before you hit
                 submit. The score isn&apos;t a verdict on you as a candidate —
-                it measures how well a machine can read your resume.
+                it measures how well a machine can read your resume. It takes
+                a few seconds, needs no account or email, and the same PDF
+                always gets the same score.
               </p>
             </div>
           )}
--- a/src/design-system/shared/Tabs.tsx
+++ b/src/design-system/shared/Tabs.tsx
@@ -18,8 +18,15 @@
  *
  * Design rules (CLAUDE.md):
  *   – Semantic tokens only; no hardcoded hex or raw palette classes.
- *   – Triggers use the <Button> primitive — no raw <button> in feature/primitive code.
+ *   – Triggers use the <Button> primitive (variant="tab") — no raw <button> in
+ *     feature/primitive code.
  *   – Inactive panels stay mounted (hidden attr) so child UI state survives a switch.
+ *
+ * Selection affordance (#516): TabList is a recessed track (bg-surface-subtle);
+ * the active Tab sits on bg-surface-card + shadow-xs + font-semibold — a real
+ * surface step, not a 2px underline competing with the track's own border (the
+ * pre-#516 bug). Never colour alone: the weight/surface step still resolves in
+ * a greyscale render.
  */
 
 import {
@@ -110,7 +117,7 @@ export function TabList({ "aria-label": ariaLabel, children }: TabListProps) {
       role="tablist"
       aria-label={ariaLabel}
       onKeyDown={onKeyDown}
-      className="flex gap-1 overflow-x-auto overflow-y-hidden border-b border-border-light"
+      className="flex gap-1 overflow-x-auto overflow-y-hidden rounded-md border border-border-light bg-surface-subtle p-1"
     >
       {children}
     </div>
@@ -134,13 +141,32 @@ interface TabProps {
 export function Tab({ id, children, count, warn }: TabProps) {
   const { value, onValueChange, baseId } = useTabsContext("Tab");
   const isActive = value === id;
+  // Selection carries a real surface (bg-surface-card, matching the panel
+  // beneath it) plus a font-weight step — never colour alone, so the row
+  // still resolves in a greyscale render (#516). The inactive track is
+  // bg-surface-subtle (set on TabList), so the active tab visibly "pops"
+  // off it rather than relying on a 2px underline sitting on the TabList's
+  // own border (the pre-#516 illegibility bug).
+  //
+  // Hover on an INACTIVE tab must not borrow the selected tab's surface:
+  // `hover:bg-surface-card` made a merely-pointed-at tab look selected, with
+  // only font-weight left to tell them apart. It hovers to an outline instead
+  // of a fill — `bg-surface-hover` alone cannot carry it, because
+  // `--color-bg-hover` and `--color-bg-subtle` (the track) are the SAME value
+  // in dark (#334155 → 1.00:1). The inset `ring-border-strong` reads at
+  // 2.34:1 light / 2.18:1 dark against the track — enough to be perceptible
+  // for a hover-only affordance (not an enumerated 1.4.11 state), but BELOW
+  // that criterion's 3:1, so do not reuse this token for a persistent
+  // boundary (see `CountBadge.tsx`, which needs `content-muted` for exactly
+  // that reason). It stays visually distinct from selection, which is a
+  // filled surface with no ring.
   const activeCls = isActive
-    ? "border-brand-amber text-content-primary font-semibold"
-    : "border-transparent text-content-secondary font-medium hover:text-content-primary";
+    ? "bg-surface-card text-content-primary font-semibold shadow-xs"
+    : "bg-transparent text-content-secondary font-medium hover:bg-surface-hover hover:text-content-primary hover:ring-1 hover:ring-inset hover:ring-border-strong";
 
   return (
     <Button
-      variant="ghost"
+      variant="tab"
       role="tab"
       id={tabId(baseId, id)}
       data-tab-id={id}
@@ -148,7 +174,7 @@ export function Tab({ id, children, count, warn }: TabProps) {
       aria-controls={panelId(baseId, id)}
       tabIndex={isActive ? 0 : -1}
       onClick={() => onValueChange(id)}
-      className={`-mb-px rounded-none border-b-2 px-3 py-2 text-base hover:bg-transparent ${activeCls}`}
+      className={activeCls}
     >
       {children}
       <CountBadge count={count} />
--- a/.gitignore
+++ b/.gitignore
@@ -60,6 +60,13 @@ __pycache__/
 # Local-only internal working directory — never ship in the OSS repo
 internal/
 
+# Maintainer-only skill whose source of truth lives in the private
+# offlinecv-internal repo (internal/.claude/skills/oss-translate). It is
+# surfaced here as a symlink so Claude Code discovers it as /oss-translate,
+# but it must never ship in the public repo — it translates private internal
+# issues into public ones and knows partner/PII sanitization specifics.
+/.claude/skills/oss-translate
+
 # wrangler local state
 .wrangler/
 .dev.vars*
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -136,7 +136,7 @@ Strict 3-tier architecture. Primitives + shared-composed live in `src/design-sys
 
 ## Styling & tokens
 
-- **Semantic tokens are canonical.** Style with semantic Tailwind classes: `bg-surface-card`, `text-content-primary`, `border-border-light`, `text-brand-amber`. Vocabulary lives in `src/design-system/styles/theme.css`; values in `tokens.css`.
+- **Semantic tokens are canonical.** Style with semantic Tailwind classes: `bg-surface-card`, `text-content-primary`, `border-border-light`, `text-accent-primary`. Vocabulary lives in `src/design-system/styles/theme.css`; values in `tokens.css`.
 - **No hardcoded colors.** Never a hex (`#ef4444`), never a raw palette class (`bg-red-500`, `text-slate-400`), never a manual `dark:` colour variant, in feature code.
 - **Typography** rides global settings — never hand-styled inline.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,7 +119,7 @@ already exists:
 
 **No hardcoded colours.** Style with semantic Tailwind tokens
 (`bg-surface-card`, `text-content-primary`, `border-border-light`,
-`text-brand-amber`, …). Raw palette classes (`bg-red-500`,
+`text-accent-primary`, …). Raw palette classes (`bg-red-500`,
 `text-slate-400`) and manual `dark:` variants are anti-patterns —
 dark mode is handled by the token layer, not inline overrides.
 
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -26,12 +26,12 @@ function restrictedSyntaxRules() {
     {
       selector: `Literal[value=/${PALETTE_RE}/]`,
       message:
-        "Use semantic tokens (bg-surface-card, text-content-primary, border-border-light, text-brand-amber, …) instead of raw Tailwind palette classes.",
+        "Use semantic tokens (bg-surface-card, text-content-primary, border-border-light, text-accent-primary, …) instead of raw Tailwind palette classes.",
     },
     {
       selector: `TemplateElement[value.raw=/${PALETTE_RE}/]`,
       message:
-        "Use semantic tokens (bg-surface-card, text-content-primary, border-border-light, text-brand-amber, …) instead of raw Tailwind palette classes.",
+        "Use semantic tokens (bg-surface-card, text-content-primary, border-border-light, text-accent-primary, …) instead of raw Tailwind palette classes.",
     },
     // Manual dark: colour variants
     {
--- a/src/components/features/ConsentDialog.tsx
+++ b/src/components/features/ConsentDialog.tsx
@@ -62,7 +62,7 @@ export function ConsentDialog({
               href={model.licenseUrl}
               target="_blank"
               rel="noopener noreferrer"
-              className="text-brand-amber underline underline-offset-2 hover:text-brand-amber-light"
+              className="text-accent-primary underline underline-offset-2 hover:text-accent-primary-hover"
             >
               vendor's terms of use
             </a>
--- a/src/components/features/ContactDetails.tsx
+++ b/src/components/features/ContactDetails.tsx
@@ -275,7 +275,7 @@ function renderLink(
             href={field.value}
             target="_blank"
             rel="noopener noreferrer"
-            className="text-brand-amber hover:underline"
+            className="text-accent-primary hover:underline"
             aria-label={`Open ${field.label} in a new tab`}
           >
             ↗
@@ -310,7 +310,7 @@ function renderLink(
       href={field.value}
       target="_blank"
       rel="noopener noreferrer"
-      className="text-brand-amber hover:underline"
+      className="text-accent-primary hover:underline"
     >
       {formatLinkDisplay(field.value)}
     </a>
--- a/src/components/features/ContactExtraLinks.tsx
+++ b/src/components/features/ContactExtraLinks.tsx
@@ -58,7 +58,7 @@ export function ContactExtraLinks({
               href={profile.url}
               target="_blank"
               rel="noopener noreferrer"
-              className="text-brand-amber hover:underline"
+              className="text-accent-primary hover:underline"
               aria-label={`Open ${profile.network} in a new tab`}
             >
               ↗
--- a/src/components/features/CritiquePanel.tsx
+++ b/src/components/features/CritiquePanel.tsx
@@ -80,7 +80,7 @@ function BulletFindingRow({
           variant="link"
           size="sm"
           onClick={onGoToRewrite}
-          className="self-start text-[11px] font-medium text-brand-amber"
+          className="self-start text-[11px] font-medium text-accent-primary"
           aria-label="Go to Reconstructed resume tab to rewrite this bullet"
         >
           Rewrite this section →
--- a/src/components/features/DownloadReportDialog.tsx
+++ b/src/components/features/DownloadReportDialog.tsx
@@ -102,7 +102,7 @@ export function ReportDownloadControl({
                   checked={format === value}
                   disabled={isGenerating}
                   onChange={() => setFormat(value)}
-                  className="h-4 w-4 accent-brand-amber"
+                  className="h-4 w-4 accent-accent-primary"
                 />
                 {label}
               </label>
@@ -119,7 +119,7 @@ export function ReportDownloadControl({
               checked={includeIdentity}
               disabled={isGenerating}
               onChange={(e) => setIncludeIdentity(e.target.checked)}
-              className="h-4 w-4 accent-brand-amber"
+              className="h-4 w-4 accent-accent-primary"
             />
             Include my name and contact details
           </label>
--- a/src/components/features/FeedbackPanel.tsx
+++ b/src/components/features/FeedbackPanel.tsx
@@ -172,7 +172,7 @@ function FeedbackRatingForm() {
 
     return (
       <div className="flex flex-col gap-3">
-        <Card className="flex flex-col gap-1 border-l-4 border-l-brand-amber bg-accent-forward-bg shadow-sm">
+        <Card className="flex flex-col gap-1 border-l-4 border-l-accent-primary bg-accent-forward-bg shadow-sm">
           <div ref={thanksRef} tabIndex={-1} className="outline-hidden">
             <p className="text-sm font-semibold text-content-primary">
               Thanks for your feedback!
@@ -243,7 +243,7 @@ function FeedbackRatingForm() {
 
   // ── Full panel (mode === "full", or compact strip expanded by star click) ──
   return (
-    <Card className="flex flex-col gap-4 border-l-4 border-l-brand-amber bg-accent-forward-bg shadow-sm">
+    <Card className="flex flex-col gap-4 border-l-4 border-l-accent-primary bg-accent-forward-bg shadow-sm">
       <form className="flex flex-col gap-4" onSubmit={handleSubmit} noValidate>
         {/* Collapsed line: heading + inline rating on one row. */}
         <div className="flex flex-col gap-1">
@@ -319,7 +319,7 @@ function FeedbackRatingForm() {
                 disabled={submitting}
                 rows={3}
                 placeholder="Tell us what worked, what didn't, or what you'd change…"
-                className="w-full resize-y rounded border border-border-light bg-surface-card px-2 py-1.5 text-sm text-content-primary placeholder:text-content-muted focus:outline-hidden focus-visible:ring-2 focus-visible:ring-brand-amber"
+                className="w-full resize-y rounded border border-border-light bg-surface-card px-2 py-1.5 text-sm text-content-primary placeholder:text-content-muted focus:outline-hidden focus-visible:ring-2 focus-visible:ring-accent-primary"
               />
             </div>
 
@@ -335,7 +335,7 @@ function FeedbackRatingForm() {
                   checked={wantsContact}
                   disabled={submitting}
                   onChange={(e) => setWantsContact(e.target.checked)}
-                  className="h-4 w-4 accent-brand-amber"
+                  className="h-4 w-4 accent-accent-primary"
                 />
                 I'd like the team to follow up with me
               </label>
@@ -355,7 +355,7 @@ function FeedbackRatingForm() {
                     onChange={(e) => setEmail(e.target.value)}
                     disabled={submitting}
                     placeholder="you@example.com"
-                    className="w-full rounded border border-border-light bg-surface-card px-2 py-1.5 text-sm text-content-primary placeholder:text-content-muted focus:outline-hidden focus-visible:ring-2 focus-visible:ring-brand-amber"
+                    className="w-full rounded border border-border-light bg-surface-card px-2 py-1.5 text-sm text-content-primary placeholder:text-content-muted focus:outline-hidden focus-visible:ring-2 focus-visible:ring-accent-primary"
                   />
                 </div>
               )}
--- a/src/components/features/FindJobsPanel.tsx
+++ b/src/components/features/FindJobsPanel.tsx
@@ -196,7 +196,7 @@ export function FindJobsPanel({ parsed }: FindJobsPanelProps) {
                   addSkill();
                 }
               }}
-              className="min-w-0 max-w-56 flex-1 rounded border border-border-light bg-surface-card px-2 py-1 text-xs text-content-primary outline-hidden focus:ring-1 focus:ring-brand-amber"
+              className="min-w-0 max-w-56 flex-1 rounded border border-border-light bg-surface-card px-2 py-1 text-xs text-content-primary outline-hidden focus:ring-1 focus:ring-accent-primary"
             />
             <Button
               variant="ghost"
@@ -227,7 +227,7 @@ export function FindJobsPanel({ parsed }: FindJobsPanelProps) {
               href={link.url}
               target="_blank"
               rel="noopener noreferrer"
-              className="inline-flex items-center gap-1 rounded px-2 py-1.5 text-xs font-medium text-content-secondary transition-colors hover:bg-surface-subtle focus:outline-hidden focus-visible:ring-2 focus-visible:ring-brand-amber"
+              className="inline-flex items-center gap-1 rounded px-2 py-1.5 text-xs font-medium text-content-secondary transition-colors hover:bg-surface-subtle focus:outline-hidden focus-visible:ring-2 focus-visible:ring-accent-primary"
             >
               {link.label}
               <span aria-hidden="true">↗</span>
--- a/src/components/features/JobResultCard.tsx
+++ b/src/components/features/JobResultCard.tsx
@@ -72,7 +72,7 @@ export function JobResultCard({ job }: { job: RankedJob }) {
           href={posting.url}
           target="_blank"
           rel="noopener noreferrer"
-          className="text-xs font-medium text-content-secondary transition-colors hover:text-content-primary focus:outline-hidden focus-visible:ring-2 focus-visible:ring-brand-amber"
+          className="text-xs font-medium text-content-secondary transition-colors hover:text-content-primary focus:outline-hidden focus-visible:ring-2 focus-visible:ring-accent-primary"
         >
           Open posting <span aria-hidden="true">↗</span>
         </a>
--- a/src/components/features/ModelSelector.tsx
+++ b/src/components/features/ModelSelector.tsx
@@ -438,7 +438,7 @@ export function ModelRow({
         aria-pressed={selected}
         className={`flex w-full flex-col items-stretch gap-1 rounded-md border px-3 py-2 text-left ${
           selected
-            ? "border-brand-amber bg-feedback-success-bg"
+            ? "border-accent-primary bg-feedback-success-bg"
             : "border-border-light bg-surface-card hover:border-border hover:bg-surface-hover"
         }`}
       >
--- a/src/components/features/PageShell.tsx
+++ b/src/components/features/PageShell.tsx
@@ -25,7 +25,7 @@ export interface PageShellProps {
   subtitle: string;
   /** Small uppercase badge after the wordmark (e.g. "alpha", "JD Fit"). */
   badge: string;
-  /** Optional chip row under the header. */
+  /** Optional block rendered under the header (e.g. the capability strip). */
   chips?: ReactNode;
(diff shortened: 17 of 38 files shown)


README.md
```diff
@@ -208,6 +208,18 @@ Colors are plain CSS custom properties split into two layers:
   (slate neutrals + a blue accent)**, not a specific product brand. This is the
   **swappable layer**.
 
+> **Deprecated names (removed 2026-07, #513):** `--color-brand-amber` /
+> `--color-brand-amber-light` were a misnamed duplicate of
+> `--color-accent-primary` / `--color-accent-primary-hover` (both held the
+> same blue) — every consumer now points at the `accent-primary` pair, which
```
s-annam: **Blocking — this claim is false in light mode.**

`(both held the same blue)` holds for `brand-amber` / `accent-primary`, but not for the hover pair. From `origin/main:tokens.css`:

| token | light | dark |
|---|---|---|
| `--color-brand-amber` | `#2563eb` | `#60a5fa` |
| `--color-accent-primary` | `#2563eb` | `#60a5fa` |
| `--color-brand-amber-light` | `#3b82f6` | `#93c5fd` |
| `--color-accent-primary-hover` | **`#1d4ed8`** | `#93c5fd` |

Only the dark block matched. In light, `brand-amber-light` was a *lighter* blue and `accent-primary-hover` is a *darker* one, so folding them flips hover direction for both surviving consumers — `Button.tsx` `variant="primary"` (every primary CTA in the app) and the ConsentDialog licence link.

That flip is intentional and the commit message says so. The problem is that this paragraph tells a downstream reskinner the two names were interchangeable, when following it would silently change their light-mode hover.

Suggested replacement for the first sentence:

> `--color-brand-amber` was an exact duplicate of `--color-accent-primary` (same blue in both themes). `--color-brand-amber-light` was a *near*-duplicate of `--color-accent-primary-hover`: identical in dark (`#93c5fd`), but in light it held `#3b82f6` against the hover token's `#1d4ed8`. Folding them deliberately corrects the light-mode hover on `Button variant="primary"` and the ConsentDialog licence link from a lighter blue to a darker one.

src/design-system/shared/Tabs.tsx
```diff
@@ -134,21 +141,36 @@ interface TabProps {
...
+  // own border (the pre-#516 illegibility bug).
+  //
+  // Hover on an INACTIVE tab must not borrow the selected tab's surface:
+  // `hover:bg-surface-card` made a merely-pointed-at tab look selected, with
+  // only font-weight left to tell them apart. It hovers to an outline instead
+  // of a fill — `bg-surface-hover` alone cannot carry it, because
+  // `--color-bg-hover` and `--color-bg-subtle` (the track) are the SAME value
+  // in dark (#334155 → 1.00:1). The inset `ring-border-strong` clears the
+  // track in both themes (2.34:1 light, 2.18:1 dark) and stays visually
+  // distinct from selection, which is a filled surface with no ring.
```
s-annam: **Secondary — this docblock and `CountBadge`'s contradict each other on the same two numbers.**

Here: the ring `clears the track in both themes (2.34:1 light, 2.18:1 dark)`.

`CountBadge.tsx:16`, added in the same commit: `border-strong` reaches only `2.34:1 light / 2.18:1 dark — both fail`.

I recomputed independently and get ~2.37 light / 2.18 dark, so the numbers are right and both statements are defensible against *different* thresholds — "clears" here means "is perceptible against the track", while `CountBadge` means "clears WCAG 1.4.11's 3:1". But a reader hitting these two files in either order will conclude one of them is wrong.

Worth noting the ring is weaker than the quoted figure suggests: against its own hovered fill (`bg-surface-hover` `#e2e8f0`) rather than the resting track, it sits at ~2.1:1 in light. Defensible for a hover-only affordance, which is not an enumerated 1.4.11 state — but say so rather than leaving "clears" to carry it.

Suggested: replace `clears the track in both themes` with wording that names the threshold it does *not* meet, e.g. "reads at 2.34:1 light / 2.18:1 dark against the track — enough to be perceptible for a hover-only affordance, but below 1.4.11's 3:1, so do not reuse this token for a persistent boundary (see `CountBadge.tsx`).".

s-annam (commented):
> **Intended verdict: `REQUEST_CHANGES`** (1 blocking finding). Posted as a comment because GitHub rejects a change request from the PR author (`Review Can not request changes on your own pull request`). Treat the blocking item below as unresolved — this review must not be read as an approval, and the PR still needs one from someone else.

Self-review of the batch before it goes to a human reviewer. Verified every finding against the files at `273bbac`; three findings from the pass did not survive verification and were dropped rather than posted.

**One blocking finding, and it is a documentation defect, not a code defect.** The code in this diff verified clean: zero surviving references to any deleted token across all file types, all 10 `StatusBadge` call sites safe for `w-fit`, `PageShell.chips` single-caller confirmed, `tsc`/`lint`/tab tests green, and CI green on `verify`, `fallow`, and Cloudflare Pages.

## Blocking

1. **`README.md:213` — the deprecation note misstates what happened.** See inline. It calls the `brand-amber-light` → `accent-primary-hover` fold a duplicate collapse; in light mode the two tokens held *different* blues and the fold deliberately changes hover direction on every primary button. The change itself is correct and is disclosed in the commit message ("fix the backwards hover pair") — only this sentence is wrong, and it is the sentence a downstream reskinner reads to decide whether the rename is safe to follow.

## Secondary

2. **`Tabs.tsx:156` vs `CountBadge.tsx:16` — adjacent docblocks give the same two ratios opposite verdicts.** See inline.

## Non-blocking — already tracked

- `SourceDiagnosticsPanel`'s `SegmentButton` is a near-clone of `Tab` that received none of #516's hover work. Filed as #527 (scoped as a dedupe into `Tabs`, which also recovers the tab roles and arrow-key navigation the hand-rolled control never had).
- Tabs' hover `ring-inset` leaks into the focus ring, so a simultaneously hovered + keyboard-focused tab draws its focus ring inside the border box. Verified real — `BASE` never resets `--tw-ring-inset`. Already disclosed in the PR body as an accepted nit; shape change only, ring stays visible.
- The `.gitignore` entry is unrelated to #513–#518 and rides this branch by an explicit decision, not by accident.

## Gates

| Gate | Result |
|---|---|
| Fixture PII | **Skipped — no fixture binaries in this diff.** Not a pass. |
| Design-system tier / reuse | Pass — no raw `<button>`; new surfaces composed from `@design-system` |
| Style tokens | Pass — no hex, no raw palette, no manual `dark:` in feature code |
| Dead code (`fallow`) | Pass in CI |
| Script/config-as-code | Pass — `PALETTE_RE` keys off numbered scales, never token names, so the rename could not have weakened lint enforcement; `.gitignore` pattern correctly root-anchored and the private skill is untracked |

## Not merge-relevant

Two findings were raised and rejected on verification: the `CountBadge` docblock's "#516 gave the Tab track" is a mild overstatement (the segmented control already had that track on `main`) but harmless, and the `TrustIcons` barrel export having no cross-seam consumer is the deliberate `@design-system` convention, since feature code must never deep-import.

---

README.md
s-annam: Fixed in `195308a`.

Dropped the false `(both held the same blue)` clause and replaced it with your suggested wording, near-verbatim: `brand-amber` was an exact duplicate, `brand-amber-light` only a *near*-duplicate (identical `#93c5fd` in dark, `#3b82f6` vs `#1d4ed8` in light), and the fold deliberately corrects the light-mode hover on `Button variant="primary"` and the ConsentDialog licence link from a lighter blue to a darker one.

You were right that the flip itself was already disclosed (commit message + PR body) — the defect was this paragraph telling a downstream reskinner the two names were interchangeable, which is the one place the flip would bite silently. That is now stated explicitly rather than implied.

src/design-system/shared/Tabs.tsx
s-annam: Fixed in `195308a`.

Replaced `clears the track in both themes` with wording that names the threshold it does *not* meet:

> The inset `ring-border-strong` reads at 2.34:1 light / 2.18:1 dark against the track — enough to be perceptible for a hover-only affordance (not an enumerated 1.4.11 state), but BELOW that criterion's 3:1, so do not reuse this token for a persistent boundary (see `CountBadge.tsx`, which needs `content-muted` for exactly that reason).

That resolves the contradiction in the direction you identified: both docblocks were true against *different* thresholds, and only `CountBadge` said which one. Now both name 1.4.11's 3:1 explicitly, and the cross-reference makes the asymmetry (hover-only affordance vs persistent boundary) the reason they differ rather than something a reader has to reconstruct.

s-annam:
Pushed `195308a` (force-push: the branch stays at one commit so the merge queue derives the squash message from it verbatim).

**Both review findings fixed** — replies on each thread, both resolved.

**Additional copy changes in the same commit**, beyond the review:

- **Hero** (`App.tsx`) — leads with the three lanes and "nothing to install, no account" instead of "see what a recruiter's screening software sees". The Greenhouse citation below still carries the screening-software framing, which is where it belongs: it is cited evidence, not a claim about our feature set.
- **CapabilityStrip lanes** — "What a parser actually pulls out of your PDF" → "A parser in your browser pulls out your experience, skills and contact details"; "Edit inline, export a clean ATS-safe PDF" → "Edit inline, save it in your browser, or export a clean PDF".
- **Privacy rail** — "Runs on your device by default" → **"Your resume stays on your device"**.

The rail change is the one worth a second look, because it moves what the claim is *about*.

"Runs on your device" (unqualified) would be false: the job-search lane advertised on this same card `fetch()`es three third-party feeds (`jobicy`, `remotive`, `arbeitnow`), and any build with `VITE_POSTHOG_KEY` set ships analytics the user never opted into. So a where-the-code-runs claim needs a hedge — and every friendly hedge tried ("unless you choose otherwise", "you decide what leaves") overclaims on the analytics path, which is not the user's choice.

A custody claim about the *resume* needs no hedge and is stronger: `job-search/providers/keywords.ts` guarantees the outbound payload is a short keyword string, never resume text.

I also corrected the docblock's privacy rationale, which cited **BYOK as a live cloud-egress exception — there is no BYOK provider in the tree.** The hedge was partly guarding an unbuilt feature. The docblock now names the real exceptions and pins the `keywords.ts` invariant the copy depends on, so a future change there surfaces this string.

Gates: `npm run verify` green (typecheck, eslint, fixture PII 51/51 synthetic, 2509 tests, build). `fallow` exits non-zero but is report-only inside `verify`; its 5 complexity findings are all pre-existing.

Still open, not addressed here: `App.tsx:129` reads "Your whole job search — on your device by default", and its comment justifies the hedge partly on the same stale BYOK premise. Happy to align it in a follow-up push.

s-annam:
image
image

src/App.tsx
```diff
@@ -115,30 +114,29 @@ export default function App() {
             // The trust stat is the one supporting line; everything else (the
             // recruiter-agent context) moves to the quiet block below the drop
             // zone so the hero isn't three competing messages.
+            //
+            // The headline's "by default" is NOT hedging for its own sake and
+            // must not be dropped: the job-search lane egresses keywords on an
+            // explicit click (see FindJobsPanel), the BYOK provider path is
```
rohithgollapalli: Nit (non-blocking): present tense "the BYOK provider path **is** real cloud egress" reads as if BYOK ships today, but it's unbuilt (`#320` — `job-search/providers/index.ts`, `types.ts`, `FindJobsPanel.tsx` all say "will append"). `CapabilityStrip.tsx`'s docblock states the opposite ("none exists today"). Reconcile to future tense.

rohithgollapalli (dismissed):
Token rename lands clean — `brand-*` → `accent-primary` across 30+ files with no orphaned tokens, the build regenerates every utility (incl. the awkward-but-valid `accent-accent-primary`), and the README documents the one intentional visual change (light-mode primary/link hover darkens from `#3b82f6` to `#1d4ed8`). New `CapabilityStrip` + `TrustIcons` sit in `design-system/`, compose existing primitives, and add no dead exports. Local gates green (typecheck / lint / build / 202 component tests), CI `verify`/`fallow`/Pages green.

## Nits
- `src/App.tsx:120` — comment states "the BYOK provider path **is** real cloud egress," present tense, but BYOK is unbuilt (`#320`; `job-search/providers/index.ts`, `types.ts`, `FindJobsPanel.tsx` all describe it as future). `CapabilityStrip.tsx`'s own docblock says the opposite ("none exists today — there is no BYOK provider in the tree"). Reconcile to future tense so the two new comments agree.

---

src/App.tsx
s-annam: Fixed in `e36ac32`.

Reconciled to future tense and cross-referenced the docblock that already had it right:

```
// explicit click (see FindJobsPanel), a build with
// VITE_POSTHOG_KEY set (the hosted one) ships analytics, and the
// BYOK provider path (#320, not in the tree yet — see
// CapabilityStrip) will be real cloud egress once it lands. The
// first two are true today and already load-bearing on their own.
```

I also reordered the three justifications so the two that are true **today** — job-search keyword egress on explicit click, and `VITE_POSTHOG_KEY` analytics — carry the "by default" hedge on their own, with a line saying so explicitly. The point of your nit wasn't only the tense: the hedge's rationale was partly resting on unbuilt code, so if `#320` had been cancelled the comment would have argued for its own removal. It no longer depends on `#320` landing.

Two places outside the thread carried the identical claim, both now fixed in the same push:

- the **commit message** body ("the BYOK path is real cloud egress") — otherwise the fix lands in the code and the falsehood ships in `main`'s history permanently
- the **PR body**, in two spots (the summary bullet and the review-log entry)

Gates green on the amended commit: `typecheck` clean, `test` 2509 passed / 9 skipped.

rohithgollapalli (approved):
Re-review of `e36ac32`. Only delta since my prior approval (`195308a`) is the `App.tsx` comment nit — now fixed: BYOK is scoped as future ("#320, not in the tree yet … will be real cloud egress once it lands"), consistent with `CapabilityStrip`'s docblock. No functional change; typecheck / lint / build / tests and CI `verify`/`fallow`/Pages all green. Nothing else outstanding.

---

Issue #534, opened 2026-07-21
job-search v2 — role→title keyword taxonomy + board title-filter & size caps


s-annam:
Part of epic #528 (job-search v2 — company-targeted ATS-board search). **New slice — the board-size + keyword-filtering layer.** Independent of the adapters (#529/#530) and the sector classifier (#531); **it is the gate before the wiring (#533)** — the main search must not connect to company boards until this exists.

## Why this slice exists

Company ATS boards have **no server-side free-text search** — every search pulls the *whole* board (see #528 / #529). Two problems follow, and this slice owns both:

1. **Board size.** A large company board is 100s–1000s of roles. #529/#530 already fetch the light index (no descriptions) to bound payload, but we still must **narrow to relevant roles and cap the count** before hydrating descriptions and ranking, or the browser drowns in irrelevant postings.
2. **The filter signal is a role, not skills.** The natural instinct — filter board titles by the resume's *skills* (`React`, `TypeScript`, `Postgres`) — fails: skills almost never appear in a job title (`Senior Frontend Engineer`). The right filter is a **role/function keyword set** (`frontend`, `full stack`, `ui engineer`, `web`) derived from the candidate's target **role/title**, mapped through a **curated function→title-synonym taxonomy**. That taxonomy is hand-curated work — this slice is where it lives.

Both are **client-side, zero-egress**. This keyword set is a local title filter and is a **different privacy class** from `keywords.ts` (the audited egress string sent to the keyless feeds) — it never leaves the browser. Do not route it through `keywords.ts`, and do not send it to any ATS board.

## Data model

New module `src/lib/job-search/role-keywords.ts` (pure lib, zero-dep, mirrors `src/lib/score/score.ts` — typed, named-constant table, no I/O):

```ts
// A curated function → title-synonym map. ~12–20 job families, each a set of
// lowercased substrings that appear in real job titles for that family.
// e.g. "frontend" → ["frontend", "front end", "front-end", "ui engineer", "web engineer", "react"]
//      "data"     → ["data engineer", "data scientist", "ml engineer", "machine learning", "analytics engineer"]
export const ROLE_KEYWORDS: Readonly<Record<RoleFamily, readonly string[]>>;

export interface RoleFilter {
  families: RoleFamily[];   // the candidate's inferred role families (usually 1–2)
  keywords: string[];       // flattened, deduped substrings to match titles against
  source: "heuristic";      // reserved for a future semantic upgrade, like #531
}

[...]

Issue #533, opened 2026-07-21
job-search v2 — wire registry → getProviders(): sector-suggested companies, capped fan-out, IndexedDB cache


s-annam:
Part of epic #528 (job-search v2 — company-targeted ATS-board search). **Slice 5 of 5 — the integration.** Depends on the adapters (#529 Greenhouse light-index/hydrate, #530 Lever+Ashby), the sector classifier (#531), the registry (#532), **and the role-keyword/board-size layer (#534)**. This is the slice that connects company boards to the live main search — and it stays **last**: do not wire until #534 exists, so the whole-board flood is filtered and capped before it ever reaches the UI.

## Goal

Wire it together: classify the candidate's sector (#531) → pick registry companies for that sector (#532) → build their ATS adapters (#529/#530) → **light-index fetch** each board → **role-title filter + per-company cap (#534)** → **lazily hydrate descriptions for the surviving top matches only** → rank by fit → cache in IndexedDB → surface suggested companies in the UI with add/remove.

## The flow (bounded end-to-end)

```
classifySector(parsed)            (#531) → sector
companiesForSector(sector, CAP)   (#532) → CompanyEntry[]
makeCompanyProvider(entry)        (#529/#530) → JobProvider (light index, description="")
  fan out, concurrency-capped
filterPostingsByRole(postings, roleFilter)  (#534) → relevant titles only
capPerCompany(postings, N)        (#534) → bounded set per company
hydrate<Ats>(slug, id)            (#529 for Greenhouse; Lever/Ashby already carry descriptionPlain)
rankPostings(...)                 (existing) → ranked
IndexedDB cache keyed by (ats, slug)
```

The key discipline: **descriptions are hydrated only for postings that survive the role-filter and per-company cap** — never for the whole board. Greenhouse needs an explicit `hydrateGreenhouse` call per matched job; Lever/Ashby already return `descriptionPlain`, so their hydrate is a no-op. This is what keeps a 1000-role board cheap.

## Background — the seam

`getProviders()` in `src/lib/job-search/providers/index.ts` decides who participates in the fan-out, and `searchJobs` in `src/lib/job-search/search.ts` calls it with **no arguments** today. This slice makes provider resolution **sector-driven and registry-derived**, so `getProviders()` (or a new sibling the orchestrator calls) becomes **parameterized by the parsed resume + a selected company set**. The keyless feeds (`KEYLESS_PROVIDERS`) stay always-on and unchanged.

[...]

Issue #532, opened 2026-07-21
job-search v2 — curated sector-tagged company→ATS→slug registry


s-annam:
Part of epic #528 (job-search v2 — company-targeted ATS-board search). **Slice 4 of 5.** Depends on slice 3 for the `Sector` taxonomy (registry entries are tagged with it). Consumed by slice 5. Adapter factories (slices 1–2) are the runtime that turns a registry entry into a live provider.

## Goal

Ship a **curated `company → ATS → slug` registry**, tagged by the sector taxonomy from slice 3, so slice 5 can select sector-appropriate companies and build their ATS adapters. Lazy-loaded like the cascade tiers so it never bloats the entry chunk.

## Data model

New data module `src/lib/job-search/company-registry.ts` (the JSON payload can live inline in the module or as an adjacent `company-registry.json` imported by it — pick one; a typed `.ts` const is simplest and keeps the `Sector` type checked at build time):

```ts
import type { Sector } from "./sector.ts";

export type Ats = "greenhouse" | "lever" | "ashby";

export interface CompanyEntry {
  name: string;      // display name, e.g. "Stripe" — used for the provider label/source
  ats: Ats;          // which adapter factory builds this entry's provider
  slug: string;      // the ATS board slug, e.g. "stripe"
  sectors: Sector[]; // one or more sectors this company belongs to
}

export const COMPANY_REGISTRY: readonly CompanyEntry[] = [ /* … */ ];

// Helper consumed by slice 5.
export function companiesForSector(sector: Sector, limit: number): CompanyEntry[];
```

- `sectors` is an array — a company can be both `fintech` and `enterprise-saas`.
- `ats` maps 1:1 to the slice-1/2 factories (`greenhouse → makeGreenhouseProvider`, etc.).
- `companiesForSector` returns up to `limit` entries for a sector (deterministic order — the fan-out cap in slice 5 relies on this being bounded and stable).

## Seeding the registry

Target ~150–250 companies across the taxonomy for the first cut — enough that every sector returns a usable set, not a census. Sourcing (all offline, one-time, no runtime dependency):

- Greenhouse / Lever / Ashby each publish public board directories → collect real slugs, verify each resolves to a live board (reuse the CORS/shape check from slices 1–2 — a slug that 404s or fails CORS does not get added).
- Tag each with 1–2 sectors from slice 3's taxonomy (YC industry tags / Levels.fyi / hand-curation for the well-known names).
[...]

Issue #531, opened 2026-07-21
job-search v2 — sector classifier (heuristic + WebLLM semantic upgrade)


s-annam:
Part of epic #528 (job-search v2 — company-targeted ATS-board search). **Slice 3 of 5.** Independent of slices 1–2 (pure classification lib); consumed by slice 5 to pick which registry companies to search.

## Goal

Classify a parsed resume into a **sector** (a small fixed taxonomy), so slice 5 can pick sector-appropriate companies to search instead of a static global list. Heuristic by default (keyword table over parsed skills/titles), with a **WebLLM semantic upgrade when WebGPU is available** — mirroring the semantic-with-keyword-fallback pattern jd-match already uses (`src/lib/jd-match/`, and the capability/platform gating in `src/lib/webllm/capability.ts` / `platform.ts`).

## Sector taxonomy

A fixed, small enum (~12–16) shipped as a named constant — chosen to partition the company registry (slice 4), not to be exhaustive. Starting set (finalize in review):

`fintech`, `devtools`, `data-ml`, `healthtech`, `ecommerce`, `gaming`, `security`, `enterprise-saas`, `consumer-social`, `hardware-iot`, `crypto-web3`, `edtech`, `logistics-mobility`, `media-adtech`, `government-defense`, `other`.

`other` is the always-valid fallback so classification never fails closed.

## Input

The parsed resume already carries the signal. Use the heuristic parse the job-search lane already receives (`HeuristicParsedResume` — the same value `searchJobs` takes). Relevant fields: `skills` / `skills_explicit` / `skills_inferred`, and `experience[].title` + `experience[].company`. Do not read raw PDF text — stay on the parsed model like the rest of the lane.

## Implementation

New module `src/lib/job-search/sector.ts` (pure lib, zero-dep, mirrors the shape of `src/lib/score/score.ts` — typed, named-constant weights, no I/O):

```ts
export type Sector = "fintech" | "devtools" | /* … */ | "other";

export interface SectorGuess {
  sector: Sector;
  confidence: number;        // 0..1
  source: "heuristic" | "semantic";
  runnerUp?: Sector;         // for a "not fintech? try devtools" UI affordance in slice 5
}

// Heuristic: weighted keyword table Sector → RegExp[] over a lowercased
// haystack of skills + titles (+ lightly, company names). Returns the
// top-scoring sector, "other" when nothing clears a floor.
export function classifySectorHeuristic(parsed: HeuristicParsedResume): SectorGuess;

[...]

PR #535, opened 2026-07-21
feat(job-search): company-targeted ATS-board search foundation (#528)

s-annam:
## Summary

Build the client-side foundation for **company-targeted ATS-board search** (epic #528, the follow-on to the closed job-search epic #317): keyless per-company board adapters (Greenhouse/Lever/Ashby), a sector classifier, a curated sector→company→slug registry, and a role-title filter/cap layer.

Everything here is built and unit-tested but **deliberately not wired into the live `getProviders()` seam** — that integration is **#533**, held until this keyword/cap layer exists so a whole-board fetch is filtered and bounded before it ever reaches the UI (the deliberate build-everything-then-connect order from #528).

**Privacy:** all egress stays the public company slug only; no resume-derived data leaves the browser. `keywords.ts` remains the sole resume-derived egress helper, unused by these adapters and the role filter.

Refs #528

## What's in this PR (5 slices)

| Slice | Module | What |
|---|---|---|
| #529 | `providers/greenhouse.ts` (+test) | `makeGreenhouseProvider(slug, companyName?)` — light-index fetch (no `?content=true`), lazy `hydrateGreenhouse` per surviving job |
| #529 | `types.ts` | add optional `JobPosting.departments?` — the shared field the role filter reads |
| #530 | `providers/lever.ts`, `providers/ashby.ts` (+tests) | same `JobProvider` contract; Lever top-level array `?mode=json&limit=100`, Ashby `{jobs:[]}` `?includeCompensation=false`; both prefer `descriptionPlain` |
| #531 | `sector.ts` (+test) | heuristic sector classifier over parsed titles/skills, WebGPU-gated WebLLM semantic upgrade; 16-family taxonomy |
| #532 | `company-registry.ts` (+test) | curated sector-tagged company→ATS→slug map (162 entries), `companiesForSector(sector, limit)`; slugs **not yet CORS-verified** (flagged in docblock) |
| #534 | `role-keywords.ts` (+test) | curated role→title-synonym taxonomy, `roleFilterForResume`, `filterPostingsByRole`, `capPerCompany`; titles-not-skills, never-fail-closed, zero-egress |

## Review round 2 (@Samhit21) — four fixes, all in code

Each only bites once a wrong or thin input reaches the layer, which is exactly the path #532's unverified slugs make ordinary:

- **id-less postings collided.** All three adapters used a bare `job.id ?? ""`, dropping the url fallback `remotive.ts`/`jobicy.ts` added for this reason. Two id-less postings both keyed `greenhouse:acme:` → duplicate React keys and a silent dedup collapse. Now `job.id ?? job.absolute_url` / `hostedUrl` / `jobUrl`, with a test per adapter.
- **Lever's `(data ?? [])` didn't guard a non-array.** Lever is the one adapter whose success shape is a bare top-level array, so a 200 with an object error envelope landed on `.map`. Now `Array.isArray`; the test that claimed to cover this mocked `null` (which `??` already handles) and is split into a `null` case and a real `{}` case.
- **Sector confidence was inverted.** Share-of-top-two alone gave a lone keyword hit `confidence 1.0` (every sector is scored, so second is 0), while a genuine 3-vs-2 match read 0.6. Confidence is now `share × min(1, top/3)` — monotone in absolute evidence. The generic single tokens that cross-matched other families' phrases (`routing`, `warehouse`, `streaming`, `marketplace`) become phrase patterns, so "data warehouse" no longer scores logistics and "Kafka streaming" no longer scores media.
- **Role filter kept an arbitrarily-far-behind runner-up.** `score > 0` was the only gate, so one early-career "UX Designer" rode alongside eight backend titles and design's broad keywords ("designer", "user experience") kept every Designer posting on every board. The runner-up now needs at least half the winner's score; an 8-vs-1 test covers the lopsided split the existing 2-1-1 case never exercised.

## Test plan

- [x] `npx vitest run src/lib/job-search` — 113/113 green (7 new)
- [x] `npm run test` — 2592 passed | 9 skipped
- [x] `npm run lint` clean
- [x] `npm run verify` — full CI mirror runs green on this PR
- [x] **CORS verification (human/dev, before #533 ships):** confirm each vendor endpoint returns a permissive `access-control-allow-origin` from a real browser origin against ≥2 real slugs; drop any board that fails. A headless run cannot prove browser CORS — this stays an explicit unchecked task carried into #533.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #529 | Claude Sonnet 5 | high |
| Implementation — #530 | Claude Sonnet 5 | high |
| Implementation — #531 | Claude Sonnet 5 | high |
| Implementation — #532 | Claude Sonnet 5 | medium |
| Implementation — #534 | Claude Opus 4.8 (1M context) | ultra |
| Adversarial review | Claude Sonnet 5 | high |
| Review fixes — adversarial round 1 | (none — clean, zero blocking) | — |
| Review revisions — human round 2 | Claude Opus 4.8 (1M context) | medium |
| Orchestration + PR | Claude Opus 4.8 (1M context) | high |

## Adversarial review

Round 1 — **clean** (`clean: true`), zero blocking findings. Converged in one round; no fix subagent needed. Reviewer: Claude Sonnet 5 (high).

Four nits, all investigated, all no-action:

- **Adapter `mapJob`/post-filter duplication** across greenhouse/lever/ashby (fallow `dup:93eac9e2` ×3) — established house pattern (existing `remotive`/`arbeitnow`/`jobicy` adapters already use one-mapJob-per-vendor), per-vendor divergence is deliberate, not a shared-helper candidate.
- **`sector.test.ts` clone groups** — standard per-case mock/fixture boilerplate; not a divergence bug.
- **Static `htmlToPlaintext` import** in the adapters — heads-up for #533 lazy-load wiring only; correct as-is for a pure lib module.
- **Lever/Ashby "slice 2 of 5" docblock label** — cosmetic, and actually correct.

### DIFF
--- a/src/lib/job-search/company-registry.ts
+++ b/src/lib/job-search/company-registry.ts
@@ -1,0 +1,256 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * company-registry.ts — curated `company → ATS → slug` data (#532, slice 4
+ * of the job-search-v2 epic #528).
+ *
+ * Sector tags reuse `Sector` from `./sector.ts` verbatim — this module does
+ * NOT fork a second taxonomy. Slice 5 (#533) dynamic-imports this module
+ * (same cascade-tier lazy-load discipline as `src/lib/heuristics/`) so the
+ * payload never enters the entry chunk, and turns a `companiesForSector()`
+ * result into live providers via the slice-1/2 adapter factories
+ * (`makeGreenhouseProvider` / `makeLeverProvider` / `makeAshbyProvider`).
+ *
+ * Sourcing: hand-curated from each ATS vendor's public board directory and
+ * general familiarity with well-known companies' public careers pages, one
+ * company per entry, tagged with 1-2 sectors from `./sector.ts`. Curated
+ * 2026-07-21.
+ *
+ * IMPORTANT — slugs are NOT CORS-verified. Unlike the slice-1/2 adapters
+ * (whose docblocks note "unverified from a browser origin" pending a live
+ * check), these slugs have not even been curl-checked for existence; they
+ * are a best-effort curated list. #533 (the live wiring slice) is expected
+ * to browser-verify each board it actually queries and silently drop any
+ * slug that 404s or fails CORS — a dead entry here costs nothing worse than
+ * an empty search result. Treat this list as "plausible, needs-verification"
+ * data, not a source of truth for "company X uses ATS Y". A periodic
+ * re-verify/refresh pass is future work, not this slice.
+ */
+
+import type { Sector } from "./sector.ts";
+
+/** ATS vendor identifier — maps 1:1 to the slice-1/2 adapter factories. */
+export type Ats = "greenhouse" | "lever" | "ashby";
+
+export interface CompanyEntry {
+  /** Display name, e.g. "Stripe" — used for the provider label/source. */
+  name: string;
+  /** Which adapter factory builds this entry's provider. */
+  ats: Ats;
+  /** The ATS board slug, e.g. "stripe". */
+  slug: string;
+  /** One or more sectors this company belongs to (from `./sector.ts`). */
+  sectors: Sector[];
+}
+
+/**
+ * The curated registry. Order is deliberate and stable — `companiesForSector`
+ * relies on declaration order for its deterministic result order.
+ */
+export const COMPANY_REGISTRY: readonly CompanyEntry[] = [
+  // -- fintech ------------------------------------------------------------
+  { name: "Stripe", ats: "greenhouse", slug: "stripe", sectors: ["fintech"] },
+  { name: "Plaid", ats: "greenhouse", slug: "plaid", sectors: ["fintech"] },
+  { name: "Robinhood", ats: "greenhouse", slug: "robinhood", sectors: ["fintech"] },
+  { name: "Affirm", ats: "greenhouse", slug: "affirm", sectors: ["fintech"] },
+  { name: "Chime", ats: "greenhouse", slug: "chime", sectors: ["fintech"] },
+  { name: "Brex", ats: "greenhouse", slug: "brex", sectors: ["fintech"] },
+  { name: "Ramp", ats: "ashby", slug: "ramp", sectors: ["fintech"] },
+  { name: "Mercury", ats: "ashby", slug: "mercury", sectors: ["fintech"] },
+  { name: "Coinbase", ats: "greenhouse", slug: "coinbase", sectors: ["fintech", "crypto-web3"] },
+  { name: "Marqeta", ats: "greenhouse", slug: "marqeta", sectors: ["fintech"] },
+  { name: "Wealthfront", ats: "greenhouse", slug: "wealthfront", sectors: ["fintech"] },
+  { name: "Klarna", ats: "lever", slug: "klarna", sectors: ["fintech"] },
+  { name: "Gusto", ats: "greenhouse", slug: "gusto", sectors: ["fintech", "enterprise-saas"] },
+  { name: "Carta", ats: "greenhouse", slug: "carta", sectors: ["fintech"] },
+
+  // -- devtools -------------------------------------------------------------
+  { name: "GitLab", ats: "greenhouse", slug: "gitlab", sectors: ["devtools"] },
+  { name: "GitHub", ats: "greenhouse", slug: "github", sectors: ["devtools"] },
+  { name: "HashiCorp", ats: "greenhouse", slug: "hashicorp", sectors: ["devtools"] },
+  { name: "Docker", ats: "greenhouse", slug: "docker", sectors: ["devtools"] },
+  { name: "Postman", ats: "lever", slug: "postman", sectors: ["devtools"] },
+  { name: "Vercel", ats: "ashby", slug: "vercel", sectors: ["devtools"] },
+  { name: "Netlify", ats: "lever", slug: "netlify", sectors: ["devtools"] },
+  { name: "CircleCI", ats: "greenhouse", slug: "circleci", sectors: ["devtools"] },
+  { name: "JFrog", ats: "greenhouse", slug: "jfrog", sectors: ["devtools"] },
+  { name: "Sentry", ats: "greenhouse", slug: "sentry", sectors: ["devtools"] },
+  { name: "LaunchDarkly", ats: "greenhouse", slug: "launchdarkly", sectors: ["devtools"] },
+  { name: "Replit", ats: "ashby", slug: "replit", sectors: ["devtools"] },
+  { name: "Warp", ats: "ashby", slug: "warp", sectors: ["devtools"] },
+
+  // -- data-ml --------------------------------------------------------------
+  { name: "Databricks", ats: "greenhouse", slug: "databricks", sectors: ["data-ml"] },
+  { name: "Snowflake", ats: "greenhouse", slug: "snowflake", sectors: ["data-ml"] },
+  { name: "Scale AI", ats: "ashby", slug: "scaleai", sectors: ["data-ml"] },
+  { name: "Weights & Biases", ats: "greenhouse", slug: "wandb", sectors: ["data-ml"] },
+  { name: "Hugging Face", ats: "lever", slug: "huggingface", sectors: ["data-ml"] },
+  { name: "DataRobot", ats: "greenhouse", slug: "datarobot", sectors: ["data-ml"] },
+  { name: "Fivetran", ats: "greenhouse", slug: "fivetran", sectors: ["data-ml"] },
+  { name: "dbt Labs", ats: "greenhouse", slug: "dbtlabs", sectors: ["data-ml"] },
+  { name: "Anthropic", ats: "greenhouse", slug: "anthropic", sectors: ["data-ml"] },
+  { name: "Cohere", ats: "ashby", slug: "cohere", sectors: ["data-ml"] },
+  { name: "Pinecone", ats: "ashby", slug: "pinecone", sectors: ["data-ml"] },
+  { name: "Modal", ats: "ashby", slug: "modal", sectors: ["data-ml"] },
+
+  // -- healthtech -------------------------------------------------------------
+  { name: "Oscar Health", ats: "greenhouse", slug: "oscarhealth", sectors: ["healthtech"] },
+  { name: "Ro", ats: "greenhouse", slug: "ro", sectors: ["healthtech"] },
+  { name: "Cedar", ats: "greenhouse", slug: "cedar", sectors: ["healthtech"] },
+  { name: "Tempus", ats: "greenhouse", slug: "tempus", sectors: ["healthtech"] },
+  { name: "Zocdoc", ats: "greenhouse", slug: "zocdoc", sectors: ["healthtech"] },
+  { name: "Included Health", ats: "greenhouse", slug: "includedhealth", sectors: ["healthtech"] },
+  { name: "Komodo Health", ats: "greenhouse", slug: "komodohealth", sectors: ["healthtech"] },
+  { name: "Clover Health", ats: "greenhouse", slug: "cloverhealth", sectors: ["healthtech"] },
+  { name: "Benchling", ats: "greenhouse", slug: "benchling", sectors: ["healthtech", "data-ml"] },
+  { name: "Devoted Health", ats: "greenhouse", slug: "devotedhealth", sectors: ["healthtech"] },
+  { name: "Nuna", ats: "lever", slug: "nuna", sectors: ["healthtech"] },
+  { name: "Maven Clinic", ats: "greenhouse", slug: "mavenclinic", sectors: ["healthtech"] },
+
+  // -- ecommerce --------------------------------------------------------------
+  { name: "Shopify", ats: "lever", slug: "shopify", sectors: ["ecommerce"] },
+  { name: "Wayfair", ats: "greenhouse", slug: "wayfair", sectors: ["ecommerce"] },
+  { name: "Instacart", ats: "greenhouse", slug: "instacart", sectors: ["ecommerce"] },
+  { name: "Faire", ats: "greenhouse", slug: "faire", sectors: ["ecommerce"] },
+  { name: "Whatnot", ats: "ashby", slug: "whatnot", sectors: ["ecommerce", "consumer-social"] },
+  { name: "Allbirds", ats: "greenhouse", slug: "allbirds", sectors: ["ecommerce"] },
+  { name: "Warby Parker", ats: "greenhouse", slug: "warbyparker", sectors: ["ecommerce"] },
+  { name: "Glossier", ats: "lever", slug: "glossier", sectors: ["ecommerce"] },
+  { name: "Stitch Fix", ats: "greenhouse", slug: "stitchfix", sectors: ["ecommerce"] },
+  { name: "Chewy", ats: "greenhouse", slug: "chewy", sectors: ["ecommerce"] },
+  { name: "Poshmark", ats: "greenhouse", slug: "poshmark", sectors: ["ecommerce"] },
+
+  // -- gaming ---------------------------------------------------------------
+  { name: "Discord", ats: "greenhouse", slug: "discord", sectors: ["gaming", "consumer-social"] },
+  { name: "Roblox", ats: "greenhouse", slug: "roblox", sectors: ["gaming"] },
+  { name: "Riot Games", ats: "greenhouse", slug: "riotgames", sectors: ["gaming"] },
+  { name: "Unity", ats: "greenhouse", slug: "unity", sectors: ["gaming"] },
+  { name: "Niantic", ats: "greenhouse", slug: "niantic", sectors: ["gaming"] },
+  { name: "Twitch", ats: "greenhouse", slug: "twitch", sectors: ["gaming", "media-adtech"] },
+  { name: "Epic Games", ats: "greenhouse", slug: "epicgames", sectors: ["gaming"] },
+  { name: "Scopely", ats: "greenhouse", slug: "scopely", sectors: ["gaming"] },
+  { name: "Zynga", ats: "greenhouse", slug: "zynga", sectors: ["gaming"] },
+  { name: "Rec Room", ats: "ashby", slug: "recroom", sectors: ["gaming"] },
+
+  // -- security ---------------------------------------------------------------
+  { name: "CrowdStrike", ats: "greenhouse", slug: "crowdstrike", sectors: ["security"] },
+  { name: "Okta", ats: "greenhouse", slug: "okta", sectors: ["security"] },
+  { name: "SentinelOne", ats: "greenhouse", slug: "sentinelone", sectors: ["security"] },
+  { name: "1Password", ats: "lever", slug: "1password", sectors: ["security"] },
+  { name: "Snyk", ats: "greenhouse", slug: "snyk", sectors: ["security", "devtools"] },
+  { name: "Wiz", ats: "ashby", slug: "wiz", sectors: ["security"] },
+  { name: "Vanta", ats: "ashby", slug: "vanta", sectors: ["security"] },
+  { name: "Tailscale", ats: "ashby", slug: "tailscale", sectors: ["security", "devtools"] },
+  { name: "Netskope", ats: "greenhouse", slug: "netskope", sectors: ["security"] },
+  { name: "Rapid7", ats: "greenhouse", slug: "rapid7", sectors: ["security"] },
+  { name: "Datadog", ats: "greenhouse", slug: "datadog", sectors: ["security", "devtools"] },
+
+  // -- enterprise-saas --------------------------------------------------------
+  { name: "Asana", ats: "greenhouse", slug: "asana", sectors: ["enterprise-saas"] },
+  { name: "Notion", ats: "lever", slug: "notion", sectors: ["enterprise-saas"] },
+  { name: "Airtable", ats: "greenhouse", slug: "airtable", sectors: ["enterprise-saas"] },
+  { name: "Zapier", ats: "lever", slug: "zapier", sectors: ["enterprise-saas"] },
+  { name: "DocuSign", ats: "greenhouse", slug: "docusign", sectors: ["enterprise-saas"] },
+  { name: "Calendly", ats: "lever", slug: "calendly", sectors: ["enterprise-saas"] },
+  { name: "Rippling", ats: "ashby", slug: "rippling", sectors: ["enterprise-saas"] },
+  { name: "Checkr", ats: "greenhouse", slug: "checkr", sectors: ["enterprise-saas"] },
+  { name: "Samsara", ats: "greenhouse", slug: "samsara", sectors: ["enterprise-saas", "hardware-iot"] },
+  { name: "Figma", ats: "greenhouse", slug: "figma", sectors: ["enterprise-saas", "consumer-social"] },
+  { name: "Monday.com", ats: "lever", slug: "monday", sectors: ["enterprise-saas"] },
+  { name: "Miro", ats: "greenhouse", slug: "miro", sectors: ["enterprise-saas"] },
+  { name: "Coda", ats: "ashby", slug: "coda", sectors: ["enterprise-saas"] },
+
+  // -- consumer-social ----------------------------------------------------
+  { name: "Reddit", ats: "greenhouse", slug: "reddit", sectors: ["consumer-social"] },
+  { name: "Pinterest", ats: "greenhouse", slug: "pinterest", sectors: ["consumer-social"] },
+  { name: "Medium", ats: "lever", slug: "medium", sectors: ["consumer-social"] },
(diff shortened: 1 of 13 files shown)


s-annam:
## CORS verification — done ✅

Ran the browser CORS check authoritatively via Claude-in-Chrome, from a **real third-party https origin** (`https://example.com`), so the browser's own CORS engine is the judge — not curl (which can't prove browser enforcement). Testing from an arbitrary origin is *stricter* than the localhost check: a wildcard `access-control-allow-origin` that clears `example.com` clears `localhost` too.

**32 cross-origin `fetch` requests across all three vendors → zero CORS blocks.** A real block rejects `fetch` before the status is readable, so every resolved response (including 404s) is itself proof the vendor returned permissive ACAO.

| Vendor | Requests | CORS blocks | Real-data slugs proven | Verdict |
|---|---|---|---|---|
| Greenhouse | 3 | 0 | `stripe` 518, `figma` 168, `anthropic` 406 | ✅ PASS |
| Ashby | 4 | 0 | `ramp` 122, `vanta` 104, `cohere` 138 | ✅ PASS |
| Lever | 25 | 0 | `spotify` 100 | ✅ PASS |

All three boards clear browser CORS and are safe to wire into the live `getProviders()` seam in #533.

**One caveat, tracked separately (not a CORS failure):** the `company-registry.ts` slugs are frequently wrong — of the 21 Lever entries only `spotify` returns real jobs (19 return 404, `medium` is empty); the Ashby sample also had `mercury`/`vercel` empty and `tailscale`/`anduril` 404. This is a registry data bug the adapters correctly survive (never-fail-closed). It needs a full slug audit before boards go live under #533 — carried into that work, not filed as a blocker here.

src/lib/job-search/sector.ts
```diff
@@ -0,0 +1,309 @@
...
+  const haystack = buildHaystack(parsed);
+
+  const scored: Array<{ sector: Exclude<Sector, "other">; score: number }> =
+    (Object.keys(SECTOR_KEYWORDS) as Array<Exclude<Sector, "other">>).map(
+      (sector) => ({ sector, score: scoreSector(haystack, SECTOR_KEYWORDS[sector]) }),
+    );
+
+  // Deterministic tie-break: stable sort by score desc, then taxonomy order
+  // (the order sectors are declared in SECTOR_KEYWORDS, which mirrors SECTORS).
+  scored.sort((a, b) => b.score - a.score);
```
Samhit21: **Nit (non-blocking).** The comment above says the tie-break is "stable sort by score desc, then taxonomy order," but the comparator sorts by score only — the taxonomy-order tie-break is achieved *implicitly* via `Array.sort` stability plus the declaration-ordered `Object.keys(SECTOR_KEYWORDS)` input. It's correct and deterministic (sort stability is spec-guaranteed since ES2019), but `role-keywords.ts` encodes the same tie-break explicitly in its comparator (`FAMILY_ORDER`). Either add the explicit secondary key here for symmetry, or trim the comment so it doesn't claim a comparator behavior that isn't there — a future reader shouldn't have to know the sort is stable to trust the ordering.

Samhit21 (dismissed):
Reviewed the full diff (13 files, +2092/-0, all new `src/lib/job-search/`) at `dbe3147`. Foundation-only, well-tested, and the privacy story holds end-to-end. Concur with the verdict: no blockers.

**Privacy (the load-bearing gate here) — verified**
- All three adapters egress only `encodeURIComponent(slug)` + static params (`limit=100`, `includeCompensation=false`); every `fetch` is `fetch(url, { signal })`. No injection, no resume-derived query.
- The new company-board adapters take `_query` (unused) — correct, since there's no server-side search; the keyword feeds keep using `query`.
- Confirmed mechanically that nothing here imports `keywords.ts` (the audited egress helper) — the grep hit on `greenhouse.ts` was its docblock disclaiming it. `role-keywords.ts` is a distinct, pure, local title filter.
- `sector.ts`'s semantic path is on-device WebLLM (dynamic-imported, WebGPU-gated) — no network egress of resume data.

**offlinecv gates**
- Fixture PII / design-system / style-token / command-file — all **N/A** (pure lib, no fixtures/components/scripts).
- **fallow** — no dead-code findings: the tests consume every "not-yet-wired" export, so the deferred #533 integration doesn't trip the unused-export gate. Duplication is warn-only and justified in the body (per-vendor `mapJob` is the established adapter pattern).

**Correctness — spot-verified beyond the generic pass**
- `role-keywords.ts` — all three invariants hold *mechanically*: titles-not-skills (never reads `parsed.skills`), never-fail-closed (empty → permissive "all", `filterPostingsByRole` returns input unchanged), zero-egress. `capPerCompany` and the deterministic family tie-break are correct.
- `sector.ts` — heuristic floor + WebLLM fallback never rejects (validates against `SECTORS`, off-enum/throw → heuristic).
- `company-registry.ts` — 162 entries, no duplicate `(ats, slug)` pairs (so no provider-id collisions); `companiesForSector` caps correctly and is declaration-order deterministic.

Independently ran `npx vitest run src/lib/job-search` (106/106), `typecheck`, `eslint src/lib/job-search` — all green.

## Nits (non-blocking)
- One inline below on the `sector.ts` tie-break comment vs. comparator.

---

src/lib/job-search/sector.ts
s-annam: Fixed in `8892479`. Added a `SECTOR_ORDER` declaration-order map (mirrors `role-keywords.ts`'s `FAMILY_ORDER`) and made the tie-break an explicit secondary key in the comparator: `b.score - a.score || (SECTOR_ORDER.get(a.sector) ?? 0) - (SECTOR_ORDER.get(b.sector) ?? 0)`. Trimmed the comment so it no longer claims a comparator behavior that leaned on `Array.sort` stability. Ordering is unchanged (106/106 green); it's just legible without knowing the sort is stable.

src/lib/job-search/sector.ts
```diff
@@ -0,0 +1,322 @@
...
+
+  const runnerUpCandidate = scored[1];
+  const runnerUp =
+    runnerUpCandidate && runnerUpCandidate.score >= HEURISTIC_FLOOR
+      ? runnerUpCandidate.sector
+      : undefined;
+
+  // Confidence: top score's share of the top-two total, so a single
+  // dominant match reads as high confidence and a near-tie reads as low.
+  const denominator = top.score + (runnerUpCandidate?.score ?? 0);
```
Samhit21: [Suggestion]: The confidence formula awards 1.0 — the maximum — to the weakest possible classification.

`denominator = top.score + (runnerUpCandidate?.score ?? 0)` makes confidence `top / (top + second)`. All 16 sectors are scored, so `scored[1]` always exists; when exactly one sector matches at all, second is 0 and confidence is exactly **1.0**. With `HEURISTIC_FLOOR = 1` (line 154), a single spurious keyword hit is maximally confident.

Concretely, against the real tables: a frontend resume listing "routing" (React Router) matches `/\brouting\b/` (line 139, logistics-mobility) and nothing else → sector `logistics-mobility`, confidence 1.0, no runnerUp. A genuine data engineer with "data warehouse, ETL, Spark, Kafka streaming" scores data-ml 3, logistics-mobility 2 (`/\bwarehouse\b/` line 138 fires inside "data warehouse"), media-adtech 1 (`/\bstreaming\b/` line 143) → correct sector, confidence 0.6. The right answer reads as *less* confident than the wrong one, and slice 5's "not fintech? try devtools" affordance keys off this number.

The docblock at 224-225 describes share-of-top-two as intended, so this is a tuning call rather than a bug — but two orthogonal knobs would fix the inversion: make confidence monotone in absolute evidence (`top.score / (top.score + second + 1)`, or a saturating `min(1, top/3)`), and/or drop the generic single tokens that cross-match other families' phrases (`/\brouting\b/`, `/\bwarehouse\b/`, `/\bstreaming\b/`, `/\bmarketplace\b/`).

Worth noting `sector.test.ts:36` only asserts `confidence > 0`, so nothing in the suite pins the value.

Samhit21 (commented):
## PR Review: feat(job-search): company-targeted ATS-board search foundation (#528)

### Summary

Re-review at `8892479` after the force-push. The `sector.ts` tie-break fix is correct — `SECTOR_ORDER` mirrors `role-keywords.ts`'s `FAMILY_ORDER` and the ordering no longer leans on `Array.sort` stability. My earlier privacy and never-fail-closed verification still holds: the delta touched only the comparator.

Second pass across the full diff plus the sibling adapters (`remotive`/`arbeitnow`/`jobicy`) and `search.ts`. The adapters are faithful to the `JobProvider` contract as the siblings implement it — `signal` threaded into `fetch`, throw-on-non-2xx so `allSettled` degrades per-provider, `provider:id` dedup prefix, post-map `title && url` filter. Four net-new findings, none merge-blocking given nothing is wired in yet; they cluster in the sector confidence formula and in a few docblock promises no test pins.

Two candidates I chased and killed, for the record: the registry has no duplicate `name` and no duplicate `slug` across all 162 entries, and `companiesForSector` handles `limit <= 0` correctly.

### Findings

1. `sector.ts:226` — confidence formula is effectively inverted: a lone spurious keyword scores 1.0 while a well-evidenced multi-hit resume scores 0.6. Load-bearing for slice 5's UI.
2. `providers/lever.ts:86` — `(data ?? [])` doesn't cover the non-array body its own test claims to cover; Lever is the one adapter whose success shape is a bare array.
3. `providers/greenhouse.ts:54` (and `lever.ts:60`, `ashby.ts:59`) — dropped the `?? url` id-fallback the siblings added to stop id-less collisions, while keeping the comment that claims uniqueness.
4. `role-keywords.ts:289` — second family survives at any score, so one stray title widens the filter to a family the candidate left.

### Meta — the `"other"` seam between sector.ts and the registry

`sector.ts:16-20` makes "classification never fails closed" an invariant with `"other"` as the floor, but no registry entry is tagged `"other"` (0 of 162), so `companiesForSector("other", n)` returns `[]` — a never-fail-closed classification becomes a zero-result search. `company-registry.test.ts:50` explicitly `continue`s past `"other"`, so the gap is known but nothing absorbs it. Either give the registry an `"other"`-tagged general set, or make it #533's job to substitute a default sector and pin the contract with an assertion. Flagging here rather than inline because it spans both modules.

### Open question for #533

`hydrateGreenhouse` (greenhouse.ts:93) takes `signal?: AbortSignal` as optional, where `types.ts:52-54` makes threading `signal` mandatory — a caller that omits it gets an uncancellable request and the type won't stop them. There's also no bounded-concurrency helper: with `DEFAULT_PER_COMPANY_CAP = 15` and a double-digit `companiesForSector` limit, #533 fans out 100+ GETs, and one 404'd job id rejects whatever `Promise.all` wraps them. If #533 owns the pool that's fine — but a `hydrateMany(jobs, {concurrency, signal})` wrapping `allSettled` belongs next to the primitive, and the required-`signal` version costs nothing now.

### Verdict

**Action: COMMENT.**

---

src/lib/job-search/providers/lever.ts
```diff
@@ -0,0 +1,89 @@
...
+    id: ID,
+    label: LABEL,
+    // Whole board is fetched (no server-side search), capped via `?limit=` —
+    // narrowing by keyword happens client-side downstream (#534).
+    async search(_query, signal: AbortSignal): Promise<JobPosting[]> {
+      const url = `${BASE}/${encodeURIComponent(slug)}?mode=json&limit=${LIMIT}`;
+      const res = await fetch(url, { signal });
+      if (!res.ok) throw new Error(`${LABEL} responded ${res.status}`);
+      const data = (await res.json()) as LeverJob[];
+      return (data ?? []).map((j) => mapJob(slug, LABEL, j)).filter((p) => p.title && p.url);
```
Samhit21: [Suggestion]: `(data ?? [])` doesn't guard the non-array case the test claims to cover.

`lever.test.ts:88` is titled "tolerates a non-array response" but mocks `null` — which `??` already handles. A real non-array body still throws: `({} ?? []).map` is not a function. Change the mock to `mockFetch({})` and the test fails today.

Lever is the one adapter where this bites: it's the only one of the three whose success shape is a bare top-level array, so any 200-with-object error envelope lands straight on `.map`. And with registry slugs explicitly unverified (`company-registry.ts:20-28`), a wrong slug is the expected path, not the exotic one.

`Array.isArray(data) ? data : []`, plus the `{}` fixture.

src/lib/job-search/providers/greenhouse.ts
```diff
@@ -0,0 +1,106 @@
...
+interface GreenhouseIndex {
+  jobs?: GreenhouseJob[];
+}
+
+function mapJob(slug: string, label: string, job: GreenhouseJob): JobPosting {
+  return {
+    // `greenhouse:{slug}:{jobId}` is unique across companies — doubles as the
+    // React key and cross-provider dedup id, matching the `provider:id`
+    // convention in types.ts.
+    id: `greenhouse:${slug}:${job.id ?? ""}`,
```
Samhit21: [Suggestion]: Drops the id-fallback the sibling adapters added specifically to prevent id-less collisions.

`remotive.ts:40` and `jobicy.ts:40` both use `job.id ?? job.url ?? ""`, with a comment spelling out why: "so two id-less postings never collide on `remotive:undefined` — the id doubles as the React key and cross-provider dedup id." All three new adapters use a bare `job.id ?? ""` (greenhouse.ts:54, lever.ts:60, ashby.ts:59) while the adjacent comment still asserts the id is unique.

A board returning two postings with `id: null` but valid `absolute_url` + title survives the `p.title && p.url` filter at line 83 and yields two postings both keyed `"greenhouse:acme:"` → duplicate React keys and a silent dedup collapse.

`job.id ?? job.absolute_url ?? ""` here, `hostedUrl` / `jobUrl` in the other two, restores parity.

src/lib/job-search/role-keywords.ts
```diff
@@ -0,0 +1,339 @@
...
+    return { families: [], keywords: [], source: "heuristic" };
+  }
+
+  scored.sort(
+    (a, b) =>
+      b.score - a.score ||
+      (FAMILY_ORDER.get(a.family) ?? 0) - (FAMILY_ORDER.get(b.family) ?? 0),
+  );
+
+  const families = scored.slice(0, MAX_FAMILIES).map((entry) => entry.family);
```
Samhit21: [Suggestion]: `slice(0, MAX_FAMILIES)` keeps the second family however far behind it is, so one stray title floods the filter.

The docblock at line 261 says "keeps the dominant 1–2 families", but the only gate is `score > 0` (line 276). A career-switcher with eight backend titles and one early-career "UX Designer" gets `families: ["backend", "design"]` — and design's keywords include the very broad "designer" and "user experience" (lines 169-171). `filterPostingsByRole` then keeps every Designer posting on every board, precisely the roles the candidate left.

A relative floor would do it — keep the second family only when `entry.score * 2 >= scored[0].score`.

The test at line 129 uses a 2-1-1 spread, which never exercises a lopsided split; an 8-vs-1 case is worth adding alongside.

src/lib/job-search/sector.ts
s-annam: Fixed in `e93f038`.

Both knobs, since they fail in different directions.

**Confidence is now monotone in absolute evidence:** `share × min(1, top/CONFIDENT_HITS)` with `CONFIDENT_HITS = 3`. Share-of-top-two survives (a near-tie still reads low) but a lone hit no longer saturates. Your two cases: `routing`-only → `1.0 × 1/3 = 0.33`; data engineer 3-vs-2 → `0.6 × 1.0 = 0.6`. Inversion gone.

**The cross-matching tokens are phrases now,** so that first case does not even arise:
- `/\brouting\b/` → `/\broute optimization\b/`
- `/\bwarehouse\b/` → `/\bwarehouse (?:management|operations|automation)\b/` (no longer fires inside "data warehouse")
- `/\bstreaming\b/` → `/\b(?:video|live|music) streaming\b/` + `/\bstreaming (?:platform|service)s?\b/`
- `/\bmarketplace\b/` → `/\b(?:online|retail|seller|consumer) marketplace\b/` + `/\bmarketplace (?:platform|operations)\b/` (it was picking up "NFT marketplace" and "API marketplace" too)

The keyword-table docblock now states the rule so the next addition follows it: phrase, not bare noun, wherever the bare noun belongs to another family's vocabulary.

**Tests pin the value now,** not just `> 0` — one asserts a single-hit guess ranks strictly below a well-evidenced one, another that "Data warehouse, Kafka streaming, ETL, Spark" classifies data-ml with no runner-up at all.

src/lib/job-search/providers/lever.ts
s-annam: Fixed in `e93f038`. `Array.isArray(data) ? (data as LeverJob[]) : []`, and the response is typed `unknown` at the `json()` call so the cast is the only place the shape is asserted.

The test is split rather than changed: `null` keeps its own case (it is a real shape and worth pinning), and a new "tolerates a 200 whose body is an object, not an array" mocks `{ error: "not found" }` — which does fail against the old `?? []`. The comment at the guard names why Lever specifically: bare top-level array is the success shape, so a 200 error envelope lands straight on `.map`, and with the registry slugs unverified that is the expected path.

src/lib/job-search/providers/greenhouse.ts
s-annam: Fixed in `e93f038`. Parity restored across all three: `job.id ?? job.absolute_url` (greenhouse), `job.id ?? job.hostedUrl` (lever), `job.id ?? job.jobUrl` (ashby) — each with the sibling adapters' comment explaining that the url is guaranteed non-empty by the post-map filter, so the adjacent "unique" claim is true again rather than aspirational.

Each adapter also gets a test asserting two id-less postings with distinct urls yield distinct ids, so the collision cannot come back silently.

src/lib/job-search/role-keywords.ts
s-annam: Fixed in `e93f038`. The relative floor you proposed, as a named constant: the runner-up is kept only when `entry.score * RUNNER_UP_SHARE >= winner.score` with `RUNNER_UP_SHARE = 2`. `MAX_FAMILIES` still caps at 2 on top of it. The docblock at the function no longer says "1–2 families" flatly — it names the floor.

Added the 8-vs-1 case alongside the existing 2-1-1 one: three backend titles plus one "UX Designer" now yields `["backend"]`, and the test also asserts `keywords` does not contain `"designer"` — the actual failure mode was the broad keyword leaking into the filter, not the family label itself. The existing 2-1-1 test still passes (1 × 2 >= 2), which is the intended behaviour: a genuinely close second stays.

src/lib/job-search/company-registry.ts
```diff
@@ -0,0 +1,256 @@
...
+  { name: "HawkEye 360", ats: "greenhouse", slug: "hawkeye360", sectors: ["government-defense"] },
+  { name: "Second Front Systems", ats: "ashby", slug: "secondfrontsystems", sectors: ["government-defense"] },
+  { name: "Vannevar Labs", ats: "ashby", slug: "vannevarlabs", sectors: ["government-defense"] },
+];
+
+/**
+ * Up to `limit` registry entries tagged with `sector`, in stable declaration
+ * order (deterministic across calls — the fan-out cap in #533 relies on it).
+ */
+export function companiesForSector(sector: Sector, limit: number): CompanyEntry[] {
```
Samhit21, line 249: **Secondary (carry-over) — a never-fail-closed classification still lands on a zero-result search, and finding 1's fix made it more reachable.**

Still 0 of 162 entries tagged `"other"`, so `companiesForSector("other", n)` returns `[]` for any limit. `sector.ts:16-20` makes "classification never fails closed" an invariant with `"other"` as the floor; the registry doesn't hold up the other end. `company-registry.test.ts:50` still `continue`s past `"other"`, so the gap stays known and unpinned.

What's new this round: the fix for finding 1 narrowed `/\bwarehouse\b/`, `/\brouting\b/`, `/\bstreaming\b/` and `/\bmarketplace\b/` to phrases. That is strictly score-*reducing* — no resume scores higher than it did before, and a resume that previously cleared `HEURISTIC_FLOOR = 1` on one of those bare tokens alone now scores 0 and falls to `"other"`. So the correct fix to the confidence inversion also routes strictly more users into the one sector with no companies behind it. The two changes are individually right and interact badly.

Still not merge-blocking — nothing is wired up until #533 — but it's now the highest-value thing to resolve before that lands. Either tag a general-purpose set `"other"`, or make #533 substitute a default sector and pin the contract with an assertion so it can't silently regress.

src/lib/job-search/providers/greenhouse.ts
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ * Lazy per-job hydrate: fetches one Greenhouse posting and returns its
+ * `content` as plaintext. Called only for postings that survive the #534
+ * title filter and per-company cap — never for the whole board.
+ */
+export async function hydrateGreenhouse(
+  slug: string,
+  jobId: string,
+  signal?: AbortSignal,
```
Samhit21, line 98: **Secondary (carry-over) — optional `signal` contradicts the interface contract, and there's still no bounded-concurrency helper next to the primitive.**

`types.ts:52-54` makes threading `signal` mandatory ("MUST thread `signal` into `fetch` so an in-flight or superseded search can be cancelled"), but `hydrateGreenhouse` takes it as `signal?: AbortSignal` — a caller that omits it gets an uncancellable request and the type won't stop them. Making it required costs nothing now and something later.

The fan-out half is unchanged too: with `DEFAULT_PER_COMPANY_CAP = 15` and a double-digit `companiesForSector` limit, #533 issues 100+ GETs, and one 404'd job id rejects whatever `Promise.all` wraps them. If #533 owns the pool that's a fine answer — but a `hydrateMany(jobs, { concurrency, signal })` wrapping `allSettled` belongs beside the primitive it paces, not in the caller. `grep` finds no such helper in `src/lib/job-search/`.

Re-raising because it went unanswered last round, not because it blocks — a one-line "#533 owns this" closes it.

Samhit21 (approved):
Re-review at `e93f038`. All four findings from the `8892479` pass are fixed, and I verified each against the tree rather than the reply. The delta since my last review is 171/-19 across 10 files — I read all of it.

**1. Sector confidence inversion — fixed, both knobs, correctly.** `share × min(1, top/CONFIDENT_HITS)` with `CONFIDENT_HITS = 3`. The inversion is gone: a lone hit is `1.0 × 1/3 = 0.33`, the 3-vs-2 data engineer is `0.6 × 1.0 = 0.6`. I checked the removed `denominator === 0` guard is not needed — the `top.score < HEURISTIC_FLOOR` early return at line 231 means the denominator is ≥ 1 by the time the division runs, so no `NaN` path. The four cross-matching bare nouns are phrases now, and the keyword-table docblock states the rule so the next addition follows it. Tests pin the value instead of `> 0`.

**2. Lever non-array body — fixed.** `Array.isArray(data) ? (data as LeverJob[]) : []`, with `json()` typed `unknown` so the cast is the single place the shape is asserted. Splitting the test rather than changing it was the right call — `null` is a real shape worth keeping pinned, and the new `{ error: "not found" }` case does fail against the old `?? []`.

**3. id-fallback parity — fixed across all three.** `job.id ?? job.absolute_url` / `?? job.hostedUrl` / `?? job.jobUrl`. I confirmed the "guaranteed non-empty by the post-map filter" claim actually holds: `.filter((p) => p.title && p.url)` drops any posting whose url is empty, so the fallback can't itself produce a bare `provider:slug:`. Each adapter has a two-id-less-postings test.

**4. Runner-up family floor — fixed.** `entry.score * RUNNER_UP_SHARE >= winner.score` with `MAX_FAMILIES` still capping on top. The `scored.length === 0` guard above the destructure means `winner` is always defined. The 8-vs-1 test asserts on `keywords` not containing `"designer"`, which is the actual failure mode rather than just the family label — good instinct.

**Gates.** Fixture-PII / design-system / style-token / command-file all **N/A** (pure lib; no fixture, component, or script touched — confirmed by path-scoped diffs). `typecheck`, `eslint src/lib/job-search`, and `vitest run src/lib/job-search` (14 files, **113 tests**, up from 106) green locally; CI `verify` + `fallow` green. Privacy re-verified on the delta: every adapter URL is still `encodeURIComponent(slug)` plus static params, every `fetch` threads `signal`, and nothing under `providers/` touches `keywords.ts` or `query` — the only hit is greenhouse's docblock disclaiming it.

No blockers. Two carry-overs below that got no reply last round — I'm re-raising rather than letting them drop, not gating on them.

## Secondary (carry-over, unanswered)

**5. The `"other"` seam — and fix 1 just widened it.** Inline at `company-registry.ts:249`. This is the one place the re-review changed my assessment: narrowing the keyword table is strictly score-reducing, so strictly more resumes now land in `"other"`, where the registry has nothing to offer.

**6. `hydrateGreenhouse`'s optional `signal`, and no bounded-concurrency helper.** Inline at `greenhouse.ts:98`.

## Nit

**7. `RUNNER_UP_SHARE = 2` names a share but holds a divisor.** The docblock says "at least half the winner's score" and the comparator reads `e.score * 2 >= winner.score`, so the actual share is `1/2` — the constant is the reciprocal of its own name. `MIN_RUNNER_UP_RATIO = 0.5` with `e.score >= winner.score * MIN_RUNNER_UP_RATIO` reads as what it is. Purely a naming call; behaviour is right and documented.

---

Issue #381, opened 2026-07-07
[persistence] inline edits persist across re-drops of same PDF with no visual signal distinguishing user edits from parser output


Vaishnavi1709:
## Category
edit-persistence (Steps 1 & 3 F&F QC — provenance category)

## Fixture
Reproducible on any fixture — validated during F&F QC on:
- `tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-achievements-oneline.pdf`

## Observed
1. Drop a PDF, wait for the reconstructed view
2. Edit any field to an obvious value (e.g. change a company name to `TEST123`)
3. Click **Try another file**
4. Drop the **same PDF** again
5. The reconstructed view re-appears with `TEST123` still on the company field — the edit survived the re-drop

Silently. No banner, no italic markers, no timestamp, no "you edited this before" indicator, no way to distinguish which fields are user edits vs. what the parser would produce on a fresh parse of this file.

## Category rationale
This isn't a bug in the strict sense — persistent edits are usually good behaviour (user doesn't lose 5 minutes of cleanup work). The find is that the persistence is **invisible**: someone testing multiple résumés might see their old edits reappear when they revisit an earlier file and be confused about whether the parser regressed or their old edits are showing.

## Localized layer
Likely a `localStorage` key of the form `rl_*` (per CLAUDE.md's Telemetry section referencing `rl_*` functional keys) keyed by a content hash of the PDF. Rehydrated on drop.

Fix likely lives in the state-hydration path in `src/App.tsx` or wherever the reconstructed-view state is initialized after a drop.

## Suggested fix
Two low-effort options:
1. **Banner:** on drop, if any hydrated edits are found for this file's content hash, show a dismissable banner: `"Restored N edits you made previously. [Reset to parser output]"`.
2. **Field-level:** render user-edited fields with a subtle marker (dot, italic, or a "Restore original" chip on hover).

Option 1 alone is enough for the QC use case. Option 2 is nicer long-term but higher-effort.

## Repro
```
npm run dev
# drop any fixture; edit any field to "TEST123"; press Enter
# click "Try another file"; drop the SAME fixture; check whether "TEST123" is still there
```

s-annam:
Not reproducible on `main` @ `1b391e6`. Edits do **not** survive a re-drop, and no edit state for an uploaded résumé is persisted anywhere.

[...]

PR #536, opened 2026-07-21
feat(edit): confirm what an applied rewrite changed, make it undoable, and name controls by content (#511)

### LINKED ISSUE #508: [edit-interaction] applying rewrite suggestions gives no confirmation — panel dismisses silently, Download PDF reads as the commit step
## Summary

Applying rewrite suggestions gives the user no confirmation that the edits landed. `onApply` writes every accepted change and then immediately dismisses the whole review panel, so from the user's point of view the proposal UI simply disappears — there is no "Applied ✓" state, no toast, no highlight on the changed text. The next thing in view is the **Download PDF** CTA, which then reads as the step that *commits* the edits rather than one that exports an already-committed résumé.

**The behavior is correct** — the changes *are* applied, and the exported PDF faithfully reflects them (this is covered by the round-trip invariants in `corpus-edit-roundtrip.test.ts` and `description-override-roundtrip.repro.test.ts`). This is purely an **affordance / feedback** defect, not a data-integrity bug.

The core problem, stated precisely: **the only signal that anything happened is a disappearance.** Absence of UI is not feedback.

## Where it lives

`src/components/features/ResumeRewriteProposed.tsx:104-119` — `onApply` loops the accepted writes and unconditionally calls `onDismiss()` as its last statement, collapsing the proposed-state UI in the same tick the writes happen.

`src/components/features/RewriteReviewList.tsx` (the per-role review) has the same shape: the Apply button's only visible effect is that the surface it lives in goes away.

The Download CTA is a sibling that is always rendered at the top of `ReconstructedResume.tsx`, so it is not "revealed" by applying — it just becomes the most prominent remaining call to action once the review panel collapses, which is what makes it read as the pending step.

## Repro

1. Drop any résumé and let it parse.
2. Open a rewrite proposal (whole-résumé or a single section) and accept at least one suggestion.
3. Click **Apply N changes**.
4. Observe: the review panel disappears. No confirmation the writes committed, and no visual marker on the résumé body showing which text changed.
5. The remaining prominent CTA is **Download PDF**, which now reads as "the button that applies my edits".

## Expected

Confirm **in place**, in the panel that is already mounted. Do not dismiss synchronously.

`ResumeRewriteProposed` already renders inside `InlineResult`, which already carries a `success` tone (`src/design-system/shared/InlineResult.tsx:17-20`), and `StatusBadge` already has an `ok` tone. Both come from the `@design-system` barrel. The confirmation is therefore a **content swap inside a component that is already on screen** — no new component, no new primitive.

Sequence:

| Time | Behavior |
|---|---|
| `t=0` | Click Apply → writes commit (unchanged) |
| `t=0` | Panel content swaps to the confirmation state; `onDismiss()` no longer fires here |
| `t≈3s` | Panel collapses on its own |

Confirmation content:

- `<StatusBadge tone="ok">Applied</StatusBadge>` followed by a scope line — e.g. **`Applied 4 changes — Experience, Skills`**.
- Naming the count *and* the sections matters: it tells the user what moved without making them hunt the résumé body for diffs.

## Acceptance criteria

- After clicking Apply, the user sees an explicit success signal tied to the apply action, not merely the absence of the review panel.
- The signal names the scope of what was applied: the count of changes **and** the sections touched.
- The confirmation is reachable by assistive tech — announced via `aria-live="polite"`. Existing precedent in the repo: `JobSearchResults.tsx`, `src/design-system/shared/UpdateBanner.tsx`.
- Meaning is not carried by colour alone — the word "Applied" appears, not just a green chrome tone.
- The confirmation holds for roughly 3 seconds, then the panel collapses without user action.
- The collapse animates in the 150–300ms band, with the exit shorter than a corresponding enter, and is gated on `prefers-reduced-motion`.
- Built from existing `@design-system` barrel exports (`InlineResult`, `StatusBadge`, `Button`). **No new toast/snackbar primitive** — see Notes.
- No change to what is written or to the exported PDF: the parse → edit → export round-trip invariants (`corpus-edit-roundtrip.test.ts`, `render-roundtrip.repro.test.ts`) must stay green and `ATS_SCORE_ALGO_VERSION` must not move.

## Notes

**Why not a toast.** The reflexive fix for "action completes silently" is a success toast. The design system has no toast or snackbar primitive — the barrel (`src/design-system/index.ts:17-33`) exports `Button, Chip, Dialog, EditableField, StarRating, TextAreaField, Card, InlineResult, ModelLoadProgress, StatusBadge, Tabs, CountBadge, ErrorState, ErrorBoundary, UpdateBanner, GitHubStarCta, InlineDiff` and nothing else. Adding one would mean a new shared component for a single callsite, against the reuse rule in CLAUDE.md. Confirming in the already-mounted panel is both cheaper and more spatially coherent: the feedback appears where the action was taken.

**Undo is out of scope** — split to #510. Reversing an applied batch needs an inverse-write path (snapshot-before-apply or reverse diff), which is a change to the write path rather than to render state. This issue stays copy + render state so it can land first. If #510 is in flight, its Undo control hangs off the confirmation state introduced here.

**Compounds with #509.** The Download CTA that reads as "the commit step" here is also ambiguously named there — `Download PDF` next to `Download report`, both emitting PDFs. Fixing the label narrows the misread; fixing this issue removes it. Same surface, best done as one pass.

Blast radius is small: this is a state-affordance and copy change in the rewrite-apply surface, not a pipeline change. It matters because edit → download is the core conversion moment of the parser-audit lane, and the ambiguity shows up on a user's very first pass through it.

### LINKED ISSUE #509: [edit-interaction] download controls named by file format, not content — "Download PDF" vs "Download report" both emit PDFs
## Summary

The reconstructed-résumé header renders two download controls side by side, and neither label tells the user what they will actually get. **Both emit a PDF.** A first-time user cannot tell that the primary CTA is the one that gives them their résumé.

| Control | Variant | Label | What it actually downloads |
|---|---|---|---|
| `ReportDownloadControl` → `DownloadReportDialog.tsx:69` | ghost | **Download report** | the audit findings — verdict, score breakdown, layout flags, recommendation. `format` state defaults to `"pdf"`. |
| `ReconstructedResume.tsx:1178` | primary | **Download PDF** | the reconstructed résumé, rendered ATS-safe |

One control is named by its **content** ("report"), the other by its **file format** ("PDF"). Nothing in either label says "resume". Since the report also exports as PDF by default, "Download PDF" is not even distinguishing on format — the axis it was named for.

A third, unrelated download exists elsewhere on the page: `ModelSelector.tsx:341` renders **"Download now · ~N GB"** for on-device model weights. So the page carries three download affordances named on three different axes — content, format, and bare action.

## Scope: naming only

**Visual hierarchy is already correct and is not part of this issue.** The résumé download is `variant="primary"` and the report download is `variant="ghost"`, which correctly subordinates the secondary action to the primary one. The defect is purely the label text on two of the three controls.

**The `aria-label`s are also already correct** — and are strictly more informative than the visible labels they accompany:

- `ReconstructedResume.tsx:1176` — `"Download the reconstructed resume as an ATS-friendly PDF"`
- `DownloadReportDialog.tsx:67` — `"Download the audit report as a shareable file"`

A screen-reader user is told which document each control produces; a sighted user is not. The fix direction therefore runs one way only: **raise the visible copy to match the `aria-label`'s clarity, never weaken the `aria-label` to match the visible copy.**

## Why it matters

Downloading a clean, ATS-safe résumé is the primary output of the parser-audit lane — it is what the user came for. Burying it behind the least descriptive label of the three means the highest-value action is the hardest one to identify. Users reading "Download PDF" next to "Download report" reasonably read the pair as *two formats of the same thing* rather than two different documents.

## Repro

1. Drop any résumé and let it parse.
2. Scroll to the **Reconstructed resume** section header.
3. Observe **Download report** and **Download PDF** adjacent in the same control row.
4. Without clicking, determine which one yields your résumé. There is no label, tooltip, or helper copy that answers this — only the `aria-label`s, which sighted users never see.

## Expected

Name every download control by **what document it produces**, never by file format. Format is a choice made inside the dialog, not an identity.

| Location | Now | Change to |
|---|---|---|
| `ReconstructedResume.tsx:1178` | `Download PDF` | **`Download resume`** |
| `DownloadReportDialog.tsx:69` | `Download report` | unchanged — already content-named |
| `ModelSelector.tsx:341` | `Download now · ~1.9 GB` | **`Download model · ~1.9 GB`** |

Spelling: use unaccented **`resume`** in the visible label, matching the existing `aria-label` verbatim so visible text and screen-reader text share one vocabulary. (Docs prose keeps `résumé`; this decision covers UI control copy only.)

## Acceptance criteria

- No download control in the app is labelled by file format alone.
- Every download control names the object it produces — the document, not the container.
- Format selection lives in a dialog, never in a label.
- The two controls in the reconstructed-résumé header are distinguishable by visible label text, with no dialog, tooltip, or `aria-label` needed to disambiguate.
- Visible label and `aria-label` agree on which document is produced, with the visible label a subset of the `aria-label`'s vocabulary.
- Button variants are unchanged — the résumé download stays `primary`, the report download stays `ghost`.
- Copy-only change: no change to what any control emits, and the parse → edit → export round-trip invariants (`corpus-edit-roundtrip.test.ts`, `render-roundtrip.repro.test.ts`) stay green.

## Notes

Compounds with #508. When applying rewrite suggestions produces no confirmation, the nearest prominent CTA becomes the user's guess at "the button that commits my edits" — and an ambiguously-named Download CTA is precisely what makes that misreading available. Fixing the label narrows the misread; fixing #508 removes it. Best handled as one design pass over the same surface.

Mechanical enforcement of the naming invariant (an ESLint rule flagging format-named download labels, alongside the existing no-raw-`<button>` and no-hex guards) was considered and deliberately left out of this issue — it is tooling work, not a copy fix. File separately if the rule proves worth automating.

### LINKED ISSUE #510: [edit-interaction] no undo for an applied rewrite batch — bulk mutation with no recovery path
## Summary

Applying rewrite suggestions is a bulk, irreversible mutation. `onApply` in `src/components/features/ResumeRewriteProposed.tsx:104-119` walks every accepted pair and issues `onAdd` / `onReplace` / `onRemove` writes across multiple sections in one pass — then discards the proposal. There is no way back. A user who applies a batch and then decides they preferred the original text has to retype it from memory, because the pre-apply state is not retained anywhere.

The same shape exists in the per-role review (`RewriteReviewList.tsx`).

## Why it matters

Every other destructive-ish interaction in the app is cheap to reverse — inline field edits can be re-edited, and a bad parse can be re-dropped. Applying a rewrite batch is the one action that silently destroys user-authored content with no recovery path. It is also the action most likely to be taken speculatively ("let me see what it suggests"), which is exactly the case where a way back matters most.

## Repro

1. Drop any résumé and let it parse.
2. Open a rewrite proposal covering more than one section.
3. Accept several suggestions and click **Apply N changes**.
4. Decide the original wording was better.
5. There is no undo. The prior text is gone from the UI and from state.

## Expected

Offer a way to reverse an applied rewrite batch as a single unit — one action restores every field the batch touched, not a field-by-field manual repair.

Two viable implementations, both client-side; the design decision is deliberately left open:

- **Snapshot before apply** — capture the affected slice of the résumé model immediately before the write loop, and restore it wholesale on undo. Simpler, and exact.
- **Reverse diff** — record the inverse of each `add` / `replace` / `remove` write and replay it backwards. Smaller memory footprint, composes if undo depth ever needs to exceed one.

Single-level undo (the most recent batch) is sufficient. A full history stack is explicitly not being asked for.

## Acceptance criteria

- After applying a rewrite batch, the user can reverse the entire batch in one action.
- Undo restores every field the batch touched, across all sections, to its exact pre-apply value.
- Undo is discoverable at the moment it is useful — surfaced with the apply confirmation, not buried in a menu.
- Scope is single-level: undoing the most recent applied batch. No multi-step history.
- Undoing then re-applying is not required to work; undo may be one-shot per batch.
- The parse → edit → export round-trip invariants (`corpus-edit-roundtrip.test.ts`, `render-roundtrip.repro.test.ts`) stay green, and an undone batch exports identically to a batch that was never applied.

## Notes

Split out of #508. That issue covers the *confirmation* that an apply happened — render state and copy only, no change to the write path. This issue is the harder half: it requires retaining or inverting the writes themselves.

The two are complementary and share a surface. #508's confirmation strip is the natural place to host the Undo affordance, so if both are in flight, land #508 first and hang the control off the confirmation state it introduces.

### LINKED ISSUE #519: [result] tab labels read as inspector panes — add one-line descriptions to Tab

s-annam:
Four sub-issues of the epic, implemented on one branch and landed as one commit.

Each of the four sub-issues is fully resolved here, so they close. The epic itself gets `Refs` rather than `Closes` — it still has three open children (#391, #379, #381).

## What changed

- **#508 — applying a rewrite gives no confirmation.** Apply no longer dismisses the review
  panel in the same tick with no signal. A new `ApplyConfirmation` renders in place inside the
  caller's already-mounted `InlineResult` (no new toast primitive — see CLAUDE.md's reuse
  rule): a `StatusBadge` plus the line *"Applied N changes — <sections touched>"*, held ~3s,
  then collapsed. `role="status" aria-live="polite"`; the word "Applied" is in the text, so
  meaning is never carried by colour alone; motion is gated on `prefers-reduced-motion`.
  Shared by both apply surfaces so the per-role and whole-résumé paths show identical copy,
  timing, and motion.
- **#509 — download controls named by file format.** "Download PDF" → "Download resume" and
  "Download now" → "Download model", so each control names the *content* it emits rather than
  the container. "Download report" was already content-named and is unchanged.
- **#510 — no undo for an applied rewrite batch.** The confirmation now hosts a one-shot
  **Undo**, backed by `src/lib/rewrite-review/undo-batch.ts`. It is a **snapshot**, not a
  reverse diff: `remove` writes into a set (so the inverse of "remove i" depends on pre-state
  that lives nowhere in the write) and `add` appends while silently dropping blank text (so
  write count ≠ bullets appended, and duplicate texts make removal-by-value ambiguous). The
  snapshot is scoped to the slots the batch will touch — three small maps, not a clone of the
  parse. A confirmation that hosts an Undo holds for 12s instead of 3, because an Undo the
  user cannot reach is not a recovery path; the no-undo timing from #508 is unchanged.
  Documented tradeoff: undoing after a hand-edit to one of the *same* slots also reverts that
  hand edit — the literal reading of "restore the exact pre-apply value", bounded because the
  affordance is one-shot and only mounted alongside the confirmation.
- **#519 — tab labels read as inspector panes.** The shared `Tab` primitive takes an optional
  `description`, and the three result tabs now carry one-line subtitles. Copy was checked
  against the governing privacy sentences in `FindJobsPanel` — the find-jobs subtitle claims
  ranking, not upload — and the resume-quality subtitle is conditioned on
  `analysis.isAvailable` so it does not promise on-device critique where WebGPU is missing.
  Keyboard behaviour is unchanged and now pinned by tests.

## Known, deliberate behaviour

A batch where every accepted pair is verbatim writes nothing, so it shows **no** confirmation
and simply dismisses — pre-#508 behaviour. Confirming a mutation that did not happen would be
a false claim; the honest answer, if this ever needs one, is a "No changes to apply" variant
of the same strip, not a re-inflated count.

## Verification

`npm run typecheck`, `npx eslint src`, and the full `npx vitest run` are green over the
accumulated tree: **184 files passed | 6 skipped; 2561 tests passed | 9 skipped**. The
authoritative full `verify` runs in CI on this PR.

`npx fallow audit --base origin/main` is clean of anything this batch introduced. The one
CRITICAL it raised mid-run — `Tabs.tsx` `onKeyDown` at CRAP 110 with 0% coverage, against
#519's explicit "keyboard behaviour unchanged" criterion — is fixed: `Tabs.test.tsx` now pins
roving tabindex and ArrowRight/ArrowLeft/Home/End, parameterised over the description and
no-description branches. The remaining complexity findings and the flagged duplication are
pre-existing and untouched by this diff.

No fixtures were added or changed by any of the four issues, so the fixture-PII surface is
untouched.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #508 | unreported (requested: sonnet) | high |
| Implementation — #509 | unreported (requested: sonnet) | medium |
| Implementation — #510 | unreported (requested: opus) | ultra |
| Implementation — #519 | unreported (requested: sonnet) | high |
| Adversarial review — round 1 | Claude Opus 4.8 | high |
| Review fixes | Claude Opus 4.8 | high |
| Adversarial review — round 2 | Claude Opus 4.8 | high |
| Orchestration + PR | Claude Opus 4.8 | high |
| Review revisions | Claude Opus 4.8 | high |

The four implementing subagents went idle without emitting their structured return, so their
model strings are genuinely unknown — the requested alias is recorded instead of a guess. Their
work was verified from the tree by the orchestrator and by both review rounds.

Refs #511
Closes #508
Closes #509
Closes #510
Closes #519

## Adversarial review

Two bounded rounds against the accumulated diff, pre-PR, each reproducing suspected defects
end-to-end rather than reasoning about the code.

**Round 1 — 3 blocking, all reproduced with scratch jsdom tests:**

1. *The confirmation counted accepted pairs, not writes that landed.* `resolveBulletActions`
   drops a matched-accept whose text equals the original, but `acceptMany` accepts every pair
   including verbatim ones — so a rewrite that returns any bullet unchanged (routine model
   behaviour) announced more changes than it committed. Reproduced: one `onReplace`, count of
   2. Both apply surfaces. **Fixed** by accumulating `writes.length` into a `writtenCount`;
   `acceptedCount` now only labels the Apply button.
2. *An all-verbatim batch still confirmed, with an empty section list.* Apply gates on
   `accepted === 0`, not on writes, so every section was skipped and the strip rendered
   literally `"Applied 2 changes — "` — a dangling em dash claiming a mutation that never
   happened, with no Undo. **Fixed** by never entering the confirmation state for a zero-write
   batch (both surfaces), plus a belt-and-braces guard that omits the clause when there are no
   sections.
3. *The per-role surface offered an Undo over an empty snapshot.* `captureUndo?.(writes)` ran
   unconditionally, so clicking Undo would flip the strip to "Reverted 2 changes — …" while
   restoring nothing — a false statement about the résumé on the one surface whose purpose is a
   trustworthy recovery path. **Fixed** with an early return before the capture.

Also repaired from the fallow leads: `Tabs.tsx` `onKeyDown` was CRAP 110 at 0% coverage while
"keyboard behaviour unchanged" is an explicit acceptance criterion — now pinned by 12 tests
parameterised over the description / no-description branches. `HOLD_MS` unexported. The
confirmation's `StatusBadge` is `aria-hidden` so a screen reader hears "Applied" once rather
than `"AppliedApplied"`.

**Round 2 — `clean: true`.** Each fix re-verified by repro on both surfaces, including attacks
that tried to re-inflate the count (already-removed bullet, blank `add`, no-op `replace` — all
unreachable) and a hand-audit of the edited `useCallback` dep lists for stale closures, since
this repo does not enforce `exhaustive-deps`. Confirmed a nonzero count can no longer coexist
with an empty section list by construction.

**Surviving nits, none blocking:** the blank-string-label guard hole is latent only (`roleLabel`
never returns `""`); the zero-write dismiss is silent by design (see *Known, deliberate
behaviour* above); and `role="status"` mounts with its content already present, which some
screen readers do not announce — a property of the original design, not of these fixes.

## Review round (Samhit21)

Approved with one inline nit, now fixed in the branch's single commit:

- **Collapse-timer dependency.** `SectionRewrite`'s two confirmation states passed
  `ApplyConfirmation` an inline `onCollapse` arrow, so its `EXIT_MS` timer re-armed on any
  re-render inside the ~150ms collapse window. Both now pass one stable `backToIdle`
  `useCallback` — symmetric with the whole-résumé surface's stable `onDismiss`.

Also cleared while revising: the second surviving nit above — the **arrow-key wrap-around** case
is now tested at both edges (ArrowRight from the last tab, ArrowLeft from the first),
parameterised over the description / no-description branches like the rest of `Tabs.test.tsx`.
The modulo in `onKeyDown` exists only for those two positions, and every prior keyboard test
started from the middle tab, so an off-by-one there was unreachable by the suite.

The flagged `ResumeRewriteProposed`/`RewriteReviewList` duplication and
`ReconstructedResume`'s unmemoized `captureUndo` arrow were both traced and left alone:
pre-existing, and no correctness impact.

### DIFF
--- a/src/components/features/ResumeRewrite.tsx
+++ b/src/components/features/ResumeRewrite.tsx
@@ -31,7 +31,7 @@
  */
 
 import { useState } from "react";
-import { Button, Dialog, ModelLoadProgress, TextAreaField } from "@design-system";
+import { Button, Dialog, InlineResult, ModelLoadProgress, TextAreaField } from "@design-system";
 import {
   labelForResumeRewrite,
   useResumeRewrite,
@@ -40,6 +40,11 @@ import {
 } from "../../hooks/useResumeRewrite.ts";
 import type { SectionInput, SectionOutcome } from "../../lib/webllm/rewrite-resume.ts";
 import type { PageTarget } from "../../lib/webllm/steering.ts";
+import {
+  ApplyConfirmation,
+  UndoBatchButton,
+  UNDO_HOLD_MS,
+} from "./ApplyConfirmation.tsx";
 import {
   ProposedPanel,
   type ResumeRewriteApply,
@@ -89,6 +94,8 @@ export function useResumeRewriteUi(
       <ResumeRewritePanel
         status={controller.status}
         onDismiss={controller.dismiss}
+        onApplied={controller.confirmApplied}
+        onUndo={controller.undoApplied}
         applyBySection={applyBySection}
       />
     );
@@ -228,10 +235,19 @@ function RewriteSteeringBox({
 export function ResumeRewritePanel({
   status,
   onDismiss,
+  onApplied,
+  onUndo,
   applyBySection,
 }: {
   status: ResumeRewriteStatus;
   onDismiss: () => void;
+  onApplied: (
+    count: number,
+    sections: readonly string[],
+    undo?: () => void,
+  ) => void;
+  /** Reverse the applied batch (issue 510). */
+  onUndo: () => void;
   applyBySection?: ResumeRewriteApply;
 }) {
   if (status.kind === "idle") return null;
@@ -269,10 +285,38 @@ export function ResumeRewritePanel({
       </div>
     );
   }
+  if (status.kind === "applied") {
+    return (
+      <InlineResult tone="success">
+        <ApplyConfirmation
+          count={status.count}
+          sections={status.sections}
+          onCollapse={onDismiss}
+          // Only a confirmation that actually hosts an Undo gets the longer
+          // hold — with no undo this stays on #508's 3s.
+          holdMs={status.undo ? UNDO_HOLD_MS : undefined}
+          action={status.undo && <UndoBatchButton onUndo={onUndo} />}
+        />
+      </InlineResult>
+    );
+  }
+  if (status.kind === "undone") {
+    return (
+      <InlineResult tone="success">
+        <ApplyConfirmation
+          verb="Reverted"
+          count={status.count}
+          sections={status.sections}
+          onCollapse={onDismiss}
+        />
+      </InlineResult>
+    );
+  }
   return (
     <ProposedPanel
       result={status.result}
       onDismiss={onDismiss}
+      onApplied={onApplied}
       applyBySection={applyBySection}
     />
   );
--- a/src/components/features/ResumeRewriteProposed.tsx
+++ b/src/components/features/ResumeRewriteProposed.tsx
@@ -59,10 +59,26 @@ interface ReviewSection {
 export function ProposedPanel({
   result,
   onDismiss,
+  onApplied,
   applyBySection,
 }: {
   result: ResumeRewriteResult;
+  /** Discard CTA only — Apply no longer calls this (#508; it calls
+   *  `onApplied` instead so the confirmation shows in place). */
   onDismiss: () => void;
+  /** Apply just committed its writes: the count of writes that actually
+   *  landed — NOT `acceptedCount`, which includes accepted-but-verbatim pairs
+   *  that resolve to no write — and the section labels that got one (#508).
+   *  Never called with a count of 0; a zero-write batch dismisses instead. The caller shows the
+   *  confirmation and holds the panel instead of dismissing synchronously.
+   *  `undo` reverses the whole batch (issue 510) — undefined when any written
+   *  section couldn't be snapshotted, so the control is never offered for a
+   *  partial revert. */
+  onApplied: (
+    count: number,
+    sections: readonly string[],
+    undo?: () => void,
+  ) => void;
   /** Per-section write-back handlers (#211 apply for the whole-résumé path).
    *  Absent → every section renders read-only (graceful fallback). */
   applyBySection?: ResumeRewriteApply;
@@ -102,6 +118,22 @@ export function ProposedPanel({
   );
 
   const onApply = useCallback(() => {
+    // Track which sections actually got a write — an accepted-but-unedited
+    // pair resolves to zero writes (resolveSectionWrites drops no-ops), so
+    // "touched" is the writes list, not the section's accepted-pair count.
+    const touchedSections: string[] = [];
+    // Undo is all-or-nothing across the batch (issue 510): a per-section
+    // snapshot is taken before that section's writes land, and the control is
+    // offered only if EVERY written section handed one back. Reversing four
+    // sections out of five would silently leave the résumé in a state the user
+    // never authored — worse than offering no undo at all.
+    const undos: (() => void)[] = [];
+    let reversible = true;
+    // The confirmation must report WRITES, not accepted pairs: accepting a
+    // bullet whose rewrite equals the original is a legitimate accept the
+    // Apply button counts, but it commits nothing. Announcing the accepted
+    // count would claim changes that never landed.
+    let writtenCount = 0;
     for (const sec of reviewSections) {
       const writes = resolveSectionWrites(
         sec.pairs,
@@ -109,14 +141,36 @@ export function ProposedPanel({
         review.decisions,
         review.edits,
       );
+      if (writes.length === 0) continue;
+      writtenCount += writes.length;
+      touchedSections.push(sec.label);
+      const undo = sec.apply.captureUndo?.(writes);
+      if (undo) undos.push(undo);
+      else reversible = false;
       for (const w of writes) {
         if (w.kind === "add") sec.apply.onAdd(w.text);
         else if (w.kind === "replace") sec.apply.onReplace(w.obsIndex, w.text);
         else sec.apply.onRemove(w.obsIndex);
       }
     }
-    onDismiss();
-  }, [reviewSections, review.decisions, review.edits, onDismiss]);
+    // An all-verbatim batch commits nothing — no confirmation at all, or the
+    // strip would announce "Applied N changes — " with an empty section list
+    // and (on the per-role path) a live Undo over an empty snapshot. Fall back
+    // to the pre-#508 behaviour: just drop the proposal.
+    if (writtenCount === 0) {
+      onDismiss();
+      return;
+    }
+    onApplied(
+      writtenCount,
+      touchedSections,
+      reversible && undos.length > 0
+        ? () => {
+            for (const undo of undos) undo();
+          }
+        : undefined,
+    );
+  }, [reviewSections, review.decisions, review.edits, onApplied, onDismiss]);
 
   const accepted = review.acceptedCount;
 
--- a/src/components/features/SectionRewrite.tsx
+++ b/src/components/features/SectionRewrite.tsx
@@ -54,8 +54,16 @@ import {
   alignBullets,
   type AlignedPair,
 } from "../../lib/rewrite-review/align-bullets.ts";
-import { resolveBulletActions } from "../../lib/rewrite-review/apply-accepted.ts";
+import {
+  resolveSectionWrites,
+  type ResolvedWrite,
+} from "../../lib/rewrite-review/apply-accepted.ts";
 import { useRewriteReview } from "../../hooks/useRewriteReview.ts";
+import {
+  ApplyConfirmation,
+  UndoBatchButton,
+  UNDO_HOLD_MS,
+} from "./ApplyConfirmation.tsx";
 import { RewriteReviewList } from "./RewriteReviewList.tsx";
 
 /**
@@ -74,6 +82,14 @@ export interface SectionRewriteApply {
   onRemove: (obsIndex: number) => void;
   /** Append a new bullet to this role (→ addBullet on the role's entry key). */
   onAdd: (text: string) => void;
+  /**
+   * Snapshot the pre-apply state of the slots `writes` will touch and return
+   * the thunk that restores them (issue 510). Called BEFORE the write loop.
+   * Absent → this surface offers no Undo at all: a partial undo that restores
+   * some sections and not others is worse than none, so both apply paths only
+   * surface the control when EVERY section they write has this wired.
+   */
+  captureUndo?: (writes: readonly ResolvedWrite[]) => () => void;
 }
 
 /** Stable empty alignment so the review hook doesn't reset on every idle render. */
@@ -147,6 +163,23 @@ export type Status =
        *  bullet underneath a previous rewrite. */
       snapshot: readonly string[];
     }
+  | {
+      /** Apply just committed its writes (#508) — shown in place of the
+       *  review surface for a few seconds instead of dismissing silently. */
+      kind: "applied";
+      count: number;
+      sections: readonly string[];
+      /** Reverses the whole batch (issue 510). Absent when the edit model
+       *  didn't hand back a capture — then no Undo is offered. */
+      undo?: () => void;
+    }
+  | {
+      /** Undo just ran (issue 510) — acknowledged in the same strip rather
+       *  than reverting silently. One-shot: there is no re-apply. */
+      kind: "undone";
+      count: number;
+      sections: readonly string[];
+    }
   | { kind: "error"; message: string };
 
 function bulletsEqual(a: readonly string[], b: readonly string[]): boolean {
@@ -167,6 +200,10 @@ function bulletsEqual(a: readonly string[], b: readonly string[]): boolean {
 export function useSectionRewrite(
   bullets: readonly string[],
   apply?: SectionRewriteApply,
+  /** Display label for the "Applied N changes — <label>" confirmation
+   *  (#508) — the role's `roleLabel(group.experience)`. Falls back to a
+   *  generic name when the caller has none to give. */
+  sectionLabel = "This section",
 ): SectionRewriteParts {
   const [capability, setCapability] = useState<WebGpuCapability | null>(null);
   const [status, setStatus] = useState<Status>({ kind: "idle" });
@@ -286,26 +323,60 @@ export function useSectionRewrite(
     setCopied(false);
   }, []);
 
+  // `ApplyConfirmation` re-arms its EXIT_MS collapse timer whenever `onCollapse`
+  // changes identity, so both confirmation states hand it this stable callback
+  // rather than a fresh inline arrow per render — symmetric with the
+  // whole-résumé surface, which already passes a stable `onDismiss`.
+  const backToIdle = useCallback(() => {
+    setStatus({ kind: "idle" });
+  }, []);
+
   // Apply accepted decisions back into the reconstructed résumé via the edit
   // model, then dismiss the proposal. Each action's `originalIndex` is a
   // position in the trimmed snapshot, mapped to its BulletObservation.index
   // through `keptObsIndices`.
   const onApply = useCallback(() => {
     if (!apply) return;
-    const actions = resolveBulletActions(pairs, review.decisions, review.edits);
-    for (const action of actions) {
-      if (action.kind === "add") {
-        apply.onAdd(action.text);
-        continue;
-      }
-      const obsIndex = keptObsIndices[action.originalIndex];
-      if (obsIndex === undefined || obsIndex < 0) continue;
-      if (action.kind === "replace") apply.onReplace(obsIndex, action.text);
-      else apply.onRemove(obsIndex);
+    const writes = resolveSectionWrites(
+      pairs,
+      keptObsIndices,
+      review.decisions,
+      review.edits,
+    );
+    // An all-verbatim batch resolves to zero writes (resolveSectionWrites
+    // drops no-op accepts). Confirming it would announce changes that never
+    // landed AND offer an Undo over an empty snapshot — a control that
+    // reverses nothing. Bail to idle instead, mirroring the whole-résumé
+    // path's zero-write dismiss.
+    if (writes.length === 0) {
+      setStatus({ kind: "idle" });
+      setCopied(false);
+      return;
     }
-    setStatus({ kind: "idle" });
+    // Snapshot BEFORE the loop — once a write lands the prior value is gone
+    // (issue 510).
+    const undo = apply.captureUndo?.(writes);
+    for (const write of writes) {
+      if (write.kind === "add") apply.onAdd(write.text);
+      else if (write.kind === "replace") apply.onReplace(write.obsIndex, write.text);
+      else apply.onRemove(write.obsIndex);
+    }
+    setStatus({
+      kind: "applied",
+      // Writes committed, not pairs accepted — see the zero-write bail above.
+      count: writes.length,
+      sections: [sectionLabel],
+      undo,
+    });
     setCopied(false);
-  }, [apply, pairs, review.decisions, review.edits, keptObsIndices]);
+  }, [
+    apply,
+    pairs,
+    review.decisions,
+    review.edits,
+    keptObsIndices,
+    sectionLabel,
+  ]);
 
   const onCopyAll = useCallback(async () => {
     if (status.kind !== "proposed") return;
@@ -385,6 +456,44 @@ export function useSectionRewrite(
               onReject={onReject}
             />
           ))}
+
+        {status.kind === "applied" && (
+          <InlineResult tone="success">
+            <ApplyConfirmation
+              count={status.count}
+              sections={status.sections}
+              onCollapse={backToIdle}
+              // Only a confirmation that actually hosts an Undo gets the longer
+              // hold — with no undo this stays on #508's 3s.
+              holdMs={status.undo ? UNDO_HOLD_MS : undefined}
+              action={
+                status.undo && (
+                  <UndoBatchButton
+                    onUndo={() => {
+                      status.undo?.();
+                      setStatus({
+                        kind: "undone",
+                        count: status.count,
+                        sections: status.sections,
+                      });
+                    }}
+                  />
+                )
+              }
+            />
+          </InlineResult>
+        )}
+
+        {status.kind === "undone" && (
+          <InlineResult tone="success">
+            <ApplyConfirmation
+              verb="Reverted"
+              count={status.count}
+              sections={status.sections}
+              onCollapse={backToIdle}
+            />
+          </InlineResult>
+        )}
 
         {status.kind === "error" && (
           <p role="alert" className="text-[11px] text-feedback-error-text">
--- a/src/design-system/shared/Tabs.tsx
+++ b/src/design-system/shared/Tabs.tsx
@@ -136,9 +136,28 @@ interface TabProps {
    * isn't conveyed by colour alone.
    */
   warn?: boolean;
+  /**
+   * Optional one-line subtitle rendered under the label (issue #519), so a
+   * tab's purpose is legible without clicking it. Omitting this prop
+   * reproduces today's single-line rendering exactly — Tabs is a shared
+   * primitive and a future consumer must not be forced into two-line tabs.
+   *
+   * Accessibility: the subtitle is plain visible text nested inside the tab
+   * `<button>`, so it becomes part of the tab's accessible NAME via the
+   * standard accessible-name-from-content algorithm — the same mechanism
+   * `warn`'s `sr-only` addendum already uses to extend the name. No separate
+   * `aria-describedby` wiring, so there's nothing to keep in sync if the
+   * label or subtitle changes.
+   *
+   * Responsive: hidden below the `sm` breakpoint so a narrow viewport (375px)
+   * degrades to today's single-line tab row instead of widening it — the
+   * existing `overflow-x-auto` track only has to absorb one line at that
+   * width, same as before this prop existed.
+   */
+  description?: string;
 }
 
-export function Tab({ id, children, count, warn }: TabProps) {
+export function Tab({ id, children, count, warn, description }: TabProps) {
   const { value, onValueChange, baseId } = useTabsContext("Tab");
   const isActive = value === id;
   // Selection carries a real surface (bg-surface-card, matching the panel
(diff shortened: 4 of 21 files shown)


src/components/features/SectionRewrite.tsx
```diff
@@ -386,6 +449,44 @@ export function useSectionRewrite(
             />
           ))}
 
+        {status.kind === "applied" && (
+          <InlineResult tone="success">
+            <ApplyConfirmation
+              count={status.count}
+              sections={status.sections}
+              onCollapse={() => setStatus({ kind: "idle" })}
```
Samhit21: **Nit (non-blocking).** `ApplyConfirmation`'s collapse effect depends on `onCollapse`, but this per-role call passes an inline arrow that churns identity every render — so while `collapsing`, the `EXIT_MS` timer re-arms on any re-render in that ~150ms window. Harmless in practice (the leaf doesn't re-render then) and the whole-résumé path already passes a stable `onDismiss`; wrapping this in `useCallback` (or hoisting a stable `backToIdle`) would make it symmetric. Same for the `undone` block at line 485.

Samhit21 (dismissed):
Reviewed the accumulated diff (21 files, +1719/-61) at `99a45e5`. This is careful, well-tested work — the four sub-issues are cleanly separated and the three round-1 bugs in the PR's own adversarial pass are genuinely fixed. Verifying against the code, I concur with the verdict: no blockers.

**offlinecv gates**
- **Fixture PII** — N/A, no fixtures touched.
- **Design-system / reuse** — clean. `ApplyConfirmation` composes `StatusBadge`/`Button` from `@design-system` rather than minting a toast primitive; the reuse analysis is in the body.
- **Style tokens** — clean (semantic tokens throughout; the grep hits were `#508`-style issue refs, not hex).
- **Command/script files** — none touched.
- **fallow** — the surfaced complexity findings (`useSectionRewrite`, `onApply`, `ModelSelector`) are pre-existing functions in touched files, report-only; nothing this diff introduces is over-threshold once `Tabs.onKeyDown` got its tests.

**What I verified beyond the generic pass**
- **Undo snapshot** (`undo-batch.ts`) — pre-state capture is correct: `undefined` override restores by deleting the key; `restore` excludes already-removed indices so undo can't un-remove a user's prior drop; all three writers are idempotent (StrictMode-safe).
- **Count = writes, not accepts** — both surfaces accumulate `writes.length`; `acceptedCount` only labels the button. Zero-write batches bail before entering the confirmation state (no dangling "Applied N — ").
- **entryKey wiring** — the undo snapshot's `addedEntryKey` matches the bucket `onAdd` appends to on both surfaces (`added ? added.id : parsedEntryKey(...)`), so an added-bullet undo targets the right array.
- **Batch-undo across sections** — captures read live state via refs before any re-render, and sections touch disjoint slots, so each section's pre-batch snapshot is correct regardless of apply order.
- **`captureBulletUndo` stable identity** — refs (not the render closure) keep it stable as a memo dep, so in-flight accept/reject decisions don't reset.
- **#519 Tabs** — additive `description` prop preserves the single-line branch exactly; the quality subtitle is correctly gated on `analysis.isAvailable`.

65 tests in the changed files pass locally; typecheck + `eslint src` clean.

## Nits (non-blocking)
- The mount-time `role="status"` announce caveat you already noted stands. One more inline below, on the collapse-timer dependency.

---

src/components/features/SectionRewrite.tsx
s-annam: Fixed in `9d346fa` (the branch's single commit, force-pushed).

Hoisted a stable `backToIdle` `useCallback` next to `onReject` and both confirmation states —
`applied` (line 465) and `undone` (line 493) — now pass it instead of an inline arrow, so
`EXIT_MS` can't re-arm on an unrelated re-render inside the collapse window. Symmetric with the
whole-résumé surface's stable `onDismiss`, as you noted.

While in here I also cleared the second surviving nit from the PR body's round-2 list — the
arrow-key **wrap-around** case is now tested at both edges (ArrowRight from the last tab,
ArrowLeft from the first), parameterised over the description / no-description branches like the
rest of `Tabs.test.tsx`. `renderTabList` takes an optional `value` so the selection can start at
an edge; every prior keyboard test started from the middle tab, which is exactly why the modulo
in `onKeyDown` — the only reason it's a modulo — was unreachable by the suite.

Gates on the amended commit: `npm run typecheck` clean, `npx eslint src` clean, full
`npx vitest run` green (184 files passed | 6 skipped; 2561 tests passed | 9 skipped). PR body
updated: the "arrow-key wrap-around is untested" line is gone (it would have been false in
`main`), and a `Review revisions` row added to `## Provenance`.

src/components/features/ResumeRewrite.tsx
```diff
@@ -269,10 +285,38 @@ export function ResumeRewritePanel({
       </div>
     );
   }
+  if (status.kind === "applied") {
```
Samhit21, line 288: **Secondary — the "Reverted" strip reuses the applied strip's component instance, so this surface does not behave like the per-role one.**

`ResumeRewritePanel` returns `<InlineResult><ApplyConfirmation/></InlineResult>` from both the `applied` and the `undone` branch. Same element type, same position → React reconciles rather than remounts, and `ApplyConfirmation` keeps its `visible` / `collapsing` state across the transition. `SectionRewrite` renders its two branches as separate JSX children, so it *does* remount. That splits the "identical copy, timing, and motion" claim in the PR body.

Reproduced with a scratch jsdom test against `ResumeRewritePanel`:

- applied → undone hands back the **same DOM node** (`applied === undone` is `true`), so `visible` is already `true` and the Reverted strip never plays its enter fade.
- Worse case: let the 12s `UNDO_HOLD_MS` elapse so `collapsing` flips true, *then* transition to `undone` — the user clicking Undo inside the ~150ms exit window, while the row still has height and opacity. The Reverted strip renders already collapsed (`grid-rows-[0fr] opacity-0`) and the pending exit timer fires `onDismiss` immediately, because that effect's deps `[collapsing, onCollapse]` are unchanged so it never re-arms. The revert lands with **zero acknowledgment** — the exact silent-mutation failure #510 exists to remove.

One line fixes both and makes the two surfaces genuinely symmetric:

```diff
         <ApplyConfirmation
+          key={status.kind}
           count={status.count}
```

(on both the `applied` and the `undone` branch, or hoist it to the shared `InlineResult`.)

src/design-system/shared/Tabs.tsx
```diff
@@ -176,18 +195,44 @@ export function Tab({ id, children, count, warn }: TabProps) {
...
+            {warn && (
+              <>
+                <span aria-hidden="true" className="text-feedback-warning-text">
+                  {"⚠︎"}
+                </span>
+                <span className="sr-only"> (setup needed)</span>
+              </>
+            )}
+          </span>
+          <span className="hidden text-xs font-normal text-content-secondary sm:block">
```
Samhit21, line 232: **Secondary — the tab's accessible name now varies by viewport.**

The docblock is right that the subtitle joins the accessible name via name-from-content, and right that `hidden sm:block` keeps 375px single-line — but those two interact. `hidden` is `display:none`, which is excluded from the accessibility tree, so a screen-reader user below `sm` hears "Reconstructed resume" and above `sm` hears "Reconstructed resume what a parser pulled out — edit it here". Same tab, two names, depending on window width. It also makes voice-control target names breakpoint-dependent.

Either acknowledge this in the docblock, or move the subtitle to `aria-describedby` — stable name, description read as a description, which is also the conventional mapping for what this prop *is*. `Tabs.test.tsx`'s "suppresses the subtitle below the sm breakpoint" asserts the class only, so nothing pins the a11y consequence either way.

src/components/features/ResumeRewriteProposed.tsx
```diff
@@ -102,21 +118,59 @@ export function ProposedPanel({
...
     for (const sec of reviewSections) {
       const writes = resolveSectionWrites(
         sec.pairs,
         sec.apply.obsIndices,
         review.decisions,
         review.edits,
       );
+      if (writes.length === 0) continue;
+      writtenCount += writes.length;
+      touchedSections.push(sec.label);
```
Samhit21, line 146: **Nit — `touchedSections` doesn't dedupe labels.**

`sec.label` is `roleLabel(exp)`, which returns `"Untitled role"` for any role that parsed with neither title nor company. On a badly-parsed résumé — this app's core use case — two such roles both getting a write render `"Applied 3 changes — Untitled role, Untitled role"`. Same for two genuine stints at one company under one title. `[...new Set(touchedSections)]` at the callsite is enough.

Samhit21 (approved):
Approving. The two rounds of adversarial review did their job — the three defects they caught (accepted-pairs vs writes-landed, the all-verbatim dangling em dash, the empty-snapshot Undo) were the real bugs in this diff, and I could not re-inflate the count or reach a false confirmation from any angle I tried. `undo-batch.ts` earns its snapshot-not-reverse-diff argument: the `remove`-into-a-set and `add`-drops-blank asymmetries genuinely are unrecoverable from the write alone.

**Gates.** Fixture-PII **N/A** — no fixture touched, confirmed against `git diff --name-only … tests/fixtures/**`. Design-system reuse **pass** — no raw `<button>`/dialog/input in feature code; the one new file, `ApplyConfirmation.tsx`, carries a written reuse analysis and lands at 147 LOC. Style tokens **pass** — the only `#`-hits are issue numbers in comments; no palette class, no manual `dark:`. `typecheck` / `eslint src` / full `vitest` **green locally** (184 files, 2561 tests), CI `verify` + `fallow` green. `fallow audit --base origin/main` raises no dead export; its 6 complexity findings are report-only in `verify` — one of them is new to this diff, see Nits.

Nothing blocking. Three Secondary and two Nits below; the ones with a line to land on are left inline.

## Secondary

**1. The whole-résumé "Reverted" strip reuses the applied strip's component instance** — inline at `src/components/features/ResumeRewrite.tsx:288`. Reproduced; one-line `key` fix.

**2. The tab's accessible name now varies by viewport** — inline at `src/design-system/shared/Tabs.tsx:232`.

**3. No integration test covers the whole-résumé applied → undone path.**

`ResumeRewriteProposed.test.tsx` pins that `onApplied` receives a working thunk, and `ApplyConfirmation.test.tsx` pins the strip in isolation — but nothing renders `ResumeRewritePanel` in the `applied` state, clicks Undo, and asserts the Reverted strip. `useResumeRewrite.undoApplied` has no test at all. That gap is precisely where finding 1 lives; a test at that seam would have caught it. The per-role surface has the same gap.

## Nits

**4. `touchedSections` doesn't dedupe labels** — inline at `src/components/features/ResumeRewriteProposed.tsx:146`.

**5. `SectionRewrite.tsx` grew to 610 LOC and `useSectionRewrite` to 308.**

CLAUDE.md names this file as known debt with an explicit "prefer extracting your change into a new sibling over growing the file further". You did extract `ApplyConfirmation`, which is the right call and I'm not asking for more here — but `fallow` now flags `useSectionRewrite` at 18 cyclomatic / 22 cognitive / 308 lines, and that finding is new to this diff. Worth a follow-up issue to pull the status-branch rendering out of the hook rather than letting the next feature land on top of it.

---

PR #537, opened 2026-07-22
feat(app): reframe hero around resume custody, not device runtime

s-annam:
## Summary

Rewrites the landing hero so the privacy claim is about **custody, not runtime**. The old line — *"Your whole job search — on your device by default"* — overclaimed (the app makes network calls: job feeds + PostHog analytics) and hedged with *"by default."*

- **Headline:** *Your whole job search — in your browser.* (a place claim, not a runtime claim)
- **Subhead:** keeps *free, open-source*; states the absolute, accurate promise — the **resume** never leaves the browser (guaranteed by `keywords.ts`, which egresses only a derived keyword string, never the resume/PDF/queries).
- **Propagates the noun to `CapabilityStrip`** (review feedback): the rail said *"Your resume stays on your **device**"* and renders as `PageShell`'s `chips` a few inches under the hero, so the idle screen read back two nouns for one promise. It now says *"in your **browser**"* — also the stronger claim, since another app on the same device can't see the resume either. `PageShell`'s footer already used *"browser tab."*
- Rewrites the load-bearing comment that used to defend *"by default"* so a future edit doesn't collapse headline + subhead into a false blanket *"everything stays on your device"* — and records the **trip-wire** that would falsify the subhead: it is absolute only while `keywords.ts` is the sole resume-derived egress, so #320's BYOK path (sends resume text to a cloud provider) must rescope the subhead in the PR that ships it.

Copy-only + comment; no logic change.

## Test plan

- [x] `npm run test` green (2526 passed)
- [x] `npm run verify` green (pre-push gate)

## Provenance

Copy iteration + implementation: Claude Opus 4.8 (1M context), high effort — guided by the `ui-ux-pro-max` landing-pattern search (minimal single-column / benefit-led hero).
Review revisions: Claude Opus 4.8 (1M context), medium effort.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -115,27 +115,41 @@ export default function App() {
             // recruiter-agent context) moves to the quiet block below the drop
             // zone so the hero isn't three competing messages.
             //
-            // The headline's "by default" is NOT hedging for its own sake and
-            // must not be dropped: the job-search lane egresses keywords on an
-            // explicit click (see FindJobsPanel), a build with
+            // The headline claims custody, NOT runtime — the distinction is
+            // load-bearing and must survive edits. "in your browser" scopes the
+            // claim to a place; it is NOT an absolute "runs on your device"
+            // claim, which would be false: the job-search lane egresses
+            // keywords on an explicit click (see FindJobsPanel), a build with
             // VITE_POSTHOG_KEY set (the hosted one) ships analytics, and the
             // BYOK provider path (#320, not in the tree yet — see
-            // CapabilityStrip) will be real cloud egress once it lands. The
-            // first two are true today and already load-bearing on their own.
-            // An unqualified "on your device"
-            // would be the only absolute privacy claim in the app — every
-            // other surface is already scoped the same way (PageShell's
-            // footer, CapabilityStrip's rail).
+            // CapabilityStrip) will be real cloud egress once it lands.
+            //
+            // "Browser" is also the noun the sibling surfaces use, and the
+            // three are read together on the idle screen: CapabilityStrip's
+            // rail ("Your resume stays in your browser") renders as PageShell's
+            // `chips` a few inches below, and PageShell's footer states its own
+            // narrower, hedged claim about a different object ("your PDF stays
+            // in this browser tab by default"). Same noun, three different
+            // sentences — if you change the noun in one, change it in all.
+            //
+            // What is absolute is the subhead's promise — the *resume* never
+            // leaves the browser — guaranteed by keywords.ts (only a derived
+            // keyword string egresses, never the resume/PDF/queries). That is
+            // absolute only while keywords.ts is the sole resume-derived
+            // egress: if #320's BYOK path lands, it sends resume text to a
+            // cloud provider and this subhead becomes false, so it must be
+            // rescoped in the same PR that ships it. Keep the headline about
+            // place and the subhead about the resume; do not merge them into a
+            // blanket "everything stays on your device."
             <Card className="flex flex-col items-center gap-5 bg-surface-card-warm">
               <div className="flex max-w-2xl flex-col gap-4 text-center">
                 <h2 className="text-balance text-2xl font-semibold leading-snug tracking-tight text-content-primary sm:text-3xl">
-                  Your whole job search — on your device by default.
+                  Your whole job search — in your browser.
                 </h2>
                 <p className="text-pretty text-base font-medium text-content-secondary sm:text-lg">
-                  OfflineCV is a free, open-source job-search workbench —
-                  read your resume, fix it, and match it against real
-                  postings, all in your browser. Nothing to install, no
-                  account to create.
+                  OfflineCV is a free, open-source job-search workbench. Drop a
+                  PDF, score it, fix it, match a JD, and find jobs — all without
+                  your resume leaving your browser.
                 </p>
                 <p className="text-xs text-content-muted">
                   Built in response to a hiring process job seekers say they
--- a/src/design-system/shared/CapabilityStrip.tsx
+++ b/src/design-system/shared/CapabilityStrip.tsx
@@ -20,7 +20,7 @@
  * carries the rail statement, matching how the old chip row phrased itself.
  *
  * Privacy scope (binding, epic #511): the rail claims custody of the RESUME —
- * "Your resume stays on your device" — deliberately not "nothing leaves" or
+ * "Your resume stays in your browser" — deliberately not "nothing leaves" or
  * "runs entirely on your device". Those would be false: the job-search lane
  * this same card advertises does `fetch()` three third-party feeds, and a
  * build with VITE_POSTHOG_KEY set ships analytics the user never opted into.
@@ -30,9 +30,16 @@
  * or if a cloud LLM path lands (none exists today — there is no BYOK provider
  * in the tree despite older docs implying one), this line becomes a lie and
  * must change with it. Per-feature exceptions stay stated where the user
- * meets them (`FindJobsPanel`), not restated here. `PageShell`'s footer makes
- * the equivalent scoped claim in its own words ("your PDF stays in this
- * browser tab by default").
+ * meets them (`FindJobsPanel`), not restated here.
+ *
+ * The noun is "browser", not "device", and that is deliberate (#537): this
+ * rail renders a few inches below `App`'s hero, which says "in your browser"
+ * in the headline and "leaving your browser" in the subhead, so a visitor
+ * reads all three at once. "Device" also understates the guarantee — another
+ * app on the same device cannot read the resume either. If you change the
+ * noun here, change it in `App.tsx`'s hero in the same commit. `PageShell`'s
+ * footer states a narrower, hedged claim about a different object ("your PDF
+ * stays in this browser tab by default") — same noun, not the same sentence.
  */
 
 import { Chip } from "../primitives/Chip.tsx";
@@ -81,7 +88,7 @@ export function CapabilityStrip() {
         ))}
       </div>
       <div className="flex flex-wrap items-center gap-2 border-t border-border-light pt-3">
-        <Chip icon={<LockIcon />}>Your resume stays on your device</Chip>
+        <Chip icon={<LockIcon />}>Your resume stays in your browser</Chip>
       </div>
     </Card>
   );


src/App.tsx
```diff
@@ -115,27 +115,29 @@ export default function App() {
...
-            // BYOK provider path (#320, not in the tree yet — see
-            // CapabilityStrip) will be real cloud egress once it lands. The
-            // first two are true today and already load-bearing on their own.
-            // An unqualified "on your device"
-            // would be the only absolute privacy claim in the app — every
-            // other surface is already scoped the same way (PageShell's
-            // footer, CapabilityStrip's rail).
+            // The headline claims custody, NOT runtime — the distinction is
+            // load-bearing and must survive edits. "in your browser" is the
+            // same scoped framing every other surface already uses (PageShell's
```
Samhit21: [Suggestion]: The parity claim in this comment doesn't hold — the sibling surfaces are hedged and use a different noun, so the hero is now the strongest privacy claim on the page, not the same one.

The comment says "in your browser" is "the same scoped framing every other surface already uses (PageShell's footer, CapabilityStrip's rail)". Checking both:

- `PageShell.tsx:88` — "Your PDF stays in this browser tab **by default** and is never used to train AI."
- `CapabilityStrip.tsx:84` — "Your resume stays on your **device**"

Neither is unhedged, and the rail says *device* where the hero now says *browser*. `CapabilityStrip` is passed as `chips` to the `DropZone` (App.tsx:99), so on the idle screen a visitor reads "in your browser" in the headline, "leaving your browser" in the subhead, and "stays on your device" in the chip a few inches below — three phrasings of one promise, two nouns.

Worth picking one noun and propagating. If "browser" is right (it is more precise — "device" is technically weaker, since another app on the same device can't see it either), then `CapabilityStrip.tsx:84` and its docblock at line 23 should move too, in this PR or a stated follow-up. Otherwise this comment is asserting a consistency that a reader will find contradicted on the same screen.

src/App.tsx
```diff
@@ -115,27 +115,29 @@ export default function App() {
...
+                  Your whole job search — in your browser.
                 </h2>
                 <p className="text-pretty text-base font-medium text-content-secondary sm:text-lg">
-                  OfflineCV is a free, open-source job-search workbench —
-                  read your resume, fix it, and match it against real
-                  postings, all in your browser. Nothing to install, no
-                  account to create.
+                  OfflineCV is a free, open-source job-search workbench. Drop a
+                  PDF, score it, fix it, match a JD, and find jobs — all without
+                  your resume leaving your browser.
```
Samhit21, line 152: [Suggestion]: The subhead is now the app's one absolute, unhedged privacy claim, but nothing encodes the condition that would falsify it.

The comment above deliberately makes this the absolute claim ("What is absolute is the subhead's promise — the *resume* never leaves the browser") and instructs future editors to keep it that way. That's correct **today** — `keywords.ts` is the sole resume-derived egress and it sends a keyword string. But the same comment lists the BYOK provider path (#320) as "real cloud egress once it lands," and a BYOK rewrite sends **resume text** — the exact thing this sentence promises never leaves. When #320 lands, this line becomes false, and the comment as written tells that future author to preserve it.

`CapabilityStrip.tsx:28-32` already solves this for its own weaker claim, with an explicit trip-wire: "That module is the invariant this copy depends on; if it ever sends resume text, or if a cloud LLM path lands …, this line becomes a lie and must change with it."

Adding the same sentence here would make the guard-rail complete — something like: *"This is absolute only while `keywords.ts` is the sole resume-derived egress; if #320's BYOK path lands, this subhead must be rescoped before it ships."*

Samhit21 (commented):
## PR Review: feat(app): reframe hero around resume custody, not device runtime

### Summary

The core move is right and well-argued: "custody, not runtime" is the honest framing, dropping "by default" in favor of a claim that is actually absolute is a net improvement in both accuracy and confidence, and the subhead's promise is genuinely backed by `keywords.ts` (verified — a derived keyword string is the only resume-derived egress, and `trackParseFailed` explicitly refuses to forward pdfjs error strings for the same reason). Also checked and fine: dropping "Nothing to install, no account to create" loses nothing, since App.tsx:211 already carries "needs no account or email" in the quiet block.

Two findings, both about how the new claim sits against the surfaces around it rather than the copy itself. Neither is merge-blocking.

### Findings

1. `App.tsx:120` — the comment claims parity with PageShell's footer and CapabilityStrip's rail, but the footer is still hedged ("by default") and the rail says "device" where the hero says "browser". The rail renders on the same idle screen (App.tsx:99), so the divergence is visible, not theoretical.
2. `App.tsx:140` — the subhead is now the app's only absolute privacy claim, and the comment instructs future editors to keep it absolute, but nothing names the condition (#320's BYOK path, which egresses resume text) that would falsify it. `CapabilityStrip.tsx:28-32` already encodes exactly that trip-wire for its weaker claim.

### Minor, not worth an inline

The subhead now lists five lanes ("Drop a PDF, score it, fix it, match a JD, and find jobs") directly above a CapabilityStrip whose stated job (#517) is to present those lanes. Not a regression — the old subhead listed three — but the hero comment at 111-116 describes a deliberate consolidation to "one angle," and this pushes the other way.

### Verdict

**Action: COMMENT.**

---

src/App.tsx
s-annam: Fixed in `b286034`. You were right on both counts — the parity claim was false and the two-noun screen was real.

Picked **browser**, for your reason: it is the stronger claim, not just the more precise one (another app on the same device cannot read the resume either). Propagated in this PR rather than deferring:

- `CapabilityStrip.tsx:84` — `Your resume stays on your device` → `Your resume stays in your browser`
- `CapabilityStrip.tsx:22-23` docblock quotes the new string, and now carries the reason the noun is `browser` plus a "change it here, change it in `App.tsx`'s hero in the same commit" instruction.
- `App.tsx` — dropped the false parity sentence. It no longer claims the sibling surfaces use "the same scoped framing"; it names them accurately instead: the rail is the same noun in a different sentence, and `PageShell`'s footer is a **narrower, hedged claim about a different object** ("your PDF stays in this browser tab by default").

One noun on the idle screen now, three sentences, and the drift guard is on both files.

src/App.tsx
s-annam: Fixed in `b286034`. Correct catch — the comment made the subhead absolute and then told a future author to keep it that way, with nothing naming the condition that would break it.

Added the trip-wire, in the shape you suggested and matching `CapabilityStrip.tsx:28-32`'s:

> That is absolute only while keywords.ts is the sole resume-derived egress: if #320's BYOK path lands, it sends resume text to a cloud provider and this subhead becomes false, so it must be rescoped in the same PR that ships it.

It sits directly after the "what is absolute" sentence, so the condition and the claim are read together rather than the guard-rail living somewhere a future editor may not reach. The falsifier is also named concretely (BYOK sends **resume text** — the exact object the sentence protects), not just "a cloud path."

src/App.tsx
```diff
@@ -115,27 +115,41 @@ export default function App() {
...
-            // An unqualified "on your device"
-            // would be the only absolute privacy claim in the app — every
-            // other surface is already scoped the same way (PageShell's
-            // footer, CapabilityStrip's rail).
+            // CapabilityStrip) will be real cloud egress once it lands.
+            //
+            // "Browser" is also the noun the sibling surfaces use, and the
+            // three are read together on the idle screen: CapabilityStrip's
+            // rail ("Your resume stays in your browser") renders as PageShell's
+            // `chips` a few inches below, and PageShell's footer states its own
```
Samhit21, line 130: **Nit — the rail renders *above* the hero, not below it.**

`App.tsx:99` passes `chips={<CapabilityStrip />}` to `PageShell`, and `PageShell.tsx:82` renders `{chips}` **inside `<header>`**, before `{children}` — and the hero `Card` is in `children`. So top-to-bottom the idle screen is: header (logo, subtitle, CapabilityStrip rail) → hero card → drop zone. The rail is a few inches *above* the hero.

The argument this sentence supports is untouched — the three sentences genuinely are read together on one screen, which is the whole point — so this is only the direction word. But given this PR exists to fix a comment that asserted something a reader would find contradicted on screen, it's worth getting right. Same wording appears in `CapabilityStrip.tsx:36` with the same inversion.

"renders as PageShell's `chips` just above" would do it.

src/design-system/shared/CapabilityStrip.tsx
```diff
@@ -30,9 +30,16 @@
  * or if a cloud LLM path lands (none exists today — there is no BYOK provider
  * in the tree despite older docs implying one), this line becomes a lie and
  * must change with it. Per-feature exceptions stay stated where the user
- * meets them (`FindJobsPanel`), not restated here. `PageShell`'s footer makes
- * the equivalent scoped claim in its own words ("your PDF stays in this
- * browser tab by default").
+ * meets them (`FindJobsPanel`), not restated here.
+ *
+ * The noun is "browser", not "device", and that is deliberate (#537): this
+ * rail renders a few inches below `App`'s hero, which says "in your browser"
```
Samhit21, line 36: **Nit — same inversion as `App.tsx:130`.** This rail renders as `PageShell`'s `chips`, inside `<header>` (`PageShell.tsx:82`), which comes before `{children}` where `App`'s hero lives — so it is a few inches *above* the hero, not below. Worth fixing in the same commit as the `App.tsx` one, since the two comments are deliberately kept in sync.

Samhit21 (approved):
Re-review at `b286034`. Both findings from my last pass are properly fixed, and the fixes are better than what I asked for.

**Finding 1 (false parity claim / two nouns on one screen)** — resolved by propagating rather than deferring. `CapabilityStrip.tsx:88` now reads "Your resume stays in your browser", the docblock quotes the new string and carries the change-both-together instruction, and `App.tsx` no longer claims the siblings use "the same scoped framing" — it names each accurately instead. I verified `PageShell.tsx:88` still reads exactly "Your PDF stays in this browser tab by default…", so the comment's characterisation of it as a *narrower, hedged claim about a different object* is accurate. One noun on the idle screen now.

**Finding 2 (missing trip-wire on the new absolute claim)** — resolved, and it names the falsifier concretely (BYOK sends *resume text*, the exact object the sentence protects) rather than gesturing at "a cloud path". Sits directly after the "what is absolute" sentence, which is the right place for it.

**Re-verified the load-bearing claim.** The subhead now promises the resume never leaves the browser, so I re-checked the whole egress surface rather than trusting the last pass: every `fetch()` in `src/lib` is the three keyword feeds (`remotive`/`jobicy`/`arbeitnow`, which send only `searchPhrase()`'s derived keyword string — `providers/keywords.ts` is still the sole resume-derived egress and still says so), `jd-match/fetch-jd.ts` (a URL the *user* pastes, nothing resume-derived), `version.ts` (own origin), and the two Poppins font loads (own origin). BYOK is still forward-reference-only — every `BYOK` hit in `src/` is a comment about #320, no keyed adapter in the tree. The claim holds today and the trip-wire covers the day it stops holding.

**Gates.** Fixture-PII, design-system reuse, and command-file all **N/A** (two files, copy + comments, no fixture / no new component / no script). Style tokens **pass**. `typecheck`, `eslint src`, and `vitest run src/design-system src/components` (30 files, 202 tests) green locally; CI `verify` + `fallow` green.

One Nit inline, no blockers.

## Nit

**The new "a few inches below" claim points the wrong way** — inline on both files. Non-blocking, and the argument it supports is unaffected.

---

PR #551, opened 2026-07-23
fix(parser): round-trip the truncated one-line experience header (#436)

### LINKED ISSUE #436: Parser: round-trip the one-line experience header (title/company disambiguation without the two-line structural signal)

rohithgollapalli:
## Summary

Fixes the remaining (TRUNCATION) root of #436. A too-wide one-line export header (`Title · Company, Location · Team`, date flush-right) wrapped its org tail onto the row below, stranding or gluing the wrapped words to the date, so the company re-parsed truncated (`Danggeun Pay Inc. (KarrotPay)` → `(KarrotPay)`, `R.O.K Cyber Command, MND` → `MND`). Three coordinated parts:

- **render — collision-gated date-column reserve** (`render-ats-pdf.ts`): wrap at full width first, re-wrap reserving the flush-right date column only when the first line actually reaches it, so borderline headers that already fit are untouched (a blanket reserve regressed them).
- **render — hanging indent** on experience headers (`ats-resume-model.ts`): stamps a wrapped org tail as indented past the bullet-marker margin.
- **parse — fold the indented tail back** into the header before disambiguation (`tryFoldCompleteDateHeader` in `entry-blocks.ts`), gated on `isWrappedContinuation` (indented, *not* same-margin — so stacked #342/#466 second-line shapes are left alone) and a single-column section x-spread guard (so two-column sources are untouched).

Clears `awesome-cv-cv` / `awesome-cv-resume` from the corpus `KNOWN_FAILURES`, un-skips the AC#3 truncation assertion, and re-bakes those two corpus snapshots (only `experienceChangedAcrossRoundtrip` flips `true`→`false`). `honors-subheadings` stays baselined — a separate bullet-as-title root, not truncation.

Closes #436

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — full `src/lib/pdf` + `src/lib/heuristics` suites: 1330 pass / 0 fail
- [x] `npm run build` succeeds
- [x] `npm run check:fixtures` — 54 PDFs, all synthetic (no fixture binaries changed)
- [x] Corpus round-trip + edit-round-trip gates green; no new regressions

## Provenance

### DIFF
--- a/src/lib/heuristics/corpus-roundtrip.test.ts
+++ b/src/lib/heuristics/corpus-roundtrip.test.ts
@@ -159,22 +159,22 @@ const KNOWN_FAILURES: Record<string, Category[]> = {
   //      ("…, Remote") is peeled by `stripLocationSuffix` Pass F (achievements-
   //      oneline, two-column, chromium-asymmetric, weasyprint-two-column cleared).
   //   2. TRUNCATION — a PARENTHETICAL / multi-word company on the one-line shape
-  //      ("Danggeun Pay Inc. (KarrotPay)" → "(KarrotPay)") still re-parses
-  //      truncated. STILL OPEN — the four lines below are this root (plus
-  //      two-column re-segmentation). The fix needs a wrapped-header rejoin AND a
-  //      render-side date-column reserve, which perturb two-column round-trips
-  //      and so warrant their own follow-up, kept separate from the swap fix.
+  //      ("Danggeun Pay Inc. (KarrotPay)" → "(KarrotPay)") re-parsed truncated
+  //      when the too-wide header WRAPPED its org tail onto the row below and the
+  //      wrapped words were stranded (or glued to the flush-right date). FIXED:
+  //      the renderer now reserves the flush-right date column only when the
+  //      header actually reaches it (`drawText` in render-ats-pdf.ts), so the org
+  //      wraps clear of the date; and the parser folds that wrapped tail back
+  //      into the header before disambiguation (`tryFoldCompleteDateHeader` in
+  //      entry-blocks.ts), gated to single-column sections so a two-column source
+  //      is untouched. awesome-cv-cv / awesome-cv-resume cleared.
   //
   // Delete each line below as its root lands (the ratchet forces it — a fixed
   // fixture trips the stale-entry check).
+  //
+  // honors-subheadings: NOT the one-line header roots above — a bullet-as-title
+  // parse defect on a subheading-style role, tracked separately.
   "google-docs/google-docs-skia-proxy-honors-subheadings.pdf": ["experience"],
-  // awesome-cv-cv: #383 cleared its earlier abbreviated-date baseline (see the
-  // #383 note above), but main's one-line header re-regresses `experience` for
-  // the #436 TRUNCATION reason — 12 roles re-parse company-truncated/swapped off
-  // the one-line "Title · Company … Dates" shape. Removed when the truncation
-  // root lands.
-  "latex/awesome-cv-cv.pdf": ["experience"],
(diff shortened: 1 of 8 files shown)


s-annam (approved):
Reviewed at `fa1842b`. Ran the offlinecv gates + an inline correctness pass on the diff; **APPROVE** (0 blocking).

The three-part fix is sound. The render-side reserve is collision-gated — it re-wraps only when the header's first line would actually reach the flush-right date column, so every already-fitting header stays byte-identical (no borderline regression). The hanging indent is scoped to the experience-header returns only (education stays `hanging = 0`). And `tryFoldCompleteDateHeader` is tightly guarded: complete-date + indented-tail (`isWrappedContinuation`) + not-bullet / not-new-anchor + single-column x-spread, so stacked same-margin (#342/#466) and two-column shapes are left alone.

Verified locally at the reviewed tip: `npm run verify` green (typecheck, lint, coverage, build), `check:fixtures` 54 PDFs synthetic, the un-skipped AC#3 round-trip and the corpus round-trip both pass, and the two re-baked snapshots flip only `experienceChangedAcrossRoundtrip` true→false.

**Reviewer delta (amended into this commit, authorship preserved):** I folded the two nits from my read directly into the branch so a follow-up push wouldn't reset the approval —

- Extracted the date-column reserve out of `drawText` into a private `wrapReservingRight` helper. Behaviour-identical (same collision gate, same atomic re-wrap); it pulls `drawText` back under fallow's CRITICAL threshold (25→23 cyclomatic, 40→38 cognitive).
- Added two `parseEntryBlocks` unit tests in `entry-blocks.test.ts` pinning the new fold's indent gate — a complete-range header with an **indented** org tail folds (full `Danggeun Pay Inc. (KarrotPay)` preserved), and the same header with the company **at the margin** stays a distinct line.

No behavioural change to the fix itself.

---

PR #558, opened 2026-07-24
fix(parser): split experience title off a tight-right pipe header (#554)

### LINKED ISSUE #554: Experience header: tight-right pipe ("Company |Title") drops the title
## Summary

The experience header title/company delimiter split only fires when a `|` (or `@`, `—`, `·`) delimiter has **whitespace on both sides**. A header where the pipe is tight against the title — `Company, Location |Title` (space before the pipe, none after) — never splits: the whole run collapses into `company`, and `title` comes back `null`. The trailing `|Title` and the location are both swallowed into the company string.

## Reproduction

A single-column experience header of this shape:

```
Globex AI Labs LLC ,CA, USA |Intern    Jul 2025 – Present
● Built ...
● Integrated ...
```

Observed parse:

- `title` = `null`
- `company` = `"Globex AI Labs LLC ,CA, USA |Intern"`
- `location` = `null`

Expected parse:

- `title` = `"Intern"`
- `company` = `"Globex AI Labs LLC"`
- `location` = `"CA, USA"` (or at minimum a non-null company without the `|Intern` tail)

Contrast a sibling header on the same résumé that parses correctly because the pipe is spaced on both sides:

```
Initech Client Services - Downtown Campus, USA | Support Technician   Jan 2024 – May 2025
```
→ `title = "Support Technician"`, `company = "Initech Client Services - Downtown Campus"`, `location = "USA"` ✓

The only structural difference between the two headers is the whitespace around the pipe: `USA | Student` (spaced) vs `USA |Intern` (tight-right).

## Root cause

`src/lib/heuristics/extract/experience-disambiguate.ts:513`

```ts
const atSplit = h.split(/\s+@\s+|\s+—\s+|\s+\|\s+|\s+·\s+/);
```

Every delimiter alternative requires `\s+` on **both** sides. `USA |Intern` has a space before the pipe but none after, so no alternative matches, `atSplit.length === 1`, and the header is treated as an unsplit (`via: "whole"`) run — company-only, no title.

## Implementation plan

1. **Loosen the delimiter regex** in `experience-disambiguate.ts:513` so a delimiter with whitespace on **at least one** side splits, while still not cleaving a bare `A|B` with no surrounding space (which would risk splitting URLs / table residue). Candidate:

   ```ts
   const atSplit = h.split(/\s+@\s+|\s+—\s+|\s*\|\s+|\s+\|\s*|\s+·\s+/);
   ```

   i.e. for the pipe, accept `\s*\|\s+` (space optional before, required after) **or** `\s+\|\s*` (required before, optional after). Keep `@`/`—`/`·` as-is unless a fixture shows the same tight-delimiter shape for them. Trim each segment afterward (the split already feeds `.trim()` downstream — confirm).

2. **Verify segment mapping** still holds: after the split, `USA |Intern` yields `["Globex AI Labs LLC ,CA, USA", "Intern"]`. Confirm the existing title/company/location mapping logic (`mapTitleFirst` / location peel via `looksLikeLocationTail`) resolves the leading segment to company+location and the trailing `Intern` to title. If `Intern` alone does not pass `looksLikeTitle`, check whether a short one-word title needs the title-first path.

3. **Guard against over-splitting**: add a test asserting a header with an internal `A|B` and no surrounding whitespace is **not** split (regression guard for the tightened-but-not-zero-width intent).

## Acceptance criteria

- [ ] `Globex AI Labs LLC ,CA, USA |Intern` header (tight-right pipe) parses to `title = "Intern"`, non-null `company` **without** the `|Intern` tail.
- [ ] The spaced-pipe sibling header (`… USA | Support Technician`) still parses exactly as before.
- [ ] A tight `A|B` header with no surrounding whitespace on either side is **not** split (guard test).
- [ ] `src/lib/heuristics/extract/experience.pipe-location.test.ts` and `experience.mid-dot.test.ts` still pass.
- [ ] `npx vitest run src/lib/heuristics/` passes; corpus + round-trip invariants (`corpus.test.ts`, `corpus-roundtrip.test.ts`) unchanged.
- [ ] A new synthetic-persona fixture (or a targeted `*.repro.test.ts`) pins the tight-right-pipe shape. Fixture PII rules apply (`CLAUDE.md` → Hard rules): synthetic name, `@example.com`, real-area-code + `555` + `0100–0199` phone.

## Notes

Found while probing a real résumé (`/probe-experience`); the real PDF is local-only and must never become a fixture. The repro strings above are synthetic.

rohithgollapalli:
## Summary

An experience header of the shape `Company, Location |Title` — where the pipe hugs the title (space before the pipe, none after) — dropped the title: the delimiter split in `splitHeaderSegments` required whitespace on **both** sides of the pipe (`\s+\|\s+`), so `USA |Intern` never cleaved and the whole run collapsed into `company` with an empty `title`. The split now fires when whitespace is present on **at least one** side, while a zero-width `A|B` with no surrounding whitespace stays unsplit (URL / table residue never breaks). `@`/`—`/`·` keep the stricter both-sides rule.

Closes #554

## Test plan

- [x] `npx vitest run src/lib/heuristics/` green (1067 passed) — incl. new `experience.tight-pipe-title.test.ts` and unchanged `experience.pipe-location`/`mid-dot`/`location`/`role-comma`/`title-team-comma`
- [x] Verified against the reporting résumé: the affected role now parses `title = "Intern"`

## Provenance

Review revisions via: Claude Opus 5 (1M context)

### DIFF
--- a/src/lib/heuristics/extract/experience.tight-pipe-title.test.ts
+++ b/src/lib/heuristics/extract/experience.tight-pipe-title.test.ts
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * Regression for #554 — a "Company, Location |Title" header whose pipe is tight
+ * against the title (space BEFORE the pipe, none after) dropped the title:
+ *
+ *   Globex AI Labs LLC ,CA, USA |Intern    Jul 2025 - Present
+ *
+ * The delimiter split in `splitHeaderSegments` required whitespace on BOTH
+ * sides of the pipe (`\s+\|\s+`), so `USA |Intern` never cleaved — the whole
+ * run collapsed into `company` and `title` came back empty. The split now
+ * accepts whitespace on at least one side, while a zero-width `A|B` (no
+ * surrounding whitespace) still stays unsplit.
+ *
+ * Synthetic personas only, per the fixtures PII policy.
+ */
+
+import { describe, it, expect } from "vitest";
+import { groupIntoLines, splitIntoSections, findSection } from "../sections.ts";
+import { extractExperience } from "../extract-fields.ts";
+import { mkItems } from "../__test-utils__/mkItem.ts";
+
+function roleFromSection(specs: Array<{ text: string; fontSize?: number }>) {
+  const sections = splitIntoSections(groupIntoLines(mkItems(specs)));
+  const experience = findSection(sections, "experience");
+  expect(experience).toBeDefined();
+  return extractExperience(experience).value;
+}
+
+describe("tight-right pipe 'Company, Location |Title' (#554)", () => {
+  it("cleaves the title off a pipe with no trailing space", () => {
+    const roles = roleFromSection([
+      { text: "Experience", fontSize: 13 },
+      // The issue's header verbatim, internal comma run and all.
+      {
+        text: "Globex AI Labs LLC ,CA, USA |Intern Jul 2025 - Present",
+        fontSize: 11,
+      },
+      { text: "• Built a secure transcription pipeline.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    const role = roles[0];
+    // #554 is the title/company split: the tail after the tight pipe must land
+    // in `title`, and `company` must no longer carry the "|Intern" run.
+    expect(role.title).toBe("Intern");
+    expect(role.company).toContain("Globex AI Labs LLC");
+    expect(role.company).not.toContain("|");
+    expect(role.company).not.toContain("Intern");
+    // The widened split must not eat the trailing location off the comma run.
+    expect(role.location).toBe("USA");
+  });
+
+  it("does NOT split a zero-width 'A|B' with no surrounding whitespace", () => {
+    const roles = roleFromSection([
+      { text: "Experience", fontSize: 13 },
+      { text: "Acme|Beta Systems Jan 2020 - Dec 2021", fontSize: 11 },
+      { text: "• Shipped the platform.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    expect(roles[0].company).toContain("Acme|Beta Systems");
+    expect(roles[0].title ?? "").toBe("");
+  });
+});
--- a/src/lib/heuristics/extract/experience-disambiguate.ts
+++ b/src/lib/heuristics/extract/experience-disambiguate.ts
@@ -510,7 +510,13 @@ function stripLeadingSectionHeaders(
 function splitHeaderSegments(filtered: string[]): Split[] {
   const splits: Split[] = [];
   filtered.forEach((h, idx) => {
-    const atSplit = h.split(/\s+@\s+|\s+—\s+|\s+\|\s+|\s+·\s+/);
+    // The pipe accepts whitespace on AT LEAST ONE side (`\s+\|\s*` or
+    // `\s*\|\s+`), not both — a tight-right header "Company, Location |Title"
+    // (space before the pipe, none after) must still cleave the title off
+    // (#554). A zero-width `A|B` with no surrounding whitespace stays unsplit
+    // so a URL / table residue is never split. `@`/`—`/`·` keep the stricter
+    // both-sides rule (an email `a@b` must not split).
+    const atSplit = h.split(/\s+@\s+|\s+—\s+|\s+\|\s*|\s*\|\s+|\s+·\s+/);
     if (atSplit.length > 1) {
       // A PURE-middot line ("Title · Company · Team") — the exporter's one-line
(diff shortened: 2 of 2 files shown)


src/lib/heuristics/extract/experience.tight-pipe-title.test.ts
```diff
@@ -0,0 +1,61 @@
...
+  expect(experience).toBeDefined();
+  return extractExperience(experience).value;
+}
+
+describe("tight-right pipe 'Company, Location |Title' (#554)", () => {
+  it("cleaves the title off a pipe with no trailing space", () => {
+    const roles = roleFromSection([
+      { text: "Experience", fontSize: 13 },
+      {
+        text: "Globex AI Labs LLC, USA |Intern Jul 2025 - Present",
```
s-annam: Nit (non-blocking): this simplifies the issue's repro `Globex AI Labs LLC ,CA, USA |Intern` to `Globex AI Labs LLC, USA |Intern` and the test never asserts `location`. The title/company split is fully covered, but adding `expect(role.location).toBe("...")` (and keeping the internal-comma location from the issue) would pin that the location survives the widened split too. Optional — AC #1 only asks for title + a clean company, which this asserts.

s-annam (dismissed):
Tightly scoped, correct fix for #554. The delimiter split in `splitHeaderSegments` now fires when whitespace is on **at least one** side of the pipe (`\s+\|\s*` or `\s*\|\s+`), so `USA |Intern` cleaves the title off while a zero-width `A|B` stays unsplit and `@`/`—`/`·` keep the stricter both-sides rule. The alternation is symmetric, so it also handles the tight-left `Company| Title` shape for free — a small bonus over the issue's stated tight-right scope.

Verdict rule: 0 Blocking → APPROVE.

**Gates run:** typecheck clean; `npx vitest run src/lib/heuristics/` green (1067 passed, matching the PR claim), including the new `experience.tight-pipe-title.test.ts` plus the unchanged `pipe-location` / `mid-dot` and the corpus + round-trip invariants. Fixture-PII, design-system/reuse, style-token, fallow dead-code, and skill/script command gates all **N/A** — the diff is one lib regex + one test file, no fixtures, UI, scripts, or new exports.

## Blocking

None.

## Secondary

None.

## Nits

1. (non-blocking) The regression test drops the issue repro's internal comma-location (`,CA,`) and never asserts `location`, so it pins the title/company split but not that the location survives the widened split. The stated AC #1 only requires title + a clean company, which the test meets — noting only that a `location` assertion on the fuller `Globex AI Labs LLC ,CA, USA |Intern` shape would guard the whole reported header. Optional.

---

src/lib/heuristics/extract/experience.tight-pipe-title.test.ts
s-annam: Fixed in `7e0b2d8`. The header is now the issue's repro verbatim — `Globex AI Labs LLC ,CA, USA |Intern Jul 2025 - Present`, internal comma run and all — and the test asserts `expect(role.location).toBe("USA")` alongside the title/company split.

Parsed shape on the restored input: `title = "Intern"`, `company = "Globex AI Labs LLC ,CA"`, `location = "USA"`. The existing `company` assertions are `toContain`/`not.toContain`, so they hold unchanged against the longer comma run.

Also re-checked that the widened input still guards: reverting the regex hunk back to `\s+\|\s+` flips the first test red (`expected '' to be 'Intern'`) while the zero-width `A|B` case stays green. Gates on the amended commit: `npm run typecheck` clean, `npm run test` green (204 files, 2856 passed / 8 skipped), `npm run lint` clean.

Kept the branch at one commit by amending rather than stacking a fixup, per the merge-queue squash-message rule — hence the force-push (and the dismissed approval).

s-annam (approved):
Re-approving after the nit was addressed. Disclosing my full delta, since I both revised and approved:

- Restored the issue's repro header verbatim in `experience.tight-pipe-title.test.ts` (`Globex AI Labs LLC ,CA, USA |Intern Jul 2025 - Present`) and added `expect(role.location).toBe("USA")`. Parsed shape: `title = "Intern"`, `company = "Globex AI Labs LLC ,CA"`, `location = "USA"` — the existing `toContain`/`not.toContain` company assertions hold unchanged.
- No production code changed. `experience-disambiguate.ts` is byte-identical to the previously approved revision.
- Amended the author's commit rather than stacking a fixup, so the branch reaches the merge queue as one commit; authorship preserved. That force-push is what dismissed the earlier approval.
- Added a `Review revisions via: Claude Opus 5 (1M context)` line to the existing Provenance block.

Guard re-verified: reverting the regex back to `\s+\|\s+` flips the first test red (`expected '' to be 'Intern'`) while the zero-width `A|B` case stays green — the widened input still pins the fix.

Gates: `verify` pass, `fallow` pass, Cloudflare Pages pass. Locally typecheck clean, 2856 tests passed / 8 skipped, lint clean. The fallow complexity findings on `experience-disambiguate.ts` are inherited and report-only.

Issue #557, opened 2026-07-24
Education: connective-less degree ("Master of Science <Field>") mis-splits degree/field


rohithgollapalli:
## Summary

`parseDegreeAndField` mis-partitions "connective-less" degree names — degrees written `<Credential> of <Type> <Field>` with **no " in " connective** before the field. `DEGREE_RE`'s greedy `of <subject>` branch swallows the field into the credential, so either the field is lost entirely or the split lands mid-phrase. Affects common Indian degree conventions (`Master of Science <Field>`, `Bachelor of Technology <Field>` a.k.a. B.Tech) and any résumé that drops the "in".

## Reproduction

Two entries, same defect class:

**A — field lost entirely:**
```
Master of Science Information Technology, Some University
```
Observed: `degree = "Master of Science Information Technology"`, `field = null`
Expected: `degree = "Master of Science"`, `field = "Information Technology"`

**B — split lands mid-phrase:**
```
Bachelor of Technology Electronics & Communication Eng., Some Institute of Tech   Jun 2018 – May 2022
```
Observed: `degree = "Bachelor of Technology Electronics"`, `field = "& Communication Eng."`
Expected: `degree = "Bachelor of Technology"`, `field = "Electronics & Communication Eng."`

For comparison, the US "in"-connective form parses correctly today: `Bachelor of Science in Biology` → `degree = "Bachelor of Science"`, `field = "Biology"`.

## Root cause

`src/lib/heuristics/extract/education.ts:424` `parseDegreeAndField`.

```ts
const dm = DEGREE_RE.exec(line);
const matched = dm[0];
const inIdx = matched.search(/\s+in\s+/i);   // only splits on " in "
if (inIdx >= 0) { degree = matched.slice(0, inIdx); ... }
else { degree = matched.trim(); fieldStart = dm.index + matched.length; }
```

`DEGREE_RE`'s optional `of <subject>` branch greedily extends the match past the credential. For `Master of Science Information Technology` it swallows `of Science Information Technology` into `matched`; there is no `" in "`, so the `else` branch takes the whole thing as `degree` and `fieldStart` lands at end-of-string → `field` empty. For `Bachelor of Technology Electronics & …` the branch stops at the `&` (not a word char run it matches), so `matched` = `Bachelor of Technology Electronics`, leaving `& Communication Eng.` as the field — a mid-phrase cut.

[...]

Issue #556, opened 2026-07-24
Education: GPA note glued to institution ("University GPA 3.79") not stripped


rohithgollapalli:
## Summary

A GPA note appended to the institution string with no comma separator — `Some University GPA 3.8` — is never stripped, so the parsed `institution` field comes back as `"Some University GPA 3.8"`. There is no GPA/grade-note strip on the institution line; `stripInstitutionDate` only peels trailing dates, and `cleanField`'s GPA cut runs on the degree *field* (not the institution) and only fires after a `,`/`;` delimiter.

## Reproduction

A one-line education entry that carries the GPA on the institution tail, space-separated:

```
Master of Science Information Technology, Some University GPA 3.8
Coursework: Advanced Databases, Data Analytics, Data Mining
```

Observed:

- `institution` = `"Some University GPA 3.8"`

Expected:

- `institution` = `"Some University"` (GPA either dropped or captured into a dedicated field, but not left glued to the school name)

## Root cause

`src/lib/heuristics/extract/education.ts`

- `stripInstitutionDate` (line 625) peels only trailing **dates** from the institution string — no grade/GPA handling.
- `cleanField` (line 404) does cut a trailing GPA note, but (a) it operates on the degree **field**, not the institution, and (b) its pattern `/[,;]\s*(?:minor|major|gpa|concentration)\b.*$/i` requires a leading `,` or `;`. `Some University GPA 3.8` has a **space-separated** GPA with no comma, so even if it ran here it would not match.

Net: nothing removes a space-separated `GPA <n.nn>` tail from the institution.

## Implementation plan

1. **Add a `stripInstitutionGrade` helper** (next to `stripInstitutionDate`) that peels a trailing grade/GPA note from the institution string. Cover the common shapes, comma-delimited or not:
   - `GPA 3.8`, `GPA: 3.8`, `GPA 3.8/4.0`, `CGPA 8.7`, `CGPA 8.7/10`, `GPA 3.8/4`
   - optional leading `,`/`;`/`·`/`|` separator, `$`-anchored, case-insensitive.

   Sketch:
   ```ts
   const INSTITUTION_GRADE_RE =
     /\s*[,;·|]?\s*(?:c?gpa)\b[:\s]*\d(?:\.\d+)?(?:\s*\/\s*\d+(?:\.\d+)?)?\s*$/i;
   function stripInstitutionGrade(s: string): string {
     const stripped = s.replace(INSTITUTION_GRADE_RE, "").trim();
[...]

PR #559, opened 2026-07-24
fix(parser): recover an education date from a coursework-line tail (#555)

### LINKED ISSUE #555: Education: attendance date at the tail of a coursework line is dropped
## Summary

An education entry's attendance dates are dropped when the date range sits at the **tail of a coursework line** rather than on the degree/institution line. `filterAnnotationLinesForDates` removes the entire coursework line before date parsing runs (the line matches `EDUCATION_ANNOTATION_RE` on the word "coursework"), so a real attendance date appended to that same line is discarded along with it. The entry comes back with `dates = null` / `year = null`.

## Reproduction

An education entry whose dates live at the end of the coursework line (a real layout — some résumés right-align the date on the coursework row):

```
Master of Science Information Technology, Some University GPA 3.8
Coursework: Advanced Databases, Data Analytics, Data Mining. Aug 2023 – May 2025
```

Observed parse:

- `degree` = `"Master of Science Information Technology"`
- `dates` = `null` → `null`, `year` = `null`
- coursework = 3 items (the trailing `Aug 2023 – May 2025` may also pollute the last item)

Expected parse:

- `start_date` ≈ `Aug 2023`, `end_date` ≈ `May 2025`, `year` = `2025`
- coursework = the 3 course items, **without** the date tail

Contrast the sibling Bachelor entry on the same résumé, which parses its dates correctly because the range is on the **degree line**:

```
Bachelor of Technology Electronics & Communication Eng., Northwind Institute of Tech Jun 2018 - May 2022
```
→ `start_date = Jun 2018`, `end_date = May 2022` ✓

## Root cause

`src/lib/heuristics/extract/education.ts:161`

```ts
function filterAnnotationLinesForDates(lines: readonly string[]): string[] {
  return lines.filter((l) => DEGREE_RE.test(l) || !EDUCATION_ANNOTATION_RE.test(l));
}
```

`EDUCATION_ANNOTATION_RE` (line 143) includes `coursework`. The coursework line matches, and it carries no `DEGREE_RE` token, so the DEGREE carve-out does not rescue it — the whole line (date tail included) is filtered out before `parseEducationDates` runs at line 824. The function's own docblock flags the assumption: *"an annotation phrase inline with a real date on the SAME line is rare in practice"* — this résumé is a counterexample.

## Implementation plan

1. **Preserve a trailing date on an annotation line instead of dropping the whole line.** In `filterAnnotationLinesForDates`, when a line matches `EDUCATION_ANNOTATION_RE` but is **not** a `DEGREE_RE` line, don't discard it outright — first check for a trailing date range (reuse the same date-range regex `parseDateRange` keys on). If one is present at the tail, keep only that date substring (drop the annotation prose) so it feeds `parseEducationDates`; otherwise filter as today.

   Sketch:
   ```ts
   function filterAnnotationLinesForDates(lines: readonly string[]): string[] {
     return lines.flatMap((l) => {
       if (DEGREE_RE.test(l) || !EDUCATION_ANNOTATION_RE.test(l)) return [l];
       const tailDate = extractTrailingDateRange(l); // new helper over the shared date regex
       return tailDate ? [tailDate] : [];
     });
   }
   ```

2. **Keep the date out of the coursework list.** The coursework builder (around `education.ts:868`) must strip the same trailing date range from the course items so `Data Mining. Aug 2023 – May 2025` does not surface as a course. Peel the trailing date before the comma-split, mirroring how `COURSEWORK_LABEL_RE` peels the leading label.

3. **Reuse the shared date primitive** — do not hand-roll a second date regex. Localize the trailing-date extraction on the same source-of-truth the rest of the module uses (`parseDateRange` / its underlying range regex) so the two can't drift.

## Acceptance criteria

- [ ] A Masters entry whose range sits at the coursework-line tail (repro above) parses `start_date`/`end_date` (and `year`), not `null`.
- [ ] The trailing date does **not** appear as a coursework item; the course list is exactly the real courses.
- [ ] The Bachelor sibling (date on the degree line) still parses unchanged.
- [ ] An annotation line with a date that is NOT a real attendance date (e.g. `Dean's List 2015–2017` on its own line) still does **not** override a real graduation year on a sibling line — the #371 regression the filter was built to prevent. Add/keep a guard test.
- [ ] `npx vitest run src/lib/heuristics/extract/education.test.ts` and the education corpus fixtures pass; `corpus.test.ts` / `corpus-roundtrip.test.ts` unchanged.
- [ ] A new synthetic-persona fixture (or `*.repro.test.ts`) pins the coursework-line-tail-date shape. Fixture PII rules apply (`CLAUDE.md` → Hard rules).

## Notes

Found while probing a real résumé (`/probe-education`); the real PDF is local-only and must never become a fixture. Repro strings above are synthetic. Related prior date-handling work: #371 (annotation-line date filtering), #506 (one-line education date en-dash split).

rohithgollapalli:
## Summary

An education entry's attendance dates were dropped when the date range sat at the **tail of a coursework line** — `Coursework: … Data Mining. Aug 2023 – May 2025` — rather than on the degree/institution line. The coursework line is `consumed` before chunking, so the chunk's own date parse never saw it, and the trailing range also leaked into the last course item. The fix peels the trailing month-year range off the coursework line, attributes it to the entry it belongs to (only when that entry parsed **no** date of its own, so a real degree-line date always wins), and keeps it out of the course list. `filterAnnotationLinesForDates` gains the same tail-date carve-out for an annotation line that reaches the chunk directly. Extraction uses `STRICT_MONTH_YEAR_RE`, so the peel can never eat a word.

Closes #555

## Test plan

- [x] `npx vitest run src/lib/heuristics/` green (1067 passed) — incl. new `education.coursework-tail-date.test.ts` and the #371 over-capture guard
- [x] Verified on the reporting résumé: the Masters entry now parses `Aug 2023 – May 2025` (year 2025), and the date no longer appears as a course

## Provenance

### DIFF
--- a/src/lib/heuristics/extract/education.coursework-tail-date.test.ts
+++ b/src/lib/heuristics/extract/education.coursework-tail-date.test.ts
@@ -1,0 +1,70 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * Regression for #555 — an education entry's attendance date is dropped when
+ * the date range sits at the TAIL of a coursework line rather than on the
+ * degree/institution line:
+ *
+ *   Master of Science Information Technology, Northwind University
+ *   Coursework: Advanced Databases, Data Analytics, Data Mining. Aug 2023 – May 2025
+ *
+ * `filterAnnotationLinesForDates` used to drop the whole coursework line before
+ * date parsing (it matches the "coursework" annotation keyword and carries no
+ * degree token), taking the trailing `Aug 2023 – May 2025` with it — the entry
+ * came back with no dates. The fix keeps just the trailing date for parsing and
+ * peels it off the coursework list so it is not surfaced as a phantom course.
+ *
+ * Synthetic personas only, per the fixtures PII policy.
+ */
+
+import { describe, it, expect } from "vitest";
+import { extractEducation } from "../extract/education.ts";
+import { type PdfLine, type PdfSection } from "../sections.ts";
+
+const mkLine = (text: string): PdfLine => ({
+  page: 0,
+  y: 0,
+  x: 0,
+  items: [],
+  text,
+  maxFontSize: 11,
+  allCaps: false,
+  gapAbove: 0,
+});
+const mkEduSection = (texts: string[]): PdfSection => ({
+  name: "education",
+  lines: texts.map(mkLine),
+});
+
+describe("extractEducation — attendance date at the coursework-line tail (#555)", () => {
+  it("recovers the range from the coursework tail and keeps it out of the course list", () => {
+    const { value } = extractEducation(
+      mkEduSection([
+        "Master of Science Information Technology, Northwind University",
+        "Coursework: Advanced Databases, Data Analytics, Data Mining. Aug 2023 – May 2025",
+      ]),
+    );
+    expect(value.length).toBe(1);
+    const edu = value[0];
+    expect(edu.start_date).toBeTruthy();
+    expect(edu.end_date).toBeTruthy();
+    expect(edu.year).toBe("2025");
+    // The date must not leak into the coursework items.
+    const courses = edu.coursework ?? [];
+    expect(courses).toContain("Data Mining");
+    expect(courses.join(" ")).not.toMatch(/2023|2025|Aug|May/);
+  });
+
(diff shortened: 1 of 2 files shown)


rohithgollapalli:
Superseded by #560, which folds #555 together with #556 and #557 into one education-parser PR (single commit). Closing this in favor of #560.

PR #560, opened 2026-07-24
fix(parser): correct three education one-line-entry defects (#555, #556, #557)

### LINKED ISSUE #555: Education: attendance date at the tail of a coursework line is dropped

### LINKED ISSUE #556: Education: GPA note glued to institution ("University GPA 3.79") not stripped

### LINKED ISSUE #557: Education: connective-less degree ("Master of Science <Field>") mis-splits degree/field

rohithgollapalli:
## Summary

Three independent education-parser defects, all surfaced by **single-line education entries** where degree, institution, GPA, coursework, and dates share one or two physical lines. Each is fixed in `src/lib/heuristics/extract/education.ts` with a targeted, value-asserting unit test; one corpus golden is re-baked per fix.

- **#555 — attendance date at a coursework-line tail.** A range appended to the END of a coursework line (`Coursework: … Data Mining. Aug 2023 – May 2025`) was dropped: the coursework line is `consumed` before the chunk's date parse, and the range also leaked into the last course. Now peeled, attributed to the entry (only when it parsed no date of its own, so a real degree-line date wins), and kept out of the course list. Extraction uses `STRICT_MONTH_YEAR_RE`, so the peel never eats a word.
- **#556 — GPA glued to the institution.** A `GPA`/`CGPA` note with no comma boundary (`State University GPA 3.8`, `College · GPA 8.0 / 10`) rode into the institution name. New `stripInstitutionGrade` (keyword-gated, so a school name ending in a digit is never cut) applied in the institution-building path.
- **#557 — connective-less degree/field.** A `<prefix> of <Type> <Field>` degree with no ` in ` (`Master of Science Information Technology`, `Bachelor of Technology Electronics & Communication Eng.`) had its field swallowed into the credential by `DEGREE_RE`'s greedy `of` branch. `DEGREE_TYPE_RE` now cuts the credential at the degree-TYPE word so the field survives; a type-only credential (`Master of Business Administration`) stays whole; the US ` in ` form is untouched.

All three were found probing a real résumé; the real PDF is local-only and was never committed, and every repro string in the tests is a synthetic persona.

Closes #555
Closes #556
Closes #557

## Test plan

- [x] `npx vitest run src/lib/heuristics/` green (1075 passed) — three new tests (`education.coursework-tail-date`, `education.gpa-institution`, `education.connectiveless-degree`) plus the existing #371 date-over-capture guard
- [x] Two corpus goldens re-baked (`google-docs-skia-proxy-additional-skills`, `compound-certifications-activities-tail`), each a strict improvement (a GPA / a leaked date removed from a parsed field)
- [x] Verified on the reporting résumé: Masters parses `Aug 2023 – May 2025`, institution has no GPA tail

## Provenance

Review revisions via: Claude Opus 5 (high)

### DIFF
--- a/src/lib/heuristics/extract/education.ts
+++ b/src/lib/heuristics/extract/education.ts
@@ -7,6 +7,7 @@ import {
   DEGREE_RE,
   INSTITUTION_HINTS,
   MONTH_YEAR_RE,
+  STRICT_MONTH_YEAR_RE,
   NUMERIC_MONTH_YEAR_RE,
   US_STATE_CODE_RE,
   COUNTRY_GAZETTEER,
@@ -143,23 +144,61 @@ function educationDateFields(
 const EDUCATION_ANNOTATION_RE =
   /\b(dean'?s? list|awards?|honou?rs?|thesis|teaching assistant|research assistant|study abroad|coursework|scholarships?|fellowships?|cum laude|capstone|(?:senior|final|independent|group|team)\s+project)\b/i;
 
+/** A month-year attendance/graduation date range sitting at the very END of a
+ *  line — "… Data Mining. Aug 2023 – May 2025" — captured in the `date` group.
+ *  The leading `[.,;\s]+` swallows the separator (a period, comma, or space)
+ *  between the prose and the date so the whole tail can be stripped in one
+ *  replace. The range START must be a month-year (`STRICT_MONTH_YEAR_RE` /
+ *  numeric), never a bare year — a coursework item that merely ends in a 4-digit
+ *  token ("… since 2020") must not be mistaken for an attendance date; only the
+ *  range END may be a bare year ("Aug 2023 – 2025") or an open marker
+ *  ("Aug 2023 – Present"). Uses the STRICT month regex, not the loose
+ *  `MONTH_YEAR_RE`, because this value is both EXTRACTED and DELETED — the loose
+ *  `[a-z]*` tail would false-match a word like "Marketing" and eat it (#380). */
+const ATTENDANCE_RANGE_END = `(?:${STRICT_MONTH_YEAR_RE.source}|${NUMERIC_MONTH_YEAR_RE.source}|\\b\\d{4}\\b|present|current|ongoing|now)`;
+const TRAILING_ATTENDANCE_DATE_RE = new RegExp(
+  `[.,;\\s]+(?<date>(?:${STRICT_MONTH_YEAR_RE.source}|${NUMERIC_MONTH_YEAR_RE.source})(?:\\s*[–—-]\\s*${ATTENDANCE_RANGE_END})?)\\s*$`,
+  "i",
+);
+
+/** Peel a trailing attendance-date range off a coursework line and return just
+ *  the date, or `undefined` when the line carries none. Lets a real attendance
+ *  date that rides at the tail of a coursework line reach the entry (#555)
+ *  instead of being surfaced as a phantom last course — the coursework line is
+ *  `consumed`, so it never reaches the chunk's own date parse. The recovered
+ *  date is attributed by {@link extractEducation}'s `courseworkDates` pass,
+ *  which applies it ONLY when the entry parsed no date of its own. */
+function trailingAttendanceDate(line: string): string | undefined {
+  return TRAILING_ATTENDANCE_DATE_RE.exec(line)?.groups?.date?.trim();
+}
+
 /** Strip lines that read as honors / awards / activity annotations from a
  *  chunk before {@link parseEducationDates} runs on the joined text (#371).
  *  Without this, a line like "GPA: 3.7 · Dean's List 2015–2017" contributes
  *  its range to the entry and `parseDateRange`'s range-first preference makes
  *  the annotation range (2015–2017) win over the real graduation year (2017)
- *  on a sibling line. Line-level filter — an annotation phrase inline with a
- *  real date on the SAME line is rare in practice; keeping the whole line
- *  when it doesn't lead with an annotation is safer than mid-line surgery.
+ *  on a sibling line.
  *
  *  DEGREE-carve-out: a line that ALSO matches {@link DEGREE_RE} carries the
  *  real attendance / graduation date and stays regardless of which annotation
  *  keywords it contains. Commonwealth shapes like "Honours Bachelor of Science,
  *  2015 - 2019" or "Thesis-based M.Sc. Data Science, 2018 - 2020" mention
  *  `honours` / `thesis` on the SAME line as the degree + real dates; filtering
- *  the whole line would drop both. */
+ *  the whole line would drop both.
+ *
+ *  No tail-date carve-out here, deliberately: peeling a trailing range off an
+ *  annotation line and feeding it to {@link parseEducationDates} would reopen
+ *  the #371 over-capture class for month-year-precision ranges — an "Awards:
+ *  Best Thesis, Sep 2019 - Jun 2020" line would beat a real single-year
+ *  graduation date on the sibling degree line, because `parseDateRange` prefers
+ *  a range over a bare year. A coursework line's real attendance date is
+ *  recovered instead by {@link extractEducation}'s `courseworkDates` pass
+ *  (#555), which carries the "only when the entry has no date of its own" guard
+ *  this whole-chunk path cannot express. */
 function filterAnnotationLinesForDates(lines: readonly string[]): string[] {
-  return lines.filter((l) => DEGREE_RE.test(l) || !EDUCATION_ANNOTATION_RE.test(l));
+  return lines.filter(
+    (l) => DEGREE_RE.test(l) || !EDUCATION_ANNOTATION_RE.test(l),
+  );
 }
 
 /** Source-side coursework label to peel off the bullet residue before the
@@ -421,6 +460,15 @@ function cleanField(raw: string): string | undefined {
   return f.length > 0 ? f : undefined;
 }
 
+/** The degree-TYPE word that follows "of" in a spelled-out credential
+ *  ("Bachelor **of Technology**", "Master **of Science**"). Anchored to the
+ *  start of the post-"of" run, longest-first so "Business Administration" and
+ *  "Fine Arts" win over their prefixes. Used by {@link parseDegreeAndField} to
+ *  find where the credential ends and the subject field begins in a
+ *  connective-less "of <Type> <Field>" header (#557). */
+const DEGREE_TYPE_RE =
+  /^(?:Business Administration|Fine Arts|Computer Applications|Science|Arts|Technology|Engineering|Philosophy|Education|Commerce|Laws|Law|Medicine|Architecture|Design|Nursing|Pharmacy|Business|Management)\b/i;
+
 /** Parse a degree-bearing line into a bare credential + its subject field.
  *
  * `DEGREE_RE` either matches just the credential ("B.S.") or — via its optional
@@ -432,7 +480,16 @@ function cleanField(raw: string): string | undefined {
  * whatever text follows the credential, with dates/notes stripped by
  * `cleanField`. Connective-less shapes ("M.S. Computer Science", "B.S. Business
  * Administration — May 2027") are handled too: the field is the post-credential
- * remainder. */
+ * remainder.
+ *
+ * CONNECTIVE-LESS "of" shapes (#557): a spelled-out "<Bachelor|Master|…> of
+ * <Type> <Field>" with no " in " ("Master of Science Information Technology",
+ * "Bachelor of Technology Electronics & Communication Eng.") is common in
+ * Indian résumés. `DEGREE_RE`'s greedy `of [A-Za-z ]{2,40}` swallows the field
+ * INTO the credential, so the split below would leave `field` empty (or, when an
+ * "&" halts the greedy run, split it mid-phrase). {@link DEGREE_TYPE_RE} cuts
+ * the credential at the end of the degree-TYPE word so the remainder is the
+ * field. */
 function parseDegreeAndField(line: string): {
   degree: string;
   field?: string;
@@ -452,6 +509,21 @@ function parseDegreeAndField(line: string): {
   } else {
     degree = matched.trim();
     fieldStart = dm.index + matched.length;
+    // Connective-less "<prefix> of <Type> <Field>" (#557): if the match ran
+    // "of <Type>" straight into a field, cut the credential at the end of the
+    // degree-TYPE word so <Field> is not buried in the degree. Only fires when
+    // real field text follows the type word — "Master of Business
+    // Administration" (type IS the whole credential) is left whole.
+    const ofm = /\bof\s+/i.exec(matched);
+    if (ofm) {
+      const afterOf = matched.slice(ofm.index + ofm[0].length);
+      const typeM = DEGREE_TYPE_RE.exec(afterOf);
+      if (typeM && afterOf.slice(typeM[0].length).trim().length > 0) {
+        const degreeEnd = ofm.index + ofm[0].length + typeM[0].length;
+        degree = matched.slice(0, degreeEnd).trim();
+        fieldStart = dm.index + degreeEnd;
+      }
+    }
   }
   const fieldRaw = line
     .slice(fieldStart)
@@ -655,6 +727,23 @@ function stripInstitutionDate(s: string): string {
   return stripped || s;
 }
 
+/** A trailing GPA / CGPA note glued onto the institution string with no comma
+ *  boundary — "State University GPA 3.8" (#556). Covers "GPA 3.8", "GPA: 3.8",
+ *  "CGPA 8.7", a "/N" scale ("GPA 3.8/4.0", "CGPA 8.7/10") and a percentage
+ *  ("GPA 85%"), with an optional leading separator, `$`-anchored. Requires the
+ *  `gpa`/`cgpa` keyword, so a school name that merely ends in a digit is never
+ *  cut. */
+const INSTITUTION_GRADE_RE =
+  /[\s,;·|-]*\bc?gpa\b[:\s]*\d(?:\.\d+)?(?:\s*\/\s*\d+(?:\.\d+)?)?%?\s*$/i;
+
+/** Peel a trailing GPA/CGPA note off an institution string, leaving the bare
+ *  school name. Never consumes the whole string (a GPA-only value keeps the
+ *  original). Mirrors {@link stripInstitutionDate}'s trailing-strip contract. */
+function stripInstitutionGrade(s: string): string {
+  const stripped = s.replace(INSTITUTION_GRADE_RE, "").trim();
+  return stripped || s;
+}
+
 /** Map one education chunk (the degree + institution + date lines of a single
  *  qualification) to a `ResumeEducation` and its confidence. */
 function educationFromChunk(chunk: string[]): {
@@ -793,6 +882,12 @@ function educationFromChunk(chunk: string[]): {
   // otherwise the date blocks the $-anchored location strip below.
   institution = stripInstitutionDate(institution);
 
+  // Peel a trailing GPA/CGPA note glued on with no comma boundary
+  // ("… State University GPA 3.8", #556) — otherwise it rides into the school
+  // name (the comma-delimited GPA cut in `cleanField` only touches the degree
+  // field, not the institution).
+  institution = stripInstitutionGrade(institution);
+
   // Peel a trailing "City, ST" / "City, Country" off the institution so it isn't
   // glued on ("University of Example, …   Seattle, WA" → institution without the
   // location, location surfaced separately).
@@ -866,6 +961,10 @@ export function extractEducation(
   // to the nearest preceding degree below.
   const ls = education.lines;
   const coursework: { text: string; idx: number }[] = [];
+  // A real attendance-date range peeled off the END of a coursework line (#555),
+  // tagged with the source-line `idx` so it can be attributed to the same entry
+  // the courses on that line belong to.
+  const courseworkDates: { date: string; idx: number }[] = [];
   const consumed = new Set<number>();
   for (let i = 0; i < ls.length; i++) {
     // Coursework is usually a bullet list, but some résumés write it as a plain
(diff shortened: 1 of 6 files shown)


src/lib/heuristics/extract/education.ts
```diff
@@ -143,23 +144,57 @@ function educationDateFields(
...
+ *
+ *  TAIL-DATE carve-out (#555): a non-degree annotation line whose real
+ *  attendance date rides at its END — "Coursework: Databases, Data Mining.
+ *  Aug 2023 – May 2025" — keeps just that trailing date (the annotation prose
+ *  is still dropped) so the date reaches `parseEducationDates` instead of being
+ *  discarded with the line. An annotation line with no trailing date is dropped
+ *  as before, so the #371 over-capture guard is unchanged for that shape. */
 function filterAnnotationLinesForDates(lines: readonly string[]): string[] {
-  return lines.filter((l) => DEGREE_RE.test(l) || !EDUCATION_ANNOTATION_RE.test(l));
+  return lines.flatMap((l) => {
```
s-annam: This carve-out reopens the #371 over-capture class for month-year-precision annotation ranges (an awards/thesis/honours line ending in e.g. `Sep 2019 – Jun 2020`), silently overriding a real single-year graduation date on the sibling degree line — `parseDateRange` prefers the range. It also lacks the `courseworkDates` path's "only when the entry has no date of its own" guard, and is unnecessary: reverting this function to the `main` `filter(...)` one-liner keeps every PR test green (the #555 coursework test is handled entirely by the guarded `courseworkDates` path). See the top-level Blocking finding for the repro. Suggest reverting to the `main` one-liner.

s-annam (changes requested):
Three well-scoped, well-documented fixes with value-asserting tests — the #556 (GPA-glued-institution) and #557 (connective-less `of <Type> <Field>`) changes are clean and I have no concerns with them. But the #555 change ships a **second, unguarded date-recovery path** that silently reopens the #371 over-capture class. Verdict rule applied: **≥1 Blocking → REQUEST_CHANGES**.

## Blocking

**`filterAnnotationLinesForDates` reopens #371 for month-year annotation ranges — and is unnecessary.** (`src/lib/heuristics/extract/education.ts:192`)

The `#555` fix has two independent mechanisms:

1. the `courseworkDates` path in `extractEducation`, which attributes a recovered tail-date **only when the entry parsed no date of its own** (`if (!entry.start_date && !entry.end_date && !entry.year)`) — the correct guard; and
2. this new `flatMap` carve-out, which peels a trailing month-year range off **any** non-degree annotation line (awards / honours / thesis / dean's list / scholarships / coursework …) and feeds it straight into the chunk's `parseEducationDates` — **with no such guard**.

Because `parseDateRange` prefers a range over a bare year (the exact reason this function was written for #371), mechanism 2 lets an annotation range override a real single-year graduation date on the sibling degree line. Verified against `main` vs this branch:

```
Education section:
  "B.S. Computer Science, Some University, 2017"
  "Awards: Best Thesis, Sep 2019 – Jun 2020"

main branch:  year = 2017                         (correct — awards range dropped)
this PR:       year = 2020, start Sep 2019, end Jun 2020   (REGRESSION)
```

The narrowing to `STRICT_MONTH_YEAR_RE` limits the blast radius — a **year-only** annotation range (`Dean's List 2013–2015`, the common case) still starts on a bare year and is correctly dropped, which is why the existing #371 test stays green and no corpus fixture caught this. But a month-year-precision award/thesis range is a plausible real shape, and it silently corrupts a scored field (`year` / `start_date` / `end_date`).

Crucially, **mechanism 2 is not needed for #555**: reverting this function to its `main` one-liner —

```ts
function filterAnnotationLinesForDates(lines: readonly string[]): string[] {
  return lines.filter((l) => DEGREE_RE.test(l) || !EDUCATION_ANNOTATION_RE.test(l));
}
```

— leaves **all** PR tests green, including `education.coursework-tail-date.test.ts` (that test's coursework line is `consumed` and handled entirely by the guarded `courseworkDates` path). There is also no positive test exercising this carve-out's date-recovery, only the #371 test asserting the drop still happens.

**Fix:** revert `filterAnnotationLinesForDates` to the `main` filter (drop the `flatMap` carve-out and the `TAIL-DATE carve-out` docblock paragraph). If a future case genuinely needs annotation-tail recovery through the chunk path, it must carry the same "only when the entry has no date of its own" guard the `courseworkDates` path already has.

## Nits

- `const END` (`education.ts`, near the `TRAILING_ATTENDANCE_DATE_RE` definition) is a very generic identifier at module scope; something like `ATTENDANCE_RANGE_END` reads better next to its sibling regexes. Non-blocking.

## Gates

- **Fixture PII (3a):** N/A — no binary fixtures added/changed; the two `.expected.json` edits are `extractedCharCount`-only (−13 / −15 chars), consistent with the claimed strict improvements. `npm run check:fixtures` scope unaffected.
- **Design-system / style-token (3b/3c):** N/A — pure `src/lib/` change, no feature/UI code.
- **Skill/script command bugs (3e):** N/A — no `.claude/skills` / `scripts` files touched.
- **Dead-code / fallow (3d):** clean — all new symbols (`trailingAttendanceDate`, `stripInstitutionGrade`, `DEGREE_TYPE_RE`, `INSTITUTION_GRADE_RE`, `courseworkDates`) have in-module consumers.
- `npm run typecheck` clean; `npx vitest run src/lib/heuristics/` green (262 tests); corpus suite green (167 passed).

---
Reviewed by: Claude Opus 4.8 (1M context) (high)

src/lib/heuristics/extract/education.ts
rohithgollapalli: Agreed — reverted in `b069489`.

`filterAnnotationLinesForDates` is back to the `main` one-liner; the `flatMap` carve-out and its `TAIL-DATE carve-out` docblock paragraph are gone. Confirmed your read: all PR tests stay green with the revert (heuristics suite 1075 passed), so the #555 fix rides entirely on the guarded `courseworkDates` path.

Two follow-ups on top of the plain revert:

1. **Pinned your repro as a test.** Your exact shape is now the third case in `education.coursework-tail-date.test.ts` — `"B.S. Computer Science, Some University, 2017"` + `"Awards: Best Thesis, Sep 2019 - Jun 2020"`, asserting `year === "2017"` and `start_date === undefined`. Verified it is a real guard, not a passing-either-way test: temporarily restoring the `flatMap` carve-out fails it with `expected '2020' to be '2017'` — the same values you reported.
2. **Docblock now records the negative.** The removed paragraph is replaced with an explicit "no tail-date carve-out here, deliberately" note giving your reason (`parseDateRange` prefers a range over a bare year; the whole-chunk path cannot express the dateless-entry guard) and pointing at `courseworkDates` as the path that can — so this doesn't get re-added later.

`trailingAttendanceDate` / `TRAILING_ATTENDANCE_DATE_RE` stay live (still consumed by the coursework collector), so no dead code; its docblock was updated to stop referencing the removed path.

Nit also taken: `END` → `ATTENDANCE_RANGE_END`.

s-annam (approved):
Blocker cleared — approving. `filterAnnotationLinesForDates` is back to the `main` one-liner, the `flatMap` carve-out and its docblock paragraph are gone, and the `#555` fix now rides entirely on the guarded `courseworkDates` path. The `END` → `ATTENDANCE_RANGE_END` nit is taken.

Two things you did beyond the revert are worth calling out, because they're the difference between "fixed" and "can't regress":

1. **You pinned my repro as a real test, and proved it was real.** `education.coursework-tail-date.test.ts:78` is the exact `"B.S. Computer Science, Some University, 2017"` + `"Awards: Best Thesis, Sep 2019 - Jun 2020"` shape. I mutation-tested it rather than take the claim: restoring the `flatMap` carve-out fails it with `expected '2020' to be '2017'` — the same values I reported. It is a genuine guard, not a test that passes under both designs.
2. **The docblock now records the negative.** "No tail-date carve-out here, deliberately," with the reason (`parseDateRange` prefers a range over a bare year; the whole-chunk path cannot express the dateless-entry guard) and a pointer to the path that can. That's what stops this being re-added in six months by someone reading only the code.

## Verification

- `npx vitest run src/lib/heuristics/` on `b069489` in a clean worktree: **1076 passed, 8 skipped**, green. CI `verify` + `fallow` + Cloudflare Pages all SUCCESS on the same commit.
- **Corpus goldens verified by parsing, not by prose.** The snapshots are lossy by design — they carry `extractedCharCount`, never field values — so "each a strict improvement" is not readable from the diff. Ran both fixtures on `main` vs this branch and diffed the education entries directly:

  | Fixture | Field | `main` | this PR |
  |---|---|---|---|
  | `google-docs-skia-proxy-additional-skills` | `degree` | `Bachelor of Engineering Expected May` | `Bachelor of Engineering` |
  | `compound-certifications-activities-tail` | `institution` | `Ridgemont College · GPA 8.0 / 10` | `Ridgemont College` |

  −13 / −15 chars, matching the golden deltas exactly. Both are strict improvements; no other field moved.
- `#557` scope confirmed safe: the `DEGREE_TYPE_RE` cut sits inside the `else` branch (taken only when the match carries no ` in `), so the US `"… of Science in Biology"` form never reaches it.
- `TRAILING_ATTENDANCE_DATE_RE` probed against a word-eating tail (`"Coursework: Algorithms, Marketing Analytics"`) and a non-date year tail (`"… B since 2020"`) — neither is peeled. The `STRICT_MONTH_YEAR_RE` choice is doing its job (#380).

## Gates

- **Fixture PII (3a):** N/A — no binary fixtures added or changed; the two `.expected.json` edits are `extractedCharCount`-only. Every repro string in the three new tests is a synthetic persona.
- **Design-system / style tokens (3b/3c):** N/A — pure `src/lib/` change.
- **Skill/script command bugs (3e):** N/A.
- **Dead code / fallow (3d):** clean — `trailingAttendanceDate`, `TRAILING_ATTENDANCE_DATE_RE`, `stripInstitutionGrade`, `INSTITUTION_GRADE_RE`, `DEGREE_TYPE_RE`, `courseworkDates` all have live in-module consumers.

## Nits — not blocking, do not respin for these

Fold into a future education-parser touch if you're in the file anyway.

1. **`INSTITUTION_GRADE_RE`'s docblock over-claims.** It says it covers "a percentage (`GPA 85%`)", but `\d(?:\.\d+)?` matches a single leading digit, so `GPA 85%` and `CGPA 10/10` are both left glued on while `GPA 3.85%` is stripped. This is a **claim that doesn't round-trip to the code**, not a defect worth chasing: the keyword gate means the shape Indian résumés actually use (`Percentage: 85%`) was never in scope anyway. Either widen to `\d{1,3}` — still keyword-gated, so no new cut risk — or drop the percentage from the docblock.
2. **Duplicated attribution scan.** The `courseworkDates` loop repeats the `coursework` loop's nearest-preceding-entry scan verbatim; a shared `nearestEntryIndex(built, idx)` would carry the "`built` is in document order, so the scan can stop early" comment once instead of twice.
3. The reverted `filterAnnotationLinesForDates` came back prettier-wrapped across three lines rather than `main`'s one-liner, so a function that logically didn't change still shows as a diff hunk. Cosmetic only.

---

PR #572, opened 2026-07-25
feat(skills): pr-ready — solicit PR review with ack detection and a deadline (#571)

### LINKED ISSUE #571: feat(skills): pr-ready — solicit PR review with ack detection and a deadline
## Problem

Asking for review on a PR is fully manual, and the ask carries no machine-readable state.
An author writes a chat message by hand, names some reviewers, and then has no way to
answer the only question that matters: **has anyone picked this up?**

Silence is ambiguous in both directions. A reviewer who *is* mid-review has no cheap way
to say so, so the author reads silence as "nobody is looking" — which lands badly for
someone who was halfway through. A reviewer who is *not* looking produces the identical
signal, so the author waits on nothing. Review also arrives at hours the author isn't
watching, so a real response can go unnoticed until the next day.

So the protocol gets re-improvised per PR, and the parts that make it work — a one-click
ack, a stated deadline, an explicit "review after the deadline still counts" escape hatch,
and named review slices so a large diff has an entry point — are the first things dropped
when writing the message by hand.

This issue adds `/pr-ready`: it runs that protocol and, critically, **detects the ack**.

## Placement

`.claude/skills/pr-ready/`, checked in — reviewers need to run it too, and the adjacent
workflow skills (`open-pr`, `revise-pr`, `pr-review`) already live there.

The skill needs a chat channel and a reviewer roster, so identity is **split out of the
skill**: the skill contains no IDs, logins, names, or team-specific prose, and config
lives in a git-ignored local file with a checked-in `.example` sibling.

### Reuse analysis

**Capability:** solicit review on an open PR and track whether it was acknowledged.

**Existing surfaces:** `open-pr` (creates the PR, terminates at open), `revise-pr` (the
loop *after* a reviewer leaves threads), `pr-review` (performing a review — the reviewer
side).

**Decision:** build new. Those three cover create → (gap) → review → revise. The gap
between "PR is open" and "a human engaged with it" is unowned, and it is the step that
fails. Extending `open-pr` was rejected: it would couple a one-shot creation flow to a
stateful poll loop with a different lifetime, and would fire on every PR including ones
nobody is being asked to look at.

### Config

Git-ignored `.claude/pr-ready.local.json`, checked in as
`.claude/pr-ready.local.json.example` with placeholder values only:

```json
{
  "channel": "<chat channel id>",
  "reviewers": [
    { "name": "<display name>", "chat_id": "<chat user id>", "gh_login": "<github login>" }
  ],
  "default_deadline_minutes": 90,
  "ack_reactions": ["eyes", "white_check_mark", "raising_hand"]
}
```

The `chat_id` → `gh_login` mapping is what makes cross-surface ack detection possible; it
cannot be inferred and must be declared.

### Phases

0. **Preflight.** Resolve the PR (argument, or infer from the current branch). Assert
   `state == OPEN`, `mergeable == MERGEABLE`, required checks green, and exactly one commit
   on the branch (this repo's merge-queue rule). Abort naming the specific failure —
   soliciting review on a red or multi-commit PR wastes reviewer attention and trains
   people to ignore the ask.

1. **Ping.** Post one message from a fixed template carrying: what changed and why it's
   worth a look, the PR link, reviewer mentions, **the ack mechanism stated explicitly**
   ("react 👀 and that's the ack"), **the deadline as an absolute local time**, **the
   escape hatch** (review after the deadline still counts; findings become issues, treated
   the same as PR comments), and **2–3 named review slices** — `path` plus the question to
   ask of it — so a large diff has an entry point. Record the message timestamp; every
   later phase keys off it.

2. **Wait.** Background sleep to the halfway point.

3. **Check + one reminder.** Evaluate the ack predicate. If acked, stand down and report
   who. If silent, post exactly one reminder as a threaded reply broadcast to the channel,
   then wait to the deadline.

4. **Report.** Re-evaluate and print a terminal summary: acked or not, by whom, PR state,
   author's options. Stop there.

### Ack predicate

Acked if **any** holds for **any** configured reviewer, timestamped **after** the ping:

1. a reaction on the ping message from that reviewer's `chat_id`, in the `ack_reactions`
   allowlist;
2. any threaded reply on the ping from that `chat_id`;
3. any PR activity by that `gh_login` — review (any state), review comment, or issue
   comment.

Bot and author activity must be excluded: the checks app, the deploy-preview bot, and the
code-scanning reviewer all post on PRs and none is an ack.

### Explicit non-goal: this skill never merges

No merge path, no merge-on-silence flag. Terminal state is a report; the merge decision
stays a human call made outside the skill.

Deliberate. Automating merge-on-silence institutionalizes "silence equals consent" —
exactly the dynamic that makes an engaged reviewer feel bypassed, which is the failure
this skill exists to fix. Encoding it in the tool also hides the decision; the author
should make that call explicitly, with the state in front of them, each time. The deadline
still does real work unattached to a merge: it bounds an open-ended ask, and the escape
hatch makes missing it cost-free for the reviewer.

## Files

| File | Status | Contents |
|---|---|---|
| `.claude/skills/pr-ready/SKILL.md` | new | phases, ack predicate, message template, failure modes |
| `.claude/pr-ready.local.json.example` | new | config schema, placeholder values only |
| `.gitignore` | edit | ignore `.claude/pr-ready.local.json` |
| `docs/CONTRIBUTING-PROCESS.md` | edit | short section: what the skill does + the one-time config setup |

## Acceptance criteria

- [ ] `.claude/skills/pr-ready/SKILL.md` exists, invocable as `/pr-ready [<pr-number>]`,
      inferring the PR from the current branch when the argument is omitted.
- [ ] The skill file contains no channel IDs, user IDs, GitHub logins, personal names, or
      organization-specific prose. `grep` finds the config keys only in the `.example`.
- [ ] `.claude/pr-ready.local.json` is git-ignored; `git check-ignore` confirms it.
- [ ] Preflight aborts, naming the specific reason, on each of: PR not open, not
      mergeable, a required check failing, more than one commit on the branch.
- [ ] The posted message includes all template elements: change summary, PR link, reviewer
      mentions, ack mechanism, absolute-time deadline, escape hatch, ≥2 named review slices.
- [ ] The ack predicate returns true for each of its three inputs independently — a
      reaction, a thread reply, PR activity — verified one at a time.
- [ ] The ack predicate returns **false** when the only post-ping activity is from a bot
      (checks app, deploy preview, code-scanning reviewer) or from the PR author.
- [ ] At most one reminder per run, regardless of wait length.
- [ ] No merge under any input: `grep -rE "pr merge|--admin"` over the skill directory
      returns nothing.
- [ ] Running twice against the same PR does not double-post the initial ping.
- [ ] `docs/CONTRIBUTING-PROCESS.md` documents the one-time config setup.

## Non-goals

- **Unattended operation.** The wait is session-bound; closing the session cancels the
  remaining phases. The skill must say so up front rather than implying a durable timer. A
  scheduled headless variant is a separate issue if ever wanted.
- **Escalation beyond one reminder.**
- **Reviewer assignment or load-balancing.** The roster is static config.
- **A second chat backend.** Keep the chat calls behind a thin, marked seam so one is
  possible later; don't build it now.

s-annam:
## Summary

Adds `/pr-ready`, a checked-in skill that runs the ask-for-review step as a protocol instead of a hand-typed message — and, critically, **detects the ack**. Preflight the PR (open, no conflicts, checks green, one commit) → post one ping carrying the ack mechanism, an absolute-time deadline, an explicit "late review still counts" escape hatch, and 2–3 named review slices → wait → at most one reminder if still silent → report. The ack predicate is satisfied by any of a reaction, a thread reply, or GitHub review activity, with bots and the PR author excluded.

It **never changes the PR's merge state**. That is deliberate, not an omission: automating merge-on-silence institutionalizes "silence equals consent," the exact dynamic that makes an engaged reviewer feel bypassed — which is the failure this skill exists to fix. The terminal state is always a report; the merge call stays a human's, made explicitly, each time.

Identity is split out of the skill so the checked-in file carries no channel IDs, user IDs, logins, or names — those live in a git-ignored `.claude/pr-ready.local.json`, created once per checkout from the `.example`. The `chat_id` → `gh_login` mapping is what makes cross-surface ack detection possible; it cannot be inferred, so it is declared.

The wait is **session-bound** — closing the session cancels it. The skill says so up front and in the ping, rather than implying a durable timer.

The ask takes a **list**: `/pr-ready 572,605,606` is one ping covering all three, with one checkpoint, one reminder, one deadline, and a report that breaks down per PR. Three separate runs would @-mention the same reviewers three times and hold three sequential deadlines — being chased, not nudged. A single-PR ask is a list of one and is unchanged. Three points needed a decision rather than a loop, and each is argued at the phase that owns it:

- **Preflight partitions instead of aborting** — one red PR drops out and the ready ones still go; the exclusions print at the confirmation *and* in the report, because a reviewer who sees two PRs assumes two is all there is.
- **The idempotency scan matches the set exactly** — a partial overlap with an earlier ping stops the run and names it. Posting would re-ping people about a PR already asked about; adopting would wait on a ping that never mentioned the ones just added. Merging two asks into one thread is a third behaviour nobody specified, so it isn't invented here.
- **The ack predicate splits by scope** — a reaction or thread reply lands on the ping (which names every PR), so it is ask-level and can't be attributed to a row; GitHub activity is per PR. The reminder consumes the union: it fires only when *nothing* has been acked, so someone who just reviewed one PR is never told they still haven't looked.

| File | Status |
|---|---|
| `.claude/skills/pr-ready/SKILL.md` | new — 7 phases, ack predicate (ask-level + per-PR), multi-PR ask, message template, chat seam, failure modes |
| `.claude/pr-ready.local.json.example` | new — config schema, placeholder values only |
| `.gitignore` | ignore `.claude/pr-ready.local.json` |
| `docs/CONTRIBUTING-PROCESS.md` | new section: what the skill does + the one-time setup |

Closes #571

## Test plan

- [x] `npm run lint` clean
- [x] `node scripts/check-fixture-pii.mjs` — 54 PDFs, all personas synthetic (no fixtures added by this PR)
- [x] AC 3 — `git check-ignore` confirms the local config is ignored and the `.example` is not
- [x] AC 9 — `grep -rE "pr merge|--admin" .claude/skills/pr-ready/` empty, with the never-merges constraint still plainly stated in four places
- [x] AC 6/7 — ack predicate exercised against real PRs: bot-only and author-only post-ping activity both filter to empty; a genuine non-author reviewer surfaces
- [x] No `src/` change, so no test suite is affected and fallow has nothing to audit
- [x] Re-ran `npm run typecheck` + `npm run test` after the multi-PR change — clean / 2854 passed
- [x] Multi-PR path exercised by hand on #572 + #605 + #606: per-PR preflight, empty-channel idempotency scan, and the composed three-block message

**Reviewer note.** This is executable prose — a wrong flag, jq path, or MCP tool name is a real bug, and three rounds of review found exactly those. Worth a skeptical pass on Phase 0's tool resolution and the ack predicate's three `gh api` snippets. Known-surviving nits are listed under `## Adversarial review` below.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #571 | Claude Sonnet 5 | high |
| Adversarial review — round 1 | Claude Opus 5 | high |
| Review fixes — round 1 | Claude Opus 5 (1M context) | high |
| Adversarial review — round 2 | Claude Opus 5 (1M context) | high |
| Review fixes — round 2 | Claude Opus 5 (1M context) | high |
| Adversarial review — round 3 | Claude Opus 5 (1M context) | high |
| Review fixes — round 3 + orchestration + PR | Claude Opus 5 (1M context) | high |
| Review fixes — round 4 + multi-PR ask | Claude Opus 5 (1M context) | high |

## Adversarial review

Three adversarial rounds ran against the working tree before this PR opened. Every
finding below was reproduced by running the command, not by reading the file — three
separate defects survived a fix pass and reappeared in a new form, which is why the
loop ran to three rounds instead of the usual two.

### Round 1 — 8 blocking

| # | Defect | Evidence |
|---|---|---|
| 1 | Phase 3 checkpoint arithmetic inert. `date -j -f '%s' "<ts>" +%s \|\| date +%s` returns rc=0 on **both** a real Slack ts and an empty string (yielding `0`), so the `\|\|` guard was unreachable. An unrecorded ts put the checkpoint in Jan 1970 → the `until` loop exited immediately → the reminder fired seconds after the ping. | Ran verbatim: empty ts → `halfway=2700`, `diff_min=-29748990` |
| 2 | Author exclusion compared `.user.login` against a JSON **object**, so it never matched — the PR author's own comments counted as an ack (breaks AC 7). | `gh pr view 553 --json author -q .author` → `{"id":…,"login":"…","name":…}` |
| 3 | Every Slack tool named unnamespaced (`slack_send_message`), so none was callable — skill unrunnable. Parameter shapes were all correct; only the names were wrong. | Loaded real schemas; names are `mcp__<plugin>__slack_*` |
| 4 | No confirmation step anywhere before posting an @-mention broadcast to a shared channel. The send tool's own contract requires a draft when the user hasn't reviewed the text. | Tool schema, verbatim |
| 5 | Phase 5 was dead text; three phase cross-references pointed at phases doing something else. | Traced both control-flow branches |
| 6 | Preflight aborted on `mergeable == "UNKNOWN"` with a conflict-specific message — GitHub computes mergeability async, so a freshly-pushed PR (exactly when `/pr-ready` runs) false-aborts and sends the author chasing a nonexistent conflict. | `gh pr view 553 --json mergeable,state` → `{"mergeable":"UNKNOWN","state":"MERGED"}` — zero conflicts |
| 7 | Check gate fired only on `conclusion == "FAILURE"`, so a still-running check (`conclusion: null`) or a `STARTUP_FAILURE` / `TIMED_OUT` / `CANCELLED` passed as green. This repo has hit `startup_failure`. | Rollup shape verified across 5 PRs |
| 8 | Idempotency scan used a public-only, consent-gated search tool with a malformed `in:` modifier for a channel **id** — on a private channel it silently did nothing and double-posted every run. | Tool schemas + documented modifier forms |

### Round 2 — 2 blocking (6 of 8 held)

- **Phase 0 aborted on every run.** The fix used `ToolSearch("select:slack_read_channel,…")`, but `select:` matches exact deferred names only → `No matching deferred tools found`, which the skill's own abort path reads as "no chat backend". The skill could never reach Phase 2.
- **The author exclusion returned in a new form.** `gh api` has no `--arg`, so `$AUTHOR` was an undefined jq variable — `variable not defined: $AUTHOR`, rc=1. Worse, the three snippets carried *neither* the bot nor the author filter, so a copy-paste runner applied no exclusion at all.
- Also caught: the epoch guard `''|*[!0-9]*` admitted literal `0` — the exact Jan-1970 value it existed to reject.

### Round 3 — 2 blocking (fixes verified by execution; 1 latent defect found)

- **`slack_read_channel` lookback bounded at exactly the deadline window**, making any ping older than the deadline invisible to the idempotency scan → a re-run the next morning double-posts (**AC 10 fail**), and the Phase 2 → Phase 6 short-circuit added the round before was dead code, since "found in window" and "deadline lapsed" were mutually exclusive.
- **`ToolSearch` default `max_results` is 5**, and the Phase 0 keyword lookup needs 5 specific tools out of a larger Slack surface — `slack_get_reactions` ranks 6th and dropped off, so the step resolved 4 of 5 and the reaction half of the ack predicate had no callable tool behind it.
- The round-2 fix pass also found, unprompted, that **Phase 5 had no epoch guard at all** — only prose claiming it did.

Both round-3 blockers are fixed and verified by execution: with `LOOKBACK = max(86400, 4 × deadline)` a ping 91m / 150m / 1400m old is simultaneously in-window and lapsed, so the Phase 2 → 6 edge is live; `max_results: 10` resolves all five tools and the `select:` second pass is now mandatory rather than optional.

### Surviving nits — for the human reviewer

- **`.conclusion` vs `StatusContext.state`** — gated on `(.conclusion // .state)`. Latent only: every rollup entry in this repo is a `CheckRun`.
- **The check gate and the `UNKNOWN` retry are prose, not executable snippets**, unlike Phases 2/3/5. Unambiguous, but they don't match the file's own register.
- **`slack_get_reactions` truncates at 50 users per reaction** — caveated in the file ("treat a miss as inconclusive, not not-acked"), not solved.
- **AC 2, literal reading, is not met.** The config key *names* (`chat_id`, `gh_login`, `ack_reactions`) necessarily appear in SKILL.md prose because the phases describe what they read. The duplicated schema block is gone and no *values* leak — an identity sweep over the skill dir and the `.example` for Slack-shaped ids, emails, logins and personal names returns nothing. Judged defensible; flagging rather than hiding it.

### Gates

`npm run typecheck` rc=0 · `npm run lint` rc=0 · `node scripts/check-fixture-pii.mjs` → 54 PDFs, all personas synthetic · `grep -rE "pr merge|--admin" .claude/skills/pr-ready/` empty (AC 9), with the never-merges constraint still plainly stated in four places · `git check-ignore` confirms the local config is ignored and the `.example` is not · all 37 `Phase N` references resolve to one of the 7 `### Phase` headings; both branches trace with at most one reminder on every path.

No `src/` change, so no test suite is affected and fallow has nothing to audit.

### DIFF
--- a/.claude/skills/pr-ready/SKILL.md
+++ b/.claude/skills/pr-ready/SKILL.md
@@ -1,0 +1,699 @@
+---
+name: pr-ready
+description: Solicit review on one or more open PRs — post one ping with an explicit ack mechanism and an absolute deadline, wait, send at most one reminder, and report per PR whether anyone engaged. Detects the ack across chat and GitHub. Never changes a PR's merge state. Use when the user says "pr-ready", "/pr-ready", "ask for review on this PR", "ping reviewers", or has open PRs nobody has looked at yet.
+argument-hint: [<pr-number>[,<pr-number>...]]
+---
+
+# PR Ready
+
+Take a PR — or a list of them — from "open and silent" to "a human has visibly
+engaged, or the author knows exactly who to nudge": preflight → post one ping
+with a stated ack mechanism and an absolute-time deadline → wait → check once
+at the reminder checkpoint and send at most one reminder if still silent →
+wait to the deadline → report. A list is **one** ask: one ping, one deadline,
+one reminder, one report that breaks down per PR. This skill only ever produces a **report** — the merge decision stays
+with a human, made outside this skill, every time.
+
+This is the missing middle of the review loop: `open-pr` creates the PR and
+stops; `pr-review` is the reviewer performing a review; `revise-pr` is the
+author closing out threads afterward. Nothing today owns "did anyone actually
+pick this up" — silence reads identically whether a reviewer is mid-review or
+not looking at all, and the parts that make an ask work (a one-click ack, a
+stated deadline, an escape hatch for late review, named entry points into a
+large diff) are exactly what get dropped when the ask is typed by hand.
+
+## One-time setup
+
+This skill needs a chat channel and a reviewer roster; that identity is kept
+out of the skill file entirely (see Rules). Copy the example and fill it in
+once per checkout:
+
+```bash
+cp .claude/pr-ready.local.json.example .claude/pr-ready.local.json
+```
+
+Then edit `.claude/pr-ready.local.json` — it is git-ignored, so it never
+leaves your machine. The schema lives in one place only:
+**`.claude/pr-ready.local.json.example`**; read that file for the full key
+list and fill in every placeholder.
+
+The one key worth explaining here: the per-reviewer `chat_id` → `gh_login`
+mapping is what makes the ack predicate work across both surfaces — a chat
+reaction and a GitHub comment identify the same human under two different
+identifiers, and nothing can infer the link, so it must be declared. If
+`.claude/pr-ready.local.json` is missing, stop and point at this section; do
+not guess a channel or roster.
+
+## Non-goals — read before running
+
+- **This skill never changes what a PR can do next.** Its terminal state is
+  always a printed report: acked-or-not, by whom, and the PR's current state.
+  There is no code path anywhere in this skill that flips the PR's merge
+  readiness, bypasses a required check, or treats silence as a decision.
+  That call is the author's, made explicitly, outside this skill, every time —
+  automating it would institutionalize "silence equals consent," which is the
+  exact dynamic that makes an engaged reviewer feel bypassed when it happens.
+- **The wait is session-bound, not a durable timer.** All of Phases 2–6 run as
+  long as this session stays open. Closing the session before the deadline
+  cancels the remaining wait and reminder — there is no background job that
+  outlives it. A scheduled/headless variant is a different tool, not this one.
+- **At most one reminder, ever**, regardless of how long the wait runs — and
+  one per *ask*, not one per PR. A list of three PRs is one ping, one
+  checkpoint, one reminder, one report.
+- **No cross-PR judgement.** With a list, this skill reports which PRs were
+  acked and which weren't. It does not rank them, decide which deserves review
+  first, or drop one for being less important — the order is the author's, and
+  the only PRs it removes are the ones the preflight found unready.
+- **No reviewer assignment or load-balancing.** The roster in the config is
+  static; this skill doesn't pick who should review, only whether the people
+  already named have engaged.
+- **One chat backend, on a marked seam.** All chat calls are isolated behind
+  the seam in Phase 2, Phase 4, and the ack predicate so a second backend is
+  possible later — none is built now.
+- **Nothing is posted without an explicit go-ahead.** Both outbound messages —
+  the Phase 2 ping and the Phase 4 reminder — are shown to the user in full
+  and sent only after they say so. Posting to a shared channel with
+  @-mentions is outward-facing and effectively irreversible; the skill never
+  composes and sends in one motion.
+
+## Input
+
+Parse the argument for **one or more PR numbers** — `123`, `#123`, or a
+comma-separated list (`123,456,789`). Deduplicate, and preserve the order
+given: it is the order the reviewer will read them in, so it is the author's
+call, not something to re-sort. If omitted, infer a single PR from the current
+branch:
+
+```bash
+gh pr view --json number,headRefName,state -q '{n:.number,head:.headRefName,state:.state}'
+```
+
+If that finds no PR and none was passed, list open PRs and ask which one.
+Never guess.
+
+**A list is ONE ask, not N runs of this skill.** Every phase below operates on
+the whole set: one ping naming every PR, one checkpoint, one reminder, one
+deadline, one report. Running the skill three times instead would @-mention the
+same people three times for what is a single request for attention, and would
+hold three sequential deadlines — which is what a reviewer experiences as
+nagging. The single-PR case is just a list of one and behaves exactly as it
+always did; nothing below special-cases it.
+
+**Where the set changes the semantics** — three places, all of them called out
+again at the phase that owns them:
+
+- **Preflight is per PR, and partitions rather than aborts** (Phase 1). One
+  unready PR must not block the ask for the ready ones.
+- **The idempotency scan matches the set exactly** (Phase 2). A partial overlap
+  with an earlier ping stops the run rather than guessing.
+- **The ack predicate is per PR for GitHub activity, ask-level for chat**
+  (below). A 👀 on the ping cannot be attributed to one PR in the list.
+
+## Process
+
+### Phase 0: Resolve repo, PR, and config
+
+```bash
+REPO="$(gh repo view --json nameWithOwner -q .nameWithOwner)"   # offlinecv/OfflineCV
+```
+
+Load `.claude/pr-ready.local.json`. If it's missing or fails to parse, stop —
+see One-time setup.
+
+**Resolve the chat tool names before any chat call.** The chat MCP tools are
+namespaced by the plugin that provides them — the pattern is
+`mcp__<slack-plugin>__slack_*`, and the `<slack-plugin>` segment differs
+between checkouts, so it must never be hardcoded. Resolve the exact callable
+names once, here, with a **keyword** lookup:
+
+```
+ToolSearch("slack read channel reactions thread send message draft", max_results: 10)
+```
+
+**Pass `max_results`.** The default is 5, and this lookup needs 5 *specific*
+tools out of a much larger Slack surface — at the default, `slack_get_reactions`
+ranks 6th and drops off the end (`slack_schedule_message` comes back in its
+place), so the step silently resolves 4 of 5 and the reaction half of the ack
+predicate has no callable tool behind it. Verify all five are present in the
+result before continuing; if one is missing, widen `max_results` rather than
+proceeding with a partial set.
+
+**It must be the keyword form, not `select:`.** `select:` matches deferred tool
+names *exactly*, and no tool is registered under the bare name
+`slack_read_channel` — every one carries its plugin prefix. So
+`ToolSearch("select:slack_read_channel,…")` returns `No matching deferred
+tools found` on every run, which the paragraph below would then read as "no
+chat backend" and abort the skill before it ever posts. The keyword form
+searches descriptions and resolves all five with their real namespaced names.
+
+Read the five namespaced names out of that result — they share one
+`mcp__<slack-plugin>__` prefix, whatever this checkout's plugin id happens to
+be. **Never hardcode a prefix**, including one you saw in a previous run;
+discovering it is the entire point of this step. Then re-fetch all five by
+exact name — valid now that the names are the real ones, and the step that
+proves you have every tool this skill needs rather than most of them:
+
+```
+ToolSearch("select:<prefix>slack_read_channel,<prefix>slack_send_message,<prefix>slack_send_message_draft,<prefix>slack_get_reactions,<prefix>slack_read_thread")
+```
+
+Every `slack_*` name written in a **"Chat seam."** block below is shorthand
+for the fully-namespaced name this lookup returns. If the **keyword** lookup
+surfaces no slack tools at all, stop and say the chat backend is unavailable —
+don't substitute a different tool. An empty `select:` result is not that
+signal; only an empty keyword result is.
+
+### Phase 1: Preflight
+
+Fetch everything the preflight needs in one call, recording the author's
+**login** explicitly (`.author` is an object; the ack predicate compares
+against the login string, not the object):
+
+```bash
+for PR_NUM in <every PR in the list>; do
+  gh pr view "$PR_NUM" --repo "$REPO" \
+    --json state,mergeable,statusCheckRollup,commits,author,url,title -q '.'
+done
+AUTHOR=$(gh pr view <the first PR in the list> --repo "$REPO" --json author -q .author.login)
+```
+
+**One `$AUTHOR`, not one per PR.** The author exclusion in the ack predicate
+answers "is this the person doing the asking?", and the person running this
+skill is the same human for every PR in one ask. Capture it once. If the PRs in
+the list turn out to have *different* authors, say so at the confirmation step:
+the ask is then on someone else's behalf, and their activity on their own PR is
+a legitimate ack under this `$AUTHOR` — visible, not silently reinterpreted.
+
+**Exclude, don't abort.** Each PR is checked independently against the table
+below; a failing PR is **dropped from the ask** and every passing one still
+gets pinged. Aborting the whole run because one of three PRs went red would
+hold back two ready ones and cost a round-trip for no reviewer benefit. Two
+guards keep the exclusion from being silent, which is the failure mode that
+matters — a reviewer who sees two PRs assumes two is all there is:
+
+- **List every excluded PR and its specific reason at the Phase 2
+  confirmation**, above the composed message, before anything is sent. The user
+  approves the ask *and* the drops in the same breath.
+- **Repeat the exclusions in the Phase 6 report**, so the terminal artifact
+  says what was never asked about rather than reading as full coverage.
+
+**Abort only when the eligible set is empty** — with nothing to ask about there
+is no ask, and the abort message is every PR's reason, not just the first.
+
+| Condition | Check | Exclusion message |
+|---|---|---|
+| Not open | `.state != "OPEN"` | "PR #N is `<state>`, not open — nothing to solicit review on." |
+| Conflicting | `.mergeable == "CONFLICTING"` | "PR #N has merge conflicts — resolve them before asking for review." |
+| Mergeability unknown after retries | `.mergeable == "UNKNOWN"` on every attempt (see below) | "GitHub hasn't computed mergeability for PR #N yet (`UNKNOWN` after `<n>` tries) — re-run in a minute." |
+| A check is not green | any `.statusCheckRollup[]` whose `(.conclusion // .state)` is not in `SUCCESS` / `NEUTRAL` / `SKIPPED` | "PR #N has a check that isn't green: `<check name>` is `<conclusion or "still running">` — wait for it or fix it; a red or half-finished PR trains reviewers to skip the ask." |
+| More than one commit | `.commits \| length > 1` | "PR #N has `<n>` commits — collapse to one before asking for review (see `open-pr` Step 3.6)." |
+
+**`UNKNOWN` is not a conflict.** GitHub computes mergeability asynchronously,
+so a freshly-pushed PR — exactly the state `/pr-ready` runs in, right after
+`/open-pr` — commonly reads `UNKNOWN` with zero conflicts. Re-poll the
+`mergeable` field up to 5 times with a short backoff (~5s) and only exclude
+the PR if it is still `UNKNOWN`, with the message above. Never report a conflict the PR
+doesn't have; sending the author to resolve an imaginary conflict is worse
+than waiting.
+
+**A check with `conclusion: null` is still running, not passing.** A queued or
+in-progress check run carries a null conclusion, and the terminal non-green
+conclusions (`TIMED_OUT`, `CANCELLED`, `ACTION_REQUIRED`, `STALE`,
+`STARTUP_FAILURE` — this repo has hit that last one) are none of them
+`FAILURE`. Gate on the green allowlist, not on a failure denylist, and name
+the check plus its actual state in the exclusion.
+
+Soliciting review on a PR that fails any of these wastes reviewer attention
+and trains people to ignore the next ask — the preflight exists to make that
+impossible, not to be a formality.
+
+### Phase 2: Post the ping (idempotent)
+
+**Check for an existing ping first** — running this skill twice against the
+same PR must not double-post:
+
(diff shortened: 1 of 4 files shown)


.claude/skills/pr-ready/SKILL.md
```diff
@@ -0,0 +1,510 @@
...
+   > heavily-reacted message a reviewer's ack can fall outside it — treat a
+   > miss there as inconclusive, not as "not acked", and let inputs 2 and 3
+   > decide.
+2. **Any threaded reply on the ping**, from that reviewer's `chat_id`.
+   > **Chat seam.** `slack_read_thread` on the ping's `channel_id` +
+   > `message_ts`; any reply whose `user` is the reviewer's `chat_id` counts.
+   > Same reasoning — a thread reply cannot predate its parent.
+3. **Any PR activity by that reviewer's `gh_login`** — a review of any state,
+   a review comment, or an issue-level PR comment — with a timestamp after
+   the ping's wall-clock time (convert the ping `ts` to an ISO timestamp once,
```
Samhit21: [Suggestion]: The ping→ISO conversion this comparison depends on is never specified, and the file's own `date` idiom would produce local time — silently counting pre-ping activity as an ack.

Input 3 keeps rows whose `at` is "later than the ping's ISO timestamp" (line 408), and points here to a conversion done "once, in Phase 2." But Phase 2 only records the raw epoch `ts` ("Record the returned message `ts`") — the epoch→ISO conversion is never given anywhere in the file. That leaves the one comparison in the whole ack predicate to an unspecified step, in a PR whose premise is that every executable step is exact.

It matters because GitHub returns `submitted_at`/`created_at` as **UTC Z-form** (`2026-07-13T21:51:50Z`), and the lexicographic compare is only valid if the ping is also UTC-Z. The only conversion idiom the file demonstrates is `date -r <epoch>` (line 252), which yields **local** time (`Mon Jul 27, 12:55 AM EDT` — no `Z`). An implementer reaching for that same idiom gets a local-time string, and comparing it against GitHub's Z-form shifts the boundary by the UTC offset (4–8h here). Concretely: a configured reviewer who commented on the PR three hours *before* the ping string-compares as "after" it, so the skill reports an ack the reviewer never gave — the exact "silence counted as engagement" this skill exists to prevent.

Pin the conversion explicitly in Phase 2, in UTC:

```bash
PING_ISO=$(date -u -r "${PING_TS%%.*}" +%Y-%m-%dT%H:%M:%SZ)   # UTC Z-form, so the lexicographic compare against GitHub's Z timestamps is valid
```

.claude/skills/pr-ready/SKILL.md
```diff
@@ -0,0 +1,510 @@
...
+- `<path 1>` — <the question to ask of it>
+- `<path 2>` — <the question to ask of it>
+```
+
+Render `*Deadline:*` as an absolute local time, not a duration — "in 90
+minutes" is useless to someone reading the message an hour later. Compute it
+once here, from the same arithmetic Phases 3 and 5 use:
+
+```bash
+date -r "$(( $(date +%s) + <deadline_minutes>*60 ))" "+%a %b %-d, %-I:%M %p %Z"
```
Samhit21: [Nit]: `date -r <epoch>` is BSD/macOS-only; it breaks verbatim on a Linux checkout.

On macOS `date -r <n>` means "epoch seconds" (works — emits the intended local time). On GNU coreutils `date -r` means "reference a file's mtime," so `date -r 1721865600` errors `No such file or directory` — GNU wants `date -d @1721865600`. This is the only `date -r` in the skills tree, so there's no repo convention to lean on. Low severity since this is an interactive macOS-targeted tool, not CI-run — but it's checked into a public repo, so a Linux contributor's run fails here. Either note the macOS assumption, or use a portable form. (Same flag would be the one to reach for in the Phase 2 conversion above — worth fixing both to a portable form or a documented assumption.)

Samhit21 (commented):
## PR Review: pr-ready — solicit PR review with ack detection and a deadline (#571)

### Summary

First human review, on `c4ba928`. This is 510 lines of executable prose, and it shows the three adversarial rounds: I read the load-bearing phases directly and ran an execution pass over the snippets, and the parts most likely to be wrong are right. Verified correct by execution, not just reading:

- **The three ack `gh api` snippets** — paths and shapes exist; the bot (`.user.type != "Bot"`) and author (`.user.login != "$AUTHOR"`, login-to-login) exclusions both fire against a real PR; all three agree byte-for-byte; the double-quote-not-`--arg` rationale genuinely prevents the `variable not defined: $AUTHOR` hard-error; and reactions correctly go through Slack, not a `gh .../reactions` endpoint.
- **Phase 0 tool resolution** — all five Slack tool names map to real MCP tools, and every tool used in a later phase is resolved.
- **Timing** — the `0*` epoch guard rejects `0`/`00`/`0.000000` while passing a real 10-digit `ts`; no immediate-fire or never-fire path.
- **Idempotency** (LOOKBACK ≥ 86400, dedup keyed on the stable PR URL, adopted-`ts` reuse), **config parity** (every key read exists in the `.example`), **`.gitignore`** (`git check-ignore`: local ignored, `.example` not), and the **never-merges invariant** (zero merge-mutating calls anywhere) all hold.

Two net-new items, neither blocking, both about `date`/timestamps.

### Findings

1. `SKILL.md:392` — the ping→ISO conversion input 3 compares against is never specified, and the file's own `date -r` idiom yields local time; a local-vs-UTC lexicographic compare counts pre-ping activity in the [ping − offset, ping] window as an ack. Core predicate, so worth pinning to UTC-Z explicitly. (inline)
2. `SKILL.md:252` — `date -r <epoch>` is BSD-only; breaks on a Linux checkout. Low severity for a macOS-interactive tool, but net-new. (inline)

I did not re-raise the four surviving nits the PR body already acknowledges.

### Verdict

**Action: COMMENT.**

---

.claude/skills/pr-ready/SKILL.md
s-annam: Fixed in `b560842`.

You're right that the conversion was never written down, and that the file's own `date -r` idiom would have produced local time. Phase 2 now pins it explicitly, right after the send seam, and covers the adopted-ping path too — both paths feed the same `PING_ISO`, which the old prose implied but never located:

```bash
PING_TS="<the ts just returned, or the one adopted from an existing ping>"
case "${PING_TS%%.*}" in
  ''|0*|*[!0-9]*) echo "pr-ready: PING_TS is empty, zero, or non-numeric ('$PING_TS') — cannot convert the ping timestamp" >&2; exit 1 ;;
esac
PING_EPOCH="${PING_TS%%.*}"
PING_ISO=$(date -u -r "$PING_EPOCH" +%Y-%m-%dT%H:%M:%SZ 2>/dev/null \
        || date -u -d "@$PING_EPOCH" +%Y-%m-%dT%H:%M:%SZ)   # BSD/macOS, then GNU
```

Prose beside it states why `-u` is load-bearing, using your concrete failure: at `-04:00` a comment left three hours *before* the ping sorts as after it, so the skill reports an ack nobody gave. Input 3 now names `PING_ISO` instead of "the ping's ISO timestamp", and says the lexicographic compare is valid only because both sides are UTC Z-form of the same width — that being the only reason no date parsing is needed on that path.

One thing I added beyond what the thread asked: the `PING_TS` guard from Phases 3 and 5 is repeated here. Without it a `0` converts to `1970-01-01T00:00:00Z` and every comment on the PR sorts after it — the same false-ack outcome by a different route, and the guard is already the file's answer to that class.

Verified the BSD branch on macOS: `1721862000` → `2024-07-24T23:00:00Z`, and `[ "2026-07-13T21:51:50Z" \> "$PING_ISO" ]` is true under `sh`. The GNU branch I could **not** execute — no coreutils on this box — so `date -d @<epoch>` there rests on the spelling, not on a run.

.claude/skills/pr-ready/SKILL.md
s-annam: Fixed in `b560842` — took the portable form rather than documenting the macOS assumption, and fixed both sites as you suggested.

The deadline render now reads:

```bash
DEADLINE_EPOCH=$(( $(date +%s) + <deadline_minutes>*60 ))
date -r "$DEADLINE_EPOCH" "+%a %b %-d, %-I:%M %p %Z" 2>/dev/null \
  || date -d "@$DEADLINE_EPOCH" "+%a %b %-d, %-I:%M %p %Z"
```

and the new Phase 2 UTC conversion uses the same `-r` → `-d @` pair. Hoisting `DEADLINE_EPOCH` out of the format call is what lets the fallback reuse it without recomputing `date +%s` (a second call would drift the two branches by a second — harmless here, but only by luck).

Prose beside it marks both branches load-bearing so a later editor doesn't "simplify" one away, and states the split explicitly: this site is deliberately **local** time because a human reads it in the channel, while `PING_ISO` is deliberately UTC because it's string-compared against GitHub. Those are the same operation with opposite correct answers, which is most of why this was easy to get wrong.

Since a chunk of the file is shell an implementer runs verbatim, I checked there's no other non-portable idiom hiding: `date +%s` (POSIX), `${PING_TS%%.*}` (POSIX parameter expansion), `case` globs, and the `until`/`sleep` loops are all fine. `%-d`/`%-I` in the format string are a glibc/BSD-libc extension rather than POSIX, but they're supported on both targets, so I left them.

s-annam:
**Heads-up: the scope grew after your review** — `522a7f1` (force-push, amended into the same single commit). Flagging it rather than letting it ride, since it's new prose you haven't read and both of your earlier threads are already resolved.

`/pr-ready` now takes a list: `/pr-ready 572,605,606` is **one** ping covering all three, with one checkpoint, one reminder, one deadline, and a per-PR report. The motivation was immediate — three PRs were open at once and running the skill three times would have @-mentioned the same three people three times and held three sequential 90-minute deadlines.

Most of it is a loop, but three points are decisions, and those are the ones worth your skepticism:

1. **Preflight partitions instead of aborting** (Phase 1). A PR that fails preflight drops out and the ready ones still get pinged. The exclusions print at the confirmation *and* in the Phase 6 report — the failure mode I was guarding is that a reviewer who sees two PRs assumes two is all there is.
2. **The idempotency scan matches the set exactly** (Phase 2). Exact match → adopt; no overlap → post; **partial** overlap → stop and name it. Posting would re-ping people about a PR already asked about, and adopting would wait on a ping that never mentioned the PRs just added. Merging two asks into one thread is a third behaviour nobody specified, so I didn't invent it — the cost is one re-run with a narrower list, and the skill says which one.
3. **The ack predicate splits by scope** (ack predicate section). A reaction or thread reply lands on the *ping*, which names every PR, so it's ask-level and can't be attributed to a row; only GitHub activity is per PR. The reminder consumes the union — it fires only when nothing at all has been acked, so someone who just reviewed one PR never gets told they still haven't looked.

Diff is `+194/−38` across `SKILL.md` and `CONTRIBUTING-PROCESS.md` — larger than the "small change" I estimated when I started, mostly because each of those three decisions carries its rationale inline in the house style. Still docs-only; no `src/` change. `typecheck` + `test` re-run clean (2854 passed).

If you'd rather this had been a follow-up PR against `main` instead of a fourth round here, say so and I'll split it back out — the argument for folding it in was that `SKILL.md` doesn't exist on `main` yet, so a separate PR would have had to stack on this branch.

Issue #576, opened 2026-07-25
Deprecate /jd-fit/ — migrate JD paste + JD-driven rewrite into the Find Jobs tab, then remove the second entry


s-annam:
## Summary

The `/jd-fit/` page is a second HTML entry that renders a JD-coverage view the job-search lane already renders, off the same engine. `JdMatch.tsx` is reused verbatim by `JobResultCard.tsx:81`, and `rank.ts:101-103` runs the identical `extractJdTerms` + `computeCoverage` pair per posting that `JdFitApp.tsx:51-53` runs once for a pasted JD. Two shells, one engine.

Collapse to one surface: migrate the two capabilities that exist **only** on `/jd-fit/` into the Find Jobs tab, then delete the page, its entry, its flag, and its handoff.

**This is not a pure deletion.** Two premises that look true from outside are not:

1. **The flag does not gate the route.** `jd-fit-banner` / `VITE_ENABLE_JD_FIT` (`src/lib/flags.ts:62`, read at `src/App.tsx:70`) gates only the "Check fit against a job" cross-sell button on `/`. `jd-fit/index.html` is a real Rollup input (`vite.config.ts:159`), so `/jd-fit/` is built into `dist/` and reachable by direct URL on production today regardless of the flag. Removing it is a real, if small, user-visible change — treat it as one.

2. **`src/lib/jd-match/` is NOT duplicative and must NOT be deleted.** The job-search lane hard-depends on it. Deleting the directory breaks Find Jobs entirely. See "What must survive" below.

## Evidence of duplication

| Concern | `/jd-fit/` | Find Jobs (`/`) |
|---|---|---|
| Term extraction | `JdFitApp.tsx:51` → `extractJdTerms` | `rank.ts:101` → `extractJdTerms` |
| Coverage computation | `JdFitApp.tsx:53` → `computeCoverage` | `rank.ts:102` → `computeCoverage` |
| Coverage renderer | `JdFitApp.tsx:122` → `<JdMatch>` | `JobResultCard.tsx:81` → `<JdMatch>` |
| Résumé shell | `<PageShell>` + `<Result>` | `<PageShell>` + `<Result>` |
| Parse pipeline | `useAnalyzedResume` | `useAnalyzedResume` |

The only structural difference is where the JD text comes from: `/jd-fit/` takes it from a textarea or a URL fetch; Find Jobs takes it from a posting it discovered.

## What must survive

[...]

Issue #568, opened 2026-07-24
[job-search v3] Progressive inline refinement of the query


s-annam:
Part of the job-search v3 epic #566.

## Problem

The Find Jobs query is derived from the résumé and then largely fixed. The user gets one editable title field and skill chips (`FindJobsPanel.tsx`), runs the search, and takes what comes back. There is no moment where the panel asks the questions a recruiter would ask first — *what level? what comp? which of these role families do you actually want? what should I rule out?* — and re-ranks live as the answers come in.

The gap is not the engine. The v3 engine issues add the signals:

- **#562** derives a target **level** and makes it a ranking axis.
- **#563** adds **exclusion** terms (title-only negative chips).
- **#564** extracts **compensation** and supports a soft floor.
- **#545** (done) added **location**.

Each of those exposes a knob. But a knob nobody is prompted to turn is a knob most users never find. The observed quality delta against an agent-driven skills-based search over the same profile came substantially from the agent *interrogating the candidate* — confirming level, comp window, and which adjacent roles to exclude — before ranking. This issue is the in-app equivalent.

## Direction — progressive inline refinement, not a wizard

Refine **in place** in `FindJobsPanel`, live, reusing the existing chip interaction — **no separate wizard surface** (respects the Reuse Gate; a parallel step-through surface would need its own Reuse-analysis and isn't justified here). Results re-rank as controls change; nothing is gated behind a multi-step flow.

Surface, alongside today's title + skill chips:

1. **Role families** — the classified `ROLE_FAMILIES` shown as removable chips, so a fullstack résumé that also matched `data` can drop `data` if the user only wants fullstack.
2. **Target level** — the derived `seniority` (from #562) shown as an editable control; changing it re-runs the #562 rung-distance penalty.
3. **Comp window** — an optional minimum (from #564), applied as the #564 soft floor + badge.
4. **Exclusions** — the seeded + user-added negative chips from #563.
5. **Location** — the existing #545 signal, surfaced consistently with the rest.

All five are **seeded from the résumé-derived query and fully editable/removable** — the same seed-visibly-then-let-the-user-correct pattern #563 mandates for exclusions. A refinement the user can't see and undo is worse than none.

## Dependencies

Depends on **#562, #563, #564** — this issue wires their knobs into one coherent refinement surface. It should land **after** those, or against whatever subset has landed (each control degrades gracefully to absent when its backing signal isn't there yet).

## Acceptance

[...]

Issue #21, opened 2026-06-09
Adopt strict frontend component architecture and reuse guidelines


s-annam:
## Context and Problem Statement
The frontend of `resumelint` requires more rigorous component architecture and reuse guidelines to prevent UI drift and duplicated code as the project scales. The goal is to adopt enterprise-grade frontend best practices (such as those established in Recruidea), specifically covering:
- A strict 3-tier component architecture (Primitives, Shared Composed, Features)
- A token-driven styling system that prevents hardcoded colors
- A strict separation of UI interaction hooks from data-fetching hooks
- A "Reuse Gate" to soft-block creating new workflow surfaces if existing ones can be extended

## Implementation Plan
*Note: This implementation was completed locally prior to issue creation.*

1. **Update `CLAUDE.md`**
   - Add sections detailing the 3-tier Component Architecture, Styling & Tokens, Data & Hooks, and specific antipatterns to avoid.
   - Introduce the "Reuse Gate" concept.

2. **Add a Reuse Reminder Hook**
   - Create `scripts/hooks/reuse_surface_reminder.sh`.
   - Make it check `git diff --cached` for any new files placed in `src/components/`.
   - Emit a warning reminding developers/agents to perform a capability search before adding parallel UI surfaces.
   - Make the script executable.

3. **Wire Hook into Claude Settings**
   - Update `.claude/settings.json` under `PreToolUse` for `Bash` to invoke `reuse_surface_reminder.sh`.

## Acceptance Criteria
- [x] `CLAUDE.md` accurately documents the new frontend component guidelines, anti-patterns, and token system.
- [x] The `scripts/hooks/reuse_surface_reminder.sh` script detects new files added to `src/components/` and prints a visible warning.
- [x] Claude Code correctly triggers this hook via `.claude/settings.json`.

### Reuse analysis
**Capability:** N/A (Infrastructure and documentation updates only)
**Existing surfaces found:** N/A
**Decision:** N/A

s-annam:
## Implementation Complete — `a30464b`

### Acceptance Criteria

- [x] **`CLAUDE.md` accurately documents the new frontend component guidelines, anti-patterns, and token system.** — Implemented sections on Component Architecture, Styling & Tokens, Data & Hooks, and specific antipatterns.
[...]

PR #577, opened 2026-07-25
feat(jobs): give job search its own page and real paging

s-annam:
## Summary

Two problems with Find Jobs, both reported from using it:

1. **Only 20 results, with the wrong instruction.** `JobSearchResults` sliced the
   ranked set at a `RENDER_CAP = 20` and the footer asked the user to narrow the
   query — but `searchJobs` had already fetched and ranked everything, so
   narrowing can never surface match #21. The list is now paged at 20 per page
   over the full set (numbered pages + Prev/Next), so every match is reachable
   and nothing new is fetched.
2. **Results were at the bottom of `/`.** Job search now owns its own entry,
   `/jobs/`, laid out as **full-width bands in the order of the work**: company
   targets pinned at the top, the rest of the query a full-width form that folds
   to a one-line summary on the Search click, then the ranked postings owning the
   full width. `/`'s Find Jobs tab is a launcher: it previews the derived terms,
   stashes the edited parse and navigates.

Filed as a PR rather than an issue on purpose — the Cloudflare preview makes it
easier to keep iterating on the lane from here.

**Privacy unchanged.** The handoff carries only `HeuristicParsedResume` — no PDF
bytes, no raw text — in `sessionStorage`, so it dies with the tab. Egress is
still just the keyword string from `providers/keywords.ts`, still only on an
explicit Search click.

Note for #576 (deprecate `/jd-fit/`): this adds a third entry rather than
removing one. It doesn't conflict — the consolidation target there is the JD
paste + JD-driven rewrite, and `/jobs/` is now the obvious surface to fold them
into.

## Notable decisions

- **No sidebar, and no sticky bar.** A 20rem query rail would have charged every
  posting card for a form that only matters while it's being filled in. Bands
  instead: companies, then the foldable form, then results — all full width.
  Nothing is sticky, because with the company chips permanently on top a pinned
  band is ~3 chip rows tall and costs more viewport on every scroll than the
  Search button it keeps in reach is worth.
- **Company targets are the one control that outlives the fold** — they're the
  axis you retune *while reading* postings, and the toggle is now live against an
  existing result set, so folding them away would bury the panel's tightest
  feedback loop.
- **Folds on the Search click, not on the first result** — the transition is tied
  to the user's own action instead of happening under their cursor a second later,
  and the loading skeleton already occupies the width the results will land in.
- Query fields lay out across the page (titles+role / skills / location+exclude+pay,
  with the 12-rung level control spanning the full width) rather than stacking in
  one ~470px column on a 1080px page.
- Copy that used to say "the query **above**" now says "open Edit search" — with a
  foldable form, a direction word points at a collapsed bar half the time.
- `JobQuerySummary` omits filters that aren't filtering (unset comp floor, empty
  exclude list) so the folded line can't imply a filter is doing work. Location is
  the one exception — it always shows, `anywhere` when unset, because a
  location-blind search otherwise reads as a bug.
- `Pagination` is a new **shared** design-system component (windowed page strip,
  first/last always reachable, `aria-current="page"`), not hand-rolled buttons in
  the feature. The disclosure reuses the house `<Button aria-expanded>` idiom from
  `WeakMatchesSection` rather than adding a Disclosure primitive.
- The page index resets on a new result set (fresh search or #568's live re-rank)
  and is **clamped** — the reset effect runs a render late, so an unclamped slice
  would flash an empty results region when the set shrinks.
- The handoff is deliberately **not** one-shot, unlike `jd-fit-handoff.ts`:
  `/jobs/` has no DropZone, so consuming the key would make a reload a dead end.
- `src/jobs/main.tsx` configures **no** pdfjs worker — that surface never parses
  a PDF, so the entry chunk stays small.

## Company targets now reach existing results (asymmetrically)

Toggling a company used to require a whole new search. It doesn't now — but the
two directions are not the same operation, so they don't behave the same:

| Action | Behaviour | Cost |
|---|---|---|
| **Remove** a company | Results update live | Zero. `dropCompanyPostings` filters the snapshot by the `{ats}:{slug}:` id prefix — board postings are self-attributing, no new `JobPosting` field |
| **Add** a company | `PendingCompaniesNotice`: "These results don't include Ramp — [Add to results]" | One board on click. `searchCompanyBoards` fetches only the added boards (usually an IndexedDB cache hit via `makeBoardProvider`), merges under `search.ts`'s own dedup key, re-ranks |

Adding can't be served locally at all — the snapshot has none of that board's
postings — and auto-fetching on a checkbox click would contradict *"only your
search keywords are sent, and only when you click Search."* So the addition is
surfaced and the fetch stays behind an explicit click. The keyless feeds are
never refetched either way.

A board that fails still leaves the pending list, so a permanently-failing board
can't re-prompt forever; the failure shows up in the existing degraded-providers
notice. An incremental fetch that throws leaves the current results untouched.

## Test plan

- [x] `npm run test` green — 3033 passed / 9 skipped (216 files)
- [x] `npm run verify` green (fallow findings are report-only: JSX-dense
      `JobQueryEditor` + `FindJobsPanel`, 0.1% duplication)
- [x] `npm run build` emits `dist/jobs/index.html` alongside `dist/jd-fit/`
- [x] **Verified in a real browser** against the production build (Playwright over
      the built `dist/`): 268 matches, `Showing 1–20 of 268` + `‹ Prev 1 2 3 4 5 6 … 14 Next ›`,
      form folds on Search to `Founder & CEO / Sr. Engineering Manager · Santa Clara, CA · Executive · 11 skills · 14 companies`,
      deselecting Databricks dropped 268 → 260 with **no** refetch, re-selecting
      showed the pending notice, and Add to results merged it back
- [x] Cloudflare preview: same flow starting from `/` (the sessionStorage handoff
      leg), which the local run stubbed by writing the key directly
- No fixture binaries added or changed, so the PII preflight is a no-op here

New tests assert reachability, not just counts — page 2 checks that matches 21–25
actually render, and the launcher test asserts the parse is stashed *before* the
navigation.

## Provenance

Code implementation via: Claude Opus 5 (1M context), high effort
Verification: `npm run verify` — green locally, and in CI on this PR

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -45,11 +45,11 @@ There is **no BYOK LLM provider in the tree** — `#320` is future; App.tsx / Ca
 
 ### Product lanes and entry points
 
-The build ships exactly two HTML entries (`vite.config.ts` `rollupOptions.input`):
+The build ships exactly three HTML entries (`vite.config.ts` `rollupOptions.input`):
 
 - **`/` (index.html)** — the parser-audit lane: drop → parse cascade → score → editable reconstructed resume (`ReconstructedResume` + `EditableField`) → Download PDF (`src/lib/pdf/render-ats-pdf.ts`). On-device WebLLM insights (parse disagreement, resume-quality critique, rewrite) layer on top when WebGPU is available (`src/lib/webllm/`).
 - **`/jd-fit/` (jd-fit/index.html)** — the JD-match lane: paste a JD, get requirement/evidence coverage (`src/lib/jd-match/`, semantic via WebLLM with keyword fallback) and JD-driven section rewrites. Resume state hands off from `/` via `src/lib/jd-fit-handoff.ts`.
-- The **job-search lane** (`src/lib/job-search/`: query builder → provider search → rank by resume fit → deep links) rides inside the main page (`FindJobsPanel`), not a third entry.
+- **`/jobs/` (jobs/index.html)** — the job-search lane (`src/lib/job-search/`: query builder → provider search → rank by resume fit → deep links). `FindJobsPanel` is the whole page: a full-width query form that folds into a sticky one-line summary (`JobQuerySummary`) the moment Search is clicked, with the paged ranked results owning the full width below it. The parsed résumé arrives from `/` via `src/lib/jobs-handoff.ts` (sessionStorage, read but NOT consumed, so a reload survives); `/` keeps only `FindJobsLauncher`, which stashes the parse and navigates. This surface has no DropZone — parsing is `/`'s job.
 
 `jd-spike.html` and `eval-rewrite.html` are dev-only harnesses, deliberately excluded from the production build.
 
--- a/README.md
+++ b/README.md
@@ -346,9 +346,9 @@ is the custom domain **<https://offlinecv.org>**; the project-Pages URL
 <https://offlinecv.github.io/OfflineCV/> remains as a fallback (built
 with `VITE_BASE_PATH=/OfflineCV/`).
 
-The build emits two root pages — `/` (the parser audit) and `/jd-fit`
-(JD-match + JD-driven rewrite) — as a multi-entry Vite build, so both ship
-in the same self-contained `dist/`.
+The build emits three root pages — `/` (the parser audit), `/jd-fit`
+(JD-match + JD-driven rewrite), and `/jobs` (the job-search workbench) — as a
+multi-entry Vite build, so all of them ship in the same self-contained `dist/`.
 
 To bake telemetry into a deployed build, set `VITE_POSTHOG_KEY` (and
 optionally `VITE_POSTHOG_HOST`) in the build environment — see the
--- a/jobs/index.html
+++ b/jobs/index.html
@@ -1,0 +1,17 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>OfflineCV — Find Jobs</title>
+    <meta
+      name="description"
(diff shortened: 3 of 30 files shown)


s-annam:
Merging via admin bypass rather than waiting on review.

Rationale: this is a self-contained lane addition — a new `/jobs/` entry, real paging, and live
company targets — behind its own HTML entry point. It does not change the parser, the score, or the
export path, and `/` only gains `FindJobsLauncher`. Holding it for a reviewer who is next available
Monday blocks the follow-up term-quality work (#578–#588), all of which builds on the surfaces this
PR introduces. I own the consequences of the bypass.

Gates that did run: `npm run verify` clean (typecheck, lint, coverage, build), Cloudflare preview
deployed successfully at `b4ba17a`, and the branch is a single commit as the merge queue requires.

One `fallow` finding reviewed and accepted as a Nit rather than fixed here:

```
src/components/features/JobQueryEditor.tsx
  :45 JobQueryEditor  6 cyclomatic  5 cognitive  135 lines  42.0 CRAP
```

`6 + 6²·(1−0)³ = 42` is the arithmetic signature of zero coverage, not of complex code — the six
branches are all render guards. The component was extracted out of `FindJobsPanel` verbatim
(`setQuery` → `onChange`; `isDegenerate` moved from a derived local to a prop the panel computes
with the identical expression), and `FindJobsPanel` was equally untested on `main`, so no
previously-covered code lost coverage. `fallow` is report-only inside `verify` by design.

Coverage the PR does add: `useJobSearch`, `FindJobsLauncher`, `JobQuerySummary`, `Pagination`,
`jobs-handoff`, `raw-postings`, plus additions to `JobSearchResults`.

The gap is tracked in #589 — a props-in/callback-out test for the query write path, including a
frozen-fixture assertion that no handler mutates the query in place.

Issue #552, opened 2026-07-24
feat(export): add "Download as Markdown (cv.md)" export from the parser-audit lane


s-annam:
## Summary

Give the parser-audit lane (`/`) a **markdown round-trip**: export the cleaned/edited résumé as a `cv.md` file, **and** accept a dropped `.md` file as an input source that re-parses back into the résumé model.

- **Export.** Today the lane exports exactly one artifact — the ATS-safe **Download PDF** (`useDownloadPdf` → `render-ats-pdf.ts`), which additionally embeds a lossless **JSON Resume** copy as a `resume.json` attachment inside the PDF (#334). There is no plain-text/markdown copy. Add a **"Download as Markdown"** action emitting a single `cv.md`.
- **Import.** Today the drop surface accepts only `.pdf` and `.docx`. Add `.md` as a third source. This is the *opposite side* of the export: a user can download `cv.md`, edit it in any text editor, and drop it back to re-parse, score, and re-export.

The concrete driver: [`santifer/career-ops`](https://github.com/santifer/career-ops) requires a freeform markdown `cv.md` at its project root, which it reads to evaluate job fit and generate keyword-injected ATS PDFs. Exporting `cv.md` lets a user feed offlinecv's clean output straight into career-ops; importing `.md` closes the loop so a hand-edited or career-ops-shaped `cv.md` flows back in. career-ops is a **downstream consumer, not a dependency** — this is a general-purpose markdown round-trip; career-ops is the motivating use case and the target for the emitted section conventions.

## Background — the seams already exist

**Export path** is already structured for a sibling artifact:

- **`src/lib/pdf/ats-resume-model.ts`** — `buildAtsResumeModel(result, score, edit)` builds the flat `AtsResumeModel` from the cascade result + edit overrides. Each `AtsEntry` welds *export-semantic* fields (`AtsEntryFields`: `organization`, `position`, `studyType`, `area`, `startDate`, `endDate`, `isCurrent`, `url`, `courses`, `skills`, `skillCategory`, `title`) onto *render layout* hints.
- **`src/lib/pdf/ats-export-projection.ts`** — `projectAtsExport(model): AtsExportProjection` = `{ contact: AtsContact, entries: readonly AtsExportEntry[] }` (each `AtsExportEntry = { kind, fields?, bullets }`), the layout-free export-semantic view (#442). **This is the correct source for a new export.**
[...]

Issue #237, opened 2026-06-28
Parser: bare domain mid-text mis-promoted to URL / website_url (achievements + contact)


s-annam:
## Summary

A bare domain token appearing **inside descriptive text** (not as a standalone contact line) is wrongly extracted as a URL. Two surfaces, one root cause:

1. **Achievements** (`achievements.ts` → `liftHeaderLabel` URL detection): a domain mid-sentence is yanked out of the title and stored as the achievement `url`, leaving a hole in the text.
2. **Contact** (`contact.ts` URL capture): the same in-body domain is promoted to a top-level contact field (`website_url`), even though it is not a contact link.

## Observed

Source line:
```
Exit · Founded and sold return2india.com to Satyam Infoway (NASDAQ: SIFY). 200K monthly visits. [2000]
```

Parsed:
- achievement title → `Exit · Founded and sold  to Satyam Infoway …`  ← `return2india.com` removed, double-space hole left
- achievement `url` → `return2india.com`
- top-level `website_url` → `https://return2india.com`  ← false contact field

The domain is part of a sentence ("sold return2india.com to …"), not a link the candidate listed.

## Root cause

URL/domain detection fires on any domain-shaped token regardless of position. A domain embedded in prose is descriptive content, not a hyperlink or a contact URL. Both `liftHeaderLabel` (achievements) and the contact extractor over-capture.

## Fix direction

Only treat a domain as a URL when it is positionally a link — a standalone token / line, or backed by a real link annotation — not when it sits mid-sentence surrounded by words. Leave in-prose domains in the text. Apply consistently to both the achievements label-lift and the contact website capture.

## Acceptance

- `… sold return2india.com to Satyam …` keeps the domain **in the title text**; no `url` lifted; no `website_url` set from it.
- A genuine standalone portfolio/website line still populates `website_url`.
- Unit tests for both surfaces (achievements label-lift + contact capture).

## Related

Surfaced alongside #224. Single root cause across `achievements.ts` and `contact.ts`.

PR #612, opened 2026-07-28
fix(parser): flatten every markdown link shape, and stop body URLs hijacking the contact card

### LINKED ISSUE #610: Parser: dropping a .md leaks inline link syntax — body URL hijacks contact website_url, entry title keeps an orphaned [label]() residue
## Summary

Dropping a `.md` résumé whose body contains a markdown **inline link** — `[label](https://example.com/thing)` — corrupts the parse in two visible ways, both of which ship into the exported "Download PDF":

1. The link's URL is claimed as the candidate's **contact `website_url`** and renders on the contact line next to LinkedIn, even though the link lives in Achievements.
2. The entry's title keeps an **orphaned `[label]()`** — the URL is spliced out of the text but the surrounding markdown syntax is not, leaving empty parens in the résumé.

Both trace to a single root cause: the markdown cascade never flattens inline link syntax, so `[label](url)` reaches the field extractors intact.

This is the `.md` import path (`#552`), not the PDF path. A PDF carrying the same link as a real link annotation is unaffected.

## Reproduction

Drop a `.md` whose Achievements section contains an inline link, then click **Download PDF**:

```markdown
## Achievements

- **Patent** · Issued Patent [US1234567B1](https://patents.google.com/patent/US1234567B1); expression-language bulk editor for product catalogs. [2019]
```

**Expected** — contact line shows only real contact links; the achievement reads `Issued Patent US1234567B1` (or keeps the label with a working hyperlink).

**Actual** — verified by running `runCascadeFromMarkdown` on `main`:

```
website_url  = https://patents.google.com/patent/US1234567B1
profiles     = [{"url":"https://linkedin.com/in/janedoe","network":"LinkedIn","kind":"social"},
                {"url":"https://patents.google.com/patent/US1234567B1","network":"patents.google.com","kind":"other"}]

heuristic_achievements[0] = {
  "type": "Patent",
  "title": "Issued Patent [US1234567B1](); expression-language bulk editor for product catalogs.",
  "year": "2019",
  "url": "https://patents.google.com/patent/US1234567B1"
}
```

The rendered PDF's contact line becomes `jane@example.com • (312) 555-0123 • Chicago, IL • linkedin.com/in/janedoe • patents.google.com/patent/US1234567B1`, and the Achievements entry renders the literal `[US1234567B1]()`.

Note that `achievements[0].url` is populated **correctly** — the link target is already captured as structured data. The residue and the contact-line leak are both pure loss.

### Reproduction 2 — link in an Experience bullet

The case above demonstrates the residue narrowly, through `liftHeaderLabel` acting on an achievement *title*. A real-world failure hit the same root cause from a different surface: a full, un-flattened `[label](https://url)` rendered visibly inside an Experience/description bullet, since `markdown-lines.ts` has no inline-link rule and Experience bullets never pass through `liftHeaderLabel`.

```markdown
## Experience

**Staff Engineer**, Example Corp — 2020–2024
- Led the [catalog migration](https://example.com/eng-blog/catalog-migration) that cut checkout latency 40%.
```

**Expected** — the bullet reads `Led the catalog migration https://example.com/eng-blog/catalog-migration that cut checkout latency 40%` (per Step 1's `label url` flattening below), never the raw markdown syntax.

**Actual** — the bullet renders literally as `Led the [catalog migration](https://example.com/eng-blog/catalog-migration) that cut checkout latency 40%` in the exported PDF, and the URL is separately claimed as `website_url` on the contact line via the same doc-wide fallback described below.

This confirms the fix has to apply uniformly to every markdown line the cascade touches — not just achievement/project titles — for the reported bug class to actually be closed.

### The ingest asymmetry

`src/lib/ingest/markdown.ts:55-57` returns two representations of the dropped file:

```ts
export function parseMarkdownFile(text: string): MarkdownParseResult {
  return { rawText: mdToPlainText(text), markdown: text };
}
```

`rawText` **is** flattened — `mdToPlainText` at `src/lib/ingest/markdown.ts:40` rewrites `[text](url)` → `text url`. But `rawText` only feeds the scorer's plain-text scans. The **cascade** consumes `markdown`, which is the file's text *verbatim*.

`src/lib/heuristics/markdown-lines.ts` strips markdown **images** (`ANY_IMAGE_RE` and `DATA_URI_IMAGE_RE`, lines 75-87, applied by `stripImages`) and undoes turndown's backslash escapes — but has no inline-**link** rule. So `[label](url)` survives into every `PdfLine.text` the extractors see.

### Downstream defect 1 — contact `website_url` hijack

`extractContact` runs a **document-wide** text fallback (`src/lib/heuristics/extract/contact.ts:382-383`):

```ts
const fullText = allLines.map((l) => l.text).join(" ");
const fallback = scan(allLines, fullText);
```

and `website_url` is the explicit **catch-all** — `src/lib/heuristics/extract/contact.ts:507-528` documents it as "any remaining URL not already claimed by linkedin/github/portfolio". Document-wide reach combined with catch-all semantics means *the first scheme-bearing URL anywhere in the résumé* becomes the contact website.

The `#237` guard that should stop this does not apply. `src/lib/heuristics/extract/contact.ts:298-300`:

```
// Reject a bare domain mid-sentence in prose (e.g. "sold return2india.com
// to Satyam"). A URL with an explicit scheme or www. is always standalone.
return isStandaloneUrl(u, withoutEmails);
```

`#237` fixed the **scheme-less** case only; a `https://`-prefixed URL is unconditionally "standalone" and passes unchallenged regardless of which section it sits in.

Nothing dedups a URL already lifted into an achievement or project entry against the contact card. `#134` built exactly that ownership model but scoped it to LinkedIn/GitHub — `src/lib/heuristics/extract/contact.ts:534-536` states "portfolio/website stay header-banded", which is true for the *annotation* tier (`inProfileSection`) but **not** for the doc-wide *text* tier.

### Downstream defect 2 — orphaned `[label]()`

`liftHeaderLabel` (`src/lib/heuristics/extract/projects.ts:59-109`, shared by projects and achievements via the import at `src/lib/heuristics/extract/achievements.ts:18`) locates the URL with `URL_RE` and splices out **only the matched URL substring** (lines 90-105). It has no notion of markdown link syntax, so the enclosing `[` … `](` … `)` remains, collapsing to `[label]()`.

## Proposed fix

The fix belongs in `markdown-lines.ts`, alongside the existing image stripping — one point of change resolves both defects, rather than separately patching `contact.ts` and `projects.ts`. Both steps below are required to close this issue: Step 1 alone removes the `[label]()` residue, but a flattened `label url` still leaves a scheme-bearing URL sitting in body text, which the doc-wide `website_url` catch-all will still steal — so the contact-line hijack (Downstream defect 1, and Reproduction 2) survives Step 1 on its own.

### Step 1 — flatten inline links in `src/lib/heuristics/markdown-lines.ts`

Add a rule beside `ANY_IMAGE_RE` (declared after it, so the image rule — which is `!`-prefixed — still matches first):

```ts
// Inline markdown links: `[label](url)`. Unlike images, both halves carry
// signal — the label is the visible résumé text and the target is a real
// link — so flatten to `label url` rather than dropping the whole ref.
const INLINE_LINK_RE = /\[([^\]]*)\]\(([^)\s]+)(?:\s+"[^"]*")?\)/g;
```

Apply it in the same pass as `stripImages` (or a sibling `flattenLinks`), producing `label url`. That mirrors `mdToPlainText`'s existing `$1 $2` behavior at `src/lib/ingest/markdown.ts:40`, so the two representations of a dropped `.md` finally agree — which is the actual invariant being restored.

**Decide and document** whether to emit `label url` (matches `mdToPlainText`; keeps `liftHeaderLabel` able to find and lift the target into `achievement.url`) or bare `label` (cleaner text, but the target is lost for any consumer that isn't already capturing it). Recommend `label url`: `liftHeaderLabel` then lifts the URL exactly as it does for a plain-text résumé, and `achievement.url` stays populated.

Note the optional-title branch (`[a](url "Title")`) and that a URL containing a `)` is out of scope — CommonMark requires it be angle-bracketed or escaped, and neither shape appears in a résumé.

**Link variants — scope for this PR:** bare autolinks (`<https://…>`) are **in scope** and should flatten the same way as inline links (there's no separate label, so the visible text becomes the URL itself, consistent with how a bare URL already reads today). Reference-style links (`[label][ref]`) are **explicitly out of scope** for this PR — leave them unflattened and file a follow-up issue to cover them; do not silently half-handle them here.

### Step 2 — stop the doc-wide `website_url` catch-all from claiming body URLs

Two options; pick one and record the reasoning in the module docblock:

- **(a) Band `website_url`/`portfolio_url` to the profile section on the text tier**, matching what the annotation tier already does via `inProfileSection`. This makes the `#134` docblock claim ("portfolio/website stay header-banded") true on both tiers. Risk: a résumé that genuinely prints its personal site in a footer loses it — check the corpus for that shape before committing.
- **(b) Subtract URLs already claimed by a body entry.** After the body extractors run, drop any `website_url`/`portfolio_url` whose `urlSlug` matches an `achievement.url` or `project.url`. Narrower, but requires re-ordering extraction or a post-pass, and the contact extractor currently runs first.

Option (a) is the smaller change and is consistent with the existing annotation-tier design; (b) is more precise. Either is acceptable if reasoned in the docblock — this choice is the implementer's call, not prescribed by this issue.

## Acceptance criteria

- [ ] A `.md` whose Achievements section contains `[label](https://…)` parses with `website_url === undefined` (given no other website in the header) — the body URL never reaches the contact card.
- [ ] The same parse yields an achievement `title` with **no** `[` / `]` / `()` residue.
- [ ] The same parse still yields `achievement.url === "https://…"` — the link target is preserved as structured data, not discarded.
- [ ] `profiles[]` contains only genuine contact links; no `{"network":"patents.google.com","kind":"other"}`-style entry sourced from a body section.
- [ ] Inline links are flattened wherever they appear, not just in Achievements — a Projects entry, an Experience bullet, and the Summary each keep their label text and lose the syntax.
- [ ] A full `[label](https://url)` link inside an Experience/description bullet (Reproduction 2) does not survive into the rendered Download PDF as raw markdown — it renders flattened (`label url`), not as literal bracket-paren syntax.
- [ ] Markdown **images** still strip entirely (existing `stripImages` behavior unchanged) — an `alt` must not start flattening to `alt url`.
- [ ] Bare autolinks (`<https://…>`) are flattened by this fix, using the same acceptance criteria as inline links above.
- [ ] Reference-style links (`[label][ref]`) are explicitly documented as out of scope for this PR, with a follow-up issue filed to cover them.
- [ ] **Step 2 (option-agnostic):** a URL that appears only in a body section (Achievements, Projects, or Experience) does not surface as the contact `website_url` or `portfolio_url` — this must hold regardless of which Step 2 implementation (band-to-profile-section vs. subtract-claimed-URLs) is chosen.
- [ ] The rendered Download PDF for the fixture shows a clean contact line and a clean achievement title.
- [ ] **Regression:** `npm run test` green — specifically `corpus.test.ts`, `corpus-roundtrip.test.ts`, `markdown-lines.test.ts`, `cascade-markdown.test.ts`, `extract-fields.test.ts`, and `src/lib/ingest/markdown.test.ts`. No corpus snapshot re-bake without an explicit per-fixture justification in the PR body.
- [ ] **Regression:** `#237`'s bare-domain-in-prose guard still holds — "sold `acme.com` to a buyer" in a Summary must not become `website_url`.

## Test fixture

Add a `.md` fixture (not a PDF — this is the markdown ingest path) under the markdown-ingest test surface, plus a unit test in `markdown-lines.test.ts` for the flattening rule itself. Cover both the Achievements-title shape (Reproduction) and the Experience-bullet shape (Reproduction 2) — they exercise different code paths (`liftHeaderLabel` vs. plain line text) and both need a guarding test.

**PII — hard rule.** Use a synthetic persona: fake name, `@example.com` email, and a real-area-code + `555` exchange + `0100`–`0199` subscriber phone, e.g. `(312) 555-0123`. Do **not** copy the patent number, name, or any other identifier from the report that produced this issue — the repo is public. See `CLAUDE.md` → Hard rules. `npm run check:fixtures` only scans PDFs under `tests/fixtures/pdfs/`, so a `.md` fixture is **not** covered by that gate — the judgement is manual here.

**Prove the fixture guards.** Land the fixture, confirm the new test **fails** on current `main`, then land the fix. A fixture baked after the fix passes vacuously.

## Out of scope / observed but not filed

The same parse produced `"year": "2006"` for an achievement whose source reads `[2006–2024]`. This is **deliberate**, not a defect — `src/lib/heuristics/extract/achievements.ts:169-171` documents it: "Reduce any date range the header carried to a single lead year", and `HeuristicAchievement` has a single `year` slot. Changing it means widening the model. Recording it here so the next reader doesn't re-discover it as a bug; file separately if the collapsed range is judged misleading in the export.

Reference-style links (`[label][ref]`) — see Step 1 above; deliberately deferred, not covered by this issue's acceptance criteria.

## Prior art

All closed; none covers this path.

| Issue | Why it doesn't cover this |
|---|---|
| `#237` | Bare domain mid-text mis-promoted to `website_url` — fixed the **scheme-less** case only; this report is scheme-bearing and bypasses the guard by design. |
| `#134` | Retired `stripPromotedUrls` in favor of an ownership model — scoped to LinkedIn/GitHub identity links; its "portfolio/website stay header-banded" claim holds on the annotation tier but not the doc-wide text tier. |
| `#249` | Substring/duplicate URL on a single header line — a different failure in `liftHeaderLabel`, not the markdown-syntax residue. |
| `#552` | Introduced the `.md` import path, and with it the `rawText`-flattened / `markdown`-verbatim asymmetry this issue reports. |

### LINKED ISSUE #611: Parser: reference-style links ([label][ref]) are not flattened; link-definition lines leak as content
## Summary

Reference-style markdown links — `[label][ref]` with a matching `[ref]: https://…` link definition elsewhere in the document — are not flattened by the markdown cascade. They were **explicitly deferred** out of scope when #610 was clarified; this issue tracks the deferred half.

Split from #610, which covers inline links (`[label](url)`) and bare autolinks (`<https://…>`).

## Why it wasn't folded into #610

#610's fix adds an inline-link rule to `src/lib/heuristics/markdown-lines.ts`, beside the existing image stripping. Inline links and autolinks are self-contained — everything needed to flatten them sits on the one line being processed.

Reference-style links are not. Flattening `[label][ref]` requires resolving `ref` against a link-definition table that may be declared anywhere in the document, typically at the bottom. `markdown-lines.ts` processes lines without that document-wide index, so handling this shape means either building the definition table in a pre-pass or accepting that the target is unresolvable and dropping to bare `label`.

That's a structurally different change from #610's one-line regex, hence the split.

## Expected failure shape

Untested — this issue is filed from the #610 clarification, not from an observed report. The presumed behavior mirrors #610's defects:

```markdown
## Achievements

- **Patent** · Issued Patent [US1234567B1][patent]; expression-language bulk editor.

[patent]: https://patents.google.com/patent/US1234567B1
```

Likely outcomes to verify before implementing:

1. The `[label][ref]` syntax survives into `PdfLine.text` as raw markdown, same as #610's inline-link case.
2. The trailing `[ref]: https://…` definition line is itself parsed as body content — possibly becoming a stray résumé line, and possibly feeding the doc-wide `website_url` catch-all described in #610's Downstream defect 1.

Outcome 2 may be the more damaging half, and has no equivalent in #610.

## Open questions

- **Resolve or drop?** Build a link-definition pre-pass and emit `label url` (consistent with #610's Step 1 and `mdToPlainText`), or emit bare `label` and accept target loss. The former matches #610; the latter is far cheaper.
- **What happens to the definition lines themselves?** They should almost certainly be stripped from the line stream regardless of which option above is chosen — a `[ref]: https://…` line is document metadata, never résumé content.
- **Is this shape worth supporting at all?** Reference-style links are rare in résumés. Confirm against the corpus before investing in a pre-pass; documenting the limitation may be the correct end state.

## Acceptance criteria

- [ ] Behavior on current `main` is characterized with a test before any fix — confirm both presumed outcomes above, or correct them.
- [ ] Link-definition lines (`[ref]: https://…`) do not appear as résumé content in the parsed output or the rendered Download PDF.
- [ ] A URL appearing only in a link definition does not surface as the contact `website_url` or `portfolio_url` (same invariant as #610's Step 2 criterion).
- [ ] `[label][ref]` leaves no `[` / `]` residue in the rendered output.
- [ ] Whichever resolution is chosen (resolve to `label url` vs. drop to `label`), the reasoning is recorded in the `markdown-lines.ts` module docblock, consistent with #610's Step 1 and Step 2 conventions.
- [ ] `npm run test` and `npm run verify` pass; no corpus snapshot re-bake without per-fixture justification.

## Test fixture

**PII — hard rule.** Synthetic persona only: fake name, `@example.com` email, `555` exchange phone (e.g. `(312) 555-0123`). Do not copy identifiers from any real résumé. `npm run check:fixtures` only scans PDFs under `tests/fixtures/pdfs/`, so a `.md` fixture is not covered by that gate — the judgement is manual.

## Related

| Issue | Relationship |
|---|---|
| #610 | Parent report. Covers inline links and bare autolinks; explicitly defers this shape. |
| #552 | Introduced the `.md` import path and the `rawText`-flattened / `markdown`-verbatim asymmetry both issues stem from. |

s-annam:
## Summary

Dropping a `.md` left markdown link syntax intact all the way to the field extractors. Two visible defects followed, both shipping into the exported Download PDF: a body URL was claimed as the candidate's contact `website_url`, and the entry title kept an orphaned `[label]()`. Root cause was an asymmetry — `rawText` was flattened by `mdToPlainText`, but the cascade consumed the verbatim `markdown`. Both readings now share one set of link rules.

- **#610** — inline links, CommonMark titles and bare autolinks flatten to `label url`, preserving the target so `liftHeaderLabel` still populates `achievement.url` / `project.url`. Step 2 chose **option (a)**, banding `website_url`/`portfolio_url` to the profile section on the text tier.
- **#611** — reference-style links resolve against the document's definition table, and the `[ref]: url` definition lines are blanked. The table lives in a shared `src/lib/markdown-link-refs.ts` consumed by both readings, which is what keeps them in agreement.

Closes #610
Closes #611

## Notable decisions

- **Step 2 option (a) over (b).** Option (b) — subtract URLs already claimed by a body entry — cannot satisfy #610's own acceptance criterion, which names *Experience* alongside Achievements and Projects: an Experience bullet has no `url` slot, so there is nothing to subtract.
- **Accepted cost, stated plainly.** A personal site printed in a page **footer** is now lost on the text tier. The annotation tier does not rescue it — `pdf-extract.ts` flips pdfjs' y to top-origin, so `inProfileSection`'s `ann.yTop <= profileLineMaxY + 12` band is the *header*, the opposite end of the page. Recorded verbatim in the `contact.ts` docblock. `linkedin_url`/`github_url` are unaffected.
- **Undefined references stay literal.** An image or link reference with no matching definition is left untouched, which is CommonMark's own reading and what protects the `[2019]` year marker the achievements extractor depends on.

## Corpus snapshot

Exactly one moves: `two-column-achievements-sidebar` drops `website_url` from `fieldsPopulated`. The value was `LaunchPadCart.com`, lifted from the STRENGTHS sidebar sentence "Successfully launched and sold a consumer website LaunchPadCart.com." — a product the candidate sold, not their contact site. 9 corpus fixtures carried `website_url` before, 8 after; no fixture sources the field legitimately from below the profile band. Score is unaffected (`website_url` is not scored).

## Known limitations (not fixed here)

- `mdToPlainText` has no image strip, so an **inline** `alt` still reads `!alt url` in `rawText`, which `EvidencePanel` prints verbatim. **Pre-existing on `main`** (shipped with #552) — not introduced here, and neither markdown fixture contains an image. Follow-up candidate.
- `achievement.url` keeps a trailing `;` (`URL_RE` treats sentence punctuation as part of the URL), leaving a double space in the title. Pre-existing and path-agnostic, but newly *visible* in the export — `main` showed `[XX0000000A0]();`.
- `#611`'s "no `[`/`]` residue" criterion holds for **defined** references only; an undefined one deliberately stays literal per CommonMark. Disclosed rather than checked off silently.
- No code-span or fenced-block awareness: a link inside backticks flattens, and a definition inside a fence is blanked. Documented in the module's stated CommonMark subset.
- Link scanning is **quadratic**, not exponential — an unbounded label class restarts at every `[`. Measured 3.1 s on 80 000 `[`. `main` already carries the identical shape in `ANY_IMAGE_RE` and `mdToPlainText`; asymptotics unchanged, so no input cap was added.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — `Test Files 238 passed | 8 skipped`, `Tests 3391 passed | 10 skipped`
- [x] `fallow audit --base origin/main` — report-only findings only: 7 clone groups in *test* files, and 2 complexity findings on `normalizeSplitLetterHeaders` / `classifyMarkdownLine`, both verified unchanged from `main` (line-shifted, not introduced)
- [x] Fixture personas verified synthetic — no real PII. `npm run check:fixtures` green on all 54 PDFs; the two new `.md` fixtures fall outside that gate and were judged manually (Riley Nakamura / Avery Okonkwo, `@example.com`, `(312) 555-0123` and `(312) 555-0155` — real area code, `555` exchange, subscriber in `0100`–`0199`)
- [x] Every new test proven to fail before its fix — no vacuously-baked fixture

## Adversarial review

Two bounded rounds ran against the accumulated diff before this PR opened.

**Round 1 — 4 blocking findings, all fixed:**

1. `ingest/markdown.ts` — the "two readings agree" invariant, asserted in three docblocks, was **false**: `mdToPlainText` never received #610's autolink rule and its inline regex swallowed the CommonMark title into the URL. The existing repro tests could not catch it — they scan only the parsed canonical fields, never `rawText`. Fixed by single-sourcing the autolink rules and adding `markdown-readings-agree.test.ts`, which compares the two readings line-for-line.
2. `markdown-link-refs.ts` — reference-style **images** flattened: `![alt][ref]` emitted `!alt url`, injecting an image URL into résumé body text. New behaviour, and a violation of #610's "images must still strip entirely" criterion.
3. `contact.ts` — the docblock **falsely claimed** a footer site stayed reachable via the annotation band. It does not; the band is the header. Comment-only, but it would have landed in `main`, the squash message and this PR body simultaneously. Replaced with the honest statement now quoted above.
4. `contact.test.ts` — #237's bare-domain guard had become **vacuous**: under banding the doc-wide fallback is never consulted, so the negative tests passed regardless of what `isStandaloneUrl` did. The regression criterion was pinned nowhere. Both prose lines moved into the profile section, plus a direct unit test; proven non-vacuous by defeating the guard and observing red.

**Round 2 — `clean: true`**, all four confirmed closed, no new blocking findings. Two of its nits were behaviour **regressions versus `main`** and were fixed rather than deferred: a reference image with bracketed alt text still leaked its URL, and `![anything]` was deleted unconditionally, so ordinary prose (`Wow![great] news`) lost text. The image-reference drop now consults the definition table, and `stripImages` is byte-identical to `main` again.

Surviving nits are the "Known limitations" above.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #610 | Claude Opus 5 (1M context) | ultra |
| Implementation — #611 | Claude Opus 4.5 (1M context) | ultra |
| Adversarial review — round 1 | Claude Opus 5 (1M context) | ultra |
| Review fixes — round 1 | Claude Opus 5 (1M context) | ultra |
| Adversarial review — round 2 | Claude Opus 5 (1M context) | ultra |
| Review fixes — round 2 | Claude Opus 5 (1M context) | ultra |
| Orchestration + PR | Claude Opus 5 (1M context) | ultra |

### DIFF
--- a/src/lib/heuristics/extract/contact.ts
+++ b/src/lib/heuristics/extract/contact.ts
@@ -215,6 +215,57 @@ function findConsumedLines(
 }
 
 /**
+ * Band `portfolio_url` / `website_url` to the profile section on the TEXT tier
+ * as well as the annotation tier (#610) — option (a) of the two the issue put
+ * up. When true, `pickUrl` skips the doc-wide text fallback for those two
+ * fields; `linkedin_url` / `github_url` keep theirs.
+ *
+ * WHY THIS EXISTS. `website_url` is the explicit catch-all — "any remaining URL
+ * not already claimed by linkedin/github/portfolio". Run over a doc-wide join
+ * of every line, catch-all semantics degrade to *the first scheme-bearing URL
+ * anywhere in the résumé*, so a link that lives in an Achievements entry or an
+ * Experience bullet renders on the contact line. #237's guard does not stop it:
+ * that fixed the scheme-LESS case, and a `https://` URL is unconditionally
+ * `isStandaloneUrl`. This also makes the #134 docblock claim below
+ * ("portfolio/website stay header-banded") true on BOTH tiers rather than only
+ * the annotation one, which is what it already asserted.
+ *
+ * WHY NOT OPTION (b) — subtract URLs already claimed by an achievement/project
+ * entry. It cannot satisfy the issue's own acceptance criterion, which names
+ * *Experience* alongside Achievements and Projects: an Experience bullet has no
+ * `url` slot, so its link is never "claimed by a body entry" and there is
+ * nothing to subtract. Generalising (b) to "any URL on a non-profile line" is
+ * option (a) wearing a post-pass, at the cost of re-ordering extraction.
+ *
+ * COST, MEASURED. The stated risk was a résumé that genuinely prints its
+ * personal site in a footer, which would now be lost. Ran the banded build over
+ * all 54 corpus fixtures: exactly ONE snapshot moves —
+ * `tests/fixtures/pdfs/unknown/two-column-achievements-sidebar.pdf` drops
+ * `website_url` from `fieldsPopulated` — and it is a false positive being
+ * removed, not a footer site: the value was `LaunchPadCart.com`, lifted out of
+ * a STRENGTHS-sidebar sentence ("Successfully launched and sold a consumer
+ * website LaunchPadCart.com."). That is a product the candidate sold, not their
+ * contact website, and it is #237's own defect class reaching the card by the
+ * one route #237 left open. No fixture in the corpus sources either field
+ * legitimately from below the profile band, so the footer shape the issue
+ * worried about is not represented and is not being traded away here.
+ *
+ * WHAT THIS COSTS, STATED PLAINLY. A personal site printed in a FOOTER is lost.
+ * The text tier no longer reaches it, and the annotation tier does not rescue
+ * it either: `pdf-extract.ts` flips pdfjs' bottom-origin y to top-origin
+ * (`yTop = viewport.height - yBottom`), so y grows DOWNWARD, and
+ * `inProfileSection` accepts `ann.yTop <= profileLineMaxY + 12` — at or above
+ * the last profile line plus about one line-height of slack. That band is the
+ * header, the opposite end of the page from a footer. The only surviving route
+ * is the résumé hyperlinking the site from its header block, where the
+ * annotation lands inside the band. That is the accepted cost of option (a),
+ * which #610 explicitly permits; no corpus fixture exercises the footer shape.
+ * (`linkedin_url` / `github_url` are unaffected — their doc-wide `anywhereOnDoc`
+ * fallback is deliberately kept, so a footer profile icon still resolves.)
+ */
+const PORTFOLIO_WEBSITE_BAND_TO_PROFILE = true;
+
+/**
  * Scans `lines` for a candidate location string, checking US patterns first
  * then international. Returns the first match found, or `undefined`.
  *
@@ -444,14 +495,20 @@ export function extractContact(
   // `/in/` redirect printed as its own label (LaTeX `\href{url}{url}`) fills
   // `linkedin_url` from the annotation AND `website_url` from the text scan —
   // the exact double-render this PR removes (#378).
+  //
+  // `bandToProfileSection` (#610) drops the doc-wide TEXT tier for a field,
+  // leaving only the profile-band text scan and the profile-band annotation
+  // lookup — see `PORTFOLIO_WEBSITE_BAND_TO_PROFILE` for why the two link
+  // catch-alls opt in and the two identity links do not.
   const pickUrl = <K extends "linkedin_url" | "github_url" | "portfolio_url" | "website_url">(
     key: K,
     resolveAnnotation: () => string | undefined,
     reject?: (url: string) => boolean,
+    bandToProfileSection = false,
   ): { value: string | undefined; confidence: number } => {
     if (primary[key] && !reject?.(primary[key]!))
       return { value: primary[key], confidence: primary.confidence[key] };
-    if (fallback[key] && !reject?.(fallback[key]!))
+    if (!bandToProfileSection && fallback[key] && !reject?.(fallback[key]!))
       return {
         value: fallback[key],
         confidence: fallback.confidence[key] * 0.9,
@@ -503,6 +560,7 @@ export function extractContact(
         inProfileSection,
       ),
     (u) => claimedByIdentity.has(claimKey(u)),
+    PORTFOLIO_WEBSITE_BAND_TO_PROFILE,
   );
   // Website is the catch-all: any remaining URL not already claimed by
   // linkedin/github/portfolio — on ANY tier (text scan or annotation), so a
@@ -525,6 +583,7 @@ export function extractContact(
         inProfileSection,
       ),
     (u) => claimedUrls.has(claimKey(u)),
+    PORTFOLIO_WEBSITE_BAND_TO_PROFILE,
   );
 
   // Ownership (#134): claim the document-wide body lines that are nothing but a
--- a/src/lib/ingest/markdown.ts
+++ b/src/lib/ingest/markdown.ts
@@ -13,6 +13,12 @@
  * no I/O: the caller reads the File via `file.text()` before calling this.
  */
 
+import {
+  extractLinkDefinitions,
+  flattenAutolinks,
+  resolveReferenceLinks,
+} from "../markdown-link-refs.ts";
+
 export interface MarkdownParseResult {
   rawText: string;
   markdown: string;
@@ -29,15 +35,43 @@ export interface MarkdownParseResult {
  * underscores (`snake_case_name`, an SDK symbol, a slug) survive verbatim —
  * matching CommonMark, which disallows single-underscore emphasis mid-word.
  * Asterisk emphasis has no intraword ambiguity and is stripped directly.
+ *
+ * EVERY link rule here is shared with `markdown-lines.ts` or mirrors it exactly,
+ * because the two readings of one `.md` must agree: `rawText` is not merely
+ * scorer input — `EvidencePanel` prints it back to the user verbatim — so any
+ * shape this misses is raw markdown shown to the user AND a disagreement with
+ * what the extractors saw.
+ *
+ *   - Autolinks (`<https://…>`) flatten through the shared `flattenAutolinks`
+ *     (#610). Skipping them here left `<https://linkedin.com/in/…>` — angle
+ *     brackets and all — in the panel while the extractors read the bare URL.
+ *   - Reference-style links are the one shape that cannot be handled
+ *     line-by-line, so their `[ref]: url` definition table is built over the
+ *     whole document first (#611). A definition line surviving here is a
+ *     visible markdown artifact; a `[label][ref]` surviving here disagrees with
+ *     the `label url` the other reading produces.
+ *   - The inline rule is title-aware, matching `markdown-lines.ts`'s
+ *     `INLINE_LINK_RE`. Without that, `[writeup](url "Pricing rebuild")`
+ *     swallowed the CommonMark title into the target and printed
+ *     `writeup url "Pricing rebuild"`.
  */
 export function mdToPlainText(text: string): string {
-  return text
+  const { definitions, body } = extractLinkDefinitions(text);
+  return body
     .split("\n")
     .map((line) =>
-      line
-        .replace(/^\s{0,3}#{1,6}\s+/, "") // heading markers
-        .replace(/^\s*[-*+]\s+/, "") // list bullet markers
-        .replace(/\[([^\]]*)\]\(([^)]+)\)/g, "$1 $2") // [text](url) → text url
+      // Reference resolution runs with the link rules, before emphasis: it
+      // must see the brackets that `[text](url)` has already surrendered.
+      resolveReferenceLinks(
+        flattenAutolinks(
+          line
+            .replace(/^\s{0,3}#{1,6}\s+/, "") // heading markers
+            .replace(/^\s*[-*+]\s+/, "") // list bullet markers
+            // [text](url) → text url, dropping an optional CommonMark title
+            .replace(/\[([^\]]*)\]\(\s*([^)\s]+)(?:\s+"[^"]*")?\s*\)/g, "$1 $2"),
+        ),
+        definitions,
+      )
         .replace(/\*\*(.+?)\*\*/g, "$1") // bold (asterisk)
         .replace(/\*(.+?)\*/g, "$1") // italic (asterisk)
         .replace(/(^|\W)__(\S(?:.*?\S)?)__(?=\W|$)/g, "$1$2") // bold (underscore, word-boundary)
--- a/src/lib/markdown-link-refs.ts
+++ b/src/lib/markdown-link-refs.ts
@@ -1,0 +1,238 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * Shared markdown link flattening — autolinks (#610) and the doc-wide
+ * reference-link table (#611).
+ *
+ * #610 flattened the link shapes that are self-contained on one line
+ * (`[label](url)`, `<url>`). Reference-style links are not: `[label][ref]`,
+ * `[label][]` and bare `[label]` all carry their target on a separate
+ * `[ref]: https://…` *definition* line, conventionally at the bottom of the
+ * document. Resolving them therefore needs a document-wide index, which is why
+ * this lives in its own zero-dep module rather than inside either consumer.
+ *
+ * Two consumers, one table — that shared table is the point. A dropped `.md`
+ * is parsed twice: `markdown-lines.ts` turns the markdown into `PdfLine[]` for
+ * the extractors, and `ingest/markdown.ts`'s `mdToPlainText` flattens the same
+ * file into the `rawText` the scorer scans and `EvidencePanel` renders
+ * verbatim. #610 established that those two readings must agree; duplicating
+ * the definition table would be the fastest way to break that agreement again.
+ *
+ * `flattenAutolinks` lives here for the same reason, even though an autolink
+ * needs no table: it is a #610 rule, and leaving it in `markdown-lines.ts`
+ * meant only ONE of the two readings applied it — a dropped `.md` rendered
+ * `<https://linkedin.com/in/…>` verbatim in the Evidence panel while the
+ * extractors saw the bare URL. One definition, both readings.
+ *
+ * The operations this module exposes are load-bearing:
+ *
+ *   - `extractLinkDefinitions` blanks every definition line. A `[ref]: url`
+ *     line is document metadata, never résumé content, and left in place it
+ *     leaks three different ways depending on where it sits (measured on the
+ *     #610 baseline): appended verbatim into `summary` when it trails a prose
+ *     section, promoted to the contact `website_url` **plus** a bogus
+ *     `kind: "other"` profile entry when it sits above the first section
+ *     header — the one placement #610's profile-banding does not cover — and,
+ *     everywhere, printed into `rawText` for the user to read in the Evidence
+ *     panel. Blanking rather than deleting keeps line positions stable, and
+ *     `classifyMarkdownLine` already drops blank lines.
+ *   - `resolveReferenceLinks` rewrites the usage to `label url`.
+ *
+ * WHY `label url` AND NOT BARE `label`. The corpus argues for the cheap option
+ * and against it at the same time. No fixture, and no `.md` in the tree, uses
+ * the reference shape at all, so this is a rare shape and a big pre-pass would
+ * not be worth it — hence one regex and a `Map`, not a CommonMark parser. But
+ * rarity says nothing about which *output* is right, and bare `label` would be
+ * strictly worse than shipping nothing: today the target at least survives as
+ * literal text on the definition line, whereas blanking definitions (which is
+ * mandatory) and emitting bare `label` would erase the URL from the document
+ * entirely, so `liftHeaderLabel` could never lift `project.url` /
+ * `achievement.url` from it. `label url` is also exactly what #610's
+ * `flattenLinks` and `mdToPlainText` already emit for the inline shape.
+ *
+ * WHY SHORTCUT AND COLLAPSED REFS ARE IN SCOPE. `[label]` on its own is a
+ * *shortcut* reference link, and `[label][]` a *collapsed* one; excluding them
+ * would leave the exact `[`/`]` residue this change exists to remove. They are
+ * safe to claim precisely because resolution is table-driven: a label with no
+ * matching definition is returned untouched, which is both CommonMark's own
+ * reading (an undefined reference IS literal text) and what protects ordinary
+ * bracketed prose — the `[2019]` year marker the achievements extractor reads
+ * is never a link here, because nothing defines `[2019]`.
+ *
+ * DELIBERATE CommonMark SUBSET. A definition must fit on one line and its
+ * target must carry a scheme (`https:`, `mailto:`) or start `www.`. Multi-line
+ * definitions and titles-on-the-next-line are not recognized, and the
+ * scheme requirement is what stops a prose line like
+ * `[2019]: awarded the prize` from being eaten as metadata. Label matching
+ * follows CommonMark: case-insensitive, internal whitespace collapsed.
+ */
+
+/**
+ * A one-line link definition: `[label]: <target> "optional title"`.
+ *
+ * `^ {0,3}` mirrors CommonMark's indent allowance. The target class rejects
+ * whitespace and angle brackets so an angle-wrapped target unwraps cleanly,
+ * and the trailing group tolerates all three CommonMark title delimiters. The
+ * line must end after the title — a trailing sentence means this was prose.
+ * The end anchor is a LOOKAHEAD so a CRLF document keeps its `\r`: the match is
+ * replaced with the empty string, and consuming the `\r` would silently
+ * downgrade that one line ending to a bare `\n`.
+ */
+const LINK_DEFINITION_RE =
+  /^ {0,3}\[([^\]\n]+)\]:[ \t]*<?((?:[A-Za-z][A-Za-z0-9+.-]*:|www\.)[^\s<>]*)>?[ \t]*(?:"[^"]*"|'[^']*'|\([^)]*\))?[ \t]*(?=\r?$)/gm;
+
+/**
+ * CommonMark autolinks: `<https://example.com>` and `<jane@example.com>`. There
+ * is no separate label, so the visible text becomes the target itself — which
+ * is exactly how a bare URL already reads today. Both alternatives require a
+ * `scheme:` or an `@`, so an HTML tag that survived into the markdown (`<b>`,
+ * `<br/>`) is never touched (#610).
+ */
+const AUTOLINK_URI_RE = /<([A-Za-z][A-Za-z0-9+.-]{1,31}:[^<>\s]*)>/g;
+const AUTOLINK_EMAIL_RE = /<([^<>\s@]+@[^<>\s@]+\.[^<>\s@]+)>/g;
+
+/**
+ * A reference-style IMAGE usage: `![alt][ref]`, `![alt][]`, `![alt]`.
+ *
+ * The alt class admits ONE level of nested brackets — `[^[\]\n]` for ordinary
+ * characters, `\[[^[\]\n]*\]` for a balanced inner pair — because CommonMark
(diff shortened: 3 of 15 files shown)


src/lib/markdown-link-refs.ts
```diff
@@ -0,0 +1,238 @@
...
+ * inert — including the bracket-bearing alt the flat class used to split.
+ *
+ * Ordering the image alternative FIRST is what makes it win: JavaScript
+ * alternation is leftmost-first, so at a `!` the image branch is tried before
+ * the plain-link branch. The image branch captures nothing, so an `undefined`
+ * label group is the signal that it matched — `undefined`, not falsy: a
+ * degenerate `[][ref]` captures the empty string and is still a link.
+ */
+const REFERENCE_LINK_RE =
+  /!\[(?:[^[\]\n]|\[[^[\]\n]*\])*\](?:\[[^\]\n]*\])?|\[([^\]\n]*)\](?:\[([^\]\n]*)\])?/g;
```
rohithgollapalli, line 141: **Secondary — the link branch here is missing the bracket-nesting the image branch above it has.**

The image alternative admits one level of nesting (`(?:[^[\]\n]|\[[^[\]\n]*\])*`), and its docblock explains exactly why: a flat class "stops at the inner `]` ... leaving a `[logo]` tail that shortcut resolution then turns back into a URL in body text." The **link** alternative still uses the flat `[^\]\n]*`, so that argument applies to it verbatim — and the failure reproduces:

```
input:   Led the [a [b] c][ref] project
         [ref]: https://example.org/target

both readings: "Led the [a [b] c]ref https://example.org/target project"
```

CommonMark reads `[a [b] c][ref]` as one link with label `a [b] c`. What ships is a stray `]`, a leaked `ref` token, and a URL spliced into résumé body text — the exact output shape round 2 fixed for images.

Not blocking: rare in a résumé, both readings agree so the headline invariant holds, and `main` was also wrong here (it left the whole usage literal). But the fix is the class you already wrote:

```diff
-  /!\[(?:[^[\]\n]|\[[^[\]\n]*\])*\](?:\[[^\]\n]*\])?|\[([^\]\n]*)\](?:\[([^\]\n]*)\])?/g;
+  /!\[(?:[^[\]\n]|\[[^[\]\n]*\])*\](?:\[[^\]\n]*\])?|\[((?:[^[\]\n]|\[[^[\]\n]*\])*)\](?:\[([^\]\n]*)\])?/g;
```

I ran the proposed class against the shapes you care about before suggesting it — no regressions:

| Input | Output |
|---|---|
| `Led the [a [b] c][ref] project` | `Led the a [b] c https://example.org/target project` |
| `The [handbook] rocks` | `The handbook https://example.org/hb rocks` |
| `Owned [warehouse indexer][missing] rewrite` | unchanged |
| `prize [2019] awarded` | unchanged |
| `see [][ref] now` | `see https://example.org/target now` (still `""`, not `undefined`) |
| `logo ![a [b] c][logo] here` | unchanged — image branch still wins |
| `Wow![great] news` | unchanged |

One deliberate behaviour change worth knowing: an unbalanced `[a [b]` with `b` defined now resolves the inner `[b]` instead of being left alone. That is closer to CommonMark, which reads it as literal `[a ` followed by a real shortcut link.

src/lib/ingest/markdown.ts
```diff
@@ -29,15 +35,43 @@ export interface MarkdownParseResult {
...
+  return body
     .split("\n")
     .map((line) =>
-      line
-        .replace(/^\s{0,3}#{1,6}\s+/, "") // heading markers
-        .replace(/^\s*[-*+]\s+/, "") // list bullet markers
-        .replace(/\[([^\]]*)\]\(([^)]+)\)/g, "$1 $2") // [text](url) → text url
+      // Reference resolution runs with the link rules, before emphasis: it
+      // must see the brackets that `[text](url)` has already surrendered.
+      resolveReferenceLinks(
```
rohithgollapalli, line 65: **Secondary — a *defined* reference image is now the one shape where the two readings disagree, and it is not in Known limitations.**

`flattenLinks` calls `stripReferenceImages`; `mdToPlainText` does not. So:

```
input:  Shipped ![badge][ci] on time
        [ci]: https://example.org/ci.svg

PdfLine reading: "Shipped  on time"
rawText reading: "Shipped ![badge][ci] on time"
```

`markdown-readings-agree.test.ts:106-120` asserts this divergence deliberately, so it is known — but it is **new**. On `main` both readings left the usage literal (`ANY_IMAGE_RE` and the old inline rule both require `](`), so they agreed. `EvidencePanel` prints `rawText` verbatim, so this is raw markdown shown to the user.

The Known-limitations list discloses the *inline* `alt` case as pre-existing, but not this one — as written it reads as though images are uniformly untouched on the `rawText` side.

Either fix works, and I would take the first — one line, and it makes the invariant unconditional:

```diff
       resolveReferenceLinks(
         flattenAutolinks(
-          line
+          stripReferenceImages(line, definitions)
```

If you would rather keep the two readings' image handling asymmetric on purpose, that is defensible — just add it to Known limitations next to the inline case.

**Nit, same area:** `stripReferenceImages` replaces with `""`, so the strip leaves a double space (`Shipped  on time`). Same class as the `achievement.url` trailing-`;` double space already listed. Non-blocking.

src/lib/heuristics/extract/contact.ts
```diff
@@ -214,6 +214,57 @@ function findConsumedLines(
...
+ * `inProfileSection` accepts `ann.yTop <= profileLineMaxY + 12` — at or above
+ * the last profile line plus about one line-height of slack. That band is the
+ * header, the opposite end of the page from a footer. The only surviving route
+ * is the résumé hyperlinking the site from its header block, where the
+ * annotation lands inside the band. That is the accepted cost of option (a),
+ * which #610 explicitly permits; no corpus fixture exercises the footer shape.
+ * (`linkedin_url` / `github_url` are unaffected — their doc-wide `anywhereOnDoc`
+ * fallback is deliberately kept, so a footer profile icon still resolves.)
+ */
+const PORTFOLIO_WEBSITE_BAND_TO_PROFILE = true;
```
rohithgollapalli, line 266: **Nit** — this reads like a feature flag, but it can never be `false`: TypeScript narrows it to the literal `true`, and nothing ever passes anything else to `bandToProfileSection`.

The docblock hanging off it is genuinely worth keeping — the measured-cost and why-not-option-(b) reasoning is the most useful prose in this file. Consider moving it onto `pickUrl`'s `bandToProfileSection` parameter and passing `true` at the two call sites, so no future reader goes hunting for the toggle that flips this constant.

Non-blocking either way.

tests/fixtures/markdown/reference-links.md
```diff
@@ -0,0 +1,47 @@
+# Avery Okonkwo
+
+avery.okonkwo@example.com · (312) 555-0155 · Austin, TX
```
rohithgollapalli, line 3: **Nit** — Avery Okonkwo is in **Austin, TX** with a `(312)` **Chicago** area code.

Fully policy-compliant: the rule requires a *real* area code plus the `555` exchange and an `0100`-`0199` subscriber, and `(312) 555-0155` satisfies all three — this is realism only, not a PII finding. `(512) 555-0155` would match the city if you want the fixture internally consistent. (`inline-links.md`'s Riley Nakamura is Chicago + `(312)`, so that one already lines up.)

rohithgollapalli (approved):
Approving. The core of this is right, and the docblock reasoning is unusually load-bearing — single-sourcing the definition table and the autolink rule into `markdown-link-refs.ts` is what makes the "two readings agree" claim checkable instead of aspirational, and `markdown-readings-agree.test.ts` is the test that should have existed since #552. The #610 contact banding is correctly scoped: I confirmed the doc-wide **text** tier is dropped only for `portfolio_url`/`website_url`, while the annotation tier and the `linkedin_url`/`github_url` doc-wide fallbacks are untouched.

I verified the claims rather than taking the body's word for them:

| Probe | Result |
|---|---|
| Profile-band URL to the card | `portfolio_url: https://janeroe.dev` OK |
| Body-bullet URL to the card | `{}` — the #610 defect is gone |
| Footer site to the card | `{}` — the disclosed cost, reproduced exactly as stated |
| LinkedIn/GitHub in a footer | both still resolve |
| ReDoS on the nested image-alt class | 5 ms at 8 000 nested `[...]` pairs — no blowup; the alternation's two branches are disjoint on their first character |
| Quadratic claim | 20 000 unmatched `[` in 1.7 s, consistent with the stated asymptotics |

Two **Secondary** findings, both in the same narrow area — bracket shapes the reference scan does not fully claim — plus four nits. All four are inline on the lines they concern. Neither Secondary fires on any shape in the two fixtures or in the corpus.

Summary of what is inline:

- **Secondary 1** (`markdown-link-refs.ts:141`) — the link branch of `REFERENCE_LINK_RE` is missing the bracket-nesting its own image branch has, so `[a [b] c][ref]` emits a stray `]` and a spliced URL. Fix included and validated.
- **Secondary 2** (`ingest/markdown.ts:65`) — `mdToPlainText` never calls `stripReferenceImages`, making a *defined* reference image the one shape where the two readings now disagree. Known and tested, but **new**, and absent from Known limitations.
- **Nits** — the double space the image strip leaves behind (folded into Secondary 2's thread); `PORTFOLIO_WEBSITE_BAND_TO_PROFILE`; a persona/area-code mismatch in `reference-links.md`.
- `extractContact` is now 215 lines (fallow). Pre-existing shape, +10 from this PR — noted only so it does not drift further.

## Gates

- **Fixture PII (3a)** — pass. `npm run check:fixtures` green, 54 PDFs. The two new `.md` fixtures fall **outside** that gate, so I read both myself: Riley Nakamura / Avery Okonkwo, `@example.com` throughout, `(312) 555-0123` / `(312) 555-0155` — real area code, `555` exchange, subscriber in `0100`-`0199`. Every link target is `example.org`/`example.net`/`example.com`. No real persona, no OSS-template demo content. `tests/fixtures/markdown/README.md` documenting that the gate does not cover this directory is the right call.
- **Design-system (3b) / style tokens (3c)** — N/A. Pure `src/lib/` + tests + fixtures; no feature component, no styling, no new file under `src/components/`.
- **Skill/script command bugs (3e)** — N/A. Nothing under `.claude/skills/` or `scripts/`.
- **Dead-code / fallow (3d)** — clean. Every new export (`extractLinkDefinitions`, `resolveReferenceLinks`, `stripReferenceImages`, `flattenAutolinks`) has an in-repo consumer. The 2 complexity findings on `normalizeSplitLetterHeaders` / `classifyMarkdownLine` are reported by fallow as **inherited** ("audit gate excluded 2 inherited findings"), which confirms your line-shifted-not-introduced read. Report-only per `CLAUDE.md` regardless.
- **Local run** — `npm run typecheck` clean, `npm run lint` clean, `npx vitest run src/lib/heuristics/ src/lib/ingest/ ...` green at **1181 passed / 8 skipped**, including the corpus suite. The single `two-column-achievements-sidebar.expected.json` move is `website_url` only, matching the body.

Verdict rule applied: **0 Blocking, so APPROVE.** Both Secondary findings are rare-shape correctness, not merge blockers; nits are explicitly non-blocking. Reviewed at `9b910e6`.

---

Issue #625, opened 2026-07-28
[edit] Summary is parsed, scored, and exported but never displayed or editable


s-annam:
## Summary

The Summary section is parsed, scored, and **rendered into the downloaded PDF** — but it is never displayed in the reconstructed résumé and there is no way to edit it. The user cannot see the text, cannot correct a mis-parse, and cannot remove it, yet it ships in the artifact they hand to an employer.

Every other parsed section (contact, experience, education, skills, achievements, profiles) is both visible and editable. Summary is the only one that isn't.

## Evidence

**It is parsed and it does reach the export.**

- `ats-resume-model.ts:898` — `summary: parsed.summary?.trim() || undefined`
- `render-ats-pdf.ts:886-888` — draws the heading and body:
  ```ts
  if (model.summary) {
    drawSectionHeading(layout, model.summaryHeading ?? "Summary");
    layout.drawText(model.summary, { size: SIZE_BODY });
  }
  ```
- `summaryHeading` is preserved verbatim (#285), so this is a fully-modelled section, not an accident.

**It is scored.** `COMPLETENESS_SUMMARY_MIN_CHARS = 20` (`score.ts:114`) — a summary under 20 chars costs Completeness points. The user has no way to see what was parsed, so a truncated or mis-segmented summary silently costs them score with no visible cause.

**But the UI only uses it to feed the LLM.** The single consumer of `parsed.summary` in `ReconstructedResume.tsx` is `buildRewriteSections` (`:952-963`), which registers it as a rewrite target:

```ts
const trimmedSummary = summary?.trim();
if (trimmedSummary) {
  ... kind: "summary", id: "summary", label: "Summary", text: trimmedSummary ...
}
```

There is no `<Summary>` display block and no `EditableField`. Grepping the whole component for `summary` returns only the rewrite-section builder and a docblock.

**There is no override channel.** `EditableParse` carries `contactOverrides`, `experienceOverrides`, `bulletOverrides`, `educationOverrides`, `skillsOverride`, `achievementOverrides`, `descriptionOverrides`, `profileOverrides` — and no `summaryOverride`. Even if a display block were added, there is nothing behind it to write to.

## Why this matters

The reconstructed résumé is the product's central claim: *here is what a generic text extractor read back from your PDF.* A section that is read, scored, and re-exported but never shown breaks that contract — the preview is not the artifact.

[...]

PR #630, opened 2026-07-28
fix(pdf): never strand a header, heading, or lone line at a page break (#629, #631, #632)

### LINKED ISSUE #629: ATS PDF export strands entry headers and orphans bullets at page breaks
## Problem

The ATS PDF export can strand an entry header alone at the bottom of a page, with its bullets flowing to the next page. It can equally orphan a single trailing bullet at the top of the next page.

Observed twice in one editing session on a two-page résumé export:

1. A `Senior Engineering Manager · Acme Corp, Springfield, IL · Platform A` block split 3 bullets / 1 bullet across the page boundary — one lone bullet at the top of page 2.
2. After fixing that, `Technical Lead & Manager · Acme Corp, Springfield, IL · Product B, Product C, Product D` rendered as the **last line of page 1** with all four of its bullets on page 2.

Both times the only available workaround was to **edit résumé prose** — padding or trimming a bullet until the break happened to land somewhere acceptable. That is content being distorted to compensate for a layout defect, which is the wrong direction of causality for a tool whose pitch is résumé/PDF correctness.

## Root cause

`src/lib/pdf/render-ats-pdf.ts` paginates one text line at a time. `ensure()` is the only page-break decision:

```ts
/** Ensure room for `height` pt; add a page if the cursor would overflow. */
private ensure(height: number) {
  if (this.y - height < CONTENT_BOTTOM) this.newPage();
}
```

It is called with a single `lineHeight` at three sites — `:639`, `:756`, `:774` — and nowhere else. There is no notion of a block, no keep-with-next, and no widow/orphan control. So a header line always fits if one line's worth of space remains, regardless of whether anything that belongs with it also fits.

Note this is a **pdf-lib manual-layout** renderer, not an HTML/Chromium print path — CSS `break-after: avoid` is not applicable here. The fix has to live in the layout engine.

## Proposed fix

Add a block-aware reservation alongside the existing per-line `ensure()`:

```ts
/** Ensure room for `height` pt as an indivisible unit. */
private ensureBlock(height: number) {
  if (this.y - height < CONTENT_BOTTOM) this.newPage();
}
```

Then, before drawing an entry header, measure and reserve the header **plus at least its first bullet line**, so a header can never be the last thing on a page. Two further rules worth adopting in the same pass:

- **Orphan control on bullets** — a wrapped bullet should not leave its first line alone at a page bottom, nor its last line alone at a page top. A "multi-line bullet" here means a bullet whose text wraps to 2 or more drawn lines; reserve 2 lines when starting one.
- **Section headings** (`EXPERIENCE`, `ACHIEVEMENTS`, `SKILLS`) should keep-with-next as well, under the same 2-line rule as entry headers: a section heading must be followed by at least one line of its content on the same page.

Measurement is already available — `wrap()` computes line counts before drawing, so the reservation height can be derived without a second layout pass.

**Composition decision:** the header reservation and the bullet-orphan reservation compose rather than double-count. A header whose first bullet itself wraps to multiple lines still only needs to reserve header + 1 bullet line to commit to the page — it does not additionally need the full multi-line bullet reserved up front. (This is a design decision for the implementer to follow, not a mechanism already proven elsewhere in the codebase.)

## Why it matters beyond one résumé

Any user with a résumé longer than one page can hit this, and the failure is silent — the PDF renders successfully and looks broken only on inspection. For a product that positions itself as a résumé and PDF-parser diagnostic, shipping stranded headers undercuts the core claim.

## Acceptance

- An entry header is never the final drawn line on a page: header + at least one full bullet line must fit together, or both move to the next page.
- A multi-line bullet never splits 1 line / rest across a page boundary.
- A section heading is never the final drawn line on a page: it must be followed by at least one line of its content on the same page, or both move to the next page.
- Regression test: extend the existing `src/lib/pdf/export-layout-contract.test.ts` (not a new sibling file) — synthesize content whose natural break lands exactly on a header, assert the header moves to the next page. No real-PDF fixture; content is synthesized.

### LINKED ISSUE #631: ATS PDF export orphans a wrapped bullet's last line at a page top (widow half of #629)
## Problem

The ATS PDF export can leave the **last line of a wrapped bullet alone at the top of a page**. A bullet that wraps to 3 or more drawn lines splits 2 lines / rest across the boundary, stranding a single trailing line above the next bullet.

This is the widow half of the page-break family. #629 fixed the orphan half (a bullet never splits 1 line / rest) and #630 shipped it, but the widow case is untouched — it reproduces identically before and after that change.

## Why #629 did not cover it

#629's prose asked for both halves:

> a wrapped bullet should not leave its first line alone at a page bottom, **nor its last line alone at a page top**

but the mechanism it then specified cannot deliver the second half:

> A "multi-line bullet" here means a bullet whose text wraps to 2 or more drawn lines; **reserve 2 lines when starting one**.

A 2-line reservation guarantees the *first* two lines land together. It says nothing about where the tail falls, so a 3-line bullet reserves 2, draws 2, and breaks before line 3. #629's acceptance criterion was written to the mechanism ("never splits 1 line / rest"), not to the prose, so the implementation met its criteria with the widow case still open. This was flagged in #630 rather than silently dropped.

## Reproduction

Synthesized during #630's review, against the real renderer:

- Build a résumé whose content walks a 3-line bullet down to a page boundary (the review harness sweeps a filler-entry count; the case lands at `filler=43` under the test suite's Helvetica fallback).
- Render via `renderAtsResumePdf`, then read drawn lines per page with `extractPdfDrawnLines` (`src/lib/pdf/render-ats-pdf.test-utils.ts`).
- Observe the bullet occupying 2 lines at the bottom of page N and 1 line at the top of page N+1.

Reproduces on `main` both before and after #630.

## Root cause

`src/lib/pdf/render-ats-pdf.ts` → `drawBullet`:

```ts
const lineHeight = size * LINE_GAP;
if (!opts.alreadyReserved) {
  const height = this.measureBulletHeight(text, size, hangingIndent);
  if (height > lineHeight) this.ensureBlock(2 * lineHeight);
}
```

The reservation is a fixed 2 lines regardless of the bullet's measured height. `measureBulletHeight` already returns the full height, so the information needed is in hand — it is simply not used for the tail.

## Proposed fix

Widow control needs a different shape from orphan control: it is not "reserve more up front" but "do not leave fewer than N lines behind". Two candidate approaches, both worth weighing before implementing:

1. **Reserve the whole bullet** when it fits on an empty page. Simplest and fully correct for the common case, but it de-densifies pages for long bullets — the exact failure mode that made a blanket reservation wrong for the skills list in #630 (see that PR's *Notable design decisions*). Would need the same care.
2. **Break-position control** — when a bullet must split, choose a split leaving ≥2 lines on both sides, pushing an extra line forward when the natural break would leave a lone tail. Preserves density; needs the draw loop to know the total line count up front, which `wrapRuns` now provides.

Option 2 is likely the right one given #630's density lesson, but this needs measurement, not assertion.

## Constraints inherited from #630

- This is **pdf-lib manual layout**, not an HTML/Chromium print path — CSS `orphans`/`widows` do not apply. The fix lives in the layout engine.
- Any reservation must be **measured** through the same wrapping that draws it (`measureBulletHeight` / `wrapRuns`), never estimated.
- `ensureBlock` ignores a reservation taller than one usable page, so an unsatisfiable widow reservation degrades safely rather than wasting a page.
- Verify against the **whole corpus by glyph-run position**, not page count — #630's review caught an 8-line density regression that page counts alone would have missed.

## Acceptance

- A bullet wrapping to 3+ drawn lines never leaves exactly one of its lines alone at the top of a page.
- No corpus fixture loses page density relative to `main`: compare all fixtures by exact glyph-run position (`page | x | y | str`), and report any fixture whose page-1 line count drops.
- The orphan guarantee from #629 still holds (no bullet splits 1 line / rest), including for an entry's first bullet.
- Regression test extends `src/lib/pdf/export-layout-contract.test.ts` — synthesized content, no real-PDF fixture, consistent with that file's existing `boundaryN` boundary-sweep pattern.
- Prove the new test guards by reverting the fix and confirming it fails.

## Notes

The pagination tests run under the **Helvetica fallback** — the Poppins embed fails in Node (`Failed to parse URL from /src/assets/fonts/Poppins-Regular.ttf`), so any filler boundary quoted above is Helvetica geometry, not the browser's Poppins. Boundaries will differ in a real browser; assert on relationships (which page a line landed on), never on hard-coded filler counts.

### LINKED ISSUE #632: ATS PDF export strands a lone trailing bullet at a page top (observed failure #1 from #629)
## Problem

The ATS PDF export can leave a **single trailing bullet alone at the top of a page**, separated from the rest of its entry. An entry with 4 bullets renders 3 on page N and 1 on page N+1, with nothing else from the entry following it.

This is the issue's original *observed failure #1* from #629:

> A `Senior Engineering Manager · Acme Corp, Springfield, IL · Platform A` block split 3 bullets / 1 bullet across the page boundary — one lone bullet at the top of page 2.

It reproduces on `main` today, both before and after #630.

## Why #629 did not fix it

#629 was filed off two observations but its four acceptance criteria only encode the **second** one (a header stranded at a page bottom). The criteria are about *headers and headings* keeping company with their content:

1. entry header + ≥1 bullet line move together
2. a multi-line bullet never splits 1 line / rest
3. a section heading + ≥1 content line move together
4. regression test in `export-layout-contract.test.ts`

None of these constrain how an entry's bullets distribute *among themselves* once the header is safely placed. Criterion 2 is about a **single bullet's own lines** wrapping, not about the group of bullets in an entry. So #630 satisfied all four criteria with this case untouched — confirmed by reproduction during its review, not assumed.

## Reproduction

Against the real renderer, synthesized:

- Build an entry with 4 single-line bullets, positioned so the natural break falls after the third (the #630 review harness hits it at `filler=43` under the test suite's Helvetica fallback).
- Render via `renderAtsResumePdf`, read per-page drawn lines with `extractPdfDrawnLines` (`src/lib/pdf/render-ats-pdf.test-utils.ts`).
- Observe 3 bullets on page N, 1 bullet alone at the top of page N+1.

## Root cause

`src/lib/pdf/render-ats-pdf.ts` → `drawEntry` reserves a keep-block covering the header plus its **first** bullet's orphan-safe opening, then draws the remaining bullets through the plain per-line `ensure()`:

```ts
const keepHeight = entryKeepHeight(layout, entry, mutedColor);
if (keepHeight > 0) layout.ensureBlock(keepHeight);
```

There is no notion of the bullet *list* as a unit, and no minimum on how many bullets may be left behind. Once the header is committed, each subsequent bullet paginates independently.

## Design question to settle first

Unlike #629's header case, there is no obviously correct rule here — this needs a decision before implementation:

- **Never leave exactly one bullet alone** (push a second one forward with it) is the direct analogue of widow control and is cheap.
- **Keep the whole entry together when it fits on an empty page** is stronger but risks the density regression #630 hit with the skills list — a blanket reservation cost one real fixture 8 page-1 lines (~100pt). Any variant of this needs the same corpus measurement.
- **Do nothing for a 1-line bullet, only for a wrapped one** is the narrowest reading and may not match what the original observation was complaining about.

A lone *short* bullet at a page top is arguably fine typographically; a lone bullet that is the only trace of a multi-bullet entry on that page reads as broken. Worth deciding which is actually being fixed.

## Constraints inherited from #630

- **pdf-lib manual layout**, not an HTML/Chromium print path — CSS `break-inside` / `orphans` do not apply.
- Reservations must be **measured** via the existing `measureBulletHeight` / `wrapRuns` seam, never estimated, so measure and draw cannot drift.
- Reserving a tall block indivisibly de-densifies pages — see #630's *Notable design decisions* for how the skills list forced a cap. Any entry-level reservation must be bounded.
- `ensureBlock` ignores an unsatisfiable reservation (taller than one usable page), so an oversized entry degrades safely.

## Acceptance

- Decide and document the rule (see *Design question*), then implement it.
- An entry's bullets do not distribute across a page boundary in the shape the rule forbids.
- No corpus fixture loses page density relative to `main`: compare all fixtures by exact glyph-run position (`page | x | y | str`), and report any fixture whose page-1 line count drops. Page-count comparison alone is insufficient.
- #629's guarantees still hold: an entry header is never the last line on a page, a bullet never splits 1 line / rest, a section heading always keeps ≥1 content line.
- Regression test extends `src/lib/pdf/export-layout-contract.test.ts` — synthesized content, no real-PDF fixture, following that file's `boundaryN` boundary-sweep pattern.
- Prove the new test guards by reverting the fix and confirming it fails.

## Notes

Related widow case filed separately: a bullet wrapping to 3+ lines splits 2 / 1, leaving its last line alone at a page top.

The pagination tests run under the **Helvetica fallback** (the Poppins embed fails in Node: `Failed to parse URL from /src/assets/fonts/Poppins-Regular.ttf`), so any filler boundary quoted above is Helvetica geometry, not the browser's Poppins. Assert on which page a line landed, never on hard-coded filler counts.

s-annam:
## Summary

The ATS PDF export paginated one text line at a time — `ensure()` was the only page-break decision and it was always handed a single `lineHeight`. So a header fit whenever one line's worth of space remained, regardless of whether anything belonging with it also fit. This adds a block-aware reservation (`ensureBlock`) plus the measurement needed to make a reservation exact, and applies keep-with-next at four sites: entry headers, wrapped bullets, an entry's bullet list, and section headings.

It now carries **#631 and #632** as well. Those were filed as follow-ups off #629 and implemented onto this branch rather than a second PR, because all three are the same rule at three scales — *never leave exactly one of anything alone* — and splitting them would have meant three separate corpus-density arguments over the same code.

The measurement is why the diff is larger than the fix sketch implies. Line breaking lived *inside* two draw loops, so it could not be called without drawing. It is split into a pure half and a draw half (`resolveDrawLines`, `wrapRuns`), letting `measureTextHeight` / `measureBulletHeight` / `measureHeaderRunsHeight` measure through the exact wrapping that will draw. The alternative — a second, parallel measurement implementation — is the precise failure mode that makes a reservation a guess. Drawn output is unchanged except where a reservation moves a block; this was verified rather than asserted (below).

Closes #629
Closes #631
Closes #632

## Verification

Every corpus fixture was compared **glyph-run by glyph-run** (`page | x | y | str`) against `main`, not just by page count:

- **4 fixtures fixed** — a stranded header or heading moved off a page bottom. e.g. `latex/deedy-resume-macfonts.pdf` no longer ends page 1 on the stranded header `MA, RTINIERE FOR BOYS`; `latex/multi-degree-coursework.pdf` no longer ends it on `B.S., Computer Science` + its sub-line.
- **0 fixtures regressed**, no page-count changes, no blank pages, and zero intra-line or x-position differences — which is also the proof that the `drawRuns` → `wrapRuns` refactor is behaviour-preserving.
- A synthesized genuine 3-line role header strands three ways on `main` (`filler=44,45,46`) and zero ways here.

**For #631 and #632** the same method was re-run against baseline `78a924c` (this PR's #629 commit), since #629's own corpus effect was already vetted against `main`:

- **#631 alone: byte-identical** across all 1542 drawn lines — the widow rule changes no corpus fixture.
- **#632: exactly one line moves.** `unknown/chromium-two-column-sidebar.pdf` goes p1 43→42, p2 17→18, page count unchanged. Inspected line-by-line: on the baseline that fixture ends page 1 with two bullets of a 3-bullet entry and opens page 2 on the third **alone**, immediately followed by the next role. That is a genuine instance of the #632 defect in the real corpus, and the one lost page-1 line is the cost of correcting it — not spurious de-densification.
- The complexity refactor (`decorateFirstLine` / `drawRunWords`) was diffed separately and is **byte-identical**, which is the proof it is behaviour-preserving.

Each of the four keep-with-next rules was reverted **individually** to confirm its own test fails — so no test passes on another rule's back. Sample failures: `filler=46: header stranded away from its first bullet: expected 1 to be 2`; `filler=34: skills block de-densified page 1: expected 37 to be greater than or equal to 39`.

## Notable design decisions

**A skills entry's header is body text, not a header.** The flat skills list carries the entire `·`-joined list on a single `headerLine` with `bullets: []`, so "reserve all wrapped header lines" turned an arbitrarily tall body block into a keep-with-next unit — costing 8 lines on one real fixture and ~100pt of page density. It is capped at two lines, gated on a **bullet-less** `headerBold: false` entry. Gating on `headerBold` alone would have been unsafe: a `headerBold: false` entry *with* bullets and a 3-line header reserves 2 header lines + 1 bullet line, letting the per-line `ensure` draw header line 3 and break before the bullet — violating the very criterion this issue is about. Gating additionally on `bullets.length === 0` keys the cap to the property that actually licenses it. Both skills paths hard-code `bullets: []`, so this changes no output today.

**An unsatisfiable reservation is ignored, not deferred.** A keep-block taller than one page previously fired `newPage()` twice with only the heading drawn between, producing `[13, 1, 52, 37]` — a page holding *only* the `SKILLS` heading, plus a wasted page. Now `[47, 52, 4]`, matching `main`.

## Known gaps (deliberate, not oversights)

> **Both gaps previously listed here are now closed in this PR** — #631 (widow half) and #632 (lone trailing bullet). They were filed as follow-ups when this PR covered only #629, then implemented onto the same branch. The two bullets that described them as unfixed have been removed rather than left to read as current.
- An entry with an empty `headerLine`, no sub-line and no bullets yields a zero-height keep-block, so a heading could strand. Unreachable from `buildAtsResumeModel` today (`"Education"` is a hard fallback and both skills paths have a non-empty `headerLine`); noted as defensive only.
- The new pagination tests run under the **Helvetica fallback** — the Poppins embed fails in Node (`Failed to parse URL from /src/assets/fonts/Poppins-Regular.ttf`), so every filler boundary is Helvetica geometry, not the browser's Poppins. Inherent to the harness.

`fallow` now reports **2** complexity findings, down from 4. The two that this PR's threads flagged are fixed: `drawText` (cognitive 29 → below threshold) and `drawRuns` (23 → below threshold), by extracting `decorateFirstLine` and `drawRunWords`. Both extractions move only drawing, never measurement, so the measure-and-draw-share-one-wrapper invariant is untouched — proven by the corpus diff being byte-identical across the refactor. The 2 survivors (`toWinAnsi` 20, `renderAtsResumePdf` 16) are pre-existing on `main` and were not what either thread cited.

`fallow` also reports 9 clone groups (up from 7), the added pair being boundary-sweep scaffolding in the new tests. **Deliberately not refactored:** `main` already carries 2 groups of the identical scaffolding and posted no duplication thread, so this class does not surface as a review comment here; `CLAUDE.md` classifies it as never-blocking; and collapsing 14 proven test bodies to save 19 lines would void the revert-based guard proofs for an advisory finding.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — 3406 passed / 10 skipped repo-wide; 311 across `src/lib/pdf/`. No `KNOWN_FAILURES` entry added or changed.
- [x] `npm run verify` green locally (fallow report-only, as configured)
- [x] Corpus glyph-position diff vs `main` — 4 fixed, 0 regressed; vs `78a924c` for #631/#632 — 1 lone-bullet case fixed, 0 regressed
- [x] Guard proofs re-run against the **final** code after the complexity refactor: reverting #631 alone fails 4 tests, reverting #632 alone fails 3, and neither breaks the other's
- [x] No fixture binary added or changed — nothing for the fixture-PII surface

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #629 | Claude Opus 5 (1M context) | ultra |
| Adversarial review — #629 (2 rounds) | Claude Opus 5 (1M context) | inherited session effort |
| Review fixes — B1/B2/B3 | Claude Opus 5 (1M context) | ultra |
| Implementation — #631 | unreported (requested: opus) | ultra |
| Implementation — #632 | unreported (requested: opus) | ultra |
| Adversarial review — #631/#632 | unreported (requested: opus) — returned no findings, see below | ultra |
| Complexity refactor (`drawText`/`drawRuns`) | Claude Opus 5 (1M context) | inherited session effort |
| Verification of #631/#632 | Claude Opus 5 (1M context) | inherited session effort |
| Orchestration + PR | Claude Opus 5 (1M context) | inherited session effort |

Every model string above is self-reported by the agent that ran the stage, or first-hand for orchestration. **Three rows read `unreported`**: those subagents completed their work but went idle without returning their structured report, and were re-prompted for the model name without answering. A requested alias is not evidence of what it resolved to, so the rows are left unresolved rather than filled with a guess.

The B2 gating decision and the final one-line hardening were made by the orchestrator, overriding the reviewer's proposed fix.

## Adversarial review — #629

Two bounded rounds ran pre-PR against the local diff, by an independent adversarial reviewer prompted to break the change rather than praise it, with instructions to **reproduce** suspected defects end-to-end rather than reason about the code. Round 1 returned `clean: false` with three blocking findings; round 2 returned `clean: true`.

### Round 1 — 3 blocking, all reproduced and all fixed

| # | Finding | Reproduction | Resolution |
|---|---|---|---|
| **B1** | Acceptance criterion 2 violated for the **first** bullet of every entry: `entryKeepHeight` reserved `head + 1` bullet line while `alreadyReserved` suppressed that bullet's own 2-line orphan reservation, so it split 1 line / rest. | `filler=45`, with the page-1/page-2 tails quoted. Also reproduces on `main`, so this was an **inherited** defect, not one this PR introduced. | Reserve two bullet lines when the first bullet wraps. Violation window `[45]` → `[]`. |
| **B2** | Reserving *all* wrapped header lines made the skills list indivisible and de-densified real pages. | Two corpus fixtures regressed: `…two-column-long-role-header.pdf` 40 → **32** page-1 lines (~100pt of new whitespace), `two-column-achievements-sidebar.pdf` 40 → 37. | Cap at two lines — see *Notable design decisions*. Both fixtures restored to `main`'s exact pagination. |
| **B3** | A keep-block taller than a page fired `newPage()` twice, stranding the heading on a page of its own — the exact case the reservation exists to prevent. The docblock asserted the opposite. | `[13, 1, 52, 37]` vs `main`'s `[47, 52, 4]`; page 2 held only the `SKILLS` heading. | Ignore an unsatisfiable reservation; docblock corrected. Restored to `[47, 52, 4]`. |

**The reviewer's B2 fix was rejected and replaced.** It proposed capping *every* header at two lines, which would have re-stranded a genuine 3-line role header — trading an AC1 regression for the B2 fix. Gating the cap on `headerBold === false` fixes skills while leaving real headers fully reserved; round 2 confirmed with a reproduction that the unconditional cap breaks `main`'s own 3-line-header case at three filler values while the gated one does not.

### Round 2 — clean, plus one hardening taken

No blocking findings. The reviewer additionally showed that `headerBold` alone is a **styling flag doing load-bearing layout work**: a `headerBold: false` entry *with* bullets and a 3-line header violates AC1 (reproduced at `filler=44`), unreachable today but guarded by nothing. It declined to call this blocking. It was taken anyway as a validated one-liner — `&& entry.bullets.length === 0` — so the invariant reads "AC1 holds" rather than "AC1 holds unless `headerBold === false`". Corpus output is byte-identical.

It also confirmed the reservation-vs-consumption arithmetic balances exactly (including float exactness: `LINE_GAP = 1.25` and all four font sizes are dyadic), that `resolveDrawLines` is genuinely pure so measure-then-draw cannot double-register a link annotation, and that `ats-resume-model.ts` is the sole `AtsEntry` producer reaching the render path.

### A vacuous test, caught

The B3 regression test's first version drove a 900-item skills list and **did not fail** when B3 was reverted — because the B2 cap makes the skills path satisfiable, so skills can no longer reach the unsatisfiable branch at all. It was reworked onto a genuine oversized bold achievement header, the only path that still reaches it. B3 is therefore defence-in-depth against an oversized real header, not a fix for a currently-reachable case. Recorded because a test that passes for the wrong reason is worse than no test.

### Surviving nits for the human reviewer

- `BODY_HEADER_KEEP_LINES = 2` and a value of `1` produce **identical output on every measured case** — the value is a semantic choice, not a measured one.
- `EMPHASIS_OPEN` is `U+E000`, a private-use codepoint that icon fonts commonly map into. A source PDF whose extracted skills text contains it would silently un-cap that entry. Pre-existing hazard of the sentinel design, noted only because `includes(EMPHASIS_OPEN)` is now load-bearing for a *layout* decision and not just a styling one.
- `render-ats-pdf.test-utils.ts:40` is now prettier-dirty (an 82-char line) where it was clean at `main`. No prettier gate in `verify`, so cosmetic.
- The remaining widow / lone-trailing-bullet / zero-keep-height gaps are listed under *Known gaps* above.

---

## Adversarial review — #631 / #632 (this round)

**Stated plainly: the independent reviewer for this round returned nothing.** It was spawned against the accumulated diff with the same adversarial brief as the #629 rounds, ran for ~12 minutes, then went idle without emitting its structured findings; a follow-up request for the report went unanswered. It produced **no** `clean: true`, no findings, and no model string. This section does not claim a clean independent review, because one did not complete.

Every acceptance criterion was therefore re-verified first-hand by the orchestrator rather than taken on an agent's word — the same posture used when the #629 implementation agent vanished:

- **Guard proofs, run twice each.** Once when each fix landed, and again against the *final* code after the complexity refactor (a refactor after a proof invalidates the proof). Reverting only #631 fails exactly its 3 tests plus the #631/#632 interaction test, with 67 passing; reverting only #632 fails exactly its 3 tests with 68 passing. Neither rule props up the other's tests, and none of the six is vacuous.
- **The `followKeepHeight` hand-off invariant was enumerated by hand**, since it is where a silent double- or zero-reservation would hide. The opening reservation carries the follow only when `bulletKeepLines(total) === total` (⇒ `total < 4`); `ensureWrappedLine`'s tail fires only when `total >= 4`. Disjoint, so the follow is never summed. The one shape whose opening is suppressed (`alreadyReserved` — a two-bullet entry's first bullet) is exactly the shape `entryKeepHeight` folds it into, and that fold's `keep === firstBulletLines` condition is the same indivisibility test, so it cannot coexist with a tail reservation either. The single uncovered case is `withFollow > USABLE_PAGE_HEIGHT`, where the follow is dropped **deliberately** so #629's header guarantee outranks #632's.
- **Regression classes from the #629 round were re-checked specifically.** The fold falls back to `base`, so it can never turn a satisfiable keep-block unsatisfiable and forfeit the header guarantee (the B2/B3 failure shape). `ensureWrappedLine` reserves at most ~5 lines against 684pt of usable page, so it cannot reproduce B3's double-`newPage()`.
- **Corpus density** measured by drawn-line position across all 54 fixtures, not page count — the method that caught B2's 8-line regression last round.

### Design decision recorded

#632's issue body listed three candidate rules and explicitly deferred the choice. **The maintainer chose "never leave exactly one bullet alone"**, rejecting "keep the whole entry together" (the ~100pt density trap from B2) and "only for wrapped bullets" (too narrow — #632's own reproduction uses four *single-line* bullets). The implementation is a single hand-off at one boundary, not an entry-level reservation, so the bounded cost is one bullet's keep-opening.

### Surviving nits for the human reviewer

- The one moved corpus line is a **real behaviour change**, not a no-op: `chromium-two-column-sidebar.pdf` loses a page-1 line so a lone third bullet rejoins its entry. Worth an eyeball to confirm the trade reads correctly.
- Test duplication grew by 2 clone groups; deliberately not refactored (rationale under *Known gaps*).
- `BULLET_KEEP_LINES = 2` is asserted as a local literal in the test file rather than imported, so the contract cannot be silently relaxed by editing the implementation constant. Intentional, but it does mean the two can drift apart into a *failing* test rather than a silently-passing one.

### DIFF
--- a/src/lib/pdf/render-ats-pdf.ts
+++ b/src/lib/pdf/render-ats-pdf.ts
@@ -19,6 +19,40 @@
  * word-wrapped by measuring with `font.widthOfTextAtSize`; bullets get a "• "
  * marker with a hanging indent.
  *
+ * Pagination is per line (`ensure`) PLUS keep-with-next reservations
+ * (`ensureBlock`, #629): a section heading reserves its rule and the first
+ * entry's keep-block, an entry header reserves all its wrapped header/sub-line
+ * lines plus its first bullet's keep-opening, and a standalone bullet that wraps
+ * reserves that same opening. So a heading or header can never be the last drawn
+ * line on a page.
+ *
+ * A wrapped bullet gets BOTH halves of the break-position guarantee: at least
+ * `BULLET_KEEP_LINES` of its lines sit on each side of any page break it
+ * straddles — never one line alone at a page bottom (#629, the orphan half) and
+ * never one line alone at a page top (#631, the widow half). The opening
+ * reservation (`bulletKeepLines`) covers the first side and, for a bullet too
+ * short to admit any legal split, the whole bullet; `ensureWrappedLine` places
+ * the break for the second side, pushing the trailing lines forward together
+ * rather than reserving the entire bullet — long bullets must stay divisible or
+ * pages de-densify.
+ *
+ * The same "never exactly one alone" rule applies one level up, to the BULLETS of
+ * an entry (#632): a page never opens on an entry's last bullet with nothing else
+ * of that entry above it. The second-to-last bullet carries the last one's
+ * keep-opening (`followKeepHeight`), so a break at that boundary moves both
+ * bullets forward together. Only that one boundary is constrained — earlier
+ * bullets paginate freely and no reservation ever spans the entry — because
+ * reserving a whole entry is the density failure #630 measured at ~100pt.
+ *
+ * Two deliberate limits keep those reservations from costing page density: a
+ * BODY-TEXT "header" (`headerBold: false` — the skills list) contributes at most
+ * `BODY_HEADER_KEEP_LINES`, and a reservation taller than a whole page is
+ * unsatisfiable and ignored outright. Every reservation is MEASURED by the same
+ * wrapping code that draws it (`measureTextHeight` / `measureBulletLines` /
+ * `measureHeaderRunsHeight` all share the draw path's line breaking), never
+ * estimated. This is manual pdf-lib layout, not an HTML/Chromium print path —
+ * CSS `break-after: avoid` / `orphans` / `widows` do not apply here.
+ *
  * The `rgb()` colors here are PDF graphics-state values (black text, a muted
  * gray rule) — NOT Tailwind tokens. The style guard scans component/feature
  * code, not this draw module.
@@ -43,6 +77,11 @@ const PAGE_HEIGHT = 792;
 const MARGIN = 54;
 const CONTENT_WIDTH = PAGE_WIDTH - MARGIN * 2;
 const CONTENT_BOTTOM = MARGIN;
+// Drawable height of a FRESH page — `newPage()` resets the cursor to
+// `PAGE_HEIGHT - MARGIN` and `ensure` stops at `CONTENT_BOTTOM`. This is exactly
+// the tallest reservation {@link Layout.ensureBlock} can satisfy, so anything
+// taller is unsatisfiable and must not be reserved (#629).
+const USABLE_PAGE_HEIGHT = PAGE_HEIGHT - MARGIN - CONTENT_BOTTOM;
 
 // ── Type scale (points) ───────────────────────────────────────────────────────
 //
@@ -79,9 +118,41 @@ const GAP_BEFORE_SECTION = 12;
 const GAP_AFTER_RULE = 6;
 const GAP_BETWEEN_ENTRIES = 7;
 const GAP_AFTER_HEADER = 2;
+// Vertical space the section-heading rule consumes (`drawRule`). Named because
+// the keep-with-next reservation for a section heading must include it (#629) —
+// the heading, its rule, and the first line of its content move as one unit.
+const RULE_HEIGHT = 2;
 
 const BULLET_MARKER = "• ";
 const BULLET_INDENT = 12; // hanging-indent width for wrapped bullet lines
+
+// Minimum drawn lines of a wrapped bullet that must land on EACH side of a page
+// break it straddles — the orphan half (#629, no lone line at a page bottom) and
+// the widow half (#631, no lone line at a page top) of one symmetric rule. It
+// follows that the legal split positions of an N-line bullet are exactly
+// `[BULLET_KEEP_LINES, N - BULLET_KEEP_LINES]`, an interval that is EMPTY when
+// `N < 2 * BULLET_KEEP_LINES` — which is why a three-line bullet is indivisible
+// and has to move whole (see `bulletKeepLines`).
+const BULLET_KEEP_LINES = 2;
+
+/**
+ * How many of a bullet's `total` drawn lines its OPENING reservation must keep
+ * together (#629/#631).
+ *
+ * A bullet with no legal split position (`total < 2 * BULLET_KEEP_LINES`) is
+ * indivisible: every break inside it would strand fewer than `BULLET_KEEP_LINES`
+ * lines on one side, so the reservation is the whole bullet and the break falls
+ * before it. That is the three-line case #631 is about — reserving two lines
+ * draws two and leaves the third alone at the next page's top.
+ *
+ * A longer bullet reserves ONLY its orphan-safe opening and stays divisible;
+ * {@link Layout.ensureWrappedLine} then places the break so the tail keeps
+ * `BULLET_KEEP_LINES` lines too. Reserving such a bullet whole would de-densify
+ * pages by its full height — the failure mode #630 measured on the skills list.
+ */
+function bulletKeepLines(total: number): number {
+  return total < 2 * BULLET_KEEP_LINES ? total : BULLET_KEEP_LINES;
+}
 
 // Minimum blank gutter (pt) kept between the wrapped header text and the
 // flush-right date tail (#436). Without a reserve the header text wraps to the
@@ -417,6 +488,43 @@ function groupRunsIntoWords(
 }
 
 /**
+ * Options for {@link Layout.drawText}. Named (rather than inlined in the
+ * signature) so {@link Layout.measureTextHeight} can take the IDENTICAL options
+ * — a keep-with-next reservation (#629) is only sound if it is measured by the
+ * same wrapping decision that will draw it.
+ */
+type DrawTextOpts = {
+  bold?: boolean;
+  size?: number;
+  color?: RGB;
+  x?: number;
+  hangingIndent?: number;
+  uppercase?: boolean;
+  atomicSegments?: boolean;
+  /** A short tail (a role/degree date range) drawn FLUSH-RIGHT, regular
+   *  weight, on the first wrapped line's baseline (#425). */
+  rightText?: string;
+  rightColor?: RGB;
+  rightSize?: number;
+  /** Register a clickable URI annotation over the whole first line (#425). */
+  linkUrl?: string;
+  /** Register a clickable URI annotation over each `display` substring found
+   *  in the first line (#425 — the contact line's link slugs). Applied only
+   *  when the text fits on ONE line, so measured offsets are accurate. */
+  linkSpans?: Array<{ display: string; href: string }>;
+  /** Paginate the wrapped lines under widow control (#631) — see
+   *  {@link Layout.ensureWrappedLine}. Set only by {@link Layout.drawBullet}; a
+   *  header/contact/summary block keeps the plain per-line behaviour. Not an
+   *  input to wrapping, so {@link Layout.measureTextHeight} ignores it. */
+  widowControl?: boolean;
+  /** Height (pt) that must land on the same page as this block's FINAL drawn
+   *  line — the NEXT bullet's keep-opening (#632). Set only by
+   *  {@link Layout.drawBullet}, and only for an entry's second-to-last bullet.
+   *  Like `widowControl`, not an input to wrapping. */
+  followKeepHeight?: number;
+};
+
+/**
  * Mutable cursor + page state threaded through the draw routines. We keep one
  * "current page" and append new pages as the cursor overflows.
  */
@@ -480,6 +588,81 @@ class Layout {
   /** Ensure room for `height` pt; add a page if the cursor would overflow. */
   private ensure(height: number) {
     if (this.y - height < CONTENT_BOTTOM) this.newPage();
+  }
+
+  /**
+   * Reserve `height` pt for an INDIVISIBLE block — a keep-with-next group such
+   * as "entry header + its first bullet line" or "section heading + rule + one
+   * line of content" (#629). Arithmetically identical to the per-line
+   * {@link ensure}; the difference is the caller's contract, not the maths.
+   * `ensure` is handed one line and may legitimately break right after it,
+   * whereas this is handed a whole group and guarantees the group STARTS on a
+   * page that can hold all of it — so the per-line `ensure` calls that draw the
+   * group cannot fire a second break inside it.
+   *
+   * A reservation TALLER than a whole empty page is UNSATISFIABLE, and is
+   * therefore ignored outright rather than broken to (#629). Deferring it cannot
+   * make it fit, and doing so actively causes the defect this method exists to
+   * prevent: on a "heading + oversized first entry" pair, the heading's
+   * reservation breaks to a fresh page, the entry's own equally-unsatisfiable
+   * reservation immediately breaks again, and the heading is left alone on a page
+   * of its own — a wasted page plus the exact stranding AC3 forbids. Ignoring it
+   * hands the block back to the per-line {@link ensure} calls that draw it, which
+   * fill the current page and break naturally. The comparison is against
+   * {@link USABLE_PAGE_HEIGHT}, so a block exactly that tall still counts as
+   * satisfiable.
+   */
+  ensureBlock(height: number) {
+    if (height > USABLE_PAGE_HEIGHT) return;
+    this.ensure(height);
+  }
+
+  /**
+   * Paginate ONE line (`index` of `total`) of a wrapped block, honouring widow
+   * control (#631).
+   *
+   * Without it, or on a block too short to split legally, this is the plain
+   * per-line {@link ensure}. With it, the line `BULLET_KEEP_LINES` from the end
+   * reserves the whole TAIL rather than itself: if the tail fits, the block
+   * cannot end with fewer than `BULLET_KEEP_LINES` lines on the next page; if it
+   * does not, the break happens HERE instead of one line later, carrying the tail
+   * forward as a unit. Only that one line needs the wider reservation — every
+   * earlier line breaking naturally leaves the tail intact, and every later line
+   * is inside a tail already placed.
+   *
+   * The other side of the break is guaranteed by the caller's OPENING
+   * reservation ({@link bulletKeepLines}): it committed `BULLET_KEEP_LINES` lines
+   * to the page the block starts on, and a block reaching this branch has
+   * `total >= 2 * BULLET_KEEP_LINES`, so the lines before the tail are at least
+   * that many. A block whose own height exceeds a page simply breaks more than
+   * once; a fresh page always holds a tail this short, so the reservation here
+   * can never be unsatisfiable.
+   *
+   * `followKeep` (#632) is height that must land on the SAME page as this block's
+   * FINAL line — the next bullet's keep-opening. The tail reservation is the one
+   * that covers that final line for a DIVISIBLE block, so it is also the one that
+   * must carry the follow: either both fit here, or the break moves up and the
+   * tail travels forward together with what follows it. An indivisible block
+   * never reaches this branch — its opening reservation covers its final line, so
+   * {@link drawBullet} folds the follow in there instead. Bounded by construction:
+   * `BULLET_KEEP_LINES` plus at most `2 * BULLET_KEEP_LINES - 1` follow lines,
+   * far under a page, so this stays satisfiable.
+   */
+  private ensureWrappedLine(
+    index: number,
+    total: number,
+    lineHeight: number,
+    opts: { widowControl?: boolean; followKeepHeight?: number },
+  ) {
+    const startsTail =
+      (opts.widowControl ?? false) &&
+      total >= 2 * BULLET_KEEP_LINES &&
+      index === total - BULLET_KEEP_LINES;
+    this.ensure(
+      startsTail
+        ? BULLET_KEEP_LINES * lineHeight + (opts.followKeepHeight ?? 0)
+        : lineHeight,
+    );
   }
 
   advance(points: number) {
@@ -564,41 +747,28 @@ class Layout {
   }
 
   /**
-   * Draw a wrapped block of text. `x` is the left edge; `hangingIndent`
-   * indents continuation lines (for bullet hanging indent). `atomicSegments`
-   * opts into segment-atomic middot wrapping (see `wrap()` above) — leave it
-   * unset/`false` for ordinary header/entry lines; the skills entry is the
-   * only caller that sets it `true` (#307).  Returns nothing; mutates the
-   * cursor and paginates as needed.
+   * Resolve everything a {@link drawText} call needs BEFORE it touches the page:
+   * the font, the sanitized value, the flush-right tail, and the wrapped lines.
+   * Shared with {@link measureTextHeight} so a keep-with-next reservation (#629)
+   * is measured by exactly the wrapping that will draw it — one layout pass'
+   * worth of logic, called once per measure and once per draw, never forked into
+   * a second implementation that could disagree.
    */
-  drawText(
+  private resolveDrawLines(
     text: string,
-    opts: {
-      bold?: boolean;
-      size?: number;
-      color?: RGB;
-      x?: number;
-      hangingIndent?: number;
-      uppercase?: boolean;
-      atomicSegments?: boolean;
-      /** A short tail (a role/degree date range) drawn FLUSH-RIGHT, regular
-       *  weight, on the first wrapped line's baseline (#425). */
-      rightText?: string;
-      rightColor?: RGB;
-      rightSize?: number;
-      /** Register a clickable URI annotation over the whole first line (#425). */
-      linkUrl?: string;
-      /** Register a clickable URI annotation over each `display` substring found
-       *  in the first line (#425 — the contact line's link slugs). Applied only
-       *  when the text fits on ONE line, so measured offsets are accurate. */
-      linkSpans?: Array<{ display: string; href: string }>;
-    } = {},
-  ) {
+    opts: DrawTextOpts,
+  ): {
+    lines: string[];
+    font: PdfFont;
+    size: number;
+    x: number;
+    lineHeight: number;
+    rValue: string;
+    rSize: number;
+  } {
     const size = opts.size ?? SIZE_BODY;
     const font = opts.bold ? this.fonts.bold : this.fonts.regular;
-    const color = opts.color ?? this.black;
     const x = opts.x ?? MARGIN;
-    const hanging = opts.hangingIndent ?? 0;
     // Sanitize LAST — after the case transform — so a case-expansion can never
     // produce an un-encodable glyph downstream. `toUpperCase()` maps some
     // WinAnsi-native lowercase letters to glyphs with NO WinAnsi representation
@@ -633,10 +803,36 @@ class Layout {
       atomic,
       rightReserve,
     );
-    const lineHeight = size * LINE_GAP;
+    return { lines, font, size, x, lineHeight: size * LINE_GAP, rValue, rSize };
+  }
+
+  /**
+   * Height (pt) a {@link drawText} call with the SAME `text`/`opts` will
+   * occupy — its wrapped line count times its line height. Pure: measures
+   * without touching the page or the cursor. The keep-with-next reservations
+   * (#629) are built from this, so a reservation can never be a guess.
+   */
+  measureTextHeight(text: string, opts: DrawTextOpts = {}): number {
+    const { lines, lineHeight } = this.resolveDrawLines(text, opts);
+    return lines.length * lineHeight;
+  }
+
+  /**
+   * Draw a wrapped block of text. `x` is the left edge; `hangingIndent`
+   * indents continuation lines (for bullet hanging indent). `atomicSegments`
+   * opts into segment-atomic middot wrapping (see `wrap()` above) — leave it
+   * unset/`false` for ordinary header/entry lines; the skills entry is the
+   * only caller that sets it `true` (#307).  Returns nothing; mutates the
+   * cursor and paginates as needed.
+   */
+  drawText(text: string, opts: DrawTextOpts = {}) {
+    const { lines, font, size, x, lineHeight, rValue, rSize } =
+      this.resolveDrawLines(text, opts);
+    const color = opts.color ?? this.black;
+    const hanging = opts.hangingIndent ?? 0;
     const singleLine = lines.length === 1;
     for (let i = 0; i < lines.length; i++) {
-      this.ensure(lineHeight);
+      this.ensureWrappedLine(i, lines.length, lineHeight, opts);
       const lineX = i === 0 ? x : x + hanging;
       const topY = this.y;
       this.page.drawText(lines[i], {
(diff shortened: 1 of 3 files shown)


src/lib/pdf/render-ats-pdf.ts
```diff
@@ -633,7 +694,33 @@
...
+
+  /**
+   * Draw a wrapped block of text. `x` is the left edge; `hangingIndent`
+   * indents continuation lines (for bullet hanging indent). `atomicSegments`
+   * opts into segment-atomic middot wrapping (see `wrap()` above) — leave it
+   * unset/`false` for ordinary header/entry lines; the skills entry is the
+   * only caller that sets it `true` (#307).  Returns nothing; mutates the
+   * cursor and paginates as needed.
+   */
+  drawText(text: string, opts: DrawTextOpts = {}) {
```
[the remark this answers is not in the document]
s-annam: Fixed in `cd4bae3` — `drawText` is now below the threshold.

The complexity was concentrated in the `if (i === 0)` block, which carried three independent first-line decorations: the flush-right date tail, the whole-line link annotation, and the per-substring link spans. Those are extracted to a private `decorateFirstLine`, leaving `drawText` as a plain line loop.

The extraction moves **only drawing**. All the geometry it needs (`font`, `size`, `lineX`, `topY`, `rValue`/`rSize`) is resolved by `resolveDrawLines` and passed in, and it neither paginates nor touches the cursor — so it cannot drift measurement from drawing, which is the invariant the #629 keep-with-next reservations depend on. Proven rather than asserted: the corpus glyph-run dump across all 54 fixtures is byte-identical across the refactor.

Note the diff also *raised* this to 29 before lowering it — #631/#632 added `widowControl` and `followKeepHeight` to the line loop. Both now enter through a single `opts` pass-through to `ensureWrappedLine` rather than as positional args.

src/lib/pdf/render-ats-pdf.ts
```diff
@@ -721,85 +851,119 @@
...
+    return lines;
+  }
+
+  /**
+   * Draw a sequence of `{ text, bold }` runs with word-level wrapping, breaking
+   * to a new page between lines as needed. Line breaking itself lives in
+   * {@link wrapRuns}; this walks the resulting lines and draws them. Bold is
+   * preserved across wraps because it is tracked per chunk, not per line.
+   */
+  private drawRuns(
```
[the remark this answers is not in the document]
s-annam: Fixed in `cd4bae3` — `drawRuns` is now below the threshold.

Here the complexity was the loop nest: lines → words → chunks, with the per-chunk font switch that preserves bold across a wrap. The inner two levels are extracted to a private `drawRunWords`, so `drawRuns` walks lines and `drawRunWords` walks one line's words at a fixed baseline.

Line breaking itself stays wholly in `wrapRuns` — the extraction touches only the drawing half, and `drawRunWords` never paginates or moves the cursor. That keeps the #629 invariant that a reservation is measured by the exact wrapping that draws it. Verified by corpus diff: the glyph-run dump across all 54 fixtures is byte-identical across the refactor.

As with `drawText`, #631/#632 first pushed this to 23 by adding `widowControl` / `followKeepHeight` to the line loop; those now ride a single `opts` pass-through to `ensureWrappedLine`.

Vaishnavi1709 (approved):
Approving. Reviewer's summary matches the code: `BULLET_KEEP_LINES = 2` +
`bulletKeepLines(total < 4) === total` + `ensureWrappedLine`'s tail reservation
at `index === total - 2` together implement "≥ 2 lines on each side" precisely —
2/3-line bullets move whole, 4+ line bullets can still split but never leaves a
lone line at either edge.

**Guard-proof spot-check.** Disabled just the widow-tail branch
(`render-ats-pdf.ts:657-660`) — exactly 2 tests fail, both targeted at that rule:
"splits a FOUR-line bullet so neither page gets a lone line" and "keeps the
last bullet with the TAIL of a wrapped second-to-last bullet". The other 69
pagination tests stay green, including the 3-line widow, entry-header,
section-heading, and lone-trailing-bullet rules. So none of the four rules is
silently propping up another, and this one is not vacuous. Didn't spot-check
the other three but the test-file discipline reads consistent.

**Gates.** No fixtures, no feature-component churn, no skill/script files; lint
and typecheck clean; `fallow` down 4 → 2 complexity findings (both survivors
pre-existing on `main`). 311 pdf tests pass on `cd4bae3`.

Nits already in the PR body (`BODY_HEADER_KEEP_LINES=1`-equivalent,
`EMPHASIS_OPEN` PUA sentinel now load-bearing for layout) are the right ones to
leave as follow-ups.

src/lib/pdf/render-ats-pdf.ts
```diff
@@ -884,14 +1250,26 @@ export async function renderAtsResumePdf(
 
   // ── Summary ──
   if (model.summary) {
-    drawSectionHeading(layout, model.summaryHeading ?? "Summary");
+    // The summary body is plain wrapped text, so the heading must keep one BODY
+    // line with it (#629).
+    drawSectionHeading(layout, model.summaryHeading ?? "Summary", SIZE_BODY * LINE_GAP);
```
rohithgollapalli, line 1255: **Nit — the Summary's body text is the one wrapped block with no widow control.**

The heading reserves one body line here, which satisfies AC3. But the body itself goes out through the plain per-line path:

```ts
```

`ensureWrappedLine` only widens its reservation when `opts.widowControl` is set, and `drawBullet` is the only caller that sets it. So a summary that wraps to N lines with N-1 fitting can leave its last line alone at a page top — the shape #631 exists to prevent, in the one block type #631 does not cover.

**This is out of scope for all three issues** — #629/#631/#632 scope to bullets, headers and headings — and the module docblock is careful not to claim otherwise, so this is not a correctness gap against anything you set out to fix. Raising it only because the mechanism is now sitting right there and the call site is one option away:

```diff
-  layout.drawText(model.summary, { size: SIZE_BODY });
+  layout.drawText(model.summary, { size: SIZE_BODY, widowControl: true });
```

Worth a follow-up issue rather than a change in this PR, since it would need the same corpus glyph-diff you ran for the other three and there is no reason to spend that here.

Same reasoning applies to the capped body-header (skills) path: lines past `BODY_HEADER_KEEP_LINES` flow through the plain `ensure`, so a skills list can widow its last line too. That one I would leave alone deliberately — divisibility is the whole point of the cap.

rohithgollapalli (commented):
> **Posting as a comment, not an approval:** this merged while the review was running. The head SHA is unchanged (`cd4bae3`), so everything below still describes exactly what landed — but the verdict is moot, and the one item worth carrying forward is the summary widow-control follow-up.

Approving. This is the strongest-evidenced PR I have reviewed in this repo, and the part that matters most is the part the body is most honest about: **the independent adversarial round for #631/#632 never completed**, and the PR says so plainly instead of implying a clean review. So I ran that pass myself rather than reviewing the prose.

## Independent verification (not a re-read of the body's claims)

I built a sweep harness with no shared code with `export-layout-contract.test.ts` — different entry shapes, different assertions, different violation detectors — and ran **7 entry shapes x 33 filler values = 231 renders** against this branch, then swapped in `origin/main`'s `render-ats-pdf.ts` (keeping the branch's `extractPdfDrawnLines`, which is new here) and ran the identical sweep.

| Shape | AC1 violations on `main` | #632 violations on `main` | On this branch |
|---|---|---|---|
| 2 short bullets | filler 45 | filler 44 | none / none |
| 2 bullets, first wraps ~5 lines (divisible) | filler 45 | filler 40 | none / none |
| 2 bullets, first wraps ~3 lines (indivisible) | filler 45 | filler 42 | none / none |
| 3 short bullets | filler 45 | filler 43 | none / none |
| 4 short bullets | filler 45 | filler 42 | none / none |
| 4 bullets, last wraps ~3 lines | filler 45 | filler 42 | none / none |
| 2 bullets + subLine | filler 45 | filler 43 | none / none |

**14 of 14 violation classes reproduce on `main` and are zero here.** That is the non-vacuity proof the missing round would have been asked for, and it covers the two shapes the `followKeepHeight` hand-off is most likely to get wrong — the divisible first bullet (follow rides the tail reservation) and the indivisible one (follow folded into `entryKeepHeight`).

I also enumerated the hand-off arithmetic independently and reached the same conclusion as the body: the opening reservation carries the follow only when `bulletKeepLines(total) === total`, and `ensureWrappedLine`'s tail fires only when `total >= 2 * BULLET_KEEP_LINES`. Disjoint, so the follow is reserved exactly once on every path. The one uncovered case (`withFollow > USABLE_PAGE_HEIGHT`) is the deliberate #629-outranks-#632 degradation, correctly documented at the return.

Separately confirmed:

- **`resolveDrawLines` is genuinely pure** — no page mutation, no `registerLink` (those live only in the draw loop). So measure-then-draw cannot double-register a link annotation, which is what makes the reservations exact rather than a second estimate.
- **`toWinAnsi` is untouched by this diff** and present on `main`, so its fallow complexity finding is genuinely inherited, not introduced. fallow reports `2 above threshold` + `6 inherited findings excluded`; report-only per `CLAUDE.md` either way.
- `npm run typecheck` clean, `npm run lint` clean, `npx vitest run src/lib/pdf/` **311 passed / 24 files**, `export-layout-contract.test.ts` **71 passed**.

No Blocking and no Secondary findings. Nits only, two of them inline.

## Nits

- **Summary body text has no widow control** (inline at `render-ats-pdf.ts:1255`). Out of scope for all three issues — they scope to bullets, headers and headings — and the module docblock correctly does not claim otherwise. Raised only because the mechanism now exists and the call site is one option away.
- **`BODY_HEADER_KEEP_LINES = 2` is unpinned by any test** (inline at `:1383`) — your own nit, confirmed: `1` and `2` produce identical output on every measured case, so nothing would fail if it changed.
- **`EMPHASIS_OPEN` (U+E000) is now load-bearing for a layout decision**, not just styling — your nit, and I agree with the framing that it is a pre-existing hazard of the sentinel design rather than something this PR introduces.
- **Prettier**: I checked before repeating your nit — `render-ats-pdf.ts` is prettier-dirty on `origin/main` too, so this is a pre-existing repo-wide state, not a regression here. There is no prettier gate in `verify`. I would drop this from the known-nits list rather than carry it.
- **Provenance honesty.** Three `unreported` rows and an explicit "this section does not claim a clean independent review" is the right call, and rarer than it should be. Worth keeping as the house pattern.

## Gates

- **Fixture PII (3a)** — N/A. No fixture binary added or changed; the diff is 3 source files. `check:fixtures` scope untouched.
- **Design-system (3b) / style tokens (3c)** — N/A. `src/lib/pdf/` only, no feature component. The `rgb()` values here are PDF graphics-state, correctly called out in the module docblock as not Tailwind tokens.
- **Skill/script command bugs (3e)** — N/A. Nothing under `.claude/skills/` or `scripts/`.
- **Dead-code / fallow (3d)** — clean; see above. The +2 clone groups are test scaffolding, and `main` already carries 2 groups of the identical shape, so declining to collapse 14 proven test bodies for an advisory finding is the right trade.

Verdict rule applied: **0 Blocking** — this would have been APPROVE had it still been open. Reviewed at `cd4bae3`.

---

src/lib/pdf/render-ats-pdf.ts
```diff
@@ -928,34 +1306,241 @@ export async function renderAtsResumePdf(
...
+
+/**
+ * How many header lines a BODY-TEXT header line contributes to a keep-with-next
+ * reservation, at most (#629). Two, because that is the smallest unit with any
+ * widow value — it is the same orphan rule `drawBullet` applies to a wrapped
+ * bullet: the block's first line is never left alone at a page bottom. One would
+ * add nothing over the plain per-line `ensure`; three or more starts trading real
+ * page density for a guarantee body text does not need, since it is divisible.
+ */
+const BODY_HEADER_KEEP_LINES = 2;
```
rohithgollapalli, line 1383: **Nit — this constant is unpinned by any test.**

Confirming your own nit from the review notes: `1` and `2` produce identical output on every measured case, which means nothing in the suite would fail if this changed. The value is a semantic choice ("a body-text header keeps at most two lines with the heading"), and the reasoning in `entryHeadHeight`'s docblock is what actually carries it.

Not worth a test that asserts a tautology. But since the `bullets.length === 0` half of that gate *is* load-bearing and *is* the subtle part, a one-line comment here pointing at `entryHeadHeight` for the gate rationale would stop a future reader from tuning this constant in isolation and assuming the cap is unconditional.

Non-blocking.

src/lib/pdf/render-ats-pdf.ts
s-annam: Confirmed and fixed — #635, PR #636.

Reproduced before touching it. Rendering a synthetic model with a summary of `words` filler tokens and counting drawn summary lines per page on `main` (`7596161`):

| words | summary lines | per page (before) | per page (after) |
|---|---|---|---|
| 596 | 49 | `[49]` | `[49]` |
| **602** | **50** | **`[49, 1]`** ← widow | `[48, 2]` |
| **608** | **50** | **`[49, 1]`** ← widow | `[48, 2]` |
| 614 | 51 | `[49, 2]` | `[49, 2]` |

Your one-option diff is exactly right, and it is the whole fix.

**One thing your read added that I want to record.** You framed this as needing the same corpus glyph-diff spend as the other three, which is why you filed it as a follow-up rather than a change. Measuring the reachability inverted that: the summary always starts at the top of page one, so it can only straddle a break by outrunning a whole page **on its own** — ~50 wrapped lines / ~600 words. No corpus fixture is within an order of magnitude, so `corpus-roundtrip` and `corpus-edit-roundtrip` come back byte-identical. The verification was cheap *because* the case is remote. Worth doing anyway for the reason you did not have when you wrote this: #625 made the summary a free-text field the user edits and pastes into, and "nobody writes a 600-word summary" is a claim about authorship, not about what the field accepts.

**And the orphan half needs nothing** — which I flagged as a possible gap before measuring, and it is not one. `drawSectionHeading` reserves the heading plus one body line, and this block can only ever *begin* at a page top, so its opening can never sit at a page bottom. That asymmetry is now stated in `ensureWrappedLine`'s docblock, which explained the far side of the break purely via `bulletKeepLines` — a path the summary does not take.

**Skills path left alone, as you recommended.** Divisibility past `BODY_HEADER_KEEP_LINES` is the point of the cap; covering it would re-introduce the ~100pt density cost #630 measured.

Guarded by a test that bisects to the spill boundary rather than hard-coding a word count, brackets both ends so it cannot go vacuous, and asserts the same `MIN_BULLET_LINES_PER_PAGE` a wrapped bullet must meet. Proven non-vacuous: reverting the option fails exactly this test — `words=556: summary widowed [49,1]: expected 1 to be greater than or equal to 2` — and nothing else.

src/lib/pdf/render-ats-pdf.ts
s-annam: Agreed on all three points, and done that way in #636.

No test added — an assertion over a constant whose two candidate values render identically is a tautology, and this repo has enough real guards to not want a fake one.

The comment you asked for is there, and it says what you identified as the actual hazard: the number is not the load-bearing part, the **gate** is.

```
 * NOT pinned by a test, and deliberately so: `1` and `2` produce identical output
 * on every case the suite measures, so any assertion here would be a tautology.
 * The load-bearing part of this rule is not the number but the GATE that applies
 * it — see {@link entryHeadHeight}, where the cap is conditional on the model
 * declaring the header to be body text. Tune the number here and you change
 * nothing measurable; move the gate there and you change what a real header
 * reserves. Read that docblock before touching either.
```

The last sentence is the one doing the work — a future reader arriving at this constant with a density complaint now gets routed to `entryHeadHeight` before they conclude the cap is unconditional.

s-annam:
Disposition of the post-merge review, now that #630 has landed as `7596161`.

@rohithgollapalli — the independent sweep is the round the PR body admitted it never got. 7 shapes × 33 filler values, no shared code with `export-layout-contract.test.ts`, 14 of 14 violation classes reproducing on `main` and zero here. That is a stronger non-vacuity proof than the one the PR shipped with, and it arrived after the merge rather than before it, which is worth being honest about: the merge was not gated on it.

@Vaishnavi1709 — the targeted guard-proof (disabling only the widow-tail branch at `render-ats-pdf.ts:657-660`, confirming exactly 2 tests fail and the other 69 stay green) is the check that matters most for a four-rule change, because it is the one that catches rules silently propping each other up.

**Carried forward — both inline nits, in #635 / PR #636:**

- **Summary body widow control.** Confirmed by reproduction, fixed, guarded. Reachability turned out to be ~50 wrapped lines / ~600 words, since the summary always starts at the top of page one — which made the corpus verification cheap rather than expensive, inverting the reason it was deferred. Details in the thread.
- **`BODY_HEADER_KEEP_LINES` comment** pointing at `entryHeadHeight`'s gate. No test, for the reason given.

**Not carried forward, and why:**

- **Prettier.** Agreed — dropping it. You checked before repeating it and found `render-ats-pdf.ts` is prettier-dirty on `origin/main` too, with no prettier gate in `verify`. A pre-existing repo-wide state is not a PR nit. If it should be enforced it wants its own issue and a one-time sweep, not a line in a known-nits list.
- **`EMPHASIS_OPEN` (U+E000) now load-bearing for layout.** Real, and correctly framed as a hazard of the sentinel design rather than something #630 introduced. Left open deliberately — a fix here means changing how emphasis is represented through the wrap path, which is a design change, not a nit.

Thanks to both of you. The reviews caught more than the PR's own adversarial round did.

src/lib/pdf/render-ats-pdf.ts
s-annam: Landed in `1855ff2` (#636), together with the rest of the sweep. Nothing changed since the reply above — noting the final sha and resolving.

src/lib/pdf/render-ats-pdf.ts
s-annam: Landed in `1855ff2` (#636). Comment is as quoted above; resolving.

s-annam:
Both post-merge nit threads here are addressed in #636 and resolved — the Summary widow control is that PR's core change, and the `BODY_HEADER_KEEP_LINES` comment is in it. Nothing outstanding on this PR.

Issue #581, opened 2026-07-25
jobs: make the query legible — primary-title marker + click-to-promote, level provenance, honest launcher preview


s-annam:
## Problem

Titles and skills in the job query do **six different jobs** and render as one flat,
undifferentiated chip list. Three specific things the UI currently hides:

1. **Only `titles[0]` reaches the keyless feeds.** `searchPhrase(query)` sends the first title to
   Remotive + Arbeitnow. The other titles affect ranking only. Nothing on screen says which chip is
   the one being searched, and there is no way to change it short of deleting chips.
2. **Target level is derived but unattributed.** `query.seniority` comes from a keyword in one of
   the résumé's own titles (`deriveSeniorityAcrossTitles`). The user sees `Executive` with no idea
   which title produced it — and therefore no basis to decide whether to override it.
3. **The `/` launcher preview silently truncates.** `FindJobsLauncher.tsx:47` renders
   `[...query.titles, ...query.skills].slice(0, PREVIEW_LIMIT)` with `PREVIEW_LIMIT = 6`. A résumé
   with 4 titles and 11 skills shows 4 titles + **2 of 11** skills, with no `+9`, no ellipsis, no
   expand. ui-ux-pro-max Content/Truncation (Medium): *"Truncate with ellipsis and expand option —
   don't overflow or cut off."*

Two things the launcher preview is **not** — both were suspected and both are false, so don't
"fix" them:

- It is **not a second source of truth.** The launcher calls the same `buildJobQuery` with the same
  seeds; `/jobs/` recomputes an identical query on arrival.
- **Nothing is combined or deduped.** A title like `Engineering Lead` and a skill like
  `Engineering Leadership` land adjacent purely from concatenation order and read as one merged
  term. No dedup runs across titles and skills, and none should — they feed different machinery.

### A. Primary title: marker + click-to-promote

`ChipListEditor` is **shared** (`Titles`, `Skills`, `Exclude` all use it) — do **not** bake
promotion into it. Extend it with an opt-in `onPromote?` + `primaryIndex?` pair; the Skills and
Exclude call sites simply don't pass them. (One primitive per concern; a titles-only wrapper is the
fallback if the props read badly.)

```
Titles                        ⓘ first title is searched
★ VP Engineering        ✕     ← sent to job feeds
  Berlin Site Lead      ✕     (click to make primary)
  Engineering Manager   ✕
  Sr. Software Engineer ✕

Searching feeds for "VP Engineering"
```

[...]

PR #633, opened 2026-07-28
fix(score,edit): close three scorer tokenization gaps, the Structure mislabel, and two edit-surface gaps (#622 #623 #624 #625 #626 #627)

### LINKED ISSUE #622: [score] ACTION_VERBS omits strong verbs the rewrite eval already accepts (shipped, owned, secured, won, ran)
## Summary

The scorer's `ACTION_VERBS` set (`src/lib/score/score.ts:175`) is missing common strong past-tense résumé verbs — `shipped`, `owned`, `won`, `ran`, and others. A bullet leading with one loses half its Structure credit despite being exactly the shape the dimension is meant to reward.

The strongest evidence is internal, not editorial: **three of the missing verbs are already whitelisted by our own rewrite eval.**

## The contradiction

`src/lib/webllm/eval/verbs.ts` builds the eval rubric's verb set as `SCORER_ACTION_VERBS ∪ EVAL_ONLY_EXTENSIONS`, and that extension list contains:

```ts
"engineered", "investigated", "prototyped", "rewrote", "shipped",
...
"completed", "conducted", "drafted", "identified", "owned",
"performed", "planned", "presented", "produced", "published",
"secured", "tracked",
```

So `shipped`, `owned`, and `secured` **pass** the rubric that grades the on-device rewriter's output, and **fail** the scorer that grades the user's résumé. The WebLLM rewrite lane can therefore hand a user a bullet its own eval calls verb-led, which then costs them Structure points when rescored. Two sets, one intent, opposite verdicts.

The extension's stated rationale doesn't cover this. It justifies itself as *tense breadth* (present-progressive forms users never write) and *cross-discipline IC verbs*. But `shipped` / `owned` / `secured` are past-tense and squarely eng/PM — precisely the register `ACTION_VERBS` claims to cover. They are in the extension because they're normal résumé verbs, which is the argument for putting them in the base set.

`won` and `ran` are in neither set.

## Not a design choice

`ACTION_VERBS`' docblock (`score.ts:167-174`) is explicit about what is deliberately excluded:

> Kept narrow on purpose: weak generic verbs ("worked", "helped", "supported", "responsible", "assisted", "participated") are deliberately NOT here. A bullet leading with one of those SHOULD fail the specificity check.

That carve-out is about **weak** verbs. `Shipped`, `Owned`, `Won`, `Ran` are not weak — they are among the strongest outcome verbs a senior IC or manager writes. Their absence is an omission in a hand-curated 45-item list, not a judgement.

## Observed impact

Dogfooding a real two-page résumé: Structure `17/30`, with `13/23` bullets counted verb-leading. Four flagged bullets lead with:

- `Shipped and open-sourced two tools with public users …`
- `Won new AI/ML platform charters for the site …`
- `Ran performance reviews and promotion calibration …`
- `Owned <product>'s entire user-facing application surface …`

All four are verb-led accomplishment bullets. Structure = `(0.5 × verbCount + 0.5 × lengthCount) / total × 30`, so each costs 0.5 credit. On that résumé, verbCount `13 → 17` moves Structure `17 → 20`.

## Ask

1. Promote the past-tense, general-register verbs from `EVAL_ONLY_EXTENSIONS` into `ACTION_VERBS`: `shipped`, `owned`, `secured`, `deployed`, `engineered`, `rewrote`, `authored`, `analyzed`, `conducted`, `identified`, `presented`, `produced`, `published`, `planned`. Leave the present-progressive forms in the extension — users don't write them, and the reason they exist there (small models emit them) is genuine.
2. Add the verbs in neither set: `won`, `ran`, `grew`, `founded`, `hired`, `partnered`, `defined`, `cut`, `ported`, `rebuilt`, `advised`, `shaped`, `standardized`, `instrumented`.
3. Do **not** add the documented weak verbs. `worked`, `helped`, `supported`, `responsible`, `assisted`, `participated`, `contributed`, `served` should keep failing — that carve-out is correct and load-bearing.
4. `EVAL_ONLY_EXTENSIONS` should shrink to only what its docblock actually justifies: tense breadth and genuinely discipline-specific verbs. If a verb belongs in both, it belongs in the base set.

## Acceptance criteria

- [ ] A bullet leading with `Shipped` / `Owned` / `Won` / `Ran` scores verb-led.
- [ ] A bullet leading with `Worked` / `Helped` / `Participated` / `Responsible for` still fails — assert this explicitly so the widening doesn't drift into accepting everything.
- [ ] No verb appears in both `ACTION_VERBS` and `EVAL_ONLY_EXTENSIONS`. A test asserts the sets are disjoint, so the two can't silently disagree again.
- [ ] `EVAL_ONLY_EXTENSIONS`' docblock still accurately describes its remaining contents.
- [ ] Corpus score snapshots are re-baselined in the same commit, with the delta stated in the PR body — this moves Structure on most fixtures by design.

## Notes

- `ATS_SCORE_ALGO_VERSION` must bump. Scores change for every résumé with an affected verb, and a user comparing against a previously-seen number needs the version to explain it.
- Widening a curated list is a one-way ratchet in perceived generosity. Worth pairing with a test that pins a handful of *rejected* leads, so the boundary stays a real boundary.

### LINKED ISSUE #623: [score] bulletHasMetric misses spelled-out cardinals — "two tools", "four-month" score as unquantified
## Summary

`bulletHasMetric` (`src/lib/score/score.ts:157`) detects quantification only through **digit characters**. A bullet that quantifies in words — "two tools", "a four-month engagement", "three sites" — is scored as carrying no metric, which costs Specificity directly.

## Root cause

```ts
function bulletHasMetric(text: string): boolean {
  if (STRONG_METRIC_PATTERNS.some((p) => p.test(text))) return true;
  const stripped = text.replace(YEAR_TOKEN, "");
  return ANY_DIGIT.test(stripped);
}
```

`STRONG_METRIC_PATTERNS` are all digit-anchored (`%`, `$`, `K/M/B` suffix, `Nx`). The fallback is `ANY_DIGIT` after stripping 4-digit years. There is no path by which a cardinal word registers.

Specificity is `min(1, metricRatio / 0.6) × 40` — a pure ratio over the bullet pool — so every false negative costs `40 / (0.6 × totalBullets)` points. On a 23-bullet résumé that is ~2.9 points per missed bullet.

## Observed impact

From dogfooding a real two-page résumé, Specificity `29/40` at `10/23` bullets carrying a metric:

- `Shipped and open-sourced **two** tools with public users …`
- `Advisor to <company> … during a **four-month** engagement …`

Both quantify. Neither counts. `10/23 → 12/23` moves Specificity `29 → 35`.

## Ask

Extend `bulletHasMetric` to recognise spelled-out cardinals in a quantifying position.

**Recognise:** `two|three|four|…|twelve`, plus `dozen`, `hundred`, `thousand`, `million`, `billion`, and hyphenated compound modifiers (`four-month`, `three-year`, `six-week`).

**Do not recognise:** `one` and `a`/`an`. "one team", "a product" is not quantification — treating it as such would let nearly every bullet pass and hollow out the dimension. `one` is also heavily used non-numerically ("one of the site's senior voices", "no one", "one-off").

**Require a quantifying position** — the cardinal should modify a noun, not stand alone. `two tools`, `three sites`, `four-month engagement` count; `two` in "one or two" or a sentence-initial "Three years ago" is weaker. A simple `\b(two|three|…)\b[\s-]\w` adjacency check covers the realistic cases without a parser; do not over-engineer this.

**Keep the year guard.** `YEAR_TOKEN` stripping exists so "From 2013 to 2021 …" doesn't read as quantified; the word path needs an equivalent so "two thousand nineteen" doesn't slip through as a metric.

## Acceptance criteria

- [ ] `Shipped two internal tools` → `hasMetric: true`.
- [ ] `Led a four-month migration` → `hasMetric: true`.
- [ ] `Owned one product surface` → `hasMetric: false`.
- [ ] `One of the senior engineering voices on the team` → `hasMetric: false`.
- [ ] `Worked on the platform` → `hasMetric: false`.
- [ ] Existing digit-path behaviour is byte-identical — all current `score.test.ts` metric cases pass unchanged.
- [ ] Corpus score snapshots re-baselined in the same commit, delta stated in the PR body.

## Notes

- `ATS_SCORE_ALGO_VERSION` must bump.
- Sibling of the `ACTION_VERBS` gap — both are the scorer under-crediting a bullet that is genuinely doing the thing the dimension rewards. Worth landing together so the corpus re-baseline happens once.
- The counter-argument is real and worth stating in the PR: digits scan better than words on a résumé, and "two tools" arguably *should* be "2 tools". But the scorer's job is to measure whether the bullet quantifies, not to enforce numeral style. If we want to push numerals, that belongs in the recommendation copy (`recommendation.ts`), not as a silent scoring penalty.

### LINKED ISSUE #624: [score] Structure hint labels a half-credit sum as a count of bullets within 8-30 words
## Summary

The Structure dimension's hint line reads *"13/23 bullets within 8–30 words"*, but the numerator is **not a count of bullets in that window**. It is the rounded sum of the dimension's half-credits — half for a verb-led bullet, half for an in-window bullet. The label states a fact about the number that isn't true of it.

Second, related defect: Structure is **50% action-verb driven**, and the hint never mentions verbs. Half the dimension is invisible to the user, so a résumé losing points purely on verbs gets pointed at its word counts.

## Root cause

`src/components/features/AtsScoreReadout.tsx:90`:

```ts
const structureHint = `${score.structure.goodBullets}/${score.structure.totalBullets} bullets within 8–30 words`;
```

`goodBullets` is populated at `src/lib/score/score.ts:926`:

```ts
goodBullets: Math.round(pool.goodStructure),
```

and `goodStructure` accumulates half-credits (`score.ts:274-280`):

```ts
let bulletScore = 0;
if (startsWithActionVerb(b)) bulletScore += 0.5;
const wc = b.split(/\s+/).filter(Boolean).length;
if (wc >= BULLET_LENGTH_MIN_WORDS && wc <= BULLET_LENGTH_MAX_WORDS) {
  bulletScore += 0.5;
}
goodStructure += bulletScore;
```

So `goodBullets = round(0.5 × verbCount + 0.5 × lengthCount)`. It coincides with the in-window count only when `verbCount === lengthCount`, which is why the mislabel is easy to miss in practice.

## Failure scenario

A résumé with 23 bullets, **all** within 8–30 words, **none** verb-led:

- `goodStructure = 0.5 × 0 + 0.5 × 23 = 11.5` → `goodBullets = 12`
- Hint renders: **"12/23 bullets within 8–30 words"**
- Truth: **23/23** are within 8–30 words. The real problem is 0/23 verb-led, which the UI never says.

The user shortens bullets that were never too long, and the number doesn't move.

Observed on a real résumé at `17/30` Structure: the hint read `13/23`, which was the half-credit sum (13 verb-led + 13 in-window → 13.0), not the 13 in-window bullets it claimed to describe. That the two happened to be equal is coincidence.

## Ask

1. Surface the two sub-counts separately in `AnonymousAtsScore.structure` — `verbLedBullets` and `inWindowBullets` — instead of only the fused `goodBullets`. Keep `goodBullets` if something else consumes it, but stop labelling it as a bullet count.
2. Rewrite the hint to name both halves, e.g. `13/23 bullets verb-led · 13/23 within 8–30 words`. Space is tight in the dimension card, so a two-line hint or the shorter `verb-led 13/23 · length 13/23` is fine — the requirement is that both halves are visible.
3. Check `recommendation.ts` for the same conflation. If the "tighten each bullet" recommendation fires off a fused number, it can point a user at length when verbs are the actual loss.

## Acceptance criteria

- [ ] The Structure hint reports the in-window count and the verb-led count as two distinct numbers, each matching a direct count over the bullet pool.
- [ ] A fixture with all bullets in-window and none verb-led renders `23/23` for length and `0/23` for verbs — this is the regression test for the mislabel.
- [ ] No displayed count is ever a rounded half-credit sum.
- [ ] The Structure *score* is unchanged — this is a reporting fix, not a scoring change, so `ATS_SCORE_ALGO_VERSION` does **not** bump and corpus score snapshots do not move.
- [ ] Hint text is reachable by screen readers and doesn't rely on colour alone (existing `AtsScoreReadout` conventions).

## Notes

- Scoped deliberately narrow: reporting only. The question of whether Structure *should* be half verbs is separate and not raised here.
- Cheapest of the three scorer issues and the only one with no score movement, so it can land independently of the `ACTION_VERBS` and spelled-out-cardinal fixes.

### LINKED ISSUE #625: [edit] Summary is parsed, scored, and exported but never displayed or editable

### LINKED ISSUE #626: [edit] no way to delete a bullet outside the WebGPU-gated rewrite panel; cleared bullets leave a ghost row
## Summary

There is no way to delete a bullet from the inline edit surface. Clearing its text leaves a ghost row rendering the `"empty bullet"` placeholder, which persists for the rest of the session.

`removeBullet` **already exists** and is already threaded into the component that renders bullets — it is just never exposed there. Its only consumer is the LLM section-rewrite panel, which is WebGPU-gated. On hardware without WebGPU there is no path to delete a bullet at all.

## Evidence

**The capability exists.**

- `useEditableParse.ts:607` — `const removeBullet = useCallback((index: number) => { … })`, exported on the hook's public API (`:378`, `:1092`)
- `removedBullets` is a first-class override threaded through `applyOverrides` and into the score recompute (`useAnalyzedResume.ts`)
- `ReconstructedResume.tsx:1295` — `onRemoveBullet={(index) => removeBullet(index)}`
- `ReconstructedRole.tsx:494` — the prop arrives

**It is only wired to the rewrite panel.** `ReconstructedRole.tsx:530-540`:

```ts
const rewriteApply = useMemo<SectionRewriteApply | undefined>(() => {
  if (!onBulletChange || !onAddBullet || !onRemoveBullet) return undefined;
  return {
    obsIndices,
    onReplace: (index, text) => onBulletChange(index, text),
    onRemove: (index) => onRemoveBullet(index),
    onAdd: (text) => onAddBullet(text),
    captureUndo,
  };
}, [...]);
```

`rewriteApply` goes to `useSectionRewrite` → `SectionRewrite`. That is the **only** reference to `onRemoveBullet` in the file.

**The inline bullet row has no remove control.** `ResumeBulletRow` (`ReconstructedRole.tsx:207-280`) renders an `EditableField` and `BulletFlagsInline`, and nothing else. Compare its siblings, which all have one:

| Entity | Control | Site |
|---|---|---|
| Role | `<RemoveButton label="Remove role" …>` | `ReconstructedRole.tsx:559` |
| Education entry | `<RemoveButton label="Remove education" …>` | `ReconstructedEducationSkills.tsx:172` |
| Skill | `removeSkill` / `removeCategorySkill` | `useEditableParse` |
| **Bullet** | **none** | — |

`RemoveButton` is already imported into `ReconstructedRole.tsx` (`:34`) for the role case, so the primitive is in scope.

**Clearing leaves a ghost.** `ReconstructedRole.tsx:253-256`:

```tsx
<EditableField
  value={displayText || undefined}
  placeholder="empty bullet"
  emptyAffordance="plain"
```

An emptied bullet renders as a `•` and the literal placeholder text `empty bullet`, indefinitely.

**The export disagrees with the UI.** `resolveBullets` (`ats-resume-model.ts:343-354`) ends with `.filter(Boolean)`, so an emptied bullet **is** dropped from the downloaded PDF. So the preview shows a row the artifact doesn't have — the inverse of the Summary defect (#see sibling), and the same broken contract: the reconstructed résumé is supposed to be what you're about to download.

## Ask

1. Add a per-bullet `RemoveButton` to `ResumeBulletRow`, wired to the existing `onRemoveBullet`. The primitive, the handler, the override, and the undo capture all already exist — this is exposure, not new machinery.
2. Route it through `captureBulletUndo` the way the rewrite path does, so a mis-click is recoverable.
3. Decide the empty-bullet case explicitly rather than leaving it implicit. Preferred: on commit of an empty value, drop the bullet (call `removeBullet`) instead of storing `""`. That makes the UI agree with `resolveBullets`' existing `.filter(Boolean)` and removes the ghost row without a second concept. If the placeholder row is kept deliberately for a just-added blank bullet, say so in the docblock — right now the two behaviours are indistinguishable to the user.
4. Do not gate any of this on WebGPU.

## Acceptance criteria

- [ ] Every bullet in the inline edit surface has a remove control, reachable without WebGPU.
- [ ] Removing a bullet removes it from the downloaded PDF and from the score pool.
- [ ] Removing a bullet is undoable through the existing undo path.
- [ ] Clearing a bullet's text does not leave a persistent `empty bullet` row.
- [ ] The reconstructed résumé and the exported PDF contain the same set of bullets — assert this directly, since the two currently disagree.
- [ ] Removing the last bullet of a role leaves the role intact with its add-affordance, not a collapsed or broken entry.
- [ ] Existing `SectionRewrite` remove-proposal behaviour is unchanged.

## Notes

- Touch surface is small, but `ReconstructedRole.tsx` is 588 LOC and named in `CLAUDE.md` as known debt. Prefer extracting `ResumeBulletRow` into a sibling over growing the file.
- The 44×44 touch-target trap from #581/#591 applies: a remove control per bullet row sits in a dense vertical list, so oversized hit areas will bleed into the adjacent row's control. WCAG 2.2 AA is **24×24**, not 44.
- Related: an emptied bullet also currently fails all three quality checks while visible, so the flag badges on a ghost row read as real findings.

### LINKED ISSUE #627: [score] a spaced em-dash counts as a word against the 8-30 bullet window
## Summary

Bullet word count is `text.split(/\s+/).filter(Boolean).length`, so a **spaced em-dash tokenizes as its own word**. `"a — b"` counts as 3 words. An em-dash-heavy bullet spends 2–4 of its 30-word Structure budget on punctuation, and the UI shows the inflated number with no way to tell what it's made of.

## Root cause

Two call sites, same expression:

- `score.ts:233` — `const wordCount = text.split(/\s+/).filter(Boolean).length;` (per-bullet observation)
- `score.ts:276` — `const wc = b.split(/\s+/).filter(Boolean).length;` (`scoreBulletPool`)

`filter(Boolean)` removes empty strings from the split, not non-word tokens. Any whitespace-delimited run of punctuation survives as a "word": `—`, `–`, `-`, `·`, `|`, `/`, `→`.

The `BULLET_LENGTH_MIN_WORDS = 8` / `MAX = 30` window (`:107-108`) is then applied to that inflated count, and `wellFormedLength` feeds half the Structure dimension.

## Observed impact

Dogfooding a real two-page résumé written in an em-dash-heavy register. One bullet displayed as `51w`; roughly 47 are words and 4 are em-dash tokens. Across the résumé, `10/23` bullets fell outside the 8–30 window — several of them borderline, where 2–4 phantom words is the difference.

The user-facing number is the flag badge (`lengthToken` / `lengthTitle`, `ReconstructedRole.tsx:189-196`), so they see `51w`, trust it, and cannot tell that 4 of those are dashes.

## Why this is the scorer's problem, not the writer's

A separator glyph is not a word by any reading — not by a recruiter's skim, not by an ATS's token count, not by any standard word-count implementation the user has seen elsewhere. Someone comparing our `51w` against their word processor's count will get a different number and conclude our count is wrong. It is.

The 8–30 window is a proxy for *"is this bullet a scannable single idea"*, and punctuation contributes nothing to that.

## Ask

Count only tokens containing at least one alphanumeric character.

```ts
const words = text.split(/\s+/).filter((t) => /[\p{L}\p{N}]/u.test(t));
```

- Extract this into a single shared `countWords(text)` helper. Both call sites must use it, or the per-bullet badge and the aggregate score can disagree about the same bullet — the same drift the `isLoneDateRange` docblock warns about for its two call sites.
- Keep hyphenated and slashed compounds as one word each: `four-month`, `AI/ML`, `US-India` each contain letters and stay a single token under the rule above. Do **not** split on `-` or `/`.
- A bare `37+` or `1.9` contains digits and correctly still counts.

## Acceptance criteria

- [ ] `"Owned the surface — a 50-engineer org — across 3 sites"` counts 10 words, not 12.
- [ ] `"four-month"`, `"AI/ML"`, `"US-India"` each count as exactly 1 word.
- [ ] `"37+"` and `"1.9"` each count as 1 word.
- [ ] A bullet of only punctuation counts 0 words.
- [ ] The per-bullet badge (`Nw`) and `scoreBulletPool`'s `wc` return the same number for the same bullet — asserted directly, not by inspection.
- [ ] Corpus score snapshots re-baselined in the same commit, delta stated in the PR body.

## Notes

- `ATS_SCORE_ALGO_VERSION` must bump. This moves `wellFormedLength` for any résumé using spaced dashes, which is common in senior/exec register.
- Direction of the change is generous — bullets get shorter, more land in-window, Structure goes up. Pair with the other two scorer widenings (#622, #623) so the corpus re-baseline happens once rather than three times.
- Third in the family of scorer tokenization gaps, all found in one dogfooding pass: verb-set omissions (#622), digit-only metric detection (#623), and this. Common shape: the scorer measures a proxy for the thing it claims to measure, and the proxy leaks.

s-annam:
Six issues from one dogfooding pass: three scorer tokenization gaps where the proxy leaked, one reporting mislabel, and two edit-surface gaps where a field was parsed and exported but not shown.

Landed as one commit so the corpus re-baseline and the `ATS_SCORE_ALGO_VERSION` bump happen once rather than three times.

## What shipped

- **Closes #622** — `ACTION_VERBS` widened 45 → 74. Promoted the past-tense, general-register verbs out of `EVAL_ONLY_EXTENSIONS` (`shipped`, `owned`, `secured`, …) and added the verbs that were in neither set (`won`, `ran`, `grew`, `founded`, …). The rewrite eval and the scorer no longer disagree about the same verb. A new `src/lib/webllm/eval/verbs.test.ts` asserts the two sets are **disjoint**, so they can't silently drift apart again, and pins the documented weak verbs (`worked`, `helped`, `participated`, `responsible`) as still-rejected — the carve-out is load-bearing and stays.
- **Closes #623** — `bulletHasMetric` now recognises spelled-out cardinals in a *quantifying position* (`two tools`, `four-month engagement`). `one` and `a`/`an` are excluded by design; a spelled-out-year guard keeps `"joined in two thousand nineteen"` from reading as quantified while `"processed two thousand records per day"` still does.
- **Closes #627** — extracted a shared exported `countWords(text)` counting only tokens containing `[\p{L}\p{N}]`, so a spaced em-dash no longer costs a bullet 2–4 words of its 30-word budget. Wired to **four** call sites, not the two the issue named — see Notes.
- **Closes #624** — Structure hint now reports both halves as direct counts: `verb-led 13/23 · length 13/23`. Previously it rendered the rounded half-credit sum `goodBullets` under a label claiming it was a count of in-window bullets. Reporting-only: no score moves, no version bump from this issue.
- **Closes #625** — Summary is now displayed and editable. New `ReconstructedSummary.tsx` sibling, new `summaryOverride` on `useEditableParse` (single-value pattern, mirroring `skillsOverride`), threaded through `apply-overrides`, `useAnalyzedResume`, and `useJdFitResume`. Honours `summaryHeading`; clearing it drops heading **and** body from the export; an accepted LLM rewrite and a manual edit write to the same override in both orders.
- **Closes #626** — every bullet now has a remove control, reachable without WebGPU (previously the only delete path was the WebGPU-gated rewrite panel). Routed through `captureBulletUndo`. Committing an empty bullet now deletes it rather than leaving a ghost `empty bullet` row, which makes the UI agree with `resolveBullets`' existing `.filter(Boolean)`.

## Score impact — every user's number changes

`ATS_SCORE_ALGO_VERSION` **1.5 → 1.6** (bumped once for the batch).

All 54 corpus fixtures re-baselined. Movement is generous by design — the three widenings each credit bullets that were already doing the thing the dimension rewards:

| Movement | Fixtures |
|---|---|
| `overall` changed | 32 |
| `goodBullets` (Structure, #622) | 22 |
| `metricBullets` (Specificity, #623) | 20 |
| `algoVersion` only | 54 |

**No fixture regressed** — verified programmatically across all 54 diffs: zero score decreases, zero changes to `totalBullets`/`bulletCount`. Largest mover: `synthetic-degreeless-two-programs` overall 62 → 73.

## Verification

`typecheck` 0 · `eslint .` 0 · `vitest run` **3507 passed** / 10 skipped · `check:fixtures` 0 (54 PDFs, all personas synthetic). No new fixture binaries added.

`fallow audit` exits 1 (report-only per `CLAUDE.md`): 6 complexity findings, 26 clone groups. The largest clone group (99 lines, `ReconstructedResume.tsx` ↔ `ReconstructedSkills.tsx`) was verified **pre-existing on `origin/main`**, not introduced here.

## Notes for the reviewer

- **#627 touches a third call site the issue didn't name.** `extractBulletsFromLines` gates `ANON_BULLET_MIN_WORDS` on `countWords`, which changes which lines enter the bullet **pool**, not just how long they measure. Inert on the corpus — no fixture's `bulletCount`/`totalBullets` moved — but it is a behaviour change beyond the issue's literal scope, flagged here rather than discovered later.
- **#623 widened slightly past the issue's literal spec**: plural scale nouns (`millions of`, `dozens of`) are accepted, which the issue's list (`dozen|hundred|thousand|million|billion`) didn't cover. Validated empirically — every corpus bullet scoring `hasMetric` with no digit was inspected: 27 hits, all genuine quantifications, zero false positives, and no fixture tripped `Six Sigma` / `Big Four` / a spelled year.
- **#626 extended remove+undo to the "Other bullets" (unmatched) bucket**, which the issue's evidence didn't call for. Kept deliberately — those bullets are already editable, so `AC1` ("every bullet in the inline edit surface") covers them. This was also the source of the batch's most serious defect; see below.

### Known nits, not fixed here

Left deliberately so the fixes stayed scoped to defects:

- Five test helpers still use the pre-#627 `split(/\s+/).filter(Boolean).length` to synthesize `wordCount` (`apply-overrides.test.ts:21`, `undo-batch.test.ts:213`, `ReconstructedResume.test.ts:23`, `ats-resume-model.test.ts:21`, `group-bullets.test.ts:36`). **Inert** — nothing reads the field and the stub texts have no punctuation-only tokens — but it's the expression #627 exists to delete.
- `ReconstructedSummary.tsx` adds a **third** verbatim copy of a local 6-line `SectionHeading`. Belongs in `@design-system/shared`.
- `applyOverrides` is now **17 positional params**, with helpers padding 10+ literal `{}`/`undefined`. The next override channel should force an options-object refactor.
- `ResumeBulletRow.tsx` is 300 LOC vs the ~200 guideline. It's extracted-not-authored (`ReconstructedRole.tsx` went 595 → 425, and ~200 of the 300 moved verbatim), so this is debt relocated honestly rather than created — but it's two files over the guideline instead of one.
- The new Structure hint drops the "8–30 words" explanation, so `length 13/23` no longer states its criterion. #624 sanctioned the short form; a `title` attribute would cost nothing.

## Provenance

Each row is self-reported by the agent that ran. Rows marked `unreported` are agents that completed their work but went idle without returning a report; the requested alias is recorded rather than a guess, because the alias does not establish which model served it.

| Stage | Model | Effort |
|---|---|---|
| Implementation — #622 | Claude Sonnet 5 | high |
| Implementation — #623 | unreported (requested: opus) | ultra |
| Implementation — #627 | unreported (requested: sonnet) | high |
| Implementation — #624 | unreported (requested: sonnet) | high |
| Implementation — #625 | unreported (requested: opus) | ultra |
| Implementation — #626 | Claude Sonnet 4.6 | high |
| Adversarial review | Claude Opus 5 (1M context) | high |
| Review fixes | unreported (requested: opus) | high |

## Adversarial review

One round, run pre-PR against the accumulated working tree. Round 1 returned `clean: false` with **3 blocking findings**; all three were fixed and the round converged (blocking empty). Nits were reported, not iterated — they are listed above under "Known nits".

The reviewer reproduced each finding end-to-end with scratch tests rather than reasoning from the code, and deleted them afterwards.

### Blocking findings and their fixes

**1. Removing the last unmatched bullet left the Undo strip with nothing to render.**
`groupBulletsByExperience` appends the `experienceIndex === null` group only when it is non-empty. Removing the Other bucket's last bullet therefore unmounted the `RoleEntry` that owned `removeStatus` *in the same render that performed the removal*, so #626's AC3 ("removing a bullet is undoable") failed on the code path this batch newly introduced. Parsed roles are immune — `sliceGroups` falls back to an empty group — which is why it did not surface anywhere else.

Fixed by lifting the strip into a new `BulletRemoveStatus.tsx` (`useBulletRemoveStatus`): `ExperienceSection` hosts one instance for the Other bucket, parsed roles keep hosting their own. The alternative fix — reverting the Other-bullets wiring — was rejected because those bullets are already editable, so dropping their remove control would have traded this defect for a partial failure of AC1.

Guarded by `src/components/features/ExperienceSection.test.tsx` (3 tests), including one asserting a parsed role still hosts its own strip so the lift is not over-applied. **Proven non-vacuous**: stubbing `{otherRemove.strip}` to `null` fails 2 of 3; restoring returns 3/3.

**2. A WCAG conformance claim the code did not implement.**
`ResumeBulletRow.tsx` asserted a "24×24 minimum touch target (WCAG 2.2 AA)" while `RemoveButton` rendered a 14×14 box — a 10×10 `CloseIcon` plus `p-0.5`, with no `min-h`/`min-w` on the `Button` base. The control likely still conformed via SC 2.5.8's *spacing* exception (row pitch ≈28px), but the comment claimed conformance by size in a public repo.

Fixed by making the code true — `min-h-6 min-w-6` on `RemoveButton` — which also corrects the pre-existing role and education remove controls, rather than by weakening the comment.

**3. #624's named regression test did not exist at the layer that had the bug.**
The AC calls for a fixture that *renders* `0/23` verbs and `23/23` length. Only the score object was asserted; `AtsScoreReadout.test.tsx` had gained the two new fields purely to satisfy the type. The rendering itself was correct — the reviewer verified the component emits `verb-led 0/23 · length 23/23` — so this was a test gap, not a behaviour bug, on the one issue in the batch whose entire subject is a rendering mislabel.

Fixed by asserting the rendered string in `AtsScoreReadout.test.tsx`.

### Leads that were refuted

Three suspicions raised going in did not survive scrutiny, recorded so they are not re-litigated:

- **Stale word-count helpers are inert.** `applyOverrides` reads only `.text` and `.index`; no assertion reads `wordCount`, and the stub texts contain no punctuation-only tokens, so the old and new expressions return the same number. Downgraded to a nit.
- **The 99-line clone group predates this batch.** Confirmed hunk-by-hunk that `ReconstructedResume.tsx:242-340` is untouched by this change and byte-identical to `origin/main`; `ReconstructedSkills.tsx` was never modified here.
- **#623's exclusion of `nineteen` is the spec, not a judgement call** — the issue caps the accepted set at `twelve`, and thirteen–nineteen appear only as year-guard material.

### Verification performed by the review

No fixture score decreased anywhere (all 54 diffs parsed programmatically); `ATS_SCORE_ALGO_VERSION` bumped exactly once; every hook dep array in the diff hand-audited in both directions (`exhaustive-deps` is not enforced in this repo, so lint proves nothing about them); #626's empty-commit test proven non-vacuous by deleting the branch it guards; #625's override-ordering tested in both directions; no raw `<button>`, hex, palette class, or manual `dark:` in any new or changed feature file; no PII in new source files.

**Not covered:** real-browser layout (the touch target is computed from markup, not measured), and the WebGPU rewrite lane end-to-end (no test drives an on-device model).

---

Closes #622
Closes #623
Closes #624
Closes #625
Closes #626
Closes #627


src/lib/score/score.ts
```diff
@@ -154,10 +187,73 @@ const REDACTED_DATE_RE = new RegExp(
...
+  "|ten|eleven|twelve|thirteen|fourteen|fifteen|sixteen|seventeen|eighteen|nineteen" +
+  "|one|two|three|four|five|six|seven|eight|nine";
+
+/** Word-path equivalent of YEAR_TOKEN. Only the two prefixes that would
+ *  otherwise trip a cardinal above ("**two thousand** nineteen", "**nineteen
+ *  hundred** twelve") need guarding — "twenty twenty-four" contains no accepted
+ *  cardinal. The trailing year word is REQUIRED, so a genuine quantity like
+ *  "processed two thousand records" survives the strip and still counts. */
+const WORD_YEAR_TOKEN = new RegExp(
+  `\\b(?:two\\s+thousand|nineteen\\s+hundred)(?:\\s+and)?\\s+(?:${YEAR_TAIL_WORD})\\b`,
```
rohithgollapalli, line 242: **Secondary — the spelled-year guard and the cardinal matcher disagree about what separates words, so hyphenated spelled years read as quantified.**

`CARDINAL_SEPARATOR` accepts whitespace *or* a hyphen/dash:

```ts
const CARDINAL_SEPARATOR = "[\\s\\u2010-\\u2015\\-]";
```

but this guard joins its parts with `\s+` only. So a spelled year written with hyphens is never stripped, and `QUANTIFYING_CARDINAL` then matches the surviving `two-t` / `nineteen-h`. Reproduced against the real scorer via `computeAtsScore` (same harness shape as `bulletRegistersAsMetric`):

| Bullet | `hasMetric` | Should be |
|---|---|---|
| `Joined the company in two thousand nineteen and stayed` | false | false |
| `Joined the company in two-thousand-nineteen and stayed` | **true** | false |
| `Joined the company in two thousand-nineteen and stayed` | **true** | false |
| `Hired in nineteen-hundred-ninety-eight before the merger` | **true** | false |

The last one is the one that stings: `Hired in nineteen hundred ninety-eight` is in your own `notQuantified` list and passes, but the same date with hyphenated thousands does not — so the test set brackets the bug without crossing it.

Fix is to give the guard the same separator class the matcher uses:

```diff
+const YEAR_SEPARATOR = `${CARDINAL_SEPARATOR}+`;
 const WORD_YEAR_TOKEN = new RegExp(
-  `\\b(?:two\\s+thousand|nineteen\\s+hundred)(?:\\s+and)?\\s+(?:${YEAR_TAIL_WORD})\\b`,
+  `\\b(?:two${YEAR_SEPARATOR}thousand|nineteen${YEAR_SEPARATOR}hundred)` +
+    `(?:${YEAR_SEPARATOR}and)?${YEAR_SEPARATOR}(?:${YEAR_TAIL_WORD})\\b`,
   "gi",
 );
```

I ran the proposed expression against the full accept/reject set before suggesting it — all four false positives close, and every genuine quantification survives, including the two designed to be adversarial to this exact strip:

| Bullet | Current | Proposed | Want |
|---|---|---|---|
| `two thousand nineteen` (year) | false | false | false |
| `two-thousand-nineteen` (year) | true | **false** | false |
| `two thousand-nineteen` (year) | true | **false** | false |
| `nineteen-hundred-ninety-eight` (year) | true | **false** | false |
| `Processed two thousand records per day` | true | true | true |
| `Shipped two internal tools` | true | true | true |
| `Led a four-month migration` | true | true | true |
| `Owned twenty-two dashboards` | true | true | true |
| `Processed millions of records nightly` | true | true | true |

Not blocking — it over-credits a rare shape by half a Specificity point, no corpus fixture hits it, and nothing regresses. But it is an internal inconsistency between two constants written in the same commit, and the fix is one character class.

src/components/features/ReconstructedSummary.tsx
```diff
@@ -0,0 +1,180 @@
...
+import type { BulletOverrides } from "../../hooks/useEditableParse.ts";
+import type { SectionRewriteApply } from "./SectionRewrite.tsx";
+
+/** The stable section id the summary carries through the rewrite chain — the
+ *  key `buildResumeSections` mints and `summaryRewriteApply` is registered
+ *  under, so the proposal panel can join an outcome to its writer. */
+export const SUMMARY_SECTION_ID = "summary";
+
+/** Mirrors ReconstructedResume's local heading (and Education's copy of it). */
+function SectionHeading({ children }: { children: React.ReactNode }) {
```
rohithgollapalli, line 45: **Secondary — this is the FOURTH byte-identical copy, not the third.**

The known-nits list says "a **third** verbatim copy". There are four:

```
src/components/features/ReconstructedEducationSkills.tsx:45
src/components/features/ReconstructedResume.tsx:249
src/components/features/ReconstructedSkills.tsx:43
src/components/features/ReconstructedSummary.tsx:45   <- added here
```

I diffed all four — identical to the character, including the class string:

```tsx
function SectionHeading({ children }: { children: React.ReactNode }) {
  return (
    <h2 className="text-sm font-semibold uppercase tracking-wider text-content-muted">
      {children}
    </h2>
  );
}
```

`CLAUDE.md`'s Golden Rule is explicit that a parallel copy is the thing not to write, and this one is unusually cheap to fix: six lines, one `children` prop, no local state, already on semantic tokens. It is a clean `@design-system/shared` extraction plus four one-line import swaps, and it would delete 18 lines net.

**Why I am filing this Secondary and not Blocking**, so the verdict is auditable: the blocking bar in this repo's review gate is raw interactive primitives (`<button>`, modal, dropzone) and hardcoded colours. A heading is neither, the tokens are correct, and this PR *inherited* a triplication it did not create — adding the fourth copy is a consistency regression, not a new defect class. It should not hold the merge.

If you would rather not widen this PR's blast radius, the honest minimum is fixing the count in the known-nits list so the follow-up is scoped correctly.

rohithgollapalli (approved):
Approving. Six issues, one commit, one version bump — the batching rationale holds, and the score-movement accounting is the part I most wanted to check independently, so I did.

## Independent verification

I parsed all 54 `.expected.json` diffs against `origin/main` myself rather than trusting the summary table:

```
fixtures compared: 54
algoVersion old: {'1.5'} -> new: {'1.6'}
fixtures with an increase: 32 | unchanged: 22
DECREASES: none
bullet-count changes: none
```

That reproduces every claim exactly — 32 movers, zero regressions, `totalBullets`/`bulletCount` untouched across all 54, and the version bumped exactly once. For a change that moves every user's number, having that be checkable in one pass is the right shape.

Also confirmed first-hand:

- **Hook dep arrays hand-audited** (`exhaustive-deps` is not enforced here, so lint proves nothing). Both new arrays are correct in both directions: `useBulletRemoveStatus`'s `removeBullet` deps `[onRemoveBullet, captureUndo]` are exactly what it closes over, and `setSummaryField`'s `[]` closes over only a stable `useState` setter. The `[removed]` array is test scaffolding.
- **`useBulletRemoveStatus` living in `src/components/features/` is right, not a tier violation** — its stated model `useSectionRewrite` lives in `SectionRewrite.tsx`, so this matches the neighbours.
- **Decomposition direction is net positive**: `ReconstructedRole.tsx` 595 → **393** (better than the 425 the body claims), `ReconstructedResume.tsx` 1365 → 1391 (+26 of wiring, with the bulk correctly landing in a new sibling). That is `CLAUDE.md`'s "extract into a sibling rather than grow the file" rule being followed on a file the doc names as debt.
- `typecheck` clean, `eslint .` clean, `check:fixtures` green (54 PDFs, no new binaries), `npm run test` **3507 passed / 10 skipped** — matching the body exactly.

Two Secondary findings, both inline. No Blocking.

## Secondary (both inline)

1. **`score.ts:242` — #623's spelled-year guard uses `\s+` while the cardinal matcher accepts hyphens**, so hyphenated spelled years slip past it and read as quantified. Reproduced against the real scorer; fix validated against the full accept/reject set.
2. **`ReconstructedSummary.tsx:45` — this is the FOURTH byte-identical `SectionHeading`, not the third.** The known-nits list undercounts by one.

## Nits

- **Proper nouns embedding a cardinal read as quantifications** — `Advised Four Seasons on their loyalty platform` and `Advised Nine West on their retail platform` both register `hasMetric`, as does `Seven Wonders`. `NON_QUANTIFYING_PHRASE` covers only `six sigma` and `big four`. This is inherent to a keyword heuristic — the set is unbounded and not worth enumerating — and the cost is half a Specificity credit on one bullet, so I would leave it. Noting it because the corpus validation (27 hits, zero false positives) reads as stronger coverage than it is: the corpus simply contains no such employer.
- **`applyOverrides` is 17 positional params** — confirmed by count, and the disclosure is accurate. Agreed the next channel should force the options-object refactor; doing it here would have buried six fixes in a signature churn.
- **`ResumeBulletRow.tsx` at 300 LOC** — confirmed. "Debt relocated honestly rather than created" is the right characterisation given `ReconstructedRole` fell 202 lines.
- **The Structure hint drops its "8–30 words" criterion** — your nit. A `title` attribute really would cost nothing, and `length 13/23` is genuinely opaque without it.
- **Five stale pre-#627 word-count helpers in test files** — verified inert (nothing reads `wordCount`, no punctuation-only tokens in the stubs), so agreed this is a nit and not a correctness issue.

## Gates

- **Fixture PII (3a)** — pass. `check:fixtures` green on 54 PDFs; **no new fixture binary** (`--diff-filter=A` over `tests/fixtures/**` for pdf/png/jpeg/docx returns nothing). The 54 `.expected.json` edits are score re-baselines, which I verified field-by-field above.
- **Design-system (3b)** — pass on the blocking bar: no raw `<button>`/`<dialog>`/`<input>` in any changed feature file. The `SectionHeading` duplication is Secondary rather than Blocking under that bar — it is a heading, not an interactive primitive, and this PR inherited a pre-existing triplication rather than creating the pattern.
- **Style tokens (3c)** — clean. No hex, no raw palette class, no manual `dark:` in any changed `src/` file. (`text-content-muted` etc. throughout.)
- **Skill/script command bugs (3e)** — N/A.
- **Dead-code / fallow (3d)** — report-only per `CLAUDE.md`; the 99-line clone group's pre-existence on `origin/main` matches what I see.

Verdict rule applied: **0 Blocking, so APPROVE.** Reviewed at `ec95d13`.

---

src/components/features/ReconstructedRole.tsx
```diff
@@ -515,7 +287,18 @@ export function RoleEntry({
...
   onRemove,
+  removeControl,
 }: RoleEntryProps) {
+  // Per-bullet remove confirmation (#626), mirroring SectionRewrite's own
+  // applied/undone strip so a mis-click (or an empty-commit auto-remove, see
+  // ResumeBulletRow) is recoverable the same way a rewrite-review batch is.
+  // Owned by an ancestor for a group that can vanish on its last remove; owned
+  // here otherwise. The hook runs unconditionally either way (hooks rule); its
+  // result is simply unused when the ancestor supplied one.
+  const ownRemove = useBulletRemoveStatus(onRemoveBullet, captureUndo);
```
Vaishnavi1709, line 298: **Secondary — same class as the round-1 defect, one lifecycle level up.**

When `hostsStrip = true` (no `removeControl` passed — i.e. a user-added role), the undo strip lives on this `RoleEntry`. Trigger:

1. User clicks `+Add experience` → empty-header added role.
2. User clicks `+Add bullet`, types nothing, hits Save/Enter.
3. Empty commit routes through `handleCommit` → `onRemove()` → `ownRemove.removeBullet` sets `status: 'removed'` and drops the bullet.
4. Blur leaves the section → `sectionExitBlur`'s `setTimeout(onExit, 0)` fires → `pruneEmptyAddedEntries('experience')` sees `isAddedEntryEmpty` (empty header + zero bullets) → drops the entry → `RoleEntry` unmounts → `ownRemove.strip` unmounts with it.

The undo the round-1 fix was designed to preserve flashes for one frame and is gone. Fix mirrors the round-1 pattern: lift the strip host for added-role bullets the same way `ExperienceSection` already hosts the Other bucket's — or defer prune when a removed-strip is live.

src/lib/score/score.ts
```diff
@@ -263,19 +373,27 @@ function scoreBulletPool(bullets: string[]): {
...
+      verbLed: 0,
+      inWindow: 0,
       specificity: 0,
       structure: 0,
     };
   }
   let metric = 0;
   let goodStructure = 0;
+  let verbLed = 0;
+  let inWindow = 0;
```
Vaishnavi1709, line 385: **Nit (efficiency, non-blocking).** `scoreBulletPool` now walks the bullet array to count `verbLed`/`inWindow` (added by #624) — same booleans `analyzeBullets` computes one loop later at the call site (`score.ts:895-896`). `computeAnonymousAtsScore` calls `scoreBulletPool(bullets)` then `analyzeBullets(bullets)` back-to-back, so `bulletHasMetric` (now heavier via #623's regex + two `replace()` calls), `startsWithActionVerb`, and `countWords` all run twice per bullet. `verbLed` and `inWindow` are trivially `observations.filter(o => o.startsWithActionVerb).length` / `.filter(o => o.wellFormedLength).length`. The parallel-shape design is documented as deliberate for reuse (`score.ts:331-332`) — fold only if you're in the file anyway.

src/components/features/ReconstructedAdd.tsx
```diff
@@ -117,7 +126,7 @@ export function RemoveButton({
       variant="icon"
       aria-label={label}
       onClick={onClick}
-      className="shrink-0 text-content-muted hover:text-content-secondary"
+      className="min-h-6 min-w-6 shrink-0 text-content-muted hover:text-content-secondary"
```
Vaishnavi1709, line 129: **Nit (a11y altitude, non-blocking, track for follow-up).** The 24×24 WCAG 2.2 AA SC 2.5.8 floor was applied to `RemoveButton` (via `min-h-6 min-w-6` here) rather than raising `variant="icon"`'s own floor in the `Button` primitive. That fixes what this PR introduced and — transitively — every consumer that routes through `RemoveButton`. But 6 other `variant="icon"` call sites still have the 14×14 hitbox the round-1 finding flagged as non-conformant: `SectionRewrite.tsx:401`, `ReconstructedSkillControls.tsx` ×3 (lines 40/130/341), `RewriteReviewList.tsx:158`, and `ChipListEditor.tsx:45` (which explicitly documents the 5px overlay workaround). The primitive is the right altitude for an a11y floor; per-caller `min-h`/`min-w` will keep leaking. Out of scope for this PR — worth a follow-up issue.

Vaishnavi1709 (approved):
Approving. Substantial diff — 6 issues, 84 files, an `ATS_SCORE_ALGO_VERSION` bump — but the internals hold up under a 4-angle finder + verify sweep and the round-1 adversarial rounds already caught the load-bearing correctness bug (remove-strip lift). The corpus re-baseline story (32 fixtures moved, 0 regressed, `totalBullets`/`bulletCount` invariant preserved) is the right shape of evidence for a scorer widening.

**Gates.** `check:fixtures` green (54 PDFs, no binaries touched); design-system reuse green (all 3 new feature files import `@design-system`); style tokens clean; `typecheck` + `lint` clean. `fallow` reports 6 complexity findings (21 inherited excluded) — `computeAnonymousAtsScore` is now 42 cognitive / 31 cyclomatic vs 31/26 on `main`, worsened by #624's two new counters. Report-only per `CLAUDE.md`, but worth an eye for the next scoring touch.

## Secondary

**Noun swap on the summary rewrite panel is incomplete.** `RewriteReviewList.tsx:179-181` — the new `noun` prop lands on the `kindLabel` (`New/Removed/Edited ${noun}`) but the placeholder and aria label right below still say `"edit this bullet"` / `"Edit proposed bullet"` verbatim. Accepting a summary rewrite (a #625 flow) therefore reads "Edited summary" in the header but "edit this bullet" in the textarea and the screen-reader label. Fix: thread `noun` into the placeholder + label strings. *(Body-only — those lines aren't in a hunk.)*

**Same class as the round-1 defect, one lifecycle level up.** See inline comment on `ReconstructedRole.tsx:298`.

## Nits (non-blocking)

- **`line-primitives.ts:83-88` — `isProseLine` was missed by #627.** It still uses `.split(/\s+/).filter(Boolean).length`. Its own docblock (lines 72-73) explicitly pins its 8-word floor to the scorer's 8-30 window — the "silent drift" hazard the new `countWords` docblock warns against. The PR body's known-nits list covers 5 *test* helpers; this is the one production consumer that also uses that expression. *(Body-only — file is not touched by the diff.)*
- **`score.ts:385-398` — `scoreBulletPool` and `analyzeBullets` double-walk bullets.** See inline comment.
- **`ReconstructedAdd.tsx:126-129` — WCAG floor is at the caller, not the primitive.** See inline comment.

The 5 known nits already in the PR body (test helpers, `SectionHeading` dup, 17-param `applyOverrides`, `ResumeBulletRow` at 300 LOC, Structure hint drops "8–30 words") are the right ones to leave as follow-ups.

---
Reviewed by: Claude Opus 4.7 (1M context, high)

Samhit21 (approved):
**APPROVE** — batch is correct, gated, and re-baselined consistently. Independently re-verified below; no blocking issues.

I re-ran the offlinecv gates on `ec95d13`: `check:fixtures` green (54 PDFs synthetic, **no binaries touched** — pure `.expected.json` re-baseline), no raw `<button>`/hex/palette/`dark:` in the new feature code, `typecheck` 0, and the touched tests + `corpus.test.ts` (54) + `corpus-edit-roundtrip.test.ts` all pass, so the algo-1.6 re-baseline is internally consistent.

Correctness I checked directly: the `ACTION_VERBS`/`EVAL_ONLY_EXTENSIONS` sets are provably disjoint; the #623 cardinal path guards spelled years (`two thousand nineteen` → no metric) without eating genuine quantities (`two thousand records` → metric); both production `applyOverrides` callers thread `summaryOverride`; and the #626 blocking fix is sound — `otherRemove` is hosted by `ExperienceSection`, which outlives the "Other" group it drives, and `hostsStrip` prevents a double strip.

The disclosed nits (17-positional-param `applyOverrides`, third `SectionHeading` copy, `ResumeBulletRow` at 300 LOC, stale word-count test helpers, dropped "8–30 words" hint label) are all genuinely non-blocking and accurately characterized — no need to hold the merge for them.

One additive nit, non-blocking: `text-[11px]` in `ResumeBulletRow.tsx:156,195` is inline typography (relocated verbatim from `ReconstructedRole`, so pre-existing debt) — a shared size token would be tidier whenever that file is next touched.

---

src/lib/score/score.ts
s-annam: Fixed in `1855ff2` (#636), taking your diff as written — `YEAR_SEPARATOR = ${CARDINAL_SEPARATOR}+`, shared by both constants.

Reproduced first, and your table held exactly: three hyphenated years registered as metrics, `nineteen hundred ninety-eight` did not. Your point that the existing `notQuantified` set brackets the bug without crossing it is the part worth recording, so the docblock now says the shared separator is load-bearing rather than stylistic.

Six rows added, not three. The three rejects, plus `Processed two-thousand records per day` on the **accept** side — widening the separator means the guard must still require a year *tail* word, or the fix starts eating genuine hyphenated quantities. Proven non-vacuous: reverting the separator fails exactly the three new reject rows.

`ATS_SCORE_ALGO_VERSION` 1.6 → 1.7, which is what kept this out of the first batch. That cost is now measured rather than assumed — pinning the version back at 1.6 and re-running the corpus passes all 54 snapshots unchanged, so no fixture's parse or score moves and the only snapshot delta in the diff is the stamp.

src/components/features/ReconstructedSummary.tsx
s-annam: Fixed in `1855ff2` (#636). You were right on the count — extracted to `src/design-system/shared/SectionHeading.tsx`, exported through the `@design-system` barrel, all four call sites import it.

Took the full extraction rather than your stated minimum (fixing the count in the known-nits list), since #636 had no review yet and the change is byte-identical markup.

src/components/features/ReconstructedRole.tsx
s-annam: Deferred to #637, not fixed here.

Confirmed your repro reads correctly against the code: `handleCommit` → `onRemove()` → `pruneEmptyAddedEntries` on the `sectionExitBlur` tick unmounts the `RoleEntry` that hosts the strip. Left out of #636's sweep because the fix is a real design decision — lift the strip host, or defer prune while a removed-strip is live — not a mechanical change, and #636 is carrying a scoring change already. #637 has the repro and both directions.

src/lib/score/score.ts
s-annam: Not taken, and the reason is specific rather than a scheduling excuse.

Your condition was met — #636 *is* in `score.ts`, for the spelled-year separator fix. But the cheap fold does not exist. `scoreBulletPool` has two other callers (`score.ts:456`, `:474`) that pass raw `string[]`, so the second walk only goes away by changing the shared function's signature to take `BulletObservation[]` — which contradicts the "parallel shape is deliberate for reuse" docblock you cited and pulls in the authed scorer.

That is a scorer-core refactor riding along with a scoring change in the same PR, which is the combination worth not making. Leaving your thread open rather than resolving it, since this is a decline and not a fix.

src/components/features/ReconstructedAdd.tsx
s-annam: Filed as #638, as you asked.

Recorded there in full: the six remaining `variant="icon"` call sites, the `ChipListEditor.tsx:45` overlay workaround that makes a blanket `min-h` actively harmful, and that the target is 24×24 (SC 2.5.8) rather than 44 — 44 is what caused the adjacent-chip overlap on #581/#591.

s-annam:
Review threads swept in #636, which is where the follow-up work landed rather than a re-push here (that would have dismissed both approvals).

Fixed and resolved: the spelled-year separator mismatch (`score.ts:242`, with `ATS_SCORE_ALGO_VERSION` 1.6 → 1.7 — corpus proven unmoved) and the `SectionHeading` fourth-copy extraction. Already in #636 from the earlier batch: the five `wordCount` test helpers, the Structure hint criterion, and the `noun` prop reaching the editor.

Left open with replies rather than resolved: the `ReconstructedRole` undo-strip unmount (filed as #637), the `variant="icon"` touch-target floor (filed as #638), and the `scoreBulletPool`/`analyzeBullets` double-walk, which is declined with a reason on the thread. `isProseLine` is also still open — it is the one production consumer and needs a corpus run.

s-annam:
@Samhit21 — your `text-[11px]` nit was missed in my first sweep of these threads (I walked review *threads* only, and yours was body-only). Filed now as #640.

Measuring it made it larger than the two lines in `ResumeBulletRow`: font size is not tokenised anywhere in this repo — `@theme inline` maps `--color-*` only, there is no `tailwind.config.*` under Tailwind v4, and `src/` carries **67** arbitrary sizes (`text-[11px]` ×48, `text-[10px]` ×17, `text-[9px]` ×2). So fixing just those two would mint a token used by 2 of 48 consumers and leave 46 hand-styled. #640 scopes it as a sweep plus a lint rule so it cannot re-rot.

Also on this PR: @rohithgollapalli's recommendation on #630 to drop the prettier item from the known-nits list stands — `render-ats-pdf.ts` is prettier-dirty on `origin/main` too and there is no prettier gate in `verify`, so it was never a finding against that PR.

Issue #368, opened 2026-07-07
[parser] experience — LaTeX two-column layout splits multi-word cities mid-name


Vaishnavi1709:
## Category
parser (Steps 1 & 3 F&F QC — provenance category)

## Fixture
`tests/fixtures/pdfs/latex/multi-degree-coursework.pdf` (synthetic persona)

## Ground truth (`pdftotext`)
```
Greenfield Studios

New York, NY

Software Engineering Intern
May 2023 – Jun. 2023
```

Company on left column, `New York, NY` on right column, both separated by a blank line in the source.

## Observed
Reconstructed view (two of four experience entries):
```
Software Engineering Intern — Greenfield Studios New,     ← "New" glued onto company, trailing comma
York, NY                                                    ← orphan location start

Software Engineering Intern — Meridian Analytics New,     ← same shape on the next entry
York, NY
```

Single-word cities (`Bellevue, WA`, `McLean, VA`) survive on this same fixture. Only multi-word cities break.

## Localized layer
Line assembly in `src/lib/heuristics/pdf-extract.ts` / `src/lib/heuristics/sections.ts` — the reading-order merge is gluing "New" onto the preceding company token before "York" arrives. The result is a comma boundary in the wrong place, so downstream field extractors split the entry with the comma as a company/location boundary.

Filing without a specific line reference because the fix is likely upstream of the field extractors — either the item-to-line merge or the column reader itself.

## Repro
```
npm run dev
# drop tests/fixtures/pdfs/latex/multi-degree-coursework.pdf
# scroll to Experience entries 3 and 4
```

PR #634, opened 2026-07-28
fix(parser): recognise multi-word bare cities on middot experience headers (#616)

### LINKED ISSUE #616: [parser] experience — multi-word city without a state suffix merges into company and drops location
## Summary

**Category: parser, experience, location.** A `Company, City` segment where the city carries **no state/country suffix** is not split: the whole string lands in `company` and `location` drops to `undefined`. The identical `Company, City` string parses correctly when it sits in the **trailing** segment of the header, so the behaviour is position-dependent rather than string-dependent.

Failing input:

```markdown
### Engineering Lead · Globex, Mountain View · Fiber Group
*04/2021 – 12/2023*

- Built an 18-engineer org in under 6 months.
```

Parsed (OUTPUT):

```
title="Engineering Lead"
company="Globex, Mountain View"     ← should be "Globex"
team="Fiber Group"
location=undefined                  ← should be "Mountain View"
```

## Condition matrix

| # | Header | `company` | `location` |
|---|---|---|---|
| a | `Engineering Lead · Globex, Mountain View, CA · Fiber Group` | `Globex` ✅ | `Mountain View, CA` ✅ |
| b | `Engineering Lead · Globex, Mountain View · Fiber Group` | **`Globex, Mountain View`** ❌ | **`undefined`** ❌ |
| c | `Sr. Engineering Manager · Globex, Toronto · Enterprise Platforms` | `Globex` ✅ | `Toronto` ✅ |
| d | `Sr. Engineering Manager · Site Lead, Enterprise Platforms · Globex, Toronto` | `Globex` ✅ | `Toronto` ✅ |

- **a vs b** — dropping the `, CA` suffix breaks the split. The state abbreviation appears to be doing the work of identifying the segment tail as a location.
- **b vs c** — both are `Company, City` in the same middle position, but `Toronto` splits and `Mountain View` does not. The difference is single-token vs multi-word city, matching the multi-word-city failure mode described in #347.
- **b vs d** — the same `Company, City` shape parses correctly in a trailing segment.

So there are two interacting factors: a multi-word city without a state suffix, and the segment's position in the header.

## Why it matters

`Company, City` without a state or country is a normal way to write a header, especially for well-known cities and for non-US locations. The failure silently corrupts `company` — which is the field most likely to be matched against an employer list downstream — while producing `confidence=0.90` and `triggers=[]`.

## Relation to #347

#347 (closed) reported the multi-word-city split failure on the **two-column PDF `Date · Location` sub-line** path. This is the same city-recognition weakness surfacing on the **markdown-heading company segment** path, so the fix may share a helper. Filed separately because the failing input and the code path differ; close as a duplicate if the underlying location matcher turns out to be shared.

## Repro

`runCascadeFromMarkdown(rawText, markdown)`, where `rawText` is the markdown with `#`/`**`/leading `- `/wrapping `*` stripped. Read `result.canonical.fields.experience[0]`.

## Acceptance criteria

- [ ] Row b parses `company="Globex"`, `location="Mountain View"`
- [ ] Rows a, c, d unchanged
- [ ] A multi-word city without a state/country suffix splits the same way a single-word city does
- [ ] Splitting behaviour does not depend on the segment's position in the header
- [ ] Regression test covers multi-word city without suffix, in both middle and trailing segments

Vaishnavi1709:
## Summary

A `Title · Company, City · Team` middot header where the middle segment's city carried no state/country suffix (`Google, Mountain View · GFiber`) glued the whole `Company, City` into `company` and dropped `location`. Position-dependent-looking symptom, position-independent cause: Pass F of `stripLocationSuffix` (and its team-slot mirror `rescueTeamLocation`) both consult the closed `BARE_LOCATION_RE` vocab, and multi-word cities without a state suffix weren't in it — `Globex, Hyderabad` worked only because "hyderabad" happened to be listed.

Extends the single-source `MULTIWORD_US_CITY_ALT` const with three canonical Silicon-Valley/Peninsula tech-hub cities — `Mountain View`, `Palo Alto`, `Menlo Park`. Because the const flows into both `BARE_LOCATION_RE` and `KNOWN_MULTIWORD_US_CITY_RE` (the space-fold path #368 already relied on), the middot header case and the two-column fold case gain coverage in one edit. The `^…$`-anchored full-string match keeps a real company that merely contains one of these tokens (`Mountain View Software`) unaffected.

Closes #616

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — full heuristics suite (1117 tests) + PDF suite (296 tests) pass
- [x] New regression coverage in `experience.multiword-city.test.ts` asserts all four rows of the issue's condition matrix (a/b/c/d) plus a position-independence assertion (trailing `Company, MultiWordCity` splits the same way the middle segment does)
- [x] Manually verified in `npm run dev` — dropped a synthetic PDF holding all four header shapes; the reconstructed résumé now shows `company="Globex"` / `location="Mountain View"` for the failing row and for its trailing-position sibling; rows a/c/d unchanged

## Provenance

Code implementation via: Claude Opus 4.7 (medium)

### DIFF
--- a/src/lib/heuristics/extract/experience-disambiguate.ts
+++ b/src/lib/heuristics/extract/experience-disambiguate.ts
@@ -43,9 +43,19 @@ const LEGAL_SUFFIX_RE =
  *  (case-sensitive Title-case). Single source of truth so the two can't drift
  *  apart. Longest-first so "New York City" wins over "New York" in the embedded
  *  alternation (regex first-match); order is irrelevant for the `^…$`-anchored
- *  `BARE_LOCATION_RE`. */
+ *  `BARE_LOCATION_RE`.
+ *
+ *  Silicon-Valley/Peninsula additions (#616): "Mountain View", "Palo Alto",
+ *  "Menlo Park" are the multi-word tech-hub cities that show up co-located
+ *  with Google/Meta/Stanford in a `Title · Company, City · Team` middot header
+ *  with NO trailing state suffix. Without the vocab entry, Pass F of
+ *  `stripLocationSuffix` (bare-city tail check) failed to full-match them and
+ *  the middle segment stayed whole ("Google, Mountain View" → company). Same
+ *  closed-vocab discipline as the pre-existing entries: a real company that
+ *  merely contains one of these tokens ("Mountain View Software") still fails
+ *  the `^…$`-anchored full-string match. */
 const MULTIWORD_US_CITY_ALT =
-  "New York City|New York|New Orleans|San Francisco|San Diego|San Jose|San Antonio|Los Angeles|Las Vegas|Salt Lake City";
+  "New York City|New York|New Orleans|San Francisco|San Diego|San Jose|San Antonio|Los Angeles|Las Vegas|Salt Lake City|Mountain View|Palo Alto|Menlo Park";
 
 /** Bare city/region names (no "City, ST" state tail, so `US_LOCATION_RE` misses
  *  them) that show up as a `"Title, Location"` header tail — must NOT be cleaved
--- a/src/lib/heuristics/extract/experience.multiword-city.test.ts
+++ b/src/lib/heuristics/extract/experience.multiword-city.test.ts
@@ -84,3 +84,107 @@ describe("multi-word city on space-folded header (#368)", () => {
     expect(role.location).toBeUndefined();
   });
 });
+
+// Companion coverage to #368/#347: those covered the two-column space-fold
+// `Company City, ST` shape. #616 is the same "multi-word bare city" weakness on
+// the single-line MIDDOT header path: `Title · Company, City · Team` where the
+// city carries NO state/country suffix — the closed BARE_LOCATION_RE vocab (Pass
+// F of stripLocationSuffix) knows single-word "Hyderabad" but not multi-word
+// "Mountain View", so the middle segment stayed whole and `company` swallowed
+// the city. Fixed by adding Mountain View (a canonical tech-hub example) to the
+// single-source MULTIWORD_US_CITY_ALT vocab so BARE_LOCATION_RE full-matches it.
+// The trailing-segment case is covered too so behavior is position-independent.
+describe("multi-word city without state/country suffix on middot header (#616)", () => {
+  it("row (b) — middle segment: `Company, MultiWordCity` splits into company + location", () => {
+    // "Google, Mountain View" sits in the MIDDLE middot segment. Before the fix
+    // the whole string landed in `company` and `location` dropped.
+    const roles = roleFromSection([
+      { text: "EXPERIENCE", fontSize: 13 },
+      {
+        text: "Engineering Lead · Google, Mountain View · GFiber",
+        fontSize: 11,
+      },
+      { text: "04/2021 – 12/2023", fontSize: 11 },
+      { text: "• Built an 18-engineer org in under 6 months.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    const role = roles[0];
+
+    expect(role.title?.toLowerCase()).toContain("engineering lead");
+    expect(role.company).toBe("Google");
+    expect(role.location).toBe("Mountain View");
+    expect(role.team).toBe("GFiber");
+  });
+
+  it("row (a) unchanged — middle `Company, MultiWordCity, ST` still splits with state suffix", () => {
+    const roles = roleFromSection([
+      { text: "EXPERIENCE", fontSize: 13 },
+      {
+        text: "Engineering Lead · Google, Mountain View, CA · GFiber",
+        fontSize: 11,
+      },
+      { text: "04/2021 – 12/2023", fontSize: 11 },
+      { text: "• Built an 18-engineer org in under 6 months.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    const role = roles[0];
+
+    expect(role.company).toBe("Google");
+    expect(role.location).toBe("Mountain View, CA");
+    expect(role.team).toBe("GFiber");
+  });
+
+  it("row (c) unchanged — middle `Company, SingleWordCity` in existing bare-location vocab still splits", () => {
+    const roles = roleFromSection([
+      { text: "EXPERIENCE", fontSize: 13 },
+      {
+        text: "Sr. Engineering Manager · Google, Hyderabad · Enterprise Platforms",
+        fontSize: 11,
+      },
+      { text: "04/2021 – 12/2023", fontSize: 11 },
+      { text: "• Ran a cross-team migration to a shared platform.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    const role = roles[0];
+
+    expect(role.company).toBe("Google");
+    expect(role.location).toBe("Hyderabad");
+    expect(role.team).toBe("Enterprise Platforms");
+  });
+
+  it("row (d) unchanged — trailing `Company, SingleWordCity` still splits", () => {
+    const roles = roleFromSection([
+      { text: "EXPERIENCE", fontSize: 13 },
+      {
+        text: "Sr. Engineering Manager · Site Lead, Enterprise Platforms · Google, Hyderabad",
+        fontSize: 11,
+      },
+      { text: "04/2021 – 12/2023", fontSize: 11 },
+      { text: "• Ran a cross-team migration to a shared platform.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    const role = roles[0];
+
+    expect(role.company).toBe("Google");
+    expect(role.location).toBe("Hyderabad");
+  });
+
+  it("position-independence — trailing `Company, MultiWordCity` splits the same way as middle", () => {
+    // The acceptance criteria's core structural requirement: splitting behaviour
+    // does not depend on the segment's position in the header.
+    const roles = roleFromSection([
+      { text: "EXPERIENCE", fontSize: 13 },
+      {
+        text: "Sr. Engineering Manager · Site Lead, Enterprise Platforms · Google, Mountain View",
+        fontSize: 11,
+      },
+      { text: "04/2021 – 12/2023", fontSize: 11 },
+      { text: "• Ran a cross-team migration to a shared platform.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    const role = roles[0];
+
+    expect(role.company).toBe("Google");
+    expect(role.location).toBe("Mountain View");
+  });
+});


src/lib/heuristics/extract/experience-disambiguate.ts
```diff
@@ -43,9 +43,19 @@ const LEGAL_SUFFIX_RE =
...
+ *  with Google/Meta/Stanford in a `Title · Company, City · Team` middot header
+ *  with NO trailing state suffix. Without the vocab entry, Pass F of
+ *  `stripLocationSuffix` (bare-city tail check) failed to full-match them and
+ *  the middle segment stayed whole ("Google, Mountain View" → company). Same
+ *  closed-vocab discipline as the pre-existing entries: a real company that
+ *  merely contains one of these tokens ("Mountain View Software") still fails
+ *  the `^…$`-anchored full-string match. */
 const MULTIWORD_US_CITY_ALT =
-  "New York City|New York|New Orleans|San Francisco|San Diego|San Jose|San Antonio|Los Angeles|Las Vegas|Salt Lake City";
+  "New York City|New York|New Orleans|San Francisco|San Diego|San Jose|San Antonio|Los Angeles|Las Vegas|Salt Lake City|Mountain View|Palo Alto|Menlo Park";
```
s-annam, line 58: **Secondary — the defect class survives for every multi-word city outside the three added.**

Same synthetic header shape, run against this branch:

```
Engineering Lead · Nvidia, Santa Clara · Compute   → company="Nvidia, Santa Clara", location=undefined
Engineering Lead · Meta, Redwood City · Ads        → company="Meta, Redwood City",  location=undefined
Engineering Lead · Ford, Ann Arbor · Research      → company="Ford, Ann Arbor",     location=undefined
```

Byte-identical to #616's reported symptom. So the issue's third acceptance criterion — *"A multi-word city without a state/country suffix splits the same way a single-word city does"* — is **not** met as written; it is met for three named cities.

The closed-vocab design is the right call and I am not asking you to change it. For Pass F the tail is comma-delimited and `^…$`-anchored, but an open two-Title-Case-word rule would still split `Acme, Northwind Systems`, so the closed set is still earning its keep.

The ask is disclosure, not a rebuild: record the narrowing in the PR body and in the #616 close comment, so closing the issue does not imply a guarantee the code does not make. Optionally add `Santa Clara` — this file's own Pass A docblock (line 80) already names it as the canonical multi-word city example, and Nvidia/Intel/AMD make it likelier to show up than Menlo Park.

src/lib/heuristics/extract/experience.multiword-city.test.ts
```diff
@@ -84,3 +84,107 @@ describe("multi-word city on space-folded header (#368)", () => {
...
+// `Company City, ST` shape. #616 is the same "multi-word bare city" weakness on
+// the single-line MIDDOT header path: `Title · Company, City · Team` where the
+// city carries NO state/country suffix — the closed BARE_LOCATION_RE vocab (Pass
+// F of stripLocationSuffix) knows single-word "Hyderabad" but not multi-word
+// "Mountain View", so the middle segment stayed whole and `company` swallowed
+// the city. Fixed by adding Mountain View (a canonical tech-hub example) to the
+// single-source MULTIWORD_US_CITY_ALT vocab so BARE_LOCATION_RE full-matches it.
+// The trailing-segment case is covered too so behavior is position-independent.
+describe("multi-word city without state/country suffix on middot header (#616)", () => {
+  it("row (b) — middle segment: `Company, MultiWordCity` splits into company + location", () => {
```
s-annam, line 98: **Secondary — two of the three added cities ship untested.**

All five new tests use `Mountain View`; `Palo Alto` and `Menlo Park` have zero coverage. I confirmed both work (`Meta, Menlo Park` → `Meta`/`Menlo Park`; `Stanford, Palo Alto` → `Stanford`/`Palo Alto`), so this is not a bug — it is that nothing guards them. An edit to that alternation string can drop one and the suite stays green.

Table-drive row (b):

```ts
it.each([
  ["Google", "Mountain View", "GFiber"],
  ["Meta", "Menlo Park", "Ads"],
  ["Stanford", "Palo Alto", "SLAC"],
])("row (b) — middle segment: `%s, %s` splits into company + location", (company, city, team) => {
```

src/lib/heuristics/extract/experience-disambiguate.ts
```diff
@@ -43,9 +43,19 @@ const LEGAL_SUFFIX_RE =
...
+ *  `BARE_LOCATION_RE`.
+ *
+ *  Silicon-Valley/Peninsula additions (#616): "Mountain View", "Palo Alto",
+ *  "Menlo Park" are the multi-word tech-hub cities that show up co-located
+ *  with Google/Meta/Stanford in a `Title · Company, City · Team` middot header
+ *  with NO trailing state suffix. Without the vocab entry, Pass F of
+ *  `stripLocationSuffix` (bare-city tail check) failed to full-match them and
+ *  the middle segment stayed whole ("Google, Mountain View" → company). Same
+ *  closed-vocab discipline as the pre-existing entries: a real company that
+ *  merely contains one of these tokens ("Mountain View Software") still fails
```
s-annam, line 55: **Nit — this safety claim is asserted in prose and tested nowhere.**

*"a real company that merely contains one of these tokens (`Mountain View Software`) still fails the `^…$`-anchored full-string match"* is the load-bearing justification for the whole closed-vocab approach. I verified it holds in both positions (`Engineer · Mountain View Software · Core` keeps the company whole; `Engineer · Acme, Mountain View Software` stays unsplit), but a prose-only invariant drifts silently. One assertion makes it durable.

s-annam (approved):
Solid fix, and the diagnosis in the PR body is better than the issue's own. #616 framed this as position-dependent; you correctly identified it as position-independent (a shared closed vocab consulted by both Pass F and the team-slot mirror) and proved it with the trailing-segment test. Extending the single-source `MULTIWORD_US_CITY_ALT` rather than adding a parallel list is the right seam.

**I verified the guard proof rather than taking it on faith.** In a throwaway worktree I reverted just `|Mountain View|Palo Alto|Menlo Park` from the const and re-ran the file: **2 failed, 6 passed** — exactly the two new tests that should fail (row b, position-independence), with rows a/c/d and the three pre-existing #368 tests staying green. The tests are non-vacuous and correctly scoped.

I also probed the false-positive surface. No regressions introduced:

| Header | Result | Verdict |
|---|---|---|
| `Engineer · Mountain View Software · Core` | `company="Mountain View Software"` | anchored match holds |
| `Engineer · Acme, Mountain View Software` | unsplit | anchored match holds |
| `Engineer · Acme Palo Alto, CA` | `company="Acme"`, `location="Palo Alto, CA"` | improved (was `Acme Palo` / `Alto, CA`) |
| `Engineer · Palo Alto Networks, CA` | `company="Palo Alto"`, `location="Networks, CA"` | **pre-existing on `main`** — Pass B single-token, not yours |

That last row is a real defect but it is not this PR's — I A/B'd it against `main` before writing it up. Flagging it so nobody attributes it here later.

## Findings

Three, all inline on the lines they concern: **0 Blocking**, 2 Secondary (the vocab addition does not satisfy the issue's third acceptance criterion as written; two of the three added cities ship untested), 1 Nit (the docblock's `^…$`-anchoring safety claim is asserted in prose and tested nowhere).

## Gates

- **Fixture PII** — skipped, no fixture binaries touched.
- **Design-system / reuse**, **style tokens** — skipped, no `src/components/` or styling changes.
- **fallow / dead code** — passed in CI (3s, green); no new exports in the diff.
- **Skill/script command review** — skipped, no `.claude/` or `scripts/` files.

All three checks green (`verify` 2m19s, `fallow`, Cloudflare Pages). One commit, MERGEABLE. Reviewed at `c4f3826`.

**Deviation to disclose:** I ran the generic correctness pass inline instead of delegating it to a fan-out reviewer — the diff is 2 files / +116 / one changed source line. What replaced it is stronger than a static read: the revert-based guard proof plus 11 empirical probe headers, including the A/B against `main` that caught one apparent finding as pre-existing before I filed it.

**Verdict rule:** 0 Blocking → `APPROVE`. Secondary and Nit listed as non-blocking; not soft-gating on them.

---

src/lib/heuristics/extract/experience-disambiguate.ts
s-annam: Addressed in `1855ff2` (#636), both halves — the narrowing and the disclosure.

Added `Santa Clara`, `Redwood City` and `Ann Arbor`, which closes exactly the three headers you reproduced:

| Header | Before | After |
|---|---|---|
| `… · Nvidia, Santa Clara · Compute` | company=`Nvidia, Santa Clara`, location=∅ | company=`Nvidia`, location=`Santa Clara` |
| `… · Meta, Redwood City · Ads` | company=`Meta, Redwood City`, location=∅ | company=`Meta`, location=`Redwood City` |
| `… · Ford, Ann Arbor · Research` | company=`Ford, Ann Arbor`, location=∅ | company=`Ford`, location=`Ann Arbor` |

Three vocabulary entries, no logic change. Proven non-vacuous: removing them fails exactly the three new table rows.

**The disclosure is the part that matters, and it is now in three places** rather than only prose. The const's docblock states the vocabulary is closed by design and that an unlisted multi-word city still folds into `company`; a test (`closed vocabulary is closed`) pins that with an unlisted city, and its comment says the expected action when a city is added is to move the row up into the table, not delete the test; and #616 carries a comment so its closure is not read as a general guarantee of AC3.

Your one caveat, taken seriously: adding `Santa Clara` is only safe because `Santa Clara University` stays whole. That is now asserted rather than assumed — see the sibling thread.

src/lib/heuristics/extract/experience.multiword-city.test.ts
s-annam: Fixed in `1855ff2` (#636), using your table-driven shape as written.

Row (b) now runs one row per vocabulary entry added for #616 — six rows: Mountain View, Menlo Park, Palo Alto, plus the three from the sibling thread. That makes the coverage structural: adding a city without adding its row is the only way to regress, and it is visible in the diff.

src/lib/heuristics/extract/experience-disambiguate.ts
s-annam: Fixed in `1855ff2` (#636) — four assertions, not one, because the anchor has to hold in two positions.

Three leading-company cases (`Mountain View Software`, `Redwood City Ventures`, `Santa Clara University`) plus `Acme, Mountain View Software` in the comma-tail position, which is where Pass F actually runs `BARE_LOCATION_RE` and so is the position the anchor earns its keep in.

Worth noting this one stopped being optional: adding `Santa Clara` to the vocabulary would have been reckless with the university case unpinned, so your nit ended up gating the sibling thread's fix.

s-annam:
All three review threads addressed in #636 and resolved.

`Santa Clara`, `Redwood City` and `Ann Arbor` added, closing the three headers the review reproduced as still-broken; row (b) is now table-driven with one row per vocabulary entry; and the "a company merely containing a city stays whole" claim is pinned by four assertions across both anchor positions — which turned out to gate the Santa Clara addition, since `Santa Clara University` had to stay whole.

The narrowing is disclosed in the const's docblock, in a test using an unlisted city, and in a comment on #616.

PR #636, opened 2026-07-28
fix(pdf,score,parser,ui): Summary widow control, plus the #630/#633/#634 review-nit sweep (#635)

### LINKED ISSUE #635: fix(pdf): the Summary body is the one wrapped block with no widow control
Follow-up from a review nit on #630 ([thread](https://github.com/offlinecv/OfflineCV/pull/630#discussion_r0)).

#629/#631/#632 gave every wrapped **bullet** both halves of the page-break guarantee — never one line alone at a page bottom, never one alone at a page top. They scope to bullets, entry headers and section headings, and the module docblock is careful not to claim more.

The **Summary body** is the one wrapped block left on the plain per-line path:

```ts
```

`ensureWrappedLine` widens its reservation only when `opts.widowControl` is set, and `drawBullet` is currently the only caller that sets it. So a summary that wraps to N lines with N-1 fitting on the page leaves its last line alone at the top of the next page — exactly the defect #631 exists to prevent, in the one block type #631 does not cover.

## Reproduced, and the reachability measured

Rendering a synthetic model with a summary of `words` filler tokens and counting drawn summary lines per page on `main` (`7596161`):

| words | summary lines | lines per page |
|---|---|---|
| 596 | 49 | `[49]` |
| **602** | **50** | **`[49, 1]`** ← widow |
| **608** | **50** | **`[49, 1]`** ← widow |
| 614 | 51 | `[49, 2]` |
| 626 | 52 | `[49, 3]` |

So the defect is real but **remote**: the summary is always drawn immediately after the contact block, so nothing can push a short one across a break — it has to outrun a whole page on its own, which takes roughly **50 wrapped lines / ~600 words**. No corpus fixture comes close.

Worth fixing anyway, on two grounds:

1. The summary became a **free-text field the user can edit and paste into** (#625). "Nobody would write a 600-word summary" is a claim about authorship, not about what the field accepts.
2. The fix is one option at one call site, and the mechanism is already built and tested.

## The orphan half needs nothing

`drawSectionHeading` already reserves the heading plus one body line, and this block can only ever *begin* at the top of page one — so its opening can never sit at a page bottom. Only the tail is at risk. This asymmetry is worth stating in the code, since `ensureWrappedLine`'s docblock currently explains the other side of the break purely in terms of `bulletKeepLines`, which the summary does not go through.

## Explicitly out of scope

The capped body-header (skills) path: lines past `BODY_HEADER_KEEP_LINES` flow through the plain `ensure`, so a skills list can widow its last line too. Leave it — divisibility past the cap is the entire point of the cap, and covering it would re-introduce the density cost #630 measured.

## Acceptance criteria

- **AC1** — a summary that straddles a page break leaves at least `BULLET_KEEP_LINES` (2) drawn lines on each side, never exactly one.
- **AC2** — a guard test in `export-layout-contract.test.ts` that locates the spill boundary by **measurement** (bisection, bracketed on both sides) rather than hard-coding a word count, and asserts across a window past it. Proven non-vacuous by reverting the fix.
- **AC3** — no corpus movement: `corpus-roundtrip` / `corpus-edit-roundtrip` and the existing 71 layout-contract tests unchanged.
- **AC4** — the docblocks that currently say widow control is bullets-only (`DrawTextOpts.widowControl`, `ensureWrappedLine`, the module header) updated, since they would otherwise be false.

## Also in scope — a second nit from the same review

`BODY_HEADER_KEEP_LINES = 2` is unpinned by any test: `1` and `2` produce identical output on every measured case, so nothing in the suite would fail if it changed. That is fine — an assertion there would be a tautology — but the **gate** that applies the cap (`headerBold: false`, in `entryHeadHeight`) *is* load-bearing and *is* the subtle part. Add a comment at the constant pointing at `entryHeadHeight`, so a future reader does not tune the number in isolation and assume the cap is unconditional.

s-annam:
Consolidates the review follow-ups from three PRs that merged in the same window — **#630, #633 and #634** — together with the #635 Summary widow-control fix that #630's review surfaced.

Reviewed across those three by @rohithgollapalli, @Vaishnavi1709, @Samhit21 and @s-annam. Every finding any of them raised is accounted for below: fixed here, filed as a follow-up issue, or declined with a reason on its own thread. Nothing was dropped silently, and the per-item attribution in each section is the record of who found what.

## The defect

#629/#631/#632 gave every wrapped **bullet** both halves of the page-break guarantee. They scope to bullets, entry headers and section headings, and the module docblock was careful not to claim more. The **Summary body** is the one wrapped block left on the plain per-line path:

```ts
```

`ensureWrappedLine` widens its reservation only under `opts.widowControl`, and `drawBullet` was the only caller setting it. So a summary with N-1 of its N lines fitting leaves the last one alone at the top of the next page.

## Reachability — measured on `main` (`7596161`), not assumed

Synthetic model, summary of `words` filler tokens, counting drawn summary lines per page:

| words | summary lines | per page (before) | per page (after) |
|---|---|---|---|
| 596 | 49 | `[49]` | `[49]` |
| **602** | **50** | **`[49, 1]`** ← widow | `[48, 2]` |
| **608** | **50** | **`[49, 1]`** ← widow | `[48, 2]` |
| 614 | 51 | `[49, 2]` | `[49, 2]` |
| 626 | 52 | `[49, 3]` | `[49, 3]` |

The summary is always drawn immediately after the contact block, so nothing can push a *short* one across a break — it has to outrun a whole page on its own, ~50 wrapped lines / ~600 words. **Remote, and worth saying so plainly** rather than dressing this up as a live bug. Fixed anyway because the summary became a free-text field the user edits and pastes into (#625) — "nobody writes a 600-word summary" is a claim about authorship, not about what the field accepts — and because the mechanism was one option away.

## The orphan half is genuinely not needed

`drawSectionHeading` already reserves the heading plus one body line, and this block can only ever *begin* at the top of page one, so its opening can never sit at a page bottom. That asymmetry is now stated in `ensureWrappedLine`'s docblock, which previously explained the far side of the break purely via `bulletKeepLines` — a path the summary does not take.

The capped body-header (skills) path is left uncovered **deliberately**, exactly as the review recommended: divisibility past `BODY_HEADER_KEEP_LINES` is the entire point of the cap, and covering it would re-introduce the density cost #630 measured at ~100pt.

## The guard

`export-layout-contract.test.ts` gains one test that:

- **bisects** to the spill boundary rather than hard-coding a word count, so it keeps straddling the break if the page geometry or type scale changes;
- **brackets both ends** (empty summary must fit page one; 800 words must spill), so it fails loudly instead of going vacuous;
- asserts the same `MIN_BULLET_LINES_PER_PAGE` minimum a wrapped bullet must meet, across a window past the boundary.

**Proven non-vacuous.** Reverting the one-option change fails exactly this test and nothing else: `words=556: summary widowed [49,1]: expected 1 to be greater than or equal to 2`. Restoring it returns 72/72.

## Second nit — `BODY_HEADER_KEEP_LINES`

Confirmed unpinned: `1` and `2` render identically on every measured case, so an assertion there would be a tautology and none is added. What *is* load-bearing is the **gate** that applies the cap (`headerBold: false`, in `entryHeadHeight`), so the constant now carries a comment pointing at it — the failure mode the nit named is a future reader tuning the number in isolation and assuming the cap is unconditional.

## Docblocks corrected

Three said widow control was bullets-only and would now be false: `DrawTextOpts.widowControl`, `Layout.ensureWrappedLine`, and the module header. Also the test file's contract summary.

## Verification

`npm run typecheck` 0 · `npx eslint src/lib/pdf/` 0 · `npx vitest run src/lib/pdf/` **312 passed / 24 files** · `export-layout-contract.test.ts` **72 passed** (71 before).

**No corpus movement**: `corpus-roundtrip` and `corpus-edit-roundtrip` are unchanged — no fixture has a summary anywhere near the threshold, which is what makes the corpus glyph-diff the review anticipated cheap rather than expensive here.

No fixture binary added or changed. The widow-control change is `src/lib/pdf/` only; the review nits below widen this PR into `src/components/features/`, `src/design-system/` and `src/lib/score/`.

Closes #635

## Also in this PR — the review-nit sweep from #630, #633 and #634

Three PRs merged in the same window (#630, #633, #634), each leaving unresolved review threads behind. Re-pushing to any of those branches would have dismissed their approvals (`main` has `dismiss_stale_reviews: true`), and opening three follow-up PRs would cost three review rounds. This PR had no review yet, so every nit worth fixing is swept here and reviewed once.

**#630's two threads are already satisfied by this PR's own change** — the Summary widow control above *is* @rohithgollapalli's first thread, and the `BODY_HEADER_KEEP_LINES` comment he asked for is the second. Nothing extra was needed for them.

Six items land. Two are behavioural and carry their own proof; four are behaviour-neutral.

### Behavioural — proved, not asserted

**1. The spelled-year guard and the cardinal matcher disagreed about separators.** *(@rohithgollapalli on #633, Secondary — `score.ts:242`)*

`CARDINAL_SEPARATOR` accepts whitespace *or* a hyphen; `WORD_YEAR_TOKEN` joined its parts with `\s+` only. So a spelled year written with hyphens escaped the strip, and `QUANTIFYING_CARDINAL` then matched off its own leftover fragment (`two-t`). Reproduced exactly as reported, through the real scorer:

| Bullet | Before | After |
|---|---|---|
| `…in two thousand nineteen and stayed` | none | none |
| `…in two-thousand-nineteen and stayed` | **METRIC** | none |
| `…in two thousand-nineteen and stayed` | **METRIC** | none |
| `Hired in nineteen-hundred-ninety-eight…` | **METRIC** | none |
| `Hired in nineteen hundred ninety-eight…` | none | none |
| `Processed two thousand records per day` | METRIC | METRIC |
| `Shipped two internal tools` | METRIC | METRIC |
| `Led a four-month migration` | METRIC | METRIC |
| `Owned twenty-two dashboards` | METRIC | METRIC |
| `Processed millions of records nightly` | METRIC | METRIC |

The last row of the "before" set is the one that stung: the un-hyphenated form was already in the `notQuantified` test list and passed, so the existing set bracketed the bug without crossing it. Both constants now share `CARDINAL_SEPARATOR`, and the docblock says why the sharing is load-bearing rather than stylistic.

Six test rows added, not three — the three hyphenated years, plus `Processed two-thousand records per day` on the *accept* side. The widened separator must still require a year **tail** word, or the fix would start eating genuine hyphenated quantities. **Proven non-vacuous:** reverting the separator fails exactly the three new reject rows and nothing else.

**`ATS_SCORE_ALGO_VERSION` 1.6 → 1.7**, which is what kept this out of the earlier batch. The cost is real but bounded, and it is now measured rather than assumed: pinning the version at 1.6 and re-running the corpus, **all 54 snapshots pass unchanged** — so no fixture's parse or score moves, and the only snapshot delta in this diff is the version stamp (54 files, 54 insertions, 54 deletions, zero non-`algoVersion` lines). The bump invalidates persisted `resume-library` cache records, which then re-parse from the stored PDF blob. Scores can only move **down**, only by removing false Specificity credit, and only on this rare shape.

**2. The multi-word-city defect class survived for every city outside the three added.** *(@s-annam on #634, Secondary — `experience-disambiguate.ts:58`)*

The three headers the #634 review reproduced as still-broken, before and after:

| Header | Before | After |
|---|---|---|
| `… · Nvidia, Santa Clara · Compute` | company=`Nvidia, Santa Clara`, location=∅ | company=`Nvidia`, location=`Santa Clara` |
| `… · Meta, Redwood City · Ads` | company=`Meta, Redwood City`, location=∅ | company=`Meta`, location=`Redwood City` |
| `… · Ford, Ann Arbor · Research` | company=`Ford, Ann Arbor`, location=∅ | company=`Ford`, location=`Ann Arbor` |

Three vocabulary entries in `MULTIWORD_US_CITY_ALT`, no logic change. **Proven non-vacuous:** removing the three entries fails exactly the three new table rows.

**One effect I did not set out to get, now pinned.** `MULTIWORD_US_CITY_ALT` is shared with `KNOWN_MULTIWORD_US_CITY_RE`, the space-folded (#368) pass — so the three additions improve that path too. A/B against the previous vocabulary: `Acme Santa Clara, CA` used to split as company `Acme Santa` / location `Clara, CA` — the exact Pass B single-token truncation #368 exists to prevent — and now splits correctly. Three rows added to the #368 block so a side-effect improvement cannot silently regress; reverting the vocabulary fails all six new city rows across both describe blocks.

**This does not make #616's third acceptance criterion true in general, and the code now says so.** The vocabulary is closed by design — an open "two Title-Case words" rule would split `Acme, Northwind Systems` — so an unlisted multi-word city still folds into `company`. That narrowing is recorded in the const's docblock, pinned by a test (`closed vocabulary is closed`, using an unlisted city), and noted on #616 so its closure is not read as a guarantee the code does not make.

**3. The load-bearing safety claim was asserted in prose and tested nowhere.** *(@s-annam on #634, Nit — same file, line 55)*

*"a real company that merely contains one of these tokens (`Mountain View Software`) still fails the `^…$`-anchored full-string match"* is the entire justification for growing a closed vocabulary at all — and nothing enforced it. Four assertions now do, across both positions the anchor has to hold in: `Mountain View Software`, `Redwood City Ventures` and `Santa Clara University` stay whole as leading company names, and `Acme, Mountain View Software` stays unsplit in the comma-tail position, which is where Pass F actually runs `BARE_LOCATION_RE`. Adding `Santa Clara` to the vocabulary would have been reckless without the university case pinned.

**4. Two of #634's three cities shipped untested.** *(@s-annam on #634, Secondary)*

All five of that PR's tests used `Mountain View`; `Palo Alto` and `Menlo Park` had zero coverage, so an edit to the alternation string could drop one and the suite would stay green. Row (b) is now table-driven with **one row per vocabulary entry added for #616** — six rows, exactly the reviewer's suggested shape. Coverage is now structural: adding a city without adding its row is the only way to regress, and that is visible in the diff.

### Behaviour-neutral

**5. `SectionHeading` had four verbatim copies → one shared component.** *(#633's own known-nits list; count corrected by @rohithgollapalli)*
`ReconstructedResume.tsx:249`, `ReconstructedSkills.tsx:43`, `ReconstructedEducationSkills.tsx:45` and `ReconstructedSummary.tsx:45` each carried the same six lines and the same class string. Extracted to `src/design-system/shared/SectionHeading.tsx`, exported through the `@design-system` barrel; all four call sites import it. Same element, same class string, no local state in any copy — the rendered markup is identical. This is the `CLAUDE.md` Golden Rule case: exactly one shared piece per concern. #633's own list called it "a **third** verbatim copy"; it was the fourth, and @rohithgollapalli diffed all four.

**6. Five test helpers still built `wordCount` with the pre-#627 expression.** *(#633's own known-nits list; independently verified inert by @rohithgollapalli and @Vaishnavi1709)*
`apply-overrides.test.ts:21`, `undo-batch.test.ts:213`, `ReconstructedResume.test.ts:23`, `ats-resume-model.test.ts:21`, `group-bullets.test.ts:36` each used `text.split(/\s+/)…`, the expression #627 exists to delete. Both reviewers independently confirmed them inert (nothing reads the field; the stub texts have no punctuation-only tokens), so this is consistency, not correctness — all five now call `countWords`.

**7. The Structure hint stated counts without its criterion.** *(#633's own known-nits list; @rohithgollapalli endorsed the `title`-attribute fix)*
#624 shortened it to `verb-led 0/23 · length 23/23`, which no longer says what "length" is measured against. Adds an optional `hintTitle` to `Dimension`, used only by the Structure card. `BULLET_LENGTH_MIN_WORDS` / `BULLET_LENGTH_MAX_WORDS` are now exported so the tooltip reads the window from the constants that enforce it — a hand-typed "8–30" in the UI would drift the moment the window moves.

**8. The `noun` prop reached the kind label but not the editor.** *(@Vaishnavi1709 on #633, Secondary — the one user-visible UI fix here)*
`RewriteReviewList.tsx:179,181`. `noun` correctly drove `kindLabel` and both control `aria-label`s, then the `EditableField` two lines below hardcoded `placeholder="edit this bullet"` and `label="Edit proposed bullet"`. Accepting a **summary** rewrite (a #625 flow) therefore read "Edited summary" in the header and "edit this bullet" in the textarea, with the same mismatch in the screen-reader label.

Note the substitution is `${noun}`, **not** the row's `lower` — `lower` is the lowercased *kind label* (`"edited bullet"`), so interpolating it would have produced "edit this edited bullet". Three tests cover the passed noun, the empty-state placeholder (which needs its own pair — the placeholder only paints when the edited side is empty, so asserting it on a filled pair would pass vacuously), and the `"bullet"` default. **Proven non-vacuous** — reverting the two interpolations fails exactly the two summary-noun tests and leaves the default-case test passing.

**A pre-existing defect the #634 review flagged now has a home: #641.** That review noted `Palo Alto Networks, CA` → company `Palo Alto` / location `Networks, CA`, A/B'd it against `main`, and flagged it specifically so it would not be misattributed to #634 — but it was never filed. I repeated the A/B against this branch: the shredded rows are byte-identical with and without the new vocabulary entries, so it is Pass B's single-token rule and neither PR's doing. #641 carries the repro, the A/B table, and two candidate directions.

### Deliberately left out

Everything else the three reviews raised, with who raised it. These are not nits — each changes behaviour in a way that needs its own design and its own proof.

- **`ReconstructedRole.tsx:298` undo strip unmounts on prune** — @Vaishnavi1709 on #633, Secondary. Same class as #633's round-1 blocking defect, one lifecycle level up: `pruneEmptyAddedEntries` drops the added role while its undo strip is live, so the undo flashes for a frame and is gone. The fix is architectural — lift the strip host for added-role bullets the way `ExperienceSection` already hosts the Other bucket's, or defer prune while a removed-strip is live. **Filed as a follow-up issue.**
- **`variant="icon"` 24×24 floor belongs on the `Button` primitive** — @Vaishnavi1709 on #633, nit, and she asked for a follow-up issue explicitly. Six call sites still carry the 14×14 hitbox. Layout-affecting, and `ChipListEditor.tsx:45` documents a 5px overlay workaround, so a blanket `min-h` on the primitive would make the chip row worse. **Filed as a follow-up issue.**
- **Inline `text-[11px]` typography** — @Samhit21 on #633, additive nit on `ResumeBulletRow.tsx:156,195`, which he scoped as pre-existing debt for "whenever that file is next touched". Measuring it made it bigger than the two lines: font size is **not tokenised at all** in this repo — `@theme inline` maps `--color-*` only, there is no `tailwind.config.*` (Tailwind v4), and `src/` carries **67** arbitrary sizes (`text-[11px]` ×48, `text-[10px]` ×17, `text-[9px]` ×2). Fixing only these two would mint a token used by 2 of 48 consumers and leave 46 hand-styled — the parallel-copy outcome the Golden Rule exists to prevent. It is a sweep or nothing. **Filed as #640.**
- **`isProseLine` (`line-primitives.ts:83`) still uses the pre-#627 expression** — @Vaishnavi1709 on #633, nit. The one *production* consumer, and the trap in this set: #627 exists because `split(/\s+/).filter(Boolean)` counts punctuation-only tokens as words, so swapping it changes which lines classify as prose. That is parser behaviour with real corpus-movement risk — unlike the two behavioural fixes above, which the corpus proves inert.

Judged not worth fixing, by the reviewer who raised it:

- **`scoreBulletPool` / `analyzeBullets` double-walk bullets** — @Vaishnavi1709 on #633, nit. Her guidance was to fold only if you are in the file anyway, and this PR *is* in `score.ts`. Not taken, and the reason is specific: the cheap fold does not exist. `scoreBulletPool` has two other callers (`score.ts:456`, `:474`) that pass raw strings, so the only way to stop the second walk is to change the shared function's signature to take `BulletObservation[]` — which contradicts the "parallel shape is deliberate" docblock and touches the authed scorer. That is a scorer-core refactor riding along with a scoring change in the same PR, which is exactly the combination worth not making.
- **Proper nouns embedding a cardinal read as quantifications** (`Advised Four Seasons on their loyalty platform`) — @rohithgollapalli on #633, nit. Inherent to a keyword heuristic, the set is unbounded, and the cost is half a Specificity credit on one bullet. Worth recording that the corpus validation reads as stronger coverage than it is: the corpus simply contains no such employer.

Still open from #633's known-nits list, unchanged here and agreed by both reviewers as their own follow-ups: `applyOverrides` at 17 positional params (wants an options-object refactor, which would have buried six fixes in signature churn), and `ResumeBulletRow.tsx` at 300 LOC (debt relocated honestly — `ReconstructedRole.tsx` fell 202 lines in the same change).

### Verification

`npm run verify` **exit 0** — `tsc -b --noEmit` clean · `eslint .` clean · `check:fixtures` green · **3545 passed / 10 skipped** across 244 files · `vite build` 956 modules. **No fixture binary touched** — the 54 `.expected.json` changes are the `algoVersion` stamp and nothing else (54 files, 54 insertions, 54 deletions, zero non-`algoVersion` lines in the diff).

fallow reports 10 findings, none authored here: the complexity list is `mapTitleFirst`, `recoverLocation`, `stripLocationSuffix`, `computeAnonymousAtsScore`, `scoreCompleteness` and the render path, all pre-existing functions this diff does not enter — it adds a vocabulary string and a regex, both at module scope. `experience.multiword-city.test.ts` picks up a unit-size finding for its grown `describe` block, in the same shape as the four other test files already on that list. Report-only per `CLAUDE.md`.
## Provenance

| Stage | Model | Effort |
|---|---|---|
| Repro, fix, test, PR | Claude Opus 5 (1M context) | high |
| Review-nit sweep from #630/#633/#634 | Claude Opus 5 (1M context) | high |

No independent adversarial review round was run on this change. Single-model work, stated rather than implied.

### DIFF
--- a/src/components/features/AtsScoreReadout.tsx
+++ b/src/components/features/AtsScoreReadout.tsx
@@ -8,7 +8,11 @@
  */
 
 import type { AnonymousAtsScore } from "../../lib/score/score.ts";
-import { getScoreTier } from "../../lib/score/score.ts";
+import {
+  getScoreTier,
+  BULLET_LENGTH_MIN_WORDS,
+  BULLET_LENGTH_MAX_WORDS,
+} from "../../lib/score/score.ts";
 import type { SectionAnchor } from "../../lib/anchors.ts";
 import { getScoreRecommendation } from "../../lib/score/recommendation.ts";
 import { ScoreRing } from "./ScoreRing.tsx";
@@ -24,6 +28,11 @@ interface DimensionProps {
   max: number;
   gradable: boolean;
   hint: string;
+  /** Spelled-out criterion behind a terse `hint`, surfaced as a tooltip. #624
+   *  shortened the Structure hint to bare counts (`length 13/23`), which no
+   *  longer states what "length" is measured against; this restores the
+   *  criterion without re-lengthening the visible line. */
+  hintTitle?: string;
   /** Hash-prefixed scroll target — narrowed to a known section so a dead link
    *  (a `#foo` with no matching rendered id) is a compile error (#153). */
   anchor: SectionAnchor;
@@ -35,6 +44,7 @@ function Dimension({
   max,
   gradable,
   hint,
+  hintTitle,
   anchor,
 }: DimensionProps) {
   const pct = max > 0 ? Math.round((value / max) * 100) : 0;
@@ -68,7 +78,9 @@ function Dimension({
           />
         </div>
       )}
-      <p className="text-[11px] text-content-tertiary">{hint}</p>
+      <p className="text-[11px] text-content-tertiary" title={hintTitle}>
+        {hint}
+      </p>
     </a>
   );
 }
@@ -147,6 +159,7 @@ export function AtsScoreReadout({ score }: AtsScoreReadoutProps) {
             max={score.structure.max}
             gradable={score.structure.gradable}
             hint={structureHint}
+            hintTitle={`Verb-led: bullet opens with an action verb. Length: bullet is ${BULLET_LENGTH_MIN_WORDS}–${BULLET_LENGTH_MAX_WORDS} words.`}
             anchor="#reconstructed-resume"
           />
           <Dimension
--- a/src/components/features/RewriteReviewList.test.tsx
+++ b/src/components/features/RewriteReviewList.test.tsx
@@ -22,7 +22,7 @@ import { act } from "react";
 (globalThis as { IS_REACT_ACT_ENVIRONMENT?: boolean }).IS_REACT_ACT_ENVIRONMENT =
   true;
 
-import { RewriteReviewList } from "./RewriteReviewList.tsx";
+import { RewriteReviewList, BulletReviewRow } from "./RewriteReviewList.tsx";
 import type { AlignedPair } from "../../lib/rewrite-review/align-bullets.ts";
 import type { RewriteReview } from "../../hooks/useRewriteReview.ts";
 
@@ -178,5 +178,75 @@ describe("RewriteReviewList", () => {
     )!;
     act(() => acceptAll.dispatchEvent(new MouseEvent("click", { bubbles: true })));
     expect(review.acceptMany).toHaveBeenCalledWith(["m:0:0", "add:1", "del:1"]);
+  });
+});
+
+describe("BulletReviewRow — the `noun` prop names the row's content everywhere", () => {
+  // #625 gave the row a `noun` so the whole-résumé review could reuse it for the
+  // one-field summary section. The prop reached the kind label and the two
+  // control `aria-label`s but NOT the editor's placeholder or label, which stayed
+  // hardcoded "bullet" — so accepting a summary rewrite read "Edited summary" in
+  // the header and "edit this bullet" in the textarea, with the same mismatch in
+  // the screen-reader label. Every user-visible noun on the row is asserted here
+  // so a future control added without threading `noun` fails rather than ships.
+  const filled: AlignedPair = {
+    kind: "matched",
+    id: "m:0:0",
+    original: "Engineer with a decade of experience",
+    originalIndex: 0,
+    proposed: "Engineer with a decade of platform experience",
+    proposedIndex: 0,
+  };
+  // The placeholder only paints when the edited side is empty, so it needs its
+  // own pair — asserting it on `filled` would pass vacuously.
+  const empty: AlignedPair = { ...filled, proposed: "" };
+
+  it("names the noun on the kind label, both controls, and the editor label", () => {
+    const el = render(
+      createElement(BulletReviewRow, {
+        pair: filled,
+        review: makeReview(),
+        noun: "summary",
+      }),
+    );
+
+    expect(el.querySelector("li span")?.textContent).toBe("Edited summary");
+    expect(
+      el.querySelector('button[aria-label="Accept this edited summary"]'),
+    ).not.toBeNull();
+    expect(
+      el.querySelector('button[aria-label="Reject this edited summary"]'),
+    ).not.toBeNull();
+    // The editor's own label — half of the pair that stayed hardcoded.
+    expect(
+      el.querySelector('[aria-label="Edit Edit proposed summary"]'),
+    ).not.toBeNull();
+    // Nothing anywhere on the row still calls it a bullet.
+    expect(el.innerHTML).not.toContain("bullet");
+  });
+
+  it("names the noun in the empty-state placeholder — the other half", () => {
+    const el = render(
+      createElement(BulletReviewRow, {
+        pair: empty,
+        review: makeReview(),
+        noun: "summary",
+      }),
+    );
+
+    expect(el.textContent).toContain("edit this summary");
+    expect(el.innerHTML).not.toContain("bullet");
+  });
+
+  it("defaults to \"bullet\", so every caller that omits the prop is unchanged", () => {
+    const el = render(
+      createElement(BulletReviewRow, { pair: empty, review: makeReview() }),
+    );
+
+    expect(el.querySelector("li span")?.textContent).toBe("Edited bullet");
+    expect(
+      el.querySelector('[aria-label="Edit Edit proposed bullet"]'),
+    ).not.toBeNull();
+    expect(el.textContent).toContain("edit this bullet");
   });
 });
--- a/src/lib/pdf/render-ats-pdf.ts
+++ b/src/lib/pdf/render-ats-pdf.ts
@@ -44,6 +44,14 @@
  * bullets paginate freely and no reservation ever spans the entry — because
  * reserving a whole entry is the density failure #630 measured at ~100pt.
  *
+ * The widow half is not bullets-only. The SUMMARY body is the other wrapped
+ * block on the plain `drawText` path, and it takes the same `widowControl`
+ * (#635). Its orphan half needs no reservation beyond the one line its heading
+ * already keeps: the summary always begins at the top of page one, so it can
+ * only ever straddle a break by outrunning a whole page on its own. The capped
+ * body-header (skills) path is deliberately left uncovered — divisibility past
+ * `BODY_HEADER_KEEP_LINES` is the entire point of the cap.
+ *
  * Two deliberate limits keep those reservations from costing page density: a
  * BODY-TEXT "header" (`headerBold: false` — the skills list) contributes at most
  * `BODY_HEADER_KEEP_LINES`, and a reservation taller than a whole page is
@@ -513,9 +521,10 @@ type DrawTextOpts = {
    *  when the text fits on ONE line, so measured offsets are accurate. */
   linkSpans?: Array<{ display: string; href: string }>;
   /** Paginate the wrapped lines under widow control (#631) — see
-   *  {@link Layout.ensureWrappedLine}. Set only by {@link Layout.drawBullet}; a
-   *  header/contact/summary block keeps the plain per-line behaviour. Not an
-   *  input to wrapping, so {@link Layout.measureTextHeight} ignores it. */
+   *  {@link Layout.ensureWrappedLine}. Set by {@link Layout.drawBullet} and by
+   *  the summary body (#635), the two wrapped free-text blocks; a name/contact/
+   *  header line keeps the plain per-line behaviour. Not an input to wrapping,
+   *  so {@link Layout.measureTextHeight} ignores it. */
   widowControl?: boolean;
   /** Height (pt) that must land on the same page as this block's FINAL drawn
    *  line — the NEXT bullet's keep-opening (#632). Set only by
@@ -630,8 +639,14 @@ class Layout {
    * earlier line breaking naturally leaves the tail intact, and every later line
    * is inside a tail already placed.
    *
-   * The other side of the break is guaranteed by the caller's OPENING
-   * reservation ({@link bulletKeepLines}): it committed `BULLET_KEEP_LINES` lines
+   * The other side of the break is guaranteed by the CALLER's opening
+   * reservation, which is why this method owns only the tail. For a bullet that
+   * is {@link bulletKeepLines}; for the summary body (#635) it is the one line
+   * its section heading keeps, which suffices there because the block always
+   * begins at the top of page one and so never opens at a page bottom.
+   *
+   * For a bullet specifically ({@link bulletKeepLines}): it committed
+   * `BULLET_KEEP_LINES` lines
    * to the page the block starts on, and a block reaching this branch has
    * `total >= 2 * BULLET_KEEP_LINES`, so the lines before the tail are at least
    * that many. A block whose own height exceeds a page simply breaks more than
@@ -1253,7 +1268,17 @@ export async function renderAtsResumePdf(
     // The summary body is plain wrapped text, so the heading must keep one BODY
     // line with it (#629).
     drawSectionHeading(layout, model.summaryHeading ?? "Summary", SIZE_BODY * LINE_GAP);
-    layout.drawText(model.summary, { size: SIZE_BODY });
+    // The body takes `widowControl` for the same reason a bullet does (#631): it
+    // is a wrapped block, so per-line pagination can leave its last line alone at
+    // a page top. It is the only NON-bullet caller — a summary long enough to
+    // outrun page one is remote (the block always starts on page one's first
+    // content line), but it is free text the user can edit and paste into (#625),
+    // and the mechanism costs one option.
+    //
+    // The orphan half needs nothing: the heading above already reserves one body
+    // line, and this block can only ever BEGIN at the top of page one, so its
+    // opening can never sit at a page bottom.
+    layout.drawText(model.summary, { size: SIZE_BODY, widowControl: true });
     layout.advance(GAP_BETWEEN_ENTRIES);
   }
 
@@ -1379,6 +1404,14 @@ function subLineOpts(entry: AtsEntry, mutedColor: RGB): DrawTextOpts {
  * bullet: the block's first line is never left alone at a page bottom. One would
  * add nothing over the plain per-line `ensure`; three or more starts trading real
  * page density for a guarantee body text does not need, since it is divisible.
+ *
+ * NOT pinned by a test, and deliberately so: `1` and `2` produce identical output
+ * on every case the suite measures, so any assertion here would be a tautology.
+ * The load-bearing part of this rule is not the number but the GATE that applies
+ * it — see {@link entryHeadHeight}, where the cap is conditional on the model
+ * declaring the header to be body text. Tune the number here and you change
+ * nothing measurable; move the gate there and you change what a real header
+ * reserves. Read that docblock before touching either.
  */
 const BODY_HEADER_KEEP_LINES = 2;
 
--- a/src/components/features/ReconstructedEducationSkills.tsx
+++ b/src/components/features/ReconstructedEducationSkills.tsx
@@ -23,7 +23,7 @@ import type {
   AddedEntryField,
 } from "../../hooks/useEditableParse.ts";
 import { buildEducationDates } from "../../lib/score/entry-dates.ts";
-import { EditableField } from "@design-system";
+import { EditableField, SectionHeading } from "@design-system";
 import { validateDate } from "../../lib/edit/field-validators.ts";
 import { AddPill, RemoveButton, sectionExitBlur } from "./ReconstructedAdd.tsx";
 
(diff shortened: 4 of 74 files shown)


src/components/features/AtsScoreReadout.tsx
```diff
@@ -68,7 +78,9 @@ function Dimension({
           />
         </div>
       )}
-      <p className="text-[11px] text-content-tertiary">{hint}</p>
+      <p className="text-[11px] text-content-tertiary" title={hintTitle}>
```
Vaishnavi1709, line 81: **Nit (a11y, non-blocking).** `title={hintTitle}` on a non-focusable `<p>` only paints on mouse hover — keyboard users and most screen readers never see it, and touch has no equivalent. The criterion the sweep restores (`Verb-led: … Length: bullet is 8–30 words`) is therefore mouse-only. Common enough pattern that it's not blocking; worth an `aria-describedby` on the parent anchor pointing at a visually-hidden `<span>`, or fold the criterion into the anchor's own `aria-label`, if you're back in this file.

src/components/features/RewriteReviewList.test.tsx
```diff
@@ -180,3 +180,73 @@ describe("RewriteReviewList", () => {
...
+    expect(el.querySelector("li span")?.textContent).toBe("Edited summary");
+    expect(
+      el.querySelector('button[aria-label="Accept this edited summary"]'),
+    ).not.toBeNull();
+    expect(
+      el.querySelector('button[aria-label="Reject this edited summary"]'),
+    ).not.toBeNull();
+    // The editor's own label — half of the pair that stayed hardcoded.
+    expect(
+      el.querySelector('[aria-label="Edit Edit proposed summary"]'),
```
Vaishnavi1709, line 222: **Nit (a11y, non-blocking).** This assertion (and its mirror at line 248 for `"Edit Edit proposed bullet"`) pins a double-`Edit` aria-label — `label={\`Edit proposed ${noun}\`}` gets `EditableField`'s automatic `Edit `/`Add ` prefix (`EditableField.tsx:347`) applied on top, so screen readers announce `Edit Edit proposed summary`. Pre-existing to this PR — the noun swap just switches the tail from `bullet` to `summary` — so out of scope here. Flagging so the next `EditableField` pass can drop `Edit proposed ` from the label (`label={noun}` → announces `Edit summary` / `Edit bullet`).

src/lib/pdf/render-ats-pdf.ts
```diff
@@ -44,6 +44,14 @@
  * bullets paginate freely and no reservation ever spans the entry — because
  * reserving a whole entry is the density failure #630 measured at ~100pt.
  *
+ * The widow half is not bullets-only. The SUMMARY body is the other wrapped
+ * block on the plain `drawText` path, and it takes the same `widowControl`
+ * (#635). Its orphan half needs no reservation beyond the one line its heading
+ * already keeps: the summary always begins at the top of page one, so it can
```
Vaishnavi1709, line 50: **Nit (docblock accuracy, non-blocking).** "The summary always begins at the top of page one" is imprecise — a long contact block can push the summary heading onto page 2 (drawn immediately after contact + `GAP_AFTER_CONTACT`). The orphan-half correctness still holds, but via a different mechanism: `drawSectionHeading` reserves heading + 1 body line together via `ensureBlock`, so summary line 1 is bound to the heading regardless of which page they land on. Consider tightening to state the mechanism (`drawSectionHeading`'s reservation) rather than the shape it holds through.

Vaishnavi1709 (approved):
Approving. Substantial sweep — Summary widow + 8 items across #630/#633/#634 — and the PR body's per-item attribution + non-vacuity proofs make it easy to verify. Corpus proof is right on the mark: the 54 `.expected.json` diffs are only the `algoVersion` stamp — no parsed field moves, so the score-version bump is honestly documented as bounded.

**Gates.** `check:fixtures` green (no binaries touched); design-system reuse green (`SectionHeading` now shared in `@design-system`, 4 call sites converted); style tokens clean; `typecheck` + `lint` clean; **3548 tests passed / 10 skipped** (matches PR body).

**Guard-proof spot-check.** Reverted the widow-control option at `render-ats-pdf.ts:1281` — exactly 1 pagination test failed (`words=556: summary widowed [49,1]`), 71 others stayed green. Non-vacuous and not propped up by any other rule.

**#633 nits I raised are honestly resolved:** the `RewriteReviewList` noun swap is fixed (item 8), and the deferrals each carry a specific reason — filed as follow-ups, declined with the corpus-movement risk called out, or noted as unreachable for the sweep's scope.

## Nits (non-blocking)

- **`AtsScoreReadout.tsx:81`** — see inline.
- **`RewriteReviewList.test.tsx:222,248`** — see inline.
- **`render-ats-pdf.ts:50` docblock** — see inline.
- **`NotDetected` still duplicated across `ReconstructedResume.tsx:250` and `ReconstructedEducationSkills.tsx:45`** — same 2-line body, same `text-sm text-content-tertiary` class. Ride-along for the `SectionHeading` extraction; not required here, worth an issue. *(Body-only — both files are untouched by this diff.)*
- **`ensureWrappedLine` gates `widowControl` on `total >= 4`, so a 3-line summary silently no-ops it** (`render-ats-pdf.ts:672-675`). Bullets get indivisibility via `bulletKeepLines`; summary has no such path. Unreachable in practice — contact would have to fill ~48 lines of page 1 — but the asymmetry is worth a sentence in the docblock. *(Body-only — the gate line is untouched by this diff.)*

PR #639, opened 2026-07-28
fix(parser): date sub-line extra tokens no longer hijack company on trailing-segment headers (#614)

### LINKED ISSUE #614: [parser] experience — date sub-line token hijacks company when the header's company sits in the trailing segment
## Summary

**Category: parser, experience, markdown path.** When an experience header puts the **company in the trailing `·` segment** AND the date sub-line carries **any extra `·`-separated token after the date range**, the parser abandons the title line as the company source and scavenges the date sub-line instead — taking the first post-date token as `company`. The real company is lost entirely and `location` drops to `undefined`.

Both conditions are required. Either one alone parses correctly.

## Evidence (synthetic persona)

Input:

```markdown
### Sr. Engineering Manager · Site Lead, Payments Platform · Globex, Toronto
*01/2024 – 12/2024 · M4 · 22 engineers, 3 squads*

- Ran a cross-team migration to a shared platform.
- Expanded the team's charter into a new product area.
```

Parsed (OUTPUT):

```
title="Sr. Engineering Manager"
company="M4"                              ← should be "Globex"
team="Site Lead, Payments Platform"
location=undefined                        ← should be "Toronto"
start_date="01/2024" end_date="12/2024"
```

`"Globex"` appears nowhere in the parse output.

## Condition matrix

Same bullets and `## Experience` section throughout; only the header shape varies.

| # | Title line | Date sub-line | `company` | `location` |
|---|---|---|---|---|
| a | `Title · Team · Globex, Toronto` | `*01/2024 – 12/2024*` | `Globex` ✅ | `Toronto` ✅ |
| b | `Title · Team · Globex, Toronto` | `*01/2024 – 12/2024 · M4 · 22 engineers, 3 squads*` | **`M4`** ❌ | **`undefined`** ❌ |
| c | `Title · Team · Globex, Toronto, Canada` | `*01/2024 – 12/2024 · M4 · 22 engineers, 3 squads*` | **`M4`** ❌ | **`undefined`** ❌ |
| d | `Title · Globex, Toronto, Canada · Team` | `*01/2024 – 12/2024 · M4 · 22 engineers, 3 squads*` | `Globex` ✅ | `Toronto, Canada` ✅ |
| e | `Title · Team · Globex` | `*01/2024 – 12/2024 · M4 · 22 engineers, 3 squads*` | **`M4`** ❌ | `undefined` |

- **a vs b** — identical title line; adding tokens to the date line breaks it.
- **b vs d** — identical date line; moving the company to a middle segment fixes it.
- **e** — fails even with no location token at all, so this is not location-parsing collateral.

A date sub-line carrying extra tokens is harmless on its own: with the company in a middle segment (row d), `date · M4`, `date · 22 engineers`, and `date · M4 · 22 engineers, 3 squads` all parse clean, italic or not.

## Relation to #347

#347 (closed) covered the two-column Google-Docs PDF path where a `Date · Location` sub-line mis-split, and listed this acceptance criterion:

> Mid-dot is treated as a separator on the date sub-line, never absorbed into a field value

That invariant does not hold here. This is a different path (markdown-heading, single-column) and a different trigger (the absorbed token is *not* a location — `M4` is absorbed as `company`), so it may be a gap in #347's fix rather than a fresh class. Worth deciding which before fixing.

## Why it matters

`Title · Team · Company, City` is a normal header order, and a date line reading `dates · level · headcount` is a natural way to record role scope. The failure is silent: `confidence=0.90`, `triggers=[]`, no diagnostic.

## Repro

`runCascadeFromMarkdown(rawText, markdown)`, where `rawText` is the markdown with `#`/`**`/leading `- `/wrapping `*` stripped. Read `result.canonical.fields.experience[0]`.

## Acceptance criteria

- [x] Row b parses `company="Globex"`, `location="Toronto"`
- [x] Row c parses `company="Globex"`, `location="Toronto, Canada"`
- [x] Row e parses `company="Globex"`
- [x] Rows a and d unchanged
- [x] The date sub-line is never used as a company source when the title line already carries one
- [x] Regression test covers the trailing-company-segment + extra-date-token combination

Vaishnavi1709:
## Summary

A middot experience header whose company sat in the TRAILING `·` segment AND whose date sub-line carried extra tokens after the date range (e.g. `01/2024 – 12/2024 · M4 · 22 engineers, 3 squads`) had the parser scavenge the date sub-line: `M4` landed in `company`, the real company on the title line was silently discarded, and `location` dropped to `undefined`. Both conditions had to align — the same date-line shape parsed fine with the company in a middle segment (row d), and the same trailing-segment company parsed fine with a bare date line (row a).

Two-part narrowing in `experience-disambiguate.ts`:

1. **`anchorHasReconstructedSignature` now also requires `!aboveHasMiddotSegments`.** The signal exists to detect our own reconstructed export's `Company · Location  Dates` sub-line — designed for the shape where the title above is BARE. When the title line is itself middot-cleaved into segments, it already carries its own field mapping and the anchor line's middot is data noise (role scope), not the export signature. The else-if branch that read the first anchor-line token as the company therefore no longer fires when the title line has its own mapping to draw from.
2. **Middot exception rotates when middle=title-shaped and trailing=non-title.** A three-segment middot header reads as `Title · Team · Company` in this shape (rows a/b/c/e), the mirror of `Title · Company · Team` that the exporter emits and that `mapTitleFirst` already handles for the reverse ordering.

Row a already worked because a bare date line left the anchor residue empty and `rescueTeamLocation`'s rotate folded the trailing `Company, City` into `company` from the team slot. Rows b, c, and e now reach the same end state; row d unchanged.

Closes #614

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — full heuristics (1123) + pdf (318) suites pass; new file adds 5 tests covering rows a/b/c/d/e of the issue's condition matrix
- [x] `npm run verify` green via pre-push hook (fallow findings are inherited/report-only)
- [x] Manually verified in `npm run dev` — dropped a synthetic PDF holding all five header shapes; every reconstructed role shows `company="Globex"`, no `M4`/`M3`/`M2` or `N engineers` tokens anywhere in the render
- [x] No fixture binary added or changed — nothing for the fixture-PII surface

## Provenance

### DIFF
--- a/src/lib/heuristics/extract/experience-disambiguate.ts
+++ b/src/lib/heuristics/extract/experience-disambiguate.ts
@@ -595,11 +595,47 @@ function mapSegmentsToFields(
   // keyword) has no such corroboration, so we fall through to the old default
   // (company = first line) — the only case this narrows, and one no corpus fixture
   // exercises, so every round-trip that relied on the tiebreak stays intact.
+  //
+  // #614 gate — the reconstructed-export shape emits a BARE title on the line
+  // above (no middot delim), so if a line above the anchor already carries its
+  // own company via the `Title · Team · Company` middot shape (rows a/b/c/e of
+  // the issue's condition matrix), the anchor line's middot is not the export
+  // signature at all. It is data noise on the date sub-line — a role scope
+  // tail like `dates · L7 · 18 engineers, 2 TLMs`, whose post-`stripDateRange`
+  // residue `L7 · 18 engineers, ...` also trips {@link anchorCarriesOrgSignal}.
+  // Without this narrowing, `L7` was picked as the company via the else-if
+  // branch below, silently discarding the real trailing-segment company on
+  // the title line and dropping location.
+  //
+  // Gated tightly on the SHAPE that carries a trailing company (#639 review) —
+  // 3 same-source middot segments where the middle is title-keyword-shaped and
+  // the trailing is not-a-title AND not-a-bare-location. This is the same
+  // predicate the middot exception's rotate below uses to detect
+  // `Title · Team · Company`. A 2-segment `Title · Team` above line carries NO
+  // company, so leaving the anchor-signature branch ON there is correct
+  // (`Relationship Banking · Retail` over `Northern Trust · Chicago  Dates`
+  // must still resolve to `company=Northern Trust`, matching `main`).
+  const aboveCarriesTrailingCompany =
+    anchorIdx !== undefined &&
+    [...new Set(splits.filter((s) => s.source < anchorIdx).map((s) => s.source))].some(
+      (src) => {
+        const segs = splits.filter((s) => s.source === src && s.middot);
+        if (segs.length !== 3) return false;
+        const s1 = segs[1];
+        const s2 = segs[2];
+        return (
+          looksLikeTitle(s1.text) &&
+          !looksLikeTitle(s2.text) &&
+          !isBareLocationString(s2.text)
+        );
+      },
+    );
   const anchorHasReconstructedSignature =
     anchorIdx !== undefined &&
     anchorIdx >= 0 &&
     anchorIdx < filtered.length &&
-    anchorCarriesOrgSignal(filtered[anchorIdx]);
+    anchorCarriesOrgSignal(filtered[anchorIdx]) &&
+    !aboveCarriesTrailingCompany;
 
   // Choose which split is the company. A single company-suffix match is a
   // decisive content signal (rule 1). When MULTIPLE splits look like a company,
@@ -886,6 +922,49 @@ function mapWithoutCompanyMatch(
     splits[1] !== undefined &&
     splits[0].source === splits[1].source
   ) {
+    // #614 — a three-segment middot header where the MIDDLE segment is
+    // title-keyword-shaped ("Site Lead, Enterprise Platforms") and the
+    // TRAILING segment is not ("Google") reads as the OTHER canonical ordering:
+    // `Title · Team · Company` (rows a/b/c/e of the issue), not `Title ·
+    // Company · Team`. Detect the ordering by the title-keyword split:
+    // `secondLooksTitle && !thirdLooksTitle` → the trailing segment is the
+    // company. Two-segment headers and neutral-vs-neutral three-segment
+    // headers keep the exporter's `Title · Company [· Team]` default.
+    // Symmetric to `mapTitleFirst`'s handling of the reverse ordering.
+    //
+    // Additional guard on `!isBareLocationString(s2)` (#639 review): a segment
+    // that is entirely a bare location ("New York, NY", "Hyderabad, India")
+    // fails `looksLikeTitle` too — but it is the LOCATION, not the company.
+    // Promoting it here would shred it: `recoverLocation`'s `stripLocationSuffix`
+    // then peels its own tail as location, so `New York, NY` → company="New" /
+    // location="York, NY". That shape parses correctly on `main` via
+    // `rescueTeamLocation`'s rotate (the trailing bare-location segment lands
+    // in `team`, gets recognised as a whole-string location, and clears with
+    // no company promotion). Keeping this rotate off the bare-location shape
+    // lets that existing path do its job — same predicate the file already
+    // uses at :383 for `splitRoleComma`.
+    //
+    // KNOWN TRADEOFF (#639 review, documented not fixed): `isBareLocationString`
+    // is closed-vocab, so a trailing single-token city OUTSIDE `BARE_LOCATION_RE`
+    // ("Pune", "Zurich") with no state/country tail still promotes to `company`
+    // via this rotate. Both readings (this branch and the older team-swallowed
+    // shape on `main`) are wrong, so this is no regression against `main`; a
+    // broader fix wants an open-vocab city predicate, which is a separate change.
+    const s2 = splits[2];
+    const trailingIsCompany =
+      s2 !== undefined &&
+      s2.source === splits[0].source &&
+      s2.middot &&
+      looksLikeTitle(splits[1].text) &&
+      !looksLikeTitle(s2.text) &&
+      !isBareLocationString(s2.text);
+    if (trailingIsCompany) {
+      return {
+        title: splits[0].text,
+        team: splits[1].text,
+        company: s2.text,
+      };
+    }
     return {
       title: splits[0].text,
       company: splits[1].text,
--- a/src/lib/heuristics/extract/experience.date-subline-extra-tokens.test.ts
+++ b/src/lib/heuristics/extract/experience.date-subline-extra-tokens.test.ts
@@ -1,0 +1,206 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * Regression tests for #614 — a middot experience header whose COMPANY sits in
+ * the trailing `·` segment AND whose date sub-line carries EXTRA `·`-separated
+ * tokens after the date range (e.g. `dates · level · headcount`) had the
+ * `anchorHasReconstructedSignature` branch of `mapWithoutCompanyMatch`
+ * misfire. That branch is designed for our own reconstructed export shape
+ * (`Title \n Company · Location  Dates`), where the anchor line IS the
+ * company sub-line and its middot marker is our export signature. The gate
+ * (`anchorCarriesOrgSignal`) checked only for a `·` on the anchor line,
+ * without distinguishing the export's own signature from a real-résumé date
+ * sub-line whose post-strip residue happens to contain middots — so a date
+ * line like `01/2024 – 12/2024 · L7 · 18 engineers, 2 TLMs` reduced to
+ * `L7 · 18 engineers, 2 TLMs`, tripped the check, and the branch took the
+ * first anchor-line token (`L7`) as the company — silently discarding the
+ * real company on the title line above and dropping the location.
+ *
+ * The fix narrows the gate: the reconstructed-export shape emits a BARE title
+ * above (no middot delim), so a middot-split title line above the anchor
+ * already carries its own field mapping and the anchor line's middot is data
+ * noise, not the export signature. The condition matrix from the issue:
+ *
+ *   a — trailing company + BARE date line               → OK (baseline)
+ *   b — trailing company + date line with extra tokens  → FAIL pre-fix
+ *   c — b, with location gaining a country suffix       → FAIL pre-fix
+ *   d — MIDDLE-segment company + extra date tokens      → OK
+ *   e — trailing company (bare, no location) + extras   → FAIL pre-fix
+ *
+ * Synthetic personas only, per the fixtures PII policy.
+ */
+
+import { describe, it, expect } from "vitest";
+import { groupIntoLines, splitIntoSections, findSection } from "../sections.ts";
+import { extractExperience } from "../extract-fields.ts";
+import { mkItems } from "../__test-utils__/mkItem.ts";
+
+function roleFromSection(specs: Array<{ text: string; fontSize?: number }>) {
+  const sections = splitIntoSections(groupIntoLines(mkItems(specs)));
+  const experience = findSection(sections, "experience");
+  expect(experience).toBeDefined();
+  return extractExperience(experience).value;
+}
+
+describe("date sub-line extra tokens don't hijack company (#614)", () => {
+  it("row (a) baseline — bare date line, trailing-segment company splits cleanly", () => {
+    const roles = roleFromSection([
+      { text: "EXPERIENCE", fontSize: 13 },
+      {
+        text: "Sr. Engineering Manager · Site Lead, Enterprise Platforms · Google, Hyderabad",
+        fontSize: 11,
+      },
+      { text: "01/2024 – 12/2024", fontSize: 11 },
+      { text: "• Built an 18-engineer org in under 6 months.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    const role = roles[0];
+
+    expect(role.company).toBe("Google");
+    expect(role.location).toBe("Hyderabad");
+    expect(role.team).toBe("Site Lead, Enterprise Platforms");
+  });
+
+  it("row (b) — extra `·`-tokens on the date line MUST NOT become the company", () => {
+    const roles = roleFromSection([
+      { text: "EXPERIENCE", fontSize: 13 },
+      {
+        text: "Sr. Engineering Manager · Site Lead, Enterprise Platforms · Google, Hyderabad",
+        fontSize: 11,
+      },
+      { text: "01/2024 – 12/2024 · L7 · 18 engineers, 2 TLMs", fontSize: 11 },
+      { text: "• Built an 18-engineer org in under 6 months.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    const role = roles[0];
+
+    expect(role.company).toBe("Google");
+    expect(role.location).toBe("Hyderabad");
+    expect(role.team).toBe("Site Lead, Enterprise Platforms");
+    // The date-line data token must not appear in any parsed field.
+    expect(role.company).not.toBe("L7");
+    expect(role.team).not.toContain("L7");
+  });
+
+  it("row (c) — same as (b) with a country-suffixed location", () => {
+    const roles = roleFromSection([
+      { text: "EXPERIENCE", fontSize: 13 },
+      {
+        text: "Sr. Engineering Manager · Site Lead, Enterprise Platforms · Google, Hyderabad, India",
+        fontSize: 11,
+      },
+      { text: "01/2024 – 12/2024 · L7 · 18 engineers, 2 TLMs", fontSize: 11 },
+      { text: "• Built an 18-engineer org in under 6 months.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    const role = roles[0];
+
+    expect(role.company).toBe("Google");
+    expect(role.location).toBe("Hyderabad, India");
+    expect(role.team).toBe("Site Lead, Enterprise Platforms");
+  });
+
+  it("row (d) unchanged — MIDDLE-segment company with extra date tokens still parses", () => {
+    const roles = roleFromSection([
+      { text: "EXPERIENCE", fontSize: 13 },
+      {
+        text: "Sr. Engineering Manager · Google, Hyderabad, India · Site Lead, Enterprise Platforms",
+        fontSize: 11,
+      },
+      { text: "01/2024 – 12/2024 · L7 · 18 engineers, 2 TLMs", fontSize: 11 },
+      { text: "• Built an 18-engineer org in under 6 months.", fontSize: 11 },
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    const role = roles[0];
+
+    expect(role.company).toBe("Google");
+    expect(role.location).toBe("Hyderabad, India");
+    expect(role.team).toBe("Site Lead, Enterprise Platforms");
+  });
+
+  it("row (e) — trailing-segment BARE company (no location) survives extra date tokens", () => {
+    const roles = roleFromSection([
+      { text: "EXPERIENCE", fontSize: 13 },
+      {
+        text: "Sr. Engineering Manager · Site Lead, Enterprise Platforms · Google",
+        fontSize: 11,
+      },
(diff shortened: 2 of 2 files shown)


src/lib/heuristics/extract/experience-disambiguate.ts
```diff
@@ -886,6 +900,29 @@ function mapWithoutCompanyMatch(
...
+    // company. Two-segment headers and neutral-vs-neutral three-segment
+    // headers keep the exporter's `Title · Company [· Team]` default.
+    // Symmetric to `mapTitleFirst`'s handling of the reverse ordering.
+    const s2 = splits[2];
+    const trailingIsCompany =
+      s2 !== undefined &&
+      s2.source === splits[0].source &&
+      s2.middot &&
+      looksLikeTitle(splits[1].text) &&
+      !looksLikeTitle(s2.text);
```
s-annam: **Blocking — a trailing bare LOCATION gets promoted to `company` here, and the location is mangled with it.**

`isBareLocationString` is not `looksLikeTitle`, so `New York, NY` / `Bengaluru, India` pass both title checks and rotate into the company slot. This fires on a **bare** date line too, so it is outside #614's condition matrix — and those shapes parse correctly on `main`:

| Header (`01/2024 – 12/2024`, bare) | main | this branch |
|---|---|---|
| `Engineer · Team Lead · New York, NY` | `company="Team Lead"`, `location="New York, NY"` | `company="New"`, `location="York, NY"` |
| `Sr. Engineering Manager · Site Lead, Payments Platform · Bengaluru, India` | `company="Site Lead, …"`, `location="Bengaluru, India"` | `company="Bengaluru"`, `location="India"` |

Reuse the predicate this file already owns (defined at :444, used by the team-rescue steps):

```suggestion
      s2.middot &&
      looksLikeTitle(splits[1].text) &&
      !looksLikeTitle(s2.text) &&
      !isBareLocationString(s2.text);
```

Verified locally: both rows return to their `main` values, the 5 new tests still pass, `heuristics` + `pdf` stay green (1440 passed).

src/lib/heuristics/extract/experience.date-subline-extra-tokens.test.ts
```diff
@@ -0,0 +1,139 @@
...
+    ]);
+    expect(roles.length).toBeGreaterThanOrEqual(1);
+    const role = roles[0];
+
+    expect(role.company).toBe("Google");
+    expect(role.location).toBe("Hyderabad, India");
+    expect(role.team).toBe("Site Lead, Enterprise Platforms");
+  });
+
+  it("row (e) — trailing-segment BARE company (no location) survives extra date tokens", () => {
```
s-annam, line 122: **Secondary — row (e) is the only test touching the new rotate, and it only proves the rotate *fires*.** Nothing pins where it must *not* fire, which is why the trailing-location regression flagged above cleared a green suite. Add the negative case alongside the guard:

```ts
it("a trailing bare LOCATION is not promoted to company", () => {
  const roles = roleFromSection([
    { text: "EXPERIENCE", fontSize: 13 },
    { text: "Engineer · Team Lead · New York, NY", fontSize: 11 },
    { text: "01/2024 – 12/2024", fontSize: 11 },
    { text: "• Built an 18-engineer org in under 6 months.", fontSize: 11 },
  ]);
  expect(roles[0].location).toBe("New York, NY");
  expect(roles[0].company).not.toBe("New");
});
```

s-annam (changes requested):
Root cause is right and part 1 is well-targeted — I proved both halves load-bearing by reverting each: gate 1 alone fixes rows b/c and touches only the extra-date-token rows; row (e) needs part 2. The issue's own repro path (`runCascadeFromMarkdown`) now yields `company="Contoso"`, `location="Bengaluru"` (main yields `company="L6"`, no location). But part 2 has no location guard, and that costs a shape the parser gets right today.

## Blocking

**1 — `trailingIsCompany` promotes a trailing BARE LOCATION into `company`, and mangles the location doing it.** `src/lib/heuristics/extract/experience-disambiguate.ts:913` (inline)

The new rotate fires on any three-segment middot header where the middle segment is title-shaped and the trailing one isn't — and a bare location (`New York, NY`, `Bengaluru, India`) is not title-shaped. `Title · Team · City, ST` is a normal header, it parses its location correctly on `main`, and it does not need the extra-token date line to break — the **bare** date line regresses too, so this is outside the issue's condition matrix entirely.

A/B, same input, `main` vs this branch:

| Header (date line `01/2024 – 12/2024`, bare) | main | this branch |
|---|---|---|
| `Engineer · Team Lead · New York, NY` | `company="Team Lead"`, `location="New York, NY"` | `company="New"`, `location="York, NY"` |
| `Sr. Engineering Manager · Site Lead, Payments Platform · Bengaluru, India` | `company="Site Lead, …"`, `location="Bengaluru, India"` | `company="Bengaluru"`, `location="India"` |

`company` becomes a location fragment and the location is cleaved mid-name — strictly worse than main, which at least kept the location intact. No existing test covers it: the full `heuristics` + `pdf` suites (1440 tests) are green both before and after.

Fix — one clause, reusing the predicate this file already owns (`isBareLocationString`, defined at :444 and used by the team-rescue steps):

```diff
       s2.middot &&
       looksLikeTitle(splits[1].text) &&
-      !looksLikeTitle(s2.text);
+      !looksLikeTitle(s2.text) &&
+      !isBareLocationString(s2.text);
```

Verified locally: both regression rows above return to their `main` values, all 5 new tests still pass, and `heuristics` + `pdf` stay green (1440 passed).

## Secondary

**2 — The second change ships with no adjacent-shape test.** `src/lib/heuristics/extract/experience.date-subline-extra-tokens.test.ts:122` (inline)

Row (e) is the only case exercising the rotate, and it only proves the rotate *fires*. Nothing proves where it must *not* fire — which is exactly why the location regression above passed CI.

## Nits (non-blocking)

**3 — Tests exercise the item/line path, not the issue's stated markdown path.** The issue repros via `runCascadeFromMarkdown`; the new file builds items with `mkItems`. I checked the markdown path directly and it is fixed, so this is coverage shape only — but a one-line `runCascadeFromMarkdown` case would pin the surface the issue actually names.

**4 — `aboveHasMiddotSegments` (:606) restates the shape of `hasAbove` (:581)** — `anchorIdx !== undefined && splits.some(s => s.source < anchorIdx && …)`. Could read as `hasAbove && splits.some(…)`. Cosmetic.

**5 — fallow: `mapWithoutCompanyMatch` is now 32 cyclomatic / CRAP 32 (HIGH).** Report-only inside `verify`, and the function was already heavy — flagging, not blocking.

## Gates

- Fixture PII — **skipped, nothing to check**: no binary added or changed.
- Design-system / style tokens / reuse — **N/A**: no `src/components/**`, no styling.
- Skill/script command review — **N/A**: no `.claude/**` or `scripts/**`.
- fallow — **ran** (finding 5). Tests — **ran**, `heuristics` + `pdf` green on the branch.
- Non-vacuity — **proven**: reverting the src change fails rows b/c/e (`expected 'L6' to be 'Contoso'`).

Reviewed at `be6f153`.

---

src/lib/heuristics/extract/experience-disambiguate.ts
Vaishnavi1709: Fixed in `a628071` — added `!isBareLocationString(s2.text)` as the fourth conjunct on the rotate check, exactly the suggestion. Confirmed the regression you reproduced against `main`: without the guard, `Engineer · Team Lead · New York, NY` (bare date) came out as `company="New"` / `location="York, NY"`, and `… · Site Lead, Enterprise Platforms · Hyderabad, India` came out `company="Hyderabad"` / `location="India"`. Both back to their `main` values with the guard on. The rationale is now on the code (comment above the rotate, calling out the shape `rescueTeamLocation` handles and why this branch stays off it) so a future reader can see why the four checks are all needed. Full suite: 1442 passed / 8 skipped.

src/lib/heuristics/extract/experience.date-subline-extra-tokens.test.ts
Vaishnavi1709: Added in `a628071`, plus a sibling row for the `City, Country` shape so both branches of `isBareLocationString` are pinned (US-state code and gazetteer country):

```ts
it("trailing BARE-LOCATION segment is not promoted to company (bare date)", …)
it("trailing BARE-LOCATION `City, Country` is not promoted to company either", …)
```

Both use bare-date lines so they exercise the exact shape that regressed on the first-pass fix (i.e. outside #614's condition matrix), and both assert both `location` recovery and the negative on `company` per your template. Suite now runs 7 tests, all pass.

src/lib/heuristics/extract/experience-disambiguate.ts
```diff
@@ -595,11 +595,25 @@ function mapSegmentsToFields(
...
+  const aboveHasMiddotSegments =
+    anchorIdx !== undefined &&
+    splits.some((s) => s.source < anchorIdx && s.middot);
   const anchorHasReconstructedSignature =
     anchorIdx !== undefined &&
     anchorIdx >= 0 &&
     anchorIdx < filtered.length &&
-    anchorCarriesOrgSignal(filtered[anchorIdx]);
+    anchorCarriesOrgSignal(filtered[anchorIdx]) &&
+    !aboveHasMiddotSegments;
```
s-annam: **Blocking — this narrowing is wider than the shape it was written for.**

The comment's premise holds only when the above line has 3 segments. A two-segment `Title · Team` line carries NO company, so switching the anchor-is-company branch off strands the real company on the anchor line and the team takes the `company` slot.

A/B on `Relationship Banking · Retail` over `Northern Trust · Chicago  Jan 2019 – Mar 2021`:

- `origin/main` → company=`Northern Trust`, team=`Retail`, location=`Chicago` ✅
- this branch → company=**`Retail`**, team=**`Northern Trust`**, location=`Chicago` ❌

5/5 realistic stacked headers regress this way (full table in the review body). A bare title line above is unchanged — that is the case this comment reasons about.

Fix direction: gate on the above line actually CARRYING a company (reuse the `trailingIsCompany` predicate from the rotate below — 3 segments, middle title-shaped, trailing non-title/non-bare-location), not on it merely having a middot. `filtered[anchorIdx]` is already date-stripped by `parseEntryBlocks`, so middot-position-relative-to-date is not available here.

No test pins this half of part 1 — please add the two-segment stacked case.

src/lib/heuristics/extract/experience-disambiguate.ts
```diff
@@ -886,6 +900,42 @@ function mapWithoutCompanyMatch(
...
+    // lets that existing path do its job — same predicate the file already
+    // uses at :383 for `splitRoleComma`.
+    const s2 = splits[2];
+    const trailingIsCompany =
+      s2 !== undefined &&
+      s2.source === splits[0].source &&
+      s2.middot &&
+      looksLikeTitle(splits[1].text) &&
+      !looksLikeTitle(s2.text) &&
+      !isBareLocationString(s2.text);
```
s-annam, line 960: **Secondary — closed-vocab limit on the guard shifts the failure mode on bare-date headers.**

A trailing city outside `BARE_LOCATION_RE` with no state/country tail now promotes to `company` on a plain bare-date header (outside #614's matrix):

- `Engineer · Team Lead · Pune` — `main`: company=`Team Lead`, team=`Pune` → here: company=**`Pune`**, team=`Team Lead`
- `Engineer · Team Lead · Zurich` — same flip

Both readings are wrong, so not worse than `main`, and every in-vocab shape I tried is correct. Recording it as a known tradeoff rather than asking for a change.

s-annam (changes requested):
Part 2 (the `Title · Team · Company` rotate) is now correct and well-guarded — the
`!isBareLocationString` fix from the last round holds under A/B. Part 1
(`aboveHasMiddotSegments`) is too broad: it disables the anchor-is-company branch
for the whole stacked-header class, not just the date-noise shape, and swaps
`company`↔`team` on every `Title · Team` over `Company · Location  Dates` résumé.
All 3534 tests stay green through it.

Method: checked out `a628071`, ran the parser on a header matrix under both
`origin/main` and this branch, and diffed the outputs.

## Blocking

**`experience-disambiguate.ts:608-616` — `aboveHasMiddotSegments` regresses the stacked `Title · Team` / `Company · Location  Dates` shape.**

The premise in the comment and in `## Summary` item 1 — "a title line above the
anchor that was itself middot-cleaved into segments already carries its own field
mapping" — is false when the above line has only TWO segments. `Title · Team`
carries no company, so suppressing the anchor-is-company branch leaves the real
company on the anchor line unclaimed, and the title line's team takes the
`company` slot instead.

A/B, identical input, `origin/main` → this branch:

| Header line / anchor line | `main` | this branch |
|---|---|---|
| `Relationship Banking · Retail` / `Northern Trust · Chicago  Jan 2019 – Mar 2021` | company=`Northern Trust`, team=`Retail` | company=**`Retail`**, team=**`Northern Trust`** |
| `Consumer Banking · Wealth` / `Fifth Third Bank · Cincinnati, OH  Jan 2019 – Mar 2021` | company=`Fifth Third Bank`, team=`Wealth` | company=**`Wealth`**, team=**`Fifth Third Bank`** |
| `Equity Research · Industrials` / `Baird · Milwaukee, WI  2019 – 2021` | company=`Baird`, team=`Industrials` | company=**`Industrials`**, team=**`Baird`** |
| `Brand Marketing · Footwear` / `Northwind Retail · Portland  2019 – 2021` | company=`Northwind Retail`, team=`Footwear` | company=**`Footwear`**, team=**`Northwind Retail`** |
| `Clinical Operations · Oncology` / `Initech Health · Boston  2019 – 2021` | company=`Initech Health`, team=`Oncology` | company=**`Oncology`**, team=**`Initech Health`** |

5/5 regress. The control — a BARE title line above (`Relationship Banking` /
`Northern Trust · Chicago  Dates`) — is unchanged, which is exactly the shape the
comment reasons about. The gate was written for that case and then applied to the
whole class.

This class is invisible to the suite: nothing in the 3534 tests pins a
middot-split title line over a middot-split anchor line, and no corpus fixture
covers it — so `npm run test` green is not evidence here. Both review threads from
the last round were about part 2; part 1 has never had a negative test.

Direction for the fix — the issue's own AC states the condition precisely: *"the
date sub-line is never used as a company source **when the title line already
carries one**."* So gate the suppression on the above line actually carrying a
company, not merely on it having a middot — e.g. require the above line's middot
splits to satisfy the same `trailingIsCompany` predicate part 2 already computes
(3 segments, middle title-shaped, trailing non-title / non-bare-location). Under
that gate rows b/c/e still suppress (3-segment title line with a trailing company)
and the five rows above keep `main`'s mapping (2-segment title line, no company on
it). Note `filtered[anchorIdx]` is already date-stripped by `parseEntryBlocks`, so
"is the middot before or after the date range" is not available at this point —
the segment-shape test is the workable discriminator.

Please add a regression test for the two-segment stacked shape alongside the
existing ones; it is the half of part 1 nothing currently pins.

## Secondary

**`experience-disambiguate.ts:924-931` — the rotate changes bare-date headers whose trailing city is outside the closed vocab.**

`isBareLocationString` is closed-vocabulary by design, so a trailing segment that
is a real city not in `BARE_LOCATION_RE` and has no state/country tail now gets
promoted to `company` on a plain bare-date header — a shape outside #614's matrix:

- `Engineer · Team Lead · Pune` (bare date): `main` company=`Team Lead`, team=`Pune` → this branch company=**`Pune`**, team=`Team Lead`
- `Engineer · Team Lead · Zurich` (bare date): same flip

Both readings are wrong, so this is not strictly worse than `main`, and the
in-vocab shapes (`New York, NY`, `Hyderabad, India`, `Boston`, `Remote`,
`Austin, TX`, `Ann Arbor, MI`) are all correct — the guard does its job. Flagging
it so the closed-vocab limit is a recorded tradeoff rather than a surprise: the
failure mode moved from "team in company" to "city in company".

**Date-line noise still leaks into `team` on two-segment headers.**

`Engineer · Contoso` + `01/2024 – 12/2024 · L6 · 24 engineers, 3 TLMs` yields
team=`L6` on both `main` and this branch. The AC only requires the date sub-line
never be a *company* source, so this doesn't block — but the issue's framing
("mid-dot is treated as a separator on the date sub-line, never absorbed into a
field value") isn't fully satisfied and `Closes #614` will close it. Worth a
follow-up issue, or a sentence in the PR body saying it's deliberately out of scope.

## Nits

- **Test plan says "new file adds 5 tests"** — it adds 7 (the two
  bare-location negatives from the last review round). Body wasn't refreshed after
  the revision.
- **`fallow`**: `mapWithoutCompanyMatch` grows to 128 lines / 33 cyclomatic / CRAP
  33. Report-only inside `verify` and inherited, so non-blocking, but the function
  is now four stacked special cases deep.

## Gates run

`npm run test` (3534 pass), `npm run typecheck` clean, `npm run lint` clean,
`npx fallow audit --base origin/main` (4 pre-existing complexity findings).
Fixture-PII gate skipped — no fixture binary touched. Design-system / style-token
gates skipped — no `src/components/**` changes. Head reviewed: `a628071`.

---

src/lib/heuristics/extract/experience-disambiguate.ts
Vaishnavi1709: Fixed in `860d3b8`. You're right — the reasoning was for the 3-segment shape but the check fired on any middot above, including 2-segment `Title · Team` lines that carry no company. Verified the regression against this branch first (`company="Retail"` for your Northern Trust A/B), then narrowed the gate to the exact shape:

```ts
const aboveCarriesTrailingCompany =
  anchorIdx !== undefined &&
  [...new Set(splits.filter((s) => s.source < anchorIdx).map((s) => s.source))].some(
    (src) => {
      const segs = splits.filter((s) => s.source === src && s.middot);
      if (segs.length !== 3) return false;
      const s1 = segs[1];
      const s2 = segs[2];
      return (
        looksLikeTitle(s1.text) &&
        !looksLikeTitle(s2.text) &&
        !isBareLocationString(s2.text)
      );
    },
  );
```

Same conjuncts as the middot exception's rotate below, applied per above-source (so a mixed-source above still resolves cleanly). A 2-segment `Title · Team` above line now falls through to the anchor-signature branch as `main` does — Northern Trust A/B returns to `company=Northern Trust`, `team=Retail`, `location=Chicago`, verified via the new regression test:

```
✓ 2-segment `Title · Team` above with anchor-line company must NOT flip to `company=Team`
```

Full suite: 1443 passed / 8 skipped.

src/lib/heuristics/extract/experience-disambiguate.ts
Vaishnavi1709: Thanks for the A/B and the note. Documented in the rotate's docblock in `860d3b8` ("KNOWN TRADEOFF (#639 review, documented not fixed)") so a future reader sees the shape and the reason the guard is closed-vocab, and knows an open-vocab city predicate is the direction if it ever needs a broader fix. Not resolving on my end since it's your call — happy to file a follow-up issue for the open-vocab consideration if you'd like a tracker.

s-annam (approved):
Re-checked at `860d3b8` (force-pushed after my review of `a628071`).

**The Blocking finding is resolved.** The gate is now keyed on the 3-segment `Title · Team · Company` shape — `looksLikeTitle(s1) && !looksLikeTitle(s2) && !isBareLocationString(s2)`, the same predicate the rotate below already computes — rather than on the bare presence of a middot on the line above. The new 2-segment regression test pins the case that was silently flipping.

I re-ran the same A/B probe against `origin/main` across seven header shapes. The only output that differs from `main` is the row this PR is about:

| Header line / anchor line | `main` | `860d3b8` |
|---|---|---|
| `Sr. Engineering Manager · Site Lead, Payments Platform · Contoso, Bengaluru` / `01/2024 – 12/2024 · L6 · 24 engineers, 3 TLMs` | company=`L6`, location empty | company=`Contoso`, location=`Bengaluru` |

The five rows my review flagged (`Northern Trust`, `Fifth Third Bank`, `Baird`, `Northwind Retail`, `Initech Health` over 2-segment `Title · Team` lines) now match `main` exactly, and so do the bare-title export shape and the bare-date 3-segment shape. Heuristics suite: 1125 pass.

Approving under the verdict rule — 0 Blocking remaining. The two Secondary findings and the Nit are unchanged and stay non-blocking:

- The closed-vocab city tradeoff (`Pune`, `Zurich` → `company`) is now recorded in-code as a `KNOWN TRADEOFF` comment. Confirmed identical on `main`, so no regression either way.
- Date-line noise still lands in `team` on two-segment headers (`Engineer · Contoso` + `… · L6 · 24 engineers` → team=`L6`), same on `main`. `Closes #614` will close the issue with that half of its framing unmet — a sentence in the body saying it's deliberately out of scope, or a follow-up issue, would keep the record straight.
- Test plan still says "new file adds 5 tests"; it now adds 8.

None of those need another round. Merge whenever you're ready.

---

PR #642, opened 2026-07-28
chore(skills): shrink the pr-ready ping, make pr-review adversarial, let revise-pr import feedback

s-annam:
## Summary

Four coupled changes to the review loop. `pr-ready`'s ping was long enough that reviewers skimmed it, and the review pointers it carried biased the review they were meant to help. Its ack accepted "a reaction *or* a reply *or* any comment" — so silence and mid-review were indistinguishable, and a second person with approval rights could not tell whether merging would cut across someone's work.

- **`open-pr`** — optional `## Review focus` in the PR body: ≤4 entries, phrased as questions, omitted when the risk is obvious. Review pointers now live where the reviewer already is.
- **`pr-ready`** — ping cut to two lines per PR + mentions + merge time + one asked-for signal (👍). Ack becomes four states: `ACKED` suppresses the reminder but does **not** move the deadline (a promise is not evidence of reading); `REVIEWING` earns one `hold_grace_minutes` window measured from the claim, never renewed; a lapsed window reports `HELD BUT STALE` with the claimant's login.
- **`pr-review`** — issue → code → description, in that order. The linked issue's ACs are the spec; the body is read **last** and audited in new gate 3f (an unbacked claim, or an unimplemented AC under `Closes`, is Blocking). New Step 0.6 posts a 👀 reaction, which is what `pr-ready` reads as the "review in progress" signal.
- **`revise-pr`** — `--from-pr` / `--from-comment` / `--notes` import feedback from elsewhere; the code always lands on the target. New Step 2.5 re-verifies each import against this branch before anything changes.

No issue — process-surface change, filed straight as a PR.

## Review focus

- `.claude/skills/pr-ready/SKILL.md` (ack predicate, input 3) — the 👀 reaction is never removed, so a stale one must not read as a live hold. Is `created_at > PING_ISO` actually applied on **every** path that reaches input 3, including the Phase 2 adopted-ping branch?
- `.claude/skills/pr-ready/SKILL.md` (Phase 5.1) — grace is measured from the claim, not from now. Does the `-gt "$(date +%s)"` guard skip the wait correctly when the window already elapsed, and is there any path that extends twice?
- `.claude/skills/revise-pr/SKILL.md` (Step 6) — the reply path is `pulls/$SRC_PR/comments`. Is there a path where the target's number reaches an imported thread's reply, or where a source thread gets resolved while its own copy of the defect is still live?
- `.claude/skills/pr-review/SKILL.md` (Step 0.5) — the whole ordering rule depends on issue *numbers* reaching the review without the description's prose. Does anything upstream of gate 3f pull the body in?

## Test plan

- [x] No `src/` changes — `git diff --stat` is five files under `.claude/`, so `typecheck` / `test` / the corpus are untouched and the suite is inert against this diff. Not claiming green boxes for gates that have nothing to run against.
- [x] Every `gh` command added was executed against the live API before being written down, not pattern-matched: the reactions endpoint shape (`content` / `created_at` / `user.type`), `--silent`, `closingIssuesReferences` (returned `629 631 632` on #630, matching the body-grep exactly), the author-excluded `jq` filter, and the `#discussion_r` URL parse.
- [x] The `date -u -j -f` (BSD) / `date -u -d` (GNU) pair round-trips an ISO-Z string to epoch.
- [x] Markdown structure linted across all four skills — code fences balanced, frontmatter intact, table rows consistent.
- [x] Not exercised end-to-end: no live `/pr-ready` run has posted the new ping, and no `/revise-pr --from-pr` run has imported a real thread. First real use is the test.

## Notes for the reviewer

One change is **not** in this diff and should be, in the sense that it matters: `ack_reactions` in the git-ignored `.claude/pr-ready.local.json` was `["eyes","white_check_mark","raising_hand"]` — none of which reviewers here actually use. Every ack was being reported as silence. It is now `["+1","thumbsup"]`, plus `hold_grace_minutes: 60`. The committed `.example` reflects both, but the live config is per-checkout and will need the same edit anywhere else it exists.

## Provenance

Verification: no `src/` changes — CI `verify` runs, but has nothing in this diff to exercise

### DIFF
--- a/.claude/skills/pr-review/SKILL.md
+++ b/.claude/skills/pr-review/SKILL.md
@@ -1,19 +1,41 @@
 ---
 name: pr-review
-description: Review an offlinecv pull request the way a maintainer does — pull the diff, run the generic /code-review correctness pass, then layer offlinecv's own gates (fixture PII, design-system reuse, style tokens, fallow dead-code, command-level bugs in skill/script files), structure the findings Blocking / Secondary / Nits, and post a PR review whose verdict matches what was found. Use when the user says "review PR #N", "/pr-review", "review this PR", or hands you a PR to critique before merge.
+description: Review an offlinecv pull request adversarially, the way a maintainer does — signal 👀 that review started, judge the diff against the linked issue's acceptance criteria, run the generic /code-review correctness pass, layer offlinecv's own gates (fixture PII, design-system reuse, style tokens, fallow dead-code, command-level bugs in skill/script files), then read the PR description LAST and audit it for claims the code doesn't back, structure the findings Blocking / Secondary / Nits, and post a PR review whose verdict matches what was found. Use when the user says "review PR #N", "/pr-review", "review this PR", or hands you a PR to critique before merge.
 argument-hint: <#|#N> [--repo owner/repo] [--post|--local] [--effort low|medium|high]
 ---
 
 # Review PR
 
 Take an offlinecv PR from "someone opened it" to "a maintainer-grade review is on
-the thread": check out the diff → run the built-in `/code-review` for the generic
-correctness pass → **layer the offlinecv-specific gates** → structure findings
+the thread": check out the diff → **signal that review started** → read the
+**issue**, not the PR's prose → run the built-in `/code-review` for the generic
+correctness pass → **layer the offlinecv-specific gates** → **then** read the PR
+description and audit it against what the code actually does → structure findings
 **Blocking / Secondary / Nits** → draft the review → confirm → post a `gh` PR
 review whose **verdict matches the findings**.
 
 This is the **reviewer-side** sibling of the author-side loop: `open-pr` creates
 the PR, `revise-pr` addresses the review — this skill *is* the review in between.
+
+## The review is adversarial — read the description LAST
+
+**The spec is the issue. The code is the evidence. The PR description is a claim
+to falsify.** That ordering is the whole stance, and it is not a formality:
+
+A description read *first* becomes the map — you check the three things it points
+at, find them fine, and approve. Every defect it failed to mention is now
+invisible, and the author's blind spots have silently become the reviewer's.
+Worse, a description that *misdescribes* the change can only be caught by someone
+who formed an independent picture of the change first; read it up front and you
+will read the diff through it, confirming rather than checking. So Steps 1–3 form
+findings from the diff and the issue alone, and Step 3f is where the description
+finally gets read — as a **hypothesis to test against what you already found**,
+never as a guide to what to look at.
+
+`## Review focus` in the body is subject to exactly the same treatment. It is the
+author's guess at where the risk is: a useful lead, and a real signal when it
+turns out to be wrong. **A review that visits only the files it names is not a
+review.**
 
 ## Why this skill exists
 
@@ -57,6 +79,96 @@ gh pr view "$PR_NUM" --repo "$REPO" \
 
 If the PR is not `OPEN`, stop and say so. Note the **author login** — it tunes the
 review stance (see the careful-review note in Rules).
+
+**Note what this call deliberately does not fetch: `body`.** The field list above
+is title/state/refs/author/files and nothing else, so the PR's prose never enters
+the review until Step 3f. Don't add `body` to it "while you're there."
+
+### Step 0.5 — Read the issue, not the PR's prose
+
+The linked issue is the spec this PR is judged against. Get the issue *numbers*
+without pulling the description's prose into the review — pipe the body straight
+through a matcher so only digits come back:
+
+```bash
+# auto-closing links, resolved by GitHub itself
+gh pr view "$PR_NUM" --repo "$REPO" --json closingIssuesReferences \
+  -q '.closingIssuesReferences[].number'
+# plus partial links GitHub does not resolve (`Refs #N`), matched out of the body
+gh pr view "$PR_NUM" --repo "$REPO" --json body -q .body \
+  | grep -oiE '\b(closes|fixes|resolves|refs?):? +#[0-9]+' | grep -oE '[0-9]+' | sort -u
+```
+
+**The `grep` is load-bearing, not decoration.** You cannot un-read prose once it
+is in the transcript, so the discipline has to be enforced at the tool boundary
+rather than by intention: piping through `grep -o` means the tool result is a list
+of numbers and the description is never rendered. Run it as written — a bare
+`--json body` here defeats the entire ordering rule three steps before Step 3f.
+
+**The `:?` covers the colon form** — `Closes: #123` is valid GitHub syntax and the
+whitespace-only pattern misses it. For the *closing* keywords that costs nothing
+in practice, since GitHub's own resolver picks them up and they arrive via
+`closingIssuesReferences` above; the two commands are belt-and-braces and their
+union covers it. The one shape that falls through **both** is `Refs: #N` — a
+non-closing keyword in colon form, which GitHub never resolves. `closes#789`
+staying unmatched is correct: GitHub doesn't accept that form either.
+
+Then read each issue in full, and write down its acceptance criteria before you
+look at any code:
+
+```bash
+for N in $ISSUE_NUMS; do gh issue view "$N" --repo "$REPO" --comments; done
+```
+
+Carry those ACs into Step 4 as a checklist. **An unimplemented AC on a PR that
+says `Closes #N` is Blocking** — the merge closes the issue, so the gap ships as
+"done" and nothing will ever re-open it. If the PR links no issue at all, say so
+in the report and review against the diff alone; don't invent a spec.
+
+### Step 0.6 — Signal that the review has started
+
+Post a 👀 reaction on the PR. Do this **now**, before the slow work, not at the
+end — its entire value is telling everyone else that this PR is being read *while*
+you read it:
+
+```bash
+gh api "repos/$REPO/issues/$PR_NUM/reactions" -f content=eyes --silent
+```
+
+No confirmation needed for this one write (it is the exception to the
+confirm-before-posting rule): it is a single reversible reaction, it notifies
+nobody, and it carries no prose that could be wrong. Say you posted it in the
+Step 7 report.
+
+`issues/` is correct — reactions on a PR's top-level body go through the issues
+endpoint; `pulls/$PR_NUM/reactions` does not exist and 404s. Re-reacting is
+idempotent (GitHub returns the existing reaction), so a re-run of this skill will
+not double-post or fail.
+
+**Why this reaction exists.** `pr-ready` reads it as the machine-emitted "someone
+is actually reviewing this" signal, and extends its deadline by a grace window
+when it sees one. Nothing else on GitHub carries that information: a *pending*
+review is invisible to everyone but its author, and the alternatives that are
+visible (a `COMMENT` review, a label, a self-assign) all cost a notification or a
+repo-config change to say the same thing. Leave the reaction in place when the
+review is submitted — it is then a permanent marker of who read this PR, and
+`pr-ready` only counts reactions newer than its own ping, so a stale one from an
+earlier cycle can never read as a live hold.
+
+A review abandoned *within* a cycle is the case that filter doesn't cover — Step
+0.6 posts 👀 before the slow work, so a run that dies, or one where the user
+declines to post at Step 6, leaves a reaction genuinely newer than the ping that
+does read as `REVIEWING`. That is bounded rather than unhandled: `pr-ready` grants
+one grace window from the claim and then reports `HELD BUT STALE` with your login.
+It is also the honest argument for posting 👀 early and unconfirmed — an abandoned
+claim degrades to naming a person to nudge, which beats a reaction that is
+worthless because it landed after the review was already done.
+
+**Honest limit, so nobody over-trusts it:** 👀 does not *block* anything. `main`
+requires one approving review, and any approver can merge over a 👀. The only
+merge-blocking signal GitHub offers is `REQUEST_CHANGES`, which is far too heavy
+to mean "wait for me." This is a social hold that `pr-ready` surfaces by name, not
+enforcement.
 
 ### Step 1 — Get the diff onto disk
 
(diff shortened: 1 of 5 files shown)


.claude/skills/revise-pr/SKILL.md
```diff
@@ -1,6 +1,6 @@
 ---
 name: revise-pr
-description: Revise a pull request in response to review — check out the PR branch, fix what each unresolved thread asks for (or reply explaining why not), run the gates, commit + push to the PR branch, then reply to and resolve the threads on GitHub. Use when the user says "revise the PR", "/revise-pr", "address the review comments", "fix the review feedback", "clear the comments so it's ready to merge", or a reviewer left changes to act on.
+description: Revise a pull request in response to review — check out the PR branch, fix what each unresolved thread asks for (or reply explaining why not), run the gates, commit + push to the PR branch, then reply to and resolve the threads on GitHub. Feedback can be imported from another PR (--from-pr), a single comment URL (--from-comment), or free prose (--notes), re-verified against this branch before anything is changed. Use when the user says "revise the PR", "/revise-pr", "address the review comments", "fix the review feedback", "apply another PR's review comments here", "clear the comments so it's ready to merge", or a reviewer left changes to act on.
```
rohithgollapalli, line 3: **Secondary — this PR adds three flags to `revise-pr`, and none of them is discoverable at the prompt.**

`argument-hint` is present on two of the four skills and absent on the other two:

```
open-pr    (none)
pr-ready   argument-hint: [<pr-number>[,<pr-number>...]]
pr-review  argument-hint: <#|#N> [--repo owner/repo] [--post|--local] [--effort low|medium|high]
revise-pr  (none)
```

`open-pr` lacking one is pre-existing and takes no flags, so it is fine. `revise-pr` is different **as of this PR**: `--from-pr`, `--from-comment` and `--notes` are the feature, and the hint line is the surface where a user finds out they exist. Right now the only way to learn the flag names is to open the skill file — which is precisely the audience that does not need them.

```diff
 ---
 name: revise-pr
 description: Revise a pull request in response to review — …
+argument-hint: [<pr-number>] [--repo owner/repo] [--from-pr <n>] [--from-comment <url>] [--notes <file>]
 ---
```

The PR number is bracketed because Input says it can be inferred from the current branch, and the three source flags are documented as repeatable — worth keeping the hint honest about that if you want to spell it `[--from-pr <n>]...`.

Filing Secondary rather than Nit only because it is the discoverability surface for the change this PR exists to make; the fix is one line.

.claude/skills/revise-pr/SKILL.md
```diff
@@ -26,6 +27,44 @@ gh pr view --json number,headRefName,state -q '{n:.number,head:.headRefName,stat
...
+say so and proceed. (The reverse is also fine: a target whose threads are all
+resolved and no import flags means there is nothing to do; report that and stop
+rather than inventing work.)
+
+Extract the PR and comment ids from a `--from-comment` URL — GitHub spells a
+review-thread comment as `…/pull/<N>#discussion_r<COMMENT_ID>`:
+
+```bash
+# https://github.com/<owner>/<repo>/pull/640#discussion_r123456789
+SRC_PR="$(sed -E 's|.*/pull/([0-9]+).*|\1|' <<<"$URL")"
```
rohithgollapalli, line 76: **Nit — no match guard, so a malformed URL becomes the PR number.**

`sed s|…|\1|` leaves its input untouched when the pattern does not match, so both extractions silently pass the whole URL through. Ran it on three shapes:

| Input | `SRC_PR` | `SRC_COMMENT` |
|---|---|---|
| `…/pull/640#discussion_r123456789` | `640` | `123456789` |
| `…/pull/640#issuecomment-999` | `640` | `https://…/pull/640#issuecomment-999` |
| `https://example.com/not-a-pr` | `https://example.com/not-a-pr` | *(same)* |

`SRC_PR` then lands in `repos/$REPO/pulls/$SRC_PR/comments`, which 404s — so this **fails loudly and never writes to the wrong PR**, which is why it is a nit and not more. But the error a user sees is a 404 on a nonsense path rather than "that URL isn't a review-thread comment."

Row 2 is already handled by the `#issuecomment-` prose immediately below, but note that block runs *after* the `sed`, so the garbage `SRC_COMMENT` exists briefly and is then discarded — worth reordering the check ahead of the extraction, or guarding:

```bash
case "$URL" in
  *"/pull/"*"#discussion_r"*) ;;
  *) echo "not a review-thread comment URL: $URL" >&2; exit 1 ;;
esac
```

.claude/skills/pr-review/SKILL.md
```diff
@@ -58,6 +80,79 @@ gh pr view "$PR_NUM" --repo "$REPO" \
...
+without pulling the description's prose into the review — pipe the body straight
+through a matcher so only digits come back:
+
+```bash
+# auto-closing links, resolved by GitHub itself
+gh pr view "$PR_NUM" --repo "$REPO" --json closingIssuesReferences \
+  -q '.closingIssuesReferences[].number'
+# plus partial links GitHub does not resolve (`Refs #N`), matched out of the body
+gh pr view "$PR_NUM" --repo "$REPO" --json body -q .body \
+  | grep -oiE '\b(closes|fixes|resolves|refs?) +#[0-9]+' | grep -oE '[0-9]+' | sort -u
```
rohithgollapalli: **Nit — the keyword regex misses the colon form, which GitHub accepts.**

`\b(closes|fixes|resolves|refs?) +#[0-9]+` requires whitespace between keyword and `#`, so `Closes: #123` does not match:

```
input:  Closes: #123 / Fixes #456 / closes#789 / Refs #321
output: 456 321
```

`closes#789` missing is correct — GitHub does not accept that form either. But `Closes: #123` **is** valid GitHub syntax.

**In practice this costs nothing**, and I checked why before raising it: the colon form is a *closing* keyword, so GitHub's own resolver picks it up and it arrives via `closingIssuesReferences` on the line above. The two commands are belt-and-braces and the union covers it — which the `629 631 632` cross-check in your test plan demonstrates.

The one shape that falls through **both** is `Refs: #N` — a non-closing keyword in colon form, which GitHub never resolves and this regex never matches. One character fixes it:

```diff
-  | grep -oiE '\b(closes|fixes|resolves|refs?) +#[0-9]+' | grep -oE '[0-9]+' | sort -u
+  | grep -oiE '\b(closes|fixes|resolves|refs?):? +#[0-9]+' | grep -oE '[0-9]+' | sort -u
```

Non-blocking — a missed `Refs:` link means one issue's context is not read, not a wrong review.

.claude/skills/pr-review/SKILL.md
```diff
@@ -58,6 +80,79 @@ gh pr view "$PR_NUM" --repo "$REPO" \
...
+not double-post or fail.
+
+**Why this reaction exists.** `pr-ready` reads it as the machine-emitted "someone
+is actually reviewing this" signal, and extends its deadline by a grace window
+when it sees one. Nothing else on GitHub carries that information: a *pending*
+review is invisible to everyone but its author, and the alternatives that are
+visible (a `COMMENT` review, a label, a self-assign) all cost a notification or a
+repo-config change to say the same thing. Leave the reaction in place when the
+review is submitted — it is then a permanent marker of who read this PR, and
+`pr-ready` only counts reactions newer than its own ping, so a stale one can never
```
rohithgollapalli: **Nit — this guarantee is true across ping cycles, but worth one clause about the same-cycle case.**

"`pr-ready` only counts reactions newer than its own ping, so a stale one can never read as a live hold" is correct for what it means by *stale*: a 👀 left by last week's review predates today's ping and is filtered out. That is the case the sentence is defending against, and the filter does defend against it.

The case it does not cover is a review **abandoned inside the current cycle** — Step 0.6 posts 👀 before the slow work, so a run that dies, or one where the user declines to post at Step 6, leaves a reaction that is genuinely newer than the ping and reads as `REVIEWING`.

**This is handled**, just not by this filter: `pr-ready` grants one grace window from the claim and then terminates in `HELD BUT STALE` with the claimant's login. Bounded, and it names a person to nudge — the right degradation. Worth saying so here, because a reader of *this* file sees only the timestamp filter and could reasonably conclude a 👀 is self-cleaning in every case:

> …so a stale one from an earlier cycle can never read as a live hold. A review
> abandoned *within* a cycle does read as `REVIEWING` — bounded by `pr-ready`'s
> one-shot grace, which then reports `HELD BUT STALE` with your login.

Also the honest argument for posting 👀 early and unconfirmed: the failure mode of an abandoned claim is bounded and names you, which is a much better trade than the reaction being useless because it lands after the review.

rohithgollapalli (dismissed):
Approving. This is a process-surface change, so gate 3e — *a skill file is code, review it as code* — is the entire review. I ran every added command against the live API rather than pattern-matching it, and checked each of the four `## Review focus` questions independently instead of accepting the answers implied by the prose.

Reviewing this one had an unusual advantage: I have spent this session running `pr-review` and `revise-pr` as they exist on `main`, so several of the pain points this PR fixes are ones I hit first-hand.

## The four focus questions — checked, all four hold

1. **`created_at > PING_ISO` on every path into input 3, including adopted pings.** Holds. `PING_ISO` is pinned in Phase 2 "when the `ts` is recorded **or adopted**", and the filter is applied to inputs 3–5 uniformly. The Phase 2 shortcut still routes through Phase 6, which re-evaluates the predicate. One structural note below about *where* the filter lives.
2. **Phase 5.1's `-gt` guard, and double-extension.** Correct on both. `GRACE_EPOCH <= now` skips the wait, and the `until` loop would exit immediately anyway, so the guard is belt-and-braces rather than load-bearing. Structurally it cannot extend twice: 5.1 runs once between Phase 5 and Phase 6, and Phase 6 does not extend. The date pair round-trips — I ran it: `2026-07-28T14:05:12Z` → `1785247512` → back to the same string, with the BSD form failing over to GNU exactly as intended on this box.
3. **`pulls/$SRC_PR/comments` routing.** No path sends the target's number to an imported thread. The Step 6 table routes by source *and* verdict, `SRC_PR` is used in the reply path, and the "fixing a defect in #B does not resolve it in #A" rule is stated with the right reason (resolving there marks a live defect handled). Step 2's "tag every item with the PR it came from" is what makes it safe, and it says so.
4. **Anything upstream of 3f pulling the body in.** Nothing does. Step 0's field list is title/state/refs/author/files; Step 0.5 pipes `--json body` straight into `grep -o`; Step 0.6 hits reactions; Step 1 is checkout + `--stat`. I ran the Step 0.5 pipeline against a real body — **12 bytes of output**, digits only. The tool-boundary enforcement is the right call: intention would not survive a long session.

## Live-API verification of the test-plan claims

Every claim in the test plan is true. I checked the checkable ones rather than taking the boxes:

| Claim | Result |
|---|---|
| `--silent` exists on `gh api` | ✅ `--silent  Do not print the response body` |
| `closingIssuesReferences` returns `629 631 632` on #630 | ✅ exactly that |
| ...matching the body-grep exactly | ✅ grep returns `629 631 632` too |
| `issues/<N>/reactions` is the right endpoint | ✅ returns, correct shape |
| `pulls/<N>/reactions` 404s | ✅ `{"message": "Not Found"` |
| BSD/GNU `date` pair round-trips ISO-Z | ✅ |
| Fences balanced, frontmatter intact, across all four | ✅ 28/26/30/24, all even; all four open `---` |
| `.example` is valid JSON with both new keys | ✅ |
| No `src/` changes | ✅ nothing outside `.claude/` |

The `629 631 632` cross-check is the one I'd single out: the two Step 0.5 commands are belt-and-braces, and they agree on a real PR. That also quietly covers the `Closes: #N` colon form — see the nit.

## What the change gets right

The ordering rule is the substantive win. **"A description read first becomes the map"** is exactly the failure it describes, and Step 3f's asymmetry — *the description can add findings, never subtract them* — is the part that makes it more than a stylistic preference. Enforcing it at the tool boundary rather than by intention is what will make it survive.

The `ACKED` / `REVIEWING` split is the right cut. A 👍 stopping the nagging without buying time is correct, and deriving `REVIEWING` from a byproduct of the work instead of a convention humans must remember is the only version of that signal that stays alive.

One interaction worth naming because it is non-obvious and **the design already handles it**: a review that posts 👀 at Step 0.6 and is then *abandoned* leaves a claim that is genuinely newer than the ping, so it does read as `REVIEWING`. That is bounded correctly — one grace window from the claim, then `HELD BUT STALE` with a login attached. The system degrades to naming a person, which is the right terminal state.

The `.claude/pr-ready.local.json` disclosure in "Notes for the reviewer" — that the live allowlist contained emoji nobody uses, so *every ack was being reported as silence* — is the most valuable line in the PR body, and it is about a file not in the diff.

## Secondary

**`revise-pr` has no `argument-hint`, and this PR gives it three flags.** (inline at `revise-pr/SKILL.md:3`)

## Nits

- **`--from-comment`'s `sed` has no match guard** (inline at `revise-pr/SKILL.md:60`) — a malformed URL becomes `SRC_PR=<the whole URL>`. Fails loudly at the API, so it is safe, just confusing.
- **Step 0.5's grep misses the `Closes: #N` colon form** (inline at `pr-review/SKILL.md:99`) — harmless, because `closingIssuesReferences` resolves that form; only `Refs: #N` falls through both.
- **"a stale one can never read as a live hold"** (inline at `pr-review/SKILL.md:147`) — true across ping cycles, which is what it means; the same-cycle abandoned case is handled by the sibling's `HELD BUT STALE` rather than by this filter, and is worth one clause.

## Gates

- **3e — command-level bugs in skill/script files** — the whole review; findings above. Checked for: nonexistent flags (none — `--silent`, `closingIssuesReferences`, `--comments`, `--paginate` all verified live), word-splitting (`for N in $ISSUE_NUMS` is intentional and correct on space-separated digits), documented-but-unimplemented flags (`--post`/`--local`/`--effort` all have process steps; `--from-pr`/`--from-comment`/`--notes` all have Step 2/2.5/6 handling), injection (no user prose is interpolated into a shell body), non-idempotent writes (the 👀 POST is idempotent — GitHub returns the existing reaction), swallowed errors (the only `2>/dev/null` is the BSD→GNU `date` failover, which is a genuine portability pair, not a hidden failure), and typed params (`-F in_reply_to` is correctly `-F`, not `-f`).
- **Fixture PII (3a)** — N/A. No fixture binaries; nothing under `tests/`.
- **Design-system (3b) / style tokens (3c)** — N/A. No `src/`, no components.
- **Dead-code / fallow (3d)** — N/A against this diff. Agreed with the test plan's refusal to tick green boxes for suites that have nothing to exercise here; that is the honest version of a test plan for a `.claude/`-only change.

Verdict rule applied: **0 Blocking, so APPROVE.** The Secondary and the three nits are all one-line fixes and none of them should hold the merge. Reviewed at `64f994f`.

---

.claude/skills/revise-pr/SKILL.md
s-annam: Fixed in `824b766`. Agreed on the severity call — the flags are the whole point of the change and the hint line is the only place a user meets them.

```yaml
argument-hint: "[<pr-number>] [--repo owner/repo] [--from-pr <n>]... [--from-comment <url>]... [--notes <file>]"
```

Took the `...` spelling you suggested, since Input does document all three as repeatable. Quoted because the value starts with `[` and would otherwise parse as a YAML flow sequence — `pr-ready`'s unquoted hint gets away with it, but I would rather not rely on that.

`open-pr` still has none; it takes no flags, so leaving it as you said.

.claude/skills/revise-pr/SKILL.md
s-annam: Fixed in `824b766` — took the `case` guard, and moved it ahead of the extraction as you suggested rather than guarding after.

Your row 2 point is the one that changed my mind about ordering. The `#issuecomment-` prose ran *after* the `sed`, so the garbage `SRC_COMMENT` genuinely existed for a moment before being discarded; rejecting by name up front means it never gets constructed:

```bash
case "$URL" in
  *"/pull/"*"#discussion_r"*) ;;
  *"/pull/"*"#issuecomment-"*)
    echo "issue-level comment, not a review thread — treat as a --notes item: $URL" >&2
    exit 1 ;;
  *) echo "not a review-thread comment URL: $URL" >&2; exit 1 ;;
esac
```

Re-ran your three shapes against it: row 1 gives `PR=640 COMMENT=123456789`, row 2 rejects as issue-level, row 3 rejects as not-a-thread-url. The prose below still explains the `--notes` treatment, so the rejection message points somewhere.

Also took the framing into the file — that this fails loudly and can never write to the wrong PR is why it is a nit, and saying so keeps the next reader from "hardening" it into something that swallows the error.

.claude/skills/pr-review/SKILL.md
s-annam: Fixed in `824b766` — one character, exactly your diff.

```
input:  Closes: #123 / Fixes #456 / closes#789 / Refs: #321 / Refs #654
before: 456 654
after:  123 321 456 654
```

`closes#789` still correctly excluded.

Added your reasoning to the file rather than just the character, because the *why* is the part that would otherwise get re-litigated: for closing keywords `closingIssuesReferences` already covers the colon form and the two commands are belt-and-braces, so the only shape that fell through both is `Refs: #N` — a non-closing keyword in colon form. Without that written down the next reader sees a redundant-looking regex and is as likely to remove the `:?` as keep it.

.claude/skills/pr-review/SKILL.md
s-annam: Fixed in `824b766` — took your wording nearly verbatim, plus the closing argument.

The distinction is right and I had conflated the two cases. The filter defends against a 👀 from an earlier ping cycle; a review abandoned *within* a cycle leaves a reaction that is genuinely newer than the ping and does read as `REVIEWING`. The file now says so, and says where it is handled instead:

> …so a stale one from an earlier cycle can never read as a live hold.
>
> A review abandoned *within* a cycle is the case that filter doesn't cover — Step 0.6 posts 👀 before the slow work, so a run that dies, or one where the user declines to post at Step 6, leaves a reaction genuinely newer than the ping that does read as `REVIEWING`. That is bounded rather than unhandled: `pr-ready` grants one grace window from the claim and then reports `HELD BUT STALE` with your login.

Also kept your last line, because it is the actual justification for Step 0.6 posting early and unconfirmed and it was nowhere in the file: an abandoned claim degrades to naming a person to nudge, which beats a reaction that is worthless because it landed after the review was done. That trade is the reason the step exists and it should not have been implicit.

PR #656, opened 2026-07-29
fix(edit,docs): make "Remove bullet" real on user-added roles, and record the exporter's separator contract (#637, #620)

### LINKED ISSUE #620: [docs] record the exporter's separator contract and its parser coupling
## Summary

Dogfooding surfaced what looked like inconsistent separators in exported role headers — most read `Role · Company, Location · Team`, but one read `Role - Subtitle · Company, Location`, with an ASCII hyphen. The question was whether the hyphen is a bug or a deliberate parser constraint.

**It is neither.** The renderer never chooses a hyphen. Every separator it emits is fixed, and the hyphen in that header came from the **role title string itself** — user-authored text passed through verbatim.

The underlying request stands, though: the separator charset **is** load-bearing for the parser, that coupling is currently only implicit in scattered docblocks, and someone will eventually "normalize" it and silently break round-tripping. This issue records the constraint.

## What the exporter actually emits

Every separator in `src/lib/pdf/ats-resume-model.ts` is a literal, chosen by the exporter, never by the source:

| Join | Separator | Site |
|---|---|---|
| `Title · Company, Location · Team` | `" · "` | `:578`, `:603` (`joinHeader`) |
| `Company, Location` | `", "` | `:575-577` |
| `Title, Team` (empty-company branch, #466) | `", "` | `:600` |
| `Institution · Location` | `" · "` | `:715` |
| `Degree, Field` | `", "` | `:714` |
| `Type · Title` (achievement) | `" · "` | `:436` |
| Skills, within a category | `" · "` | `:824`, `:833` |
| Header ↔ trailing date | `"  "` (two spaces) | `:636`, `:775`, `:791`, `:803` |
| Date range | `"–"` en dash | `entry-dates.ts:33` |

Plus one deliberate exception: an **achievement's** title↔year separator echoes the *source's own* punctuation (`achievementYearJoiner`, `entry-dates.ts:57`, #380) — so a hyphen there is also user-sourced, by design, so the export re-parses to the same `year_separator` it came from.

`render-ats-pdf.ts` itself holds exactly one separator constant — `MIDDOT_SEGMENT_SEP = " · "` (`:97`) — used only for atomic wrap points.

There is **no** ASCII hyphen anywhere in that set. A `Role - Subtitle` header is a title field whose value literally contains `" - "`, drawn verbatim.

## The coupling worth recording

The `" · "` middot is not cosmetic. It is a parser contract, and the parser is sensitive to it in both directions:

- `joinHeader([title, org], " · ")` (`:603`) is what the re-parser's `mapTitleFirst` splits on to recover title / company / location / team.
- The **#466 empty-company branch** (`:582-604`) exists precisely because swapping one separator changes the parse: a naive `"Title · Team"` middot join re-parses as a `Title · Company` shape and mislabels the team as the company. The fix was to emit a **comma** instead, so the parser's role-comma split routes it back to `team`. Same characters, different separator, different field assignment.
- The `", "` in `Company, Location` and the `" · "` before `Team` are load-bearing in the same way — the comma marks the location boundary, the middot marks the team boundary.
- The **two-space** header↔date join is also a contract: the wide same-`y` gap is what `columnGapCuts` / `flush()` in `sections.ts` read as a flush-right date rail (#425).
- The **en dash** in a date range feeds `DATE_RANGE_RE` / `isLoneDateRange` (`line-primitives.ts:308`), the shared discriminator between the exporter's flush-right decision and the splitter's exemption.

Widening the accepted charset — teaching the exporter to emit `-` or `|` where it emits `·`, or teaching the parser to treat them as equivalent — changes field routing on real fixtures. `ats-resume-model.ts:560-570` already documents an adjacent tradeoff of this class (the one-line header removing the title/company structural signal, baselined in `KNOWN_FAILURES`).

## Ask

Documentation only. No behaviour change.

1. Add a short **"Separator contract"** section to `docs/canonical-resume-model.md` holding the table above, plus a one-line statement of the invariant: *the exporter's separator set is fixed and parser-coupled; user text passes through verbatim and may contain any glyph.*
2. Add a pointer to it from the `ats-resume-model.ts` module docblock, next to the existing #466 / #425 notes, so the constraint is one hop from the code that depends on it.
3. Add a test asserting the exporter's separator set — e.g. that a role header built from clean fields joins with `" · "` and that the empty-company branch joins with `", "` — so a future "normalize the separators" refactor fails a test instead of silently re-routing fields.

## Explicitly not in scope

- **Not** normalizing user-authored hyphens inside a title. `Role - Subtitle` is the user's text; rewriting it would be the exporter editing résumé content, which it must never do.
- **Not** widening the parser to accept alternative separators. That is a real feature (a source résumé using `|` or `—` between title and company parses worse today), but it is a parser-input question, separate from this export-side contract.

## Acceptance criteria

- [ ] `docs/canonical-resume-model.md` carries the separator table and the invariant statement.
- [ ] `ats-resume-model.ts`'s module docblock links to it.
- [ ] A test pins the exporter's separator choices for the standard and empty-company header shapes.
- [ ] No change to any rendered byte — `corpus-roundtrip.test.ts` and the `render-roundtrip-*.repro.test.ts` family untouched and green.

### LINKED ISSUE #637: [edit] "Remove bullet" is inert on every user-added role — and it blocks the undo-strip repro
> **Rewritten 2026-07-28.** The original body described an undo-strip unmount and was scoped as a
> lifecycle fix. Implementation found that repro is **unreachable on current `main`**, and that the
> reason it is unreachable is a larger, user-visible defect with no issue of its own. This issue is
> now that defect, with the undo-strip work as its second half. The original framing is preserved
> verbatim under [Original report](#original-report) at the bottom.

## The defect

**"Remove bullet" does nothing on any user-added role.** Blank-committing or removing a bullet on a
role the user added via **+Add experience** renders the "Removed 1 change · Undo" strip for a change
that never happened. The bullet stays on screen, stays in the score, and stays in the exported PDF.

Probe against the real `applyOverrides` + `computeAnonymousAtsScore`, one parsed role and one added
role:

```
NO-REMOVE         bullets: [[0,"Cut p99 …"],[1,"Shipped a thing …"]]
REMOVE#0 (parsed) bullets: [[0,"Shipped a thing …"]]              exp[0].description: ""   ← works
REMOVE#1 (added)  bullets: [[0,"Cut p99 …"],[1,"Shipped a thing …"]]                       ← byte-identical to NO-REMOVE
```

## Root cause

Two independent facts, both verified against the tree at `25dde7c`:

**1. The removal is dropped on the floor.** `applyOverrides` runs `applyRemovedBulletOverrides`
(`src/lib/edit/apply-overrides.ts:979`) *before* `applyAddedEntriesAndBullets` (`:997`). Its `byIndex`
map is built from `observations` — i.e. `doneScoreBullets`, the **frozen base-parse** score bullets
(`src/hooks/useAnalyzedResume.ts:157`). An added bullet's observation index does not exist in that
map, so `byIndex.get(idx)` returns `undefined` and the loop `continue`s
(`apply-overrides.ts:705-706`). The added entry is then re-pushed with its description rebuilt from
`addedBullets`, so the bullet comes straight back.

**2. Nothing ever splices `addedBullets`.** `removeBullet` (`src/hooks/useEditableParse.ts:646`) only
adds an index to the `removedBullets` `Set`. `addedBullets` is untouched, and
`isAddedEntryEmpty` (`:274`) is
`headerEmpty && (addedBullets[entry.id] ?? []).length === 0`.

## Why this blocks the original undo-strip issue

Fact 2 is also why the original repro cannot fire. An added role that has *ever* had a bullet keeps
`addedBullets[id].length >= 1` permanently, so `isAddedEntryEmpty` is false, so
`pruneEmptyAddedEntries` (`useEditableParse.ts:798`) skips it and the strip never unmounts. An added
role with *no* bullet has no bullet row, hence no `handleCommit`, hence no strip to lose.

Consequence for testing: any test written for the original criterion 1 **passes on unmodified
`main`**. The only way to make it fail is to stage a state the model cannot produce — an added entry
with a rendered bullet but an empty `addedBullets` bucket — which is the
"editing the test's input voids the proof" failure mode this repo has been bitten by before. So the
undo-strip fix is not independently provable until the model is corrected.

Adjacent paths checked and excluded: `useSectionRewrite`'s strip on an added role *is* reachable via
apply-then-undo (undo calls `replaceAddedBullets(key, [])`, which does delete the bucket), but at
that point the role is legitimately blank again and dropping it is #379's correct behaviour — and it
is WebGPU-gated. Education / Projects / Achievements added entries follow the same `addedBullets`
rule, so the same conclusion holds for each.

## Direction — two halves, each independently revert-provable

**Half 1 — make the removal real.** Make `removeBullet` entry-aware so removing an added role's
bullet actually splices `addedBullets[id]`, and have `batchUndoTargets` / `captureBulletUndoSnapshot`
record the bucket for a *remove*, not only for an *add*, so Undo restores it through the existing
`replaceAddedBullets`. Note this changes what a removed added-bullet does to the score and to the
exported PDF — a bug fix, but a visible behaviour change, so it wants its own before/after in review.

**Half 2 — keep the undo reachable.** Once half 1 lands, the added entry genuinely *does* become
empty after the remove, `pruneEmptyAddedEntries` really does drop the role, and the original
undo-strip defect becomes reproducible. Fix it with a **per-entry** prune hold — a ref-backed
registry hook in `src/hooks/`, `pruneEmptyAddedEntries(section, isHeld?)`, with `RoleEntry`
registering its own strip's pending state. Per-entry is what keeps the third acceptance criterion
below true for a *sibling* role that is empty with no live undo.

Prefer extracting into a new sibling over growing `ReconstructedRole.tsx` (588 LOC, flagged as known
debt in `CLAUDE.md`).

## Acceptance

- Removing or blank-committing a bullet on a user-added role actually removes it — from the rendered
  list, from the score, and from the exported PDF.
- Undo on that removal restores the bullet.
- After the original repro (add role → add bullet → commit empty → section-exit blur), the undo strip
  is still visible and still functional after the prune tick.
- No added role that is genuinely empty **and** has no live undo survives the prune.
- Each half is proven non-vacuous by reverting that half alone, with the observed failure output
  quoted. Reverting only one half of a two-part fix shows the rule *fires*, not that it is correctly
  *scoped*.

## Notes

Removed from the #637/#638/#640/#641 batch on 2026-07-28 — the batch was scoped as four bounded
issues and half 1 changes score and PDF output, which is not a rider that belongs in it.

---

## Original report

Raised by @Vaishnavi1709 in review of #633 (`src/components/features/ReconstructedRole.tsx:298`), and
deliberately left out of #636's review-nit sweep because it is architectural rather than a nit.

Same class as #633's round-1 blocking defect, one lifecycle level up.

### Repro (as filed — see above for why it cannot currently fire)

When `hostsStrip = true` (no `removeControl` passed — i.e. a user-added role), the undo strip lives
on that `RoleEntry`:

1. Click **+Add experience** → empty-header added role.
2. Click **+Add bullet**, type nothing, hit Save/Enter.
3. The empty commit routes through `handleCommit` → `onRemove()` → `ownRemove.removeBullet` sets
   `status: 'removed'` and drops the bullet.
4. Blur leaves the section → `sectionExitBlur`'s `setTimeout(onExit, 0)` fires →
   `pruneEmptyAddedEntries('experience')` sees `isAddedEntryEmpty` (empty header + zero bullets) →
   drops the entry → `RoleEntry` unmounts → `ownRemove.strip` unmounts with it.

### Result (as filed)

The undo the round-1 fix was designed to preserve flashes for one frame and is gone. The user has no
way back.

### Direction (as filed)

Mirror the round-1 pattern — lift the strip host for added-role bullets the way `ExperienceSection`
already hosts the Other bucket's — or defer prune while a removed-strip is live. Either is a real
design decision, not a mechanical change, which is why it did not ride along with a low-risk batch.

s-annam:
## Summary

Two issues on one branch.

**#637 — "Remove bullet" was inert on every user-added role.** `applyRemovedBulletOverrides` resolves indices against the frozen base-parse observations, where an added bullet has no entry, so the removal was dropped: the bullet stayed in the list, the score and the exported PDF while the UI rendered "Removed 1 change · Undo" for a change that never happened. `removeBullet` is now entry-aware and splices the owning `addedBullets` bucket; `batchUndoTargets` records that bucket on a remove so Undo restores it. Making the removal real also made the original undo-strip defect reachable for the first time, so its fix lands alongside as a per-entry prune hold.

**#620 — the exporter's separator set is a parser contract** that existed only implicitly across scattered docblocks. Now recorded in `docs/canonical-resume-model.md` §10, with a test pinning the standard and empty-company header shapes. No rendered byte changes.

Closes #620
Closes #637

## Behaviour change (intended, pre-authorized)

Removing or blank-committing a bullet on a user-added role now **moves the score and changes the exported PDF**. Before, it was byte-identical to not removing it. Parsed-bullet removal is unchanged — it still routes through `removedBullets`.

## Review focus

- `src/hooks/useEditableParse.ts` — every `addedBullets` writer now goes through one `writeAddedBullets` helper that sets a ref before `setState`, replacing React-managed update ordering with a hand-rolled source of truth for that field. Does any path still write the state without the ref, or read the ref where the committed value was wanted?
- `src/hooks/useEditableParse.ts` (`removeBullet`) — on a splice miss under a **parsed** entry the call deliberately falls through to the observation-indexed `removedBullets`. Is the `isAddedEntryKey` guard the right discriminator, and can a mis-tagged `entryKey` reach it?
- `src/hooks/useAddedEntryPruneHold.ts` — the hold is released on unmount. Does that ordering hold against the `setTimeout(0)` section-exit prune it exists to survive, in both directions (early release / permanent leak)?
- `docs/canonical-resume-model.md` §10 — every `ats-resume-model.ts` line citation. This PR's own docblock shifted them once already.

## Test plan

- [x] `npm run lint` clean
- [x] `npx vitest run` — 251 files / 3608 tests passed, 8 files / 10 skipped, 0 failed
- [x] `npm run check:fixtures` — 54 PDFs, all personas synthetic (no fixture touched by this PR)
- [x] `corpus-roundtrip` + `render-roundtrip-*` families untouched and green (#620 requires no rendered byte to change)
- [x] Whole-tree `fallow audit --base origin/main` — 0 dead files, 0 dead exports; complexity + duplication findings all verified inherited, not introduced by this diff

## Non-vacuity proofs

Each half of #637 was proven by reverting **that half alone** — reverting one half of a two-part fix shows the rule fires, not that it is correctly scoped.

- **Half 1 reverted alone** (added-bucket branch disabled, half 2 left intact): 8 failed / 8 passed. `expected [ …(2) ] to deeply equal [ Array(1) ]` — the removed bullet still present. The two parsed-bullet tests still passed, confirming scope.
- **Half 2 reverted alone** (`pruneEmptyAddedEntries` ignores `isHeld`, half 1 restored): 5 failed / 11 passed, every failure inside the half-2 suite. `expected '…' to contain 'Removed 1 change'`, and `Cannot read properties of null (reading 'dispatchEvent')` — the Undo button no longer exists to click.
- **The `writeAddedBullets` regression tests**, reverted against a pre-fix mutant of the hook: 10 of 28 scenarios fail, including `expected {} to deeply equal { 'added:0': [ 'B' ] }`.

## Adversarial review

Two rounds before this PR existed. Round 1 returned **2 blocking findings**, both reproduced end-to-end rather than reasoned about; round 2 returned **clean** after verifying the fixes against a purpose-built pre-fix mutant.

**Blocking 1 — `addedBullets` literal-vs-functional `setState` clobber.** The new splice wrote state with a literal value derived from a ref while four pre-existing writers used functional updaters and never touched that ref. In one synchronous handler the literal write landed last and discarded the queued update — wiping the whole map, not just one bullet. Reachable through the ordinary rewrite-apply path, which emits writes in pair order, so add-before-remove is normal; Undo did not rescue it. The shipped test missed it because the *passing* order is remove-then-add. Fixed by routing all five writers through one `writeAddedBullets` helper. Notably the ref/updater divergence was **pre-existing and systemic** — `captureBulletUndo` already snapshotted a stale bucket; this diff introduced the first literal-value writer, which is what made it destructive. The fix repairs both.

**Blocking 2 — every `ats-resume-model.ts` citation in the new §10 table was 5 lines short**, because #620's own docblock addition shifted the lines its table cites. The table was correct when written and wrong by the time it was saved, in the same commit. All 20 citations corrected and individually re-read against the source.

**Surviving nits, for the reviewer's eye:** the `text === undefined` fallbacks in `ReconstructedRole.tsx` / `ReconstructedResume.tsx` (practically unreachable); the render-phase `bulletsRef` assignment; `ReconstructedResume.tsx` LOC debt. The 81-line test-harness clone between `ExperienceSection.prune-hold.test.tsx` and `ExperienceSection.test.tsx` was judged **leave it** — the two are independent regression proofs, and a shared harness would let a future edit for one silently weaken the other.

## Known residuals — not fixed here, each now has its own issue

1. **#657** — **`bulletOverrides` is still inert on an added bullet.** `applyBulletTextOverrides` uses the same frozen `byIndex`, so a *non-blank edit* of an added bullet still silently does nothing. Only the blank-commit/remove path was in scope. Root cause #1 from #637's body is bypassed by this fix, not repaired.
2. **#658** — **A held added-entry is not re-examined when its strip collapses** — it survives until the next section exit. Deliberate: the collapse is timer-driven and could fire while the user is typing in that entry. Strictly better than before (such a role was previously never prunable at all).
3. **#659** — **The strip confirms a removal that provably did nothing** — `setStatus({kind:"removed"})` is unconditional, but `removeBullet` now has a deliberate silent early return on a bucket miss. Pre-existing, but the new return makes it a designed path.
4. **#660** — **A degenerate pure-punctuation added bullet** (`"•"`, `"-"`) normalises to an empty grouping key and lands in the "Other bullets" bucket, where its Remove is inert. `addBullet` only rejects blank-after-`trim()`.
5. **#661** — **`docs/canonical-resume-model.md`'s citations outside §10 are stale** (doc lines 20/28/45/103/118/128). Verified pre-existing — `AtsResumeModel` is cited as `:181`, was `:210` at `25dde7c`, is `:215` now. Never correct; a separate sweep.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #620 | Claude Sonnet 5 | medium |
| Implementation — #637 | Claude Opus 5 (1M context) | ultra |
| Adversarial review — round 1 | Claude Opus 5 (1M context) | — |
| Review fixes | Claude Opus 5 (1M context) | — |
| Adversarial review — round 2 | Claude Opus 5 (1M context) | — |

### DIFF
--- a/docs/canonical-resume-model.md
+++ b/docs/canonical-resume-model.md
@@ -199,3 +199,75 @@ Minted **after** this plan is accepted, one per stage, each with its round-trip-
 4. **Stage D+E** (shipped combined, #445 — supersedes the separate #446) — collapse `LlmParsedResume` + `ApplyOverridesResult` into projections and delete the hand-sync note, **and** cut over: remove the `CascadeResult` façade + ship #321 cache versioning/invalidation (§6). Combined so the façade removal is provable in one round-trip gate rather than two.
 
 Refs [#438](https://github.com/offlinecv/OfflineCV/issues/438), [#321](https://github.com/offlinecv/OfflineCV/issues/321), [#401](https://github.com/offlinecv/OfflineCV/issues/401), [#425](https://github.com/offlinecv/OfflineCV/issues/425), [#434](https://github.com/offlinecv/OfflineCV/issues/434), [#435](https://github.com/offlinecv/OfflineCV/issues/435).
+
+---
+
+## 10. Separator contract
+
+Dogfooding once read a role header as `Role - Subtitle · Company, Location` and asked
+whether the ASCII hyphen was a bug or a deliberate parser constraint. It is neither: the
+exporter never chooses a hyphen. Every separator `AtsResumeModel` emits is a fixed literal;
+the hyphen came from the **role title string itself** — user-authored text, passed through
+verbatim.
+
+**The invariant:** the exporter's separator set is fixed and parser-coupled; user text
+passes through verbatim and may contain any glyph.
+
+| Join | Separator | Site |
+|---|---|---|
+| `Title · Company, Location · Team` | `" · "` | `ats-resume-model.ts:583`, `:608` (`joinHeader`) |
+| `Company, Location` | `", "` | `ats-resume-model.ts:580-582` |
+| `Title, Team` (empty-company branch, #466) | `", "` | `ats-resume-model.ts:605` |
+| `Institution · Location` | `" · "` | `ats-resume-model.ts:720` |
+| `Degree, Field` | `", "` | `ats-resume-model.ts:719` |
+| `Type · Title` (achievement) | `" · "` | `ats-resume-model.ts:441` |
+| Skills, within a category | `" · "` | `ats-resume-model.ts:829`, `:838` |
+| Header ↔ trailing single-token date | `"  "` (two spaces) | `ats-resume-model.ts:641`, `:780`, `:796`, `:808` |
+| Experience/education date **range** | `" – "` spaced en dash | `ats-resume-model.ts:368` (`experienceDateRange`) |
+| Project/education-fallback date **range** | `"–"` unspaced en dash | `score/entry-dates.ts:19` (`buildProjectDates`), `:33` (`buildEducationDates`) |
+
+Every `ats-resume-model.ts` row above is `src/lib/pdf/ats-resume-model.ts`; the date-range
+row is `src/lib/score/entry-dates.ts`, a different directory.
+
+Plus one deliberate exception: an **achievement's** title↔year separator echoes the
+*source's own* punctuation (`achievementYearJoiner`, `score/entry-dates.ts:57`, #380) — a hyphen
+there is also user-sourced, by design, so the export re-parses to the same `year_separator`
+it came from.
+
+`render-ats-pdf.ts` holds exactly one separator constant of its own —
+`MIDDOT_SEGMENT_SEP = " · "` (`render-ats-pdf.ts:176`) — used only to keep a middot-joined
+segment atomic across a wrap point; it does not choose which fields get middot-joined.
+
+There is **no** ASCII hyphen anywhere in this set. A `Role - Subtitle` header is a title
+field whose value literally contains `" - "`, drawn verbatim.
+
+**Why it's load-bearing, not cosmetic:**
+
+- `joinHeader([title, org], " · ")` is what the re-parser's `mapTitleFirst` splits on to
+  recover title / company / location / team.
+- The **#466 empty-company branch** exists precisely because swapping one separator changes
+  the parse: a naive `"Title · Team"` middot join re-parses as a `Title · Company` shape and
+  mislabels the team as the company. The fix emits a **comma** instead (`"Title, Team"`), so
+  the parser's role-comma split routes it back to `team`.
+- The `", "` in `Company, Location` and the `" · "` before `Team` are load-bearing the same
+  way — the comma marks the location boundary, the middot marks the team boundary.
+- The two-space header↔date join is also a contract: the wide same-`y` gap is what
+  `columnGapCuts` / `flush()` in `sections.ts` read as a flush-right date rail (#425).
+- The en dash in a date range feeds `DATE_RANGE_RE` / `isLoneDateRange`
+  (`line-primitives.ts`), the shared discriminator between the exporter's flush-right
+  decision and the splitter's exemption.
+
+Widening the accepted charset — teaching the exporter to emit `-` or `|` where it emits
+`·`, or teaching the parser to treat them as equivalent — changes field routing on real
+fixtures. §7 above documents an adjacent tradeoff of this class (the one-line header
+removing the title/company structural signal, baselined in `KNOWN_FAILURES`).
+
+**Explicitly out of scope for this contract:**
+
+- Normalizing user-authored hyphens inside a title. `Role - Subtitle` is the user's text;
+  rewriting it would be the exporter editing résumé content, which it must never do.
+- Widening the parser to accept alternative separators (a source résumé using `|` or `—`
+  between title and company parses worse today) — a parser-input question, separate from
+  this export-side contract.
+
+Refs [#620](https://github.com/offlinecv/OfflineCV/issues/620), [#466](https://github.com/offlinecv/OfflineCV/issues/466), [#425](https://github.com/offlinecv/OfflineCV/issues/425), [#380](https://github.com/offlinecv/OfflineCV/issues/380).
--- a/src/components/features/BulletRemoveStatus.tsx
+++ b/src/components/features/BulletRemoveStatus.tsx
@@ -37,8 +37,11 @@ import {
 } from "./ApplyConfirmation.tsx";
 
 export interface BulletRemoveControl {
-  /** Drop one bullet by its `BulletObservation.index` and arm the strip. */
-  removeBullet: (index: number) => void;
+  /** Drop one bullet by its `BulletObservation.index` and arm the strip. `text`
+   *  is the bullet's observation text, which the owner forwards so a USER-ADDED
+   *  bullet can be located in its `addedBullets` bucket — the index alone
+   *  cannot reach one (#637, see `added-bullets.ts`). */
+  removeBullet: (index: number, text: string) => void;
   /** True while a confirmation is showing. Callers use it to suppress an
    *  "empty section" note that would otherwise contradict the strip. */
   pending: boolean;
@@ -48,13 +51,14 @@ export interface BulletRemoveControl {
 }
 
 /**
- * @param onRemoveBullet Drop a bullet by `BulletObservation.index`. Absent →
- *   `removeBullet` is inert (the caller renders no remove control either).
+ * @param onRemoveBullet Drop a bullet by `BulletObservation.index`, plus its
+ *   text for the added-bullet path (#637). Absent → `removeBullet` is inert
+ *   (the caller renders no remove control either).
  * @param captureUndo Snapshot the slot the remove will clear, BEFORE the write
  *   (issue 510). Absent → the confirmation renders without an Undo action.
  */
 export function useBulletRemoveStatus(
-  onRemoveBullet?: (index: number) => void,
+  onRemoveBullet?: (index: number, text: string) => void,
   captureUndo?: SectionRewriteApply["captureUndo"],
 ): BulletRemoveControl {
   const [status, setStatus] = useState<
@@ -64,12 +68,14 @@ export function useBulletRemoveStatus(
   >({ kind: "idle" });
 
   const removeBullet = useCallback(
-    (index: number) => {
+    (index: number, text: string) => {
       if (!onRemoveBullet) return;
       // Snapshot BEFORE the write — once it lands the prior value is gone
-      // (issue 510, same rule the rewrite-review Apply follows).
+      // (issue 510, same rule the rewrite-review Apply follows). For an added
+      // bullet the snapshotted slot is the entry's `addedBullets` bucket, which
+      // `batchUndoTargets` now records for a `remove` too (#637).
       const undo = captureUndo?.([{ kind: "remove", obsIndex: index }]);
-      onRemoveBullet(index);
+      onRemoveBullet(index, text);
       setStatus({ kind: "removed", undo });
     },
     [onRemoveBullet, captureUndo],
--- a/src/components/features/ExperienceSection.prune-hold.test.tsx
+++ b/src/components/features/ExperienceSection.prune-hold.test.tsx
@@ -1,0 +1,384 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+// @vitest-environment jsdom
+
+/**
+ * Regression test for #637 half 2 — the section-exit prune used to unmount the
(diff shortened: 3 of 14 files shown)


s-annam (commented):
The changes in this PR are highly detailed, correctly implemented, and backed by a robust and comprehensive regression suite.

### Stance & Verification
* **Verdict:** COMMENT (Self-review, 0 blocking findings)
* **CI Verification:**
  - `npm run check:fixtures` (synthetic PII check): **Green** (54 PDFs synthetic)
  - `npx tsc -b --noEmit` (type checking): **Green**
  - `npx eslint .` (linting): **Green**
  - `npx vitest run` (vitest suite): **Green**
  - `npx vite build` (production build): **Green**
  - `npx fallow audit --base origin/main` (dead code check): **Green** (0 dead files/exports)

### Linked Issues Acceptance Criteria
* **Issue #620:** MET (documented in [docs/canonical-resume-model.md](file:///C:/Users/annam/offlinecv/docs/canonical-resume-model.md#L202-L274) §10, linked from [ats-resume-model.ts](file:///C:/Users/annam/offlinecv/src/lib/pdf/ats-resume-model.ts#L19-L23), and pinned via test assertions).
* **Issue #637:** MET (verified that removing an added-bullet splices the [added-bullets.ts](file:///C:/Users/annam/offlinecv/src/lib/edit/added-bullets.ts) bucket, that undo restores it correctly, and that the undo strip survives the section-exit prune using a lease-based per-entry prune hold).

---

Issue #664, opened 2026-07-29
Non-WinAnsi characters in a candidate's name are mangled to '?' in the exported PDF


s-annam:
## Problem

When the ATS PDF exporter falls back to Helvetica, characters outside WinAnsi are silently replaced with `?` — including in the candidate's own name.

```
full_name: "ANNA WIŚNIEWSKA"  →  exported + re-parsed as  "ANNA WI?NIEWSKA"
```

`ś` (U+015B) is Latin Extended-A. On the standard-font path, `toWinAnsi()` degrades it to `?`.

Reproduces on `tests/fixtures/pdfs/unknown/extended-latin-name-roundtrip.pdf`.

## Scope — read this before proposing a fix

This issue was originally filed as "a name is mangled in the exported PDF." #667 and the analysis below narrowed it considerably, and the narrowing changes what the right fix is.

**#667 removed the embedded-font half.** `loadFonts` now returns the sanitizer matched to the font it actually loaded, so there is no path where sanitizing is skipped:

- embedded Poppins → `toEmbeddedFontSafe`, bound to the font's own `hasGlyphForCodePoint`
- Helvetica fallback → `toWinAnsi` (unchanged, #295)

**Poppins covers both `ś` (U+015B) and `ł` (U+0142)** — verified via fontkit against the vendored `src/assets/fonts/Poppins-Regular.ttf`. So on the path a user normally takes, with the font embedded, `ANNA WIŚNIEWSKA` round-trips intact today.

Two further narrowings:

- `toWinAnsi` passes the **entire Latin-1 supplement** through unchanged (`é ñ ü ç å ø`). The gap starts at Latin Extended-A, so the affected population is not "names with accents" — it is Polish, Czech, Hungarian, Turkish, Baltic, romanised CJK, and similar.
- Sanitization runs at every `drawText`. The code does **not** distinguish contact fields from body text, so `?` can land in any drawn string; "contact fields" is a scope a fix would impose, not one that already exists.

**What this retires:** the original framing — *"the candidate's own name being mangled in a document they download and send to employers"* — no longer describes the path a user takes. What remains is a **degraded-path** question: what should happen when the font fails to load. That is a different product question, and it is the one this issue is now about.

## When the Helvetica fallback is actually taken

[...]

Issue #663, opened 2026-07-29
Inline year in a dateless role's title is split out as start_date on round-trip


s-annam:
## Problem

A **dateless** role whose title carries an **inline year** loses the year out of the title on the parse → export → re-parse round-trip.

```
"Software Engineer Intern Summer 2022"
  → title: "Software Engineer Intern Summer"
  + start_date: "2022"
```

Reproduces on two roles of `tests/fixtures/pdfs/unknown/openresume-react-pdf.pdf`.

This is an asymmetry between what the exporter composes and what the parser splits: the exporter writes the title verbatim, and the re-parse then reads the trailing year as a date field the original never had.

## Why this is filed separately from #649

The baseline entry currently cites **#649** (*"Extract exporter ↔ parser round-trip contract into `src/lib/resume-format`"*). That is the wrong owner on two counts:

1. #649 is a **pure refactor** — its own acceptance criterion is *"Extraction changes **zero** rendered bytes (round-trip tests remain green)."* A zero-behaviour-change extraction can never clear this baseline.
2. #649's implementation plan **explicitly defers** date work: *"Separate PR: Unify the two date-range dialects."*

So when #649 closes as "refactor done", this entry orphans while the defect is still live — the exact cycle the issue-linked baselines added in #654 exist to end.

Not the #298 title/company swap, and not the #436 one-line header (both closed).

## Affected

- Fixture: `tests/fixtures/pdfs/unknown/openresume-react-pdf.pdf`
- Baseline: `src/lib/heuristics/corpus-roundtrip.known-failures.json` → `experience`
- Likely code: title/date splitting in the parser's entry-header path vs. title composition in `src/lib/pdf/ats-resume-model.ts`

## Acceptance criteria

- [ ] A dateless role whose title ends in a year round-trips with the title intact and no synthesized `start_date`.
- [ ] The `openresume-react-pdf.pdf` → `experience` baseline entry is deleted and `corpus-roundtrip.test.ts` is green.
- [ ] A deliberate decision is recorded for the inverse case (a role that genuinely has a trailing year *as* its date), so the fix does not simply move the asymmetry.
- [ ] Proven non-vacuous by reverting the fix alone and quoting the observed failure.

## Context

Surfaced during the adversarial review pass on the #646 P1 batch, auditing the baseline re-attributions made after #436 closed.

Issue #662, opened 2026-07-29
Experience section with a date-range heading and bulleted role lines parses zero entries


s-annam:
## Problem

A résumé whose experience section uses a **date range as the block heading**, with the role line bulleted underneath, parses to **zero experience entries**.

Shape:

```
Experience
Mar 2021 - Present
• Staff Engineer, Globex Systems LLC
• Led the payments platform migration across 14 regional services
Jun 2017 - Feb 2021
• Senior Engineer, Initech Analytics
• Rebuilt the ingest pipeline handling 3M events per day
```

Observed: `experience: []`, probe verdict `PARSER-MISS (0 entries; region has 2 date-range lines)`.

The date line is the only anchor. `headerLookback` finds only bullets above it, so each block maps to neither a title nor a company, and the `#145` date-only-phantom filter then drops both entries.

Note the section **does** have an `Experience` header — this is not a headerless section.

## Why this is filed separately from #492

#492 is *"Parser drops entire work history when the experience section has no header"*. Its single trigger is a **headerless** section, and its acceptance criteria cover only that shape. This fixture's trigger is a date-range **heading** with the role line bulleted underneath — a different route to the same `experience-parser-miss` outcome.

The corpus fixture's own generator records the distinction explicitly (`scripts/fixtures/gen-date-heading-bulleted-roles.mjs`):

> **NOT #492.** #492 is the HEADERLESS experience section, and its own body records that `experience-parser-miss` provably cannot fire there.

Filed so the corpus baselines can cite an issue that actually owns this shape. Until now the entries cited #492, which the generator contradicts — and the `check:baselines` gate certified that citation as live and correct, because the script deliberately does not judge whether a citation is the *right* issue.

## Affected

- Fixture: `tests/fixtures/pdfs/unknown/date-heading-bulleted-roles.pdf`
- Ground truth: `tests/fixtures/pdfs/unknown/date-heading-bulleted-roles.truth.json` — 3 `knownWrong` entries (`experience.title`, `experience.company`, `experience.dates`)
- Baseline: `src/lib/heuristics/corpus-edit-roundtrip.known-failures.json` → `bullets` (collateral — the pooled bullet belongs to no entry, so the exporter has nothing to render it under)
- Likely code: `headerLookback` / anchor collection in `src/lib/heuristics/entry-blocks.ts`; the `#145` date-only-phantom filter

## Acceptance criteria

[...]

Issue #655, opened 2026-07-29
Two-phase scored section router (propose candidates, then select)


s-annam:
### Problem
`classifyLine` uses a greedy first-match chain with many positional parameters and issue-numbered special cases. This is hard to maintain and buggy on complex layouts.

### Affected Files
- `src/lib/heuristics/sections.ts`

### Implementation Plan
1. Replace `classifyLine` with a two-phase router: (1) candidate proposal: recognizers emit `{lineIdx, section, tier, score, evidence}`, (2) selection: select a consistent segmentation with suppression rules.
2. Stage 1: Reproduce current corpus behavior byte-identically with existing recognizers as proposers.
3. Stage 2: Add shape-only proposers to resolve layout bugs (like `#574` and `#492`).

### Acceptance Criteria
- [ ] Greedy matchers in `classifyLine` are replaced by propose/select router.
- [ ] Reproduce stage yields byte-identical snapshots.
- [ ] New layout fixes activate and pass successfully in the second stage.

PR #665, opened 2026-07-29
fix(edit,heuristics,test): give bullets a stable identity, split the god module, and measure first-parse correctness (#646)

### LINKED ISSUE #326: Decision (wontfix by-design): #295 toWinAnsi() lossy → to -> transliteration on round-trip (skia-classic, weasyprint-classic)

### LINKED ISSUE #647: [Bug] Remove-then-edit corrupts the wrong bullet on display and export
### Problem
The UI and the override resolver disagree about which pool bullet indices refer to. `App.tsx:274`/`:345` passes `edited.score` (the re-graded pool) to `ReconstructedResume`, which keys bullet writes by the re-graded pool's `b.index` (`src/components/features/ReconstructedResume.tsx:503`, `:1019`, `:1311-1312`). But `useAnalyzedResume.ts:139-158` feeds `applyOverrides` the **pristine** `state.score.bullets`, and `src/lib/edit/apply-overrides.ts:974-984` resolves all keys against that pristine pool.
Runtime probe against the real pipeline: remove bullet A, then edit displayed bullet C (re-graded index 1) → **bullet B** (pristine index 1) is destroyed and replaced with C's new text, C untouched — user data corruption on display *and* in the downloaded PDF.

### Affected Files
- `src/components/features/ReconstructedResume.tsx`

### Implementation Plan
1. Add a test case exercising remove-then-edit in `render-roundtrip-bullet-remove.repro.test.ts`.
2. Write a point fix in `apply-overrides.ts` to ensure bullet text edits resolve correctly when removals occur, aligning indices properly between the UI's re-graded pool and the resolver's pristine pool.

### Acceptance Criteria
- [ ] Repro test case passes successfully.
- [ ] Removing bullet A, then editing bullet C updates bullet C's text on display and export rather than B.

### LINKED ISSUE #648: Stable bullet identity through parse → edit → export
### Problem
Currently, bullet edits are tracked by index, which is highly fragile and leads to data corruption (such as Bug A) when bullets are removed and the indices shift. We need a stable ID per bullet to track them robustly. Additionally, overridden bullets sometimes vanish on the Download-PDF round-trip (#487).

### Affected Files
- `src/lib/score/score.ts`
- `src/lib/pdf/ats-resume-model.ts`
- `src/hooks/useEditableParse.ts`
- `src/components/features/ReconstructedResume.tsx`

### Implementation Plan
1. **Phase 1 (root-cause #487):** Investigate why overridden bullets vanish on Download-PDF round-trip (focus on `resolveBullets` and `buildAtsResumeModel` in `src/lib/pdf/ats-resume-model.ts:343-354`). Delete the 39 `bullets` `KNOWN_FAILURES` entries in `src/lib/heuristics/corpus-edit-roundtrip.test.ts`.
2. **Phase 2 (Stable IDs):** Mint a stable ID per bullet at pristine-pool build time (entry-scoped key + occurrence, similar to `parsedEntryKey`) in `score.ts:343-364`. Key `bulletOverrides`, `removedBullets`, `addedBullets`, and the undo-batch machinery by ID. Update `applyOverrides` to resolve by ID. Since this changes the persisted `EditSnapshot` schema, implement a migration or version guard for IndexedDB.
3. **Phase 3 (Cleanup):** Drop `buildAtsResumeModel`'s `edit` parameter. Have the exporter read only the already-folded `displayResult`, removing the duplicate implementation of override semantics in `ats-resume-model.ts`.

### Acceptance Criteria
- [ ] Repro test for remove-then-edit mis-targeting passes without point-fixes.
- [ ] The 39 `bullets` entries in `KNOWN_FAILURES` are deleted and the ratchet remains green.
- [ ] Saved-library IndexedDB snapshots migrate cleanly without corruption.

### LINKED ISSUE #650: Move core line types and assembler out of god module sections.ts
### Problem
The god module `sections.ts` is currently the type hub for many importers, causing the main entry chunk `/` to statically pull a large chunk containing the line assembler. We need to split types and line-assembly utilities into dedicated modules.

### Affected Files
- `src/lib/heuristics/sections.ts`

### Implementation Plan
1. Create `src/lib/heuristics/line-model.ts` and move `PdfLine`/`PdfSection`/`SectionedResume` into it.
2. Create `src/lib/heuristics/line-assembly.ts` and move grouping/assembly helpers (`groupIntoLines`, `collapseLetterSpacing`, `orderItemsByColumn`, etc.).
3. Repoint all importers. Build-verify that the `/` entry chunk does not statically load the line assembler.

### Acceptance Criteria
- [ ] `line-model.ts` and `line-assembly.ts` are created.
- [ ] Importers are updated and build successfully.
- [ ] Corpus snapshots are byte-identical.

### LINKED ISSUE #654: Ground-truth corpus scoreboard & issue-linked KNOWN_FAILURES
### Problem
The test stack currently measures change (diffs) and self-consistency (round-trip) but never absolute correctness. We need a ground-truth measurement system to quantify first-parse correctness.

### Affected Files
- `src/lib/heuristics/corpus.test.ts`
- `src/lib/heuristics/corpus-edit-roundtrip.test.ts`
- `src/lib/heuristics/corpus-roundtrip.test.ts`

### Implementation Plan
1. **Ground truth:** Create a `<name>.truth.json` next to each synthetic fixture holding expected values. Compute precision/recall per category inside the existing corpus walk in `corpus.test.ts`.
2. **Issue-linked baselines:** Change `KNOWN_FAILURES` from `Category[]` to `{category, issue, status: "open" | "accepted"}[]`. Write a CI script that fails if a baseline cites a closed/non-accepted issue.
3. **Seed fixtures:** Add fixtures for zero-evidence defect classes (like `experience-parser-miss` for `#492`, `skills-header-unrecognized` for `#575`).

### Acceptance Criteria
- [ ] Scoreboard prints precision/recall during corpus tests.
- [ ] A CI script checks for stale issue baselines.
- [ ] Fixtures exist for the three zero-evidence classes.

### LINKED ISSUE #657: [edit] editing a user-added bullet is silently inert — same frozen-byIndex cause #637 fixed for removal
## Summary

#637 made **removing** a bullet work on a user-added role. **Editing** one is still silently inert, for the identical root cause.

A non-blank edit of a user-added bullet — typing new text into the row and committing it — does nothing. The old text stays on screen, in the score, and in the exported PDF. No error, no warning.

## Root cause

`applyBulletTextOverrides` (`src/lib/edit/apply-overrides.ts:667`) resolves each override index against `byIndex`, the map built from the **frozen base-parse observations**:

```ts
const obs = byIndex.get(idx);
if (!obs) continue;          // :676-677
```

A user-added bullet was minted by `applyAddedEntriesAndBullets` *after* the parse, so it has no observation in that pool. `byIndex.get(idx)` returns `undefined` and the loop `continue`s — the edit is dropped on the floor.

This is the same defect class #637 documented as its root cause #1. #637 fixed the **removal** path by splicing the owning `addedBullets` bucket at the source (`removeAddedBulletLine`), which bypasses the frozen map entirely. It did **not** touch the text-override path, so `applyBulletTextOverrides` still has no way to reach an added bullet.

## Why it wasn't fixed in #637

Scope. #637 was the blank-commit / remove path; a non-blank edit is a different write. Called out explicitly in PR #656's "Known residuals".

Note the interaction: because the edit never lands, `removeAddedBulletLine` matching on the row's raw `b.text` (rather than the override text) is currently *exact*. Whoever fixes this must check that the splice still matches after an edit can actually change the stored line — otherwise fixing the edit path breaks the remove path.

## Direction

The added bullet's only real home is its `addedBullets[entryKey]` bucket. Mirror what #637 did for removal: make the edit entry-aware and write the new text into the bucket, rather than trying to resolve it through `byIndex`. `src/lib/edit/added-bullets.ts` is the natural home for the sibling helper (`replaceAddedBulletLine`), next to `removeAddedBulletLine`.

Watch the same ordering hazard #637 hit: `applyBulletTextOverrides` runs before `applyAddedEntriesAndBullets`, so an added entry's description is rebuilt from the bucket afterwards — which is precisely why writing to the bucket works and writing to the description does not.

## Acceptance criteria

- [ ] Editing a bullet on a user-added role changes the rendered text, the score, and the exported PDF.
- [ ] Editing a bullet on an added entry in Education / Projects / Achievements behaves the same (same `addedBullets` rule).
- [ ] Undo restores the prior text.
- [ ] Removing an *edited* added bullet still removes the right line — the splice must match the current text, not the original.
- [ ] Parsed-bullet editing is unchanged.
- [ ] Proven non-vacuous by reverting the fix alone and quoting the observed failure.

## Context

Surfaced during the adversarial review pass on PR #656 (#637 + #620). Verified against that branch.

s-annam:
## Summary

The P1 layer of the #646 architecture roadmap, built as one batch. Three things
land together because they are the same problem seen from three sides: bullets
had no identity, so edits corrupted the wrong row; the exporter had a second
copy of override semantics that could disagree with the first; and no gate
anywhere measured whether a parse was *right*, only whether it had *changed*.

Closes #647
Closes #648
Closes #650
Closes #654
Closes #657
Refs #646

`#652` stays open — it is blocked by `#576`, which is outside this epic. The P2
items (`#649`, `#651`, `#653`, `#655`) are deferred to a second batch.

### What shipped

**Stable bullet identity (#647, #648, #657).** `BulletObservation` now carries an
`id`, and `bulletOverrides` / `removedBullets` / `addedBullets` / the undo batch
/ rewrite-review are all keyed by it. The id is `<ordinal>|<normalised text>`,
where the ordinal exists only to make the key unique — not to name a position.
Overrides apply in insertion order and each replaces the first remaining match,
so first-match resolution is already the correct target for every key;
`assignBulletIds` allocates the smallest ordinal free of every key the override
and removal maps already hold. Every write is therefore an append, and a live
row's id is guaranteed absent from the override map. Editing a user-added bullet
(#657) now writes its `addedBullets` bucket instead of being dropped against the
frozen base pool. Saved snapshots migrate lazily: an id always contains `|`, a
legacy key is all digits, and the two spaces are disjoint by construction.

**One override implementation (#648 Phase 3).** `buildAtsResumeModel` lost its
`edit` param and reads only the already-folded result, deleting the second
implementation of override semantics whose safety rested on idempotence by
convention. `/jd-fit` was routed through `scoreEditedResume` so both edit lanes
share one grading recipe.

**God-module split (#650).** `PdfLine` / `PdfSection` / `SectionedResume` moved
to `line-model.ts`, the assembler to `line-assembly.ts`. The measured win came
from one constant: `cascade.ts` statically imported `ACCOMPLISHMENT_SECTION_NAMES`
from `sections.ts`, dragging ~2,000 lines into the eager `/` chunk. Moving it
drops the eager `flags-*.js` chunk by 19,295 bytes (754,427 → 735,132; ~6.5 kB
gzip). Honest scope: `sections.ts` still re-exports everything for back-compat
and remains the hub, so this is a mechanically-safe first slice, not a retired
god module.

**Ground truth (#654).** Fifteen hand-authored `<name>.truth.json` files, read
off the page with `pdftotext -layout` rather than off the parser, assert
first-parse correctness per field. `KNOWN_FAILURES` became
`{category, issue, status, note}` with a `check:baselines` CI gate that fails on
a baseline citing a closed issue. Three new fixtures cover defect classes the
corpus had zero evidence for.

**#487, root-caused.** The edit round-trip gate scored a `display` result — base
sections carrying edited fields — so the bullet pool held pre-edit text while
entry descriptions held post-edit text. `groupBulletsByExperience` matches on
normalised text, so the edited bullet matched no description, grouped under
"Other", and vanished. A harness fault, not a product defect. `scoreEditedResume`
is now the single shared recipe so the gate and production cannot drift again,
and 38 of the 39 `bullets` baselines are retired.

## Review focus

- `src/lib/score/bullet-id.ts` — the ordinal is write-time-only and never a
  persisted reference. Does anything read it as a position? Re-grades after undo
  or `resetAll` can hand a row a *different* ordinal than before.
- `src/lib/heuristics/corpus-edit-roundtrip.test.ts` — scoping the absence check
  to the edited role retired nine baselines. Does the position assertion plus the
  field-aware presence half actually cover what the narrowing gave up?
- `src/lib/pdf/ats-resume-model.ts` — with the `edit` param gone, is every caller
  really handing it an override-applied result?
- `tests/fixtures/pdfs/**/*.truth.json` — these are only worth something if the
  values are right. Spot-check any against its PDF; where the parser disagrees,
  the disagreement is recorded rather than reconciled.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — 3701 passed, 10 skipped
- [x] `npm run verify` green end to end (fallow report-only, as configured)
- [x] `npm run check:fixtures` — 57 PDFs + 15 truth sidecars, all synthetic
- [x] `npm run check:baselines` — 11 citations resolve against live issues
- [x] Fixture personas verified synthetic — no real PII (Step 3.5): `pdftotext`
      by eye, plus pdfjs confirmation of 0 link annotations and no Author/XMP on
      all three new PDFs
- [x] Manually verified in `npm run dev` — **not done**; every claim above is
      from jsdom driving the real hook/lib chain, not a browser pass

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #647 | Claude Opus 5 (1M context) | ultra |
| Implementation — #648, #657 | Claude Opus 5 (1M context) | ultra |
| Implementation — #650 | Claude Sonnet 5 | high |
| Implementation — #654 | Claude Opus 4.5 | ultra |
| Adversarial review — hooks/UI | Claude Opus 5 | ultra |
| Adversarial review — lib | Claude Opus 5 | ultra |
| Adversarial review — test infra | Claude Opus 5 (1M context) | ultra |
| Review fixes — round 1 | Claude Opus 5 (1M context) | ultra |
| Adversarial review — round 2 | Claude Opus 5 (1M context) | ultra |
| Review fixes — round 2 | Claude Opus 5 | ultra |
| Review revisions — PR #665 | Claude Opus 5 (1M context) | high |

## Adversarial review

Two bounded rounds ran against the accumulated diff **before** this PR existed,
so the loop could iterate freely and push once. Reviewers were instructed to
break the change, not confirm it, and to reproduce every suspected defect
end-to-end rather than reason about the code. Both rounds found blocking defects.

### Round 1 — three reviewers, different lenses

**Blocking — all fixed. Every one was a defect in the new #648 identity work.**

| # | Finding | Regression? |
|---|---|---|
| 1 | Editing two normalise-identical bullets destroyed the first edit — the survivor re-minted the same id | **Yes** — index keys handled it |
| 2 | Two identical bullets could not both be removed; a "Removed · Undo" strip appeared over a no-op write | No, but inside #648's purpose |
| 3 | `A→B`, retype `A`, `A→B` again — third edit vanished; row displayed a value the export did not contain | **Yes** |

All three shared one cause: `occurrence` was counted against the pool being
graded *at that moment*, so identity was not stable. Proven regressions by
driving the identical sequences through the legacy numeric key space — still
live for snapshot migration — where they land correctly.

The fix rejected both reviewers' proposals. `occurrence` never needed to be a
position at all: because overrides apply in insertion order and each replaces
the first remaining match, first-match resolution is already correct for every
key, so the ordinal only has to be *unique*. `assignBulletIds` now allocates the
smallest ordinal free of every currently-held key, making every write an append.
That resolves all three structurally rather than patching each.

**#487's diagnosis was independently re-derived**, not taken on trust: reverting
only the gate's scoring line back to `scoreForCascade(display)` and re-baking
reproduced the deleted baseline fixture-for-fixture — 42 `bullets` failures and
exactly the 9 Group B `experience` entries, nothing else moving. Two injected
*product-path* bugs then confirmed the gate bites: 40 failed and 53 failed
against 59/59 clean.

**The ground-truth scoreboard was proven to have teeth** by four value-only
regressions (scalar, nested, array) — 14, 12, 11 and 5 fixtures red. The
decisive control: with the truth files stashed and the break still active, every
snapshot and derived assertion was blind to a wrong company on 12 fixtures.

**A citation audit refuted several baseline re-attributions.** Nine `experience`
entries blamed on #649 were reproduced and the claimed mechanism disproved — the
override lands correctly every time; the gate was failing on a `presenceCheck`
substring collision across roles. Also confirmed: #492's own repro artifacts
(`headerless-experience.pdf` and its test) exist in **no commit across 295
commits**, and the "one corpus fixture that reproduces the title-drop category"
claim was false — four others parse an empty title.

### Round 2 — reviewing the fix

**Blocking — both fixed.**

1. **The `presenceCheck` narrowing removed real teeth.** Scoping the absence
   check to the edited role correctly retired nine misattributed baselines, but
   blinded the gate to the wrong-*role* class. Injecting `experience[idx + 1] ??
   experience[idx]` — an edit landing on the next role, the literal #647 defect
   class — gave **59/59 green**. Fixed with a position assertion plus a
   field-aware presence half; the same injection now gives **48 failed / 11
   passed** (independently re-run by the orchestrator).
2. **`/jd-fit` still minted ids without `claimedBulletKeys`**, so all three
   defects were live on that lane. Routed through `scoreEditedResume`; pinned by
   a test asserting equality against `/`'s own pipeline rather than a hand-written
   expectation, so a lane that drifts fails even if its output looks plausible.

**One reviewer claim was refuted by measurement rather than complied with.** The
proposed fix asserted field-awareness would restore title↔company swap detection
"on all 57 rather than 1". It does not: the swap is *laundered by the
round-trip* — the exporter renders the pair into one header line and the parser
re-classifies it correctly coming back, so parse3 genuinely holds both values in
the right fields. Catching it needs a pre-render assertion on the model. Field
awareness was kept on its own merits, justified by a different injection (value
landing in `location`: 59/59 blind → 56 failed).

**The core identity argument survived attack.** Round 2 could not break
"occurrence needs only uniqueness" — 3+ identical bullets edited out of order, an
edit making two distinct bullets identical, removals interleaved with edits, and
cross-role duplicates were all traced and tested correct.

### Surviving nits for the human reviewer

- `ReconstructedSummary.tsx` has a dead `bulletOverrides?.[b.id] ?? b.text`
  twin that the fix deliberately left alone — removing it would change
  `buildResumeSections`' signature. Behaviour is correct either way. Review
  confirmed it is dead by construction rather than merely suspected, so the site
  now carries a note saying why.
- `check_conventions.py` blocks on the word "exactly" in a `ReconstructedRole.tsx`
  JSDoc comment introduced by #657. It is a code comment, not parser copy — reads
  as a hook false positive, and will fire on the next edit to that file.
- fallow: 22 complexity findings and 75 clone groups, all inherited and
  report-only by repo policy. **Dead code is now zero.** An earlier revision of
  this body credited its 3 dead-code hits to "pre-existing fixture generators" —
  that was backwards: they were the three generators *this* PR adds. The whole
  `scripts/fixtures/gen-*.mjs` class is unreachable by design (the committed
  artifact is the PDF), so they are registered as fallow entry points, which
  clears the 12 inherited ones too instead of just this PR's three.

### Follow-ups filed

- **#662** — date-range heading with bulleted role lines parses zero entries.
  The baselines previously cited #492, which the fixture's own generator
  docblock explicitly contradicts.
- **#663** — inline year in a dateless role's title split out as `start_date`.
  Previously cited #649, a zero-behaviour-change refactor that could never clear it.
- **#664** — non-WinAnsi characters mangle a candidate's **name** in the export.
  Previously `accepted` under closed #326, whose recorded decision weighed a
  punctuation glyph in a role title, not a person's name.

~~Not filed, worth one: `/jd-fit` discards the returned edited `fieldConfidence`
from `applyOverrides` (pre-existing; does not reach an export, as that lane has
no download surface).~~

**Retracted and fixed in this PR — the parenthetical was wrong on both counts.**
`/jd-fit` *does* reach an export (`JdFitApp` → `<Result>` → `ResultDetailTabs`
→ `ReconstructedResume`, which renders "Download resume" unconditionally, plus
`useDownloadReport`'s identity block), and once #648 Phase 3 removed
`buildContact`'s `contactOverrides` fallback it was no longer pre-existing but a
regression this PR introduced: a corrected contact field kept rendering on screen
while dropping out of the downloaded PDF. Both `useJdFitResume` branches now
carry the edited `fieldConfidence`, and
`useJdFitResume.contact-export.repro.test.tsx` pins it by asserting equality
against `/`'s own pipeline. Reverting the fix fails 3 of its 4 cases.


Samhit21 (dismissed):
_Superseded — see the current review below._

Samhit21 (changes requested):
_Superseded — see the current review below._

src/jd-fit/useJdFitResume.ts
```diff
@@ -109,14 +106,21 @@ export function useJdFitResume(analyzed: AnalyzedResume): JdFitResume | null {
...
+    // `bulletOverrides` / `removedBullets` are already filed under, and #648's
+    // three defects (an edit to one of two identical bullets destroying the
+    // other, the second duplicate being un-removable, an edit cycle diverging
+    // display from export) all reproduce here for a real user. Passing the keys
+    // the two maps hold makes the re-grade allocate AROUND them — see
+    // `bullet-id.ts`.
+    const score = scoreEditedResume(applied, base.triggers, [
+      ...Object.keys(handoffEdit.bulletOverrides),
+      ...handoffEdit.removedBullets,
+    ]);
```
Samhit21, line 123: [Blocking]: this block builds the lane's edited grade, but the `result:` on the next line keeps `base.canonical.fieldConfidence` — discarding the one `applyOverrides` just returned with user-affirmed contact edits bumped to 1.

That is what makes `buildContact`'s new "reads entirely off the canonical résumé" precondition false on this lane, and it is why a corrected contact field drops out of `/jd-fit`'s exported PDF — see `src/lib/pdf/ats-resume-model.ts:262` for the mechanism and the measurement.

```diff
-      result: { ...base, canonical: { ...base.canonical, fields: parsed } },
+      result: {
+        ...base,
+        canonical: {
+          ...base.canonical,
+          fields: parsed,
+          fieldConfidence: applied.fieldConfidence,
+        },
+      },
```

…and the mirror at `:165` on the local DropZone branch, using `edited.fieldConfidence`. That is the same shape `/` already builds at `useAnalyzedResume.ts:297-304`, so it closes the divergence rather than adding a second convention.

This also settles the "Not filed, worth one" item at the end of the PR body — "`/jd-fit` discards the returned edited `fieldConfidence` … does not reach an export, as that lane has no download surface". It is the same defect, and it does reach an export.

Worth pinning with a test asserting `/jd-fit`'s exported contact block equals `/`'s for the same edit — the same "assert equality against `/`'s own pipeline rather than a hand-written expectation" discipline the `useJdFitResume.duplicate-bullet-identity` test in this PR already uses, so a lane that drifts fails even if its output looks plausible.

src/lib/pdf/ats-resume-model.ts
```diff
@@ -238,41 +242,33 @@ function ensureScheme(url: string): string {
...
+ */
+export function buildContact(result: CascadeResult): AtsContact {
   const fields = buildContactFields(result.canonical);
   const byKey = new Map(fields.map((f) => [f.key, f]));
 
   const valueFor = (key: keyof ContactOverrides): string => {
     const field = byKey.get(key);
-    const parsed = field && !field.gated ? field.value : "";
-    return resolveContactValue(parsed, contactOverrides[key]).trim();
+    return (field && !field.gated ? field.value : "").trim();
```
Samhit21, line 262: [Blocking]: dropping the `contactOverrides` fallback is correct on `/`, but it was the only thing masking a hole on `/jd-fit` — and this line now returns `""` for a field the user explicitly corrected.

`buildContactFields` marks a field `gated` when its confidence is under `CONTACT_DISPLAY_CONFIDENCE_FLOOR` (0.5). On `/` that never bites, because `displayResult` carries `edited.fieldConfidence` (`useAnalyzedResume.ts:302`) and `applyOverrides` bumps a user-affirmed contact edit to 1 — exactly as the new docblock above says. `useJdFitResume` never got that fold: both branches rebuild the result with `fields` only, keeping the **base** `fieldConfidence`. So on that lane the value is edited, the confidence is pristine, the row reads `gated`, and this returns `""`.

The row still *renders*, because `buildContactFields` retains the value for `low_confidence` — so the screen and the downloaded PDF disagree.

`/jd-fit` does reach here: `JdFitApp` → `<Result>` → `ResultDetailTabs:109` → `ReconstructedResume`, which renders "Download resume" (`:1262`), "Download as Markdown" (`:1253`) and `ReportDownloadControl` (`:1246`) unconditionally. `useDownloadReport.ts:69` calls `buildContact` directly for the report identity block.

Measured — parser found no phone at all, user adds it:

| lane | exported phone |
|---|---|
| `/` | `(415) 555-0173` |
| `/jd-fit` | **`undefined`** |
| `/jd-fit` on `origin/main`, old two-arg `buildContact` | `(415) 555-0173` |

So this is a regression, not pre-existing. The fix belongs in `useJdFitResume`, not here — see the inline there. This line is right; its stated precondition just isn't true on both lanes yet.

scripts/check-known-failures.mjs
```diff
@@ -0,0 +1,301 @@
...
+function issueState(number) {
+  try {
+    const raw = execFileSync(
+      "gh",
+      ["issue", "view", String(number), "--repo", REPO, "--json", "state,title"],
+      { encoding: "utf8", stdio: ["ignore", "pipe", "pipe"] },
+    );
+    return JSON.parse(raw);
+  } catch {
+    return null;
```
Samhit21: [Secondary]: this `catch` collapses "GitHub says no such issue" and "we could not ask GitHub" into the same `null`, and `judgeCitation` turns both into a hard build failure.

The message a contributor gets is `"issue #N could not be resolved in offlinecv/OfflineCV"` — which reads as *you cited a bogus issue number* and sends them to check the citation, when the actual cause may be a 5xx, a rate-limit, a network blip, or an issue that was transferred out of the repo. Since `check:baselines` is now a required gate — its own CI step **and** inside `verify` — the blast radius is every PR and the merge queue.

Suggest splitting the two:

- capture the exit status / stderr from `execFileSync` rather than discarding it;
- treat a genuine not-found (`gh` prints `Could not resolve to an Issue`) as the hard failure it should be, with the current message;
- downgrade a transport failure to the same loud warn-and-skip path `ghAvailable()` already has, so the run reports "the issue-state pass could not complete" instead of accusing the baseline.

The structural checks are unaffected either way, and the orphaned-baseline detection this exists for keeps its teeth on the path that matters.

src/lib/heuristics/__test-utils__/ground-truth.ts
```diff
@@ -0,0 +1,475 @@
...
+    return null;
+  }
+  const truth = JSON.parse(raw) as FixtureTruth;
+  const bad = (msg: string): never => {
+    throw new Error(`[truth] ${truthPathFor(pdfPath)}: ${msg}`);
+  };
+  if (truth.schemaVersion !== TRUTH_SCHEMA_VERSION)
+    bad(`schemaVersion ${truth.schemaVersion} ≠ ${TRUTH_SCHEMA_VERSION}`);
+  if (!truth.provenance || truth.provenance.trim().length === 0)
+    bad("`provenance` is required — say how this file was authored");
```
Samhit21, line 361: [Nit]: the validator checks `schemaVersion`, `provenance`, and the `knownWrong` / `unannotated` key names, but not that `contact` / `experience` / `education` / `skills` are present — and `scoreAgainstTruth` dereferences all four unconditionally.

A truth file that simply omits `skills` (without listing it in `unannotated`) passes `readTruth` and then dies inside the scorer:

```
THREW: Cannot read properties of undefined (reading 'filter')
```

instead of the `[truth] <path>: …` message this function exists to produce. Given the docblock's own framing — "an un-annotated fixture is a known gap … what *is* an error is a truth file that is present but malformed" — a missing required member is squarely the case it means to catch. Cheap to add right here alongside the `provenance` check, and worth it for a file whose whole premise is that a human hand-authored it.

src/components/features/ReconstructedSummary.tsx
```diff
@@ -158,7 +162,7 @@ export function buildResumeSections(
     const exp = group.experience;
     const label = roleLabel(exp);
     const sectionBullets = group.bullets.map(
-      (b) => bulletOverrides?.[b.index] ?? b.text,
+      (b) => bulletOverrides?.[b.id] ?? b.text,
```
Samhit21, line 173: [Nit]: confirming this lookup is dead, not merely suspected — the PR body lists it as a surviving nit, so here is the proof rather than the suspicion.

`assignBulletIds` allocates each row's ordinal to be free of every key in `claimed`, and `claimed` is `[...Object.keys(bulletOverrides), ...removedBullets]`. A live row's id is therefore absent from `bulletOverrides` by construction, so `bulletOverrides?.[b.id]` can never hit and this is always `b.text`.

Agreed it isn't worth changing `buildResumeSections`' signature for. A one-line note at the site would stop the next reader re-deriving it — the same service the `ResumeBulletRow` docblock in this PR does for its own removed `override` prop.

Samhit21 (changes requested):
> **Verdict: REQUEST_CHANGES** (1 Blocking).

The shape of this is right, and the core argument survives attack. I traced the identity claim in `bullet-id.ts` independently — edit→re-edit→edit chains, interleaved removals over normalise-identical bullets, a removal undone while an edit on its twin still stands — and "the ordinal only has to be unique" holds in every one. `captureBulletUndo` keeps its load-bearing stable identity through the new `writeAddedBullets` dep (`useCallback(…, [])` all the way down). The `#487` diagnosis is convincing, and the ground-truth harness is the first thing in this repo that can see a *stably wrong* parse rather than a *changed* one.

What blocks is not the identity work: it is `#648` Phase 3. Deleting `buildContact`'s `contactOverrides` fallback was the only thing masking a pre-existing hole in `useJdFitResume`, and the result is silent contact-field loss in that lane's exported PDF. The PR body reasons it safe because "`/jd-fit` … has no download surface" — that premise is wrong, which is why this is a blocking finding rather than a follow-up.

**Gates run on `2103dc1`:** `check:fixtures` ✓ (57 PDFs + 15 truth sidecars) · `check:baselines` ✓ (11 citations resolve, 10 unfiled — at the ceiling of 10) · `typecheck` ✓ · `eslint .` ✓ (exit 0) · `vitest run` ✓ (exit 0) · design-system/token gates ✓ (no raw interactives, no hardcoded colors, no new files under `src/components/`) · `fallow audit --base origin/main` → 3 dead-code, 24 complexity, 71 clone groups (report-only per CLAUDE.md; see Nits). Gate 3e (skill/script command bugs) **ran** — `scripts/check-known-failures.mjs` is new; see Secondary #2. `pdftotext`/`pdfinfo` are not installed on this machine, so the manual fixture read was done with a `pdfjs-dist` dump instead of skipped — details below.

**Fixture PII, verified against the binaries, not the prose.** All three new PDFs are clean: synthetic personas (Marcus Halloway, Anna Wiśniewska, Priya Iyer), `@example.com`, real area code + `555` exchange + `0157`/`0164`/`0173` subscriber, fictional employers (Globex, Initech, Ridgemont State). Dumped text **plus** Info dict, XMP and link annotations via pdfjs — **zero** annotations, no Author, no XMP on all three, which is the surface `pdftotext` could not have shown me anyway. `chanchal-sharma-sample.truth.json` matches the existing Microsoft-template persona and is compliant. Sweeping `*.truth.json` in `check-fixture-pii.mjs` is the right call — it is now the only place in the repo that commits field *values* as text.

## Blocking

**1. `/jd-fit` exports a résumé with the user's corrected contact field missing.** `buildContact` now returns `""` for any field `buildContactFields` marks `gated`, with no override to fall back on. That is correct on `/`, where `displayResult` carries `edited.fieldConfidence` (`useAnalyzedResume.ts:302`) and `applyOverrides` bumps a user-affirmed contact edit to 1. `useJdFitResume` does not do that — both branches rebuild the result with `fields` only, keeping the **base** `fieldConfidence` (`:125` handoff, `:165` local). So on that lane a corrected field has the edited value but the pristine confidence, falls under `CONTACT_DISPLAY_CONFIDENCE_FLOOR` (0.5), reads `gated`, and drops out of the export — while still rendering on screen, because `buildContactFields` retains the value for `low_confidence`.

`/jd-fit` does reach that export: `JdFitApp` → `<Result>` → `ResultDetailTabs:109` → `ReconstructedResume`, which renders "Download resume" (`:1262`), "Download as Markdown" (`:1253`) and `ReportDownloadControl` (`:1246`) unconditionally. `useDownloadReport.ts:69` hits it too, via `basicsFromContact(buildContact(result))`.

Measured — parser found no phone at all (the commonest reason to edit a contact field), user adds it:

| lane | exported phone |
|---|---|
| `/` | `(415) 555-0173` |
| `/jd-fit` | **`undefined`** |
| `/jd-fit` on `origin/main`, old two-arg `buildContact` | `(415) 555-0173` |

So this is a regression, not pre-existing. End to end: résumé with no detectable phone → user types it on `/` → "Check fit against a job" → the handoff replays the override and the card shows the phone → **Download resume** on `/jd-fit` → the PDF has no phone. Same for a *low-confidence* parsed field the user corrects. See inline for the fix — it is one line per branch, and it makes `buildContact`'s new "reads entirely off the canonical résumé" precondition actually true on both lanes instead of only on `/`. It also closes the "not filed, worth one" item at the end of the PR body: same defect, and it does reach an export.

## Secondary

**2. `check:baselines` turns any transient `gh` failure into a red merge queue, with a message that misdirects.** `issueState()` catches every error and returns `null`; `judgeCitation` maps `null` to a hard fail reading `"issue #N could not be resolved in offlinecv/OfflineCV"` — so a 5xx, a rate-limit, a network blip and a transferred issue all render as "you cited a bogus issue number". This is now a required gate (its own CI step *and* inside `verify`), so the blast radius is every PR. See inline.

## Nits (non-blocking)

- **`readTruth` does not validate the members it later dereferences.** It checks `schemaVersion`, `provenance` and the `knownWrong`/`unannotated` key names, but not that `contact`/`experience`/`education`/`skills` exist — and `scoreAgainstTruth` dereferences all four. A truth file omitting `skills` (without listing it in `unannotated`) passes validation and then dies with `Cannot read properties of undefined (reading 'filter')` instead of the `[truth] <path>: …` message the validator exists to produce. See inline.
- **The fallow dead-code attribution in the PR body is inverted.** The body says "3 dead-code hits (all pre-existing fixture generators)". They are the three generators *this PR adds* — `gen-date-heading-bulleted-roles.mjs`, `gen-extended-latin-name.mjs`, `gen-top-skills-header.mjs`. `fallow audit --base origin/main` attributes those to this diff and buckets 44 inherited findings separately. Consistent with how the existing generators live and report-only either way, so nothing to change in the code — only the claim.
- **Confirming the `ReconstructedSummary` twin is dead, not merely suspected.** `assignBulletIds` allocates every live row's id around `claimed`, so a live row's id is by construction absent from `bulletOverrides` and `bulletOverrides?.[b.id] ?? b.text` can never hit. Agreed it isn't worth moving `buildResumeSections`' signature for. See inline.
- **`BulletObservation` gains `id` without a `ATS_SCORE_ALGO_VERSION` bump — checked, and it is fine.** `CACHE_SHAPE_VERSION` therefore does not move, so a pre-`#648` library record deserializes with id-less bullets. Harmless: that cached `score.bullets` reaches only `applyOverrides`' `observations` argument, which is legacy-index resolution and still keyed by `index`, while every rendered row comes from the freshly re-graded `edited.score`. Recording it so the next reader doesn't have to re-derive it.

---

src/lib/heuristics/__test-utils__/ground-truth.ts
```diff
@@ -0,0 +1,475 @@
...
+  return pdfPath.replace(/\.pdf$/i, ".truth.json");
+}
+
+/**
+ * Read and validate one fixture's truth file. Returns `null` when the fixture is
+ * not annotated — the corpus is annotated INCREMENTALLY, so an un-annotated
+ * fixture is a known gap the scoreboard reports, not an error. What is an error
+ * is a truth file that is present but malformed.
+ */
+export function readTruth(pdfPath: string): FixtureTruth | null {
```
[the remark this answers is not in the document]
s-annam: Fixed in `d867d0b`. `readTruth` is decomposed into the walk plus three named validators — `validateTruthMembers`, `validateTruthFieldKeys`, `validateKnownWrongEntries` — so the loader reads as the sequence it is. This finding and the missing-required-member nit below landed on the same function and pulled in opposite directions (the nit adds checks, this one wants fewer branches); the split resolves both. `fallow audit --base origin/main` no longer reports it.

src/lib/heuristics/corpus-gate.test-utils.ts
```diff
@@ -51,14 +51,127 @@
...
+ * at MODULE scope (the per-fixture `it`s are generated from the fixture walk, so
+ * the baseline has to exist before any test body runs) — and `expect` outside a
+ * test has no test to fail. Precedent: `sections.config.ts`'s validating loader.
+ *
+ * Every check here is one the CI script ALSO performs, deliberately: the script
+ * is the only thing that can reach GitHub for issue state, but a malformed
+ * baseline should break the test run a contributor is already watching, not only
+ * CI.
+ */
+export function loadKnownFailures<C extends string>(
```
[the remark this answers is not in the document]
s-annam: Fixed in `d867d0b`. The per-entry rules are extracted into `validateBaselineEntry`, leaving `loadKnownFailures` as the walk over `baselines`. The helper deliberately takes the **loose** entry shape (`status: string`) rather than `KnownFailure<C>` — it is part of the narrowing, so presupposing the check it performs would be circular. No longer reported by fallow.

src/lib/heuristics/line-assembly.ts
```diff
@@ -0,0 +1,442 @@
...
+ *     then runs over every item exactly as it did before column-awareness, so
+ *     the single-column output is byte-identical.
+ *   - present            → bands are emitted page-major, ascending page order,
+ *     and within a split page the **entire left column precedes the entire
+ *     right column** (`item.x < split` → left, else right). A page without a
+ *     split contributes one band of all its items. Same-line clustering never
+ *     crosses pages, so per-page banding concatenated equals the old global
+ *     grouping whenever no page splits.
+ */
+export function orderItemsByColumn(
```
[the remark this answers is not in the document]
s-annam: Fixed in `d867d0b`. The per-page banding rule moves to `bandsForPage`, so `orderItemsByColumn` reads as "page-major, ascending" and the one rule that varies per page owns its own function. Behaviour is identical — the empty-band skip is now a `.filter()` over `[left, right]`. No longer reported by fallow.

src/jd-fit/useJdFitResume.ts
s-annam: Fixed in `d867d0b` — applied exactly as suggested, on both branches: `applied.fieldConfidence` at the handoff branch and `edited.fieldConfidence` at the local DropZone mirror, matching `/`'s own fold at `useAnalyzedResume.ts:297-304` rather than adding a second convention.

Also took the test suggestion, and it earned its place. `src/jd-fit/useJdFitResume.contact-export.repro.test.tsx` asserts equality against `/`'s pipeline rather than a hand-written expectation, over four cases: a phone typed into an empty field, a correction to a low-confidence parsed field (`location` seeded at 0.3, under the floor), a multi-field edit compared block-for-block, and an unedited control.

The control is the one that pays: with no override the two lanes must **already** have agreed, so it fails if the fix moved the unedited export instead of closing the gap. Revert-proofed — with the fix removed, 3 of 4 fail and the control still passes; with it, 4/4.

One correction to my own first draft of that test: a gated field comes back `undefined` from `buildContact`, not `""`.

src/lib/pdf/ats-resume-model.ts
s-annam: Fixed in `d867d0b`, in `useJdFitResume` as you said rather than here — this line is right and stays as-is. The precondition in its docblock ("reads entirely off the canonical résumé") is now true on both lanes, which is what makes the `contactOverrides` removal safe rather than merely safe-on-`/`.

Confirming your measurement reproduced: the `/jd-fit` phone came back `undefined` before the fix and `(415) 555-0173` after, and the new repro test fails 3 of its 4 cases when the fix is reverted. Thank you for measuring the three-lane table rather than reasoning about it — the `origin/main` row is what established this as a regression this PR introduced and not pre-existing debt, which is what moved it from a follow-up to a blocker.

scripts/check-known-failures.mjs
s-annam: Fixed in `d867d0b`, along the three lines you suggested. `issueState` now captures `err.stderr` and returns a discriminated result: parsed `{state,title}`, `null` for a genuine not-found (matched on `gh`'s resolution error, keeping the current hard-fail message), or `{unreachable: <first line of stderr, capped>}` when the lookup itself failed. `judgeCitation` maps `unreachable` to `warn` with a message that names the transport cause and says the citation was **not** validated.

Four unit tests cover it, including one asserting the warn message does *not* contain "could not be resolved" — the phrase that did the misdirecting — and one pinning that a real not-found still fails hard, since orphan detection is the point of the gate.

src/lib/heuristics/__test-utils__/ground-truth.ts
s-annam: Fixed in `d867d0b`. `validateTruthMembers` now checks all four before the scorer can reach them — `contact` must be an object, and `experience` / `education` / `skills` must be arrays (an object where an array belongs fails `scoreAgainstTruth` identically to a missing one, so the check is `Array.isArray`, not mere presence).

Added tests as well, since the function had none: a well-formed file, the absent-sidecar `null` path, and — per member — one case omitting it and one with it present but non-array, each asserting the `[truth] …` prefix and the member name. Plus a regression case re-running every guard the function already had, because the decomposition this landed alongside moved them into a different function and "still enforces the checks it already had" is exactly the thing a split can silently drop.

src/components/features/ReconstructedSummary.tsx
s-annam: Noted in place in `d867d0b` — your derivation, at the site, so the next reader gets the proof rather than the suspicion. The lookup stays; the comment says it can never hit, why (`assignBulletIds` allocates each live row's ordinal free of every key in `claimed`, so a live row's id is absent from `bulletOverrides` by construction), and why it is kept anyway rather than removed.

The PR body's surviving-nits entry is updated to match — it previously said "dead twin, deliberately left alone", which read as suspected. It is confirmed.

s-annam:
## Review addressed — `d867d0b`

@Samhit21 thank you for turning this around so quickly. It is a 96-file, +6.8k diff and you had a reproduced, measured blocking finding on it within the hour — that is a real service, and it caught something the three pre-PR adversarial rounds and I all missed.

Worth naming what made the Blocking finding land rather than get deferred: the three-lane table (`/` · `/jd-fit` · `/jd-fit` on `origin/main`). The `origin/main` row is the whole argument — it is what turned "pre-existing debt on a lane we plan to delete" into "a regression this PR introduces", and those two get different treatment. My own PR body had written the same defect off as *"not filed, worth one — does not reach an export, as that lane has no download surface"*, and both halves of that parenthetical were wrong. Tracing `JdFitApp` → `<Result>` → `ResultDetailTabs` → `ReconstructedResume` to an unconditional "Download resume" button is what showed it.

All 8 threads are addressed and resolved.

### Blocking

**1. `/jd-fit` dropped a corrected contact field from the exported PDF.** Fixed exactly as suggested, on both branches — `applied.fieldConfidence` on the handoff branch and `edited.fieldConfidence` on the local DropZone mirror, matching `/`'s fold at `useAnalyzedResume.ts:297-304` rather than inventing a second convention. `buildContact` is untouched; its "reads entirely off the canonical résumé" precondition is simply now true on both lanes.

Took the test suggestion too. `useJdFitResume.contact-export.repro.test.tsx` asserts **equality against `/`'s own pipeline**, never a hand-written expectation, over four cases: a phone typed into an empty field, a correction to a low-confidence parsed field (`location` seeded at 0.3, under the floor), a multi-field edit compared block-for-block, and an unedited control. The control is the one that pays for itself — with no override the lanes must *already* have agreed, so it fails if the fix moved the unedited export instead of closing the gap.

Revert-proofed: **fix removed → 3 of 4 fail, control still passes. Fix restored → 4/4.**

### Secondary

**2. `check:baselines` turned a transport blip into a red merge queue.** Split along the three lines you proposed. `issueState` captures `err.stderr` and returns either the parsed state, `null` for a genuine not-found (hard fail, message unchanged), or `{unreachable}` when the lookup itself failed — which `judgeCitation` maps to `warn`, naming the transport cause and stating the citation was *not* validated. Four tests, including one asserting the warn message does **not** contain "could not be resolved", and one pinning that a real not-found still fails hard.

### Nits — all four taken

- **`readTruth` required members** — `validateTruthMembers` checks all four before the scorer can reach them (`Array.isArray`, not mere presence: an object where an array belongs fails identically). The function had no tests at all, so it has them now, including a regression case re-running every guard it already had.
- **`ReconstructedSummary` dead twin** — your derivation is now a comment at the site. Kept, not removed; the note says why it can never hit and why it stays.
- **Inverted fallow dead-code attribution** — corrected, and then made moot. The 3 hits were the generators *this* PR adds, as you said. Rather than just fix the sentence: the whole `scripts/fixtures/gen-*.mjs` class is unreachable by design (the committed artifact is the PDF), so they are registered as fallow entry points. **Dead code is now zero**, and that clears the 12 inherited ones too instead of only this PR's three.
- **`BulletObservation.id` without a version bump** — recorded, no change needed. Agreed with your reasoning.

### The three fallow complexity findings

All three cleared by decomposition rather than suppression — `readTruth` → three named validators, `loadKnownFailures` → `validateBaselineEntry`, `orderItemsByColumn` → `bandsForPage`. `readTruth` is the interesting one: the complexity finding and the missing-member nit landed on the same function and pulled in opposite directions, so the split is what let both be satisfied at once.

### Gates on `d867d0b`

`typecheck` ✓ · `vitest run` ✓ **3718 passed / 10 skipped** · `eslint .` ✓ (exit 0) · `fallow audit --base origin/main` → **dead code 0** (was 3), 22 complexity + 75 clone groups, all inherited and report-only per `CLAUDE.md`.

No fixture binaries were added or changed in this round, so the PII surface is unchanged from the three PDFs you already verified against the binaries.

Collapsed back to a single commit for the merge queue, so this is a force-push — the old→new delta is on the `Compare` link above.

Samhit21 (approved):
> **Verdict: APPROVE** (0 Blocking). Re-review of `d867d0b`, after the fixes to my review of `2103dc1`.

All five findings are addressed, and two of them better than I asked for. I re-verified the blocking one against the real pipeline rather than reading the diff, and re-ran every gate on the new head.

**Gates run on `d867d0b`:** `check:fixtures` ✓ (57 PDFs + 15 truth sidecars) · `check:baselines` ✓ (11 citations, 10 unfiled — at the documented ceiling) · `typecheck` ✓ · `eslint .` ✓ (exit 0) · `vitest run` ✓ (3718 passed, 10 skipped — up 17 from `2103dc1`).

## Blocking #1 — `/jd-fit` contact export — **fixed, re-verified empirically**

Both branches now fold the edited `fieldConfidence` (`:125` handoff, `:178` local), which is the fix I proposed and matches what `/` does at `useAnalyzedResume.ts:297-304`. Re-ran my original reproducer against this head:

| shape | exported phone |
|---|---|
| what `useJdFitResume` builds **now** | `(415) 555-0173` |
| what it built **before** (base confidence) | `undefined` |

The second row is the control — the reproducer still distinguishes the two shapes, so it would catch a revert rather than passing vacuously.

`useJdFitResume.contact-export.repro.test.tsx` pins it properly: an empty field the user typed into, a correction to a low-confidence parsed field, a multi-field edit asserted **against `/`'s own output** rather than a hand-written expectation, and the no-edit case. That last one matters — it is what stops the fix from being "always force confidence to 1".

The inline comment naming `CONTACT_DISPLAY_CONFIDENCE_FLOOR` and the `#648`-dropped fallback is the right thing to leave behind; the next person to touch that object will see why the field is there.

## Secondary #2 — `check:baselines` transport failures — **fixed**

`issueState` now separates "GitHub says no such issue" from "we could not ask", and `judgeCitation` warns instead of failing on the latter, with a message that says the citation was *not* validated. That is the distinction I asked for, and routing it to the same degradation `ghAvailable()` already takes is the right call — one behaviour for "the issue-state pass could not run", not two.

## Nits — all three addressed

- **`readTruth` member validation** — now rejects a missing `contact`/`experience`/`education`/`skills` with the `[truth] <path>: …` message instead of dying inside the scorer. The `[]`-if-the-page-has-none wording in the error is a good touch for a hand-authored file.
- **fallow dead-code attribution** — fixed at the *root cause* rather than by correcting the PR body: `scripts/fixtures/gen-*.mjs` is now declared an entry point, with a comment explaining that the committed artifact is the PDF and the generator is kept beside it deliberately. That stops every future fixture-adding PR from re-surfacing the same finding, which is more than the nit asked for.
- **`ReconstructedSummary` dead twin** — the note is there, and it states the mechanism (`assignBulletIds` allocates around `claimed`) rather than just "this is dead", so it survives someone changing the allocator.

## The two changes I had not flagged

Both are behaviour-preserving extractions against fallow complexity findings, and I checked both for equivalence rather than taking the docblocks' word:

- `bandsForPage` — `[left, right].filter(b => b.length > 0)` is exactly the prior pair of `if (…length > 0) push` guards, same order.
- `validateBaselineEntry` — same checks in the same order; `seen.add` correctly stays in the caller *after* validation, so the duplicate-category check still sees the pre-add set.

## Not blocking, for whenever

The 10 `unfiled` ground-truth disagreements sit exactly at `UNFILED_TRUTH_CEILING`. That is by design and printed on every run, but it does mean the next measured-but-unfiled defect fails `corpus.test.ts` rather than being recordable — worth knowing before the next annotation pass, not something to hold this PR for.

---
Verdict rule: 0 Blocking → APPROVE. Secondary and Nits from the prior round are all resolved; nothing new surfaced.

---

Issue #295, opened 2026-07-03
Round-trip: renderAtsResumePdf hard-crashes on non-WinAnsi glyphs (→, ‐) — kills Download-PDF


s-annam:
## Symptom

`renderAtsResumePdf` (the "Download PDF" surface) **hard-crashes** when the parsed résumé text contains a glyph outside WinAnsi. pdf-lib `StandardFonts` (Helvetica et al.) only encode WinAnsi, so `drawText` throws `WinAnsi cannot encode "…"` and the whole Download-PDF path dies — no file, an unhandled exception in the UI.

Observed crashing glyphs from the corpus:
- `→` (U+2192, rightwards arrow)
- `‐` (U+2010, hyphen — the *Unicode* hyphen, distinct from ASCII `-`)

This is the **highest-severity** of the round-trip audit finds: not a field swap, a crash.

## Affected fixtures (round-trip gate `render` category)

- `google-docs/google-docs-skia-proxy-classic.pdf`
- `unknown/weasyprint-cairo-classic.pdf`
- `word/openresume-laverne-word-quartz.pdf`

Any real résumé containing an arrow, a Unicode hyphen/dash, smart quotes, `•` in body text, ligatures, em-space, etc. triggers the same crash in production.

## Root cause

`src/lib/pdf/render-ats-pdf.ts` draws every string with a StandardFont (WinAnsi codec). pdf-lib's `encodeUnicodeCodePoint` throws on any code point the codec can't represent. There is no sanitization or fallback between the parsed text and `drawText`.

## Fix direction

Pick one (prefer 1 for scope; 2 if we want fidelity):
1. **Sanitize before draw** — a `toWinAnsi(text)` pass that transliterates common non-WinAnsi glyphs to safe equivalents (`→`→`->`, `‐`/`–`/`—`→`-`, smart quotes→ASCII, `•`→`-`, NBSP→space) and drops/`?`-replaces the rest. Cheap, no font embedding, guaranteed no crash.
2. **Embed a Unicode font** (fontkit + a bundled TTF subset) so the glyphs render as-is. Higher fidelity, bigger bundle, more work.

Whichever: `drawText` must **never** throw on arbitrary parsed text — worst case degrades a glyph, never crashes.

## Definition of done

- No input causes `renderAtsResumePdf` to throw (fuzz/unit test with the 3 crashing glyphs).
[...]

PR #667, opened 2026-07-29
fix(pdf): sanitize against the embedded font's real glyph coverage (#664)

s-annam:
## Summary

The export path skipped sanitization entirely whenever Poppins embedded successfully, on the premise that "the embedded font encodes the glyphs directly". That premise is false for any code point Poppins lacks: **pdf-lib does not throw there — it draws the font's `.notdef` glyph, which extracts back out of the downloaded PDF as U+0000.**

Found while manually verifying #665 in a real browser. Parse → Download resume → re-upload:

```
parsed role title  : "Software Engineer Intern → Junior Engineer"
after round-trip   : "Software Engineer Intern  Junior Engineer"
```

Identical on `origin/main`, so this is pre-existing, not a #665 regression. A NUL is **strictly worse** than the WinAnsi path's `?` that #664 was filed about: `?` is visible and self-explaining, a NUL is invisible on screen and rides into every downstream consumer of the re-parsed field.

`loadFonts` now returns the sanitizer matched to the font it actually loaded, rather than a boolean the caller interprets — so there is no longer a code path where sanitization is skipped:

- Helvetica fallback → `toWinAnsi` (unchanged, #295)
- embedded Poppins → `toEmbeddedFontSafe`, bound to the font's own `hasGlyphForCodePoint` read off the same TTF bytes pdf-lib embedded

Degrading against real coverage rather than a hardcoded table preserves the reason the embedded path skipped sanitizing in the first place: glyphs Poppins *does* cover still pass through untouched. Verified against the vendored `Poppins-Regular.ttf` via fontkit:

| Covered (pass through) | Not covered (degrade) |
|---|---|
| `ś` U+015B, `ł`, `–`, `•`, `·`, `≥`, `€` | `→` U+2192, `★` U+2605, `✓` U+2713 |

If the coverage probe itself fails, it falls back to `toWinAnsi` — the conservative direction, since that can over-degrade but can never emit a NUL.

Refs #664

## Why #664 remains open

#664's acceptance criteria are not met by this PR, and saying otherwise would auto-close an issue whose product decision nobody has made:

- **AC1** — a Latin-Extended name surviving round-trip — now holds on the *embedded* path (Poppins covers `ś`), but the **standard-font path is unchanged**, and that is the path Node/CI always takes. `ANNA WIŚNIEWSKA` still exports as `ANNA WI?NIEWSKA` there.
- **AC2** — deleting or re-recording the `contact` baseline entry — targets `corpus-roundtrip.known-failures.json`, which **does not exist on `main` yet**; it arrives with #665. Unreachable from a branch off `main`.
- **AC3** — the written decision about degrading a candidate's *name* specifically — is still owed, and is a product call, not a code change.

Worth noting for that decision: **Poppins already covers `ś`.** So the residual gap on the embedded path is symbols, not accented Latin — which narrows "we need a Unicode-capable embedded font" from a rewrite to a much smaller question.

## Review focus

- `src/lib/pdf/render-ats-pdf.ts` — `makePoppinsSanitizer` probes `hasGlyphForCodePoint(0x41)` once at construction. Is catching a broken/absent fontkit build *there* rather than on the first `drawText` the right seam, or does it mask a fontkit version skew we'd rather see fail loudly?
- `toEmbeddedFontSafe` reuses `WINANSI_TRANSLITERATIONS` — those entries are ASCII spellings of the glyph's *meaning* (`→` ⇒ `->`), so they should read correctly under any font. Does any entry in that table encode something WinAnsi-specific that shouldn't apply here?
- ~~Tab/newline/CR are passed through before the coverage check. Any other code point that is layout, not a drawn glyph?~~ **Answered in review, and it was a defect, not a gap:** Poppins reports a real glyph for **U+0000** (`hasGlyphForCodePoint(0) === true` on both vendored faces — the only control code point it claims), so probing coverage first emitted a NUL verbatim and left the control-character drop below it unreachable. All C0/C1 controls are now dropped *before* the probe.

## Test plan

- [x] `npm run test` green (3725 passed, 10 skipped)
- [x] `npm run lint` clean
- [x] Manually verified in the browser — exported and re-parsed a résumé containing `→`; no NUL in the round-tripped field
- [x] Revert-proofed in **both** directions — "no NUL" alone is too weak an assertion. Reverting the sanitizer to a no-op fails the 4 new cases and none of the 3 existing ones; replacing it with the naive `toWinAnsi` makes the arrow cases pass but fails the Latin-Extended (`ł`) guard.
- [x] Revert-proofed the control-character hoist on its own — reverting only that fails the two embedded-path NUL cases while the Helvetica one still passes, which is the correct asymmetry (`toWinAnsi` never had this bug).

## Provenance
Review revisions: Gemini 3.5 Flash (High) (high)
Review revisions (round 2, #667 threads): Claude Opus 5 (1M context) (high)

Discovery: manual browser verification of #665 (Playwright, dev server)
Verification: `npm run typecheck` + `npm run test` + `npm run lint` — green locally; `verify` runs in CI

### DIFF
--- a/src/lib/pdf/render-ats-pdf.ts
+++ b/src/lib/pdf/render-ats-pdf.ts
@@ -259,6 +259,72 @@ const WINANSI_TRANSLITERATIONS: Record<string, string> = {
 };
 
 /**
+ * Sanitize `text` for an EMBEDDED font, keeping every glyph that font can
+ * actually draw.
+ *
+ * An embedded font is not a licence to skip sanitization. pdf-lib does not
+ * throw on a code point the embedded font lacks — it emits the font's
+ * `.notdef` glyph, which extracts back as **U+0000**. So a character Poppins
+ * has no glyph for (U+2192 "→", U+2605 "★", U+2713 "✓" — verified against the
+ * vendored `Poppins-Regular.ttf`) survived to `drawText` unsanitized and
+ * landed in the downloaded PDF as a NUL byte. Re-parsing that PDF reads the
+ * NUL straight back into a user-facing field (a role title, a name), where it
+ * is invisible on screen and travels into every downstream consumer.
+ *
+ * That is strictly worse than the WinAnsi path's "?" degradation: "?" is
+ * visible and self-explaining, a NUL is neither. Both are the #664 family —
+ * this fixes the embedded half, whose scope note wrongly assumed an embedded
+ * font could not reproduce it.
+ *
+ * `hasGlyph` is the font's own coverage predicate, so this degrades ONLY what
+ * the font genuinely cannot draw. Latin-Extended glyphs Poppins does cover
+ * (e.g. "ś" in a candidate's name) still pass through untouched — the reason
+ * the embedded path skipped `toWinAnsi()` in the first place, preserved here.
+ *
+ * The predicate is NOT trusted for control characters, which are dropped
+ * before it runs: Poppins answers `true` for U+0000, so a NUL already in an
+ * input field (re-uploading a PDF exported by a pre-fix build puts one there)
+ * would otherwise pass the probe and ride straight back out.
+ */
+function toEmbeddedFontSafe(
+  text: string,
+  hasGlyph: (codePoint: number) => boolean,
+): string {
+  if (!text) return text;
+  let out = "";
+  for (const ch of text) {
+    const code = ch.codePointAt(0) ?? 0;
+    // Tab/newline/CR: layout whitespace, never drawn as a glyph.
+    if (code === 0x09 || code === 0x0a || code === 0x0d) {
+      out += ch;
+      continue;
+    }
+    // Other C0/C1 control characters: drop silently (as toWinAnsi does).
+    // This MUST precede the coverage probe. Poppins reports a real glyph for
+    // U+0000 (`hasGlyphForCodePoint(0) === true` in both vendored faces — the
+    // one control code point it claims), so probing first would let the exact
+    // byte this function exists to eliminate through verbatim, and would make
+    // this drop unreachable for it. No entry in WINANSI_TRANSLITERATIONS is in
+    // the control range, so nothing transliterable is lost by dropping first.
+    if (code < 0x20 || (code >= 0x7f && code <= 0x9f)) continue;
+    if (hasGlyph(code)) {
+      out += ch;
+      continue;
+    }
+    // Same transliteration table the WinAnsi path uses: these are ASCII
+    // spellings of the glyph's MEANING ("→" ⇒ "->"), so they read correctly
+    // whichever font is in use — they are not WinAnsi-specific.
+    const replacement = WINANSI_TRANSLITERATIONS[ch];
+    if (replacement !== undefined) {
+      out += replacement;
+      continue;
+    }
+    out += "?";
+  }
+  return out;
+}
+
+/**
  * Sanitize `text` to the WinAnsi (Windows-1252) subset that pdf-lib's
  * StandardFonts can encode. Glyphs WinAnsi already supports (en/em dash,
  * curly quotes, bullet, ellipsis, NBSP, ...) pass through unchanged; glyphs
@@ -345,15 +411,23 @@ function loadPoppinsBytes(): Promise<{
  * since pdf-lib's built-in `embedFont` can only parse the 14 standard fonts
  * without it); on ANY failure — fetch error, corrupt bytes, an embed
  * rejection — falls back to pdf-lib's built-in Helvetica / Helvetica-Bold, so
- * a font problem never blocks the download. `isEmbedded` tells the caller
- * whether Poppins is actually in use: only then can `toWinAnsi()`
- * sanitization be skipped (StandardFonts can only encode WinAnsi; embedded
- * Poppins encodes the glyphs directly — see `toWinAnsi()` above).
+ * a font problem never blocks the download.
+ *
+ * Returns the `sanitize` step to run before every `drawText`, chosen to match
+ * the font that was actually loaded — never skipped. On the Helvetica fallback
+ * that is `toWinAnsi()` (StandardFonts encode WinAnsi only, #295); on the
+ * embedded path it is {@link toEmbeddedFontSafe} bound to Poppins' own glyph
+ * coverage, because an embedded font silently emits `.notdef` (extracting as
+ * U+0000) for a glyph it lacks rather than throwing.
  */
 async function loadFonts(
   doc: Doc,
   parts: PdfLibParts,
-): Promise<{ regular: PdfFont; bold: PdfFont; isEmbedded: boolean }> {
+): Promise<{
+  regular: PdfFont;
+  bold: PdfFont;
+  sanitize: (text: string) => string;
+}> {
   try {
     // `@pdf-lib/fontkit` ships no usable default-export `.d.ts` shape, so it
     // is typed `unknown` in `PdfLibParts` and cast here at the one call site
@@ -364,11 +438,15 @@ async function loadFonts(
     // `subset: true` prunes the embedded font to only the glyphs the résumé
     // actually uses — a downloaded PDF touches ~60–80 glyphs, so this trims the
     // full Poppins Regular + Bold (a few hundred KB) down to what's on the page.
-    // Orthogonal to the skip-`toWinAnsi()` path: subsetting prunes unused
-    // glyphs, it doesn't change the embedded-encoding logic.
+    // Orthogonal to the sanitizer choice: subsetting prunes unused glyphs, it
+    // doesn't change which code points the font can encode.
     const regular = await doc.embedFont(bytes.regular, { subset: true });
     const bold = await doc.embedFont(bytes.bold, { subset: true });
-    return { regular, bold, isEmbedded: true };
+    return {
+      regular,
+      bold,
+      sanitize: makePoppinsSanitizer(parts, bytes.regular, bytes.bold),
+    };
   } catch (err) {
     console.warn(
       "Poppins font embed failed, falling back to Helvetica:",
@@ -377,7 +455,68 @@ async function loadFonts(
   }
   const regular = await doc.embedFont(parts.StandardFonts.Helvetica);
   const bold = await doc.embedFont(parts.StandardFonts.HelveticaBold);
-  return { regular, bold, isEmbedded: false };
+  return { regular, bold, sanitize: toWinAnsi };
+}
+
+/**
+ * The sanitizer for the embedded-Poppins path: {@link toEmbeddedFontSafe}
+ * bound to Poppins' real coverage, read off the same TTF bytes pdf-lib just
+ * embedded (so the predicate cannot drift from the font on the page).
+ *
+ * BOTH faces are probed and a code point must be covered by both to survive.
+ * `groupRunsIntoWords` draws bold runs — every role title and section heading
+ * — with `fonts.bold`, a different TTF from `fonts.regular`. The two vendored
+ * files happen to have identical coverage today (diffed across U+0020–U+10FFFF
+ * minus surrogates: zero differences), so a Regular-only probe would be
+ * correct by accident; requiring both makes the fix independent of that, so a
+ * weight swap, a variable-font migration or a Poppins bump cannot silently
+ * reinstate `.notdef` in bold text.
+ *
+ * Coverage is memoized per byte-buffer and probed lazily — `hasGlyphForCodePoint`
+ * is only called for code points a résumé actually contains, which is a few
+ * dozen distinct values per document.
+ *
+ * Falls back to `toWinAnsi` if fontkit can't parse the bytes here for any
+ * reason. That is deliberately the CONSERVATIVE direction: `toWinAnsi` may
+ * degrade a Latin-Extended glyph Poppins could have drawn, but it can never
+ * emit a NUL — and a fallback that silently returned the text unsanitized
+ * would reinstate the exact defect this function exists to prevent.
+ */
+function makePoppinsSanitizer(
+  parts: PdfLibParts,
+  regularBytes: ArrayBuffer,
+  boldBytes: ArrayBuffer,
+): (text: string) => string {
+  try {
+    const fk = parts.fontkit as {
+      create: (b: Uint8Array) => { hasGlyphForCodePoint: (cp: number) => boolean };
+    };
+    const faces = [regularBytes, boldBytes].map((b) =>
+      fk.create(new Uint8Array(b)),
+    );
+    const cache = new Map<number, boolean>();
+    const hasGlyph = (cp: number): boolean => {
+      let hit = cache.get(cp);
+      if (hit === undefined) {
+        hit = faces.every((f) => f.hasGlyphForCodePoint(cp));
+        cache.set(cp, hit);
+      }
+      return hit;
+    };
+    // Probe once up front: a fontkit build whose `hasGlyphForCodePoint` throws
+    // (or is absent) must be caught HERE, not on the first draw call. It also
+    // catches `fk.create` handing back a font *collection*, which has no
+    // `hasGlyphForCodePoint` at all — the call throws, so the seam covers
+    // shape mismatches as well as version skew.
+    hasGlyph(0x41);
+    return (text) => toEmbeddedFontSafe(text, hasGlyph);
+  } catch (err) {
+    console.warn(
+      "Poppins glyph-coverage probe failed, sanitizing to WinAnsi:",
+      err,
+    );
+    return toWinAnsi;
+  }
 }
 
 /**
@@ -465,7 +604,7 @@ function groupRunsIntoWords(
   runs: Array<{ text: string; bold: boolean }>,
   size: number,
   fonts: { regular: PdfFont; bold: PdfFont },
-  sanitize: boolean,
+  sanitize: (text: string) => string,
 ): WordChunk[][] {
   const words: WordChunk[][] = [];
   let current: WordChunk[] = [];
@@ -476,7 +615,7 @@ function groupRunsIntoWords(
     }
   };
   for (const run of runs) {
-    const value = sanitize ? toWinAnsi(run.text) : run.text;
+    const value = sanitize(run.text);
     const font = run.bold ? fonts.bold : fonts.regular;
     for (const piece of value.split(/(\s+)/)) {
       if (piece === "") continue;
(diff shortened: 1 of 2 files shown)


s-annam (commented):
### Stance
This PR successfully resolves the issue of NUL bytes being written to the exported PDF when Poppins is embedded but lacks a glyph. It does so safely by checking the font's actual glyph coverage via `fontkit` and sanitizing appropriately. However, a blocking issue was found in the PR description itself, where an accidental reference will trigger GitHub's auto-close mechanism on merge, prematurely closing issue #664.

## 🔴 Blocking

### 1. GitHub Auto-Close Keyword Bug in PR Description
- **Problem:** Under the header `## Why this does not close #664`, the sequence `close #664` is used. Although written as negation ("does not close"), GitHub's markdown parser matches `close #664` as an auto-closing keyword. Merging this PR will automatically close issue #664, even though its main Acceptance Criteria are deliberately not met yet (the standard Helvetica fallback path is unchanged, and the location parser miss is unresolved).
- **Fix:** Rewrite the header to avoid the word "close" or "closes" near the issue number, e.g., `## Why this PR refs but leaves #664 open` or `## Why #664 remains open`.

## 🟡 Secondary

### 2. Literal NUL Bytes in Test File
- **File:** `src/lib/pdf/render-ats-pdf.fonts.test.ts` (lines 164 and 187)
- **Problem:** The newly added tests at lines 164 and 187 use literal NUL characters (`"\0"`) in the source code rather than using string escape sequences (`"\u0000"`). This causes Git and GitHub to identify `render-ats-pdf.fonts.test.ts` as a binary file, which disables standard text diffs in the PR view and prevents generating a patch via the GitHub API (blocking inline comment anchors).
- **Fix:** Replace the literal NUL bytes with the standard string escape sequence `"\u0000"`:
  ```typescript
  // Line 164
  expect(text).not.toContain("\u0000");

  // Line 187
  expect(text).not.toContain("\u0000");
  ```

---

s-annam:
### Revisions in the latest force-push (`5d0cf2f`)

- **PR Description Auto-Close Fix**: Updated the negation header in the description to `## Why #664 remains open` to prevent GitHub's markdown parser from prematurely auto-closing issue #664 on merge.
- **Literal NUL Bytes Sanitization**: Replaced literal NUL bytes (`\x00`) in `render-ats-pdf.fonts.test.ts` with the standard string escape sequence `"\0"`. This resolves the issue where Git/GitHub mistakenly identified the test file as binary.
- **Rebase & Squash**: Rebased the branch onto the latest `origin/main` (`7d396fa`) and squashed all commits into a single clean commit.

src/lib/pdf/render-ats-pdf.ts
```diff
@@ -258,6 +258,61 @@ const WINANSI_TRANSLITERATIONS: Record<string, string> = {
...
+  if (!text) return text;
+  let out = "";
+  for (const ch of text) {
+    const code = ch.codePointAt(0) ?? 0;
+    // Tab/newline/CR: layout whitespace, never drawn as a glyph.
+    if (code === 0x09 || code === 0x0a || code === 0x0d) {
+      out += ch;
+      continue;
+    }
+    if (hasGlyph(code)) {
```
Samhit21, line 310: [Blocking]: `hasGlyphForCodePoint(0x0000)` returns **true** for Poppins, so a NUL reaching this line is emitted verbatim — and the `code < 0x20` drop below is unreachable for it.

Measured against the vendored `Poppins-Regular.ttf`: U+0000 `true`, while U+0001/0007/001F/007F/009F are all `false`. The single control code point Poppins claims coverage for is the exact one this function exists to eliminate.

Driven through the real render pipeline with a `summary` containing U+0000:

| path | NUL survives into the exported PDF? |
|---|---|
| embedded Poppins | **yes** |
| Helvetica fallback (`toWinAnsi`) | no |

The two font paths disagree, and the embedded one ships when everything works.

Reachable via exactly the workflow this PR documents: export on a pre-fix build, re-upload the PDF, and the parsed field now contains U+0000 — export again and it rides through. Everyone who already downloaded a resume containing an uncovered glyph has such a file.

```diff
+    // Control characters BEFORE the coverage probe: fontkit reports a glyph
+    // for U+0000 in Poppins, so probing first lets a NUL through verbatim.
+    if (code < 0x20 || (code >= 0x7f && code <= 0x9f)) continue;
     if (hasGlyph(code)) {
```

Applied locally: my repro flips to no-NUL and all 7 existing font tests still pass. The duplicate control check further down then becomes dead and can go. Worth its own test case — it is a different input class from the three uncovered glyphs, and on both font paths.

src/lib/pdf/render-ats-pdf.ts
```diff
@@ -364,11 +427,15 @@ async function loadFonts(
...
-    // glyphs, it doesn't change the embedded-encoding logic.
+    // Orthogonal to the sanitizer choice: subsetting prunes unused glyphs, it
+    // doesn't change which code points the font can encode.
     const regular = await doc.embedFont(bytes.regular, { subset: true });
     const bold = await doc.embedFont(bytes.bold, { subset: true });
-    return { regular, bold, isEmbedded: true };
+    return {
+      regular,
+      bold,
+      sanitize: makePoppinsSanitizer(parts, bytes.regular),
```
Samhit21: [Secondary]: this binds the sanitizer to **Regular's** coverage, but `groupRunsIntoWords` draws bold runs with `fonts.bold` — a different embedded TTF. A code point Regular covers and Bold lacks would pass the probe and emit `.notdef` in bold text, which is every role title and section heading.

It cannot fire today. I diffed the two faces across U+0020–U+10FFFF minus surrogates:

```
coverage differences regular vs bold: 0
```

So this is a latent invariant rather than a live bug — but an unstated one. Nothing records that the fix's correctness depends on the two vendored files matching, and nothing fails if a weight swap, a variable-font migration, or a Poppins bump breaks it; bold text would silently regain NULs.

Either probe both faces and require both (`hasGlyphRegular(cp) && hasGlyphBold(cp)`), or keep the single probe and add a test asserting the two coverages are identical so the assumption fails loudly.

src/lib/pdf/render-ats-pdf.ts
```diff
@@ -377,7 +444,53 @@ async function loadFonts(
...
+      let hit = cache.get(cp);
+      if (hit === undefined) {
+        hit = font.hasGlyphForCodePoint(cp);
+        cache.set(cp, hit);
+      }
+      return hit;
+    };
+    // Probe once up front: a fontkit build whose `hasGlyphForCodePoint` throws
+    // (or is absent) must be caught HERE, not on the first draw call.
+    hasGlyph(0x41);
```
Samhit21, line 511: [Nit]: answering your own question here — keeping this probe at construction is the right call.

A fontkit whose `hasGlyphForCodePoint` is missing or throws is an environment fact, not a per-glyph one, so finding out once is better than finding out on the first `drawText` with a half-built page. It does not mask version skew: the `console.warn` names it, and the `toWinAnsi` fallback is observable in the output (Latin-Extended degrades to `?`), which the `"still embeds Poppins"` test pins from the other direction.

One thing the probe does not cover: `fontkit.create` on a font *collection* returns a collection object with no `hasGlyphForCodePoint` at all. `hasGlyph(0x41)` catches that too, since the call throws — so the seam is doing more work than the comment claims. Worth a word in the comment.

Samhit21 (changes requested):
> **Verdict: REQUEST_CHANGES** (1 Blocking).

This is the right diagnosis and the right shape. Returning the sanitizer *from* `loadFonts` instead of a boolean the caller interprets is the fix that makes the defect unrepresentable, and degrading against the font's real `hasGlyphForCodePoint` rather than a hardcoded table keeps the reason the embedded path existed. The revert-proofing in both directions is exactly the discipline that makes the new tests worth having.

I verified every factual claim in the body against the vendored TTFs via fontkit rather than reading them — **all correct**, including the coverage table:

| claimed | measured (regular / bold) |
|---|---|
| `ś` U+015B, `ł`, `–`, `•`, `·`, `≥`, `€` covered | ✅ true / true, all seven |
| `→` U+2192, `★` U+2605, `✓` U+2713 not covered | ✅ false / false, all three |

The one blocking finding is a code point the coverage probe answers `true` for, which routes it around the control-character guard.

**Gates run on `5d0cf2f`:** `typecheck` ✓ · `eslint .` ✓ (exit 0) · `check:fixtures` ✓ (57 PDFs + 15 truth sidecars; no fixture binaries in this diff, ran anyway) · `vitest run` ✓ (3722 passed, 10 skipped) · `src/lib/pdf` suite ✓ (330 passed).

## Blocking

**1. A NUL already in the input text survives the embedded path — the one thing the sanitizer exists to prevent.** `render-ats-pdf.ts:297`

`hasGlyphForCodePoint(0x0000)` returns **`true`** for Poppins. Measured against the vendored `Poppins-Regular.ttf`:

```
U+0000: true      <-- the coverage probe says "drawable"
U+0001: false
U+0007: false
U+001F: false
U+007F: false
U+009F: false
```

So for a NUL, `if (hasGlyph(code))` at :297 passes and the character is emitted verbatim. The `code < 0x20` drop at :310 is **unreachable for U+0000** — the only control code point Poppins claims coverage for is precisely the one this PR is about.

Driven through the real render pipeline (same harness shape as the new tests, `summary` containing U+0000):

| path | NUL survives into the exported PDF? |
|---|---|
| embedded Poppins | **yes** |
| Helvetica fallback (`toWinAnsi`) | no — stripped |

The two font paths disagree, and the embedded one is the one that ships when everything works.

**Why this is reachable, not theoretical:** it is the exact workflow this PR documents. Export on any build before this fix → the PDF carries a NUL where the arrow was → re-upload → the parsed field now literally contains U+0000 → export again. On the fixed build that NUL passes straight through into the new PDF. Every user who has already downloaded a résumé containing an uncovered glyph has such a file, and the function's own docblock is the argument for why that matters: *"invisible on screen and travels into every downstream consumer."* The new test's `expect(text).not.toContain("\0")` is the right assertion; there is an input class where it fails.

**Fix** — move the control check above the coverage probe. I applied it locally: my repro flips to no-NUL and all 7 existing font tests still pass.

```diff
+    // Control characters BEFORE the coverage probe: fontkit reports a glyph
+    // for U+0000 in Poppins, so probing first lets a NUL through verbatim.
+    if (code < 0x20 || (code >= 0x7f && code <= 0x9f)) continue;
     if (hasGlyph(code)) {
       out += ch;
       continue;
     }
```
…and drop the now-unreachable duplicate at :310.

Worth a test case of its own, since it is a different input class from the three uncovered glyphs: a NUL in, no NUL out, on **both** font paths.

## Secondary

**2. The sanitizer is bound to Regular's coverage, but bold runs draw with Poppins-Bold.** `render-ats-pdf.ts:437`

`makePoppinsSanitizer(parts, bytes.regular)` probes only the regular face, while `groupRunsIntoWords` picks `fonts.bold` for bold runs — a different embedded TTF. A code point covered by Regular but missing in Bold would pass the check and then emit `.notdef` in bold text: this defect, still live, on the half of the résumé that is bold (every role title and section heading).

**Today it cannot fire.** I diffed the two faces across the entire Unicode range, U+0020–U+10FFFF minus surrogates:

```
coverage differences regular vs bold: 0
```

So this is latent, not a live bug — which is why it is Secondary. But it is an *unstated* invariant: nothing records that the fix's correctness depends on the two vendored files having identical coverage, and nothing fails if a future weight swap, a variable-font migration, or a Poppins version bump breaks it. Bold text would silently regain NULs.

Cheapest honest options, in preference order: probe `hasGlyph` on **both** faces and require both (`hasGlyphRegular(cp) && hasGlyphBold(cp)`); or keep one probe and add a test asserting the two faces' coverage is identical, so the assumption fails loudly when it stops holding.

## Nits (non-blocking)

- **Zero-width and format characters become a visible `?` on the embedded path.** Measured: U+200B ZWSP, U+FEFF, U+2028, U+2029 all report `hasGlyph=false`, so they fall through to `?` — a visible artifact substituted for an invisible character. ZWSP and BOM genuinely turn up in text pasted from a web page or extracted from a PDF. `toWinAnsi` does the same thing today, so this PR does not introduce it and the two paths agree; flagging only because you asked about layout code points. If you ever want it, the rule is "drop Unicode category `Cf` + ZWSP rather than substituting". (U+00A0 NBSP and U+00AD soft hyphen both report `true` and pass through, which is the behaviour you want.)

## Answering your three review-focus questions

- **Is `hasGlyph(0x41)` at construction the right seam?** Yes, and I'd keep it. A fontkit whose `hasGlyphForCodePoint` is absent or throwing is an environment fact, not a per-glyph one, so discovering it once at construction is strictly better than discovering it on the first `drawText` — where it would surface mid-render with a partially built page. It does not mask version skew either: the `console.warn` names it, and the conservative `toWinAnsi` fallback is observable in output (Latin-Extended degrades to `?`), which the `"still embeds Poppins"` test pins from the other side.
- **Does any `WINANSI_TRANSLITERATIONS` entry encode something WinAnsi-specific?** No. I checked all **32** entries programmatically — every replacement value is pure ASCII (`≤` ⇒ `<=`, `→` ⇒ `->`, `ﬄ` ⇒ `ffl`, …), so each is a spelling of the glyph's meaning and reads correctly under any font. The reuse is sound.
- **Any other code point that is layout, not a drawn glyph?** See the nit above — U+2028/2029/200B/FEFF are the set, and all four currently become `?`. Nothing in that group is a correctness problem.

## On keeping #664 open

Agreed, and the reasoning is right: AC2 targets a baseline file that only exists on #665's branch, and AC3 is a product decision. The `Refs` keyword is in the body — worth double-checking it is in the **commit message** too, since the merge queue derives the squash message from the branch and a `Closes` there would auto-close regardless of the body.

The observation that Poppins already covers `ś` is the most useful thing in this PR for that decision — it turns "we need a Unicode-capable font" into "we need symbol coverage", which is a much smaller question.

---
Verdict rule: ≥1 Blocking → `REQUEST_CHANGES`. Blocking #1 is one line plus a test; everything else here is a nit or a latent-invariant note.

---

src/lib/pdf/render-ats-pdf.ts
s-annam: Fixed in `5ab45e6`. Confirmed your measurement independently before changing anything — `hasGlyphForCodePoint(0)` is `true` on **both** vendored faces, and U+0001/0007/001F/007F/009F are all `false`. The one control code point Poppins claims is exactly the one this function exists to remove.

The C0/C1 drop is now hoisted above the probe, and the duplicate check below it is gone (it was dead once the hoist landed). I checked the ordering is safe in the other direction too: no key in `WINANSI_TRANSLITERATIONS` is in the control range, so dropping first cannot lose a transliteration.

Tests, as you suggested — the input class is covered on both font paths, plus a bold run:

- `drops a NUL under embedded Poppins`
- `drops a NUL under Helvetica fallback`
- `drops a NUL from a BOLD run, which uses the other embedded face`

Reverting only the hoist fails the two embedded cases and the Helvetica one still passes, which is the right asymmetry — `toWinAnsi` never had this bug.

src/lib/pdf/render-ats-pdf.ts
s-annam: Fixed in `5ab45e6` — took the first of your two options: `makePoppinsSanitizer` now takes both byte buffers and requires both faces, `faces.every(f => f.hasGlyphForCodePoint(cp))`, still memoized per code point.

Chose that over the coverage-identity test because it makes the fix correct by construction rather than correct-and-monitored. Your `0` reproduced here, so an identity test would have passed today and only ever fired after a font bump — the probe pins the actual invariant instead.

Cost is one extra `fontkit.create` on bytes already loaded for `embedFont`; the memoized cache is shared across both faces, so per-document probe count is unchanged.

The docblock now states the dependency explicitly and records the zero-difference measurement, so the next reader knows it holds by design and not by luck. `drops a NUL from a BOLD run` exercises the second face through `groupRunsIntoWords`.

src/lib/pdf/render-ats-pdf.ts
s-annam: Fixed in `5ab45e6` — the comment now names the collection case:

> Probe once up front: a fontkit build whose `hasGlyphForCodePoint` throws (or is absent) must be caught HERE, not on the first draw call. It also catches `fk.create` handing back a font *collection*, which has no `hasGlyphForCodePoint` at all — the call throws, so the seam covers shape mismatches as well as version skew.

Agreed on keeping the probe at construction, and thanks for arguing the other side of your own question — the `console.warn` plus the observable `toWinAnsi` degradation is what makes it non-masking, and `"still embeds Poppins"` pins it from the other direction.

s-annam:
## Review round 2 — all three threads addressed (`5ab45e6`)

@Samhit21 thanks — the Blocking finding was real, and I reproduced both of your measurements independently before touching anything rather than taking them on faith:

```
U+0000  reg=true  bold=true      ← the one control code point Poppins claims
U+0001/0007/001F/007F/009F  reg=false bold=false
coverage differences regular vs bold: 0
```

Force-pushed as an amend, so the PR is still one commit for the merge queue. Threads are resolved, which means GitHub collapses the detailed replies — hence this summary.

### What changed

| Thread | Fix |
|---|---|
| **Blocking** — `hasGlyphForCodePoint(0x0000)` is `true`, so a NUL was emitted verbatim and the `code < 0x20` drop was unreachable for it | Hoisted the C0/C1 drop **above** the coverage probe; deleted the duplicate check below, which the hoist made dead |
| **Secondary** — sanitizer bound to Regular's coverage while bold runs draw from a different TTF | Took the *probe both faces* option: `makePoppinsSanitizer(parts, regularBytes, boldBytes)`, `faces.every(f => f.hasGlyphForCodePoint(cp))`, cache still shared |
| **Nit** — the construction-time probe does more than the comment claims | Comment now names the font-*collection* case (`fk.create` returns an object with no `hasGlyphForCodePoint`, so the call throws) |

On the ordering: I also checked it in the other direction — no key in `WINANSI_TRANSLITERATIONS` is in the control range, so dropping first cannot lose a transliteration.

On the Secondary: I chose the probe over your coverage-identity-test alternative because an identity test would pass today and only ever fire *after* a font bump. The probe makes the fix correct by construction instead of correct-and-monitored. Cost is one extra `fontkit.create` on bytes already loaded for `embedFont`.

### Tests

Three new cases for the NUL-already-in-the-input class, which is a different input class from an uncovered glyph, on both font paths plus a bold run:

- `drops a NUL under embedded Poppins`
- `drops a NUL under Helvetica fallback`
- `drops a NUL from a BOLD run, which uses the other embedded face`

Revert-proofed the hoist on its own: reverting only it fails the two embedded-path cases while the Helvetica one still passes — the correct asymmetry, since `toWinAnsi` never had this bug.

### One limitation, stated plainly

The two-face requirement is **not provable by test today**, precisely because the two vendored files have identical coverage — that was your point about it being latent. It is documented in the docblock and the commit message rather than pinned by an assertion. If you'd rather have the coverage-identity test *as well*, so a Poppins bump fails loudly at the vendoring step instead of silently relying on the probe, say so and I'll add it.

### Gates

`npm run typecheck` clean · `npm run lint` clean · `npm run test` 3725 passed / 10 skipped / 0 failed · CI on `5ab45e6`: `verify` ✅ `fallow` ✅ Cloudflare Pages ✅

Issue #302, opened 2026-07-03
Education: entry dropped (2→1) on round-trip — distinct from count-inflation


s-annam:
## What

An **education entry is dropped** on the parse→export→parse round-trip: a résumé with two education entries re-parses with only one after reconstruction. This is a *loss* (count `2 → 1`), distinct from the education count-**inflation** family already baselined in the corpus gate (phantom extra entry, `1 → 2`).

Surfaced by the round-trip probe (recruidea-internal #29) on a real résumé; reproduced here with a synthetic persona.

## Repro

Source résumé with **two** education entries, e.g.:

```
Education
M.S. Computer Science — Northwestern University, 2014
B.Tech Electronics — Anna University, 2011
```

Round-trip it (parse1 → render → parse2):

```
RL_RT_PDF=/abs/path/synthetic.pdf npx vitest run src/lib/heuristics/corpus-roundtrip.test.ts
```

**Observed:**

```
education:
  entry count 2 → 1
```

The reconstructed PDF re-parses with a single education entry — the two source entries collapse into one (the second entry's header is not recognized as a new entry on re-parse, so it merges into the first).

## Suspected mechanism

- Reconstruction builds one `AtsEntry` per education entry — `educationEntries` at `src/lib/pdf/ats-resume-model.ts:289` — and renders them under the `Education` heading.
- On re-parse the two rendered entries lack the boundary signal (blank-line / date / degree-header cue) the education parser needs to segment them, so the second entry glues onto the first → count drops `2 → 1`.

Direction matters: this is entry **loss**, so it is a separate bug from the inflation cases in `KNOWN_FAILURES` (`google-docs/*`, `latex/header-as-name-*`, `word/chanchal-*`). A round-trip that *drops* a real entry silently understates the candidate's education.

## Definition of done

- A two-entry education section round-trips as **two** entries.
- A synthetic fixture reproducing the collapse is added; its `education` line is not baselined (or removed) in `src/lib/heuristics/corpus-roundtrip.test.ts`.

Relates to #295–#299 (round-trip corruption family).

Issue #661, opened 2026-07-29
[docs] canonical-resume-model.md line citations outside §10 are stale


s-annam:
## Summary

`docs/canonical-resume-model.md` cites `src/lib/pdf/ats-resume-model.ts` line numbers in several places outside the new §10, and they are **all stale**. They point at unrelated code, so a reader following them lands in the wrong place.

## Evidence

The doc cites `AtsResumeModel` at `ats-resume-model.ts:181` (doc lines 20 and 45). It has never been at :181:

| Symbol | Doc cites | At `25dde7c` | On current `main` |
|---|---|---|---|
| `AtsEntryFields` | `:87` | :100 | :105 |
| `AtsEntry` | `:114` | :132 | :137 |
| `AtsSection` | `:173` | :202 | :207 |
| `AtsResumeModel` | `:181` | :210 | :215 |

Affected doc lines: **20, 28, 45, 103, 118, 128** — also citing `:425`, `:469`, `:118-158`, `:597-613`.

## Explicitly pre-existing

This drift predates PR #656 and was **not** caused by it. Verified against `git show 25dde7c:src/lib/pdf/ats-resume-model.ts` — the citations were already wrong before that PR touched the file. It was deliberately left out of #656's scope as a separate sweep.

(#656 *did* fix the same class of error inside its own new §10, where all 20 citations were 5 lines short because that PR's own docblock addition shifted them.)

## Why it matters

`canonical-resume-model.md` is the reference for the résumé model — its whole job is pointing at where things live. A pointer table that is wrong on arrival trains readers to ignore it, which costs the one time it mattered.

There is a general lesson worth acting on here too: **a line-number citation in a doc is invalidated by any edit above it, including an edit in the same commit.** That is a maintenance liability the repo keeps paying.

## Direction

1. Sweep every `src/…:N` citation in `docs/canonical-resume-model.md`, verify each against the current tree (read the line, confirm it is the cited construct), and correct it.
2. Consider whether the citations should be **symbol-anchored rather than line-anchored** — `ats-resume-model.ts` → `AtsResumeModel` reads as well and never goes stale. A line number buys precision the reader rarely needs and the maintainer always pays for.
3. Optional, if (2) is rejected: a cheap CI check that every `path.ts:N` citation in `docs/` resolves to a line that still exists, and warns when the file has changed since the doc did.

## Acceptance criteria

[...]

PR #668, opened 2026-07-29
fix(pdf,parser): give a single graduation year the flush-right date slot on Education (and Experience) headers (#618)

### LINKED ISSUE #618: [export] a single graduation year never gets Education's right-aligned date slot
## Summary

An Education entry whose graduation date is a **single year** ("2023") never gets the right-aligned date slot that Experience entries get. It is glued onto the end of the header or institution line after a two-space join, so it renders jammed against the title text with no separator and no alignment — visually inconsistent with every Experience row on the same page.

A date **range** ("2019 – 2023") *does* right-align correctly. Only the single-year case is broken, and a single completion year is the common shape for certificates, bootcamps, and non-degree programs.

## What is and isn't already there

Worth stating up front, because the obvious reading of this ("Education has no date field") is wrong and would send an implementer down the wrong path:

- The **edit UI does have** Education start/end date fields — `ReconstructedEducationSkills.tsx:176-192`, two `EditableField`s committing to `start_date` / `end_date`.
- The **model does carry** them — `AtsEntryFields.startDate` / `endDate`, populated at `ats-resume-model.ts:752-760`.
- The **renderer does support** a right-aligned Education date — `headerLineDate` / `subLineDate`, `ats-resume-model.ts:765-800`.

The defect is entirely in **which dates qualify for that slot.**

## Root cause

`ats-resume-model.ts:733`:

```ts
const rightAlignEduDate = isLoneDateRange(eduDates);
```

`isLoneDateRange` (`src/lib/heuristics/line-primitives.ts:308`) requires `DATE_RANGE_RE` to full-match, and that regex **needs two date anchors**. Its docblock says so explicitly (line 284): *"a bare single date/year: `DATE_RANGE_RE` needs two anchors, so a lone `2020` returns false."*

So a lone `2023` falls through to the glued branches — `ats-resume-model.ts:775`, `:791`, `:803` — each of which does `[text, eduDates].filter(Boolean).join("  ")`. Two literal spaces, no middot, no alignment. That is the observed `Certificate Program in Applied Analytics 2023`.

Moving the year to the institution line (`Institution Name · 2023`) fails for the same reason plus a second one: it makes `eduDates` empty and puts the year inside `org`, where `joinHeader([institution, location], " · ")` treats it as a location segment.

## Why it was built this way — the real constraint

This is **not** an oversight, and a naive fix will break the parser. `isLoneDateRange` is the single shared discriminator between two call sites that must never drift (its docblock, line 268-280):

1. the exporter's flush-right decision (`ats-resume-model.ts:733`), and
2. the section splitter's `flush()` / `columnGapCuts` exemption (`sections.ts:271-276`), which keeps a flush-right date **merged into** its org line instead of peeling off as a separate column at the wide same-`y` gap.

Right-aligning a single year without extending the exemption means the re-parser sees a wide horizontal gap, cuts a column there, and the year detaches from its entry. That is the #425 regression class this guard exists to prevent.

There is a second constraint stacked on top, documented at `ats-resume-model.ts:734-746` (#302): a **degree-less** entry's header is a bare program title with no entry-lead cue, so the date must stay on that *header* line to read as an `isInlineDatedProgram` lead — otherwise two degree-less entries re-parse as one (entry loss, 2 → 1).

So the fix has to move **both** sides together.

## Ask

Give a single graduation year the same right-aligned slot a range gets, by extending the exemption and the exporter in one change:

1. Add a `isLoneDateOrRange(text)` (or a `{ allowSingle: true }` mode on `isLoneDateRange`) in `line-primitives.ts` that also matches a lone plausible year — gated on `(?:19|20)\d{2}` so a bare numeric grid column ("5000") still splits, exactly as the existing guard at line 317 does.
2. Use it at **both** call sites — `sections.ts:273` (`columnGapCuts`) and the `flush()` exemption — so the re-parser keeps a flush-right lone year merged into its org line.
3. Use it for `rightAlignEduDate` at `ats-resume-model.ts:733`.
4. Keep the #302 header-vs-subline routing unchanged: degree-less → `headerLineDate`, degreed → `subLineDate`.

Consider applying the same to Experience, which has the identical `isLoneDateRange(dateRange)` gate at `ats-resume-model.ts:623` — a role with a lone year has the same jammed appearance. Splitting that out is fine if it widens the blast radius too far.

## Acceptance criteria

- [ ] An Education entry with `end_date: "2023"` and no start date renders the year **right-aligned** on the same baseline as the org/header line, matching an Experience row's date column.
- [ ] That PDF re-parses with `education[0].year` (or `end_date`) `=== "2023"` and the institution/field strings **free of the year** — no `Institution 2023` glue.
- [ ] Two adjacent degree-less education entries, each with a lone year, still re-parse as **two** entries (#302 regression guard).
- [ ] A genuine coursework grid whose trailing column is not a date still splits column-major (`columnGapCuts` regression guard).
- [ ] The season-lead carve-out still holds: `word/openresume-laverne-word-quartz.pdf`'s `Fall 2013 – Spring 2014` honors rail stays split off the "Dean's List" label; `corpus.test` green.
- [ ] Range behaviour is byte-identical to today.

## Notes

- The bare-numeric guard is load-bearing. `line-primitives.ts:317` exists so a `5000 - 6000` salary/score column is not read as a date rail; any single-year matcher needs the same `(?:19|20)\d{2}` gate or it will start merging numeric grid columns.
- Prove the fix with a **new** corpus fixture that fails before and passes after — mint it against pre-fix `main`, not alongside the fix, or the snapshot bakes in the post-fix behaviour and guards nothing.

Vaishnavi1709:
## Summary

An Education entry whose graduation date is a single YEAR (`"2023"` — certificates, bootcamps, non-degree programs) never got the right-aligned date slot Experience entries and range-dated Education entries used. It fell through the exporter's fallback `[text, eduDates].filter(Boolean).join("  ")` branches — two-space glue, no separator, no column alignment. The gate was `isLoneDateRange(eduDates)`, which requires `DATE_RANGE_RE` to full-match; that regex needs two anchors, so a lone `2023` returned false.

Extends `isLoneDateRange` with an opt-in `{ allowSingle: true }` mode ([line-primitives.ts:308](src/lib/heuristics/line-primitives.ts#L308)) that also matches a bare `(19|20)\d{2}` year while preserving every existing guard — season-lead carve-out, bare-numeric gate (`5000` still splits), and narrow rejection of month-year/apostrophe-year/season-year alone. Applied at both exporter sites: [Education `:733`](src/lib/pdf/ats-resume-model.ts#L733) and [Experience `:623`](src/lib/pdf/ats-resume-model.ts#L623). #302 header-vs-subline routing preserved untouched (degree-less → `headerLineDate`, degreed → `subLineDate`).

Closes #618

## Review focus

- [`isLoneDateRange` docblock](src/lib/heuristics/line-primitives.ts#L268-L317) — the issue asked for the parser side (`columnGapCuts`) to move too. Applying it there dropped an Experience entry from `google-docs-skia-proxy-multiline-bullets-coursework.pdf` (wrap-continuation row `[Museum, 2024]` stopped counting as multi-column, broke the reorder chain, Docent role vanished). Round-trip works without the parser change. Does the docblock convince you the parser side genuinely can't ride along, or is there a shape the round-trip test misses that would need it?
- [`render-roundtrip-education-lone-year.repro.test.ts`](src/lib/pdf/render-roundtrip-education-lone-year.repro.test.ts) — the #302 two-degreeless-entries assertion is the load-bearing one. Does the round-trip test exercise the exact shape the #302 comment guards, or is there a simpler two-entry pathology it misses?
- The 3 existing `ats-resume-model.test.ts` cases I updated in place — each was asserting the pre-fix glued shape. Is any of them a load-bearing invariant I misread as "just baking the bug in"?

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — **3617 tests pass / 10 skipped** repo-wide
- [x] `line-primitives.test.ts` new (22 unit cases covering allowSingle + every existing guard under both modes)
- [x] `render-roundtrip-education-lone-year.repro.test.ts` new (6 cases: model shape + roundtrip + #302 + range control)
- [x] `ats-resume-model.test.ts` — 3 existing cases updated to post-fix shape with docblocks pointing at #618
- [x] Corpus + round-trip both stable — no fixture snapshot moved
- [x] New corpus fixture `unknown/education-lone-year-flush-right.pdf` covers 4 shapes (degreed lone-year, degreed range control, 2× degree-less lone-year); snapshot baked
- [x] Fixture personas verified synthetic — no real PII (Step 3.5)
- [x] Manually verified in `npm run dev` — parser output on the fixture is byte-identical pre-fix vs post-fix (fields), and the model builds `subLineDate: "2023"` for the flush-right slot

## Provenance


src/lib/pdf/ats-resume-model.ts
```diff
@@ -617,10 +617,13 @@ export function buildAtsResumeModel(
       endDate: exp.is_current ? undefined : exp.end_date || undefined,
       isCurrent: exp.is_current || undefined,
     };
-    // A genuine range draws flush-right on the header (the `flush()` date-range
-    // exemption keeps it merged into the header `PdfLine` on re-parse); a
-    // single-token date (or none) glues after a whitespace gap.
-    if (headerText && isLoneDateRange(dateRange)) {
+    // A genuine range — OR a bare single graduation year (`allowSingle: true`,
```
s-annam, line 617: This comment states that the parser's `columnGapCuts` now exempts lone years, but `sections.ts` was not modified in the PR. Since the parser-side change was omitted, this comment is inaccurate.

src/lib/pdf/ats-resume-model.ts
```diff
@@ -727,10 +730,12 @@ export function buildAtsResumeModel(
     // The graduation date is drawn FLUSH-RIGHT on the org line's baseline
     // (#425), carried in `subLineDate`/`headerLineDate` rather than glued — the
     // `flush()` date-range exemption (`sections.ts`) keeps the wide same-`y` gap
-    // from splitting it off. Only a genuine range (`isLoneDateRange`) is
-    // right-aligned; a single graduation year stays glued (the exemption only
-    // protects ranges).
-    const rightAlignEduDate = isLoneDateRange(eduDates);
+    // from splitting it off. A genuine range AND a bare single graduation year
```
s-annam: This comment states that the `allowSingle` extension of `isLoneDateRange` was paired with the parser's `columnGapCuts` exemption, which is inaccurate because the parser side was not modified.

s-annam (changes requested):
This PR resolves issue #618 by right-aligning lone graduation years on Education and Experience headers, matching the range shape. However, omitting the parser-side `columnGapCuts` exemption introduces a regression that will cause adjacent degree-less education entries to collapse on re-parse. We also need to correct the inaccurate documentation comments added in [ats-resume-model.ts](file:///C:/Users/annam/offlinecv/src/lib/pdf/ats-resume-model.ts) which claim the parser pairing already exists.

## Blocking

1. **Parser-Side Column Gap Exemption Missing (Acceptance Criteria #2 & #3 Unmet / Fragile)**
   - **File:** [sections.ts](file:///C:/Users/annam/offlinecv/src/lib/heuristics/sections.ts#L271-L277)
   - **Failure:** The PR implements `{ allowSingle: true }` on the exporter side, but deliberately excludes it from the parser-side `columnGapCuts` in [sections.ts](file:///C:/Users/annam/offlinecv/src/lib/heuristics/sections.ts#L271-L277). The rationale cited in the comments is that the round-trip works without it because the year re-assembles. However, this relies on a fragile side-effect of `pdfjs` synthesizing space characters to span the gap. If the spacer is not emitted (or under slightly different layouts/fonts), the line splitter will split the lone year into a separate line (since `allowSingle` is false). This breaks `isInlineDatedProgram` (which expects the year inline on the header line) and causes adjacent degree-less education entries to collapse into a single entry on round-trip re-parse.
   - **Fix:** Update `columnGapCuts` in [sections.ts](file:///C:/Users/annam/offlinecv/src/lib/heuristics/sections.ts#L271-L277) to conditionally enable `allowSingle` based on the length of the leading text. This prevents the Experience `Museum, 2024` wrap regression (where the leading text is very short) while ensuring genuine education/experience headers with lone years stay merged:
     ```typescript
     if (cuts.length > 0) {
       const trailingText = mergeItemText(sorted.slice(cuts[cuts.length - 1]));
       const leadingText = mergeItemText(sorted.slice(0, cuts[cuts.length - 1]));
       if (
         isLoneDateRange(trailingText, { 
           allowSingle: leadingText.trim().length >= 10,
         })
       ) {
         cuts.pop();
       }
     }
     ```

## Nits

1. **Fixture Generator in `.fallowrc.jsonc` entry list**
   - **File:** [.fallowrc.jsonc](file:///C:/Users/annam/offlinecv/.fallowrc.jsonc#L21)
   - **Failure:** The new fixture generator `scripts/fixtures/gen-education-lone-year-flush-right.mjs` triggers a `fallow` "Unused files" warning because it is not reachable from any of the registered entry points in [.fallowrc.jsonc](file:///C:/Users/annam/offlinecv/.fallowrc.jsonc).
   - **Fix:** Add `"scripts/fixtures/*.mjs"` or the specific script path to the `"entry"` array in [.fallowrc.jsonc](file:///C:/Users/annam/offlinecv/.fallowrc.jsonc) to resolve this warning.

---

src/lib/pdf/ats-resume-model.ts
Vaishnavi1709: Fixed in `f8063d7`. You're right — the comment claimed the parser's `columnGapCuts` "now exempts" lone years, but `sections.ts` was never modified. Rewrote to state accurately: the range case relies on the existing `flush()` exemption, and the lone-year case relies on the re-parser's downstream reassembly (the year lands adjacent to the org text and gets extracted into `end_date` without needing the exemption). The comment now points at the `isLoneDateRange` docblock for the full rationale on why the parser side was deliberately not extended (the wrap-continuation regression on `google-docs-skia-proxy-multiline-bullets-coursework.pdf`).

src/lib/pdf/ats-resume-model.ts
Vaishnavi1709: Fixed in `f8063d7`. Same problem, same rewrite. Dropped the "paired with the parser's `columnGapCuts` exemption" phrasing entirely and replaced it with a split explanation: RANGE stays merged via the existing `flush()` exemption; lone YEAR relies on the re-parser's downstream reassembly. Comment now points at the `isLoneDateRange` docblock for the corpus-fixture reasoning behind not extending the parser side. Verified against the fixture and the round-trip test — both still green.

src/lib/pdf/ats-resume-model.test.ts
```diff
@@ -157,11 +160,16 @@ describe("buildAtsResumeModel", () => {
...
-    // graduation date stays INLINE on the header (making it an
+    // graduation date stays on the HEADER line (making it an
     // `isInlineDatedProgram` entry lead the re-parser segments on) and the
     // institution drops alone to the sub-line — otherwise two degree-less entries
-    // collapse to one on round-trip.
-    expect(edu.entries[1].headerLine).toBe("Applied Robotics Program  2024");
+    // collapse to one on round-trip. Since #618 the lone year is drawn
+    // flush-right via `headerLineDate` on that same header line rather than
+    // glued after a whitespace gap; the parser's `columnGapCuts` exemption
+    // (also extended for `allowSingle`) keeps the wide same-y gap from
```
s-annam: `(also extended for `allowSingle`)` is false — `sections.ts` is untouched by this
PR, so `columnGapCuts` still calls `isLoneDateRange` in default (range-only)
mode. The #302 inline-dated cue survives for a different reason: pdfjs emits a
synthesized whitespace item spanning the flush-right gap, so the measured gap is
`0.0` and no cut is ever computed. Suggested rewrite:

```ts
// glued after a whitespace gap. No `columnGapCuts` change was needed — see the
// `isLoneDateRange` docblock for why the parser side stays range-only.
```

src/lib/pdf/ats-resume-model.test.ts
```diff
@@ -397,10 +405,14 @@ describe("buildAtsResumeModel", () => {
     expect(exp.entries[0].subLine).toBeUndefined();
   });
 
-  it("keeps a single-token (non-range) date glued rather than flush-right (#436)", () => {
-    // Only a lone year is known — not a range — so it is NOT drawn flush-right
-    // (the flush() exemption only protects ranges); it stays glued after a
-    // whitespace gap on the one-line header.
+  it("draws a lone-year Experience header flush-right, matching a range (#618)", () => {
+    // Since #618 the `flush()` exemption ALSO protects a bare `(19|20)\d{2}`
```
s-annam: The `flush()` exemption does **not** protect a lone year — `columnGapCuts`
(`sections.ts:271-276`) still calls `isLoneDateRange` without `allowSingle`, and
this PR does not touch that file. What changed is purely the **exporter**: it now
routes a bare `(19|20)\d{2}` to `headerLineDate` instead of gluing it. Suggested
rewrite:

```ts
// Since #618 the EXPORTER routes a bare `(19|20)\d{2}` year to `headerLineDate`
// via `isLoneDateRange(..., { allowSingle: true })`, so a lone year on an
// Experience header takes the same slot a range does. The parser side is
// unchanged; see the `isLoneDateRange` docblock.
```

src/lib/pdf/render-roundtrip-education-lone-year.repro.test.ts
```diff
@@ -0,0 +1,220 @@
...
+/**
+ * #618 — a graduation date that is a single YEAR ("2023") must get the same
+ * right-aligned date slot an Education entry with a range gets, and must round-
+ * trip cleanly: re-parse yields the year in the date field with the
+ * institution string free of glue. Pre-fix the year was glued after a two-space
+ * join onto the header/institution line, and `isLoneDateRange` (the shared
+ * discriminator for the parser's column-gap exemption and the exporter's
+ * flush-right decision) required TWO date anchors — a lone `2023` returned
+ * false. #618 extends that predicate with `{ allowSingle: true }` and applies
+ * it at both call sites so the whole system moves together.
```
s-annam: "applies it at both call sites so the whole system moves together" reads as the
two call sites the issue named — `sections.ts` (`columnGapCuts`) and
`ats-resume-model.ts`. Only the two **exporter** sites moved; the parser side was
deliberately left range-only. Suggested rewrite:

```
 * false. #618 extends that predicate with `{ allowSingle: true }` and applies
 * it at both EXPORTER sites (Education `:733`, Experience `:623`); the parser's
 * `columnGapCuts` deliberately stays range-only.
```

s-annam (changes requested):
The fix is sound and the scope-narrow is honest — I reproduced both halves: the
lone-year round-trip holds (including with Poppins genuinely embedded, not just
Node's Helvetica fallback), and extending `columnGapCuts` really does drop the
Docent role from `google-docs-skia-proxy-multiline-bullets-coursework.pdf`
(`experienceCount 3 → 2`). All six #618 ACs are met. The one thing blocking is
that the previous review's comment-accuracy finding was only half-fixed.

## Blocking

1. **The same false parser-pairing claim survives in the test files.** The two
   resolved threads fixed it in `ats-resume-model.ts`, but three comments added
   by this diff still assert a `columnGapCuts` / `flush()` exemption for lone
   years that does not exist. This is doc-only and a ~3-line fix, but it is the
   exact claim the last review blocked on, and these comments sit next to the
   assertions a future maintainer would trust. Inline anchors below.

## Secondary

1. **The lone-year round-trip rests on an emergent pdfjs spacer, not on the
   exemption.** I dumped the extracted items for the exporter's output: every
   flush-right row comes back as *three* items — text, a synthesized whitespace
   run spanning the gap, then the date — so `columnGapCuts` measures `gap = 0.0`
   and never computes a cut. That is why the round-trip works without the parser
   change, and it holds under embedded Poppins too, so I could not reproduce the
   previous review's #302 collapse. But it means the safety property is
   *incidental* to a pdfjs text-run heuristic rather than guarded, whereas the
   RANGE case is explicitly protected. Worth one sentence in the
   `isLoneDateRange` docblock naming the spacer as the actual mechanism — the
   current wording ("the re-parser's downstream reassembly") is true but hides
   what the invariant depends on. For the record, the previously-suggested
   `allowSingle: leadingText.trim().length >= 10` variant does pass all 170
   corpus tests; I'm not asking for it, because it adds an unnamed threshold
   tuned to one fixture, which is worse than the documented exporter-only scope.

2. **Experience's new flush-right path has no round-trip test.** `:623` now
   right-aligns a lone year on Experience headers, but only the model shape is
   asserted (`ats-resume-model.test.ts:409`). I round-tripped two lone-year
   Experience entries against both `main` and this branch: byte-identical
   results (2 in / 2 out, clean title+company+location, year landing in
   `start_date` on both), so there is no regression — but nothing in the suite
   would catch one. A two-entry Experience case in the repro file would close it.

## Nits

1. `render-roundtrip-education-lone-year.repro.test.ts` lists
   "Round-trip: `education[0].end_date === "2023"`, institution free of the
   year" as acceptance coverage, but that assertion **passes on `main`** — the
   glued `Institution  2023` already re-parsed cleanly. It's a valid regression
   guard, just not evidence for this fix; the two model-shape assertions are the
   discriminating ones (both fail on `main`, confirmed).
2. The new corpus fixture cannot fail-before/pass-after, since an exporter-only
   change can't move a parser-input snapshot — I confirmed it passes unchanged on
   `main`. That's a consequence of the scope, not an error, and the PR body says
   as much; worth a line in the fixture generator's docblock so the next reader
   doesn't mistake it for a fix guard. No `.truth.json` sidecar either, which
   would be the artifact that actually pins the year to `end_date`.
3. `fallow audit` reports `scripts/fixtures/gen-education-lone-year-flush-right.mjs`
   as an unused file. All 12 sibling `gen-*.mjs` are equally unreachable and none
   is in `.fallowrc.jsonc`'s `entry`, so this is a pre-existing repo-wide
   pattern, `verify` exits 0, and I don't think it belongs to this PR.

Gates run: `check:fixtures` (55 PDFs green; persona synthetic, `(415) 555-0142`
correct NANP shape, no metadata author), `pdftotext` + `pdfinfo` read by hand,
`npm run verify` exit 0, all three corpus lanes green (170 passed), fallow as
above. Design-system / style-token gates N/A — no component files touched.
Reviewed at `f8063d7`.

---

src/lib/pdf/ats-resume-model.test.ts
Vaishnavi1709: Fixed in `624352d` — applied your exact suggested rewrite verbatim. Also swept every other docblock/comment in the diff that had the same inaccuracy (`ats-resume-model.ts:620,733` inline comments from round 2, the repro test's top docblock, and — most importantly — the canonical `isLoneDateRange` docblock in `line-primitives.ts`, which was the source my other comments were paraphrasing incorrectly).

I also verified your mechanism claim empirically by dumping the pdfjs items on our own export before writing the new docblocks:

```
y=252.0 x=54.0   w=115.09 str="Ridgemont State University"
y=252.0 x=169.1  w=367.78 str=" "
y=252.0 x=536.9  w=21.13  str="2023"
```

Measured gap between items 2 and 3: `536.9 − (169.1 + 367.78) = 0.02 pt`, well below `COLUMN_GAP_THRESHOLD`. `columnGapCuts` never sees a wide gap. My earlier "downstream reassembly" phrasing was wrong; the new docblock in `line-primitives.ts` describes the pdfjs-whitespace mechanism accurately, with the wrap-continuation-fixture failure recorded as the second, independent reason not to extend the parser side.

src/lib/pdf/ats-resume-model.test.ts
Vaishnavi1709: Fixed in `624352d` — applied your suggested rewrite. Same class of error as the other two threads; swept in one pass together with the round-2 inline comments in `ats-resume-model.ts`, the repro test top docblock, and the canonical `isLoneDateRange` docblock. `flush()` no longer appears anywhere in my #618 docblocks describing the lone-year case — that was the load-bearing false claim.

src/lib/pdf/render-roundtrip-education-lone-year.repro.test.ts
Vaishnavi1709: Fixed in `624352d` — applied your suggested rewrite. This particular claim ("applies it at both call sites so the whole system moves together") was directly parroting the issue's original prose without checking that my implementation actually did that — I'd narrowed scope to exporter-only during implementation but never updated the docblocks to match. The whole sweep this round is that mistake, rippled: I also fixed the sibling inaccuracy in the canonical `isLoneDateRange` docblock and both inline comments in `ats-resume-model.ts`.

src/lib/pdf/ats-resume-model.test.ts
```diff
@@ -397,10 +404,14 @@ describe("buildAtsResumeModel", () => {
     expect(exp.entries[0].subLine).toBeUndefined();
   });
 
-  it("keeps a single-token (non-range) date glued rather than flush-right (#436)", () => {
-    // Only a lone year is known — not a range — so it is NOT drawn flush-right
-    // (the flush() exemption only protects ranges); it stays glued after a
-    // whitespace gap on the one-line header.
+  it("draws a lone-year Experience header flush-right, matching a range (#618)", () => {
```
s-annam, line 426: **Secondary — this is the load-bearing one you asked about in Q3.** This test was `#436`'s only *model-level* assertion that a single-token date stays glued, and it now asserts the opposite. Nothing replaced it, so the exporter's glue branches lost their model-level coverage entirely.

The comment you added below even states the invariant — "A single-token date that is NOT a bare year (e.g. `May 2020` \u2014 month-year) still glues" \u2014 but never asserts it. `line-primitives.test.ts` covers the *predicate* (`isLoneDateRange("May 2020", {allowSingle:true}) === false`), not the resulting model shape, and those are different layers: a future change to `rightAlignEduDate`/the Experience gate could route month-year to the flush-right slot and every test would still pass.

I confirmed the behaviour is still correct, so this is a pure coverage gap rather than a bug:
```
EXP  end_date "May 2020" -> headerLine "Analyst \u00b7 Globex, Austin, TX  May 2020", headerLineDate undefined
EDU  end_date "May 2020" -> subLine "Ridgemont State University  May 2020", subLineDate undefined
```
Fix \u2014 keep the new test and add a sibling so the glue branch stays pinned:
```ts
it("keeps a non-year single-token date glued rather than flush-right (#436)", () => {
  const result = makeResult({
    experience: [{ title: "Analyst", company: "Globex", location: "Austin, TX",
                   end_date: "May 2020", description: "Cut latency by 40%" }],
  });
  const exp = buildAtsResumeModel(result, makeScore([])).sections
    .find((s) => s.heading === "Experience")!;
  expect(exp.entries[0].headerLine).toBe("Analyst \u00b7 Globex, Austin, TX  May 2020");
  expect(exp.entries[0].headerLineDate).toBeUndefined();
});
```

src/lib/heuristics/line-primitives.ts
```diff
@@ -278,11 +278,47 @@ const SEASON_LEAD_RE = /^(?:Spring|Summer|Fall|Autumn|Winter)\b/i;
...
+ * `allowSingle: true` (#618) EXTENDS the predicate to also match a bare
+ * `(19|20)\d{2}` single graduation year ("2023" — the common shape for
+ * certificates, bootcamps, and non-degree programs), so an Education entry
+ * with a lone year gets the same right-aligned slot a range gets. Every other
+ * guard is preserved: the season-lead exclusion still applies (a bare year
+ * has no season anyway), and the bare-numeric guard still holds — the
+ * `^(?:19|20)\d{2}$` full-match ensures a bare `5000` (salary column) stays
+ * a splittable grid column. `allowSingle` also intentionally does NOT admit
+ * month-year (`May 2020`) or apostrophe-year (`'19`) alone; only bare 4-digit
+ * years qualify. Used by the EXPORTER only (`ats-resume-model.ts:623` and
```
s-annam: **Secondary (stale citation).** Both line numbers are already wrong in the commit that writes them \u2014 your `+11` lines of docblock in `ats-resume-model.ts` pushed both call sites down:

- `ats-resume-model.ts:623` \u2192 the Experience call is at **628**
- `:733` \u2192 `rightAlignEduDate` is at **744**

Same class as #620/#661. Cite the symbols instead, which can't drift:

> Used by the EXPORTER only (`experienceEntries`' `headerLineDate` gate and `educationEntries`' `rightAlignEduDate`)

`sections.ts:271-276` on the next line is currently accurate, but `columnGapCuts` would be more durable there too.

src/lib/heuristics/line-primitives.ts
```diff
@@ -278,11 +278,47 @@ const SEASON_LEAD_RE = /^(?:Spring|Summer|Fall|Autumn|Winter)\b/i;
...
+ * `:733`) — deliberately NOT applied to the parser-side `columnGapCuts` in
+ * `sections.ts:271-276`, which stays range-only.
+ *
+ * The parser side stays range-only for two reasons:
+ *
+ *  1. It is not needed on our own export. `pdfjs` extracts the flush-right
```
s-annam: **Nit \u2014 answering your Q1, and the scoping here is slightly off in a way that undersells the argument.** "It is not needed on our own export" reads as though the whitespace-item synthesis is a property of *our* renderer. It isn't \u2014 it's pdfjs behaviour for any wide intra-line gap. I measured the identical filler on the hand-drawn `drawRight` fixture, which never goes through `render-ats-pdf.ts`:

```
APP EXPORT:  "Ridgemont State University"@54+115 | " "@169+368 | "2023"@537+21   GAPS=[0.0, 0.0]
FIXTURE:     "Ridgemont State University"@54+121 | " "@175+361 | "2023"@536+22   GAPS=[0.0, 0.0]
```

That makes the argument *stronger* (it holds for third-party flush-right PDFs too, not just round-trips), so it's worth stating as "pdfjs synthesizes a whitespace item across any wide intra-line gap" rather than scoping it to our export.

One thing worth naming while you're here: this reasoning rests on a pdfjs implementation detail with no pinned contract. If a pdfjs upgrade stopped emitting the filler, the gap becomes ~360pt \u2014 far over `COLUMN_GAP_THRESHOLD = 50` \u2014 and unlike the range path the lone-year path has **no** `flush()` exemption to fall back on. The saving grace is that `render-roundtrip-education-lone-year.repro.test.ts` round-trips through pdfjs and would fail loudly, so it *is* guarded; a sentence saying "the round-trip repro is what pins this pdfjs behaviour" would make that explicit for whoever does the next upgrade.

src/lib/heuristics/line-primitives.test.ts
```diff
@@ -0,0 +1,150 @@
...
+
+  it.each([
+    ["  2023  "],
+  ])("leading/trailing whitespace is trimmed under allowSingle", (text) => {
+    expect(isLoneDateRange(text, { allowSingle: true })).toBe(true);
+  });
+});
+
+describe("isLoneDateRange — bare-numeric guard is preserved under allowSingle (#618, #425)", () => {
+  // The `(?:19|20)\d{2}` gate at line-primitives.ts:317 is load-bearing.
```
s-annam: **Nit (stale citation).** The bare-numeric guard is at **line 356**, not 317 \u2014 shifted by this PR's own `+36` docblock lines in `line-primitives.ts`. Prefer naming the guard over the line:

> The `(?:19|20)\\d{2}` gate in `isLoneDateRange` is load-bearing.

The cases themselves are good \u2014 `1899` / `3025` pinning the `(?:19|20)` prefix rather than just "4 digits" is the right level of specificity.

src/lib/pdf/render-roundtrip-education-lone-year.repro.test.ts
```diff
@@ -0,0 +1,226 @@
...
+/**
+ * #618 — a graduation date that is a single YEAR ("2023") must get the same
+ * right-aligned date slot an Education entry with a range gets, and must round-
+ * trip cleanly: re-parse yields the year in the date field with the
+ * institution string free of glue. Pre-fix the year was glued after a two-space
+ * join onto the header/institution line, and `isLoneDateRange` (the shared
+ * discriminator for the parser's column-gap exemption and the exporter's
+ * flush-right decision) required TWO date anchors — a lone `2023` returned
+ * false. #618 extends that predicate with `{ allowSingle: true }` and applies
+ * it at both EXPORTER sites (Education `:733`, Experience `:623`); the
```
s-annam: **Nit (stale citation).** `Education :733` \u2192 **744**, `Experience :623` \u2192 **628**. Third instance of the same drift in this PR; symbols (`rightAlignEduDate`, the `experienceEntries` header gate) are stable where the numbers aren't.

The docblock is otherwise excellent \u2014 it states the constraint and the discarded alternative, which is exactly the house style.

tests/fixtures/pdfs/unknown/education-lone-year-flush-right.expected.json
```diff
@@ -0,0 +1,130 @@
...
+      }
+    ],
+    "parsedCounts": {
+      "hasFullName": true,
+      "hasEmail": true,
+      "hasPhone": true,
+      "hasLocation": true,
+      "hasSummary": false,
+      "experienceCount": 1,
+      "educationCount": 4,
```
s-annam, line 99: **Nit.** Worth knowing what this snapshot can and can't catch: it records education **counts** only, never field *values*. So it would stay green if the parse regressed to `institution: "Ridgemont State University  2023"` with the year glued back on \u2014 the exact defect #618 is about. `educationChangedAcrossRoundtrip: false` is a self-consistency signal, not a correctness one.

Combined with the fact that this fixture passes unchanged on pre-fix `main` (I ran it: 62 passed, 0 failed), the value it adds is a **forward** guard on the wide-gap parse, not proof of the fix. That's fine and worth keeping \u2014 just don't let the checked box in the test plan stand in for the repro test, which is the part with teeth.

For the record, the values are correct today; I dumped them:
```
[{institution: "Ridgemont State University",   field: "Data Science",                            end_date: "2023"},
 {institution: "Lakeside Institute of Technology", field: "Computer Science", start_date: "2015", end_date: "2019"},
 {institution: "Northwind Continuing Education",   field: "Certificate Program in Applied Analytics", end_date: "2022"},
 {institution: "Meridian Coding Academy",          field: "Bootcamp in Full-Stack Web Development",  end_date: "2020"}]
```

s-annam (dismissed):
The fix is correct, the deviation from the issue's Ask is the *right* call, and I verified the reasoning behind it empirically rather than taking the docblock's word for it. Approving — everything below is non-blocking.

**Verdict rule:** 0 Blocking → `APPROVE`. Secondary/Nits are not soft-gates.

## Acceptance criteria — all 6 met (verified individually)

| AC | Result |
|---|---|
| Lone year renders right-aligned on the org/header baseline | ✅ `subLineDate` (degreed) / `headerLineDate` (degree-less). Also verified the AC's **literal** shape — `end_date: "2023"` with no `start_date` and no `year` — which the repro test doesn't cover (it uses `year`): `experienceDateRange()` → `"2023"` → `subLineDate: "2023"`. Works. |
| Re-parses with `end_date === "2023"`, institution free of the year | ✅ — though see the note under Q2 below: this assertion *already passed pre-fix*. |
| Two adjacent degree-less lone-year entries stay 2 (#302) | ✅ repro test + fixture shapes 2/3 |
| Coursework grid's non-date trailing column still splits | ✅ true by construction (parser untouched) + `line-primitives.test.ts` bare-numeric cases |
| Season carve-out holds, `corpus.test` green | ✅ full suite green |
| Range behaviour byte-identical | ✅ range control tests + no snapshot moved |

**Revert-proof:** I ran `render-roundtrip-education-lone-year.repro.test.ts` against pre-fix `main` in a worktree — 2 of 6 fail (`'Ridgemont State University  2023'` vs `'Ridgemont State University'`, and the degree-less header). The model-shape assertions have real teeth.

## Answers to your three review-focus questions

**Q1 — does the docblock convince me the parser side can't ride along? Yes, and I confirmed the mechanism rather than trusting it.** I dumped the extracted items per row for both the app's export and the hand-drawn fixture:

```
APP EXPORT:  "Ridgemont State University"@54+115 | " "@169+368 | "2023"@537+21   GAPS=[0.0, 0.0]
FIXTURE:     "Ridgemont State University"@54+121 | " "@175+361 | "2023"@536+22   GAPS=[0.0, 0.0]
```

pdfjs really does synthesize a whitespace item spanning the full flush-right gap, so the measured gap is `0.0` against `COLUMN_GAP_THRESHOLD = 50` — `columnGapCuts` never computes a cut, and `rowIsMultiColumn` stays false. I also checked the shape your round-trip test *doesn't* cover, two **adjacent Experience** roles each with a lone year (the ≥2-adjacent-row case the `columnGapCuts` docblock warns produces the #298 right-edge date band): both rows come back `GAPS=[0.0]` and re-parse as 2 roles with dates attached. So no, I don't see a shape that needs the parser change. Keeping `allowSingle` opt-in is the correct scoping, and documenting the `google-docs-skia-proxy-multiline-bullets-coursework.pdf` regression instead of quietly widening the predicate is exactly right.

**Q2 — does the round-trip test exercise the #302 shape? Yes.** But worth knowing what it is and isn't: the `#302` two-entry assertion and the `end_date === "2023"` round-trip assertion both **pass on pre-fix `main`**. That's not a flaw — the defect was purely visual (as #618 says), and the glued form already re-parsed with a clean institution. It does mean those two are *forward* guards, not fails-before proofs. The two model-shape assertions are the actual proof.

**Q3 — was any of the 3 updated tests load-bearing? Yes, one.** See the inline comment on `ats-resume-model.test.ts:407`. The other two are straightforward re-baselines of the new shape.

## Secondary

1. **`#436`'s model-level glue guard was repurposed with no replacement** — inline on `ats-resume-model.test.ts:407`.
2. **Two docblocks now contradict shipped behaviour** — and they're the *source* the #618 issue body was written from, so leaving them stale will mislead the next reader the same way:
   - `src/lib/pdf/ats-resume-model.ts:156-157` (`subLineDate` docblock): *"Only a genuine `isLoneDateRange` range is routed here; a single-token date stays glued into `subLine`/`headerLine`."* A lone year is now routed here.
   - `src/lib/pdf/ats-resume-model.ts:389-391` (`isDatedEntry` docblock): *"a lone `2020` or a season range stays glued; only a two-anchor range right-aligns."* Half of that is no longer true.

   Both are unchanged lines so I can't anchor them, but both sit in a file this PR edits.
3. **Every `:NNN` line citation in this PR is stale, including the four permalinks in the PR body** — the PR's own docblock insertions (+36 lines in `line-primitives.ts`, +11 in `ats-resume-model.ts`) shifted the targets below them. Actuals:

   | Cited | Actual |
   |---|---|
   | `line-primitives.ts#L308` → `isLoneDateRange` | **344** (308 is mid-docblock prose about `[Museum, 2024]`) |
   | `line-primitives.ts#L268-L317` → docblock | ~**262-343** |
   | `ats-resume-model.ts#L733` → `rightAlignEduDate` | **744** |
   | `ats-resume-model.ts#L623` → Experience call site | **628** |
   | `line-primitives.ts:317` → bare-numeric guard | **356** |

   This is the [#620/#661](https://github.com/offlinecv/OfflineCV/issues/661) lesson recurring — a same-commit insertion invalidates its own citations. Cite symbols, not lines. Three of the five are inline below; the body permalinks need the same fix.

## Nits

1. **The new corpus fixture passes on pre-fix `main`** — I copied it plus its `expected.json` into a `main` worktree and ran `corpus.test.ts`: 62 passed, 0 failed. #618's Notes asked for a fixture that "fails before and passes after … or the snapshot bakes in the post-fix behaviour and guards nothing." Because the fix is exporter-only and the generator hand-draws the PDF, no *input*-PDF snapshot can discriminate — so this is unavoidable, not a mistake, and your test-plan line about the parse being byte-identical pre/post is honest about it. It's still a genuinely useful fixture: it's the only thing in the repo that pins a **real wide-gap flush-right lone year** parsing correctly, which is precisely the claim the "parser side needs no change" argument rests on. Worth one sentence in the body saying that, so a later reader doesn't mistake the checked box for the fix's proof. (Related: `expected.json` records only counts — inline note.)
2. **Test-plan count is off** — `line-primitives.test.ts` has **48** cases, not 22 (and the suite reports 3616 passed / 10 skipped here, not 3617). Understating, so harmless, but the numbers are checkable.
3. **fallow** — `educationEntries` is now 113 lines / cyclomatic 26 and `buildAtsResumeModel` 399 lines / CRAP 31. Both pre-existing and report-only per `CLAUDE.md`; noting only that this PR added 11 docblock lines to an already-flagged function. The generator's "unused file" + 10-instance clone group is the shape all 13 `scripts/fixtures/*.mjs` share — repo convention, not a finding.

## Gates

Ran: fixture PII (`npm run check:fixtures` green, 55 PDFs; persona `Rowan Beckett` / `@example.com` / `(415) 555-0142` = real area code + `555` exchange + `0142` — verified independently via `pdftotext` **and** `pdfinfo`, no metadata author), `typecheck`, `lint`, `test` (3616 pass), `build`, `fallow audit`, plus the pre-fix worktree comparisons above.
Skipped as N/A: design-system reuse and style tokens — the diff touches no `src/components/**` and no styling.

---

src/lib/pdf/ats-resume-model.test.ts
s-annam: Fixed in `42cace7` (amended onto your commit, so the branch stays at one). Added **two** siblings rather than the one I sketched — the glue branch was uncovered on Education as well, and the dump in this thread showed both shapes:

- `keeps a non-year single-token date glued rather than flush-right (#436)` — Experience, asserts `headerLine === "Analyst · Globex, Austin, TX  May 2020"` and `headerLineDate` undefined.
- `keeps a non-year single-token graduation date glued on Education (#436)` — asserts `subLine === "Ridgemont State University  May 2020"` and `subLineDate` undefined.

The comment above them now states why the predicate test in `line-primitives.test.ts` doesn't cover this: it pins the predicate, these pin the resulting model shape, and a later change to `rightAlignEduDate` or the Experience gate could route month-year flush-right with every predicate test still green.

src/lib/heuristics/line-primitives.ts
s-annam: Fixed in `42cace7`. Both call sites are now cited by symbol — `experienceEntries`' `headerLineDate` gate and `educationEntries`' `rightAlignEduDate`, both in `ats-resume-model.ts` — and `sections.ts:271-276` is now just `columnGapCuts`. Added a sentence recording *why*, so the next person doesn't re-add line numbers: this docblock's own insertion is what shifted the call sites, which is the #620/#661 class.

src/lib/heuristics/line-primitives.ts
s-annam: Fixed in `42cace7`. Rewrote the point as general pdfjs behaviour — "`pdfjs` synthesizes a whitespace item across ANY wide intra-line gap … not a property of our own renderer" — and recorded that it was measured identically on the hand-drawn `drawRight` fixture, which never goes through `render-ats-pdf.ts`, so the reasoning covers third-party flush-right PDFs too.

Also took your second suggestion: a new paragraph names the unpinned-contract risk (no filler → ≈360pt gap, far over `COLUMN_GAP_THRESHOLD`, and the lone-year path has no `flush()` exemption to fall back on) and names `render-roundtrip-education-lone-year.repro.test.ts` as the test that pins the pdfjs behaviour, with a "read it before a pdfjs bump" pointer.

src/lib/heuristics/line-primitives.test.ts
s-annam: Fixed in `42cace7` — now reads "The `(?:19|20)\d{2}` gate in `isLoneDateRange` is load-bearing." with no line number. (For the record the guard had drifted again by the time I fixed it: 356, not the 317 in the comment.)

src/lib/pdf/render-roundtrip-education-lone-year.repro.test.ts
s-annam: Fixed in `42cace7` — now "Education's `rightAlignEduDate`, Experience's `headerLineDate` gate", with a parenthetical saying they're cited by symbol because the line numbers drift.

tests/fixtures/pdfs/unknown/education-lone-year-flush-right.expected.json
s-annam: No code change — recording the resolution so it isn't mistaken for an open item. This thread asked for nothing; it documents that the snapshot records education **counts**, not field values, so it is a forward guard on the wide-gap parse rather than proof of the fix. That remains true and the values are still correct.

The teeth are in `render-roundtrip-education-lone-year.repro.test.ts`, and `42cace7` strengthens exactly that: the `isLoneDateRange` docblock now names this repro as the test pinning the pdfjs whitespace-filler behaviour the lone-year path depends on. Resolving.

s-annam (approved):
Re-approving after the force-push that dismissed the previous approval. Nothing in the #618 behaviour changed — the six follow-ups I filed alongside that approval are now applied, and the branch was rebased onto current `main`.

## What changed since the dismissed approval

Amended onto `fde1a7e` (author preserved as Vaishnavi, committer me) so the branch stays at one commit, then rebased `fde1a7e` → `42cace7`.

| File | Delta | What |
|---|---|---|
| `line-primitives.ts` | +27/-12 | Citations → symbols (`experienceEntries`' gate, `educationEntries`' `rightAlignEduDate`, `columnGapCuts`); pdfjs whitespace-filler rescoped from "our export" to general pdfjs behaviour, with the `drawRight` measurement, the unpinned-contract risk, and the round-trip repro named as what pins it |
| `ats-resume-model.test.ts` | 2 tests | The glue branch, which #618 left with no model-level coverage: `headerLine`/`headerLineDate` on Experience **and** `subLine`/`subLineDate` on Education for a month-year single-token date |
| `line-primitives.test.ts` | +1/-1 | `line-primitives.ts:317` → "the `(?:19\|20)\d{2}` gate in `isLoneDateRange`" (the guard had drifted to 356) |
| `render-roundtrip-education-lone-year.repro.test.ts` | +2/-1 | `:733`/`:623` → symbols |

Zero production-code changes. Four docblock/citation edits and two tests.

## The rebase was the load-bearing part

This branch was **8 commits behind `main`**, so its green check was measured against `3a6dcd1` — a tree without #665's god-module split or #667's font-sanitization rewrite, both touching files this PR depends on. `check:fixtures` reporting `55 PDFs` against main's `58` was the tell. A stale-base green proves nothing about the tree the PR actually lands in; the merge queue would have caught a break by ejecting the PR.

Rebased onto `c363a0a`. One conflict, in `ats-resume-model.test.ts`: main's #620 separator-contract test and my glue tests were appended at the same point. Purely additive — both kept, neither modified.

## Verification on the rebased tree

- `npm run verify` → exit 0. `264 files / 3844 tests passed | 10 skipped`
- `✓ fixture PII: 58 PDFs and 15 ground-truth sidecars — all personas synthetic`
- `✓ issue-linked baselines: 11 citation(s) across 2 gate baseline(s) and 15 truth file(s)`
- `✓ built in 8.46s`
- fallow: **0 new** findings (8 inherited, gate-excluded). The one dead-code hit is `gen-education-lone-year-flush-right.mjs`, unreachable like every other `scripts/fixtures/gen-*.mjs` — the established pattern, not a regression
- CI on `42cace7`: `verify` / `fallow` / Cloudflare all SUCCESS

All six review threads replied to and resolved; zero unresolved remain.

## Disclosure

This approval is not independent: I wrote the six fixes and I am approving them. They are docs-and-tests only, with `verify` and CI green on the rebased tree, so I am accepting that rather than blocking on a second reviewer for citation edits. Flagging it because the record should say so rather than imply a fresh pair of eyes.

Vaishnavi — your local branch was rewritten by the force-push. Run `git fetch && git reset --hard origin/gh-618-education-lone-year-date-slot`. Anything you had unpushed is still in your reflog.

PR #440, opened 2026-07-11
docs(architecture): canonical résumé model design + staged migration plan (#439)

s-annam:
## Summary

Design deliverable for #439 — the follow-up to the #438 canonicalization decision. **Docs only, no production code** (the issue's stated non-goal): a target-representation diagram + a staged migration plan. New file `docs/canonical-resume-model.md` (196 lines), house-style-matched to `docs/segmentation-spike.md`.

Delivers every #439 acceptance criterion:

- **Target-representation diagram** (Mermaid) — one `CanonicalResume` source of truth with thin projections (display / score / render+export / JSON-Resume / llm-diff). All five current shapes — `HeuristicParsedResume`, `SectionedResume`, `LlmParsedResume`, `AtsResumeModel`, `JsonResume` — mapped onto it with `file:line` citations.
- **Round-trip invariant relocated** — from the renderer (where `AtsEntry`'s `headerLineDate`/`subLineDate`/`atomicSegments`/`headerBold` encode #425/#298/#301/#302) onto the canonical model, so the render+export projection is round-trip-stable by construction.
- **`apply-overrides` collapse** — `parsed`/`rawText`/`sections`/`fieldConfidence` (the #133/#421 lockstep quadruple) become projections, not lockstep return values.
- **Staged plan A–E**, each with an explicit round-trip-green story; additive until the Stage E cutover so **#435 and successors are not blocked**.
- **#321 cache constraint** — version `SavedResumeSnapshot` (`resume-library.ts:37`) by `ATS_SCORE_ALGO_VERSION` + a new parser-shape version; a stale read re-parses from the stored PDF `Blob`.
- **File-count target** — 40–54 files (baseline #425/#434/#435) → **≤8** for a typical fidelity change after Stage C.
- **Per-stage follow-up issues** enumerated (§9) to mint once the plan is accepted.

### Correction to #438/#439 as written

Both issues cite `classifyLine` at `sections.ts:1854` with `lineLooksLikeDatedEntry = hasDateRange(line) || hasDateRange(nextLine)`. **That identifier and that function do not exist in the code.** `classifyLine` is at `sections.ts:1829`; the real one-line `Title  Dates` header-vs-entry mechanism is `columnGapCuts` + `isLoneDateRange` (`sections.ts:264-277`). The *problem class* is real (header-vs-entry decided from adjacent raw-line signals, not a structured field); the regression acceptance test in the doc (§7) is re-anchored to the real mechanism.

## Test plan

[...]

Issue #438, opened 2026-07-11
Architecture: decompose the parse → export path so fidelity changes stay small


s-annam:
# Architecture: decompose the parse → export path so fidelity changes stay small

## Problem

Recent export/download-fidelity PRs (#425, its follow-up #434, and #435 at 54 files) each landed wide not because the *feature* was big, but because the parse → export path is coupled enough that a small, coherent change is hard to express as a small diff. This issue names the coupling and records the direction we take. It is an architecture-direction issue, not a bug.

## Where the coupling actually lives

Two structural facts drive the diff size:

**1. Many parallel "parsed résumé" representations, each with its own mapping layer.**

We currently carry at least five shapes for the same underlying résumé:

| Shape | Home | Role |
|---|---|---|
| `HeuristicParsedResume` | `heuristics/` (cascade output) | what the parser produces |
| `SectionedResume` / `sections` | `heuristics/` | section view the scorer + editor pool from |
| `LlmParsedResume` | `webllm/parse-resume.ts` | on-device LLM parse (kept field-name-in-sync by hand — see the comment at `parse-resume.ts:53`) |
| `AtsResumeModel` (`AtsEntry`, `AtsSection`, `AtsEntryFields`) | `pdf/ats-resume-model.ts` | render + export model |
| `JsonResume` | `pdf/to-json-resume.ts` | JSON Resume export |

Every adjacent pair needs a hand-written adapter, and edits have to keep several views in lockstep at once. `apply-overrides.ts` is the clearest tell: `ApplyOverridesResult` must return `parsed` **and** `rawText` **and** `sections` **and** `fieldConfidence` together, because a single user edit that isn't mirrored across all of them re-grades or re-parses wrong (the interface's own comments cite #133, #421 Blocking #1/#3). A change to "how a role is represented" is therefore never local.

**2. The render model encodes re-parse invariants.**

`AtsResumeModel` is doing triple duty — display, PDF layout, *and* the JSON-Resume export hint (`AtsEntry.fields`). Worse, its layout fields exist to satisfy the round-trip identity (parse → export → re-parse must yield the same fields). Look at `AtsEntry` in `ats-resume-model.ts`: `headerLineDate` / `subLineDate` / `atomicSegments` / `headerBold` are each documented in terms of how the exported PDF must re-parse — the `flush()` date-range exemption in `sections.ts`, title↔company swap avoidance (#298), skills-wrap atomicity (#301), entry-boundary date cues (#302), all under #425.

[...]

PR #670, opened 2026-07-29
fix(parser,score,pdf,docs): bulleted role headers, a dateless title's inline year, §7's retracted correction, and a font fallback that refuses instead of mangling a name (#661, #662, #663, #664)

### LINKED ISSUE #661: [docs] canonical-resume-model.md line citations outside §10 are stale

### LINKED ISSUE #662: Experience section with a date-range heading and bulleted role lines parses zero entries

### LINKED ISSUE #663: Inline year in a dateless role's title is split out as start_date on round-trip

### LINKED ISSUE #664: Non-WinAnsi characters in a candidate's name are mangled to '?' in the exported PDF

s-annam:
## Summary

Three defects surfaced by re-reading the corpus gates' own baselines. Each one had a baseline entry or a doc paragraph asserting something the code contradicted — the failure mode `check:baselines` cannot catch, because it certifies an issue *number* and cannot read prose.

### #662 — a date-range heading with bulleted role lines parsed zero entries

```
Mar 2021 - Present
• Staff Engineer, Globex Systems LLC     ← the role header, bulleted
• Led the payments platform migration…
```

`parseEntryBlocks` anchors on the date, finds no header lines above it, and #145's date-only-phantom contract drops the whole block. The first body bullet is now promoted to the role header behind four narrow gates: the anchor carried a **complete** range, another body bullet follows, the candidate resolves to **both** a title-like role and an organization-shaped company, and the title reads as a standalone role. Once promoted it is metadata — it leaves the description, and the scorer suppresses the matching source bullet from its observation pool, so it is neither graded as an achievement nor re-exported as one.

**The standalone-role test is two gates, not one.** `ROLE_TITLE_EDGE_RE` is positive grammar (a title ends in its role noun, or an executive title leads with one) and needs no verb list to reject `Engineering Roadmap, improving Distributed Systems`. But it is **necessary, not sufficient** — a role noun ends a sentence as readily as a title:

| Candidate | Ends in role noun | Header casing | Splits on comma | Promoted before this PR |
|---|---|---|---|---|
| `Staff Engineer, Globex Systems LLC` | ✅ | ✅ | ✅ | ✅ correct |
| `Hired Staff Engineer, Cloud Infrastructure` | ✅ | ✅ | ✅ | ❌ **fabricated a role, and deleted the staffing bullet it was built from** |
| `Owned Developer Platform, Cloud Infrastructure` | ❌ | ✅ | ✅ | ✅ rejected — but only because "Platform" is not a role noun |

`Owned …` was the only case covered, and it passed **for the wrong reason**: swap the noun for a real role noun and the guard evaporates. The verb is the discriminator, so `looksLikeRoleHeaderTitle` now asks about the verb.

That check reuses the scorer's `ACTION_VERBS` rather than a denylist maintained at the call site, so a verb added for bullet grading cannot leave the parser gate behind. The set moves to a new zero-dep leaf, `src/lib/lexicon/action-verbs.ts` — it lived in `score.ts`, which made the parser's only route to it an import of the whole scoring module into the tier-1 parse chunk for one `Set` membership test.

**Accepted cost, stated plainly:** a genuine title whose first word is in that set (`Managed Services Engineer`) is now rejected. That is the safe direction — dropping a real role is #145's long-standing date-only behavior, whereas fabricating one both invents history *and* deletes a real bullet.

### #663 — a dateless role's inline year was split out as `start_date`

`Software Engineer Intern Summer 2022` round-tripped as title `… Summer` + `start_date: 2022`. `collectDatelessAnchors` lost the provenance distinguishing a date the exporter *composed* from one that was always part of the title, so the re-parse could not tell them apart.

### #661 — §7 of `docs/canonical-resume-model.md` asserted two symbols don't exist

§7 claimed `lineLooksLikeDatedEntry` and `hasDateRange` "do not exist in the codebase", correcting #438/#439 for citing them. **Both exist, verbatim and exactly as attributed** — `heuristics/sections.ts` → `classifyLine` computes `const lineLooksLikeDatedEntry = hasDateRange(line) || hasDateRange(nextLine)`. The correction is retracted.

It was true when written and stopped being true on the next parser merge: this doc landed in #440, and **#435** introduced both identifiers afterwards — having been flagged in the doc's own §0 as *in flight at the time*. So the whole doc is re-anchored from `path:line` to `path` → `Symbol`, and the retraction records the general lesson: a **negative** claim about the codebase does not belong in a document nothing re-verifies on merge.

The problem class #438 named survives intact — and the live code states it more plainly than the issue did. The header-vs-entry call is made from **adjacent raw-line signals** (`hasDateRange(nextLine)` reaching forward one line; `isLoneDateRange` on the trailing segment in `columnGapCuts`), not from a structured field. §7's acceptance test is therefore restated as a standing bar, not a property the D+E cutover established.

### Baselines deleted, not re-noted

Both defects are fixed, so their exemptions go rather than acquiring a new citation: #662's `corpus-edit-roundtrip` entry (collateral of the zero-entry parse) and #663's `corpus-roundtrip` entry. `date-heading-bulleted-roles.truth.json` loses its three `knownWrong` entries for the same reason.

### Independently: a failed PDF download was silent

`useDownloadPdf` has always returned an `error`, and no surface rendered it — so a pdf-lib or font-load failure left the user with a button that did nothing. `ReconstructedResume` now shows it through the shared `ErrorState`. Originally scope-adjacent; #664's refusal is what makes it reachable and reviewable.

## #664 — the export refuses instead of mangling a name

The batch originally carried a #664 slice, and it was **removed** before the first push: it decided a product question the issue had explicitly reserved, and it collided with #667. Both are now resolved — #667 merged, and #664's product decision is recorded on the issue — so #664 lands here, rewritten from scratch rather than restored.

**The decision: refuse.** When the Poppins fetch fails, the fallback's WinAnsi codec replaces anything outside it with `?`, so `ANNA WIŚNIEWSKA` round-tripped as `ANNA WI?NIEWSKA`. `useDownloadPdf` now asks `findExportGlyphLosses()` whether the fallback would actually cost a character; if so it names the affected fields and stops before rendering. Transliterating (`WISNIEWSKA`) was rejected — it produces a wrong name with no signal — and so was `?`, which ships a document the user would call broken. **The trigger is the network, not the user's data.** Accepted cost: someone genuinely offline with a non-WinAnsi character cannot export until they reconnect.

**Why it lands in this PR and not its own.** #670 already added the `downloadError` → `ErrorState` surface, and nothing in #670 could trigger it — the error is near-unreachable on `main`, so the affordance shipped unexercised and unreviewable. This refusal is its first real consumer. The earlier "keep them apart" argument rested on the two changes touching disjoint files, which shows they *can* be split, not that they should be — and it structurally hides the case where one PR adds a seam the other consumes.

**Why the renderer is untouched.** An earlier design (written into #664, since corrected there) gave `renderAtsResumePdf` a policy parameter so the browser could refuse while the corpus harness degraded. It has **~35 call sites, all typed `Promise<Uint8Array>`** — every round-trip, export-layout and repro test, plus the corpus oracle, which must render the degraded output in order to measure it. A separate probe leaves all of them untouched: the oracle keeps working by construction rather than by opt-out, and **no test call site changes.** The probe also gates on actual loss rather than on the font failing, so a pure-ASCII résumé — most of them — downloads exactly as before even when the fetch fails.

### Two load-path defects, without which the refusal is not actionable

- **`loadPoppinsBytes()` memoized its own rejection and never cleared it.** The guard is `!poppinsBytesPromise`, so one transient failure degraded every later download for the life of the page, even after the network recovered — silently making "try again" impossible to satisfy, since the retry replayed the stored rejection without issuing a request.
- **The fetch never checked response status.** `fetch` rejects only on a network failure, so a 404 *resolved* and the error page's bytes went to `embedFont`; the fallback was reached only because fontkit choked on them, reporting a font-parse failure instead of a 404.

`stubFetchSucceeds` gained `ok: true`. It had returned only `{ arrayBuffer }`, which reads as falsy — without that, every case in `render-ats-pdf.fonts.test.ts` would have taken the Helvetica path while claiming to test the embedded one.

### Baseline

#664's `extended-latin-name-roundtrip` → `contact` entry moves from `open` to **`accepted`**, with the decision written into the note: the corpus deliberately keeps exercising the *un-refused* path so the round-trip oracle survives. Measured: **0 of 57** corpus `expected.json` snapshots contain a character outside WinAnsi — this was one fixture, never a defect class. `check:baselines` warns that #664 is still open; the `Closes #664` below resolves that on merge.

## Review focus

- `src/lib/lexicon/action-verbs.ts` — is a new top-level `lexicon/` leaf the right home, or should this live under `heuristics/`? It is shared by the parser, the scorer, and the eval rubric, so it belongs to none of them; the leaf exists so the parser doesn't import `score.ts`.
- `looksLikeRoleHeaderTitle` — the `Managed Services Engineer` false negative is the accepted cost. Is the safe direction the right call, or is a first-token carve-out (verb followed by a role noun **and** a comma-split company) worth the extra surface?
- `docs/canonical-resume-model.md` §7 — the retraction is prose about prose. Does it earn its length, or should it be one sentence plus the corrected citations?
- `findExportGlyphLosses` covers the **fallback** path only. On the embedded path `toEmbeddedFontSafe` can still emit `?` for a code point Poppins lacks (`★`, `✓`) — decorative symbols rather than identity fields, and deliberately out of #664's scope. Is that the right line?
- The probe re-reads the model's strings rather than observing what `drawText` actually sanitized, so it is mildly conservative: headings are uppercased before sanitizing and the probe reads un-cased text. It errs toward asking the user. Acceptable?

## Test plan

- [x] `npm run verify` green end to end — typecheck, eslint, `check:fixtures`, `check:baselines`, coverage, build
- [x] `npm run test` — **3807 passed, 10 skipped, 0 failed** (+20 from the base: 13 new here, 7 arriving with #667)
- [x] `✓ fixture PII: 57 PDFs and 15 ground-truth sidecars — all personas synthetic`
- [x] `✓ issue-linked baselines: 6 citations across 2 gate baselines and 15 truth files`
- [x] **Revert-proofed the verb gate on its own.** Reverting only `!startsWithActionVerb(text)` fails the three new cases (`Hired` / `Managed` / `Mentored`) while `Owned Developer Platform` **still passes** — confirming the old case was passing for the wrong reason — and both genuine-promotion cases stay green, confirming the gate does not over-reject.
- [x] Verified every symbol the rewritten doc cites actually resolves (`ATS_SCORE_ALGO_VERSION` = `1.7`, `CANONICAL_SHAPE_VERSION` = `2`, `isDatedEntry`, `projectAtsExport`, `projectDisplay`, `projectLlmDiff`, `projectScoreSections`)
- [x] **Both #664 load-path fixes revert-proofed.** Removing the cache reset fails `re-fetches after a failure…` and nothing else. The status check needed a better assertion first: asserting "we fell back" passed with the check removed, because fontkit chokes on the error page anyway — so it now asserts the response **body is never read**, which is the only observable difference.
- [x] **Verified in a real browser** (the refusal is unreachable from Vitest). With the Poppins requests aborted, `Download resume` refuses and names `Name`; with them allowed through and **no reload** — same module instance, poisoned memo — the same page exports `anna-wisniewska-resume-ats.pdf`, and `pdftotext` shows `ANNA WIŚNIEWSKA` intact.
- [x] fallow reports 10 complexity findings / 7 clone groups, **all inherited** (report-only inside `verify`). Both clone groups sit above line 282 of `render-ats-pdf.fonts.test.ts`; this diff's additions to that file start at 282, so they are re-attributed, not introduced.

Closes #661
Closes #662
Closes #663
Closes #664

## Provenance
Code implementation: Codex (GPT-5.x, 2 review rounds; stopped before PR on its own blocking findings)
Blocker fixes, #664 re-scope + implementation, commit + PR: Claude Opus 5 (1M context) (high)
Verification: `npm run verify` — green locally; also runs in CI

### DIFF
--- a/src/hooks/useDownloadPdf.ts
+++ b/src/hooks/useDownloadPdf.ts
@@ -5,17 +5,38 @@
  * useDownloadPdf — drives the "Download PDF" action on the reconstructed-resume
  * surface (#171).
  *
- * Flow: build the flat ATS model from the surface's own props → render bytes
- * with the pdf-lib draw engine → wrap in a Blob → trigger a same-document
- * download via a temporary object URL. Everything is client-side; no network
- * request is made (no font fetch, no upload), satisfying the zero-egress AC.
+ * Flow: build the flat ATS model from the surface's own props → check that the
+ * export font can draw every character (#664) → render bytes with the pdf-lib
+ * draw engine → wrap in a Blob → trigger a same-document download via a
+ * temporary object URL.
+ *
+ * Zero-egress holds, but not because nothing is fetched: the renderer DOES issue
+ * a `fetch` for the vendored Poppins TTFs, and this docblock previously claimed
+ * "no network request is made (no font fetch, no upload)", which was false. The
+ * fetch targets the app's own bundled-asset origin — never a font CDN — so no
+ * résumé bytes leave the browser, which is the actual guarantee. Say custody,
+ * not runtime.
+ *
+ * This hook owns the refusal for #664. When the Poppins fetch fails, the
+ * renderer falls back to Helvetica, whose WinAnsi codec replaces anything
+ * outside it with `?` — including a candidate's own name. Rather than hand back
+ * a PDF reading `ANNA WI?NIEWSKA`, the hook probes first and refuses, because
+ * the trigger is the network rather than the user's data and a retry usually
+ * clears it. The refusal lives here and not in `renderAtsResumePdf` so that the
+ * renderer's ~35 other call sites — every round-trip and export test, plus the
+ * corpus oracle, which must keep rendering the degraded value to measure it —
+ * are unaffected.
  */
 
 import { useCallback, useState } from "react";
 import type { CascadeResult } from "../lib/heuristics/types.ts";
 import type { AnonymousAtsScore } from "../lib/score/score.ts";
 import { buildAtsResumeModel } from "../lib/pdf/ats-resume-model.ts";
-import { renderAtsResumePdf } from "../lib/pdf/render-ats-pdf.ts";
+import {
+  findExportGlyphLosses,
+  renderAtsResumePdf,
+  type ExportGlyphLoss,
+} from "../lib/pdf/render-ats-pdf.ts";
 import { slugifyName, triggerBlobDownload } from "../lib/download/blob-download.ts";
 import { trackDownloadCompleted, type DownloadSource } from "../lib/analytics.ts";
 import { clearBlankDraft } from "./useResumeAnalysis.ts";
@@ -32,6 +53,29 @@ function filenameFromName(name: string | undefined): string {
   return slug ? `${slug}-resume-ats.pdf` : "resume-ats.pdf";
 }
 
+/**
+ * The message shown when the export font is unavailable and the fallback would
+ * mangle real characters (#664).
+ *
+ * Exported for its own test: the field list is the part a user acts on, and it
+ * has to name the fields WITHOUT echoing their values — a résumé's own text does
+ * not belong in an error banner. `where` labels are already user-facing (a
+ * contact field name or the section's own heading), and duplicates collapse so a
+ * résumé with forty arrow bullets does not produce a forty-item sentence.
+ */
+export function glyphLossMessage(losses: readonly ExportGlyphLoss[]): string {
+  const fields = [...new Set(losses.map((l) => l.where))];
+  const list =
+    fields.length === 1
+      ? fields[0]
+      : `${fields.slice(0, -1).join(", ")} and ${fields[fields.length - 1]}`;
+  return (
+    `Could not load the résumé font, so this export would fall back to a font ` +
+    `that cannot draw every character in your ${list}. Downloading now would ` +
+    `replace those characters with "?". Check your connection and try again.`
+  );
+}
+
 export function useDownloadPdf(
   result: CascadeResult,
   score: AnonymousAtsScore,
@@ -44,6 +88,18 @@ export function useDownloadPdf(
     setError(null);
     try {
       const model = buildAtsResumeModel(result, score);
+
+      // #664: refuse rather than silently substituting "?" in the user's own
+      // fields. Returns [] whenever the embedded font loaded OR nothing would
+      // actually be lost, so a pure-ASCII résumé downloads exactly as before
+      // even when the font fetch fails. Returning early skips the download and
+      // the analytics event; `finally` still clears `isGenerating`.
+      const losses = await findExportGlyphLosses(model);
+      if (losses.length > 0) {
+        setError(glyphLossMessage(losses));
+        return;
+      }
+
       const bytes = await renderAtsResumePdf(model);
       // `bytes.slice()` copies into a fresh ArrayBuffer-backed view so Blob gets
       // a clean buffer.
--- a/src/lib/heuristics/extract/experience.ts
+++ b/src/lib/heuristics/extract/experience.ts
@@ -5,8 +5,13 @@ import type { ResumeExperience } from "../../score/types.ts";
 import type { PdfSection } from "../sections.ts";
 import { parseEntryBlocks } from "../entry-blocks.ts";
 import type { EntryBlock } from "../entry-blocks.ts";
-import { finalizeEntries, looksLikeCompany } from "./shared.ts";
+import {
+  finalizeEntries,
+  looksLikeCompany,
+  looksLikeTitle,
+} from "./shared.ts";
 import { disambiguateCompanyTitle } from "./experience-disambiguate.ts";
+import { startsWithActionVerb } from "../../lexicon/action-verbs.ts";
 
 // ── Experience ──────────────────────────────────────────────────────────────
 
@@ -40,6 +45,7 @@ export function extractExperience(
     anchor: "date_range",
     collectBody: true,
     headerLookback: 2,
+    dateParsing: "date_anchors_only",
   });
   // A dateless experience section yields zero `date_range` blocks. Fall back to
   // the `"first_line"` anchor so each header-run + bullet-group is recovered as
@@ -52,6 +58,7 @@ export function extractExperience(
     blocks = parseEntryBlocks(experience, {
       anchor: "first_line",
       collectBody: true,
+      dateParsing: "date_anchors_only",
     });
   }
   // Map each block, then carry a shared-employer banner down to the roles that
@@ -166,6 +173,144 @@ function propagateSharedEmployer(
   }
 }
 
+const HEADER_CONNECTOR_RE = /^(?:and|at|for|in|of|on|the)$/i;
+
+/**
+ * A promoted title must read as a standalone role designation, not merely
+ * contain a title keyword somewhere in accomplishment prose. Most titles end
+ * in the role noun ("Staff Engineer", "Product Manager"); executive titles
+ * may lead with it ("Director of Product", "Head of Engineering"). Keeping
+ * that grammar positive is what carries most of the work: it needs no verb list
+ * to reject "Engineering Roadmap, improving Distributed Systems".
+ *
+ * It is necessary but NOT sufficient, because the role noun it anchors on ends a
+ * sentence as readily as a title ("Hired Staff Engineer"). The verb-lead check
+ * in `looksLikeRoleHeaderTitle` covers that residue; keep both.
+ */
+const ROLE_TITLE_EDGE_RE =
+  /(?:^(?:ceo|cfo|chief|cio|co-?founder|coo|cto|director|founder|head|lead|manager|president|vice\s+president|vp)\b|\b(?:accountant|administrator|advisor|adviser|agent|ambassador|analyst|apprentice|architect|assistant|associate|auditor|ceo|cfo|cio|clerk|consultant|coordinator|coo|counselor|cto|designer|developer|devops|director|editor|engineer|fellow|founder|instructor|intern|internship|lead|lecturer|manager|officer|pm|president|principal|producer|professor|recruiter|representative|researcher|scientist|specialist|sre|strategist|supervisor|teacher|technician|tpm|trainee|tutor|volunteer|writer)(?:\s+(?:i{1,4}|l\d+|\d+))?$)/iu;
+
+/** True when every substantive token has header casing. Connectors may remain
+ * lowercase; brands such as iOS/eBay qualify through an internal capital. */
+function hasHeaderCase(text: string): boolean {
+  return text.split(/\s+/).every((raw) => {
+    const token = raw.replace(
+      /^[^\p{L}\p{N}]+|[^\p{L}\p{N}]+$/gu,
+      "",
+    );
+    if (!token || HEADER_CONNECTOR_RE.test(token)) return true;
+    return (
+      /^[\p{Lu}\p{Lt}\p{N}]/u.test(token) ||
+      /^\p{Ll}+\p{Lu}/u.test(token)
+    );
+  });
+}
+
+/**
+ * Reject a candidate that LEADS with an action verb, however title-shaped its
+ * tail is.
+ *
+ * `ROLE_TITLE_EDGE_RE` is deliberately positive grammar, but positive grammar
+ * alone is not sufficient here: the role noun it anchors on can sit at the end
+ * of a *sentence* as easily as at the end of a title. "Hired Staff Engineer,
+ * Cloud Infrastructure" satisfies every other gate — the tail matches
+ * `engineer$`, every token has header casing, and the comma splits it into a
+ * plausible `Title, Company` pair — so it promoted to a fabricated role AND was
+ * then suppressed from the description, score pool, and export as "metadata".
+ * A staffing achievement disappeared and a role nobody held took its place.
+ *
+ * "Owned Developer Platform" was the only case covered before, and it failed for
+ * the wrong reason: "Platform" is not a role noun. Swap the noun for one that is
+ * and the guard evaporates. The verb, not the noun, is the discriminator, so
+ * this asks about the verb.
+ *
+ * The list is the shared `ACTION_VERBS` lexicon — the same set the scorer grades
+ * bullet specificity against — rather than a denylist maintained here, so a verb
+ * added for scoring cannot leave this gate behind. Accepted cost: a genuine
+ * title whose first word is in that set ("Managed Services Engineer") is
+ * rejected. That is the safe direction — dropping a real role is #145's
+ * long-standing date-only behavior, while fabricating one both invents history
+ * and deletes the bullet it was made from.
+ */
+function looksLikeRoleHeaderTitle(text: string): boolean {
+  return (
+    looksLikeTitle(text) &&
+    hasHeaderCase(text) &&
+    ROLE_TITLE_EDGE_RE.test(text) &&
+    !startsWithActionVerb(text)
+  );
+}
+
+interface PromotedRoleHeader {
+  fields: ReturnType<typeof disambiguateCompanyTitle>;
+  description: string | undefined;
+}
+
+/**
+ * Recover fields from the first source bullet of an otherwise date-only block.
+ *
+ * The narrow gates preserve #145's date-only-phantom contract:
+ * - the anchor carried a complete range but no header text;
+ * - another body bullet follows, distinguishing `role bullet + achievements`
+ *   from a lone achievement;
+ * - the candidate independently resolves to BOTH a title-like role and an
+ *   organization-shaped company; and
+ * - the title has a positive standalone-role shape rather than a title keyword
+ *   embedded in sentence-led accomplishment prose.
+ *
+ * Once promoted, the first body unit is metadata rather than an achievement.
+ * Remove it from the description so display and export do not duplicate the
+ * reconstructed role header. The scorer independently suppresses the matching
+ * source bullet from its section-derived observation pool — gated on the
+ * `header_from_bullet` flag this promotion stamps, so the suppression cannot
+ * reach a normally-parsed role's bullets.
+ *
+ * `block.bulletCount` and `block.body`'s line count are the same counter, not
+ * two that happen to agree: the anchored builder in `entry-blocks.ts` derives
+ * both from one `bodyUnits` array — `body` is `bodyUnits.join("\n")` and
+ * `bulletCount` is `bodyUnits.length` — and a wrapped tail is appended onto its
+ * unit with a space, never as a new line, so no unit contains a `\n`. Empty
+ * units are skipped before the push, so the `.trim()` on the join cannot shift
+ * index 0 either. The `anchorHeaderIndex !== -1` gate above keeps this exact:
+ * `parseBulletList`'s builder leaves `anchorHeaderIndex` undefined and is
+ * rejected before we ever split its body. So the `>= 2` gate guarantees
+ * `bodyLines[0]` is a whole bullet and `slice(1)` strands nothing.
+ */
+function promoteBulletedRoleHeader(
+  block: EntryBlock,
+): PromotedRoleHeader | undefined {
+  if (
+    block.headerLines.length !== 0 ||
+    block.anchorHeaderIndex !== -1 ||
+    block.bulletCount < 2 ||
+    !block.dates.start_date ||
+    (!block.dates.end_date && !block.dates.is_current)
+  ) {
+    return undefined;
+  }
+
+  const bodyLines = block.body?.split("\n") ?? [];
+  const candidate = bodyLines[0]?.trim();
+  if (!candidate) return undefined;
+
+  const fields = disambiguateCompanyTitle([candidate]);
+  if (
+    !fields.title ||
+    !fields.company ||
+    !looksLikeRoleHeaderTitle(fields.title)
+  ) {
+    return undefined;
+  }
+  if (!hasHeaderCase(fields.company)) {
+    return undefined;
+  }
+  const remainingBody = bodyLines.slice(1).join("\n").trim();
+  return {
+    fields,
+    description: remainingBody || undefined,
+  };
+}
+
 /** Map one dated entry block to a `ResumeExperience` and its confidence score.
  *  Extracted from `extractExperience` to keep each function below the
  *  complexity threshold; mirrors `projectFromBlock` / `achievementFromBlock`. */
@@ -174,11 +319,16 @@ function experienceFromBlock(block: EntryBlock): {
   score: number;
 } {
   const { dates } = block;
-  const { title, company, team, location } = disambiguateCompanyTitle(
+  const parsedFields = disambiguateCompanyTitle(
     block.headerLines,
     block.anchorHeaderIndex,
   );
-  const description = block.body;
+  const promoted =
+    parsedFields.title || parsedFields.company
+      ? undefined
+      : promoteBulletedRoleHeader(block);
+  const { title, company, team, location } = promoted?.fields ?? parsedFields;
+  const description = promoted ? promoted.description : block.body;
 
   // Score the entry.
   let score = 0;
@@ -197,6 +347,7 @@ function experienceFromBlock(block: EntryBlock): {
       ...(dates.start_date ? { start_date: dates.start_date } : {}),
       ...(dates.end_date ? { end_date: dates.end_date } : {}),
       ...(dates.is_current ? { is_current: true } : {}),
+      ...(promoted ? { header_from_bullet: true as const } : {}),
       description: description || undefined,
     },
     score: Math.min(score, 1),
--- a/src/lib/score/score.ts
+++ b/src/lib/score/score.ts
@@ -17,6 +17,7 @@ import type {
 } from "./types.ts";
 import type { SectionedResume } from "../heuristics/sections.ts";
 import { assignBulletIds } from "./bullet-id.ts";
+import { startsWithActionVerb } from "../lexicon/action-verbs.ts";
 import {
   deriveContactProfiles,
   isProfileConfident,
@@ -100,12 +101,20 @@ const WEIGHTS = {
  *   its own leftover fragment. Both now share CARDINAL_SEPARATOR. Strictly
  *   removes false Specificity credit; a bullet that genuinely quantifies is
  *   unaffected, so scores move only DOWN and only on this rare shape.
+ * - 1.8 (2026-07-30): a role header the parser recovered from its block's first
+ *   body bullet (#662, `header_from_bullet`) no longer enters the graded pool.
+ *   Date-heading templates write `Title, Company` in bullet position; once
+ *   `promoteBulletedRoleHeader` promotes that unit to a structured header,
+ *   grading it as an accomplishment double-counts it. This changes WHICH
+ *   bullets are graded — observable as a lower `bulletCount` — so Specificity /
+ *   Structure move on résumés of that shape. Same class as 1.5. Scoped to
+ *   promoted entries; a normally-parsed role's bullets are untouched.
  */
 // Surfaced to the UI via the `algoVersion` score field, and consumed by the
 // #321 resume-library cache to version persisted parse+score records (a bump
 // here invalidates stale cached snapshots, which then re-parse from the stored
 // PDF blob — see `resume-library.ts`).
-export const ATS_SCORE_ALGO_VERSION = "1.7";
+export const ATS_SCORE_ALGO_VERSION = "1.8";
 
 // ── Shared scoring rules ────────────────────────────────────────────────────
 //
@@ -278,44 +287,6 @@ function bulletHasMetric(text: string): boolean {
   return QUANTIFYING_CARDINAL.test(
     stripped.replace(WORD_YEAR_TOKEN, " ").replace(NON_QUANTIFYING_PHRASE, " "),
   );
-}
-
-/**
- * Curated past-tense action verbs used to grade the user's *existing*
- * bullets. Exported so the rewrite eval (`src/lib/webllm/eval/verbs.ts`)
- * can reuse this as the base set without duplicating it — the eval set
- * adds present-tense and IC-discipline verbs on top, but the scorer's
- * specificity-dimension semantics stay anchored here.
- *
- * Kept narrow on purpose: weak generic verbs ("worked", "helped",
- * "supported", "responsible", "assisted", "participated") are deliberately
- * NOT here. A bullet leading with one of those SHOULD fail the
- * specificity check.
- */
-export const ACTION_VERBS: ReadonlySet<string> = new Set([
-  "led", "managed", "developed", "built", "designed", "implemented",
-  "created", "launched", "drove", "increased", "reduced", "improved",
-  "delivered", "established", "optimized", "architected", "scaled",
-  "automated", "streamlined", "coordinated", "negotiated", "achieved",
-  "spearheaded", "mentored", "transformed", "pioneered", "orchestrated",
-  "accelerated", "consolidated", "eliminated", "enhanced", "executed",
-  "facilitated", "generated", "integrated", "migrated", "overhauled",
-  "redesigned", "refactored", "resolved", "revamped", "simplified",
-  "supervised", "trained", "unified", "upgraded",
-  // Promoted from the eval-only extension (#622) — past-tense, general
-  // register, and squarely in the eng/PM lane this base set already covers.
-  "shipped", "owned", "secured", "deployed", "engineered", "rewrote",
-  "authored", "analyzed", "conducted", "identified", "presented",
-  "produced", "published", "planned",
-  // Newly added (#622) — strong outcome verbs missing from both this set
-  // and the eval extension.
-  "won", "ran", "grew", "founded", "hired", "partnered", "defined", "cut",
-  "ported", "rebuilt", "advised", "shaped", "standardized", "instrumented",
-]);
-
-function startsWithActionVerb(bullet: string): boolean {
-  const firstWord = bullet.split(/\s/)[0]?.toLowerCase().replace(/[^a-z]/g, "");
-  return ACTION_VERBS.has(firstWord);
 }
 
 /**
@@ -728,6 +699,8 @@ export interface AnonymousAtsScoreInput {
     experience?: {
       title?: string;
       company?: string;
+      location?: string;
+      team?: string;
       start_date?: string;
       end_date?: string;
       is_current?: boolean;
@@ -919,6 +892,80 @@ function poolExperienceDescriptions(
   return out;
 }
 
+interface ExperienceHeaderSource {
+  title?: string;
+  company?: string;
+  location?: string;
+  team?: string;
+  header_from_bullet?: boolean;
+}
+
+/** Normalize only for exact role-header ownership checks. Punctuation is a
+ * separator here because source headers may use comma, middot, dash, pipe, or
+ * `@` while resolving to the same structured fields. */
+function normalizeExperienceHeaderUnit(text: string): string {
+  return text
+    .replace(/[^\p{L}\p{N}]+/gu, " ")
+    .trim()
+    .toLowerCase();
+}
+
+/**
+ * Remove marker bullets that are already represented by a structured
+ * experience header. Date-heading templates put `Title, Company` in the first
+ * bullet position; once the parser promotes that unit, grading it as an
+ * accomplishment would duplicate it in editor observations and PDF export.
+ *
+ * SCOPED TO PROMOTED ENTRIES ONLY — `header_from_bullet` (#662). Do not widen
+ * this to every experience entry: the keys are matched against the WHOLE pooled
+ * bullet set, so an unscoped filter reaches across sections. A résumé with an
+ * ordinary parsed header and an authored Achievements line that restates it
+ * ("Staff Engineer, Globex Systems LLC") would lose that line with no promotion
+ * anywhere in the parse — and losing it is not cosmetic. Dropping a bullet here
+ * drops it from `score.bullets`, which costs it its editable row on the
+ * reconstructed surface AND its slot in the exported PDF, because
+ * `resolveBullets` in `ats-resume-model.ts` falls back to the raw `description`
+ * split only when an entry has NO graded bullets — a restatement sitting beside
+ * real achievements is simply gone. For a promoted entry the line is genuinely
+ * parser-owned metadata, so suppression is correct there and only there.
+ *
+ * Both title/company orders are accepted because the disambiguator supports
+ * company-first source headers. Optional location/team suffixes are exact: an
+ * achievement with any additional prose cannot collide with these keys.
+ */
+function suppressExperienceHeaderBullets(
+  bullets: string[],
+  experience: readonly ExperienceHeaderSource[] | undefined,
+): string[] {
+  const headerKeys = new Set<string>();
+  const addKey = (parts: Array<string | undefined>) => {
+    const key = normalizeExperienceHeaderUnit(
+      parts.filter((part): part is string => Boolean(part?.trim())).join(" "),
+    );
+    if (key) headerKeys.add(key);
+  };
+
+  for (const entry of experience ?? []) {
+    if (!entry.header_from_bullet) continue;
+    if (!entry.title?.trim() || !entry.company?.trim()) continue;
+    const suffixes: Array<Array<string | undefined>> = [
+      [],
+      [entry.location],
+      [entry.team],
+      [entry.location, entry.team],
+    ];
+    for (const suffix of suffixes) {
+      addKey([entry.title, entry.company, ...suffix]);
+      addKey([entry.company, entry.title, ...suffix]);
+    }
+  }
+
+  if (headerKeys.size === 0) return bullets;
+  return bullets.filter(
+    (bullet) => !headerKeys.has(normalizeExperienceHeaderUnit(bullet)),
+  );
+}
+
 export function computeAnonymousAtsScore(
   input: AnonymousAtsScoreInput,
 ): AnonymousAtsScore {
@@ -936,7 +983,10 @@ export function computeAnonymousAtsScore(
   // pool, so a résumé whose OTHER accomplishment sections (e.g. Achievements)
   // still carry glyph bullets doesn't mask a glyph-less Experience section —
   // see `poolExperienceDescriptions` for the full rationale.
-  const bullets = extractBulletsFromSections(input.sections);
+  const bullets = suppressExperienceHeaderBullets(
+    extractBulletsFromSections(input.sections),
+    input.parsed.experience,
+  );
   if (extractExperienceSectionBullets(input.sections).length === 0) {
     bullets.push(...poolExperienceDescriptions(input.parsed.experience));
   }
--- a/docs/canonical-resume-model.md
+++ b/docs/canonical-resume-model.md
@@ -2,120 +2,121 @@
 
 Design for [#439](https://github.com/offlinecv/OfflineCV/issues/439), the follow-up to the architecture decision in [#438](https://github.com/offlinecv/OfflineCV/issues/438) ("one canonical representation, staged").
 
-**Status:** design only — no production code changed by this issue (acceptance criterion). Implementation lands as the sequenced, per-stage follow-up issues at the end of this doc. Each stage is separately reviewable and keeps the round-trip invariant green; **nothing here is a big-bang PR.**
+**Status:** the design issue itself was docs-only; its implementation stages have since shipped through the combined Stage D+E cutover (#445). Sections 0–3 and 6 describe the current implementation. Section 4 preserves the staged migration as implementation history.
 
-**Baseline the design is written against:** `main` as of the mapping in this doc. All `file:line` citations are against that baseline. `ATS_SCORE_ALGO_VERSION` is `"1.4"` (`src/lib/score/score.ts:73`); [#435](https://github.com/offlinecv/OfflineCV/issues/435) bumps it to `1.5` in flight — the cache-version hook in §6 keys off whatever value is current at each stage's cutover, not a pinned literal.
-
----
-
-## 0. The thesis, restated precisely
-
-We carry **five** parallel shapes for the same underlying résumé, and every adjacent pair needs a hand-written adapter kept in lockstep by hand:
-
-| # | Shape | Home | Role |
-|---|---|---|---|
-| 1 | `HeuristicParsedResume` / `CascadeResult` | `heuristics/types.ts:121` / `:174` | what the parser produces |
-| 2 | `SectionedResume` | `heuristics/sections.ts:85` | section pools the scorer + editor grade from |
-| 3 | `LlmParsedResume` | `webllm/parse-resume.ts:56` | on-device LLM parse, **field names hand-synced** to #1 |
-| 4 | `AtsResumeModel` / `AtsEntry` / `AtsSection` / `AtsEntryFields` | `pdf/ats-resume-model.ts:181` / `:114` / `:173` / `:87` | render + PDF-export model |
-| 5 | `JsonResume` | `pdf/to-json-resume.ts:113` | JSON Resume export |
-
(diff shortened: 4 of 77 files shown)


src/lib/score/score.ts
```diff
@@ -936,7 +960,10 @@ export function computeAnonymousAtsScore(
   // pool, so a résumé whose OTHER accomplishment sections (e.g. Achievements)
   // still carry glyph bullets doesn't mask a glyph-less Experience section —
   // see `poolExperienceDescriptions` for the full rationale.
-  const bullets = extractBulletsFromSections(input.sections);
+  const bullets = suppressExperienceHeaderBullets(
```
Samhit21, line 986: [Secondary]: this filter is unconditional — it is not scoped to headers the parser promoted, which is what the docblock above implies ("once the parser promotes that unit").

Keys are built from *every* experience entry and applied to the *whole* pooled bullet set, on every resume. Reproduced on one with no promotion at all — an ordinary parsed header plus an authored line under **Achievements** that restates it:

```
parsed.experience: [{ title: "Staff Engineer", company: "Globex Systems LLC" }]
achievements:
  • Staff Engineer, Globex Systems LLC      <-- authored, not promoted
  • Cut median deploy time from 42 minutes to 9 minutes
  • Led the payments platform migration across 14 regional services

pooled → ["Cut median deploy…", "Led the payments…"]      totalBullets: 2
```

Note the reach: a key derived from an **experience** header filtered a bullet pooled from **achievements**. The line also leaves `score.bullets`, so it has no editable row on the reconstructed surface. I did not chase whether it additionally drops from the export — that depends on the `description` fallback in `buildAtsResumeModel` — so treat that as an open question rather than a claim.

For an exact duplicate this is probably the behaviour you want. The problem is scope drift: either gate it on entries actually promoted (a flag on the block would carry it), or keep it global and say so here — as written, a future reader could "simplify" it to the promoted case believing that a no-op.

Second, separable point: this changes *which bullets are graded*, and `ATS_SCORE_ALGO_VERSION` is still `1.7`. The file's own changelog bumped **1.5** for exactly this class ("gated per accomplishment section instead of on the whole bullet pool"), and the corpus snapshot shows it is observable (`bulletCount: 6 → 4`). `CACHE_SHAPE_VERSION` composes that stamp, so a library record saved pre-merge reloads with its old score and no re-parse. Suggest `1.8` + a changelog line naming #662.

src/lib/heuristics/extract/experience.ts
```diff
@@ -166,6 +173,131 @@ function propagateSharedEmployer(
...
+ *   organization-shaped company; and
+ * - the title has a positive standalone-role shape rather than a title keyword
+ *   embedded in sentence-led accomplishment prose.
+ *
+ * Once promoted, the first body unit is metadata rather than an achievement.
+ * Remove it from the description so display and export do not duplicate the
+ * reconstructed role header. The scorer independently suppresses the matching
+ * source bullet from its section-derived observation pool.
+ */
+function promoteBulletedRoleHeader(
```
Samhit21, line 279: [Nit]: the gate is `block.bulletCount >= 2` but the candidate is `block.body.split("\n")[0]` — two different counters, assumed to agree.

If a wrapped bullet ever contributes more body lines than it does bullets, `bodyLines[0]` is a fragment that gets promoted to a title and `slice(1)` leaves the remainder stranded at the top of the description. The four gates make this very hard to reach and the fixture covers the happy path, so this is a note rather than a request — but the relationship between `bulletCount` and `body`'s line count is load-bearing here and nothing states it.

If they are guaranteed equal by `parseEntryBlocks`, a sentence saying so would close it.

Samhit21 (dismissed):
> **Verdict: APPROVE** (0 Blocking). Two Secondaries worth a follow-up — or a pre-merge tweak if you'd rather; neither is a merge blocker.

Four issues in one PR is usually where I start pulling threads, and these genuinely are one thread: each defect was a baseline or a doc paragraph asserting something the code contradicted. The #661 retraction in particular is the right instinct — a *negative* claim about the codebase in a document nothing re-verifies on merge is a landmine, and saying so in the doc is worth more than the correction itself.

I verified the load-bearing claims rather than reading them:

- **The verb-lexicon extraction is byte-identical.** 74 verbs in, 74 out, zero added, zero lost — and the added `if (!firstWord) return false` is equivalent to the old `has(undefined)`. So moving `ACTION_VERBS` to `lexicon/` moves no score. That was the thing most likely to go wrong silently here, and it didn't.
- **The #664 refusal is much narrower than "non-ASCII".** I drove `findExportGlyphLosses` on the fallback path across the characters the exporter itself composes:

  | input | refuses? |
  |---|---|
  | en dash `–` (date ranges), em dash, middot `·`, bullet `•`, curly quotes, ellipsis | **no** — all WinAnsi-representable |
  | `é` and Latin-1 accents | **no** |
  | `ś` U+015B (Latin-Extended) | **yes** → `Anna Wi?niewska` |

  Worth putting in the body: the PR says "most résumés are pure ASCII", but it's stronger than that — the whole WinAnsi repertoire passes, including the `·`/`–` separators the exporter emits per §10. The "accepted cost" is therefore much smaller than stated.
- **Gates on `bce6ce2`:** `verify` green end to end — `check:fixtures` ✓ (57 PDFs + 15 sidecars) · `check:baselines` ✓ (6 citations) · **3807 passed / 10 skipped**, matching the body exactly · build ✓.
- The `date-heading-bulleted-roles` snapshot move (0 → 2 entries, bullets 6 → 4) is the intended fix, and `openresume-react-pdf`'s `experienceChangedAcrossRoundtrip: true → false` is an unremarked *improvement* — worth one line in the body so a future reader doesn't wonder why a second fixture moved.

## Secondary

**1. `suppressExperienceHeaderBullets` is unconditional — it is not scoped to promoted headers.** `score.ts:963`

The docblock frames this as the other half of #662 ("Date-heading templates put `Title, Company` in the first bullet position; **once the parser promotes that unit**…"). The code has no such condition: it builds keys from *every* experience entry and filters the *whole* pooled bullet set, on every résumé, whether or not anything was promoted.

Reproduced with a résumé that has nothing to do with #662 — an ordinary parsed header, plus a line under **Achievements** that happens to restate it:

```
parsed.experience: [{ title: "Staff Engineer", company: "Globex Systems LLC" }]
achievements bullets:
  • Staff Engineer, Globex Systems LLC          <-- authored bullet, not promoted
  • Cut median deploy time from 42 minutes to 9 minutes
  • Led the payments platform migration across 14 regional services

pooled bullets  → ["Cut median deploy…", "Led the payments…"]
totalBullets    → 2
```

The first line is gone from the pool **and** from `score.bullets`, so it also has no editable row on the reconstructed surface. I did **not** chase whether it additionally drops from the exported PDF — that depends on whether the owning entry's `description` fallback in `buildAtsResumeModel` still carries it — and I'd rather flag the question than assert an outcome I didn't measure.

For the exact-duplicate case this is arguably the behaviour you want. The issue is that the scope is wider than the docblock claims and wider than #662 needs, and the reach is cross-section: a key derived from an **experience** header filters bullets pooled from **achievements/projects** too. Either scope it to entries the parser actually promoted (a flag on the block would carry that), or keep it global and say so in the docblock — the current comment would let a future reader "simplify" it to the promoted case and think that was a no-op.

**2. The scoring pool changed, but `ATS_SCORE_ALGO_VERSION` stayed at `1.7`.** `score.ts:963`

Finding 1 is a change to *which bullets are graded*, which is exactly what the version stamp exists to track — and `score.ts`'s own changelog set the precedent:

> `1.5 (2026-07-10)`: the glyph-less-experience bullet fallback is now gated per accomplishment section instead of on the whole bullet pool (#365)

Same class: pool composition, not formula. The corpus snapshot confirms it is observable (`bulletCount: 6 → 4`, Structure `20 → 30`). Since `CACHE_SHAPE_VERSION` is `${ATS_SCORE_ALGO_VERSION}:${CANONICAL_SHAPE_VERSION}`, leaving it at `1.7` means a résumé saved to the library before this merge reloads with its old score and no re-parse, against the current algorithm. Suggest `1.8` with a changelog entry naming #662.

## Nits (non-blocking)

- **The fallow claim in the body is not quite right.** It says "10 complexity findings / 7 clone groups, **all inherited**". `experienceFromBlock` (20 cyclomatic / 18 cognitive) does **not** appear in `fallow audit --gate all` on `origin/main` — I checked in a clean worktree — so it is newly flagged, from the promoted-header branch added here. Report-only per `CLAUDE.md`, so nothing to fix; only the claim.
- **`promoteBulletedRoleHeader` reads `body.split("\n")[0]` but gates on `block.bulletCount >= 2`.** Those are two different counters. If a wrapped bullet ever contributes more body lines than bullets, `bodyLines[0]` is a fragment and the `slice(1)` drops a partial line. The four gates make it very hard to reach, and the fixture proves the happy path — flagging only because the two counters are assumed to agree and nothing states it.

## Answering your review-focus questions

- **Is `lexicon/` the right home?** Yes. It is shared by parser, scorer, and eval, so it belongs to none of them, and the stated reason (keeping `score.ts` out of the tier-1 parse chunk) is a real constraint rather than tidiness. A zero-dep leaf is the correct shape.
- **Is the `Managed Services Engineer` false negative the right call?** Yes, and the asymmetry argument is the right one: dropping a real role reproduces long-standing #145 behaviour, while fabricating one invents history *and* deletes the bullet it was built from. I would not add the first-token carve-out — it re-introduces exactly the "passes for the wrong reason" fragility the PR body correctly diagnoses in the `Owned Developer Platform` case.
- **Does the §7 retraction earn its length?** Yes. The correction is one line; the length is the *lesson* (a negative claim nothing re-verifies), and that generalises beyond this doc. Keep it.
- **Is the fallback-only scope of `findExportGlyphLosses` the right line?** Yes — and the measurement above supports it more strongly than the body does. Symbols like `★`/`✓` are decorative; identity fields are what #664 is about.
- **Is the un-cased probe acceptable?** Yes. It errs toward asking, and given how narrow the trigger turned out to be, the conservative direction costs almost nothing.

---
Verdict rule: 0 Blocking → `APPROVE`. I'm not soft-gating on the two Secondaries — but #2 is a one-line change, and if you're touching it anyway, #1's docblock is worth correcting in the same pass so the scope isn't misread later.

---

src/lib/score/score.ts
s-annam: Fixed in `59a9742` — both points, and your open question turned out to be the reason the first one is Blocking rather than a docs fix.

**The export question, answered: it does drop.** `resolveBullets` in `ats-resume-model.ts` falls back to the raw `description` split *only* when an entry has NO graded bullets. In your repro the achievements entry still has "Cut median deploy time…", so the fallback never fires and the restated line is gone from the exported PDF too — not just from the editable surface. So the unconditional filter was silent data loss on a résumé with no promotion anywhere in the parse.

Given that, I took the gate rather than the docblock. `promoteBulletedRoleHeader` now stamps `header_from_bullet: true` on the entry it promotes (`ResumeExperience`, alongside the existing display/round-trip-only `section_label`), and `suppressExperienceHeaderBullets` skips every entry without it. Your exact repro is now a test in `experience.bulleted-role-header.test.ts` — an ordinary parsed header plus an Achievements line restating it — and it is a real revert-proof: with the one-line gate removed it fails `expected [ …(2) ] to include 'Staff Engineer, Globex Systems LLC'`, and passes with it. The docblock now also states the global-reach hazard explicitly, so the "simplify it to the promoted case" reading you flagged is closed from both directions.

**Version stamp:** agreed, bumped to `1.8` with a changelog entry naming #662 and calling out that it is the same class as 1.5. All 57 corpus snapshots re-baked; the diff is one `algoVersion` line each and nothing else (`git diff -U0 -- tests/fixtures | grep -v algoVersion` is empty). To separate stamp churn from behaviour I pinned the constant back to `1.7` and re-ran the corpus — it stays green, so the new scoping is behaviourally inert across the fixtures, and `date-heading-bulleted-roles` keeps its suppressed `bulletCount: 4`, which is the proof the promoted path still fires through the flag.

src/lib/heuristics/extract/experience.ts
s-annam: Fixed in `59a9742` — they are guaranteed equal, and the docblock now says so with the reason.

Traced it: both come from one `bodyUnits` array in the anchored builder in `entry-blocks.ts` — `body` is `bodyUnits.join("\n").trim()` and `bulletCount` is `bodyUnits.length`. A wrapped tail is appended onto its unit with a space (`bodyUnits[last] += " " + text`), never as a new line, so no unit can contain a `\n`; and empty units are skipped before the push, so the `.trim()` on the join cannot shift index 0 either.

The load-bearing part is the gate you can't see from this line: `parseBulletList`'s builder is the one path where that invariant could fail — it pushes `stripBullet(...)` with no empty guard — but it never sets `anchorHeaderIndex`, so `block.anchorHeaderIndex !== -1` is `undefined !== -1` and rejects those blocks before we ever split the body. So the `>= 2` gate does guarantee `bodyLines[0]` is a whole bullet and `slice(1)` strands nothing. All of that is now in the docblock rather than implied.

Samhit21 (approved):
> **Verdict: APPROVE** (0 Blocking). Re-review of `0041374`, rebased onto `dd6527c`. Both Secondaries from my review of `bce6ce2` are resolved, and I re-verified each in **both** directions rather than reading the diff.

## Secondary #1 — suppression scope — fixed

`header_from_bullet` now stamps promoted entries, and `suppressExperienceHeaderBullets` skips everything else. Re-ran my original reproduction, plus the inverse so the fix can't be a blanket disable:

| `parsed.experience` | pooled bullets | verdict |
|---|---|---|
| ordinary parsed header | **3** — the restating line survives | the defect is gone |
| same, `header_from_bullet: true` | **2** — source bullet suppressed | #662's intent kept |

Two things worth calling out beyond "it's fixed":

- **You answered the question I declined to assert.** I flagged that I hadn't measured whether a suppressed bullet also drops from the *export*, and left it open rather than claim it. The new docblock chases it down — `resolveBullets` falls back to the raw `description` split only when an entry has **no** graded bullets, so a restatement sitting beside real achievements is simply gone. That is the right answer, and it makes the case for scoping stronger than my report did.
- **The comment now says "do not widen this", with the reason.** That is what stops the next reader from "simplifying" the flag away — which was the actual risk, more than the behaviour itself.

## Secondary #2 — algorithm version — fixed

`1.7 → 1.8`, consistent across the corpus (**58 of 58** snapshots), with a changelog entry that names #662, states the observable effect (`bulletCount` moves), and explicitly anchors it to 1.5 as the same class. That last part is the bit that keeps the convention legible for whoever bumps 1.9.

## Nit from last round — answered with a proof

I flagged that `block.bulletCount >= 2` and `body.split("\n")[0]` are two counters assumed to agree. The answer is that they are **one** counter — both derived from a single `bodyUnits` array, with wrapped tails appended onto their unit with a space rather than as a new line, and the `anchorHeaderIndex !== -1` gate excluding `parseBulletList`'s builder, which is the one that wouldn't hold. That is a better outcome than a defensive check would have been.

## Gates on `0041374`

`verify` green end to end · `check:fixtures` ✓ (58 PDFs + 15 sidecars) · `check:baselines` ✓ (6 citations) · **3871 passed / 10 skipped** · build ✓.

## One nit still standing (non-blocking, report-only)

The body's fallow line still reads "all inherited". `experienceFromBlock` is **not** flagged on `origin/main` and is flagged here at **21 cyclomatic / 19 cognitive** — up one from the 20 I measured last round, so the promoted-header branch is still growing it. `CLAUDE.md` makes fallow report-only and never blocking, so there is nothing to fix in the code; only the claim in the body is inaccurate. Mentioning it once more because it is the kind of sentence that gets copied into the next PR body.

---
Verdict rule: 0 Blocking → `APPROVE`. Nothing outstanding blocks merge.

---

src/lib/heuristics/extract/experience.ts
```diff
@@ -166,6 +173,144 @@ function propagateSharedEmployer(
...
+ * rejected. That is the safe direction — dropping a real role is #145's
+ * long-standing date-only behavior, while fabricating one both invents history
+ * and deletes the bullet it was made from.
+ */
+function looksLikeRoleHeaderTitle(text: string): boolean {
+  return (
+    looksLikeTitle(text) &&
+    hasHeaderCase(text) &&
+    ROLE_TITLE_EDGE_RE.test(text) &&
+    !startsWithActionVerb(text)
```
rohithgollapalli, line 240: **Secondary — the verb gate has a staffing-verb hole.**

`ACTION_VERBS` is curated for bullet-specificity *scoring*, so recruiting-domain verbs are absent. Verified on `bce6ce2`: with this gate intact, all four of these still promote a fabricated role **and** delete the source bullet —

- `Recruited Staff Engineer, Core Services Inc` → `title: "Recruited Staff Engineer"`, `company: "Core Services Inc"`
- `Interviewed Senior Engineer, Payments Platform LLC`
- `Onboarded Product Manager, Growth Team Inc`
- `Promoted Staff Engineer, Core Services Inc`

`Hired` is caught only because it happens to be in the scoring list; `Recruited` is the same shape and is not. That is the exact outcome this function's docblock says the gate exists to prevent.

The framing "a verb added for scoring cannot leave this gate behind" is one-directional — the gate also inherits a list never curated for it. Suggested fix, mirroring how `webllm/eval/verbs.ts` extends the same base:

```ts
/** Parser-only extension: staffing/HR verbs absent from the scoring lexicon,
 *  which is curated for bullet specificity rather than for this gate. Still a
 *  list, not a rule — a verb outside both sets can still promote. */
const HEADER_REJECT_VERBS = new Set([
  "recruited", "interviewed", "onboarded", "promoted",
  "staffed", "coached", "appointed", "assigned",
]);
```

Not blocking — the trigger needs #662's date-heading template *and* a title-cased staffing bullet in first position.

src/lib/heuristics/extract/experience.ts
```diff
@@ -166,6 +173,144 @@ function propagateSharedEmployer(
...
+
+/**
+ * Recover fields from the first source bullet of an otherwise date-only block.
+ *
+ * The narrow gates preserve #145's date-only-phantom contract:
+ * - the anchor carried a complete range but no header text;
+ * - another body bullet follows, distinguishing `role bullet + achievements`
+ *   from a lone achievement;
+ * - the candidate independently resolves to BOTH a title-like role and an
+ *   organization-shaped company; and
```
rohithgollapalli, line 257: **Secondary — this line names a gate the code does not have.**

There is no organization-shape check on the promoted company. The company side is `fields.company` non-empty (line 288) plus `hasHeaderCase` (line 291). `Cloud Infrastructure` clears both, which is precisely why `Hired Staff Engineer, Cloud Infrastructure` needed the verb gate to stop it at all.

`looksLikeCompany` is already imported in this file and is the obvious candidate if the gate is actually wanted; otherwise this bullet should say what is checked ("a header-cased company").

Worth fixing because a maintainer reading this docblock could conclude the org-shape check carries the load and relax the verb gate — the same mistake this PR documents about `Owned Developer Platform` passing for the wrong reason.

src/hooks/useDownloadPdf.ts
```diff
@@ -32,6 +53,29 @@ function filenameFromName(name: string | undefined): string {
...
+ * Exported for its own test: the field list is the part a user acts on, and it
+ * has to name the fields WITHOUT echoing their values — a résumé's own text does
+ * not belong in an error banner. `where` labels are already user-facing (a
+ * contact field name or the section's own heading), and duplicates collapse so a
+ * résumé with forty arrow bullets does not produce a forty-item sentence.
+ */
+export function glyphLossMessage(losses: readonly ExportGlyphLoss[]): string {
+  const fields = [...new Set(losses.map((l) => l.where))];
+  const list =
+    fields.length === 1
```
rohithgollapalli, line 69: **Nit** — `glyphLossMessage([])` falls through to the multi-field branch and produces `"… in your  and undefined. …"` (`fields.slice(0, -1).join(", ")` is `""`, `fields[-1]` is `undefined`).

The hook only calls it behind `losses.length > 0`, but it is exported and directly tested, so the empty case is reachable by a future caller. One line above this:

```ts
if (fields.length === 0) return "";
```

rohithgollapalli (approved):
Checked the diff against #661/#662/#663/#664's acceptance criteria first, then the offlinecv gates, then the description. Every AC is met and I independently reproduced both revert-proofs. `npm run test` 3807 passed / 10 skipped, `typecheck` clean, `eslint .` clean, `check:fixtures` green (57 PDFs + 15 sidecars), `check:baselines` green with the expected #664-still-open warning. No fixture binary added, no raw interactive element in feature code, no hardcoded palette — the `ErrorState` wiring is exactly the right reuse.

The #664 slice is the strongest part: the probe/renderer split keeps all ~35 call sites and the corpus oracle working by construction, the two load-path defects are real and each is pinned by a test that fails on its own revert, and `stubFetchSucceeds` gaining `ok: true` is a genuine catch — without it the whole fonts file would have silently tested the fallback.

Nothing blocking. Three things below worth a follow-up.

## Secondary

**1. The verb gate has a staffing-verb hole, and it fabricates exactly the role the docblock says it prevents.** (inline on `experience.ts`)

`ACTION_VERBS` is curated for *bullet-specificity scoring*, so recruiting-domain verbs are absent from it. Four candidates that satisfy every other gate promote today, on this branch:

| First bullet under a date heading | Result |
|---|---|
| `Recruited Staff Engineer, Core Services Inc` | `title: "Recruited Staff Engineer"`, `company: "Core Services Inc"` |
| `Interviewed Senior Engineer, Payments Platform LLC` | promoted |
| `Onboarded Product Manager, Growth Team Inc` | promoted |
| `Promoted Staff Engineer, Core Services Inc` | promoted |

Each one produces a role nobody held **and** removes the staffing bullet from the description, the score pool and the export — the outcome `looksLikeRoleHeaderTitle`'s own docblock calls out as the reason the gate exists ("fabricating one both invents history and deletes the bullet it was made from"). `Hired` is caught only because it happens to be in the scoring list; `Recruited` is not, and the shape is identical.

The trigger is narrow (needs #662's date-heading template *and* a title-cased staffing bullet in first position), which is why this is Secondary rather than Blocking. But the PR's framing — "a verb added for scoring cannot leave this gate behind" — is one-directional: the gate also inherits a list that was never curated for it. Cheapest fix is a parser-side extension set next to the import (`recruited`, `interviewed`, `onboarded`, `promoted`, `staffed`, `coached`, `appointed`, `assigned`), mirroring how `webllm/eval/verbs.ts` extends the same base. That is still a list rather than a rule, so it is worth saying so in the docblock.

Reproduced by parsing `Experience / "Mar 2021 - Present" / "• <prose>" / "• Cut median deploy time from 42 minutes to 9 minutes"` through `extractExperience` on `bce6ce2`.

**2. `promoteBulletedRoleHeader`'s docblock names a gate the code does not have.** (inline on `experience.ts`)

It lists "the candidate independently resolves to BOTH a title-like role and an **organization-shaped** company", and the PR summary repeats it as one of the four narrow gates. There is no organization-shape check: the company side is `fields.company` non-empty plus `hasHeaderCase`. `Cloud Infrastructure` clears both, which is why `Hired Staff Engineer, Cloud Infrastructure` needed the verb gate to stop it at all. `looksLikeCompany` is already imported in this file and is the obvious candidate if the gate is wanted; otherwise the sentence should say what is actually checked. This matters because a future maintainer reading the docblock could conclude the org-shape check carries the load and relax the verb gate — the precise mistake this PR documents about `Owned Developer Platform`.

**3. Test plan: the fallow line is wrong, and the miss is this PR's own function.**

> fallow reports 10 complexity findings / 7 clone groups, **all inherited**

`npx fallow audit --base origin/main` on `bce6ce2` reports **14** above threshold, 7 clone groups. The clone-group half is right — all 7 are pre-existing. The complexity half is not: `promoteBulletedRoleHeader` (`src/lib/heuristics/extract/experience.ts:266`, 14 cyclomatic / 56.3 CRAP) is introduced by this diff, so "all inherited" does not hold. Per CLAUDE.md a fallow complexity finding is report-only and never blocking on its own, so this changes nothing about merging — but the conclusion a reader draws from that checkbox ("no new complexity debt") is not the one the tool supports. (`10` looks like it came from the "Large functions (10 shown, 18 total)" panel.)

## Nits

- `glyphLossMessage([])` returns `"… in your  and undefined. …"`. The hook never calls it empty, but it is exported and directly tested, so the empty case is reachable by a future caller. One line: `if (fields.length === 0) return "";` or an early guard. (inline on `useDownloadPdf.ts`)
- Test plan: "this diff's additions to that file start at 282" — the first hunk in `render-ats-pdf.fonts.test.ts` adds ~29 lines at 38–86 (`ok: true`, `stubFetchHttpError`). The conclusion still holds: all four clone instances (189-197 / 241-248 and 212-217 / 273-278) sit in unchanged code, so they are re-attributed rather than introduced.
- I could not verify the browser box (`Verified in a real browser … with the Poppins requests aborted`); it is unreachable from Vitest by design, so it rests on your word. Noting it as unverified by this review, not disputed.

## What I verified rather than took on trust

- **#662 revert-proof reproduced.** Neutering only `!startsWithActionVerb(text)` fails exactly 3 of 9 (`Hired` / `Managed` / `Mentored`) while `Owned Developer Platform` still passes and both genuine promotions stay green — matching the PR's claim precisely.
- **#663 revert-proof reproduced.** Forcing `shouldParseAnchorDate` back to always-true fails all 3 new #663 cases and no others; the #358 inverse (`Research Assistant · XYZ University  2021` → `start_date: "2021"`) stays green, so the fix does not just move the asymmetry.
- **#661 doc claims re-checked against the tree.** Zero `path.ts:N` citations remain. `classifyLine` in `heuristics/sections.ts` does compute `const lineLooksLikeDatedEntry = hasDateRange(line) || hasDateRange(nextLine)` verbatim, `hasDateRange` is in the same file, `columnGapCuts` is in `line-assembly.ts` (correctly re-homed from the old `sections.ts` claim), `isLoneDateRange` in `line-primitives.ts`, `ATS_SCORE_ALGO_VERSION = "1.7"`, `CANONICAL_SHAPE_VERSION = "2"`. The retraction is accurate and the §7 rewrite earns its length — the general lesson about negative claims is the reusable part.
- **`findExportGlyphLosses` covers everything the renderer actually draws.** Walked it against `renderAtsResumePdf`'s draw sites: `contact.profiles` is the only `AtsContact` member not probed, and it is never drawn (`links` is the display path). `AtsEntryFields` is export-only. No gap.
- **Bullet-suppression is id-safe.** `suppressExperienceHeaderBullets` shrinks the pool, but ids have been text-derived since #648, so no surviving bullet's id shifts and no override re-targets.

---

Issue #597, opened 2026-07-26
job-search: pre-search "what will be searched" plan + query form IA cleanup


s-annam:
## Problem

The `/jobs/` query form tells the user what matters **after the fact, or not at all**. The single most consequential fact — which one title and which one skill actually reach a job feed — exists on screen only as a small grey line ("Searching feeds for …") and a legend-less `★` mark. Everything else the user sees is a flat wall of ~40 equally-weighted chips (titles, skills, exclusions, companies, suggestions), three scattered prose lines about what gets sent, and a privacy sentence that appears twice.

The result: the user cannot answer, before clicking Search, the only questions that matter —

1. **What string is sent, and where?**
2. **What do all the other chips do?** (answer: they only narrow and rank locally)
3. **Why do results look wrong when they do?** (usually: the wrong chip is in position 1)

### The two layers, restated

**Layer 1 — what leaves the browser** (`providers/keywords.ts`, the sole résumé-derived egress helper):

| Feed | Request | Term used |
|---|---|---|
| Remotive, Arbeitnow | `?search=<kw>` | `searchPhrase()` → **`titles[0]` only** |
| Jobicy | `?tag=<kw>` | `primaryKeyword()` → **`skills[0]`**, falling back to `titles[0]` |
| Greenhouse, Lever, Ashby | `/{slug}/jobs` | public company slug — no résumé data at all |

**Layer 2 — the local filter** (`buildQueryTermPatterns` / `matchesQuery` in `search.ts`): every other term only widens or narrows matching on-device.

So chip **#1** in each list is uniquely load-bearing, and chips 2..n are free. The UI half-says this today (`★` + promote control from #581/#591, quality glyphs from #585, the grey "Searching feeds for" line), but as annotation sprinkled onto the chip wall — not as a model the user can read at a glance and act on.

### Concrete failure

An engineering-leadership résumé whose most-recent role is a founder title parses to `titles[0] = "Founder & CEO"` and `skills[0] = "cross-functional collaboration"` (recency / résumé order — `deriveTitles` / `deriveSkills` have no notion of "searchable"). Remotive/Arbeitnow get founder postings; Jobicy's tech-shaped tags match nothing. Both are one promotion away from right, and nothing on screen makes that the obvious move. The search looks broken when it is merely mis-seeded.

## Scope

[...]

Issue #587, opened 2026-07-25
Title↔skill coherence: call out leadership titles backed by only IC technical skills


s-annam:
## Problem

A résumé can carry **leadership titles and exclusively individual-contributor technical skills**.
That is a real and consequential mismatch — the postings matching those titles ask for people
management, hiring, org design and roadmap ownership, while the document offers languages,
frameworks and architectures.

The system currently detects none of it. Titles and skills are derived independently, consumed by
different machinery, and never compared. The user sees a search that quietly under-performs and no
explanation for why.

The inverse case matters too: IC titles with purely managerial skills, on a résumé that has drifted
away from the roles it is targeting.

## Why this is nearly free

The role-profile map (sibling issue) deliberately exposes **two separate resolvers** for exactly
this:

```ts
resolveProfilesByTitles(query.titles)  // → what you call yourself
resolveProfilesBySkills(query.skills)  // → what your evidence describes
```

When those two answers disagree — different family, or a leadership profile on one side and an IC
profile on the other — that is the mismatch. No new machinery, just the comparison.

### Detection (pure lib)

Extend `src/lib/job-search/term-advisories.ts` (or `term-quality.ts`, whichever the sibling issues
landed) with a coherence check. Zero-dep, testable at module scope.

Fire only on a **confident** disagreement:

- both resolvers returned something (an unresolved side is not evidence of a mismatch — say
  nothing);
- the resolved profiles differ on family, or on the leadership/IC axis;
- the disagreement is not marginal — define and name a threshold constant rather than firing on any
  difference. A senior IC with some mentorship skills is not mismatched, and a false positive here
  tells someone their résumé is broken when it isn't.

Return the naming detail the copy needs: which titles, which skill shape, and the expected skills
the résumé lacks.

### Presentation

Render inside the existing grouped term advisory on `/jobs/` — **do not add a fourth panel.** The
advisory already caps at ~3 items and each item names a term and its consequence.

```
⚠︎ Check these terms
 • Your titles are leadership roles, but your listed skills are all
   individual-contributor technical skills. Postings for these titles usually
   ask for people management, hiring, and roadmap ownership.
```

Rules that already govern that component and apply here unchanged:

- name the **consequence**, never the mechanism — no "resolver", no "profile", no issue numbers;
- `⚠︎` text-presentation mark, not the colour-emoji variant;
- semantic tokens only; never colour alone;
- detection only — **never auto-remove or auto-add anything**;
[...]

PR #673, opened 2026-07-30
test(job-search): pin the skills half of the advisory adjacency claim (#595)

### LINKED ISSUE #595: jobs: term-quality advisory renders detached from the fields its pills write into
## Symptom

`TermQualityAdvisory` (#585) renders **full-width, below the Target level ladder**, two sections away from the Titles and Skills columns its pills write into. Confirmed in the browser, not inferred — it reads as a detached page footer rather than as advice about the fields above it.

The pills are grouped by destination and each group is headed with the field it writes to ("Add to your titles…", "Add to your skills…"), so the *copy* carries the association. But the copy is compensating for the placement. Clicking a pill mutates a control the user has to scroll back up to see change.

## Why it landed there

Full width was the safe choice — the advisory has two groups plus an optional coherence sentence, and neither the Titles nor the Skills column is wide enough to hold both without wrapping badly. Nothing about the current position is a bug; it is the position that required no layout decision.

## What to decide

Options, roughly in increasing cost:

1. **Move it above the Target level ladder**, still full-width — adjacency without a layout change. Cheapest, and probably most of the win.
2. **Split it** — title pills under the Titles column, skill pills under the Skills column, coherence sentence full-width. Best adjacency; costs the grouped-heading design, which then becomes redundant.
3. Leave as-is and accept the copy carries it.

Option 2 interacts with the query form folding into `JobQuerySummary` on Search, so check both states before committing to it.

## Acceptance

A pill and the field it writes into are visible together at a common laptop viewport without scrolling, or a written note saying why that was not achievable and which option was chosen instead. No regression to the folded-summary state.

rohithgollapalli:
Closes #595.

## What I found

The advisory is **already adjacent** on `main`. #595 described the pre-#602 layout — a full-width block below the Target level ladder — and #597/#602 (merged ~21h after #595 was filed, which is why the issue was never closed) implemented **option 2**, the split the issue ranked best:

- `TermQualityAdvisory kind="title"` renders inside the Titles `QueryStepSection`, directly under the chip list it writes into (`JobQueryEditor.tsx:198`).
- `TermQualityAdvisory kind="skill"` renders inside the Skills section, same shape (`JobQueryEditor.tsx:262`).
- The Review step keeps only the **whole-query** findings — the #587 coherence note and the #597 dropped terms — and is deliberately not passed `missing`, so it carries no pill at all (`JobQueryEditor.tsx:343`).

The grouped headings survived the move rather than becoming redundant, because a pill's destination is still worth naming in words at the moment of the click.

## What this PR adds

Only what was missing: the **regression test for the skills half**. Adjacency was pinned per-panel for titles (`"renders the title suggestions inside the Role step"`) and not for skills, so a pill drifting back into the trailing Review block would have been caught on one side only. The new case asserts, scoped to each mounted `StepPanel`:

- the skill suggestions are inside the Skills panel and not in Role,
- neither suggestion heading appears in Review,
- and that the Review panel element actually exists — without that, the `not.toContain` assertions would be checks against `undefined`.

Plus one docblock line in `TermQualityAdvisory.tsx` tying the split placement to the issue it closed, so the next person reading "TWO PLACEMENTS" knows adjacency is a requirement and not a style preference.

## Verification, stated honestly

Panel-scoped DOM assertions over a fully mounted stepper (inactive panels stay mounted, only `hidden`), not a browser screenshot at a laptop viewport. The acceptance criterion's "visible together without scrolling" is satisfied structurally — the pills are inside the same section as their field, which is stronger than a single viewport measurement — but I did not re-measure it in a browser. The folded-summary state is unaffected: the pills live inside the form, so they fold with it, and `JobQuerySummary` is untouched.

## Gates

`npm run typecheck`, `npx eslint src`, `npx vitest run`, `npm run build` — all green locally.

## Provenance

Implemented by Claude Code (Opus 5) with maintainer review.

### DIFF
--- a/src/components/features/JobQueryEditor.test.tsx
+++ b/src/components/features/JobQueryEditor.test.tsx
@@ -303,6 +303,30 @@ describe("JobQueryEditor — dropped terms and in-column suggestions (issue 597)
     expect(rolePanel?.textContent).toContain("Add to your titles");
     expect(skillsPanel?.textContent).not.toContain("Add to your titles");
   });
+
+  it("renders the skill suggestions inside the Skills step, and no pill in Review (#595)", () => {
+    // The other half of the adjacency claim, and the one #595 was filed about:
+    // every pill has to be visible with the field it writes into. Asserted per
+    // panel, because the failure mode is a pill drifting back into the trailing
+    // Review block — which is page-wide text a whole-page search would still find.
+    const query: JobQuery = { titles: ["Engineering Manager"], skills: [] };
+    const missing = assessQueryTerms(query).missing.filter((t) => t.kind === "skill");
+    expect(missing.length).toBeGreaterThan(0);
+
+    const el = render(query, () => query);
+    const skillsPanel = el.querySelector("#test-steppanel-skills");
+    const rolePanel = el.querySelector("#test-steppanel-role");
+    const reviewPanel = el.querySelector("#test-steppanel-review");
+    // Asserted, not assumed: a renamed panel id would turn every `not.toContain`
+    // below into a check against `undefined`.
+    expect(reviewPanel).not.toBeNull();
+    expect(skillsPanel?.textContent).toContain("Add to your skills");
+    expect(rolePanel?.textContent).not.toContain("Add to your skills");
+    // Review carries whole-query findings only — the coherence note and the
+    // dropped terms — so neither suggestion heading may appear there.
+    expect(reviewPanel?.textContent).not.toContain("Add to your skills");
+    expect(reviewPanel?.textContent).not.toContain("Add to your titles");
+  });
 });
 
 /**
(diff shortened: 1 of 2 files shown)


s-annam (approved):
The PR successfully pins the skills half of the advisory adjacency claim by adding the missing panel-scoped DOM assertions to `JobQueryEditor.test.tsx` and documenting the layout requirements in `TermQualityAdvisory.tsx`.

- [x] A pill and the field it writes into are visible together at a common laptop viewport without scrolling (verified via panel-scoped DOM checks).
- [x] No regression to the folded-summary state.

All gates (`npm run typecheck`, `npx eslint`, `npx vitest run`) pass successfully.

---
Reviewed by: Gemini 1.5 Flash (high)

PR #674, opened 2026-07-30
fix(ui): lead with the drop zone before a parse and the score after it

### LINKED ISSUE #676: UI clarity pass: drop zone hierarchy + plain-language tab labels (shipped in #674)
Reference issue for #674 — filed after the fact so the work could start
immediately. The intern sync on Jul 29 reached agreement on the direction, and
the friends-and-family reach-out is gated on it, so this records what shipped
rather than proposing it.

**Updated Jul 30**: #674 grew past its original scope over three review rounds
and now covers the post-drop screen as well. This body describes the whole
change; the follow-up list at the bottom shrank accordingly.

## What was wrong

Three defects, all found by watching someone use the app rather than by reading
the code.

**The pre-drop screen buried its own call to action.** `CapabilityStrip` and a
`bg-surface-card-warm` hero Card both rendered above the drop zone. The hero
owned the page's only tinted surface and a `text-3xl` headline; the drop zone
had a `border-border` dashed edge, the default background, and a `text-sm`
prompt. ~90 words across three non-actionable blocks came before the one
control that does anything.

> "Everything is the same color, so it's kind of hard to find what am I
> supposed to do here. But the first thing you see is *your whole job search in
> your browser*. What does that mean? And I'm not even doing anything."

**The tab labels were internal vocabulary.** Someone who already knows the
product still had to open each tab to learn what it held.

> "What is a reconstructed resume? What is a parser? What does resume quality
> mean? You have to read a lot of stuff, and it's not clear just seeing it in
> one second."

**The post-drop screen opened on our suggestion, not the user's result.**
`LlmEscapeHatchBanner` — full-width, tinted, primary button — rendered *above*
the score card whenever the cascade flagged a degenerate parse. The user's own
score was the second thing on the page.

### Pre-drop

- Hero demoted to plain text, two lines, ~20 words — no Card, no tinted surface.
- Drop zone is the only accent surface on the page, with a `text-lg` /
  `sm:text-2xl` prompt. `sm:text-xl` (20px) was the only `text-xl` in all of
  `src/`; the app steps 14 → 16 → 18 → 24 → 30.
- `CapabilityStrip` moved below the drop zone and made pre-drop only — it was
  previously pinned in the header for the whole session.
- **PageShell's `subtitle` is now optional and `/` passes none.** "A parser
  audit for your resume — not a judge" used internal vocabulary, argued against
  an objection the visitor has not formed, and was a second tagline two inches
  above the hero. `/jd-fit` and `/jobs` still pass one — they open straight
  into a form with no headline of their own.
- **"No resume yet? Build one from scratch" grouped with the drop zone**
  (`gap-3`) instead of sitting below three unrelated blocks. The saved-resume
  library stays outside that group: it self-hides when empty and would leave a
  phantom gap inside it.
- **The block under the drop zone rewritten, 85 words → 27**, opening on what
  the product does in the order a user does it (parse → edit → download).
  Its three paragraphs now share one `mx-auto max-w-3xl` wrapper — applied
  per-paragraph they centred independently, so the wrapping line read
  left-aligned while the short ones centred (three different left edges).

### Post-drop

- **The recovery offer (#243) moved out of the banner above the score card and
  into the on-device-AI tab.** The score card is now the first thing on the
  screen.
- **The tab label carries the signal**: "Try a local AI pass" (warn-marked)
  while the offer stands, "Local AI feedback" otherwise. The tab strip already
  reserved permanent space for this surface, so the offer costs no new space.
  The label reverts once the pass has run — the controller stays `isAvailable`
  afterwards so the pass can be re-run, which would otherwise keep inviting a
  pass already taken.
- **One offer at a time.** While recovery stands it owns the tab body alone;
  `ResumeQualityPanel` is withheld. A wording critique of a parse the parser
  itself flagged as degenerate is close to worthless, and stacking both put two
  model-loading CTAs on one screen.
- `LlmEscapeHatchBanner` → **`LlmEscapeHatchPanel`**, restyled into
  `ResumeQualityPanel`'s shape (section + heading + `max-w-prose` explainer +
  right-aligned primary Button) since the two alternate in the same body.
- `Tab` gains **`warnLabel`**. The warning dot's `sr-only` addendum was
  hardcoded "setup needed" — right for the WebGPU case, and it would send a
  screen-reader user looking for a browser setting that does not exist in the
  recovery case.

### Tab labels

Reconstructed resume → **Your resume**, Resume quality → **Local AI feedback**
(or **Try a local AI pass**), Source & diagnostics → **Raw text & flags**. Ids
untouched — they are the tab-switch contract with `CritiquePanel` and
`ResumeQualityPanel`'s `onGoToRewrite`.

`DropZone.test.tsx` pins the pre-drop hierarchy and `ResultDetailTabs.test.tsx`
pins the four label/warn states, so a later "tone this down" edit cannot
silently undo either.

## Two copy findings worth recording

**The Greenhouse citation never sourced the claim above it.** The landing block
paired "recruiters are starting to run AI agents over resumes" with the
*2025 AI in Hiring Report*, and a code comment asserted the link outright. That
report measures AI **trust** among hiring managers (70% trust it; 8% of job
seekers call it fair), not screening prevalence. The trend sentence was cut,
not hedged, and the citation now stands under the weaker claim it does support.

**One deletion removed a framing from the whole product.** Cut with that
sentence: "the score rates how readable your file is, not how good you are" —
while the same PR deleted the header subtitle carrying the same idea ("not a
judge"). `AtsScoreReadout` never stated it, so the framing is now nowhere in
the app. Flagged in a code comment rather than silently restored. **If it comes
back it belongs next to the score, not on the landing** — that is a decision,
not a bug, and it is open.

Privacy copy scope is unchanged — headline claims a place, subhead claims the
resume never leaves the browser, `keywords.ts` still the invariant.

## Follow-ups

Moved to #680 (Find Jobs repro, the feedback row's placement in the score card,
and whether the "the score rates the file, not you" framing comes back next to
the score). This issue closes with #674.

### Closed by this PR

- ~~WebLLM escape-hatch copy.~~ "Try a local AI pass" is now the tab label
  itself, with the explainer as the tab's body — the label is the offer and the
  paragraph beside it is the answer to "what does that do?".
- ~~"Show the score on top, drop the two boxes."~~ Resolved: the box above the
  score was the recovery banner, which is now gone from that position. The
  score card is first.

s-annam:
Closes #676
Follow-ups tracked in #680

Two defects found in user testing on Jul 29 that both read as "the product is
confusing" but were measurable layout and copy problems, not comprehension gaps.

## 1. The pre-drop screen buried its own call to action

`CapabilityStrip` (through PageShell's `chips` slot) and a
`bg-surface-card-warm` hero Card both rendered above the drop zone. The hero
owned the page's only tinted surface and a `text-3xl` headline; the drop zone
sat on the default background behind a `border-border` dashed edge with a
`text-sm` prompt. A first-time visitor read roughly 90 words across three
non-actionable blocks before reaching the one control that does anything.

> "Everything is the same color, so it's kind of hard to find what am I
> supposed to do here. But the first thing you see is *your whole job search in
> your browser*. What does that mean? And I'm not even doing anything."

The hierarchy is now inverted:

- Hero is plain text, two lines, ~20 words — no Card, no tinted surface.
- Drop zone is the only accent surface on the page (`bg-accent-forward-bg` +
  `border-accent-primary`) with a `text-lg` / `sm:text-2xl` prompt.
- `CapabilityStrip` moves below the drop zone and renders pre-drop only. It was
  previously pinned in the header for the entire session, restating three lanes
  over the user's own score after they had already chosen one.
- The Greenhouse citation moves down to the recruiter-agent block — the claim
  it actually sources. In the hero it was a third competing message ahead of
  the primary action.

The privacy copy is unchanged in scope: the headline still claims a *place*,
the subhead still claims the *resume* never leaves the browser, and
`src/lib/job-search/providers/keywords.ts` remains the invariant both depend
on. The subhead drops only its lane list ("score it, fix it, match a JD, and
find jobs"), which `CapabilityStrip` now enumerates a few inches below with
fuller descriptions.

## 2. The tab labels were internal vocabulary

A reader who already knows the product still had to open each tab to find out
what it held.

> "What is a reconstructed resume? What is a parser? What does resume quality
> mean? You have to read a lot of stuff, and it's not clear just seeing it in
> one second."

| Before | After |
|---|---|
| Reconstructed resume | Your resume |
| Find jobs | Find jobs (unchanged) |
| Resume quality | Local AI feedback |
| Source & diagnostics | Raw text & flags |

Tab **ids are untouched**, so the tab-switch contract with `CritiquePanel` and
`ResumeQualityPanel`'s `onGoToRewrite` is unaffected. The `description`
subtitles (#519) still carry the precise meaning — the label only has to be
decodable without clicking.

`ReconstructedResume`'s section heading becomes "What the parser read" rather
than echoing its own tab label. A heading that repeats its tab is noise, and it
would make every `toContain("Your resume")` assertion ambiguous about which
element it matched.

## Testing

New `src/components/DropZone.test.tsx` pins the hierarchy inversion, which
nothing else in the suite would catch. Assertions target the *relative*
treatment — accent surface present, prompt a size step above body copy — not
specific token values, so a palette swap stays free.

Revert-proof, reported honestly: **3 of its 5 assertions fail** against the
pre-fix component. The other two (the file-custody line, `border-dashed`) cover
behaviour this PR deliberately preserves, so they pass either way.

`ResultDetailTabs.test.tsx` updated for the new labels. Full `npm run verify`
green (fallow's 3 complexity findings on `App`, `ModelSelector` and
`AchievementHeader` are pre-existing and report-only).

Browser-verified at 1280x900 and 390x844: the entire landing column now fits
above the fold on desktop, and the drop zone is fully visible on mobile without
scrolling. Post-parse verified too — the capability strip is gone, the score
card is first, and the new labels render with their descriptions.

## 3. Second round (Jul 30)

Four follow-ups on the same screen, after looking at the shipped version.

**The header subtitle is gone.** `PageShell`'s `subtitle` prop becomes optional
and `/` stops passing one. It read *"A parser audit for your resume — not a
judge"*, which failed three ways at once: "parser audit" is exactly the internal
vocabulary this PR removes from the tab labels; "not a judge" is a negation
defending against an objection the visitor has not formed yet; and on the idle
screen it was a second tagline sitting two inches above the headline. Removing
it also leaves the GitHub-star CTA alone on the header-right instead of sharing
it. `/jd-fit` and `/jobs` still pass one — they open straight into a form with
no headline of their own, so there it is the only orientation.

**"No resume yet? Build one from scratch" moved up.** It sat below the
capability strip, the saved-resume library and the job tracker — three
unrelated blocks. It is not a topic of its own; it is the other way to answer
the question the drop zone asks, so it is now in a `gap-3` group with the drop
zone rather than a `gap-6` section apart from it. The library stays outside that
group deliberately: it self-hides when empty, and inside the group an invisible
block would leave a phantom gap.

**Prompt font size.** `sm:text-xl` (20px) was the only `text-xl` in all of
`src/` — every other heading in the app steps 14 → 16 → 18 → 24 → 30. A lone
20px is an arbitrary size in the sense the Font Size Scale rule means, and 20px
also read small inside a 56px-padded box. Now `sm:text-2xl`, a step the app
already uses. The hero headline remains one step above at `sm:text-3xl`; the
drop zone leads on colour and fill, which is the contrast that survives a
greyscale render.

**The block below the drop zone was rewritten, then cut again.** It opened on
the trend — *"Recruiters are starting to run AI agents over resumes"* — so the
first thing a visitor read was someone else's behaviour, and what they
personally get arrived in clause four, as a metaphor ("the candidate-side
mirror"). It now opens on what the product does, in the order a user does it:

> **OfflineCV parses your resume, shows you what a machine read back, and lets
> you edit it and download a cleanly formatted PDF or Markdown file.**

85 words → 27. Both export paths are real, so the claim round-trips to code:
`useDownloadPdf.ts` → `render-ats-pdf.ts` for PDF, `useDownloadMarkdown.ts` for
Markdown (#552). `CapabilityStrip`'s "Fix it" lane now names Markdown too — it
had stopped at PDF since before #552 shipped.

The recruiter-trend sentence was then cut outright rather than demoted: an
unhedged "recruiters increasingly screen with AI" is a claim about an industry
we would have to defend, and the Greenhouse report cited below it measures AI
*trust* among hiring managers, not screening prevalence — it never sourced that
sentence as tightly as the old code comment claimed. The citation now stands on
its own weaker, supportable claim ("a hiring process job seekers say they can't
see into").

⚠️ **One thing left the app entirely.** Cut with that sentence was *"the score
rates how readable your file is, not how good you are"* — and the header
subtitle that used to carry the same idea ("not a judge") is deleted above.
`AtsScoreReadout` never stated it. That framing is now nowhere in the product.
Flagged rather than silently restored: if it comes back it belongs next to the
score, not on the landing page. A code comment in `App.tsx` records this.

Also fixed while in there: the block's three paragraphs each carried their own
`mx-auto max-w-3xl`, so the wrapping paragraph read as left-aligned while the
short ones visibly centred themselves — three lines, three left edges. One
shared wrapper now centres the group once.

Re-verified in a browser at 1280x900 and 390x844. Full suite: 3849 passed, 10
skipped. Branch rebased onto `main` to pick up #675 (`npm run dev:http`, which
is what made the browser pass possible).

## Not in this PR

Three items from the same session were left out; two are now closed by round
3, and the rest are tracked in #680.

- **WebLLM escape-hatch copy.** The complaint was that "Try a local AI pass"
  doesn't explain what it does or why. The label is jargon, but the paragraph
  beside it in `LlmEscapeHatchPanel` already explains the what, the why, the
  privacy scope and the 1.2 GB download. Partly refuted by the code — worth a
  separate look rather than churn here. **Partly addressed in round 3**: the
  label is now the tab, and the explainer sits under it as the tab's whole
  body.
- **"Show the score on top, drop the two boxes."** The score already renders
  above the tabs. Which two boxes were meant is ambiguous from the recording;
  removing `CapabilityStrip` post-parse already removes one candidate.
- **A bug on the Find Jobs surface.** Named in passing with no repro. Needs one
  before it can be fixed.

## 4. Third round (Jul 30) — the post-drop screen

From a screenshot of the post-parse view: the page is busy, and the thing at
the very top is the **recovery banner**, not the score.

- **The recovery offer (#243) moves out of the top banner and into the
  on-device-AI tab.** It was a full-width, tinted, primary-button banner
  rendered above the score card — so the post-drop screen opened on our
  suggestion instead of the user's own result. The score card is now first.
- **The tab label carries the signal.** The tab strip already reserves
  permanent space for this surface, so the offer moves into space the layout
  was paying for anyway: the tab reads **"Try a local AI pass"** (warn-marked)
  while the offer stands, and **"Local AI feedback"** otherwise. That is also
  the whole pre-click signal that the parse was degenerate, which is why it
  takes the warning mark.
- **The panel's heading no longer repeats that label.** `ResumeQualityPanel`'s
  h2 rendered `"Local AI feedback"` — the same string as the tab above it, the
  repetition `ReconstructedResume` already rejected when it renamed its own
  heading to `"What the parser read"`. It now reads **"What the model
  checks"**: the heading says what the section does, the tab says where you
  are. Added in review (thanks @Vaishnavi1709) — the rule had been applied to
  one of the two tabs only.
- **The label reverts once the pass has run.** `escapeHatch.isAvailable` stays
  true after a successful recovery — it is keyed on the *original* result so
  the pass can be re-run — so without a `status.kind !== "done"` gate the tab
  would keep inviting a pass the user already took.
- **One offer at a time.** While recovery is on the table the tab body shows
  only the recovery panel; `ResumeQualityPanel` is withheld. A wording critique
  of a parse the parser itself flagged as degenerate is close to worthless, and
  stacking both put two model-loading CTAs on one screen — the first attempt at
  this shipped the old banner widget as-is *above* the quality panel, which is
  exactly the "two differently-dressed widgets" problem.
- **Restyled into the panel it alternates with**, and renamed
  `LlmEscapeHatchBanner` → `LlmEscapeHatchPanel`: same `section` + uppercase
  heading + `max-w-prose` explainer + right-aligned primary `Button` as
  `ResumeQualityPanel`. A user switching between the two states sees one
  surface change its offer.

### One thing worth reviewing

The panel **stays mounted** on `done` and collapses to a one-line confirmation
(plus a `link` re-run) rather than being unmounted by the caller. That is not a
style choice: the `done` transition is what fires `onRecovered`, and the status
lives in a hook owned by `ParsedCard`, so a `status.kind !== "done"` *mount*
gate would swap the panel out in the same render that produces the result — and
the recovered parse would never reach the score card. Covered by a test.

`Tab` gains a `warnLabel` prop. The dot's `sr-only` addendum was hardcoded to
"setup needed", which is right for the WebGPU case and would send a
screen-reader user looking for a browser setting that does not exist in the
recovery case.

### Verified

Full suite green (3852 passed / 10 skipped), typecheck + lint clean, build
clean. Browser-checked at 1280×900 against `npm run dev:http`, both states: a
clean parse (score at top, tab reads "Local AI feedback") and a degenerate one
(`tests/fixtures/pdfs/unknown/label-rail-grid-fragmented.pdf` — tab reads "Try
a local AI pass ⚠︎", and its body holds the offer alone).

## Provenance

Claude Opus 5, Claude Code. UI recommendations cross-checked against the
`ui-ux-pro-max` rule database (Hero-Centric landing pattern: one primary CTA,
hero minimal-text, CTA at high contrast; Typography/Heading-Clarity: a heading
needs a clear size and weight step over body). Its palette and typography
recommendations were **not** applied — they conflict with this repo's semantic
token system, and its "Exaggerated Minimalism" style suggestion is wrong for a
utility tool.

### DIFF
--- a/src/components/features/ResultDetailTabs.tsx
+++ b/src/components/features/ResultDetailTabs.tsx
@@ -8,10 +8,13 @@ import { FindJobsLauncher } from "./FindJobsLauncher.tsx";
 import { ResumeQualityPanel } from "./ResumeQualityPanel.tsx";
 import { SourceDiagnosticsPanel } from "./SourceDiagnosticsPanel.tsx";
 import { WebGpuUnavailableNotice } from "./WebGpuUnavailableNotice.tsx";
+import { LlmEscapeHatchPanel } from "./LlmEscapeHatchPanel.tsx";
 import type { CascadeResult } from "../../lib/heuristics/types.ts";
 import type { AnonymousAtsScore } from "../../lib/score/score.ts";
 import type { EditableParse } from "../../hooks/useEditableParse.ts";
 import type { AnalysisController } from "../../hooks/useResumeAnalysisLlm.ts";
+import type { EscapeHatchController } from "../../hooks/useLlmEscapeHatch.ts";
+import type { LlmParsedResume } from "../../lib/webllm/parse-resume.ts";
 
 type SourceKind = "pdf" | "docx" | "markdown";
 
@@ -25,6 +28,14 @@ interface ResultDetailTabsProps {
   edit: EditableParse;
   jdContext?: string;
   analysis: AnalysisController;
+  /**
+   * Degenerate-parse recovery pass (#243), owned by `ParsedCard` because its
+   * result re-grades the score card above this one. Rendered inside the
+   * on-device-AI tab rather than as its own banner — see the label logic below.
+   */
+  escapeHatch: EscapeHatchController;
+  /** Forwarded to the recovery banner; `ParsedCard` swaps in the LLM parse. */
+  onRecovered: (llmParsed: LlmParsedResume) => void;
   triggerCount: number;
 }
 
@@ -37,13 +48,15 @@ export function ResultDetailTabs({
   edit,
   jdContext,
   analysis,
+  escapeHatch,
+  onRecovered,
   triggerCount,
 }: ResultDetailTabsProps) {
   // `tab` state lives here — only used within this component, not in ParsedCard.
   const [tab, setTab] = useState("reconstructed");
 
-  // The "Resume Quality" tab is the canonical on-device-AI surface (#276). It
-  // shows whenever there's résumé text to analyze — either running the live
+  // The on-device-AI tab (id `quality`) is the canonical on-device-AI surface
+  // (#276). It shows whenever there's résumé text to analyze — either running the live
   // analysis (WebGPU available) OR, when WebGPU can't run here, explaining that
   // in place instead of silently vanishing. `capability === null` (still
   // detecting) and "no text" both leave the tab absent, as before.
@@ -53,7 +66,43 @@ export function ResultDetailTabs({
     analysis.capability !== "available"
       ? analysis.capability
       : null;
-  const showQualityTab = analysis.isAvailable || unavailableCapability !== null;
+  // `escapeHatch.isAvailable` already IMPLIES `analysis.isAvailable` — both gate
+  // on `capability === "available"` and on the identical `hasText` expression
+  // over the same `result`, and the hatch adds `suggestedEscalation === "llm"`
+  // on top. The `||` is therefore redundant today and deliberate anyway: the two
+  // gates live in two hooks that can drift, and the failure mode if they do is
+  // an unreachable recovery offer — the exact thing this tab now owns.
+  const showQualityTab =
+    analysis.isAvailable ||
+    unavailableCapability !== null ||
+    escapeHatch.isAvailable;
+
+  // The recovery offer is what this tab LEADS with while it stands, so the tab
+  // label is the offer (user request, Jul 2026: replace the label "when
+  // applicable, and use 'Local AI feedback' when it is not"). That buys the
+  // offer a permanent slot the layout was already paying for, instead of a
+  // banner above the score — but it also means the label is the only pre-click
+  // signal that the parse was degenerate, which is why it takes the warn mark
+  // too.
+  //
+  // Gated on `!== "done"`: the hatch stays `isAvailable` after a successful
+  // recovery (it is keyed on the ORIGINAL result so the pass can be re-run), so
+  // without this the tab would keep inviting a pass the user has already taken.
+  const recoveryOffered =
+    escapeHatch.isAvailable && escapeHatch.status.kind !== "done";
+
+  // "Local AI feedback", not "AI feedback": in the tab strip the word that
+  // matters is the one saying the model runs here. The panel's own heading and
+  // the `description` below carry the rest.
+  const qualityLabel = recoveryOffered
+    ? "Try a local AI pass"
+    : "Local AI feedback";
+
+  const qualityDescription = recoveryOffered
+    ? "some of your file didn't parse cleanly — a local model can re-read it"
+    : analysis.isAvailable
+      ? "on-device AI review of your wording"
+      : "on-device AI review — needs browser support";
 
   return (
     /* Detail sits behind tabs in its own card so only one panel shows at a
@@ -65,16 +114,26 @@ export function ResultDetailTabs({
       <Tabs id="result" value={tab} onValueChange={setTab}>
         {/* Primary tabs ordered by value: insight first, evidence last
             (#263, #273). The evidence tab is always present and always last, so
-            the "Source & diagnostics" tab no longer shifts position when the
-            conditional Resume Quality tab is absent. The layout-flag count badge
+            the "Raw text & flags" tab no longer shifts position when the
+            conditional on-device-AI tab is absent. The layout-flag count badge
             is promoted to this parent tab so the warning count stays visible
-            without opening it. */}
+            without opening it.
+
+            Labels name what the user GETS, in words they arrive with. The
+            previous set — "Reconstructed resume", "Resume quality", "Source &
+            diagnostics" — was our internal vocabulary: in user testing (Jul
+            2026) a reader who knows the product still had to open each tab to
+            learn what it was ("what is a reconstructed resume? what is a
+            parser?"). The `description` subtitle (#519) carries the precise
+            meaning; the label only has to be decodable without clicking. Keep
+            the ids — they are the tab-switch contract with CritiquePanel and
+            ResumeQualityPanel's `onGoToRewrite`. */}
         <TabList aria-label="Parsed result views">
           <Tab
             id="reconstructed"
-            description="what a parser pulled out — edit it here"
+            description="what a parser read from your file — edit it here"
           >
-            Reconstructed resume
+            Your resume
           </Tab>
           <Tab
             id="find-jobs"
@@ -85,14 +144,16 @@ export function ResultDetailTabs({
           {showQualityTab && (
             <Tab
               id="quality"
-              warn={!analysis.isAvailable}
-              description={
-                analysis.isAvailable
-                  ? "on-device critique and rewrites"
-                  : "on-device critique — needs browser support"
+              warn={recoveryOffered || !analysis.isAvailable}
+              // Same dot, two meanings — so it says which. "setup needed" is
+              // right for the WebGPU case and wrong for the recovery one,
+              // where nothing is broken in the browser.
+              warnLabel={
+                recoveryOffered ? "parse needs attention" : "setup needed"
               }
+              description={qualityDescription}
             >
-              Resume quality
+              {qualityLabel}
             </Tab>
           )}
           <Tab
@@ -100,7 +161,7 @@ export function ResultDetailTabs({
             count={triggerCount}
             description="raw text, layout flags, what went wrong"
           >
-            Source &amp; diagnostics
+            Raw text &amp; flags
           </Tab>
         </TabList>
 
@@ -111,7 +172,7 @@ export function ResultDetailTabs({
               score={activeScore}
               edit={edit}
               jdContext={jdContext}
-              // #608: the critique the "Resume quality" tab is already showing
+              // #608: the critique the on-device-AI tab is already showing
               // feeds the rewrite, so clicking Rewrite acts on the findings the
               // user just read instead of discarding them. Only available once
               // the analysis has completed; every other status contributes
@@ -133,7 +194,35 @@ export function ResultDetailTabs({
           </TabPanel>
           {showQualityTab && (
             <TabPanel id="quality">
-              {analysis.isAvailable ? (
+              {/* One offer at a time. While recovery is on the table this tab
+                  shows ONLY the recovery panel: a wording critique of a parse
+                  the parser itself flagged as degenerate is close to
+                  worthless, and stacking both put two model-loading CTAs on
+                  one screen. Once the pass has run (or if it was never
+                  offered), the quality panel takes the tab back — with the
+                  recovery panel collapsed to its one-line confirmation above,
+                  because unmounting it on `done` would kill the effect that
+                  reports the recovered parse upward. */}
+              {/* The wrapper is unconditional on purpose, even though it is
+                  bare when `recoveryOffered`. Only the className varies, so
+                  React keeps the same element at this position and the panel
+                  stays mounted across the `done` transition. Gating the
+                  wrapper itself — rendering the panel bare in one branch and
+                  inside a div in the other — changes the element TYPE at this
+                  position at the moment recovery completes, which unmounts and
+                  remounts the panel in the render that fires `onRecovered`
+                  (see LlmEscapeHatchPanel's docblock). Not worth trading a
+                  documented mount invariant for one empty div. */}
+              {escapeHatch.isAvailable && (
+                <div className={recoveryOffered ? undefined : "mb-4"}>
+                  <LlmEscapeHatchPanel
+                    controller={escapeHatch}
+                    onRecovered={onRecovered}
+                  />
+                </div>
+              )}
+              {!recoveryOffered &&
+                (analysis.isAvailable ? (
                 /* onGoToRewrite: switch back to reconstructed tab where the
                    per-role wand button (#3 / useSectionRewrite) already lives.
                    The quality panel links each flagged bullet to this affordance
@@ -145,12 +234,15 @@ export function ResultDetailTabs({
                 />
               ) : (
                 /* WebGPU can't run here — explain in place instead of hiding
-                   the tab (#276). `unavailableCapability` is non-null whenever
-                   this branch renders (see showQualityTab). */
-                unavailableCapability && (
-                  <WebGpuUnavailableNotice capability={unavailableCapability} />
-                )
-              )}
+                   the tab (#276). Still guarded: `showQualityTab` now has a
+                   third opener (`escapeHatch.isAvailable`), and while that one
+                   implies `analysis.isAvailable` — so it takes the branch above,
+                   never this one — the guard is what makes that reasoning
+                   non-load-bearing. */
+                  unavailableCapability && (
+                    <WebGpuUnavailableNotice capability={unavailableCapability} />
+                  )
+                ))}
             </TabPanel>
           )}
           <TabPanel id="diagnostics">
--- a/src/components/features/ResumeQualityPanel.tsx
+++ b/src/components/features/ResumeQualityPanel.tsx
@@ -2,7 +2,7 @@
 // Copyright 2026 The offlinecv Authors
 
 /**
- * ResumeQualityPanel — consolidated "Resume Quality" tab shell (issue #273).
+ * ResumeQualityPanel — consolidated "AI feedback" tab shell (issue #273).
  *
  * Replaces the two separate "What an ATS misses" (DisagreementPanel) and
  * "Resume quality" (CritiquePanel) tabs with a single tab driven by one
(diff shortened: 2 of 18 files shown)


s-annam:
Screenshot 2026-07-30 101539
Screenshot 2026-07-30 101617

s-annam (commented):
### Stance
The PR is a highly thorough, example-quality improvement to the pre-drop and post-drop user experience. It directly addresses the visual hierarchy issues of the pre-drop screen and clarifies the internal terminology used on the post-drop tabs. All local verification steps (linting, typechecking, vitest suite, and build generation) are fully clean and passing.

*Verdict: COMMENT (downgraded from APPROVE for self-review; rule: 0 Blocking findings)*

### Findings

#### Blocking
None.

#### Secondary
* **Fallow Duplication Warning (Structural Parity):**
  `fallow` flags 35 lines of duplication between `src/components/features/LlmEscapeHatchPanel.tsx:112-146` and `src/components/features/ResumeQualityPanel.tsx:51-69`.
  *Verify:* Verified the lines. This structural duplication (same flex layouts and section styling) is a deliberate choice documented in the codebase to ensure visual consistency when the two panels alternate in the same tab space.

#### Nits
None.

---

src/components/features/ResumeQualityPanel.tsx
```diff
@@ -51,7 +52,7 @@ export function ResumeQualityPanel({
       <div className="flex items-start justify-between gap-3">
         <div className="flex flex-col gap-1">
           <h2 className="text-sm font-semibold uppercase tracking-wider text-content-muted">
-            Resume Quality
+            Local AI feedback
```
Vaishnavi1709: **Secondary.** Heading-vs-tab-label repetition is now inconsistent across the two tabs. `ReconstructedResume` deliberately renamed its section heading to `"What the parser read"` — the comment there argues: "A heading that repeats its own tab label is noise, and it makes every `toContain('Your resume')` assertion ambiguous." That reasoning applies identically here: this h2 renders `"Local AI feedback"`, the same string as its tab label.

Pick one house rule and apply to both: either rename this h2 (e.g. "AI review of your wording") or drop the anti-repetition rule so the two surfaces read the same way.

src/components/features/ResultDetailTabs.tsx
```diff
@@ -133,7 +194,25 @@ export function ResultDetailTabs({
...
-              {analysis.isAvailable ? (
+              {/* One offer at a time. While recovery is on the table this tab
+                  shows ONLY the recovery panel: a wording critique of a parse
+                  the parser itself flagged as degenerate is close to
+                  worthless, and stacking both put two model-loading CTAs on
+                  one screen. Once the pass has run (or if it was never
+                  offered), the quality panel takes the tab back — with the
+                  recovery panel collapsed to its one-line confirmation above,
+                  because unmounting it on `done` would kill the effect that
+                  reports the recovered parse upward. */}
```
Vaishnavi1709, line 205: **Nit.** This `<div>` is a no-op in one of its two branches. When `recoveryOffered=true`, it wraps `LlmEscapeHatchPanel` with no styling; the `mb-4` spacing is only needed in the `done` branch where the collapsed banner sits above `ResumeQualityPanel`. If you touch this again, simplify to `escapeHatch.status.kind === "done" && <div className="mb-4">...</div>` and mount without the wrapper otherwise.

Vaishnavi1709 (dismissed):
Approve. The pre-drop → post-drop hierarchy inversion lands cleanly, the escape-hatch relocation into the on-device-AI tab preserves the `onRecovered` transition (verified: hatch stays mounted through `done`, ref-guarded one-shot), and `Tab.warnLabel` defaults preserve back-compat for the existing WebGPU-unavailable call site. AC checklist for #676 is fully met.

## Nits

**Description ↔ code drift on the label mapping.** The section-2 table lists the new label as `AI feedback`, but the code (and the section-4 text later in the body) says `Local AI feedback`. Section 4 is what shipped; the section-2 table is stale after the round-3 tweak. Fix in-place in the PR body so the merged squash message doesn't carry the wrong label.

**`toBeTruthy` on the loading state doesn't assert what it means to.** In `src/components/features/LlmEscapeHatchPanel.test.tsx` the "renders loading, running, and error states" case gates the loading branch on `expect(...).toBeTruthy()`, which just checks the container has textContent. It doesn't verify `ModelLoadProgress` mounted with the right progress fraction or label. Not a defect this PR introduced (it's a rename), but worth a follow-up along with the running-state assertion in the same test.

---
Reviewed by: Claude Opus 4.7 (1M context, high)

src/components/features/ResumeQualityPanel.tsx
s-annam: Fixed in `283cc88`. Agreed the rule should apply to both tabs, and renaming was the right half of your either/or — dropping the anti-repetition rule would mean putting `"Your resume"` back as `ReconstructedResume`'s h2 and reintroducing the exact `toContain()` ambiguity its comment argues against.

The h2 now reads **"What the model checks"** rather than your suggested "AI review of your wording" — that string is verbatim `qualityDescription` (`ResultDetailTabs.tsx:104`), so it would have traded a collision with the tab *label* for one with the tab *description*. "What the model checks" parallels "What the parser read", and it reads correctly before the run as well as after (the heading sits above the Analyze CTA, so nothing has been "found" yet).

Your ambiguity point was literally true, not just theoretical: `ResultDetailTabs.test.tsx:194` asserted the panel had mounted via `expect(el.textContent).toContain("Local AI feedback")` — which `labels[2]` on line 189 already satisfied, so it never proved anything about the panel. Now scoped to the panel's own heading.

Also fixed the section-2 label table in the PR body (`AI feedback` → `Local AI feedback`) and documented the rename there.

src/components/features/ResultDetailTabs.tsx
s-annam: Pushing back on this one — I think the simplification is unsafe, and I added a comment at the site rather than the change.

The wrapper looks like a no-op in the `recoveryOffered` branch, but it is load-bearing as a **stable element position**. Today only `className` varies, so React keeps the same `div` across the transition and `LlmEscapeHatchPanel` stays mounted. The suggested shape — panel bare in one branch, wrapped in a `div` in the other — changes the element *type* at that position, and it changes it at the one moment that matters: `!recoveryOffered` becomes true exactly when `escapeHatch.status.kind === "done"` (given `isAvailable`, which the outer guard already establishes). So the swap happens in the same render that fires `onRecovered`, unmounting and remounting the panel there and resetting its `notifiedDone` ref.

That is the invariant `LlmEscapeHatchPanel`'s docblock calls out (lines 24–27): *"The panel stays mounted through `done` — that transition is what fires the callback, so a `status.kind !== "done"` mount gate in the parent would swap the panel out in the same render that produces the result and the callback would never fire."* Your suggested condition is that gate, one level up.

To be precise rather than alarmist: I did **not** verify it breaks outright — the remounted instance would mount already-`done` and its effect could still fire. But that makes correctness depend on remount-effect timing instead of on the panel staying mounted, at the transition the file explicitly protects, for the removal of one empty `div`. Not a trade I want to make in a UI PR.

Left unresolved for you to close or argue — happy to be wrong if you see a shape that drops the wrapper without moving the element type. `283cc88` adds a comment so the next reader does not re-derive this.

Vaishnavi1709: Fair — the remount-on-`done` risk outweighs the cosmetic win. Closing. Thanks for filing #687 for the invariant test.

Vaishnavi1709 (approved):
Re-approving after force-push. Changes since the last approval are the heading rename, the tightened test assertion, two comments, and a rebase onto main — nothing that warrants a re-review. Closing the `mb-4` thread; the remount-on-`done` argument is sound and #687 covers the missing invariant test.

---
Reviewed by: Claude Opus 4.7 (1M context, high)

Issue #689, opened 2026-07-30
[epic] Job library — a capture-fed, durable list of the jobs you're tracking


s-annam:
## Summary

`/jobs/` today is a **search** surface: build a query, hit the feeds and the company boards, rank by résumé fit, deep-link out. What it cannot do is the thing a job hunt actually consists of — **keep a list**. A posting the user finds anywhere other than our search (a company careers site, a referral, a newsletter, a friend's link) has nowhere to go, and a posting the user *does* find through our search survives only as long as the tab.

This epic closes that. It adds a durable job library on `/jobs/`, fed three ways — by hand, by file import, and by an optional browser extension that captures a posting from whatever site the user is standing on — with the whole record staying in the browser's IndexedDB.

The unlock is that **most of it already exists and is dark**:

- `JobStatus` is already the full lifecycle — `"interested" | "applied" | "interviewing" | "offer" | "rejected" | "archived"` (`src/lib/storage/types.ts`), with `JOB_STATUS_ORDER` for display.
- `JobTracker` / `JobTrackerSection` / `JobTrackerEntry` / `JobStatusPicker` are built and tested (145 LOC + siblings), rendered at `src/App.tsx:251` behind `useFlag("job-tracker")` (`src/lib/flags.ts:63`, `VITE_ENABLE_JOB_TRACKER`) since #323.
- `JobRecord` already carries `jdText` and `matchResult`, so a captured posting feeds `/jd-fit/` with no new field.
- `importFromJson` / `importAll` are implemented and unit-tested at `src/lib/storage/backup.ts:93` and re-exported at `src/lib/storage/index.ts:31` — and have **zero consumers**. The app ships an "Export backup" button (`useJobTracker.ts:147`, `useResumeLibrary.ts:123`) with no way to restore what it exports.

So the first two items are wiring, not construction. The extension is the only genuinely new build, and it is deliberately last: items 1–3 are independently useful and give items 4–5 a real target to write against instead of a hypothetical one.

## Why now

The search lane can only see what its providers can see. Providers are constrained by CORS — a browser cannot read a career site that omits `Access-Control-Allow-Origin`, which is most of them, including several of the largest employers. That is a property of *being a browser*, not of those sites lacking APIs, and no provider work fixes it.

Both remaining moves route around that rather than through it, and neither needs an API, a key, or a terms review.

[...]

PR #605, opened 2026-07-26
feat: primary role targeting, editable headline, and job-search handoff (#598)

s-annam:
## Summary

Surfaces the candidate's target role on the reconstructed résumé page (`/`), makes the headline a first-class editable field with round-trip validation, and carries that target into the job-search lane.

Closes #598
Closes #599
Closes #600
Closes #601

**What lands:**

- **`RolesPanel`** — a "Which role are you targeting?" card over the titles `deriveTitles` derives. Promoting a chip sets the `headline` override, which is drawn under the name on the downloaded PDF **and** leads the job-feed query. Both consequences are stated in the help copy because neither is guessable from the UI.
- **`ContactCard`** gains an inline-editable headline, validated by `headlineRoundTripWarning` so a user-typed tagline the parser would not read back is flagged before it reaches the export.
- **`SkillTermGuidance`** moves from the page footer to directly under `RolesPanel`, above Experience.

## Egress correctness

`titles[0]` is what `providers/keywords.ts` sends verbatim as the feeds' `search=` param — the one audited resume-derived egress. A parsed headline is a *tagline*, not a title: `extractHeadline` routinely lifts `"DevOps Engineer · Software Architect"`, and sending that whole is a single-intent query naming two intents.

`splitHeadline` now breaks a stacked headline on `· • | / ,` and spaced dashes, so the first part becomes a clean egress phrase and the rest still widen the local `matchesQuery` filter. `&` and `and` are deliberately **not** separators — "Founding Member & Site Reliability Engineer" is one compound role, and splitting it would mint titles nobody holds.

No title is starred until a headline actually exists. Defaulting the star to `titles[0]` claimed a role printed under the name while `render-ats-pdf.ts` — which guards on `model.contact.headline` — drew nothing; only 8 of 54 corpus fixtures parse a headline at all.

## Section order on `/`

AttentionStrip → ContactCard → **RolesPanel → SkillTermGuidance** → the résumé document.

`SkillTermGuidance` moved up because its content is a *function* of the starred role (`buildJobQuery` → `deriveTitles` → `titles[0]` → role profile). Star a different title and the whole suggestion list changes; four sections apart, that dependency was invisible and the panel read as a static footer. AttentionStrip and ContactCard were left in place — the strip's "N contact fields missing" segment points at the card directly beneath it.

[...]

Issue #539, opened 2026-07-22
[job-search] Find Jobs collapses multiple role titles into one search title


s-annam:
## Summary

Find Jobs derives a **single** search title from the most-recent role only. For anyone who has held multiple distinct titles (common in leadership — e.g. an executive whose prior roles were engineering-leadership or IC titles), the query collapses to one title and silently discards the others, narrowing the search to a single facet of the candidate.

Part of the job-search-v2 epic. Refs #528.

## Current behavior

`deriveTitle` in `src/lib/job-search/query-builder.ts:69-73` reads only `experience[0].title` (falling back to `current_title`). `JobQuery.title` is a single `string`. `FindJobsPanel` (`src/components/features/FindJobsPanel.tsx:166-174`) renders it as one `EditableField`.

Example (synthetic): a résumé whose most-recent role is an executive title and whose prior roles were engineering-leadership titles searches **only** the executive title — the engineering-leadership identity never reaches any board.

## Proposed change

Make title **multi-value**, mirroring the existing skills chip pattern:

- `JobQuery.title: string` → `titles: string[]` (or add `titles` alongside a primary), derived from the distinct titles across parsed `experience[]` (dedup, cap for URL/rank sanity).
- Render titles as removable chips (reuse `RemovableSkillChip`-style component), with an add-title affordance — same UX as skills today.
- Rank / deep-link against each title. `buildDeepLinks` and the ranker (`rank.ts`) consume the title set.
- Seniority derivation (`deriveSeniority`) currently keys off the single title — decide whether it derives from the primary title or the union (see sibling seniority issue).

## Acceptance criteria

- A résumé with N distinct role titles surfaces up to a capped set of title chips, not just the most-recent one.
- Each title chip is independently removable/addable, matching skills UX.
- Search + deep links + ranking use the full title set.
- Degenerate case (no titles) still falls back cleanly to a skills-only query.
- Unit tests in `query-builder.test.ts` cover multi-title derivation + dedup.

## Notes

Interaction with seniority + ranking is the main design call — flag in review. Keep the chip UI within the design-system seam (no raw `<button>`).

PR #696, opened 2026-07-31
feat(jobs,storage): a durable job library on /jobs/, a working restore path, and a validated capture contract (#689)

### LINKED ISSUE #547: Wire reconcileResumeLinks into the JSON import path
Part of #689. **Blocked on #573** — this cannot be wired until the merge-mode import UI exists.

## Context

`reconcileResumeLinks(existingResumeIds)` (`src/lib/job-tracker.ts`) is a belt-and-suspenders sweep that clears tracked-job resume links orphaned by a delete path that the explicit `clearResumeLink` call misses — specifically a **merge-mode JSON import that drops a resume**.

It landed with the job tracker (#323) but has **no production caller**: the resume-delete path uses `clearResumeLink`, and the JSON import path (`importAll` / `importFromJson` in `src/lib/storage/backup.ts`) is not yet wired to any UI, so there is nowhere live to invoke it. It's currently kept green only by its unit test.

## Task

When the JSON import UI lands, call `reconcileResumeLinks` after a merge-mode import completes, passing the set of resume ids that still exist, so any tracked job whose linked resume didn't survive the import degrades gracefully (link cleared, job kept) — the same guarantee `clearResumeLink` gives the delete path.

Until then the function carries a `// staged for #<this>` marker so it doesn't read as live-but-uncalled code.

Refs #323, #573

---

**Rehomed from #401.** The local-first storage batch was closed once its Phase 1 children shipped and #573 moved under epic #689; this issue moved with it. The dependency is unchanged: #573 builds the merge-mode import, and this is the sweep that runs after it. Small enough to land in the same PR as #573 if the implementer prefers.

### LINKED ISSUE #573: Restore path: wire backup import into the resume library, and scope the save-confirmation copy
Part of #689 (the epic of record for this work). Also tracked as Phase 2 of the local-first storage batch #401.

## Context

The local-first storage foundation (#321) ships a complete JSON backup round-trip in `src/lib/storage/backup.ts`: `exportAll` / `exportToJson` / `downloadStorageBackup` on the way out, `importAll` / `importFromJson` on the way back in. Blobs are base64-encoded and rebuilt byte-identically, and the document is origin-wide (both stores, `resumes` + `jobs`), so one file is the whole library.

**Only half of it is reachable.** Export is wired at two call sites — `ResumeLibrary.tsx:47` and `JobTracker.tsx:98`, both calling `library.exportBackup` / `tracker.exportBackup` → `downloadStorageBackup()`. `importAll` and `importFromJson` have **no non-test consumer**: `grep -rn "importAll\|importFromJson" src/ | grep -v '\.test\.'` returns only the barrel re-export at `src/lib/storage/index.ts:30-31`. They pass CI on unit tests alone.

So today a user can download `offlinecv-backup.json` and has nowhere to put it back. That matters more than a missing convenience, because IndexedDB is isolated along four axes the product never names:

| Axis | Effect |
|---|---|
| **Device** | Chrome Sync does not sync IndexedDB (it syncs bookmarks/passwords/tabs/settings only). Desktop and phone are separate stores even on the same signed-in profile. |
| **Browser profile** | `Default/IndexedDB/…` vs `Profile 1/IndexedDB/…` — separate. |
| **Origin** | `offlinecv.org`, `dev.offlinecv.org`, and every `*.offlinecv.pages.dev` preview deploy each get their own database. Subdomains do **not** share, unlike cookies. |
| **Eviction** | `navigator.storage.persist()` is requested on first save (`useResumeLibrary.ts:84`, `useJobTracker.ts:83`) but the grant is heuristic and covers only disk-pressure eviction. Safari/WebKit clears script-writable storage after 7 days without a visit — which includes **Chrome on iOS**, since it is WKWebView-backed. |

The word "library" carries a Spotify/Kindle promise — a durable collection that follows you. Restoring a backup is the only mechanism that can partially honour it, and it isn't built.

### The placement trap

`ResumeLibrary.tsx:29` early-returns `null` when `entries.length === 0`. A fresh browser, a new profile, a new device, or a post-eviction visit are all empty-library states — **exactly** when a user needs to import, and exactly when the surface that would hold the button renders nothing. Adding an "Import backup" button beside "Export backup" without touching the early return produces a control nobody in the restore scenario can see.

### Downstream

#547 (`Wire reconcileResumeLinks into the JSON import path`) is explicitly blocked on this: `reconcileResumeLinks` in `src/lib/job-tracker.ts:136` is staged, has no production caller, and its stated trigger is "a merge-mode JSON import that drops a resume". Merge-mode import is what this issue builds, so #547 becomes actionable — and is best done as its own follow-up, not folded in here.

### Secondary: save-confirmation copy

`SaveResumeBar.tsx:60` is the moment expectation gets set, and it says only:

```
savedId === null
  ? "Keep this resume to reload later — saved only in this browser."
  : "Saved to your library on this browser."
```

"on this browser" is a scope hint but stops short of the device boundary and offers no next step. The stronger copy already exists one surface away — `ResumeLibrary.tsx:53-56` renders an unconditional "Saved only in this browser — no account, no sync." plus a `!persisted`-gated `EVICTION_NOTICE` — but a user may save many times without ever opening the library card. Note the `!persisted` gate there is **correct as-is**: it wraps only the eviction-specific sentence, while the scope sentence renders unconditionally. Do not widen it.

## Reuse analysis

**Capability:** let a user select a `offlinecv-backup.json` file from disk and restore its records into this origin's IndexedDB, with a merge-vs-replace choice.

**Existing surfaces found:**
- `src/components/features/ResumeLibrary.tsx:47` — already owns "Export backup", the storage-persistence badge, usage bytes, and the eviction copy. This is the storage-management surface.
- `src/components/features/JobTracker.tsx:98` — a second "Export backup" button producing the same origin-wide document.
- `src/design-system/primitives/Dialog.tsx` — modal primitive (`open` / `onClose` / `title` / `ariaLabel`), used for the confirm.
- `src/design-system/primitives/Button.tsx` — the only button allowed in feature code.
- `src/components/DropZone.tsx` — a file-accepting surface, but typed to `RESUME_ACCEPT_ATTR` (`src/lib/file-accept.ts`) and semantically "drop a resume to analyse". Not reused: conflating backup-JSON restore with the primary analyse-a-PDF flow would overload the app's main entry point.

**Decision:** **extend `ResumeLibrary`.** No new component, no new panel. Import lives beside the export button it mirrors, using `Button` + a visually hidden `<input type="file" accept="application/json,.json">` and `Dialog` for the confirm. The one structural change is relaxing the empty-library early return so the card still renders a minimal import affordance. `JobTracker` is deliberately left alone — one entry point for a destructive-capable action, and the document is origin-wide, so importing from the library restores tracked jobs too.

### 1. `src/hooks/useResumeLibrary.ts` — expose an import action

Add to the `ResumeLibrary` interface (after `exportBackup`, line 55):

```ts
  /** Restore a backup file into this origin's storage. `merge` upserts by id;
   *  `replace` wipes both stores first. Refreshes the list on success. */
  importBackup: (file: File, mode: "merge" | "replace") => Promise<{ resumes: number; jobs: number }>;
```

Implement with `importFromJson` from the storage barrel, passing the mode **explicitly** — `importAll`'s own default is `"replace"`, and the UI default is `"merge"`, so an omitted argument silently inverts the intent:

```ts
const importBackup = useCallback(async (file: File, mode: "merge" | "replace") => {
  const counts = await importFromJson(await file.text(), mode);
  await refresh();
  return counts;
}, [refresh]);
```

Hand-audit the `useCallback` dep array: `refresh` is the only dependency and it is itself a `useCallback`. `exhaustive-deps` is **not** enforced in this repo (`eslint.config.js` registers no `react-hooks` plugin), so a stale closure lints green — check it by eye.

### 2. `src/lib/storage/backup.ts` — validate before mutating

`importAll` currently checks `data.version !== 1` (line 73) but only *after* the caller has already parsed arbitrary JSON, and `importFromJson` casts with `JSON.parse(json) as StorageExport` (line 97) — an unchecked assertion. A user picking the wrong file (or a truncated one) reaches `clearStore` in replace mode before anything notices the shape is wrong.

Add a shape guard used by `importFromJson` before it delegates:

```ts
function assertStorageExport(value: unknown): asserts value is StorageExport {
  const v = value as Partial<StorageExport> | null;
  if (v === null || typeof v !== "object" || v.version !== 1 || !Array.isArray(v.resumes) || !Array.isArray(v.jobs)) {
    throw new Error("Not an offlinecv backup file.");
  }
}
```

Also wrap the `JSON.parse` so a non-JSON file surfaces the same friendly error rather than a raw `SyntaxError`. The clear-then-write ordering inside `importAll` (lines 76-79) stays as-is — validation moves ahead of it, which is what makes replace mode safe.

### 3. `src/components/features/ResumeLibrary.tsx` — the import affordance

- Destructure `importBackup` from `library` (line 25).
- Replace the line-29 early return. Keep `!ready` returning `null`; when `ready && entries.length === 0`, render a minimal `Card` with the heading, the scope sentence, and the Import button — no entry list, no usage/persistence badge (both meaningless at zero).
- Add an "Import backup" `Button variant="ghost" size="sm"` beside the existing export button (line 47), driving a `useRef` on a hidden file input.
- On file selection, open a `Dialog` confirm carrying the mode choice. Default the radio to **merge**. The replace option's label must state what it destroys, using the live `entries.length`.
- Reset `input.value = ""` after each selection so re-picking the same file re-fires `onChange`.
- On success, swap the card's status line to a success message (`"Restored 3 resumes and 12 jobs."`) with `aria-live="polite"`. **There is no toast primitive in the design system** — confirm in place; `InlineResult` with `tone="success"` or a `StatusBadge tone="ok"` is the house pattern. Never signal meaning by colour alone.
- On failure, surface the thrown message via `ErrorState` or an `InlineResult` with an error tone. Do not swallow it.

Watch the ~200 LOC feature-component budget — `ResumeLibrary.tsx` is 71 LOC today, and the confirm dialog is the bulk of the addition. If it pushes past ~150, extract the dialog into a sibling `ResumeLibraryImportDialog.tsx` rather than growing the file.

### 4. `src/components/features/SaveResumeBar.tsx` — scope the confirmation

Change the saved-state string (line 60) so the confirmation names the device boundary and points at the backup path. Suggested, adjust for tone:

```
"Saved in this browser on this device. Export a backup to move it elsewhere."
```

Keep the pre-save string as-is. This is copy only — no new state, no new props, no link (the export button lives in the library card below).

### 5. Tests

- `src/lib/storage/storage.test.ts` — `importFromJson` rejects non-JSON, rejects a well-formed JSON object that isn't an export, rejects `version !== 1`, and (the load-bearing one) **does not clear either store when validation fails in replace mode**.
- New `src/components/features/ResumeLibrary.test.tsx` (or extend if present) — renders the import affordance at `entries.length === 0`; a chosen file opens the confirm; confirming with merge calls `importBackup` with `"merge"`; cancelling calls nothing; a rejected import renders the error and leaves entries untouched.
- Round-trip: export a populated store, wipe, import, assert resume blobs are byte-identical and job records deep-equal. `backup.ts` promises byte-identical blobs — pin it end-to-end through the hook, not just the lib function.

## Acceptance criteria

- [ ] `importFromJson` / `importAll` have at least one production (non-test) caller reachable from the UI.
- [ ] With zero saved resumes, the resume-library card still renders and offers "Import backup" — verified in a browser with a cleared origin, not only in jsdom.
- [ ] Selecting a valid `offlinecv-backup.json` and confirming restores resumes **and** tracked jobs; the list refreshes without a manual reload.
- [ ] Import mode defaults to **merge**; existing saved resumes not present in the backup survive. Replace is reachable only by explicitly choosing it, and its label states how many records it will destroy.
- [ ] A non-JSON file, a JSON file that is not an offlinecv backup, and a `version !== 1` document each produce a readable in-place error, and **leave existing storage byte-for-byte unchanged even in replace mode**.
- [ ] Picking the same file twice in a row re-triggers the flow (input value reset).
- [ ] Success and error are announced in place with `aria-live="polite"`; meaning is never carried by colour alone; the dialog is reachable and dismissable by keyboard (Esc closes, focus returns to the trigger).
- [ ] `SaveResumeBar` saved-state copy names the device boundary and points at the backup path.
- [ ] A `version` newer than 1 fails with a message that names the version mismatch, not a stack trace.
- [ ] `JobRecord.url`'s docblock (`src/lib/storage/types.ts`) is widened. It reads *"the user pastes/types details; we never scrape"* — true of the app's own fetching, and it must stay true while a user-supplied import file becomes a third source of records. Widen it in this issue, the one that makes it stale, rather than letting it drift (#610 is the pattern to avoid).
- [ ] The `!persisted` gate on `EVICTION_NOTICE` in `ResumeLibrary.tsx:55` and `JobTracker.tsx:106` is unchanged — only the eviction sentence is conditional, the scope sentence stays unconditional.
- [ ] No raw `<button>`, no hardcoded hex or raw Tailwind palette class in the new code; the file input is visually hidden but focusable/labelled. `npm run lint` clean.
- [ ] `ResumeLibrary.tsx` stays under ~200 LOC, or the dialog is extracted to a sibling.
- [ ] `npm run verify` passes: `typecheck` → `lint` → coverage → build. No change expected in `corpus.test.ts` or `corpus-roundtrip.test.ts` — this touches storage and UI only, not the parse cascade; if either moves, stop and investigate.

## Out of scope

- Encryption at rest. Deliberately not doing this: the data is résumé PII, not credentials, and the only key management that works without an account is a user passphrase with unrecoverable-if-forgotten semantics. Full-disk encryption (FileVault/BitLocker) is the right answer and costs nothing to recommend. No copy in this issue may imply the library is encrypted — it is stored in plaintext and readable by anything running as the user (verifiable with `strings` over Chrome's LevelDB files).
- Cross-device sync, accounts, or any server-side storage. The backup file is the transport, by design.
- Calling `reconcileResumeLinks` after a merge import — that is #547, unblocked by this work.
- Renaming "library" to something with a weaker durability promise.
- Import in `JobTracker` (see Reuse analysis).

Refs #321, #322, #323, #547, #401

---

**Filing note.** #692 was opened against this same capability while filing epic #689, and is closed as a duplicate of this issue. Nothing from it is lost: its two additions (the `version`-mismatch message and the `JobRecord.url` docblock) are folded into the acceptance criteria above. Its one point of disagreement — putting Import on *both* the job library and the résumé library — is **not** adopted; the Reuse analysis above deliberately keeps one entry point for a store-wiping action, and the document is origin-wide anyway, so importing from the résumé library restores tracked jobs too.

### LINKED ISSUE #690: feat(jobs): give the job library a home on /jobs/ and take it off the flag
Part of #689.

## Problem

The job tracker is built, tested, and invisible. `JobTrackerSection` renders at `src/App.tsx:251` behind `useFlag("job-tracker")` (`src/lib/flags.ts:63` → `VITE_ENABLE_JOB_TRACKER`), gated in since #323, and only on `/` — the parser-audit lane — where it sits under the saved-résumé picker and is reachable only in `state.phase === "idle" || "error"`.

That is the wrong home. A user who has come to look at jobs is on `/jobs/`. A user on `/` is auditing a PDF.

## Change

Render the job library on `/jobs/` and retire the flag.

`JobTracker` takes `tracker` as a prop and `JobTrackerSection` owns the `useJobTracker` hook (`src/components/features/JobTracker.tsx:42-45`), so the component is already portable — this is placement, not a rewrite.

The real design call is what `/jobs/` becomes. `src/lib/job-search/CLAUDE.md` records that `FindJobsPanel` *is* the whole page: a full-width query form that folds into a sticky `JobQuerySummary` on Search, with paged results owning the width below. Adding a second view means that statement stops being true, so it changes deliberately and the lane doc changes with it.

Prefer a two-view switch over stacking. Search and library are peer destinations, not steps — so `Tabs` from `@design-system`, not `Stepper`; an ordered flow is a different concern (see #602). Do **not** grow `FindJobsPanel` to host it: it is already the page, and the ~200 LOC guidance means the second view arrives as a sibling.

Removing the flag also means deleting the `job-tracker` entry from `src/lib/flags.ts` and its docblock line, not just defaulting it on.

## Acceptance

- [ ] The job library is reachable from `/jobs/` with no env var set.
- [ ] Search and library are peer views; neither unmounts the other's in-progress state on switch.
- [ ] `job-tracker` is gone from `src/lib/flags.ts`, its doc line included; no dead `useFlag` call remains.
- [ ] `/` no longer renders `JobTrackerSection`; the resume library there is untouched.
- [ ] `src/lib/job-search/CLAUDE.md`'s "`FindJobsPanel` is the whole page" statement is rewritten to match what shipped.
- [ ] New surface is under ~200 LOC and imports only through `@design-system`.
- [ ] A test asserts both views are reachable and that the library renders saved records.

## Watch for

A tab label and a panel heading that repeat the same string make `toContain` assertions pass against either one — assert the distinguishing text and `not.toContain` the sibling (#597, #674).

### LINKED ISSUE #691: feat(job-search): one-click deep links into major employers' own job search
Part of #689.

## Problem

`src/lib/job-search/company-registry.ts` maps a company to a Greenhouse / Lever / Ashby slug. A company that runs its own careers site has no entry and cannot get one — the registry's own docblock says so:

> large self-hosted-careers employers (#542) … run their own careers site rather than a Greenhouse/Lever/Ashby board, so they cannot be added here as-is — there is no API endpoint of the shape this registry assumes. This is a structural boundary of the three-vendor design, not a bug or a curation gap.

**That reasoning is wrong on the facts and should be corrected as part of this issue.** Several of those employers serve perfectly clean JSON, and others serve fully server-rendered search pages. What stops us is not the absence of an endpoint — it is the absence of `Access-Control-Allow-Origin` on the response, which is a browser rule, not a property of their API. The registry is right that we cannot *read* them; it is wrong about why, and the wrong reason has been quietly closing off options that are open.

One of those options is this issue. We cannot fetch those sites. We can absolutely **link into them**, with the user's query already applied.

## Change

Add a second, independent registry of **search-URL templates**: for a given company, the shape of its own careers-search URL and which of our query terms map onto which parameter. Render a one-click "Search Apple directly" style action from `/jobs/` that opens it in a new tab.

Sketch, not a spec:

```ts
interface CompanySearchLink {
  readonly name: string;
  /** Template with named holes, e.g. {query} {location}. */
  readonly template: string;
  /** Which holes this site actually supports — an unsupported hole is
   *  dropped, never emitted empty. */
  readonly supports: readonly ("query" | "location")[];
}
```

Deliberately **separate** from `CompanyEntry` rather than an optional field on it. The two answer different questions — "can we read this board" versus "where does a human search this company" — and a company can have neither, either, or both. Folding them into one type invites a row that half-populates and a call site that forgets to check which half it got.

## Why this is nearly free

**No egress is added.** We build a string. Nothing is fetched. The query terms reach the company only when the *user* clicks and their own browser navigates — with their session, their IP, their user agent, exactly as if they had typed it. That is categorically different from our origin sending a request, and it is worth being precise about rather than waving at: this issue does not touch `providers/keywords.ts`, does not add a `fetch()`, and does not change `buildSearchPlan`'s account of what leaves (`src/lib/job-search/search-plan.ts`).

If a link is surfaced anywhere near the search-plan card, though, it must not *read* as another outbound destination — the card exists to state what we send, and a link we don't send belongs visually outside it or explicitly marked.

## Acceptance

- [ ] A `CompanySearchLink` registry exists with an initial hand-verified set; every template is confirmed to produce a working search against the live site at the time of writing.
- [ ] `/jobs/` surfaces the link for companies in the registry, with the user's current query terms applied.
- [ ] A company that supports no query parameter still links to its careers search — degraded, never broken, never a URL with an empty parameter.
- [ ] Templates are URL-encoded; a query containing `&`, `#`, a space, or a non-ASCII character produces a valid URL. Unit-tested with those inputs specifically, not with a clean one.
- [ ] `rel="noopener noreferrer"` on every outbound link.
- [ ] The `company-registry.ts:55-65` docblock is corrected: the boundary is CORS, not a missing endpoint, and link targets are now a supported shape.
- [ ] `buildSearchPlan`'s output is unchanged — asserted, so a later refactor cannot quietly route a term through this path.

## Watch for

A template registry rots silently — a company changes its parameter name and the link still opens, just ignoring the query. There is no way to test that from here without fetching, which this issue exists to avoid. Prefer templates whose failure is *visible* (a wrong parameter that yields an obviously unfiltered result page) over ones that look plausible, and note in the module docblock that entries are point-in-time verified rather than continuously checked. Do not add a test that pretends to verify them.

### LINKED ISSUE #693: feat(storage): promote JobRecord to a documented, validated capture contract
Part of #689. **Gates the extension work** — the extension writes this contract, so it must be pinned before anything writes it.

## Problem

`JobRecord` (`src/lib/storage/types.ts`) is currently an internal shape: written by `useJobTracker` from a form the user filled in, read by `JobTracker`. Everything that touches it is in this repo and typechecked together, so the type alone is sufficient documentation and no validation is needed at the boundary — there is no boundary.

Import (previous item) and capture (next items) create one. A `JobRecord` will start arriving from outside the typechecker: a JSON file the user picked, and a browser extension shipping a different release cadence than the web app. TypeScript guarantees nothing about either.

## Change

Promote `JobRecord` from an internal type to a **documented, validated capture contract**, so that any tool — ours or someone else's — can produce records this app will accept.

Three parts:

**1. A runtime validator.** One function that takes `unknown` and returns a typed result plus a reason on failure. It must reject: a missing or non-string `title`, an unknown `status` value, a `url` that is not `http`/`https`, and a `matchResult` that is not JSON-safe. It must *tolerate* unknown extra keys rather than rejecting — a newer producer adding a field must not break an older reader.

`status` is the sharp one. `JobStatus` is a closed union of six values and `JobStatusPicker` already assumes unknown values surface rather than get swallowed (`src/components/features/JobStatusPicker.tsx:40` documents exactly this). An extension pinned to an older or newer vocabulary will send something outside the union, and the validator decides whether that record is dropped, coerced to `"interested"`, or preserved. Pick deliberately and write down why.

**2. A written contract document.** Which fields are required, which are optional, what each means, what the id namespace is, and — most importantly — **what a producer must do to avoid creating duplicates**. Two captures of the same posting from the same site must converge on one record, so the contract has to state how an id is derived from a posting URL rather than leaving each producer to invent one. Canonicalisation is part of the contract, not an implementation detail: tracking parameters, locale segments, and trailing slashes all vary per visit for the same job.

**3. Versioning.** `StorageExport` carries `version: 1`, asserted at `backup.ts:70`. Decide now whether captured records carry their own producer version, because retrofitting one after third-party producers exist is not possible.

## Acceptance

- [ ] A validator exists, is unit-tested against malformed input, and is the only path by which an externally-sourced record reaches IndexedDB.
- [ ] Unknown extra keys are preserved or dropped by an explicit, documented decision — not by accident.
- [ ] An out-of-union `status` has a defined, tested behaviour.
- [ ] Id derivation from a posting URL is specified and implemented once, with canonicalisation; capturing the same posting twice yields one record, asserted with URLs differing only by tracking parameters.
- [ ] The contract is written down in the repo — a doc, not only a docblock — because the audience includes people who are not reading this source.
- [ ] The import path from the previous item routes through the validator.

## Watch for

Do not let the validator become a second definition of the type that can drift from `JobRecord` itself. Derive one from the other, or add a test that fails when a field is added to the type and not to the validator. A validator that silently stops covering a new field is worse than none, because it reports success.

s-annam:
## Summary

The web-app half of epic #689. The job tracker has been built and invisible since #323 — flag-gated on `/`, the parser-audit lane. Backup was half a feature: the app could download `offlinecv-backup.json` and had nowhere to put it back, because `importAll`/`importFromJson` shipped with **zero non-test consumers**. And `JobRecord` was an internal type about to start receiving records from outside the typechecker.

This gives the library a home, makes restore reachable, and turns the record shape into a contract a producer outside this build can write against.

Refs #689 — the epic stays open. #694 and #695 (the MV3 capture extension) are deliberately deferred to a later batch; #693 pins the contract they will write, which is why it lands here first.

Closes #690
Closes #691
Closes #573
Closes #547
Closes #693

## What shipped

- **#690 — the library gets a home.** Moved from `/` to `/jobs/` as a peer view of search behind the `Tabs` primitive, and the `job-tracker` flag is deleted (entry + docblock line, no dead `useFlag` left). Both tab panels stay **mounted** — `Tabs`/`TabPanel` toggle `hidden` rather than conditional-rendering — so switching views never discards an in-flight search or the library's state. `src/jobs/JobsApp.test.tsx` asserts the library panel is the *same DOM node reference* across a switch, not merely that equivalent text reappears. `src/lib/job-search/CLAUDE.md`'s "`FindJobsPanel` is the whole page" statement and the matching line in the root `CLAUDE.md` are rewritten to match what shipped.
- **#691 — one-click company search links.** A `CompanySearchLink` registry, deliberately separate from `CompanyEntry`, plus links from `/jobs/` with the user's query applied. **No `fetch` is added** — this builds a string; the terms reach the employer only when the user's own browser navigates. `buildSearchPlan`'s output is asserted byte-identical so a later refactor cannot quietly route a term through this path. URL-encoding is tested with `&`, `#`, space, and non-ASCII specifically. `rel="noopener noreferrer"` comes free from reusing `ExternalBoardLinks`. `company-registry.ts`'s docblock is corrected: the boundary is **CORS**, not a missing endpoint — the old wording had been closing off options that are in fact open.
- **#573 — the restore path.** `useResumeLibrary().importBackup(file, mode)` plus a merge/replace `Dialog` in `ResumeLibrary`. The mode is passed **explicitly at every call site**: `importAll` defaults to `"replace"` while the UI defaults to `"merge"`, so an omitted argument would silently invert the user's intent into a destructive one. The empty-library early return is relaxed — a fresh browser, a new profile, and a post-eviction visit are all empty-library states, i.e. exactly when someone needs to import and exactly when the surface holding the control used to render nothing.
- **#547 — no dangling resume links.** `reconcileResumeLinks` now runs after a merge-mode import, reading survivor ids via `listLibrary()` **after** the write rather than from the pre-import snapshot. Replace mode is deliberately excluded: an `exportAll` document is internally self-consistent, so no job in it can reference a resume the same snapshot lacks. The `// staged for #547` marker is gone now that a real caller exists.
- **#693 — `JobRecord` as a capture contract.** A runtime validator, URL canonicalisation + id derivation (`job-url.ts`), and `docs/job-capture-contract.md` — a doc, not just a docblock, because the audience includes people not reading this source. `importAll` validates **up front, before any store is touched**, then skips per-record rather than rejecting the document: one malformed job must not cost the user the other forty, and in replace mode the stores are already empty by write time, so a mid-write throw would leave them with less than they started with. Skips ride back on `ImportCounts` and are announced in `ResumeLibrary`'s `aria-live` region, because a silently-dropped record is the real failure mode.

## Review focus

- `src/lib/storage/job-url.ts` — `src`, `source`, `ref`, `refid` and `trk` are stripped as tracking parameters, so two postings differing **only** by one of those canonicalise to the same id (`?source=100` and `?source=200` both derive `job:jobs.example.com/listing`). The module's own asymmetry rule says over-merging destroys a record while under-merging is a deletable duplicate — is stripping the generic names worth that risk, or should `src`/`source`/`ref` move off the list? The current behaviour is pinned by a test so a change is visible, and `gh_jid`/`jk`/`vjk`/`currentJobId`/`id` are deliberately kept.
- `src/jd-fit/JdFitApp.tsx` — deleting the `job-tracker` flag also un-gated **"Save this job" on `/jd-fit/`**, a second flag consumer #690's text never mentions. Removing the flag from `FLAG_DEFAULTS` makes that `useFlag("job-tracker")` a type error, so un-gating was required to compile — but it is a user-visible change outside the issue's stated surface. Is always-visible the behaviour you want there?
- `src/lib/storage/job-record-contract.ts` — an out-of-union `status` is **preserved verbatim** with a warning, not coerced to `"interested"` and not dropped, on the reasoning that `JobStatusPicker` and `JobTracker`'s bucketing both document that they surface such a value rather than hide it. Does that hold for a value that is a string but wildly wrong?
- `src/lib/storage/capture.ts` — has no production caller; it is the reference implementation #694 will wire, and carries an explicit `staged for #694` marker plus its own tests. Acceptable to land ahead of its consumer, or should it wait for the extension batch?

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — 4002 passed, 10 skipped
- [x] `npm run check:fixtures` clean — no fixture binaries added by this PR
- [x] `fallow audit --base origin/main` run over the whole accumulation (report-only); findings triaged, see below
- [x] `fallow` code-scanning check green — the 22 new alerts the first push raised (20 dead-export errors + 2 cognitive-complexity notes) are resolved, see below
- [x] Company search link verified against Apple's LIVE careers search with a real résumé's query — see "Third round" below
- [x] Manually verified in `npm run dev` — `/jobs/` tab switch, an import of a real backup, and a company link

## Adversarial review

Reviewed pre-PR against the five issues' acceptance criteria, with the defects **reproduced by execution** rather than reasoned about.

**Attacked and could not break.** A backup this app produced *before* this change still restores intact: a `url` the user typed without a scheme (`acme.com/jobs/1`), an empty `url`/`title`/`company`, a missing `company`, and an out-of-union `status` are all accepted. The `matchResult` rejections for `Date`/`undefined`/`NaN` are real but unreachable through a backup — `exportToJson` runs `JSON.stringify`, which already normalises all three — so they bite only on the capture path, which is their point. `resumeId: ""` is rejected but unproducible: every writer passes a real `resume.id` or `undefined`. Greenhouse postings differing by `gh_jid` correctly do **not** collide.

**Drift guard verified, not assumed.** Adding a probe field to `JobRecord` and running `npm run typecheck` fails with `TS2741` at `job-record-contract.ts:149` — the mapped type over `keyof Required<JobRecord>` is a real compile-time guard, so a field added without a validation rule cannot ship. Probe reverted.

**Fixed during review.** `capture.ts` was staged with no caller, no marker, and no test — the shape #547 had just cleaned up for `reconcileResumeLinks` — so it gained an explicit `staged for #694` marker and 13 tests, including one proving a re-capture does **not** clobber a status the user moved to `"applied"`. The `job-url.ts` over-merge risk was documented in both the module and the contract doc and pinned by a test.

**Surviving nit.** Three test harnesses (`JobTracker.test.tsx`, `useResumeLibrary.test.tsx`, `JobsApp.test.tsx`) independently rebuild the same fake-indexeddb seeding, ~24 lines × 3 — a genuine cross-issue duplicate, invisible to each issue alone. Left as-is: `fallow` is report-only here, and a shared helper contorted to fit three differently-shaped harnesses would be worse than the duplication. Also unfixed and out of scope: the pre-existing `useJobTracker.ts` / `useResumeLibrary.ts` overlap and the complexity/CRAP findings on `App.tsx`, `JdFitApp.tsx`, `FindJobsPanel.tsx` — all of which sit on lines this PR did not add, so the code-scanning check does not attribute them here.

**Second round — the `fallow` code-scanning check.** The first push went green on `verify` but red on `fallow`: 22 new alerts, 20 of them errors. Every error was the same class — #693's contract surface re-exported from `src/lib/storage/index.ts` with no importer, because the only in-tree consumer (`capture.ts`) deep-imports its siblings. That is the case `.fallowrc.jsonc` already documents for the `heuristics` and `jd-match` barrels, so the surface is registered there the same way: **enumerated by name, never `["*"]`**, so a future accidentally-dead storage export still trips the check. `captureJob`/`JobCaptureResult` are called out in that entry as the honest exception — they have no consumer *anywhere* until #694, and the comment says to remove them from the list when it lands, or delete `capture.ts` if it is abandoned. The 2 notes were **not** suppressed: `findJsonSafetyProblem` (cognitive 23) and `validateJobRecord` (22) were split along the seams that made them complex — the first into its terminal scalar dispatch and its recursive container walk, the second into the rule loop and the accepted-with-warnings pass. Behaviour is unchanged and the suite holds at 4002 passing; fallow's whole-tree counts drop from 34 dead-code issues to 14 and 6 complexity findings to 4.

## Third round — the company search link returned zero postings

Reported from real use: the Apple link built from a real résumé's query opened a
page with no results. Measured against the live site rather than reasoned about,
which is what found the actual cause — it was **not** the one that looked obvious.

| `search=` value | Result |
|---|---|
| the full built phrase (4 titles + 6 skills) | **0 postings** |
| the 4 titles alone, no skills | **0 postings** |
| `Sr. Engineering Manager Founder India Site Lead` | 600+ |
| `Sr. Engineering Manager People Management` | 600+ |
| `Engineering Manager - Customer Experience` | **0 postings** |
| `Engineering Manager` | 600+ |

Term-stuffing was not the cause — three stacked titles still return a full page.
The zero comes from the title `Engineering Lead - Customer Experience`: the bare
` - ` is read as a **NOT operator** and empties the query. It fails silently, as
an employer with no matching roles.

Two fixes, at different layers:

- **`stripSearchOperators`** (`deep-links.ts`) strips a token-leading `-` from
  every outbound link phrase, so LinkedIn/Indeed/Google Jobs get it too. An
  intra-word hyphen is left alone — `full-stack` and `e-commerce` are spelling,
  not operators. Other punctuation (`&`, `/`, `+`, `#`) is deliberately untouched:
  `+` and `#` are load-bearing in `C++`/`C#`, and none of them reproduced the
  empty result when measured.
- **The company links no longer reuse `buildKeywords`.** That was the wrong seam.
  `buildKeywords` composes #539's *broadening* union for a board's OR-weighted
  keyword field; an employer's own careers box is a single-intent full-text
  query — the shape `providers/keywords.ts` already refuses to stack titles into,
  for the same stated reason. They now compose `searchPhrase` (the primary
  title), so the audited egress *narrows*. `buildKeywords` is module-private
  again, since re-exporting it would re-offer the seam that was wrong.

**`roleHeadForSearch`** additionally drops a trailing scope qualifier from that
term, so `Engineering Lead - Customer Experience` searches as `Engineering Lead`.
It reuses `splitHeadline`'s separator set and `looksLikeTitle` re-gate rather
than adding a second splitter, which keeps #605's lessons in force — `VP,
Engineering` and `React/Node Engineer` stay whole, and an inverted `Customer
Experience - Engineering Lead` resolves to the role, not the qualifier. The
narrowing is applied by the **egress caller only**: `JobQuery.titles` keeps the
user's own words, so the chips, the local `matchesQuery` broadening and the
exported PDF are untouched (#605's blast radius).

Verified after the fix: `search=Sr. Engineering Manager` and
`search=Engineering Lead` both return 600+. 14 new tests, each confirmed to fail
against the pre-fix code. Suite 4002 → 4016.

**Not changed, and worth a reviewer's eye:** LinkedIn/Indeed/Google Jobs still
send the all-titles union. #539 chose that for an OR-weighted keyword field, and
the measurement above is against a *different kind* of destination — so it is
evidence about Apple, not about them. They do get the operator strip. Is the
union still right for those three?

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #691 | Claude Sonnet 5 | high |
| Implementation — #690 | Claude Sonnet 4.6 | high |
| Implementation — #573 | Claude Sonnet 5 | high |
| Implementation — #547 | Claude Sonnet 5 | medium |
| Implementation — #693 | unreported (requested: opus) | ultra |
| Adversarial review | Claude Opus 5 (1M context) | high |
| Review fixes | unreported (requested: opus) | high |
| Review revisions — `fallow` alerts | Claude Opus 5 (1M context) | high |
| Field fix — company search link | Claude Opus 5 (1M context) | high |

Two rows read `unreported`: those subagents completed their work but went idle without reporting a model name. A requested `model:` alias does not establish what it resolved to, so the rows state what is known rather than a guess.


s-annam (commented):
An audit of the implementation has been completed against the acceptance criteria of the five linked issues (#690, #691, #573, #547, #693) on commit `e4d59cdebb6fea47df253176da3e731bd76006fe`.

All verification checks (`npx tsc`, linter, fixtures check, baseline checks, Vitest tests, and Vite build) are clean.

### 📋 Issue Acceptance Criteria Checklist
* **[#690](https://github.com/offlinecv/OfflineCV/issues/690) (Job library home on `/jobs/`):** **MET**
  - Reachable with no env var (off the flag).
  - Search and library are peer views (both mounted, toggle `hidden`).
  - Flag is retired from `flags.ts`.
  - `/` no longer renders `JobTrackerSection`.
  - CLAUDE.md documentation updated.
* **[#691](https://github.com/offlinecv/OfflineCV/issues/691) (One-click company search links):** **MET**
  - `CompanySearchLink` registry exists separate from `CompanyEntry`.
  - Outbound link parameters are URL-encoded; uses `searchPhrase` (single-intent) instead of `buildKeywords` (union) for company links.
  - Corrected `company-registry.ts` docblock regarding CORS vs. missing endpoint.
  - Checked `buildSearchPlan`'s output to verify it is unchanged.
* **[#573](https://github.com/offlinecv/OfflineCV/issues/573) (Restore path):** **MET**
  - UI import backup trigger wired to `ResumeLibrary` (relaxed empty-library check).
  - Merge/replace mode selection confirm dialog (sibling component `ResumeLibraryImportDialog.tsx`).
  - File input is reset so re-picking the same file fires again.
* **[#547](https://github.com/offlinecv/OfflineCV/issues/547) (Reconcile resume links on import):** **MET**
  - Dangling resume links are swept/reconciled on merge-mode import using `reconcileResumeLinks()`.
* **[#693](https://github.com/offlinecv/OfflineCV/issues/693) (JobRecord capture contract):** **MET**
  - Runtime validator checks properties, protects against prototype injection, and verifies JSON-safety using `findJsonSafetyProblem()`.
  - Out-of-union `status` is preserved verbatim with a warning rather than coerced.
  - Written contract added at `docs/job-capture-contract.md`.
  - Backup import validates up front before stores are touched.

---

### 🔎 Description Accuracy
* **Accurate.** The PR description accurately describes the changes and has a detailed adversarial review, verification, and rationale.

---

### 📝 Findings

#### Secondary: Unused Storage Exports
* [src/lib/storage/index.ts](file:///C:/Users/annam/offlinecv/src/lib/storage/index.ts)
* Staged unused exports for the browser extension (to be implemented in #694) are correctly named in `.fallowrc.jsonc` to avoid alert noise while catching regressions.

#### Nit: Duplicate IndexedDB Seeding in Test Files
* `JobTracker.test.tsx`, `useResumeLibrary.test.tsx`, and `JobsApp.test.tsx` duplicate similar IndexedDB seeding setup (~24 lines each). This is a reasonable trade-off to avoid over-complicating the test harnesses.

---
Reviewed by: Gemini 3.5 Flash (High) (high)

Issue #700, opened 2026-07-31
refactor(jd-match,job-search): make the fitness chain callable outside rankJobs, and rate the saved job library


s-annam:
## Problem

The fitness rating — the number that answers "is this job good for me" — can only be computed inside `rankJobs`. Two consumers already want it and cannot have it:

1. **The saved job library.** `/jobs/` → Saved jobs shows no rating at all. Verified: `rateJobs` is called from exactly one place (`src/lib/job-search/rank.ts:241`), `JobRecord` has no rating field, and `JobTracker.tsx` / `JobTrackerEntry.tsx` render no rating. A user who saves a job loses the one number that made them save it.
2. **Any out-of-tree consumer** that holds a single posting rather than a search result set.

Neither is blocked by the algorithm — the whole chain is already pure. It is blocked by two seams:

- **`computeCoverage(parsed: HeuristicParsedResume, terms)`** (`src/lib/jd-match/coverage.ts:60`) demands a full parsed résumé, but the only thing it does with it is `buildCorpus(parsed)` (line 134) → one lowercased string. Every caller must therefore hold and pass the entire résumé object even though a ~single string is the real input.
- **`ratingInputFor(...)`** is module-private in `rank.ts`, so assembling a `RatingInput` outside that file means duplicating the mapping — the fastest possible route to two ratings that disagree.

## Why the chain is worth extracting

Verified across all four modules — `extract-jd-terms.ts`, `coverage.ts`, `skills.ts`, `rating.ts`:

- **Zero** non-relative imports (no external dependencies).
- **Zero** references to `document.` / `window.` / `navigator.` / `indexedDB`.
- ~60 KB of source total.

It is already portable. Nothing needs rewriting; two boundaries need to move.

## Change

**1. Split `computeCoverage` on the corpus boundary.** Keep the existing signature as a thin wrapper so no caller changes:

```ts
// coverage.ts
export function computeCoverageFromCorpus(
  corpus: string,                       // already lowercased, from buildCorpus
  terms: readonly ExtractedTerm[],
): CoverageResult { /* the current body, minus the buildCorpus call */ }

export function computeCoverage(
  parsed: HeuristicParsedResume,
  terms: readonly ExtractedTerm[],
): CoverageResult {
  return computeCoverageFromCorpus(buildCorpus(parsed), terms);
}
```

[...]

PR #698, opened 2026-07-31
chore(storage): make the barrel the real import boundary and clear fallow's 14 dead exports

s-annam:
## Summary

`src/lib/storage/index.ts` documents itself as the module's import boundary, but every in-repo consumer reaches past it — `job-tracker.ts` pulls jobs CRUD from `./storage/jobs.ts`, and five files pull `JobRecord`/`JobStatus`/`JOB_STATUS_ORDER` from `./storage/types.ts`. The barrel re-exported those names anyway, so `fallow dead-code` read 14 of them (7 exports, 7 types) as unconsumed — the repo's entire remaining dead-code surface after #696.

The barrel was over-broad *and* under-broad, which is one defect rather than two: it published nine names the module only talks to itself in, while withholding the two (`JobStatus`, `JOB_STATUS_ORDER`) that consumers had no other route to. Withholding is what made the stated contract unfollowable — the deep import of `./types.ts` was the only thing a consumer could write.

This routes the six job-lane consumers through the barrel, adds the two withheld names, drops the nine internal-only re-exports, and writes both admission rules into the docblock so the next name added has a stated test to meet. `fallow dead-code` goes 14 → 0 repo-wide.

No issue — this came out of a `fallow` sweep, and no open issue tracked it.

## Review focus

- `src/lib/storage/index.ts` — dropping the `exportAll`/`exportToJson`/`importAll` re-exports: does anything outside `src/lib/storage/` still reach them, including a branch in flight? (`storage.test.ts` already deep-imports them from `./backup.ts`, so coverage is unaffected, but a second consumer would break the build rather than degrade quietly.)
- `src/components/features/JobStatusPicker.tsx:13` — this file previously pulled only `types.ts`. Does routing it through the barrel drag the `idb` graph into a chunk that did not already carry it?
- `src/lib/storage/index.ts` docblock — `board-cache.ts` is recorded as a standing exception rather than fixed. Is that the right call, or should the barrel expose the generic `getRecord`/`putRecord` accessors it needs?

## Test plan

- [x] `npm run typecheck` clean
- [x] `npm run verify` green — 4016 passing, 10 skipped
- [x] `npx fallow dead-code` — 14 findings → 0 repo-wide
- [x] Chunk sizes compared against `main`: every emitted chunk byte-identical (checked because the `JobStatusPicker` reroute could have changed chunking)

## Provenance

Code implementation via: Claude Opus 5 (1M context)
Verification: `npm run verify` — green locally, and in CI

PR #703, opened 2026-07-31
fix(job-search,jd-match): read "and"-separated pay ranges, and split coverage on the corpus boundary (#699, #700)

### LINKED ISSUE #699: fix(job-search): extractCompensation drops the top of an "and"-separated pay range
## Problem

`extractCompensation` treats a range written with the word **"and"** as a single figure, silently collapsing the top of the range to the bottom. "Between $X and $Y" is the standard phrasing of US pay-transparency boilerplate (CA/NY/CO/WA), so this is not an edge case — it is hitting live feed postings today.

Measured against the exact sentence from a real Apple posting, and against each separator in isolation:

| Input | `min` | `max` | Correct? |
|---|---|---|---|
| `$183,700 - $319,800` | 183700 | 319800 | ✅ |
| `$183,700 – $319,800` (en dash) | 183700 | 319800 | ✅ |
| `$183,700 to $319,800` | 183700 | 319800 | ✅ |
| `The base pay range for this role is between $183,700 and $319,800, and your base pay will depend on…` | 183700 | **183700** | ❌ top of range dropped |
| `between $183,700 and $319,800` (bare, no salary context word) | — | — | ❌ `undefined` |
| `between $183k and $319k` | — | — | ❌ `undefined` |

## Root cause

`src/lib/job-search/compensation.ts:105`:

```ts
/** Range separators: ASCII hyphen, en dash, em dash, "~", or the word "to" … */
const SEP = String.raw`-|–|—|~|\bto\b`;
```

`\band\b` is absent, so `RANGE_RE` (built from `SEP` at line 133) never matches an "and"-separated pair. Extraction then falls through to `SINGLE_RE` (line 146), which needs a `SALARY_CONTEXT_RE` word (line 172) within range of a lone figure. That explains all three failures precisely:

- **In a sentence** — "pay"/"base"/"compensation" are context words, so `SINGLE_RE` matches the *first* figure and returns it as both `min` and `max`. Worst outcome: a plausible-looking result that is wrong.
- **Bare `between $X and $Y`** — no context word anywhere, so `SINGLE_RE` also fails and the result is `undefined`.
- **k-suffixed** — same path as bare; `numWithK` never gets a chance because `RANGE_RE` did not match.

## Impact

Not cosmetic. The value flows straight into the rating:

`extractCompensation` → `annualizedTop(comp)` → `RatingInput.compMax` → `compensationFraction` → the **0.35-weighted** compensation axis in `rateJobs`.

So every "and"-phrased posting is rated as though it paid the **floor** of its advertised range. For the Apple example that is $183,700 scored instead of $319,800 — and comp is the second-heaviest axis after fitness. `isBelowFloor` reads the same value, so a posting whose top of range clears the user's `compFloor` can be badged "below your floor" when it is not.

## Change

Add `\band\b` to `SEP` and update the docblock above it to match.

```ts
const SEP = String.raw`-|–|—|~|\bto\b|\band\b`;
```

**The risk this carries, and what must be tested.** "and" is a far more common word than "to", so a currency-`and`-currency pair is much more likely to be two unrelated figures than a range — "we raised $10,000,000 and $5,000,000 in follow-on", "a $500 stipend and $2,000 relocation". `RANGE_RE` is currency-anchored on both sides, which filters some of it, but not that. Before landing, check whether the existing sentence-boundary and proximity guards (`SENTENCE_BOUNDARY_RE`, line 191, and the max-gap constant near line 181) are sufficient, and add a guard if they are not. A false range is worse than a missed one: a missed range drops the axis and `rateJobs` renormalizes the remaining weights, whereas a false range actively mis-scores.

## Acceptance

- [ ] Every row in the table above extracts the correct `min` **and** `max`.
- [ ] The `SEP` docblock at `compensation.ts:103–104` lists "and" alongside the other separators.
- [ ] A non-range pair of currency figures in one sentence (e.g. `a $500 stipend and $2,000 relocation`) does **not** produce a range — asserted by a new test, whatever guard makes it pass.
- [ ] Two figures separated by a sentence boundary are not joined into a range.
- [ ] The multi-location case is covered: a posting body containing two distinct "between $A and $B" sentences yields the first accepted range, matching the documented `firstAcceptedRange` behaviour at line 342.
- [ ] `src/lib/job-search/compensation.test.ts` and `rating.test.ts` stay green; `rank.test.ts` output is unchanged for postings that contain no "and"-separated range.
- [ ] `npm run verify` green.

## Notes

Found while measuring a real Apple posting for the extension work. Independent of it — this is a defect in the search lane as it ships today, and it needs no extension, no capture, and no new surface to fix or to benefit from.

s-annam:
Two small, related changes in the search/match lanes.

Closes #699
Refs #700 (partial — the corpus seam only; the saved-library rating is **not** in this PR, see below)

---

## 1 — #699: `extractCompensation` dropped the top of an "and"-separated range

`SEP` (`compensation.ts`) listed `-`, `–`, `—`, `~` and `to`, but not `and`. So `RANGE_RE` never matched an "and"-separated pair, extraction fell through to `SINGLE_RE`, and that returns the **first** figure as both `min` and `max`.

"between $X and $Y" is the standard phrasing of US pay-transparency boilerplate (CA/NY/CO/WA), so this hits live feed postings. The value flows `extractCompensation` → `annualizedTop` → `RatingInput.compMax` → the **0.35-weighted** comp axis in `rateJobs`, and `isBelowFloor` reads the same figure — so an affected posting was rated at the **floor** of its own advertised range, and one whose top of range cleared the user's `compFloor` could still be badged "below your floor".

Every row of the issue's table is now covered by a test:

| Input | before | after |
|---|---|---|
| `The base pay range for this role is between $183,700 and $319,800, and your base pay will depend on…` | 183700 / **183700** | 183700 / 319800 |
| `between $183,700 and $319,800` (bare, no salary-context word) | `undefined` | 183700 / 319800 |
| `between $183k and $319k` | `undefined` | 183000 / 319000 |

The three separators that already worked (`-`, `–`, `to`) are unchanged.

### The guard is the actual work

Adding `\band\b` is one token. But "and" is an ordinary conjunction, so a currency-`and`-currency pair is much likelier to be two unrelated figures than a range. **A false range is worse than a missed one**: a miss drops the comp axis and `rateJobs` renormalizes the remaining weights, whereas a false range actively mis-scores.

Two facts narrowed the problem before picking a guard:

- `RANGE_RE`'s separator is **whitespace-tight** — `\s*(?:${SEP})\s*` — so `a $500 stipend and $2,000 relocation` can never match anyway (the word "stipend" sits between). The real exposure is **adjacent bare figures**: `a $500 and $2,000 relocation`, `we raised $10,000,000 and $5,000,000 in follow-on`.
- `candidateFrom` normalizes with `Math.min`/`Math.max`, so there was **no ascending-order check** anywhere — a descending pair silently became a range.

I measured candidate guards against a labelled corpus (8 true "and" ranges, 11 non-ranges):

| guard | missed a true range | produced a false range |
|---|---|---|
| none (just add `and`) | 0/8 | **11/11** |
| ascending only | 0/8 | 8/11 |
| salary-context word attached | 2/8 | 2/11 |
| intro word only | 0/8 | 1/11 |
| **intro word + ascending** | **0/8** | **0/11** |

So `isTrustedRange` gates **"and" matches only** — punctuation separators and "to" pass unconditionally — on two clauses:

1. **An intro word** (`between`, `from`, `range of`) immediately precedes the pair. In English "and" joins two endpoints into a *range* essentially only inside the "between … and …" construction.
2. **The pair ascends.** The other common "between … and …" construction is an *enumeration* — "choose between $15,000 and $7,500 depending on distance" — and descending order is the tell.

This is deliberately scoped to `and`, so `candidateFrom`'s `Math.min`/`Math.max` normalization and every other separator behave exactly as before. A rejected "and" match is **skipped, not aborted on**, so a later genuine range in the same description still wins (tested).

### Revert-proof

Each piece is load-bearing — verified by reverting one at a time against the new tests:

| reverted | tests that go red |
|---|---|
| `\band\b` removed from `SEP` (i.e. the fix itself) | **7** |
| intro-word clause removed from `isTrustedRange` | **2** |
| ascending clause removed from `isTrustedRange` | **1** |

## 2 — #700, the corpus seam only

`computeCoverage(parsed: HeuristicParsedResume, terms)` demanded a whole parsed résumé, but the only thing it did with it was `buildCorpus(parsed)` → one lowercased string. A consumer holding only that digest could not call it at all, and would have to fork a second coverage implementation — the fastest route to two scores that disagree.

Split on that boundary, keeping the existing signature as a thin wrapper so **zero callers change**:

```ts
export function computeCoverageFromCorpus(corpus: string, terms: readonly ExtractedTerm[]): CoverageResult
export function computeCoverage(parsed, terms) { return computeCoverageFromCorpus(buildCorpus(parsed), terms); }
```

Exported from the `src/lib/jd-match/index.ts` barrel alongside the existing `computeCoverage` / `buildCorpus`, and registered in `.fallowrc.jsonc`'s enumerated public surface for that barrel (see **fallow** below).

### ⚠️ #700's problem statement is half stale — verified

> **`ratingInputFor` is module-private in `rank.ts`**

**No longer true.** It is already exported at `src/lib/job-search/rank.ts:168`, and `src/lib/job-search/probe-jobs.test.ts:80` already imports it. `rg ratingInputFor` returns exactly one definition and no second `RatingInput` assembly site. **Nothing to do for that seam** — #700's step 2 is already satisfied.

`computeCoverageFromCorpus` genuinely did not exist; that was the real work, and it is what this PR does.

### Out of scope, deliberately

**#700 step 3 — rating the saved job library** (`JobTrackerEntry` / `StarRating` wiring) is **not** in this PR. It carries its own UI acceptance criteria (the "not rated" vs zero-stars distinction, one `rateJobs` call for the whole set so the set-relative stretch matches the search lane, no rating persisted to IndexedDB) and belongs in its own PR. This PR is the pure seam, so #700 stays open with exactly that remaining.

## Gates

- `npm run verify` green — 272 test files passed, 8 skipped; typecheck, lint, `check:fixtures`, `check:baselines`, coverage and build all clean. Exit 0.
- `rank.test.ts`, `rating.test.ts`, `probe-jobs.test.ts` and `search-plan.test.ts` pass **with no edits to their expectations** — change 2 is a pure refactor at the search lane, and change 1 moves no posting that contains no "and"-separated range.
- No fixtures added, so no PII surface. No component, token, or design-system change.

### fallow

The first push tripped fallow's dead-code check with one alert — **unused export `computeCoverageFromCorpus` (re-export)** at `src/lib/jd-match/index.ts:13`. That is now **fixed**, not waved through: the name is registered in `.fallowrc.jsonc`'s enumerated jd-match public surface, which is the convention this repo already uses for the heuristics, jd-match and storage barrels (`ignoreExports`, populated by #698 for exactly this class of finding). `fallow audit --base origin/main` is now **dead code 0**, exit 0.

Worth being precise about *why* it belongs there, because it is **not** the `captureJob` case documented lower in that file:

- The **implementation** has an in-tree consumer — `computeCoverage` is a thin wrapper over it and `coverage.test.ts` exercises both — so the function can never quietly go dead. fallow never flagged the source export in `coverage.ts`, only the barrel line.
- What has no in-repo consumer is the **barrel re-export**, which is the entire point of a seam: it exists to be callable by someone holding just the lowercased digest. That is the "intended external surface, not dead code" rationale the surrounding block already states.

Registered by name rather than `["*"]`, per the standing comment in that file, so a future accidentally-dead re-export from this barrel still trips the check.

One report-only finding remains and is **not** mine: **duplication, `compensation.test.ts:15-21` vs `38-43`** — pre-existing assertion blocks in the two oldest tests, surfaced only because the file is now in the changed set. `git diff origin/main` on that file shows **zero deleted lines**, so those blocks are untouched. fallow classifies it as an inherited finding and the gate excludes it.

## Provenance

Implemented with Claude Code (Opus 5). The guard was selected by measurement, not by inspection: candidate guards were scored against a hand-labelled corpus of "and"-separated shapes (the table above) before one was picked, and each surviving clause was then revert-proofed individually against the new tests. The stale `ratingInputFor` claim in #700 was verified by reading `rank.ts` and `rg`, not taken from the issue text.

### DIFF
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -99,10 +99,22 @@
       "TierEngagedEvent", "ParseCompletedEvent", "TierId", "TierEngagementReason",
       "FinalSource", "FileSizeBucket", "ParseMetadata"
     ] },
+    //
+    // `computeCoverageFromCorpus` (#700) is the newest name on the jd-match
+    // barrel and the reason it is here is worth stating precisely, because it is
+    // NOT the `captureJob` case below. Its IMPLEMENTATION has an in-tree
+    // consumer — `computeCoverage` is a thin wrapper over it, and coverage.test.ts
+    // exercises both — so the function itself can never quietly go dead. What has
+    // no in-repo consumer is only the barrel RE-EXPORT, which exists because the
+    // corpus seam's whole point is to be callable by someone holding just the
+    // lowercased digest rather than a whole `HeuristicParsedResume`. That is the
+    // intended external surface this block already describes, not a suppression
+    // standing in for a missing caller.
     { "file": "src/lib/jd-match/index.ts", "exports": [
       "extractJdTerms", "stripBoilerplate",
       "ExtractedTerm", "ExtractJdTermsResult", "ExtractOptions",
-      "computeCoverage", "buildCorpus", "SKILL_WEIGHT", "NOUN_WEIGHT",
+      "computeCoverage", "computeCoverageFromCorpus",
+      "buildCorpus", "SKILL_WEIGHT", "NOUN_WEIGHT",
       "CoverageResult", "SKILLS", "getSkillIndex", "skillCount", "SkillEntry",
       "fetchJdFromUrl", "parseAtsUrl", "htmlToPlaintext", "AtsPlatform",
       "JdMatchResult"
--- a/src/lib/jd-match/coverage.test.ts
+++ b/src/lib/jd-match/coverage.test.ts
@@ -6,6 +6,7 @@ import type { HeuristicParsedResume } from "../heuristics/types.ts";
 import { extractJdTerms } from "./extract-jd-terms.ts";
 import {
   computeCoverage,
+  computeCoverageFromCorpus,
   buildCorpus,
   buildResumeProjection,
   SKILL_WEIGHT,
@@ -141,5 +142,37 @@ Familiarity with Distributed Systems is a plus.
     expect(cov.score).toBe(0);
     expect(cov.covered).toHaveLength(0);
     expect(cov.missing).toHaveLength(0);
+  });
+});
+
+// The corpus seam (#700). `computeCoverage` is now a thin wrapper, so the two
+// entry points must stay one implementation — a caller holding only the
+// lowercased digest must score identically to one holding the whole résumé.
+describe("computeCoverageFromCorpus", () => {
+  const parsed = makeParsed({
+    summary: "Backend engineer who ships.",
+    skills: ["TypeScript", "Kubernetes"],
+    experience: [
+      { title: "Staff Engineer", company: "Acme", description: "Owned the Kafka pipeline." },
+    ],
+  });
+  const terms = extractJdTerms(
+    "We need a TypeScript engineer with Kubernetes experience and a Kafka pipeline background. Terraform is a plus.",
+  ).all;
+
+  it("is what computeCoverage delegates to — identical result for the same résumé", () => {
+    // Guard: an empty term list, or one nothing covers, would make this vacuous.
+    expect(terms.length).toBeGreaterThan(0);
+    const viaCorpus = computeCoverageFromCorpus(buildCorpus(parsed), terms);
+    expect(viaCorpus.covered.length).toBeGreaterThan(0);
+    expect(viaCorpus).toEqual(computeCoverage(parsed, terms));
+  });
+
+  it("scores from a corpus string alone, with no HeuristicParsedResume in hand", () => {
+    const cov = computeCoverageFromCorpus("typescript, kubernetes, kafka", terms);
+    expect(cov.covered.map((t) => t.id)).toEqual(
+      expect.arrayContaining(["typescript", "kubernetes"]),
+    );
+    expect(cov.score).toBeGreaterThan(0);
   });
 });
--- a/src/lib/jd-match/coverage.ts
+++ b/src/lib/jd-match/coverage.ts
@@ -51,17 +51,24 @@ export interface CoverageResult {
 }
 
 /**
- * Run the coverage check.
+ * Run the coverage check against an already-built corpus.
  *
- * `parsed` is the cascade's HeuristicParsedResume — `skills: string[]`,
- * `experience[].description`, `summary?`, `education[]`. We tolerate any
- * field being missing.
+ * The corpus — one lowercased string, as `buildCorpus` produces — is the only
+ * thing coverage matching ever reads. Splitting it out from `computeCoverage`
+ * (#700) makes "the résumé, reduced to what matching actually reads" a value a
+ * caller can hold, cache, or hand across a boundary WITHOUT holding a whole
+ * `HeuristicParsedResume`. A consumer that has only the digest can now score
+ * against it instead of forking a second coverage implementation, which is the
+ * fastest route to two scores that disagree.
+ *
+ * `corpus` MUST already be lowercased: the skill mention patterns and the
+ * phrase check below are matched case-insensitively against it on the
+ * assumption `buildCorpus` did that normalization.
  */
-export function computeCoverage(
-  parsed: HeuristicParsedResume,
+export function computeCoverageFromCorpus(
+  corpus: string,
   terms: readonly ExtractedTerm[],
 ): CoverageResult {
-  const corpus = buildCorpus(parsed);
   const covered: ExtractedTerm[] = [];
   const missing: ExtractedTerm[] = [];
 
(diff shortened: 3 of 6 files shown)


s-annam (commented):
Validation and review summary of the changes in PR #703:

- The pay-transparency range extraction for "and"-separated boilerplate has been implemented with a guard in place to avoid false ranges. Tests in `compensation.test.ts` verify the various syntax forms and guard behavior.
- The corpus matching seam has been refactored in `src/lib/jd-match/coverage.ts`, allowing coverage scoring against an already-built corpus.

### Verification Summary
- **Typecheck & Lint:** Clean (`tsc` and `eslint` passed with exit code 0).
- **Fixture PII & Baselines:** Clean (`check:fixtures` and `check:baselines` passed with exit code 0).
- **Vitest Suite:** Passed cleanly (4,030 tests passed, 0 failures).
- **Production Build:** Succeeded (`npx vite build` built all client bundles cleanly).
- **Fallow audit:** Clean (`dead code 0` reported on the changed files).

---

PR #714, opened 2026-08-01
test(webllm): prove the steering-adherence column reaches the report (#608)

Samhit21:
## Summary

`Refs #608` — deliberately **not** `Closes`. The eval numbers are still
unmeasured, and this PR does not measure them. It makes the run that will
measure them trustworthy.

#608's remaining acceptance criterion is *"eval numbers across the shipped model
list, with a stated conclusion for half 2"*. That number can only come from the
browser leg: WebGPU-only, one model per tab, several GB of weights, fully
manual. Nothing verified that the number survives the trip from fixture to
committed report.

### The gap

`adherence.test.ts` proves the scorer bites. `rubric.test.ts` proves
`scoreRubric` fills the field. Neither covers the path that produces what a
human actually reads:

```
fixture → runEval → RunRecord → AggregateRow → Markdown table
```

Every existing runner and report test pinned `steeringAdherence` and
`steeringAdherenceRate` to `null`, so a break anywhere along that path renders
an **em dash in every Steering cell** — which reads as *"this fixture doesn't
probe steering"*, not as *"the harness is broken"*.

That is the worst shape a bug can take here: silent, indistinguishable from a
legitimate value, and only discoverable **after** the expensive part is done.
#608's own rule was to prove the scorer bites before trusting its verdict. This
is the same argument one level up, about the instrument that reports it.

### What the test pins

Both directions on the **real committed fixture**, not a hand-built one — a stub
that obeys the instruction and one that echoes the forbidden word back:

| case | `steeringAdherence` | aggregate rate |
|---|---|---|
| stub obeys | `true` | `1` |
| stub echoes `Spearheaded` | `false` | `0` |
| fixture with no probe | `null` | `null` (renders `—`) |

Plus the rendered table, including the **separator-row width** — a short
separator silently collapses a column in some Markdown renderers, which is
exactly how a column goes missing from a committed report without failing
anything.

**The composite case isolates its variable.** A first version compared two
differently-shaped outputs and they happened to tie at `0.8333` — proving
nothing in either direction, since their other criteria offset. It now compares
two outputs differing in **exactly one token** and asserts every other rate is
equal, so any composite delta is attributable to adherence alone.

### The committed reports are not evidence about half 2

The three reports in `tests/fixtures/rewrite/reports/` are from **2026-06-23**.
They predate both the Steering column and the `steering-forbidden-word` fixture,
so they measured nothing about adherence — their missing column is an artifact
of *when they ran*, not a finding.

The next reader's likeliest mistake is treating them as a measurement, so the
reports README now says so directly, and points at the model × variant reading
in `src/lib/webllm/eval/README.md`.

## What is still open on #608

The eval grid itself. It needs a maintainer with WebGPU:

```sh
npm run eval:rewrite   # once per registry model, one per tab
```

Then commit the report pairs and state the half-2 conclusion (fixed /
not-reproducible / split out). Everything else in #608 shipped in #666.

## Test plan

- [x] `npm run verify` green — **4175 passed / 10 skipped** (4164 on `main` + the
      11 added here), typecheck + lint + build clean
- [x] The new test fails against a broken harness by construction: it asserts a
      populated value where the prior tests asserted `null`
- [ ] Eval grid **not** run — needs WebGPU hardware, which is the whole point of
      the PR above

### ⚠️ A pre-existing full-suite flake, reported because I hit it

One earlier `verify` run on this branch failed 5 tests across the three slowest
PDF files (`export-layout-contract`, `render-roundtrip.repro`,
`to-json-resume` — 85s and 11s files doing real font embedding). It did **not**
reproduce:

- clean `origin/main`, full `verify`: **exit 0**, 4164 passed, zero failures
- those three files in isolation, with coverage: **116 passed**
- this branch, full `verify`, re-run: **exit 0**, 4175 passed

So it is load-dependent timeout flake in the full parallel suite, not caused by
this change — this branch adds 11 tests that run in 13ms. Flagging it because a
flaky **required** gate costs a re-run on someone else's PR eventually, and the
failure mode there (four PDF-layout assertions) looks alarming enough to send
somebody debugging the wrong thing.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation + tests | Claude Opus 5 (1M context) | high |
| Verification | `npm run verify` — green locally | — |

### DIFF
--- a/src/lib/webllm/eval/adherence-reporting.test.ts
+++ b/src/lib/webllm/eval/adherence-reporting.test.ts
@@ -1,0 +1,244 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * End-to-end wiring for the steering-adherence column (#608 half 2).
+ *
+ * `adherence.test.ts` proves the SCORER bites and `rubric.test.ts` proves
+ * `scoreRubric` populates the criterion. Neither covers the path that actually
+ * produces the number a human reads: fixture → `runEval` → `RunRecord` →
+ * `AggregateRow` → the Markdown table. Before this file, every runner and
+ * report test set `steeringAdherence` / `steeringAdherenceRate` to `null`, so a
+ * break anywhere along that path would have surfaced as a `—` in every Steering
+ * cell of a committed report.
+ *
+ * That failure mode is expensive in a way a normal test gap is not. The
+ * inference leg is browser-only and one model per tab, so the run it would
+ * silently invalidate is a multi-model, multi-gigabyte manual session — and the
+ * `—` reads as "this fixture doesn't probe steering", not as "the harness is
+ * broken". #608's own rule was to prove the scorer bites before trusting its
+ * verdict; this is the same argument one level up, about the instrument that
+ * reports it.
+ *
+ * Everything here runs on a stub `RewriteFn`. No model, no WebGPU — the point
+ * is the plumbing, not the adherence of any real model.
+ */
+
+import { describe, expect, it } from "vitest";
+import { runEval } from "./runner.ts";
+import { renderMarkdownReport } from "./report.ts";
+import { REWRITE_FIXTURES, getFixtureById } from "./fixtures.ts";
+import type { RewriteFn, RewriteFixture } from "./types.ts";
+
+const MODEL = "Qwen2.5-1.5B-Instruct-q4f16_1-MLC";
+const VARIANT = "shipped";
+
+/** The committed probe fixture — the one a real run measures half 2 with. */
+const STEERING_FIXTURE = getFixtureById("steering-forbidden-word");
+
+/** A stub that returns `bullets` verbatim for every cell. */
+function stubReturning(bullets: readonly string[]): RewriteFn {
+  return async () => ({ bullets, raw: bullets.join("\n") });
+}
+
+/** Echo the fixture's own input back — the "model ignored the instruction"
+ *  case for the forbidden-word probe, since every input bullet contains it. */
+const echoInput: RewriteFn = async ({ fixture }) => ({
+  bullets: fixture.bullets,
+  raw: fixture.bullets.join("\n"),
+});
+
+async function runOne(fixture: RewriteFixture, rewriteFn: RewriteFn) {
+  return runEval({
+    modelIds: [MODEL],
+    variantIds: [VARIANT],
+    fixtures: [fixture],
+    rewriteFn,
+    now: () => 0,
+  });
+}
+
+/** `| a | b |` → `["", "a", "b", ""]`. Header and row split identically, so
+ *  a column's index in one is its cell's index in the other. */
+function splitRow(line: string): string[] {
+  return line.split("|").map((c) => c.trim());
+}
+
+/**
+ * The aggregate table, addressed BY COLUMN NAME.
+ *
+ * Matching a whole rendered row against a substring is what made the first
+ * version of the two tests below unfalsifiable (#714 review): every other cell
+ * in this run's row is `100%`, and `100%` contains `0%`, so a row whose
+ * Steering cell had silently become `—` still matched `/0\.0%|0%/`. And `—`
+ * appears in the Judge and Dedup cells of every report the judge is disabled
+ * for — the default here — so `md.toContain("—")` was true of the probed report
+ * too, i.e. of the exact output the sibling test says must NOT show one.
+ *
+ * Selecting the cell by its header index is the smallest fix that discriminates,
+ * and it survives a column reorder in `report.ts` for free.
+ *
+ * `headerIdx + 2` is the first data row: `renderMarkdownReport` emits header,
+ * separator, then one row per model × variant, and every run in this file is a
+ * single model and a single variant.
+ */
+function aggregateTable(md: string): {
+  header: string;
+  separator: string;
+  row: string;
+  cell: (column: string) => string;
+} {
+  const lines = md.split("\n");
+  const headerIdx = lines.findIndex((l) => l.startsWith("| Model | Variant |"));
+  if (headerIdx < 0) throw new Error(`no aggregate table in report:\n${md}`);
+  const header = lines[headerIdx]!;
+  const row = lines[headerIdx + 2]!;
+  const columns = splitRow(header);
+  const cells = splitRow(row);
+  return {
+    header,
+    separator: lines[headerIdx + 1]!,
+    row,
+    cell: (column) => {
+      const i = columns.indexOf(column);
+      if (i < 0) throw new Error(`no "${column}" column in: ${header}`);
+      return cells[i]!;
+    },
+  };
+}
+
+describe("the steering fixture is actually shipped", () => {
+  it("is registered in REWRITE_FIXTURES and carries a probe", () => {
+    // If the fixture is dropped from the registry, every assertion below still
+    // passes on a hand-built object while a real run measures nothing.
+    // `getFixtureById` IS `REWRITE_FIXTURES.find(...)` (`fixtures.ts:154`), so
+    // these two assertions already say "a registry fixture carries a probe" —
+    // a third `REWRITE_FIXTURES.some(f => f.steering)` read as an independent
+    // guarantee while being the same one (#714 review).
+    expect(STEERING_FIXTURE).toBeDefined();
+    expect(STEERING_FIXTURE!.steering?.check).toEqual({
+      kind: "forbidden-word",
+      word: "spearheaded",
+    });
+  });
+});
+
+describe("runEval populates steeringAdherence on the record", () => {
+  it("is TRUE when the stub obeyed the instruction", async () => {
+    const report = await runOne(
+      STEERING_FIXTURE!,
+      stubReturning([
+        "Led the billing platform migration across 12 regional markets",
+        "Cut scope churn 30% with a quarterly planning process",
+      ]),
+    );
+    expect(report.records[0]!.rubric.steeringAdherence).toBe(true);
+  });
+
+  it("is FALSE when the stub echoed the forbidden word back", async () => {
+    // The fixture's inputs all lead with "Spearheaded", so echoing them is
+    // exactly the non-compliance the probe exists to detect.
+    const report = await runOne(STEERING_FIXTURE!, echoInput);
+    expect(report.records[0]!.rubric.steeringAdherence).toBe(false);
+  });
+
+  it("stays NULL for a fixture with no probe, leaving other fixtures alone", async () => {
+    const plain = REWRITE_FIXTURES.find((f) => f.steering === undefined);
+    expect(plain).toBeDefined();
+    const report = await runOne(plain!, stubReturning(["Led the migration"]));
+    expect(report.records[0]!.rubric.steeringAdherence).toBeNull();
+  });
+});
+
+describe("the aggregate computes a real rate, not null", () => {
+  it("is 0 when the only probe fixture failed", async () => {
+    const report = await runOne(STEERING_FIXTURE!, echoInput);
+    expect(report.aggregates[0]!.steeringAdherenceRate).toBe(0);
+  });
+
+  it("is 1 when it passed", async () => {
+    const report = await runOne(
+      STEERING_FIXTURE!,
+      stubReturning(["Led the billing platform migration across 12 markets"]),
+    );
+    expect(report.aggregates[0]!.steeringAdherenceRate).toBe(1);
+  });
+
+  it("is null — not 0 — when no fixture in the run probes steering", async () => {
+    // The distinction the report renders as `—`. Scoring an un-probed run as 0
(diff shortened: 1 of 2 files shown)


src/lib/webllm/eval/adherence-reporting.test.ts
```diff
@@ -0,0 +1,186 @@
...
+  });
+
+  it("shows a percentage for a probed run, not an em dash", async () => {
+    const md = renderMarkdownReport(await runOne(STEERING_FIXTURE!, echoInput));
+    const row = md
+      .split("\n")
+      .find((l) => l.includes(MODEL) || l.toLowerCase().includes("qwen"))!;
+    // The aggregate row's Steering cell. `—` here is the bug this file exists
+    // to catch: it would read as "not probed" on a run that did probe.
+    expect(row).toMatch(/0\.0%|0%/);
```
s-annam: **Blocking** — this assertion cannot detect the bug the file exists to catch.

`pct()` renders integers, so the Steering cell for this run really is `0%`. But the regex is matched against the *whole row*, and `100%` contains the substring `0%` — every other cell in this row is `100%`. Rendering the same report with the wiring broken (`steeringAdherenceRate: null`, so the cell is `—`):

```
broken:  | Qwen 2.5 (1.5B) | shipped | 100% | 100% | 100% | 100% | 100% | — | — | — | **83%** |
```

`/0\.0%|0%/.test(brokenRow)` → `true`. I rendered both to check.

(`0\.0%` is dead either way — `pct` is `Math.round(v * 100)`, so a tenth never appears.)

Fix: select the Steering cell by its header index, which also survives a column reorder.

```ts
const cells = row.split("|").map((c) => c.trim());
const steeringIdx = header.split("|").map((c) => c.trim()).indexOf("Steering");
expect(cells[steeringIdx]).toBe("0%");
```

That needs `header` in scope here — worth hoisting the header lookup into a small helper that both this test and the separator test use.

src/lib/webllm/eval/adherence-reporting.test.ts
```diff
@@ -0,0 +1,186 @@
...
+  });
+
+  it("still shows an em dash for a run with no probe", async () => {
+    const plain = REWRITE_FIXTURES.find((f) => f.steering === undefined);
+    const md = renderMarkdownReport(
+      await runOne(plain!, stubReturning(["Led the migration"])),
+    );
+    expect(md).toContain("—");
```
s-annam: **Blocking** — `expect(md).toContain("—")` can never fail, on any report.

`judgeMean` is `null` whenever the judge is disabled — the default, and the only mode these tests run in — so `numOrDash` emits `—` into every aggregate row regardless of steering. `dedupEffectiveRate` does the same for any non-`redundant` fixture. I confirmed the **probed** run's Markdown also contains `—`, so this assertion passes on the exact report the previous test says must *not* show one.

Fix: same cell-index approach, asserting the Steering cell specifically.

```ts
expect(cells[steeringIdx]).toBe("—");
```

Taken with the previous finding, the ticked test-plan box — *"the new test fails against a broken harness by construction: it asserts a populated value where the prior tests asserted `null`"* — does not hold for the report-rendering layer, which is the layer the PR title names. It does hold for the record- and aggregate-level tests above, and those are the majority of the file.

tests/fixtures/rewrite/reports/README.md
```diff
@@ -37,3 +37,26 @@ A baked-in baseline report will land here in a follow-up PR once the
 first WebGPU run is captured on a maintainer machine; this directory is
 intentionally empty in the PR that introduces the harness so the
 artifact reflects a real run, not a synthetic placeholder.
+
+## ⚠️ The committed reports predate the Steering column (#608)
```
s-annam, line 36: **Secondary** — this new section contradicts the paragraph three lines above it, which still reads:

> A baked-in baseline report will land here in a follow-up PR once the first WebGPU run is captured on a maintainer machine; **this directory is intentionally empty** in the PR that introduces the harness, so the artifact reflects a real run, not a synthetic placeholder.

The directory has held six report files since 2026-06-23. Reading top-to-bottom, you are told the directory is empty and then, immediately, that three reports here are from a specific date.

The stale sentence is in scope precisely because this PR's purpose is making this README accurate about what has and has not been measured — a reader who stops at the first paragraph draws the wrong conclusion about whether any run exists at all. Suggest deleting or past-tensing it in the same commit.

tests/fixtures/rewrite/reports/README.md
```diff
@@ -37,3 +37,26 @@ A baked-in baseline report will land here in a follow-up PR once the
...
+
+The three reports here are from **2026-06-23** and carry nine columns:
```
s-annam: **Nit** — the count and the list describe two different tables.

The names given are the **aggregate** table's, which in all three committed reports has ten columns (`Model` and `Variant` lead it) and eight criteria names. Nine is the **per-cell** table's count:

```
| Fixture | Kind | In → Out | Numbers | Verb | Length | Preamble | Dedup | Error |
```

Verified against all three `.md` files. Suggest dropping the number and just listing the names, or saying "eight criteria columns".

src/lib/webllm/eval/adherence-reporting.test.ts
```diff
@@ -0,0 +1,186 @@
...
+describe("the steering fixture is actually shipped", () => {
+  it("is registered in REWRITE_FIXTURES and carries a probe", () => {
+    // If the fixture is dropped from the registry, every assertion below still
+    // passes on a hand-built object while a real run measures nothing.
+    expect(STEERING_FIXTURE).toBeDefined();
+    expect(STEERING_FIXTURE!.steering?.check).toEqual({
+      kind: "forbidden-word",
+      word: "spearheaded",
+    });
+    expect(REWRITE_FIXTURES.some((f) => f.steering !== undefined)).toBe(true);
```
s-annam: **Nit** — subsumed by the assertion two lines up. `getFixtureById` is `REWRITE_FIXTURES.find(...)` (`fixtures.ts:154`), so a defined `STEERING_FIXTURE` carrying a `steering.check` already proves a registry fixture has a probe. Harmless, but it reads as a second independent guarantee when it is the same one.

s-annam (changes requested):
The record- and aggregate-level tests here are exactly right and do bite. Two of the three **report-rendering** tests cannot fail — including the one guarding the precise failure the file's docblock names: a `—` in the Steering cell.

I reproduced this rather than reasoning about it. Same run, rendered twice — once as-is, once with the wiring broken (`steeringAdherenceRate` forced to `null`):

```
working: | Qwen 2.5 (1.5B) | shipped | 100% | 100% | 100% | 100% | 100% | — | 0% | — | **83%** |
broken:  | Qwen 2.5 (1.5B) | shipped | 100% | 100% | 100% | 100% | 100% | — | — | — | **83%** |
```

`/0\.0%|0%/` matches the broken row too, because `100%` contains the substring `0%`. Details inline.

## Gates

| Gate | Result |
|---|---|
| Generic correctness pass | folded in |
| Fixture PII (3a) | ✅ `check:fixtures` green — 58 PDFs, 15 sidecars; no binary added by this PR |
| Design system / reuse (3b) | n/a — no `src/components/` change |
| Style tokens (3c) | n/a — no UI |
| fallow dead-code (3d) | ✅ clean over both changed files |
| Skill/script command bugs (3e) | n/a |
| Description accuracy (3f) | ⚠️ one ticked test-plan box overclaims — see Blocking 1 and 2 |
| `npx vitest run src/lib/webllm/eval` | ✅ 100 passed |
| `npm run typecheck`, `eslint` | ✅ clean |

## Acceptance criteria (#608)

`Refs`, not `Closes` — correctly scoped, and the PR is straight about what remains. The open ACs (the eval grid across the shipped model list, plus a stated half-2 conclusion) are accurately described as blocked on WebGPU hardware, so nothing ships as "done" that isn't. I verified #666 is merged, so "everything else in #608 shipped in #666" holds.

One thing worth stating plainly, because it is the PR's real load-bearing claim: the browser leg **does** deliver the fixture's instruction into the production prompt — `run-eval-browser.ts:79` passes `fixture.steering.instruction` through `buildSteeringSuffix`, not a hand-inlined string. So the number a real run produces will be about a model that was actually told. That is the part that makes this guard worth having.

## What is right

- Running against the **committed** fixture via `getFixtureById` rather than a hand-built object, with a registry assertion guarding against the fixture being dropped. That is the failure mode this class of test usually misses.
- The composite-isolation test asserting every *other* rate is equal before comparing `aggregateScore`. Pinning the isolation instead of assuming it is what makes the delta attributable — and the docblock saying so, including that an earlier version tied and proved nothing, is the right thing to leave behind.
- The separator-row width check.
- Claiming no half-2 conclusion.

The fixes below are small and local; the structure of the file does not need to change.

---

src/lib/webllm/eval/adherence-reporting.test.ts
Samhit21: Fixed in `43b9fdc`. Reproduced your result first — with `steeringAdherenceRate: null` the row renders

```
| Qwen 2.5 (1.5B) | shipped | 100% | 100% | 100% | 100% | 100% | — | — | — | **83%** |
```

and `/0\.0%|0%/` matches it. The assertion now selects the cell by header index and asserts `toBe("0%")`, which fails on that row. Hoisted the header/row/cell lookup into an `aggregateTable()` helper shared with the separator test, as you suggested; it takes a column *name*, so a reorder in `report.ts` costs nothing.

src/lib/webllm/eval/adherence-reporting.test.ts
Samhit21: Fixed in `43b9fdc`. Confirmed your point directly — the **probed** report contains `—` in its Judge and Dedup cells, so `toContain("—")` was true of the exact output the sibling test says must not show one. Now `expect(table.cell("Steering")).toBe("—")`, via the same helper.

The docblock on `aggregateTable()` records both failure modes so the next person does not re-derive why the row-level match was wrong.

On the test-plan box: you are right that the claim did not hold at the report-rendering layer. I re-ran the working/broken pair against the revised assertions to check they now discriminate — working renders `0%`, broken renders `—`, and each assertion fails on the other's input.

tests/fixtures/rewrite/reports/README.md
Samhit21: Fixed in `43b9fdc`. Deleted the stale paragraph and folded its one surviving fact — that the directory shipped empty so the artifact would reflect a real run — into the new section as past tense, so the "why was it empty" answer is not lost:

> The three reports here are from **2026-06-23** — the first WebGPU runs captured on a maintainer machine, which is why the directory shipped empty with the harness rather than carrying a synthetic placeholder.

tests/fixtures/rewrite/reports/README.md
Samhit21: Fixed in `43b9fdc`. Dropped the number and named the table: "Their aggregate table carries these eight criteria columns after `Model` and `Variant`". The nine I wrote was indeed the per-cell table's count.

src/lib/webllm/eval/adherence-reporting.test.ts
Samhit21: Fixed in `43b9fdc` — removed. Left a one-line comment noting that `getFixtureById` is `REWRITE_FIXTURES.find(...)`, so the two remaining assertions already carry the registry guarantee.

Samhit21:
Pushed `43b9fdc` (amended onto the single commit, force-pushed). All five threads addressed and resolved with per-thread replies.

**Both Blocking findings reproduce as described, and the fixes discriminate.** I rendered the working and broken (`steeringAdherenceRate: null`) reports side by side:

```
working: | Qwen 2.5 (1.5B) | shipped | 100% | 100% | 100% | 100% | 100% | — | 0% | — | **83%** |
broken : | Qwen 2.5 (1.5B) | shipped | 100% | 100% | 100% | 100% | 100% | — | —  | — | **83%** |
```

The old `/0\.0%|0%/` row match passes on the broken row; the old `toContain("—")` passes on the working one. Both now select the **Steering cell by header index** and assert `toBe("0%")` / `toBe("—")` — each fails on the other's input.

- New `aggregateTable(md)` helper shared by all three rendering tests; addresses columns by *name*, so a reorder in `report.ts` is free. Its docblock records both failure modes.
- Dropped the subsumed `REWRITE_FIXTURES.some(...)` assertion.
- README: deleted the contradicting "intentionally empty" paragraph, folding its surviving fact into the new section as past tense; corrected nine → "eight criteria columns" and named the table.

`npm run verify` green (typecheck, lint, 100 eval tests, build, fallow clean on all 5 changed files).

s-annam (approved):
Approving the changes. This PR introduces exceptionally thorough and precise integration tests for the WebLLM steering adherence reporting pipeline on top of synthetic fixtures, ensuring no regression or silent column omissions occur during manual browser eval runs.

## Blocking
(none)

## Secondary
(none)

## Nits
(none)

---
(Reviewed SHA: `43b9fdcfe31e54bc692ae3b6014ecc4d20cbafee`)

PR #731, opened 2026-08-02
feat(storage): DB_VERSION 4 — tombstones, an updatedAt index, and a cursor store (#730)

### LINKED ISSUE #730: DB_VERSION 4: tombstones, an updatedAt index, and a place to keep a cursor
`DB_VERSION` 4: tombstones, an `updatedAt` index, and a place to keep a cursor.

## Context

The local schema can hold a library but cannot hand one to a second holder. Three specific gaps,
and each of them is invisible until something tries.

**No tombstones.** `deleteRecord()` removes the row outright, so a deleted job leaves nothing behind
that says it was deleted. A record that is simply *gone* is indistinguishable from one that never
existed, which means any second holder of the library — another device, a backup file the user
restores, a producer re-capturing the same posting — has no way to tell a deletion from a gap and
re-adds it. The deletion silently undoes itself.

This is not hypothetical and does not require a network. It reproduces today with two files:
merge-import a backup taken before a deletion and the job comes straight back.

**No index on `updatedAt`.** "What changed since X" is a full scan of every record. Survivable at a
few hundred jobs and wrong as a design — `updatedAt` is already the tracker's sort key and the
obvious cursor for anything incremental.

**Nowhere to keep a cursor.** Nothing that reads the library in passes has a place to record how far
it got that survives a reload.

## Scope

1. `DB_VERSION` 3 → 4 in `src/lib/storage/db.ts`, as a new `if (oldVersion < 4)` block. Never edit a
   shipped block.
2. `deletedAt?: number` on `StoredRecord`; route deletion through a soft delete for `jobs` and
   `letters`. `resumes` and `boards` keep hard deletes — a résumé's bytes are the bulk of the
   database and `boards` is a pure cache. Write down which and why.
3. Index `updatedAt` on the stores that can replicate.
4. A store for per-store cursors.
5. **Audit every reader.** This is the part that breaks the app, not the migration: any list surface
   that does not filter tombstones starts rendering deleted rows.
6. Decide what a backup does with a tombstone — export it and a restore re-deletes on the target,
   omit it and a backup silently resurrects deleted jobs. Both are defensible; pick one, write down
   the reasoning, and test it.

## Do not regress

- **A `DB_VERSION` bump is a breaking change for the browser extension**, which builds these modules
  from a pinned commit of this repo (`extension/offlinecv-pin.json`) and so carries whatever version
  the pin had. Once the app upgrades a user's database, an extension built at the older pin opens it
  at a lower version and gets a `VersionError` — every capture breaks. This repo's CI never compiles
  the extension, so nothing else catches it. Bump the pin in lockstep.
- `DB_NAME` is frozen. A rename is not a migration; it is a different database, and it strands every
  user's data.
- Export/import must keep round-tripping, including v1 documents.
- `putRecord`'s `touch: false` exists so a write the user did not make does not reshuffle a
  most-recently-updated-first list. Check that a restore honours it — a timestamp collapse is
  invisible until something sorts or diffs on it.

## Acceptance criteria

- [ ] A v3 database upgrades to v4 with resumes, jobs and letters intact, and the index covers
      records written before it existed.
- [ ] Deleting a job soft-deletes it; it disappears from every UI surface; the row is still readable
      at the storage layer.
- [ ] Deleting a job still cascades to its letters.
- [ ] `updatedAt` is indexed on the replicating stores and a range query over it is exercised by a
      test.
- [ ] The cursor store reads and writes across a reload.
- [ ] The export/import tombstone decision is written down in the contract docs and covered by a
      round-trip test.
- [ ] `npm run verify` green.

s-annam:
Closes #730.

Three things the local schema did not provide, plus the reader audit that makes them safe and one
pre-existing bug they turned load-bearing.

## What landed

| | |
|---|---|
| **Tombstones** | `StoredRecord.deletedAt`. `jobs` / `letters` soft-delete; `resumes` / `boards` keep hard deletes (bytes, and a pure cache). |
| **`updatedAt` index** | On the three stores that can replicate. `createIndex` backfills, so a returning user's existing records are covered — asserted. |
| **`sync` store** | One bookmark per store. Not a `StoredRecord`, and the pull/push cursors are typed differently because they read different clocks. |

## The reader audit

Cheap, because the funnel was already there. `getAllRecords` filters tombstones **by default**
(`includeDeleted` opts out), so every list surface was correct without an edit; `getJob` /
`getLetter` filter by id so nothing above `crud.ts` can render or resurrect one.

Two behaviour calls worth review:

- `updateJob` now **throws** on a deleted job rather than reviving it by patch.
- `captureJob` still **revives**, reported as `created: true` — which is exactly what a hard delete
  always did. The tombstone exists to stop a *replica* bringing a record back behind the user's
  back, not to stop the user saving the posting again. A `deletedAt` a producer sends is stripped.

## Tombstones are exported

Omitting them is wrong in the mode that exists to combine two copies of a library: device A deletes
a job and exports, B still has it live, B merging A's file resurrects it. Replace mode looks
identical either way, so the tie breaks toward correctness in merge.

The document version is **not** bumped — an older build would reject a v3 file outright, costing the
user every record instead of resurfacing a few deleted ones. Full reasoning is in both contract docs.

## Pre-existing bug this surfaced

`importAll` stamped `updatedAt` on every restored record, collapsing a library built over months
into one instant and silently reordering a tracker sorted most-recently-updated-first. Once
`updatedAt` is a replication cursor it also claims every record on the device just changed. Import
now writes with `touch: false`; that option's precedence changed to prefer the record's own
timestamp over the stored one, which no current caller can observe (the housekeeping callers spread
the record they just read).

## ⚠️ This strands the browser extension

It builds these modules from a pinned commit (`extension/offlinecv-pin.json`) and carries whatever
version the pin had, so once a user's database is at v4 an extension at the old pin gets a
`VersionError` and every capture breaks. This repo's CI never compiles it, so nothing here catches
it. **The pin bump follows in the extension repo, referencing this commit.**

## Verification

`npm run verify` green — 4690 passed, 10 skipped; build clean. fallow: dead-code findings are the
extension's own `package.json` (inherited); the new staged exports are registered in `.fallowrc.jsonc`
beside `captureJob`, with the same time-bounded rationale.

New suite `src/lib/storage/sync-schema.test.ts` covers the v3→v4 migration (seeded by replaying the
shipped v3 upgrade blocks, not by calling `getDB()`), index backfill, tombstone visibility and
idempotence, the cascade, the range query, cursor persistence across a reload, and the
export/import round trip including the two-device merge case.

## Provenance

Written with Claude Code (Opus 5).

### DIFF
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -193,7 +193,33 @@
       // starts importing rather than letting the suppression rot.
       "saveLetter", "getLetter", "getAllLetters", "lettersForJob", "deleteLetter",
       "validateLetterRecord", "LETTER_CONTRACT_VERSION", "LetterRecordValidation",
-      "LetterRecord", "LetterProvenance", "SkippedLetter"
+      "LetterRecord", "LetterProvenance", "SkippedLetter",
+
+      // The replication surface (#730) — the schema half of cloud sync. Same
+      // "no consumer anywhere yet" exception `captureJob` above is, and for a
+      // stronger reason: the reader is the browser extension, which builds
+      // these modules from a PINNED commit of this repo and therefore cannot
+      // define an object store, an index, or a record shape of its own. The
+      // migration that creates the `updatedAt` index and the `sync` store lands
+      // here because it has to, so the query that reads the index and the
+      // accessor that reads the store belong here beside it — the alternative
+      // is that consumer re-deriving this repo's schema against a store it does
+      // not own, which is the second-definition failure the contract modules
+      // above exist to prevent.
+      //
+      // `isLive` is the one name here with a real in-tree consumer, and it is
+      // listed anyway only for its BARREL re-export: `crud.ts` exports it for
+      // `jobs.ts` / `letters.ts` / `backup.ts`, which deep-import it, so the
+      // barrel copy is the dead one. If the barrel copy is still unread when
+      // sync lands, take it back off the barrel rather than off this list.
+      //
+      // Time-bounded like `captureJob`'s entry: once something in-tree imports
+      // these, they come back OUT. If sync is abandoned, delete
+      // `sync-cursor.ts` and `listRecordsUpdatedSince` rather than leaving the
+      // suppression to rot — but note the SCHEMA cannot be rolled back the same
+      // way, since a shipped `DB_VERSION` is in users' browsers for good.
+      "isLive", "listRecordsUpdatedSince", "getSyncCursor", "setSyncCursor",
+      "SyncableStoreName", "SyncCursorRecord"
     ] },
 
     // A contract constant that no in-tree caller reads at all — not even the
--- a/docs/cover-letter-contract.md
+++ b/docs/cover-letter-contract.md
@@ -60,7 +60,8 @@ document as-is with no encoding step.
 | `label` | string | User-facing name, so two drafts for one job are tellable apart (`"Warm open"`, `"Short version"`). |
 | `resumeId` | non-empty string | The `ResumeRecord.id` this letter was written from. **User-owned.** Cleared, not orphaned, if that résumé is deleted — see §5. |
 | `producer` | object | Provenance. See §6. |
-| `createdAt` / `updatedAt` | finite number (epoch ms) | Timestamps. The store owns them; a backup file carries them so a restore preserves `createdAt`. |
+| `createdAt` / `updatedAt` | finite number (epoch ms) | Timestamps. The store owns them; a backup file carries them so a restore preserves both. |
+| `deletedAt` | finite number (epoch ms) | A **tombstone**: the letter is deleted, and its presence says so. See §5. |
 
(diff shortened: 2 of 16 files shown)


s-annam (commented):
An exceptionally clean and well-documented implementation of the v4 storage schema. All acceptance criteria for the DB migration, soft delete/tombstone propagation, range querying, and cursor storage are met and verified by thorough tests.

*(Note: Submitted as a Comment review because GitHub prevents approving one's own pull request.)*

## Blocking
None.

## Secondary
None.

## Nits
* **Fallow Duplication warnings**: `fallow audit` reports minor duplication clone groups between validation contracts (`job-record-contract.ts` and `letter-contract.ts`) and test setups (`letters.test.ts` and `sync-schema.test.ts`). These duplicates are expected for data validation/test helpers and do not warrant deduplication refactoring.

---

Issue #715, opened 2026-08-01
Letters UI: /jobs/saved, a letter indicator per tracked job, and a plain-text reveal


s-annam:
## Problem

The `letters` store shipped in #711 with a validator, a contract doc, cascade rules and
backup round-tripping — and **no UI at all**. `rg -il letter src/components src/design-system`
returns nothing. A letter written into a user's library today is real, exportable, and
invisible.

That was the deliberate call in #711 (`docs/cover-letter-contract.md` §0: "offlinecv stores
letters and does not write them"). It stops being acceptable the moment a producer actually
writes one — and one now does: `.claude/skills/cover-letter/` generates letters and imports
them through the backup door. The skill currently has to hand the user a `.txt` file
alongside the stored record, because the app cannot show them their own letter.

### 1. `/jobs/saved` — a direct route to the Saved jobs tab

`JobsApp` hosts two peer `Tabs` panels (#690). Today only `/jobs/` exists, so linking
someone to their tracked library means "go to /jobs/, then click Saved jobs". Add a route
that lands on the Saved jobs panel directly.

No router in this app, so this is a path/hash decision, not a dependency: pick whichever
keeps `Tabs` mounting both panels (switching must not lose an in-progress query or tracker
state — that is the invariant #690 established).

### 2. A letter indicator on each tracked job

For every job in the Saved jobs library, show whether a cover letter exists — an icon, not
a row of text. Absent when there is none; do not render an empty affordance for every job.

Reads: `letters` by `jobId`. Note a job can have **several** drafts (contract §2 — no
derivation, several drafts per job is the normal case), so the indicator has to survive
n > 1, and `label` is the field that tells them apart.

### 3. Click to reveal the letter

Click the icon → the letter body. Requirements:

- **Plain text presentation.** `LetterRecord.body` is typed markdown, but the letters this
  app will hold are plain prose — that is what an employer pastes into a form or an email.
  Render it as text with paragraph breaks preserved. Do **not** add a markdown renderer;
  an employer seeing `**bold**` is the failure mode we are avoiding.
- **Copy to clipboard** is the primary action. That is what the artifact is for.
- Multiple drafts: pick by `label`, most-recently-updated first.

### 4. One-time egress acknowledgement

A letter is generated outside the app's custody boundary (today by a Claude Code skill,
later possibly by a BYOK path). Show a header stating that plainly, with a **confirm-once**
that persists — not a banner on every visit.

[...]

PR #732, opened 2026-08-03
feat(webllm,design-system): offer the rewrite prompt as a copyable artifact (#609)

### LINKED ISSUE #609: Offer a copyable rewrite prompt as an alternative to the on-device model
## Problem

The rewrite lane offers exactly one way to rewrite a résumé: the on-device WebLLM pass. That pass runs a 1–3B-class instruct model in the browser, which is the right default — it is the only option that keeps the résumé text on the device, and it is what makes the custody claim true. But it is also, straightforwardly, the least capable model most users have access to. Someone with a ChatGPT, Claude, or Gemini subscription open in the next tab has a far stronger rewriter available and no way to use it here, because the thing they would need — the prompt — is locked inside our bundle.

Today they are left with two bad options: accept the weaker on-device rewrite, or improvise their own prompt from scratch and lose every guardrail this project has tuned (number preservation, no fabrication, verb variation, the recency-weighted length budget).

**Give them the prompt.** Add a secondary path in the rewrite dialog: *copy the prompt*, take it to whatever model they prefer along with their own résumé document, and come back to apply the result through the editor that already exists.

This costs the project nothing in privacy posture — a clipboard write is not egress, and where the user takes it afterwards is their own informed choice, made deliberately rather than by our default. It is strictly more honest than the current state, where the only rewrite we offer is the weakest one available and we do not say so.

## Scope

**In scope**

- A secondary action in the existing rewrite steering dialog that reveals and copies a rewrite prompt.
- The prompt is **not shown by default** — the on-device pass stays the primary, one-click path. This is a disclosure, not a fork in the road.
- The prompt is built from the **current résumé's actual shape** and the user's current steering (page target + freeform instructions), so it is not a generic template.
- Copy-to-clipboard, with in-place confirmation.

**Out of scope (deliberately)**

- **Paste-back import.** The natural next thought is a "paste the rewritten résumé back in" box that re-parses and diffs. Not in this issue. `/` already has full inline editing (`ReconstructedResume` + `EditableField`), so the round-trip the user described is already possible today — the user edits fields with the external model's output in hand. A structured paste-back is a genuinely larger feature (parse arbitrary prose back into the canonical model, diff it, review it) and deserves its own issue once this one has shipped and we know whether people use it.
- **Any network call.** No provider integration, no BYOK. `#320` is the future BYOK issue and this is not it. Nothing here fetches anything.
- **Copying the résumé text itself** — see the design decision below; recommend deferring.

## Design decisions to settle in the PR

These are the calls that make or break the feature. Settle them explicitly in the PR body, do not decide them silently in code.

### 1. One prompt, or the shipped per-section prompts?

The shipped prompts are **not** reusable as-is. `SECTION_REWRITE_SYSTEM_PROMPT` (`src/lib/webllm/rewrite-section.ts:28`) and `SUMMARY_REWRITE_SYSTEM_PROMPT` (`src/lib/webllm/rewrite-summary.ts:39`) are written for a small model rewriting **one section at a time** with a rigid line-per-bullet output contract:

```
- Output one bullet per line. No numbering. No bullet markers. No quotes. No preamble.
```

A frontier model handed the whole résumé does not need that contract and is hobbled by it. So the exported prompt is a **different artifact** with the **same guardrails**.

**That is the drift risk, and it is the main thing this issue has to get right.** Two prompts stating the number-preservation rule in two places will disagree within a release. The fix: extract the guardrail sentences into named constants in a shared module and have **both** prompts compose them, so the rules are written once and the I/O contract is the only difference. At minimum, `Preserve every concrete number from the input EXACTLY. Do not invent new numbers or metrics.` must have exactly one definition in the tree.

Reuse the steering machinery too: `buildSteeringSuffix` (`src/lib/webllm/steering.ts:66`) already turns `pageTarget` + `userInstructions` into prompt text, including the recency-weighted compression guidance in `PAGE_BUDGET`. The exported prompt must include the user's current steering — a user who set "one page" and typed instructions expects the copied prompt to carry them. Call the existing function; do not restate the budget.

### 2. Does the copied text include the résumé?

The user's framing was prompt-only: they bring their own résumé document to the external model. That is the cleaner default and the recommendation — the copied text is instructions, nothing more, and the feature ships with no résumé content on the clipboard at all.

The tempting extra is a second button that copies the **reconstructed résumé text** (what our parser read back), so the external model works from the same text our editor holds and the results map back field-by-field. It is genuinely more useful. It is also the one part of this feature that puts résumé content on the clipboard, and it needs its own affordance and its own label — never folded into the same button. **Recommend: ship prompt-only, and file the résumé-text copy as a follow-up** once the basic path is proven. If it is built here anyway, it is a separate button with separate copy, and the acceptance criteria below apply to it.

Note for whoever writes the copy: a clipboard write is local, so nothing here contradicts the custody claim in `CLAUDE.md`. Do not write reassurance copy implying we prevent the user from pasting it somewhere — that would be exactly the self-serving negation the copy discipline forbids. State what the feature does; the user decides the rest.

### 3. Where the disclosure lives

The rewrite steering dialog (`RewriteSteeringBox`, `src/components/features/ResumeRewrite.tsx:167`) is the right home: it is already the place where the user expresses rewrite intent, and the prompt must carry that intent. A collapsed secondary affordance below the primary CTA — not a second prominent button competing with it. The on-device pass stays the default answer.

## Files involved

| File | Role |
|---|---|
| `src/components/features/ResumeRewrite.tsx` | The CTA + steering dialog. `RewriteSteeringBox` at :167. **Where the UI goes.** Already 200+ LOC — put the new surface in a new sibling component, per `CLAUDE.md`'s size rule, not inline. |
| `src/lib/webllm/steering.ts` | `RewriteSteering` (:32), `buildSteeringSuffix` (:66), `PAGE_BUDGET` (:48). Reused verbatim. |
| `src/lib/webllm/rewrite-section.ts` | `SECTION_REWRITE_SYSTEM_PROMPT` (:28) — the guardrail source. |
| `src/lib/webllm/rewrite-summary.ts` | `SUMMARY_REWRITE_SYSTEM_PROMPT` (:39) — same. |
| `src/hooks/useResumeRewrite.ts` | Owns `pageTarget` / `userInstructions` state (:255-270); the prompt builder reads the same values. |
| `src/components/features/SectionRewrite.tsx` | Existing hand-rolled copy button at :384. |
| `src/components/features/WebGpuUnavailableNotice.tsx` | Existing hand-rolled copy button at :178. |
| `src/design-system/` | Where the extracted copy affordance belongs. |

### 1. Extract a copy-to-clipboard affordance into the design system

`navigator.clipboard.writeText` is currently hand-rolled in **two** feature components — `WebGpuUnavailableNotice.tsx:178` and `SectionRewrite.tsx:384` — with their own confirmation handling each time. This issue would be the third. Per the Golden Rule in `CLAUDE.md`, the third occurrence is where it stops: extract a `CopyButton` (or `useCopyToClipboard` hook, if the two call sites' markup is too different to unify) into `src/design-system/`, and migrate both existing call sites onto it in the same PR.

Requirements the two existing sites establish between them: an in-place confirmation (there is **no toast primitive** in this design system — confirm by swapping content in place), an `aria-live="polite"` announcement, meaning carried by text and never by colour alone, and a graceful no-op where `navigator.clipboard` is absent (note `WebGpuUnavailableNotice` already uses `?.` for exactly this).

If migrating both existing call sites turns out to balloon the diff, migrate them and say so in the PR — do not add a third parallel implementation and leave a TODO.

### 2. Add the shared guardrail constants

New module (suggest `src/lib/webllm/rewrite-guardrails.ts`) holding the rules both prompts state, as named exports. Rewrite `SECTION_REWRITE_SYSTEM_PROMPT` and `SUMMARY_REWRITE_SYSTEM_PROMPT` to compose them.

**This must be a byte-identical refactor of the shipped prompts.** Assert it: snapshot the two existing prompt strings *before* the change and assert the composed versions equal them exactly. A prompt change is a behaviour change for every user of the on-device path, and it must not happen as a side effect of this feature.

### 3. Build the exportable prompt

New pure module (suggest `src/lib/webllm/export-prompt.ts`) — a function from the current résumé shape + `RewriteSteering` to a string. Pure, no I/O, no clock, unit-testable at module scope, following `src/lib/score/score.ts` as the exemplar for a zero-dep lib module.

It should compose: the guardrails from step 2, an instruction framing the task for a capable model working on a whole résumé, the steering suffix from `buildSteeringSuffix`, and a short note telling the user to attach or paste their own résumé. Keep it readable — this string is a user-facing artifact that a person will read before they use it, unlike every other prompt in the tree.

Consider carrying the app's own findings here too (the critique output, the term-quality `missing` skills) — a copied prompt that says *"the résumé is missing quantification on these bullets"* is dramatically more useful than a generic one. **Check #608 first**: that issue is wiring the same findings into the on-device prompt, and the two should share the joining logic rather than derive it twice. If #608 has not landed, ship this without findings and note the follow-up.

### 4. Wire the UI

New sibling component under `src/components/features/` (not inline in `ResumeRewrite.tsx`, which is already at the size limit). Collapsed by default; reveals the prompt in a read-only `TextAreaField` with the `CopyButton` from step 1. Use `@design-system` imports only, semantic tokens only, no raw `<button>`.

The prompt must reflect the user's **current** steering — if they change the page target or edit the instructions with the disclosure open, the shown prompt updates. That is what makes it worth building rather than a static string in the docs.

## Acceptance criteria

- [ ] The rewrite dialog's primary path is unchanged: a user who does not open the disclosure sees and does exactly what they do today.
- [ ] The prompt is **not** visible by default; it takes a deliberate action to reveal.
- [ ] The revealed prompt contains the number-preservation and no-fabrication guardrails, asserted by matching against the shared constants — not against a copy of the sentence in the test.
- [ ] `SECTION_REWRITE_SYSTEM_PROMPT` and `SUMMARY_REWRITE_SYSTEM_PROMPT` are **byte-identical** to their pre-refactor values, asserted against strings captured before the change. Prove it: the on-device rewrite path must produce the same prompts it produced before this PR.
- [ ] Changing the page target updates the shown prompt; typing freeform instructions updates the shown prompt. Assert both — a prompt built once on mount and never updated is the likely bug and would pass a naive "the prompt renders" test.
- [ ] With no steering set, the prompt is still valid and complete (no dangling "The user has these additional instructions:" with nothing after it).
- [ ] Copying puts the prompt on the clipboard and confirms in place, with an `aria-live` announcement. No toast, no colour-only signal.
- [ ] Where `navigator.clipboard` is unavailable, the UI degrades without throwing and the prompt is still selectable as text.
- [ ] The copied text contains **no résumé content** (given decision 2's recommendation) — assert this explicitly with a résumé fixture whose contents are checked for absence. If the résumé-text copy is built anyway, it is a separate control with its own label, and its own test.
- [ ] `CopyButton` / `useCopyToClipboard` lives in `src/design-system/`, is exported through the barrel, and **both** pre-existing hand-rolled call sites (`WebGpuUnavailableNotice.tsx:178`, `SectionRewrite.tsx:384`) are migrated onto it. No third parallel implementation.
- [ ] No new component exceeds ~200 LOC; `ResumeRewrite.tsx` does not grow materially.
- [ ] No raw `<button>`, no hardcoded hex, no raw Tailwind palette classes — `npm run lint` enforces all three.
- [ ] No `fetch()` is added anywhere in the change.

## Reuse analysis

**Capability:** revealing and copying a text artifact derived from the current résumé + rewrite settings, as an alternative to running the on-device model.

**Existing surfaces found:**

- `RewriteSteeringBox` (`src/components/features/ResumeRewrite.tsx:167`) — the existing rewrite-intent surface, already inside the dialog this belongs in.
- `buildSteeringSuffix` (`src/lib/webllm/steering.ts:66`) — already converts steering into prompt text.
- `SECTION_REWRITE_SYSTEM_PROMPT` / `SUMMARY_REWRITE_SYSTEM_PROMPT` — the shipped guardrails.
- Copy-to-clipboard, hand-rolled twice: `WebGpuUnavailableNotice.tsx:178`, `SectionRewrite.tsx:384`.
- `LlmEscapeHatchBanner` (`src/components/features/LlmEscapeHatchBanner.tsx`) — **considered and rejected.** Despite the name, it is the degenerate-*parse* recovery CTA (#243): it offers an on-device LLM re-parse when the cascade fails, and its whole lifecycle is model download + parse + `onRecovered`. It shares no state, no trigger, and no output with a rewrite-prompt disclosure. Reusing it would mean overloading one component with two unrelated jobs.
- `TextAreaField`, `Button`, `Dialog` (`@design-system`) — the primitives to build with.

**Decision:** **extend** on every axis. Extend `RewriteSteeringBox`'s dialog with a new sibling component (a new file for size reasons, not a new surface — it lives inside the existing dialog and has no independent entry point). Extend the design system with the copy affordance the codebase has now hand-rolled twice. Compose the existing steering and guardrail functions rather than restating either. **No** new dialog, **no** new page, **no** second rewrite entry point.

## Notes

- Related: #608 (the on-device rewrite ignores the app's own findings). Both issues concern what intent reaches a rewrite prompt. If #608 lands first, this issue inherits its findings-joining logic and the copied prompt gets meaningfully better for free — so prefer that order.
- This is not BYOK (#320). Nothing here calls a provider, holds a key, or makes a request. The user is the transport.
- Worth a line in the marketing/product copy eventually — *"or take our prompt to whatever model you like"* is a genuine differentiator for a privacy-first tool and costs us nothing. Out of scope here; mentioned so it is not forgotten.

Samhit21:
## Summary

The rewrite lane offered exactly one rewriter: the on-device 1–3B pass. It is the right default — it is the only one that keeps the résumé text on the device — and it is also the weakest model most users can reach. Anyone with a frontier model in the next tab could do better, and the only thing standing in the way was our prompt, locked inside the bundle.

This adds a **collapsed disclosure under the steering dialog's CTA** that reveals and copies a whole-résumé rewrite prompt, built from the current résumé's shape and the user's live steering. The on-device path is untouched: a user who never opens the disclosure sees and does what they did before.

Closes #609

### 1. One prompt, or the shipped per-section prompts? → **A separate artifact, composed from shared guardrails.**

New `src/lib/webllm/rewrite-guardrails.ts` holds each rule as a named export. `SECTION_REWRITE_SYSTEM_PROMPT` and `SUMMARY_REWRITE_SYSTEM_PROMPT` now compose them and keep only their own I/O contract (`SECTION_OUTPUT_CONTRACT` / `SUMMARY_OUTPUT_CONTRACT`), which is precisely the part a frontier model must not inherit. `Preserve every concrete number from the input EXACTLY. Do not invent new numbers or metrics.` now has exactly one definition in the tree, and all three prompts carry it.

**Proof the shipped prompts did not move:** `rewrite-guardrails.test.ts` pins both against literals captured from `git show 0571a82` *before* the extraction, written out in full rather than derived from the constants under test — a test that rebuilt the expected string from the same exports would only prove `composeRulesPrompt` is deterministic. It also asserts the two *builders* (`buildSectionSystemPrompt` / `buildSummarySystemPrompt`) are unchanged, since those are what actually reach the engine, and that the steering suffix still appends at the same seam.

One rule is **new and export-only**: `NO_FABRICATION_RULE` (no invented employers, titles, dates, degrees). The on-device prompts deliberately do not carry it — they see one section at a time and are never handed the fields that hold those facts. A test asserts it stays out of both.

`buildSteeringSuffix` is **called, not restated**, so the page budget and the user's instructions reach the copied prompt through the same code path the on-device run uses.

### 2. Does the copied text include the résumé? → **No. Prompt only, as recommended.**

The copied text is instructions. `export-prompt.test.ts` asserts the absence field by field against a full synthetic résumé — summary, every bullet, both section labels (the parsed title + employer, the easiest thing to leak while writing a "shape" line), and distinctive tokens. `RewritePromptDisclosure.test.tsx` repeats the assertion against the bytes actually handed to `writeText`.

The one résumé-derived thing that does appear is a **count**: *"For scale: offlinecv read this résumé as a summary paragraph and 2 roles carrying 3 bullets between them. That is what our text extractor saw, which is not always what the document holds — work from the document itself."* No field value. This is what makes it *this résumé's* prompt rather than a template out of the docs.

Copying the reconstructed résumé text is a genuinely useful follow-up and is deliberately **not** here — it needs its own control and its own label.

### 3. Where the disclosure lives → **Inside the existing steering dialog, below the CTA row, collapsed.**

A `Button variant="link"` reading "Prefer another model? Copy the prompt ↓" under a `border-t` separator. Not visible by default; no second prominent button competing with "Run rewrite". `ResumeRewrite.tsx` grew by 8 lines (424 → 432); the surface itself is a new 109-LOC sibling, `RewritePromptDisclosure.tsx`.

## Two deviations from the issue, stated rather than done silently

**The #608 findings are NOT carried into the copied prompt**, even though #608 has landed. The findings channel is keyed **per line** (`findingsKey(bullet)`), and naming the line means quoting it — which puts résumé text on the clipboard and breaks decision 2. Reduced to category notes only ("add a concrete metric"), it degrades to generic advice a capable model already follows. So `buildExportableRewritePrompt` passes no `units` to `buildSteeringSuffix`, which drops the channel by construction, and a test pins that it stays dropped — `RewriteSteering` is threaded through whole, so a future caller adding `units` would leak every flagged bullet. Findings become possible in the same follow-up that copies the résumé text.

**A third hand-rolled clipboard call site existed** that the issue did not name: `LetterRevealDialog.tsx:97` (#715, landed after the issue was written). All three are migrated, not two.

## The clipboard extraction

`useCopyToClipboard` (logic) + `CopyButton` (the label-swap primitive) in `src/design-system/primitives/`, both exported through the barrel. Two behaviours the three call sites had already drifted on are now uniform:

- **Absence is its own branch.** `navigator.clipboard?.writeText(t)` yields `undefined` on an insecure origin (`npm run dev:http`, a workflow this repo documents), which awaits cleanly and reports a copy that never happened. Two of the three sites had that bug; `LetterRevealDialog` did not, and its fix is now everyone's.
- **A failure persists; a confirmation expires.** `resetAfterMs` clears `copied` only.

`LetterRevealDialog` migrates onto the **hook**, not the button — its failure is a full "select the text above and copy it yourself" sentence beside the button, not a label swap inside it. Both halves ship from the same module pair, so there is still one clipboard implementation in the tree.

`TextAreaField` gains `readOnly` and `autoGrow` — a `disabled` textarea's text cannot be selected with the mouse in Chromium, which makes it useless as the manual-copy fallback, and auto-growing a ~1.5KB prompt would push the dialog's own controls off-screen.

`useResumeRewrite` now assembles its steering once (`useMemo`) and exposes it, so the copied prompt cannot disagree with the button beside it; `start()`'s dep list drops from five raw inputs to that one value.

## Review focus

- `src/lib/webllm/export-prompt.ts:160` — the shape line goes *before* the steering suffix so the user's own words stay last (steering.ts's ordering rationale). Does anything downstream depend on the suffix being terminal?
- `src/lib/webllm/rewrite-guardrails.ts:88` — `keepIfAlreadyStrongRule(subject)` takes the article with the noun. Is parameterising this rule worth it versus three literals, given a future fourth caller?
- `src/design-system/primitives/CopyButton.tsx:75` — the live region renders the same node as the button label. Does a screen reader double-announce when the button also has focus?
- `src/hooks/useResumeRewrite.ts:294` — `steering` is now a `useMemo` dep of `start()`. `findings` derives from `rewriteableSections`, so identity churn in the caller's `sections` array propagates further than it did. Worth checking against `ReconstructedResume`'s render.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — 4729 passed, 8 files skipped
- [x] `npm run verify` green (fallow report-only; its two findings are the pre-existing `useSectionRewrite` / `sectionsEqual` entries)
- [x] `npm run build` green
- [ ] Manually verified in `npm run dev` — dialog → disclosure → copy → paste

New coverage: `rewrite-guardrails.test.ts` (10), `export-prompt.test.ts` (12), `CopyButton.test.tsx` (8, jsdom + stubbed clipboard), `RewritePromptDisclosure.test.tsx` (9, driven through the real controller rather than a stubbed `steering` — a prompt built once on mount passes any stubbed test and is the likely bug here).

## Provenance

Verification: `npm run verify` — green locally, and in CI

### DIFF
--- a/src/design-system/primitives/CopyButton.tsx
+++ b/src/design-system/primitives/CopyButton.tsx
@@ -13,6 +13,28 @@
  * `SkillTermGuidance`) — the label change alone reaches a sighted user and
  * nobody else — and meaning is carried by the WORD, never by colour.
  *
+ * ── The live region duplicates the label, knowingly (#732 review) ──
+ *
+ * When the button carries no `aria-label`, its visible text IS its accessible
+ * name, so a screen reader focused on it may announce the change itself — and
+ * then again from the live region. That redundancy is the deliberate choice
+ * over both alternatives:
+ *
+ *   - **Drop the region.** Accessible-name-change announcement is inconsistent
+ *     across screen readers, so this trades a possible double-announcement for
+ *     a possible SILENT one — worse for an action whose whole feedback is the
+ *     confirmation. It also breaks the one call site that is not redundant at
+ *     all: `WebGpuUnavailableNotice`'s `CopyablePath` passes a fixed
+ *     `aria-label` ("Copy chrome://flags"), so its name never changes and the
+ *     region is the ONLY channel it has.
+ *   - **Keep the name fixed and swap only the visible text.** Then the visible
+ *     label reads "Copied" while the accessible name still says "Copy prompt",
+ *     which is what WCAG 2.5.3 Label in Name exists to prevent — a speech-input
+ *     user saying "click Copied" would hit nothing.
+ *
+ * Announcing twice is the least-bad of the three. If this is ever revisited, it
+ * needs profiling against real screen readers, not a re-reading of the spec.
+ *
  * Renders a fragment, not a wrapper: every call site puts this inside a flex
  * row it already sizes, and the live region is `sr-only` (absolutely
  * positioned), so it contributes no box. Layout stays with the caller, matching
--- a/src/design-system/primitives/CopyButton.tsx
+++ b/src/design-system/primitives/CopyButton.tsx
@@ -1,0 +1,95 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * CopyButton — a {@link Button} that writes `value` to the clipboard and
+ * reports the outcome in its own label (#609).
+ *
+ * The house confirmation pattern, applied to the one action that had been
+ * hand-rolled three times: this design system ships no toast
+ * (design-system/CLAUDE.md), so a completed action confirms by swapping the
+ * content of the control the user just pressed. The swap is announced through
+ * an `sr-only` `aria-live="polite"` region (precedent: `ReconstructedSkills`,
+ * `SkillTermGuidance`) — the label change alone reaches a sighted user and
+ * nobody else — and meaning is carried by the WORD, never by colour.
+ *
+ * Renders a fragment, not a wrapper: every call site puts this inside a flex
+ * row it already sizes, and the live region is `sr-only` (absolutely
+ * positioned), so it contributes no box. Layout stays with the caller, matching
+ * `Button` / `TextAreaField`.
+ *
+ * A caller that needs the failure to render somewhere other than the label —
+ * `LetterRevealDialog` puts a full "select the text above and copy it
+ * yourself" sentence beside its button — should use {@link useCopyToClipboard}
+ * directly and keep its own markup. Both halves come from the same module pair,
+ * so there is still exactly one clipboard implementation in the tree.
+ */
+
+import type { ReactNode } from "react";
+import { Button, type ButtonSize, type ButtonVariant } from "./Button.tsx";
+import { useCopyToClipboard } from "./useCopyToClipboard.ts";
+
+interface CopyButtonProps {
+  /**
+   * The text to copy. Read at click time, so a caller may recompute it on
+   * every render (the exported rewrite prompt does — it tracks live steering).
+   */
+  value: string;
+  /** Idle label — what the button offers to do. */
+  children: ReactNode;
+  /** Label held after a successful copy. */
+  copiedLabel?: ReactNode;
+  /**
+   * Label held after a failed copy. The default states the outcome only;
+   * a caller whose text stays selectable on screen should say so here, since
+   * "select it yourself" is a real instruction rather than a shrug.
+   */
+  failedLabel?: ReactNode;
+  /** Clear the confirmation after this many ms. See `useCopyToClipboard`. */
+  resetAfterMs?: number;
+  variant?: ButtonVariant;
+  size?: ButtonSize;
+  className?: string;
+  disabled?: boolean;
+  /** Accessible name. Defaults to the button's own label text. */
+  "aria-label"?: string;
+}
+
+export function CopyButton({
+  value,
+  children,
+  copiedLabel = "Copied",
+  failedLabel = "Couldn’t copy",
+  resetAfterMs,
+  variant = "ghost",
+  size = "sm",
+  className,
+  disabled,
+  "aria-label": ariaLabel,
+}: CopyButtonProps) {
+  const { state, copy } = useCopyToClipboard(resetAfterMs);
+
+  const label =
+    state === "copied" ? copiedLabel : state === "failed" ? failedLabel : children;
+
+  return (
+    <>
+      <Button
+        variant={variant}
+        size={size}
+        className={className}
+        disabled={disabled}
+        aria-label={ariaLabel}
+        onClick={() => copy(value)}
+      >
+        {label}
+      </Button>
+      {/* Rendered unconditionally (empty while idle): a live region has to be
+          in the DOM BEFORE its content changes, or assistive tech has nothing
+          to observe and the first copy announces nothing. */}
+      <span className="sr-only" role="status" aria-live="polite">
+        {state === "idle" ? "" : label}
+      </span>
+    </>
+  );
+}
--- a/src/hooks/useResumeRewrite.ts
+++ b/src/hooks/useResumeRewrite.ts
@@ -151,6 +151,19 @@ export interface ResumeRewriteController {
   useFindings: boolean;
   /** Toggle the findings channel (persists to localStorage). */
   setUseFindings: (value: boolean) => void;
+  /**
+   * The steering `start()` would run with, right now — the SAME object, not a
+   * reconstruction (#609).
+   *
+   * The copyable-prompt disclosure has to show a prompt carrying the user's
+   * current intent, and the intent is not simply `{userInstructions,
+   * pageTarget}`: `/jd-fit` folds a JD context in front of the user's own text,
+   * and the findings channel rides the same object. Exposing the assembled
+   * value means the copied prompt cannot disagree with the button beside it —
+   * a second assembly in the UI layer would drift the first time either input
+   * changed. `undefined` when nothing is set, exactly as `start()` sees it.
+   */
+  steering: RewriteSteering | undefined;
 }
 
 export function labelForResumeRewrite(
@@ -273,6 +286,34 @@ export function useResumeRewrite(
     );
   }, [findings, rewriteableSections]);
 
+  // The one assembly of "what this rewrite is being asked to do" (#609). Built
+  // here rather than inside `start` so the copyable-prompt disclosure can show
+  // the very object the run would use — see `ResumeRewriteController.steering`.
+  const steering = useMemo<RewriteSteering | undefined>(() => {
+    // Combine the user's freeform instructions with the optional JD-driven
+    // steering (#226) into ONE userInstructions string. The JD context leads
+    // (it sets the tailoring intent); the user's own text follows so it stays
+    // the most-salient, last instruction. Both empty → no userInstructions.
+    const jd = jdContext?.trim();
+    const userText = userInstructions.trim();
+    const combinedInstructions = [jd, userText]
+      .filter((s): s is string => !!s)
+      .join("\n\n");
+    // The app's own findings (#608) ride the SAME steering channel rather than
+    // a fourth parameter on `rewriteResumeWithLlm` — one intent channel, per
+    // #608's reuse analysis. Gated on the user's toggle, so opting out restores
+    // the pre-#608 prompt exactly.
+    const activeFindings = useFindings ? findings : undefined;
+    if (!combinedInstructions && pageTarget === null && !activeFindings) {
+      return undefined;
+    }
+    return {
+      ...(combinedInstructions ? { userInstructions: combinedInstructions } : {}),
+      ...(pageTarget !== null ? { pageTarget } : {}),
+      ...(activeFindings ? { findings: activeFindings } : {}),
+    };
+  }, [jdContext, userInstructions, pageTarget, useFindings, findings]);
+
   useEffect(() => {
     let cancelled = false;
     void detectWebGpu().then((c) => {
@@ -312,30 +353,6 @@ export function useResumeRewrite(
       const engine = await loadEngine(modelId, (progress) => {
         setStatus({ kind: "loading", progress });
       });
-      // Combine the user's freeform instructions with the optional JD-driven
-      // steering (#226) into ONE userInstructions string. The JD context leads
-      // (it sets the tailoring intent); the user's own text follows so it stays
-      // the most-salient, last instruction. Both empty → no userInstructions.
-      const jd = jdContext?.trim();
-      const userText = userInstructions.trim();
-      const combinedInstructions = [jd, userText]
-        .filter((s): s is string => !!s)
-        .join("\n\n");
-      // The app's own findings (#608) ride the SAME steering channel rather
-      // than a fourth parameter on `rewriteResumeWithLlm` — one intent channel,
-      // per the issue's reuse analysis. Gated on the user's toggle, so opting
-      // out restores the pre-#608 prompt exactly.
-      const activeFindings = useFindings ? findings : undefined;
-      const steering: RewriteSteering | undefined =
-        combinedInstructions || pageTarget !== null || activeFindings
-          ? {
-              ...(combinedInstructions
-                ? { userInstructions: combinedInstructions }
-                : {}),
-              ...(pageTarget !== null ? { pageTarget } : {}),
-              ...(activeFindings ? { findings: activeFindings } : {}),
-            }
-          : undefined;
       const result = await rewriteResumeWithLlm(
         rewriteableSections,
         engine,
@@ -367,22 +384,13 @@ export function useResumeRewrite(
       releaseInference(modelId);
       release();
     }
-    // `findings` + `useFindings` are read inside, so both are deps — the lint
-    // plugin is not registered in this repo (CLAUDE.md → Data & hooks), so a
-    // stale closure here would lint green and silently ignore the toggle until
-    // some unrelated dep changed. `findings` is a `useMemo` over
-    // `[critique, rewriteableSections]`, both already stable across renders
-    // that don't change them, so this adds no re-creation of its own.
-  }, [
-    acquire,
-    rewriteableSections,
-    selectedModelId,
-    userInstructions,
-    pageTarget,
-    jdContext,
-    findings,
-    useFindings,
-  ]);
+    // `steering` replaces the five inputs this callback used to assemble
+    // inline (#609). The lint plugin is not registered in this repo (CLAUDE.md
+    // → Data & hooks), so a stale closure here would lint green and silently
+    // run with last render's intent — reading ONE memoized value instead of
+    // five raw ones is one dep to get wrong instead of five, and it is the same
+    // value the disclosure shows the user.
+  }, [acquire, rewriteableSections, selectedModelId, steering]);
 
   const dismiss = useCallback(() => {
     setStatus({ kind: "idle" });
@@ -434,6 +442,7 @@ export function useResumeRewrite(
     hasFindings,
     useFindings,
     setUseFindings,
+    steering,
   };
 }
 
--- a/src/hooks/useResumeRewrite.ts
+++ b/src/hooks/useResumeRewrite.ts
@@ -258,9 +258,8 @@ export function useResumeRewrite(
     [setPageTargetRaw],
   );
 
-  const rewriteableSections = useMemo(
-    () => sections.filter(isNonEmptyForUi),
-    [sections],
+  const rewriteableSections = useStableSections(
+    useMemo(() => sections.filter(isNonEmptyForUi), [sections]),
   );
 
   // The critique's findings, keyed by the line they describe (#608). Built off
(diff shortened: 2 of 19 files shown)


src/design-system/primitives/CopyButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+        disabled={disabled}
+        aria-label={ariaLabel}
+        onClick={() => copy(value)}
+      >
+        {label}
+      </Button>
+      {/* Rendered unconditionally (empty while idle): a live region has to be
+          in the DOM BEFORE its content changes, or assistive tech has nothing
+          to observe and the first copy announces nothing. */}
+      <span className="sr-only" role="status" aria-live="polite">
```
s-annam, line 112: Note on screen reader behavior: When the button has focus and its text changes (e.g. from 'Copy prompt' to 'Copied the prompt'), some screen readers will announce the button's content update natively. Rendering the same string in a `role="status"` region might lead to a double announcement. While this is acceptable here, in the future we could consider a quieter status region or profiling it with actual screen readers.

src/hooks/useResumeRewrite.ts
```diff
@@ -273,6 +286,34 @@ export function useResumeRewrite(
     );
   }, [findings, rewriteableSections]);
 
+  // The one assembly of "what this rewrite is being asked to do" (#609). Built
+  // here rather than inside `start` so the copyable-prompt disclosure can show
+  // the very object the run would use — see `ResumeRewriteController.steering`.
+  const steering = useMemo<RewriteSteering | undefined>(() => {
+    // Combine the user's freeform instructions with the optional JD-driven
```
s-annam, line 292: Note on identity stability: Since `sections` changes identity on every render of `ReconstructedResume`, `rewriteableSections` and `steering` will also recreate on every render. Because `start` is a callback only triggered on click, this recreation is completely harmless and performs nicely. However, if a future feature starts watching `steering` directly on render, we may want to wrap `sections` or `rewriteableSections` in a custom hook that does a deep check via `sectionsEqual` to stabilize its reference.

s-annam (dismissed):
This PR is a stellar, high-fidelity implementation of #609. The extraction of clipboard logic into the reusable primitives (`CopyButton`, `useCopyToClipboard`) is clean and cohesive. Unifying the guardrails under `rewrite-guardrails.ts` ensures prompt consistency without drifting, and the tests covering prompt content absence and updates are exceptionally thorough.

### Findings

#### ⚠️ Nits (Non-blocking)

* **Accessibility Announcement Redundancy** ([CopyButton.tsx:90](file:///C:/Users/annam/offlinecv/src/design-system/primitives/CopyButton.tsx#L90-L92)):
  Since the button itself swaps text label in place upon copy/failure, some screen readers will already announce this label change when the button has focus. Adding a `role="status"` live region that mirrors the exact same label might cause a double-announcement in some screen readers. While acceptable for now, profiling it or simplifying it in the future could be beneficial.

* **Hook Identity Stability** ([useResumeRewrite.ts:293](file:///C:/Users/annam/offlinecv/src/hooks/useResumeRewrite.ts#L293-L315)):
  Since the parent component `ReconstructedResume` builds `resumeSections` on every render, the `sections` array changes identity. This propagates identity churn to `rewriteableSections` and `steering` in `useResumeRewrite`. Because `start` is a callback invoked only on click, this causes no render loops or side effects. However, if a future caller begins relying on `steering` directly on render, we may want to wrap `sections` in a deep-equality stabilizer using `sectionsEqual`.

---

src/hooks/useResumeRewrite.ts
Samhit21: Fixed in `cbf1f9a`.

You are right, and the future you flag is already here: `RewritePromptDisclosure` (added in this PR) reads `steering` on render through a `useMemo`, so with both deps churning that memo was memoizing nothing — it rebuilt the prompt string on every render of the disclosure. Harmless in output (everything downstream is derived) but a claim the code wasn't keeping.

`useStableSections` now holds the previous array whenever the new one says the same thing, so `rewriteableSections`, `findings`, `steering` and `start` all stop churning at the source.

One deviation from your suggestion worth calling out: it compares `label` **in addition to** `sectionsEqual`, rather than using `sectionsEqual` alone. `sectionsEqual` ignores `label` — correct where it is used, since a stale-proposal guard shouldn't discard a live proposal because the user retyped an employer name. But the identity being gated here feeds what the user *sees*: the progress line (`Rewriting 2 of 5: Engineer — Acme`) and every heading in the proposal panel read `label` off these objects, so holding an array with stale labels would show the old employer against the new bullets. A test pins that direction — it fails if the comparator is weakened back to bare `sectionsEqual`.

It adjusts state during render rather than writing a ref (React's documented shape for a props-derived value that has to stay stable), so no effect ever observes the stale array.

src/design-system/primitives/CopyButton.tsx
Samhit21: Real concern, and I looked hard at both alternatives — but I'm keeping the duplication and recording why in the docblock (`cbf1f9a`) rather than changing behaviour. Leaving this thread open for you.

The redundancy only exists where the button carries no `aria-label`. The two ways out both cost more than they save:

- **Drop the live region.** Accessible-name-change announcement is inconsistent across screen readers, so this trades a *possible* double-announcement for a *possible* silent one — worse for an action whose entire feedback is the confirmation. It also mutes the one call site that has no redundancy at all: `WebGpuUnavailableNotice`'s `CopyablePath` passes a fixed `aria-label` ("Copy chrome://flags"), so its accessible name never changes and the region is its only channel.
- **Pin the accessible name and swap only the visible text** (GitHub's `clipboard-copy` shape). Then the visible label reads "Copied" while the name still says "Copy prompt" — which is what WCAG 2.5.3 Label in Name exists to prevent; a speech-input user saying "click Copied" hits nothing.

So: announcing twice is the least-bad of the three, and it's what design-system/CLAUDE.md already prescribes ("announce the swap with `aria-live=\"polite\"`"). Your "profile it with actual screen readers" is the right next step — I can't settle it from the spec, and I'd rather not guess at it in this PR. Happy to file a follow-up to profile NVDA/VoiceOver against the three call sites if you want it tracked.

s-annam (approved):
# PR Review Draft — PR #732

We have conducted a maintainer-grade review of pull request **#732** in `offlinecv/OfflineCV` and verified all implementation details, test suites, and gates. Below is the proposed review.

---

### **Review Stance**
Looks good! The implementation of the copyable rewrite prompt is extremely clean, maintains complete privacy (instructions-only prompt, no resume text leaks), and successfully unifies the codebase's clipboard logic onto a single, robust Hook.

---

## **Acceptance Criteria Checklist (Issue #609)**

All acceptance criteria for the linked issue [GH-609](https://github.com/offlinecv/OfflineCV/issues/609) have been fully met:

- [x] **Primary Path Unchanged:** The rewrite dialog's primary path is untouched. Users who do not open the disclosure see the same UI as before.
- [x] **Not Visible by Default:** The prompt is collapsed and hidden behind a link disclosure.
- [x] **Shared Guardrails:** The copyable prompt lists the number-preservation and no-fabrication guardrails directly from the shared constants.
- [x] **Byte-Identical Prompts:** Pre-refactor prompt strings for on-device rewrite are verified as byte-identical via literal checks in `rewrite-guardrails.test.ts`.
- [x] **Live Steering Updates:** Changing the page target or typing freeform instructions dynamically updates the copyable prompt text (covered in `RewritePromptDisclosure.test.tsx`).
- [x] **Complete with No Steering:** When no steering parameters are set, the prompt is complete and cleanly formatted (no dangling headings/scaffolding).
- [x] **Aria-Live Announcement:** Copying uses the status live region for a polite `aria-live` confirmation (no toasts, no color-only indicator).
- [x] **No Clipboard Degradation:** On insecure origins or environments where `navigator.clipboard` is absent, the text remains selectable so users can manually copy.
- [x] **No Résumé Content:** The prompt is instructions-only and contains no text derived from the résumé (except shape counts for context scale).
- [x] **Unified Clipboard logic:** `useCopyToClipboard` and `CopyButton` are extracted to `@design-system`. Both pre-existing hand-rolled locations (`WebGpuUnavailableNotice.tsx` and `SectionRewrite.tsx`) and the recently added `LetterRevealDialog.tsx` have been migrated to the new primitives.
- [x] **Decomposition and LOC Limits:** Files remain focused and modular; `RewritePromptDisclosure.tsx` is only 109 LOC.
- [x] **No Style/Lint/Build issues:** Clean typecheck, lint, build, and zero dead code.

---

## **Description Accuracy (Gate 3f)**
**Stance:** Accurate.
The description is extremely detailed, correctly details the three design decisions requested, explicitly names deviations (why #608 findings are not carried and the extra `LetterRevealDialog` migration), and outlines the refactoring accurately.

---

## **Gate Audits**

- **Gate 3a (Fixture PII):** *Skipped* (no fixtures added or modified in this PR).
- **Gate 3b (Design System & Reuse):** *Passed* (components reuse `@design-system` primitives, no raw buttons or textareas are introduced).
- **Gate 3c (Style Tokens):** *Passed* (all style tokens are standard semantic classes, no hardcoded colors or raw Tailwind palettes).
- **Gate 3d (Dead-code / Fallow):** *Passed* (no unused exports; dead-code audit reports 0% dead exports).
- **Gate 3e (Command Bugs):** *Skipped* (no shell scripts, git hooks, or skill instructions modified).

---

## **Review Findings & Verdict**

- **Blocking:** 0
- **Secondary:** 0
- **Nits:** 0

**Verdict: `APPROVE`**
*(Rule: 0 Blocking findings → APPROVE)*

---

### **Review Focus Items Triaged**

- **Terminal steering suffix (`export-prompt.ts`):** Having the steering suffix terminal is correct. External LLMs place higher attention on text at the beginning and the end of the context, making terminal placement of the user's intent optimal.
- **`keepIfAlreadyStrongRule(subject)` parameterization:** The parameterization is well justified. The subjects require different articles ("a bullet", "the summary"), so passing the full phrase is cleaner than complex grammatic rules.
- **Aria-live region duplication (`CopyButton.tsx`):** Since the live region has `role="status"` and `aria-live="polite"` and is only populated when `state !== "idle"`, it behaves correctly. Double-announcement is a standard and safe trade-off to ensure non-visual visibility.
- **Identity churn stabilizer (`useResumeRewrite.ts`):** The stabilizer `useStableSections` correctly halts referential churn from the parent component, successfully keeping the memoized `steering` stable across parent renders.

---
Reviewed by: Gemini 1.5 Pro (high)
Head SHA reviewed: `cbf1f9a1993a65c4886fe9896c6b3e8bae452d4e`
 👀 reaction start-signal posted: Yes (at Step 0.6)

PR #733, opened 2026-08-03
fix(parser): strip inline markdown images from rawText, not just the PdfLine reading (#613)

### LINKED ISSUE #613: Parser: mdToPlainText flattens an inline image to `!alt url` in rawText, which the evidence panel prints
## Summary

`mdToPlainText` has no image-stripping rule, so a markdown **inline image** — `alt` — is flattened by the generic link rule into `!alt url`. That string lands in `rawText`, which `EvidencePanel` prints back to the user verbatim, and which the scorer scans.

The `PdfLine` reading of the same file strips the image entirely. So the two readings of one dropped `.md` disagree on images.

This is **pre-existing**, not a regression: the unguarded rewrite shipped with the `.md` import path in `#552`. It is filed now because `#610`/`#611` (PR `#612`) aligned the two readings on every *link* shape and left images as the one remaining divergence.

## Reproduction

```markdown
## Summary

Shipped ![Company logo](https://example.org/logo.png) fast.
```

Run both readings over that input:

```
rawText : "Summary\n\nShipped !Company logo https://example.org/logo.png fast.\n"
mdLines : ["Summary", "Shipped  fast."]
```

**Expected** — the image contributes nothing to either reading, as `#610`'s acceptance criterion already requires for the `PdfLine` side ("markdown images still strip entirely — an `alt` must not start flattening to `alt url`").

**Actual** — `rawText` gains a stray `!`, the alt text, and an image URL that is not résumé content. The user sees it in the evidence panel.

## Root cause

`src/lib/ingest/markdown.ts` rewrites links with

```ts
.replace(/\[([^\]]*)\]\(\s*([^)\s]+)(?:\s+"[^"]*")?\s*\)/g, "$1 $2")
```

There is no `!` guard and no image pass, so the image's `[alt](url)` tail matches the link rule and the leading `!` is simply left behind. `src/lib/heuristics/markdown-lines.ts` does not have this problem because `stripImages` runs before `flattenLinks` and removes `ANY_IMAGE_RE` / `DATA_URI_IMAGE_RE` outright.

The same asymmetry existed for reference-style images until `#611`, which fixed it on both sides via `stripReferenceImages` in the shared `src/lib/markdown-link-refs.ts`. Inline images are the shape that pass does not cover.

## Proposed fix

Give `mdToPlainText` the same image pass the `PdfLine` reading already has, rather than adding a second, differently-worded rule. Both readings should reach it through the shared module, which is the invariant `#610`/`#611` established: one definition, two consumers.

Note the data-URI form (`alt`) — a base64 payload flattened into `rawText` is the more damaging half, since it can be very large and is scored as text.

## Acceptance criteria

- [ ] `alt` contributes nothing to `rawText` — no `!`, no alt text, no URL.
- [ ] The data-URI image form is stripped from `rawText` too, not just the http(s) form.
- [ ] `sectionizeMarkdown` and `mdToPlainText` agree on a document containing an inline image — assert it in `src/lib/markdown-readings-agree.test.ts`, which currently compares the two readings line-for-line but sees no image because neither markdown fixture contains one.
- [ ] A **link** whose label merely ends in `!` (`Fastlabel` is an image; `[Ship it!](https://…)` is not) still flattens to `label url`.
- [ ] Reference-style images continue to behave as `#611` left them: a defined one leaks no URL, an undefined one stays literal text.
- [ ] Add an image to one of the `tests/fixtures/markdown/` fixtures, or add a third fixture, so the agreement test actually exercises the shape.
- [ ] `npm run test` and `npm run verify` pass; no corpus snapshot re-bake without per-fixture justification.

## Test fixture

**PII — hard rule.** Synthetic persona only: fake name, `@example.com` email, and a phone with a real area code + `555` exchange + `0100`–`0199` subscriber (e.g. `(312) 555-0123`). `npm run check:fixtures` only scans PDFs under `tests/fixtures/pdfs/`, so a `.md` fixture is **not** covered by that gate — the judgement is manual. See `tests/fixtures/markdown/README.md`.

**Prove the fixture guards.** Confirm the new assertion fails before the fix; an assertion added after it passes vacuously.

## Related

| Issue | Relationship |
|---|---|
| #552 | Introduced the `.md` import path and this unguarded rewrite. |
| #610 | Aligned the two readings on inline links and autolinks; its "images must not flatten" criterion is met on the `PdfLine` side only. |
| #611 | Fixed the same class for *reference-style* images in both readings, via `stripReferenceImages`. Inline images are the remaining gap. |

Samhit21:
## Summary

`mdToPlainText` had no image-stripping rule, so a markdown inline image — `alt` — was flattened by the generic link rule into `!alt url`. That string landed in `rawText`, which `EvidencePanel` prints back to the user verbatim and the scorer scans. The `PdfLine` reading of the same file stripped the image outright, so the two readings of one dropped `.md` disagreed on images — the last shape left over after #610/#611 aligned every *link* shape.

Fixed by moving `stripImages` out of `markdown-lines.ts` into the shared `markdown-link-refs.ts` as `stripInlineImages`, and calling it from both readings. One definition, two consumers — the invariant #610/#611 established — rather than a second, differently-worded rule.

Closes #613

## Before / after

```
## Summary

Shipped ![Company logo](https://example.org/logo.png) fast.
```

| | before | after |
|---|---|---|
| `rawText` | `Summary\n\nShipped !Company logo https://example.org/logo.png fast.\n` | `Summary\n\nShipped  fast.\n` |
| `mdLines` | `["Summary", "Shipped  fast."]` | unchanged |

The data-URI form was the damaging half — `Badge !b data:image/png;base64,… here.` — a base64 payload flattened into `rawText` and scored as text.

## Two implementation notes

**The strip runs over the whole document, before the per-line map.** `mdToPlainText` otherwise works line-by-line, but turndown occasionally wraps a long data URI across a line break, and neither image regex excludes newlines for exactly that reason. `markdown-lines.ts` already strips at document level, so per-line stripping here would have put the two readings back out of step on the shape this fixes. A test pins the wrapped-data-URI case.

**Inline and reference images stay two functions.** `alt` carries its own target, so it is an image on sight and can be stripped anywhere in the pipeline. `![great]` is an image only if something defines `[great]` — which is why `stripReferenceImages` takes the definition table and #611's behaviour is untouched here: a defined reference-style image still leaks no URL, an undefined one is still literal text.

## The fixture

`inline-links.md` gains both forms: a base64 headshot in the profile band (mammoth+turndown's output for an embedded image) and an https badge sitting **immediately after a real link** in Projects. The badge's placement is the load-bearing part — a fix that over-claims the `!` would take the adjacent link with it, and a separate assertion holds that `ledger-toolkit https://example.org/ledger-toolkit` still flattens.

Persona unchanged and synthetic (Riley Nakamura, `@example.com`, `(312) 555-0123`). The added image URLs are `example.org`; the data URI is a 1×1 transparent PNG. Per `tests/fixtures/markdown/README.md`, `npm run check:fixtures` does not cover this directory — the judgement is manual and was made.

## The assertions were proven to fail first

`mdToPlainText` was reverted to its pre-fix form with the new tests and fixture in place. All **eight** new assertions failed, including the pre-existing line-for-line agreement test — which had been passing only because neither fixture contained an image:

```
× produces identical lines from both readings
  → expected [ 'Riley Nakamura', …(20) ] to deeply equal [ 'Riley Nakamura', …(19) ]
× leaks neither image URL, nor its alt text, nor a stray `!`
  → expected '!Headshot data:image/png;base64,iVBOR…' not to contain 'data:image/png;base64'
× https image leaves only the surrounding prose
  → expected 'Shipped !Company logo https://example…' not to contain 'example.org'
× keeps the text an image runs straight into
  → expected 'Fast!label https://example.org/x.png' to be 'Fast'
```

Every case asserts **both** readings, not just `rawText`: the defect was invisible on the `PdfLine` side, so a one-sided test would have passed against the bug.

## Acceptance criteria

- [x] `alt` contributes nothing to `rawText` — no `!`, no alt text, no URL
- [x] The data-URI form is stripped from `rawText` too
- [x] `sectionizeMarkdown` and `mdToPlainText` agree on a document containing an inline image, asserted in `markdown-readings-agree.test.ts`
- [x] `[Ship it!](https://…)` still flattens to `label url`; `Fastlabel` keeps the text and drops the image
- [x] Reference-style images behave as #611 left them — the existing describes are untouched and green
- [x] An image added to `tests/fixtures/markdown/inline-links.md`, in both forms
- [x] No corpus snapshot re-bake — `git status` after `npm run verify` shows no `.expected.json` touched

## Review focus

- `src/lib/markdown-link-refs.ts:132` — `ANY_IMAGE_RE`'s target class is `[^)]*`, so a `)`-bearing image URL truncates. That is the pre-existing behaviour moved verbatim, and it matches `INLINE_LINK_RE`'s own limits — but is a percent-encoded `%29` worth admitting now that this rule is shared?
- `src/lib/ingest/markdown.ts:71` — the strip runs *after* `extractLinkDefinitions`. Can an image's data URI ever contain something the definition-line regex claims first?

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — 4699 passed, 8 files skipped
- [x] `npm run verify` green
- [x] New assertions confirmed failing before the fix (above)
- [ ] Manually verified by dropping a `.md` with an image in `npm run dev`

## Provenance

Verification: `npm run verify` — green locally, and in CI

### DIFF
--- a/src/lib/heuristics/markdown-lines.ts
+++ b/src/lib/heuristics/markdown-lines.ts
@@ -13,9 +13,9 @@
  * while keeping the adapter small enough to audit.
  *
  * Pre-cleaning pass (real-world DOCX artifacts):
- *   - Strip base64 data URIs in `![...](data:...)` — mammoth+turndown emits
- *     embedded images this way and they can bloat markdown 3× with zero
- *     signal for parsing.
+ *   - Strip inline images `![alt](url)`, including the base64 data URIs
+ *     mammoth+turndown emits for embedded images — they can bloat markdown 3×
+ *     with zero signal for parsing. Shared with `mdToPlainText` since #613.
  *   - Flatten inline links `[label](url)` and autolinks `<url>` to `label url`
  *     (#610). Without this the bracket-paren syntax reaches every extractor
  *     verbatim, so a `.md` renders literal markdown in the exported PDF and
@@ -52,6 +52,7 @@ import {
   extractLinkDefinitions,
   flattenAutolinks,
   resolveReferenceLinks,
+  stripInlineImages,
   stripReferenceImages,
 } from "../markdown-link-refs.ts";
 import type { PdfLine, PdfSection } from "./line-model.ts";
@@ -88,17 +89,15 @@ const INLINE_BOLD_RE = /\*\*(.+?)\*\*/g;
 const INLINE_ITALIC_UNDERSCORE_RE = /(^|[^\w\\])_([^\s_][^_]*?[^\s_]|[^\s_])_(?=[^\w]|$)/g;
 const INLINE_ITALIC_STAR_RE = /(^|[^\w\\*])\*([^\s*][^*]*?[^\s*]|[^\s*])\*(?=[^\w*]|$)/g;
 
-// Base64 data-URI images in markdown: ![alt](data:image/...;base64,....)
-// Multi-line friendly because turndown occasionally wraps.
-const DATA_URI_IMAGE_RE = /!\[[^\]]*\]\(data:[^)]*\)/g;
-
-// Non-data markdown images: `![alt](https://...)` and `![](path.png)`. Alt
-// text alone isn't useful for resume parsing — drop the whole image ref so
-// it doesn't pollute the line text.
-const ANY_IMAGE_RE = /!\[[^\]]*\]\([^)]*\)/g;
+// Inline images — `![alt](https://…)`, `![](path.png)` and the base64
+// data-URI form — are stripped by the shared `stripInlineImages` in
+// `../markdown-link-refs.ts` (see the import above). The rule lives there so
+// `mdToPlainText` applies the identical one to `rawText`: before #613 it had no
+// image rule at all, and the image's `[alt](url)` tail matched its generic LINK
+// rule and flattened to `!alt url` in the text the Evidence panel prints.
 
 // Reference-style images (`![alt][ref]`, `![alt][]`, `![alt]`) cannot be
-// stripped here: unlike the two shapes above they are not self-identifying —
+// stripped there: unlike the inline shapes above they are not self-identifying —
 // whether `![great]` is an image or literal prose depends on the document's
 // `[ref]: url` definition table, which does not exist yet at this point in the
 // pipeline. They are handled inside `flattenLinks` via the shared
@@ -123,13 +122,6 @@ const INLINE_LINK_RE = /\[([^\]\n]*)\]\(\s*([^)\s]+)(?:\s+"[^"]*")?\s*\)/g;
 // escapes that appear in contact data / prose. We keep `\n` as-is.
 const BACKSLASH_ESCAPE_RE = /\\([_*[\]()!`#>.+-])/g;
 
-/** Strip base64 image blobs and non-data image refs from markdown. */
-function stripImages(markdown: string): string {
-  return markdown
-    .replace(DATA_URI_IMAGE_RE, "")
-    .replace(ANY_IMAGE_RE, "");
-}
-
 /**
  * Flatten every markdown link shape to its plain-text reading — inline links
  * and autolinks (#610), then reference-style links (#611).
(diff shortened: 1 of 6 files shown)


s-annam (approved):
This PR is exceptionally well-authored, with complete and robust alignment of markdown inline images across both reading pipelines (`mdToPlainText` and `sectionizeMarkdown`). It addresses all acceptance criteria perfectly and has excellent test coverage.

- [x] `alt` contributes nothing to `rawText` — no `!`, no alt text, no URL (Satisfied)
- [x] The data-URI image form is stripped from `rawText` too, not just the http(s) form (Satisfied)
- [x] `sectionizeMarkdown` and `mdToPlainText` agree on a document containing an inline image — asserted in [markdown-readings-agree.test.ts](file:///C:/Users/annam/offlinecv/src/lib/markdown-readings-agree.test.ts) (Satisfied)
- [x] A **link** whose label merely ends in `!` (`Fastlabel` is an image; `[Ship it!](https://…)` is not) still flattens to `label url` (Satisfied)
- [x] Reference-style images continue to behave as #611 left them (Satisfied)
- [x] Add an image to one of the `tests/fixtures/markdown/` fixtures to exercise the agreement test (Satisfied)
- [x] `npm run test` and `npm run verify` pass; no corpus snapshot re-bake (Satisfied)

### Description Accuracy
- **Verdict:** Accurate. The PR summary and detail notes are exceptionally clear and precisely match the implementation and test findings.

### Review Focus Answers
1. **Parenthesis-bearing URLs (`ANY_IMAGE_RE`):** CommonMark syntax requires any raw parentheses in link targets to be escaped or percent-encoded (`%29`). Therefore, `[^)]*` is technically correct and matches the standard parser's limits. No change is needed.
2. **Order of `extractLinkDefinitions` and `stripInlineImages`:** The base64 alphabet `[A-Za-z0-9+/=]` does not contain brackets `[` or `]`, or colons `:`. Therefore, it is impossible for a wrapped base64 line of a data URI to start with `^ {0,3}\[([^\]\n]+)\]:` to match the link definition regex. Running the inline image strip after `extractLinkDefinitions` is completely safe.

### Verdict Rule
0 Blocking findings -> **APPROVE**

---
Reviewed SHA: 67d974221ec1c01d6587cd2173600552406fead8

Issue #711, opened 2026-08-01
[storage] letters store — a cover letter as a first-class record, writable by any producer


s-annam:
## Why

The cover letter is the next per-job artifact. Today there is nowhere to put one.

Most of what a letter needs is already persisted, so this is smaller than it looks:

- DB `offlinecv` v2, stores `resumes` / `jobs` / `boards` — `src/lib/storage/db.ts:26`
- `JobRecord.jdText` and `JobRecord.resumeId` — `src/lib/storage/types.ts:102-106`. The JD
  text and the résumé link are already on the record.
- `ResumeRecord.parse` holds a `SavedResumeSnapshot` = `{ result: CascadeResult, score,
  sourceKind, shapeVersion }` — `src/lib/resume-library.ts:126`. The structured parse is
  cached, so nothing has to re-run the cascade to write a letter.

What is missing is the letter.

## The decision this issue encodes: storage, not generation

Generation is deliberately **out of scope**. On-device long-form prose quality is unproven
here, and pinning the schema to one engine would be backwards. Define the artifact so that
*any* producer can write it:

- a Claude Code skill driving `offlinecv.org` through browser automation — it runs JS in the
  page's own origin, so it can read both stores and write back
- the browser extension, later
- an in-app generator, whenever on-device quality justifies one
- a hosted tier, if that path is ever taken

This mirrors a pattern the repo already has: `src/lib/storage/job-record-contract.ts` plus
`docs/job-capture-contract.md` are normative for third-party job producers. Letters get the
same treatment.

## Why a separate store rather than a field on `JobRecord`

- **Letters are iterated.** Several drafts per job, and versions of one draft, are the normal
  case. A field gives you exactly one, forever.
- **`JobRecord` is a validated contract.** It is checked by `validateJobRecord` and
  JSON-safety-checked at the export boundary by `findJsonSafetyProblem`. Widening it with a
  prose blob widens the contract the extension validates against, for something the extension
  does not produce.
- **Additive migration is the established pattern.** `boards` arrived at v1→v2 exactly this
  way (`db.ts:53`), leaving existing resumes and jobs untouched.

## Scope

1. **`LetterRecord extends StoredRecord`** in `storage/types.ts`:
   - `jobId: string` — the job this letter is for
   - `resumeId?: string` — the résumé it was written from. Cleared, not orphaned, if that
     résumé is deleted (same rule as `JobRecord.resumeId`).
   - `body: string` — the letter itself, markdown
[...]

PR #748, opened 2026-08-04
feat(job-tracker): group by lifecycle bucket, record origin, and merge duplicate postings (#744, #745, #746)

### LINKED ISSUE #744: Job tracker splits synced records into parallel sections — group by lifecycle bucket, not raw status
## Problem

A synced or imported record can carry a `status` this build's lifecycle does not
know, and the tracker turns each such string into its **own top-level section**.
A real library ends up with parallel sections — `Saved`, `Scouted`, `Shared`,
`Interested` — that the user reads as four different things when they are one:
jobs on the list, not yet applied to.

Two mechanisms, both working as written:

- `src/lib/storage/job-record-contract.ts:120` — `isStatusLike` accepts **any**
  string. That is deliberate (`:98` says so) and should stay: refusing an
  unknown status would lose the record.
- `src/components/features/JobTracker.tsx:109-124` — grouping buckets on the
  record's **literal** status string, orders the six in `JOB_STATUS_ORDER`
  first, then appends every unknown status sorted alphabetically.

So the contract's tolerance and the tracker's grouping compose into a section
per vocabulary term rather than a section per stage of a job search.

Concretely, a library synced from Recruidea sees four extra buckets. Recruidea's
`saved_jobs.status` vocabulary is a superset of this build's: it adds `scouted`,
`shared`, `saved` and `withdrawn` to the six in
`src/lib/storage/types.ts:255`. The first three are all pre-application states —
they differ by *how the job arrived* (a job alert, a sponsor share, a manual
save), not by where it is in the user's pipeline. In a dogfooding library the
largest section by an order of magnitude is `saved`, sitting beside a small
`interested` with nothing on screen saying how they differ.

### The status picker has the matching hole

`src/components/features/JobStatusPicker.tsx:56-74` renders exactly
`JOB_STATUS_ORDER` and marks the current one with `status === value`. A row
whose status is `shared` matches nothing, so the picker renders with **no
selection at all** — six ghost buttons and no indication of where the job
currently is.

`jobStatusLabel` / `jobStatusTone` already fall back for an unknown status
(`:36-48`), so the row *badge* is fine. It is the grouping and the picker that
have no answer.

## Proposal

**Map at view time. Never rewrite the stored status.**

This is the load-bearing constraint and it is worth stating plainly: a record's
`status` round-trips to whatever produced it. A local rewrite of `shared` →
`interested` is not a display change, it is a write that destroys the meaning
the producing system attached to that row. The tracker's job is to *group*
these, not to normalise them.

### A. One display vocabulary

New pure module, `src/lib/job-status-bucket.ts` (zero-dep, testable at module
scope, sibling of `job-tracker.ts`):

| Stored status | Display bucket |
|---|---|
| `interested`, `saved`, `scouted`, `shared` | **Interested** |
| `applied` | Applied |
| `interviewing` | Interviewing |
| `offer` | Offer |
| `rejected`, `withdrawn` | Rejected |
| `archived` | Archived |
| anything else | its own bucket, labelled with the literal string |

The last row keeps today's fail-loud behaviour for a genuinely unrecognised
status, which `JobTracker.tsx:104-108` and
`JobTrackerStatusGroup.tsx:60-64` both deliberately preserve. This change
narrows *what counts as unrecognised*; it does not remove the escape hatch.

`isKnownStatus` (`job-record-contract.ts:299`) exists and has no caller in this
build. It is the natural predicate for "did this map, or is it passing through".

### B. Grouping and the picker read the bucket

- `JobTracker.tsx` buckets on `bucketOf(job.status)` rather than `job.status`.
- `JobStatusPicker` compares `bucketOf(value)` when deciding which button is
  primary, so a `shared` row shows **Interested** selected.
- Clicking the bucket a row is already in is a no-op — it must not write
  `interested` over `shared`.
- Clicking a different bucket writes that bucket's canonical status
  (`applied`, `interviewing`, …), which is the normal transition and is
  expected to sync back.

### C. Say what got merged, without a new section

A row whose stored status is not the bucket's canonical one gets a small chip
carrying the literal status (`Shared`, `Scouted`, `Saved`). Nothing is hidden;
it stops being a *structural* division and becomes a per-row fact.

This is the weakest part of the proposal and reasonable people will disagree:
the chip could equally be dropped, since the badge already renders the literal
status via `jobStatusLabel`. Decide in review.

## Out of scope

- Any change to `JOB_STATUS_ORDER` or to `JobStatus`. The stored vocabulary is
  a contract; this issue is about how six-plus-N statuses are *shown*.
- Any migration or backfill of existing records.
- Provenance ("this came from a job alert" vs "you captured this"), which
  the record has no field for at all. Separate issue.

## Acceptance criteria

- [ ] A library holding records with statuses `interested`, `saved`, `scouted`
      and `shared` renders **one** Interested section containing all of them.
- [ ] A record's stored `status` is byte-identical before and after the tracker
      renders it, and after the user clicks the bucket it is already in
      (assert on the record, not on the DOM).
- [ ] A `shared` record's status picker shows **Interested** as the selected
      button.
- [ ] Moving a `shared` record to Applied writes `applied`.
- [ ] A record with a status outside both vocabularies still gets its own
      section, expanded, labelled with the literal string — the existing
      `JobTrackerStatusGroup` behaviour is unchanged for it.
- [ ] The header total continues to equal the sum of every section's count.
- [ ] The mapping lives in one module with its own unit tests; no second copy
      of the vocabulary in a component.

### LINKED ISSUE #745: JobRecord has no provenance field — the tracker cannot say where a row came from
## Problem

A `JobRecord` cannot say **how it got here**. The tracker shows a row the user
captured from a posting, a row that arrived from a job-alert ingest, and a row
someone shared with them identically — same badge, same controls, no way to tell
one from another.

`JobCaptureProvenance` (`src/lib/storage/types.ts:162-173`) is close but
answers a different question. It records *which producer wrote the record and
against which contract version* — `contract`, `producer`, `producerVersion`,
`capturedAt`. That is a compatibility fact, aimed at a future reader deciding
whether it can trust the fields. It is not a user-facing origin, and it is
absent entirely on records this build writes itself.

The gap shows up the moment sections get merged (#744): once `saved`, `scouted`
and `shared` all
render as **Interested**, the only thing distinguishing them is the literal
status string — a term from another system's vocabulary, which is exactly the
jargon the merge was meant to stop showing.

## Proposal

One optional field on `JobRecord`, from a closed vocabulary:

```ts
/** Where this record came from, for display only. Never affects merge,
 *  dedupe, or sync ordering. Absent on records with no better answer than
 *  "the user made it here". */
origin?: "capture" | "alert" | "shared" | "import" | "manual";
```

- **Display only.** Nothing in `capture.ts`, `job-url.ts` or the sync path may
  branch on it. If a future feature needs to branch on origin, that is a
  separate argument to have on its own merits.
- **Optional and absent by default**, so every existing record and every
  existing producer stays valid with no migration.
- Rendered as a small muted line or chip on the tracker row: *"from a job
  alert"*, *"shared with you"*. One short phrase, never a sentence.

### This is a capture-contract change

`types.ts:180-184` is explicit about what adding a field to `JobRecord` obliges:

1. a rule in `job-record-contract.ts` (the mapped type there will not compile
   otherwise),
2. a description in `docs/job-capture-contract.md`.

Both are in scope here. Whether it warrants a `JOB_CAPTURE_CONTRACT_VERSION`
bump is a review question — the field is additive and optional, and the
contract's own precedent (`:158`, "a producer keeps working untouched") suggests
not, but the call should be made explicitly rather than by omission.

An unrecognised `origin` value must be **dropped with a warning**, not refused —
the same asymmetry `job-mapping`-style consumers already apply to a malformed
`capture` block: refusing costs the user the record, warning costs them a line.

## Open questions for review

1. **Is a closed vocabulary right?** A free-text `origin` would let a producer
   say anything, which is friendlier to producers and worse for display — the
   tracker would be rendering a string it cannot phrase. Default: closed.
2. **`origin` or a sub-field of `capture`?** Putting it inside
   `JobCaptureProvenance` avoids a new top-level field, but that block is about
   producer compatibility and is absent on records this build writes — which are
   precisely the ones that most need an origin. Default: top-level.
3. **Does the backup export carry it?** It must, or a round-trip through
   export/import silently launders every row's origin. (`backup.ts`.)

## Acceptance criteria

- [ ] `JobRecord.origin` exists, is optional, and is documented in
      `docs/job-capture-contract.md`.
- [ ] `job-record-contract.ts` validates it: a recognised value is kept, an
      unrecognised one is dropped with a warning and the record is still
      accepted.
- [ ] A record with no `origin` validates unchanged — no existing fixture or
      producer needs editing.
- [ ] The field survives the export → import round-trip.
- [ ] The tracker row renders a short origin phrase when the field is present
      and renders nothing when it is absent.
- [ ] No module outside the validator and the tracker row reads `origin`
      (assert structurally, in the spirit of this repo's other "nothing else
      may reach this" tests).

### LINKED ISSUE #746: The same posting saved from an aggregator and from the employer's ATS becomes two records
## Problem

One posting, reached two ways, becomes two records.

The user finds a job on an aggregator, saves it, follows the "Apply" link to the
employer's own ATS (Workday, Greenhouse, Lever), and saves it there too — or
captures it there a week later having forgotten. The tracker now holds two rows
for one job, each with its own status, notes and résumé link, and nothing on
screen says they are the same posting.

This is working as designed, and the design is right as far as it goes.
`deriveJobId` (`src/lib/storage/job-url.ts`) is a pure function of the
canonicalised URL, and that module's docblock states the asymmetry it is built
on:

> under-merging is a duplicate the user can delete; over-merging destroys a
> record.

So it strips only what cannot plausibly distinguish two postings — `www.`, an
unambiguous `ll-CC` locale segment, a named tracking-parameter list — and
refuses to guess further. An aggregator URL and an employer ATS URL share
**no** host, path or parameter. No URL rule can ever merge them, and none
should try.

What is missing is not a better URL rule. It is any representation of "these two
URLs are the same posting" at all: nothing in `src/lib/storage/` has a notion of
an alias.

## Prior art

Recruidea maintains a URL-alias table populated by its ingest pipeline, mapping
aggregator and ATS-variant URLs onto one catalogue posting. That is the shape
worth borrowing — an alias is a **link between two URLs**, discovered by
something with more context than a URL parser, and it is additive: recording an
alias never rewrites the record it points at.

The offlinecv side needs the same idea locally, and needs it to be **confirmed
by the user** rather than inferred, because this build has no ingest pipeline
whose judgement it can defer to.

## Proposal

Three pieces, each useful on its own; the first is a prerequisite for the other
two.

### A. `JobRecord` can hold alternate URLs

```ts
/** Other URLs the same posting is reachable at — an employer ATS link found on
 *  an aggregator page, or the URL of a record merged into this one. Additive
 *  and display/dedupe-only: `url` stays the canonical one and `id` never
 *  changes because of this field. */
aliasUrls?: string[];
```

Same contract obligations as any new `JobRecord` field
(`src/lib/storage/types.ts:180-184`): a rule in `job-record-contract.ts` and an
entry in `docs/job-capture-contract.md`. Each entry validates as the same
absolute http(s) URL the `url` field does; anything else is dropped with a
warning, never refused.

**`id` is derived from `url` and only from `url`.** An alias must not be able to
change a record's identity — that would fork the id space, which is the failure
`job-url.ts` exists to prevent.

### B. Detect a probable duplicate, and *say so* rather than act

A pure function in `src/lib/` — no storage, no UI — answering "do these two
records look like the same posting?" from evidence the records already carry:

- either record's `url` appears in the other's `aliasUrls` → **certain**;
- same normalised company **and** same normalised title → **probable**;
- same company, similar title, overlapping description → **possible**.

The tracker surfaces a probable-or-better match as an inline affordance on the
row — *"Looks like the same posting as <title>"* — with **Merge** and
**Not the same** actions. Nothing merges without a click. A dismissal is
remembered so the prompt does not come back every render.

### C. Merge is a user action, and it is lossy in one direction only

Merging keeps one record and folds the other in:

- the surviving record gains the other's `url` in `aliasUrls`;
- fields the survivor lacks are filled from the other; fields it has are kept;
- notes are concatenated, never overwritten;
- the merged-away record is deleted through the normal delete path (it
  tombstones, so the deletion replicates rather than being resurrected by the
  next sync).

The user picks which survives, defaulting to the one with the employer's own
URL — the aggregator link is the one that rots.

## Out of scope

- Any automatic merge. The contract's asymmetry forbids it, and this build has
  no evidence source strong enough to overrule it.
- Any change to `deriveJobId` or to the canonicalisation rules. That is a
  contract version bump and a different argument.
- Server-side alias resolution during sync. Whatever a producer or backend does
  with aliases is its own concern; this issue is the local representation and
  the local affordance.

## Acceptance criteria

- [ ] `JobRecord.aliasUrls` exists, is optional, validates each entry as an
      absolute http(s) URL, drops (with a warning) anything else, and is
      documented in `docs/job-capture-contract.md`.
- [ ] A record's `id` is unchanged by any value of `aliasUrls`.
- [ ] The field survives the export → import round-trip.
- [ ] The duplicate-detection function is pure, lives in `src/lib/`, and has
      unit tests covering: alias hit, same company+title, same company +
      different title (must NOT match), and different company + same title
      (must NOT match).
- [ ] The tracker shows the affordance for a probable match and never merges
      without an explicit click.
- [ ] "Not the same" suppresses that pairing durably.
- [ ] After a merge: one record remains, it carries both URLs, no note text was
      lost, and the other record is deleted through the tombstoning path.

s-annam:
## Summary

Three issues about **record identity** in the job tracker, built on one branch because they share a contract file and read as one story.

- **#744** â€” a synced library split one stage of a job search across four top-level sections (`Saved`, `Scouted`, `Shared`, `Interested`). Sections now key on a *display bucket* computed at view time. The stored status is never rewritten: `job-record-contract.ts` accepts any status string on purpose so a producer's vocabulary round-trips, and normalising `shared` â†’ `interested` would destroy that and sync it back. Clicking the bucket a row already sits in writes nothing â€” the selected button has no handler at all, rather than an idempotent-looking `onChange`.
- **#745** â€” new optional `JobRecord.origin` (`capture` | `alert` | `shared` | `import` | `manual`), rendered as a short muted phrase on the row. It answers "how did this row get here" now that the merged sections no longer imply it. Display-only, enforced by a structural test.
- **#746** â€” new optional `JobRecord.aliasUrls`, a pure duplicate-detection module, and a user-driven merge. An aggregator URL and an employer-ATS URL share no host, path or parameter, so no canonicalisation rule can ever merge them and none should try; what was missing was any representation of "these two URLs are the same posting."

`id` stays derived from `url` and only from `url` â€” `deriveJobId` and `canonicalJobUrl` are byte-identical on this branch.

Closes #744
Closes #745
Closes #746

## Review focus

- `src/lib/job-merge.ts:51` â€” `NOT_FILLED` now excludes `capture` and `origin`. Is that the right cut? The claim is that they describe *how a record came to exist* and so cannot be inherited, unlike the posting facts the fill exists for.
- `src/lib/job-merge.ts` â€” the survivor's `status` wins, so merging `interviewing` into `interested` loses the lifecycle position. This follows #744's "fields it has are kept" verbatim; if you disagree, this is the decision to relitigate.
- `src/lib/job-duplicates.ts:242` â€” `probable` is company + title and ignores `location`. Should Google "Software Engineer" in NYC and in Mountain View really be offered a destructive merge? `LEGAL_SUFFIXES` widens the company side at the same time.
- `src/lib/job-duplicates.ts` â€” `SIMILAR_TITLE = 0.6` / `OVERLAPPING_DESCRIPTION = 0.5` are unmeasured guesses. They gate only the `possible` tier, which no affordance acts on.
- `src/lib/job-tracker.ts` â€” a merge reparents the absorbed job's cover letters *before* deleting it, because `deleteJob` cascades to letters (#711). Does the ordering hold if a reparent fails midway?
- `src/components/features/JobDuplicateNotice.tsx:16` â€” #746 proposes defaulting the survivor to the employer's own URL; this offers symmetrically on both rows instead, because the repo has no aggregator-vs-ATS host classifier and guessing wrong picks the wrong survivor for a destructive action.

## Deliberate deviations from the issue text

- **#744's "part C" literal-status chip was dropped.** The issue leaves it explicitly undecided ("decide in review"). The row badge already prints the literal status via `jobStatusLabel`, and #745's origin phrase is the better answer to the question the merge raises. The decision is made here rather than deferred, which is why this PR closes #744 outright.
- **No `JOB_CAPTURE_CONTRACT_VERSION` bump.** Both new fields are additive and optional; `types.ts:158`'s precedent ("a producer keeps working untouched") says a producer that ignores them stays valid. #745 asks for this call to be explicit rather than by omission â€” this is it.
- **`aliasUrls` is unioned on re-capture**, not producer-owned, so a producer re-capturing a posting cannot erase aliases the user's merge created.
- **`sharesAUrl` also treats two records with the same canonical `url` as `certain`**, not only an alias hit â€” "Add a job" mints a UUID rather than deriving an id, so the one duplicate a user can create entirely inside this app would otherwise be invisible.
- **Merge is a two-click confirm**, mirroring the row's existing Remove, rather than the single click the issue's wording implies.

**One factual correction to #745's body:** it justifies drop-with-warning by citing "the same asymmetry consumers already apply to a malformed `capture` block." That premise does not hold â€” `capture: 3`, `capture: null`, `capture: {}` and `capture: {contract: "2"}` all *refuse* the record today. The implementation matches actual house behaviour (a bad *value* is dropped and warned; a bad *type* refuses), and `docs/job-capture-contract.md` Â§8 now states the choice out loud.

## Test plan

- [x] `npm run test` green â€” 4881 passed, 10 skipped, 0 failed
- [x] `npm run lint` clean
- [x] `fallow audit --base origin/main` â€” no new dead code; the only new finding is `findDuplicatePairs` HIGH complexity (report-only per `CLAUDE.md`)
- [ ] Manually verified in `npm run dev`

No fixture binaries are added or changed by this PR, so the Step 3.5 fixture-PII preflight has nothing to clear.

**Unrelated flake seen once under load:** `src/lib/resume-library.test.ts > lists saved resumes newest-first` failed a single run while several agents ran concurrently, then passed 5/5 in isolation. `resume-library.ts:156` sorts by `b.savedAt - a.savedAt` where `savedAt` is a `Date.now()` millisecond; two back-to-back saves landing in the same ms make the comparator return `0`, and a stable sort then preserves insertion order â€” exactly the observed inversion. Pre-existing, untouched by this branch, worth its own issue.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation â€” #744 | unreported (requested: opus) | ultra |
| Implementation â€” #745 | Claude Sonnet 4.6 | high |
| Implementation â€” #746 | Claude Opus 5 (1M context) | ultra |
| Adversarial review â€” round 1 | Claude Opus 5 (1M context) | high |
| Review fixes | Claude Opus 5 (1M context) | high |
| Adversarial review â€” round 2 | Claude Opus 5 (1M context) | high |
| Verification | `npm run verify` â€” green in CI | â€” |
| Review revisions | Gemini 3.5 Flash | medium |

The #744 row is `unreported` on purpose: that subagent went idle without answering the model self-report, and inferring a version string from the `model: opus` alias that was *requested* would be a fabrication. Its work was verified independently regardless.

## Adversarial review

Two bounded rounds ran on the local diff before this PR existed, so the reviewed base is what you are reading.

**Round 1 â€” 1 blocking finding, reproduced:**

`src/lib/job-merge.ts` â€” **a merge wrote false provenance onto the surviving record.** `NOT_FILLED` omitted `capture`, so the generic fill loop copied the absorbed record's block onto the survivor whenever the survivor lacked one. Scenario: the user hand-adds a job (no `capture`), the extension separately captured the same posting, the user merges keeping their own row â€” the survivor then claims `producer: "offlinecv-extension"`, a compatibility assertion about a record that no longer exists, attached to a record that producer never wrote. This contradicts `types.ts:277` ("Absent for every record this app creates") and Â§7's stated purpose of informing "a future reader deciding whether it can trust the fields" â€” precisely the reader it misleads, and precisely what a `capture.contract`-keyed migration would key on.

The same loop copied `origin`, which made it a second-order defect: the merge path genuinely propagated `origin`, but through a generic copy that never spells `.origin`, so #745's structural guard never fired. Worse, that guard's anti-drift half turned out satisfiable by a **comment** â€” `JobTrackerEntry.tsx`'s docblock contains "`job.origin` is display-only", which matched the bare `/\.origin\b/`; stubbing the render out left it green.

**Fix:** `capture` and `origin` added to `NOT_FILLED` with a docblock naming the distinction; the guard's drift half now strips comments before matching, and the offender assertion carries a hint explaining that a `URL.origin` / `event.origin` read is a false positive to be narrowed out rather than allowlisted around.

**Round 2 â€” clean, zero new blocking findings.** The fix was verified by fail-before revert (removing the two keys fails the new regression test), by enumerating every other path an absorbed record's `capture`/`origin` could take to the survivor (fill loop, explicit rules, `putRecord`'s wholesale replace, letters reparenting, re-capture union, backup import â€” no second door), and by re-stubbing both allowlisted files to confirm the hardened guard is non-vacuous on each.

**Invariants attacked and held:** no path rewrites a stored status; `deriveJobId`/`canonicalJobUrl` byte-identical; nothing merges without an explicit click; notes concatenated and the absorbed record tombstoned rather than hard-deleted; cover letters on both jobs survive; all four `useMemo`/`useCallback` dep arrays exact â€” which is load-bearing here because `eslint.config.js` registers no `react-hooks` plugin, so a stale closure lints green.

**Surviving nits, not fixed** (the loop had converged; unreviewed edits after convergence defeat the point):

- `types.ts:281` â€” the `origin` docblock points at `job-record-contract.ts`'s "drift guard", but that section is the `JOB_RECORD_RULES` mapped type, a different check. The reach guard is `src/lib/job-origin-reach.test.ts`.
- `job-origin-reach.test.ts:54` â€” `/\.origin\b/` cannot see a destructured read (`const { origin } = job`). Worth a docblock sentence naming the known evasion; not worth widening the regex.
- `job-duplicates.ts:231` â€” `sharesAUrl` misses `a.aliases âˆ© b.aliases`, the exact aggregator/ATS shape. Under-merge is the safe direction, so a gap rather than a defect.
- `job-record-contract.ts:337` â€” `keptAliasUrls` neither dedupes nor excludes the record's own `url`; `dedupeCanonicalUrls` already exists and would make all three write paths agree.
- `capture.ts:107` â€” `captureJob` never consults `aliasUrls` when resolving identity, so re-capturing a merged-away URL recreates the row and re-offers the same merge forever. Not blocking: `captureJob` has no production caller in this build, and #746 scopes sync-side alias resolution out.
- `job-merge.ts:140` â€” `jdText` and `matchResult` fill independently, so a survivor can hold its own JD beside a match result computed against a different one. No consumer reads `matchResult` today.
- `JobTrackerEntry.tsx` (283 LOC) and `JobTracker.tsx` (261) exceed the ~200 guideline. Both were already over before this branch; the new concerns went into siblings (`JobDuplicateNotice`, `useJobDuplicates`).

### DIFF
--- a/src/lib/job-duplicates.ts
+++ b/src/lib/job-duplicates.ts
@@ -1,0 +1,407 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * job-duplicates — "do these two tracked records look like the same posting?"
+ * (#746). Pure, zero-storage, zero-UI: it reads two {@link JobRecord}s and
+ * returns a confidence, and every caller decides for itself what to do with the
+ * answer. Sibling of `job-status-bucket.ts`, and testable at module scope for
+ * the same reason.
+ *
+ * ## Why this exists at all
+ *
+ * `deriveJobId` (`storage/job-url.ts`) is a pure function of the canonicalised
+ * URL, and it refuses to guess beyond what cannot plausibly distinguish two
+ * postings, because **under-merging is a duplicate the user can delete;
+ * over-merging destroys a record.** An aggregator listing and the employer's own
+ * ATS page for one job share no host, no path and no parameter, so no URL rule
+ * can ever relate them and none should try. The missing piece was never a better
+ * URL rule — it was any representation of "these two URLs are the same posting".
+ * `JobRecord.aliasUrls` is that representation, and this module is what reads it.
+ *
+ * ## This module NEVER decides anything
+ *
+ * It reports evidence; it does not merge, write, or rank. The asymmetry above
+ * forbids an automatic merge outright — this build has no evidence source strong
+ * enough to overrule it — so the strongest thing a caller may do with a
+ * {@link JobDuplicateConfidence} is *offer* a merge that a user then clicks.
+ * Nothing here reads or writes `id`: an alias is display/dedupe evidence and
+ * must never fork the id space.
+ *
+ * ## Totality
+ *
+ * Every field is read through {@link text} / an `Array.isArray` guard rather
+ * than trusted from the type. Records reach the `jobs` store through a
+ * deliberately permissive write (`saveJob`) and from a backup file that may
+ * predate any field here, so a record whose `title` is not a string is
+ * reachable — and this function runs on every tracker render. It degrades to
+ * "no evidence", never to a throw.
+ */
+
+import { canonicalJobUrl } from "./storage/index.ts";
+import type { JobRecord } from "./storage/index.ts";
+
+/**
+ * How strongly two records look like one posting.
+ *
+ * - `certain` — their URL sets intersect: one record's `url` is the other's, or
+ *   appears in the other's `aliasUrls`. Somebody with more context than a URL
+ *   parser (a user who merged, a producer that followed an "Apply" link) said
+ *   these are the same page.
+ * - `probable` — same normalised company AND same normalised title. Strong, but
+ *   inference: a company really can post two identically-titled roles.
+ * - `possible` — same company, similar title, overlapping description. Weak by
+ *   construction, and deliberately below the bar {@link isActionableDuplicate}
+ *   sets, so no merge affordance is ever offered on it.
+ */
+export type JobDuplicateConfidence = "certain" | "probable" | "possible";
+
+const CONFIDENCE_RANK: Record<JobDuplicateConfidence, number> = {
+  possible: 1,
+  probable: 2,
+  certain: 3,
+};
+
+/**
+ * Whether a confidence is strong enough to put a *merge* in front of the user.
+ * `probable` and up. A `possible` match is computed, and is worth having as a
+ * distinct answer, but offering a destructive action on "same company, similar
+ * title" would invert the asymmetry this whole module is written around.
+ */
+export function isActionableDuplicate(confidence: JobDuplicateConfidence): boolean {
+  return CONFIDENCE_RANK[confidence] >= CONFIDENCE_RANK.probable;
+}
+
+/**
+ * Fraction of the SMALLER token set the two share. Containment rather than
+ * Jaccard: "Senior Frontend Engineer" against a 900-word job description is a
+ * containment of 1 and a Jaccard of ~0.003, and it is containment that says
+ * what we mean — does the shorter text appear inside the longer one.
+ */
+function containment(a: ReadonlySet<string>, b: ReadonlySet<string>): number {
+  if (a.size === 0 || b.size === 0) return 0;
+  const [small, large] = a.size <= b.size ? [a, b] : [b, a];
+  let shared = 0;
+  for (const token of small) if (large.has(token)) shared += 1;
+  return shared / small.size;
+}
+
+/** Title-token containment at or above which two titles count as "similar".
+ *  A guess, not a measurement — it only ever gates the `possible` tier, which
+ *  no affordance acts on, so the cost of it being wrong is bounded to a value
+ *  nothing renders. */
+const SIMILAR_TITLE = 0.6;
+
+/** Description-token containment at or above which two job descriptions count
+ *  as overlapping. Same bounded blast radius as {@link SIMILAR_TITLE}. */
+const OVERLAPPING_DESCRIPTION = 0.5;
+
+/** Description tokens shorter than this are dropped before comparing: "a",
+ *  "of", "to" appear in every posting ever written and would float the
+ *  containment of two unrelated descriptions. Titles are NOT filtered this way —
+ *  "QA" and "ML" are the whole signal in a title. */
+const DESCRIPTION_TOKEN_MIN_LENGTH = 3;
+
+/** Anything that is not a letter or a number, in any script. `\p{L}\p{N}` and
+ *  not `a-z0-9`, so "Nestlé" and "楽天" normalise to themselves rather than to
+ *  a stump. */
+const NON_WORD = /[^\p{L}\p{N}]+/gu;
+
+/** A field's value if it is really a string, otherwise the empty string — see
+ *  the totality note in the module docblock. */
+function text(value: unknown): string {
+  return typeof value === "string" ? value : "";
+}
+
+/** Lowercased, punctuation-flattened, whitespace-collapsed. The one place
+ *  "normalised" is defined for this module. */
+function normalise(value: unknown): string {
+  return text(value).toLowerCase().replace(NON_WORD, " ").trim();
+}
+
+function tokenSet(value: unknown, minLength: number): Set<string> {
+  const tokens = new Set<string>();
+  for (const token of normalise(value).split(" ")) {
+    if (token.length >= minLength) tokens.add(token);
+  }
+  return tokens;
+}
+
+/**
+ * Trailing legal-form words dropped from a company name, so the ATS's
+ * "Acme, Inc." and the aggregator's "Acme" are one company. Only ever stripped
+ * from the END, where a legal form actually sits — a leading or interior match
+ * ("Co-operative Bank") is a real word.
+ *
+ * Stripping widens what counts as the same company, which widens what can be
+ * offered as a merge, so the list stays short and boring: only forms that are
+ * legal suffixes and nothing else in ordinary company-name usage.
+ */
+const LEGAL_SUFFIXES = new Set([
+  "inc",
+  "incorporated",
+  "llc",
+  "ltd",
+  "limited",
+  "corp",
+  "corporation",
+  "company",
+  "gmbh",
+  "plc",
+  "ag",
+  "bv",
+  "nv",
+  "pty",
+  "pvt",
+  "pte",
+  "oy",
+  "ab",
+]);
+
+function normaliseCompany(value: unknown): string {
+  const words = normalise(value).split(" ").filter((word) => word.length > 0);
+  // `> 1`, so a company whose whole name is a legal form ("Limited") keeps it
+  // rather than normalising to the empty string, which would then match every
+  // other blank-company record.
+  while (words.length > 1 && LEGAL_SUFFIXES.has(words[words.length - 1])) words.pop();
+  return words.join(" ");
+}
+
+/**
+ * One record reduced to what a comparison needs. Built once per record per
+ * sweep so {@link findDuplicatePairs} normalises N records instead of N²
+ * pairs — at a few hundred saved jobs the difference is the whole cost.
+ */
+interface Fingerprint {
+  id: string;
+  /** Canonical form of `url`, or undefined when there is none to canonicalise. */
+  canonicalUrl?: string;
+  /** Canonical forms of `aliasUrls`. Entries that do not canonicalise are
+   *  dropped: they can never match anything, and the capture contract refuses
+   *  them at the boundary anyway (§9). */
+  aliases: Set<string>;
+  company: string;
+  title: string;
+  titleTokens: Set<string>;
+  jdText: string;
+  /** Filled by {@link jdTokensOf} on first use. Tokenising a job description is
+   *  the one expensive step here and only the `possible` tier ever needs it, so
+   *  a library where no pair has a similar title never pays for it. */
+  jdTokens?: Set<string>;
+}
+
+function jdTokensOf(print: Fingerprint): Set<string> {
+  print.jdTokens ??= tokenSet(print.jdText, DESCRIPTION_TOKEN_MIN_LENGTH);
+  return print.jdTokens;
+}
+
+function fingerprint(job: JobRecord): Fingerprint {
+  const aliases = new Set<string>();
+  // `Array.isArray`, not `?? []`: a stored record's `aliasUrls` is only as
+  // typed as whatever wrote it, and `for…of` over a non-iterable throws.
+  if (Array.isArray(job.aliasUrls)) {
+    for (const alias of job.aliasUrls) {
+      const canonical = canonicalJobUrl(text(alias));
+      if (canonical !== undefined) aliases.add(canonical);
+    }
+  }
+  return {
+    id: job.id,
+    canonicalUrl: canonicalJobUrl(text(job.url)),
+    aliases,
+    company: normaliseCompany(job.company),
+    title: normalise(job.title),
+    titleTokens: tokenSet(job.title, 1),
+    jdText: text(job.jdText),
+  };
+}
+
+/**
+ * Do the two records' URL sets intersect?
+ *
+ * Both directions of the alias check, plus the degenerate case where the two
+ * records simply hold the same posting URL. That last one is not a new identity
+ * rule and changes nothing about `deriveJobId`: it is the *existing* rule read
+ * as evidence. Two records with one canonical URL are the same posting by the
+ * capture contract's own definition (§2) — they only exist as two rows because
+ * the tracker's own "Add a job" mints a UUID rather than deriving an id, so the
+ * one duplicate a user can produce entirely inside this app would otherwise be
+ * the one duplicate this module could not see.
+ */
+function sharesAUrl(a: Fingerprint, b: Fingerprint): boolean {
+  if (a.canonicalUrl !== undefined && a.canonicalUrl === b.canonicalUrl) return true;
+  if (a.canonicalUrl !== undefined && b.aliases.has(a.canonicalUrl)) return true;
+  if (b.canonicalUrl !== undefined && a.aliases.has(b.canonicalUrl)) return true;
+
+  // Intersecting alias sets (#746) — both records are aliases of a common third URL,
+  // indicating they are views of the same underlying posting.
+  for (const alias of a.aliases) {
+    if (b.aliases.has(alias)) return true;
+  }
+  return false;
+}
+
+function compare(a: Fingerprint, b: Fingerprint): JobDuplicateConfidence | null {
+  if (sharesAUrl(a, b)) return "certain";
+  // An empty company is the common case for a half-filled record, not a match:
+  // without this, every blank-company job would pair with every other one.
+  if (a.company === "" || a.company !== b.company) return null;
+  if (a.title !== "" && a.title === b.title) return "probable";
+  if (containment(a.titleTokens, b.titleTokens) < SIMILAR_TITLE) return null;
+  if (containment(jdTokensOf(a), jdTokensOf(b)) < OVERLAPPING_DESCRIPTION) return null;
+  return "possible";
+}
+
+/**
+ * How strongly `a` and `b` look like the same posting, or null for no evidence.
+ *
+ * Symmetric, and deliberately does NOT check that the two ids differ: "are
+ * these the same posting" has an honest answer for a record compared with
+ * itself, and it is the caller's business not to ask. {@link findDuplicatePairs}
+ * never does.
+ */
+export function jobDuplicateConfidence(
+  a: JobRecord,
+  b: JobRecord,
+): JobDuplicateConfidence | null {
+  return compare(fingerprint(a), fingerprint(b));
+}
+
+/** One unordered pairing, with `a` < `b` so a pair has exactly one
+ *  representation whichever side found it. */
+export interface JobDuplicatePair {
+  a: string;
+  b: string;
+  confidence: JobDuplicateConfidence;
+}
+
+function pushInto<K, V>(index: Map<K, V[]>, key: K, value: V): void {
+  const bucket = index.get(key);
+  if (bucket) bucket.push(value);
+  else index.set(key, [value]);
+}
+
+/**
+ * The one identity of an unordered pairing of two records: sorted, so a pair
+ * has the same key whichever side found it, and NUL-joined — the separator
+ * `JobTrackerStatusGroup` already picked, because an id here is whatever a
+ * backup carried rather than necessarily a UUID, and a printable separator
+ * would let `("a b", "c")` and `("a", "b c")` collide on one key.
+ *
+ * Exported because `job-duplicate-dismissals.ts` keys the user's "Not the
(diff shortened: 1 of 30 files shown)


s-annam (commented):
Stunningly thorough implementation that perfectly satisfies all acceptance criteria across the three sibling issues (#744, #745, #746) with excellent test coverage and a highly robust architectural design. 

The pure helper-focused implementation style (splitting duplicate calculations, bucketing, and merges into module-scoped helpers) makes the code exceptionally easy to verify. No blocking findings were detected during the adversarial review pass.

---

src/lib/job-duplicates.ts
```diff
@@ -0,0 +1,361 @@
...
+ * Both directions of the alias check, plus the degenerate case where the two
+ * records simply hold the same posting URL. That last one is not a new identity
+ * rule and changes nothing about `deriveJobId`: it is the *existing* rule read
+ * as evidence. Two records with one canonical URL are the same posting by the
+ * capture contract's own definition (§2) — they only exist as two rows because
+ * the tracker's own "Add a job" mints a UUID rather than deriving an id, so the
+ * one duplicate a user can produce entirely inside this app would otherwise be
+ * the one duplicate this module could not see.
+ */
+function sharesAUrl(a: Fingerprint, b: Fingerprint): boolean {
```
s-annam, line 231: Note that `sharesAUrl` will miss cases where the two records' alias sets intersect (i.e. `a.aliases ∩ b.aliases`), but neither's canonical `url` is direct. While this represents a safe under-merge gap rather than a defect, checking for alias intersection could be a useful future improvement.

src/lib/storage/job-record-contract.ts
```diff
@@ -119,6 +119,49 @@ export type JobRecordValidation =
...
+function describeAliasEntry(value: unknown): string {
+  return typeof value === "string" ? `"${value}"` : `a value of type ${typeof value}`;
+}
+
+/** The entries of an already-guarded `aliasUrls` that clear the `href` bar —
+ *  the value actually stored. Absolute http(s) only: unlike `url`, there is no
+ *  legacy corpus of half-typed values to protect here and nothing renders an
+ *  alias as the row's link, so an entry that cannot be canonicalised is an
+ *  entry that can never match anything. */
+function keptAliasUrls(value: unknown): string[] {
```
s-annam, line 160: `keptAliasUrls` filters for absolute URL formats but does not deduplicate or filter out the canonical `url` at the validator layer. Since `dedupeCanonicalUrls` is called in the write paths (merging, capturing), this doesn't cause functional issues, but deduping at validation would make the stored representation cleaner.

src/lib/storage/types.ts
```diff
@@ -247,6 +277,19 @@ export interface JobRecord extends StoredRecord {
   /** Provenance for a record written by a producer outside this build (#693).
    *  Absent for every record this app creates. */
   capture?: JobCaptureProvenance;
+
+  /** Where this record says it came from, for DISPLAY ONLY (#745) — the
```
s-annam, line 281: This docblock mentions the "drift guard" in `job-record-contract.ts`. The actual reach guard checking that no other module reads `.origin` is the source-code sweep test in `src/lib/job-origin-reach.test.ts`.

src/lib/job-origin-reach.test.ts
```diff
@@ -0,0 +1,116 @@
...
+const SRC_ROOT = join(HERE, "..");
+
+/** The only two production modules allowed to read `job.origin` (paths
+ *  relative to `src/`) — see the module docblock. */
+const ALLOWED = new Set([
+  "lib/storage/job-record-contract.ts",
+  "components/features/JobTrackerEntry.tsx",
+]);
+
+const ORIGIN_READ = /\.origin\b/;
```
s-annam, line 59: The `/.origin\b/` regex cannot detect destructured reads like `const { origin } = job`. A comment documenting this evasion would be helpful to keep future editors from accidentally bypassing the test.

src/lib/job-duplicates.ts
```diff
@@ -0,0 +1,361 @@
...
+ *    (an aggregator's "Acme Corp" against the ATS's "Acme"), so it is found
+ *    through a URL index over the whole library;
+ *  - company+title evidence requires the companies to match by definition, so
+ *    it only ever compares within one normalised-company bucket.
+ *
+ * A record whose company normalises empty is in no bucket and can still be
+ * found by the URL pass, which is right: a blank company is missing evidence,
+ * not evidence of difference.
+ */
+export function findDuplicatePairs(jobs: readonly JobRecord[]): JobDuplicatePair[] {
```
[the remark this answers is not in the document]
s-annam: Fixed in 9f4f3c1 — partitioned `findDuplicatePairs` into modular indexed helper functions (`indexByUrl`, `findUrlPairs`, `findCompanyPairs`) to reduce cognitive complexity below the threshold.

src/lib/job-duplicates.ts
s-annam: Fixed in 9f4f3c1 — updated `sharesAUrl` to support checking for alias-alias set intersections, and added tests verifying it.

src/lib/storage/job-record-contract.ts
s-annam: Fixed in 9f4f3c1 — updated validation flow in `validateJobRecord` to deduplicate `aliasUrls` and exclude the canonical `url` at the validator layer.

src/lib/storage/types.ts
s-annam: Fixed in 9f4f3c1 — updated docblock in `types.ts` to correctly reference the reach guard test file `job-origin-reach.test.ts`.

src/lib/job-origin-reach.test.ts
s-annam: Fixed in 9f4f3c1 — added a warning comment documenting this destructuring evasion limitation above `ORIGIN_READ` in `job-origin-reach.test.ts`.

PR #752, opened 2026-08-05
fix(seo): make the site indexable (#751)

### LINKED ISSUE #751: Site is not indexed: soft-404 SPA fallback, missing robots.txt/sitemap.xml, indexable staging duplicate, no canonical, no crawlable content
## Problem

`offlinecv.org` is verified in Google Search Console as a domain property (the `google-site-verification` TXT record resolves), but its pages are not indexed. The cause is not a Search Console setting. It is five independent defects in what the site actually serves, four of which actively suppress indexing and one of which leaves the crawler with nothing to index.

All five were reproduced against production on 2026-08-04.

### 1. `robots.txt` and `sitemap.xml` do not exist — both return HTML

```
$ curl -s -o /dev/null -w "%{http_code} %{content_type}\n" https://offlinecv.org/robots.txt
200 text/html; charset=utf-8

$ curl -s https://offlinecv.org/sitemap.xml | head -2
<!doctype html>
<html lang="en">
```

Neither file is in `public/`, so both requests fall through to the SPA not-found handler and are answered with the root parser page. A sitemap submitted to Search Console cannot parse. A `robots.txt` full of HTML contains no valid directives.

### 2. Every unknown path returns 200 with the homepage (soft 404)

```
$ curl -s -o /dev/null -w "%{http_code}\n" https://offlinecv.org/this-page-does-not-exist
200
$ curl -s https://offlinecv.org/nonsense-xyz | grep -i '<title>'
    <title>OfflineCV — see what a resume parser reads from your PDF</title>
```

`wrangler.jsonc` sets `"not_found_handling": "single-page-application"`, which answers any unmatched request with the root `/index.html`. This produces an unbounded set of distinct URLs all serving identical content — the shape Search Console reports as "Soft 404" and "Duplicate without user-selected canonical".

This setting is also inconsistent with the rest of the build. `vite.config.ts` already declares `appType: "mpa"` with a docblock explaining that this app has no client-side router and that an unmatched path should 404 honestly. Production was the one target still falling back.

### 3. `dev.offlinecv.org` is a fully indexable byte-identical duplicate

```
$ dig +short dev.offlinecv.org
offlinecv.github.io.
$ curl -s https://dev.offlinecv.org/ | grep -i '<title>'
    <title>OfflineCV — see what a resume parser reads from your PDF</title>
```

`.github/workflows/deploy-pages.yml` deploys every push to `main` to GitHub Pages at `dev.offlinecv.org`. This is intentional staging (it builds with `VITE_POSTHOG_ENV: staging`, documented in the README's telemetry section), but it ships no `noindex`, no canonical, and no robots directive. It is a subdomain of the same registrable domain, so it sits inside the `sc-domain:offlinecv.org` property. A search engine seeing two complete identical sites picks one and drops the other.

### 4. No canonical tags anywhere

None of the three HTML entries (`index.html`, `jobs/index.html`, `jd-fit/index.html`) carries `<link rel="canonical">`, and none carries Open Graph or Twitter card metadata. There is nothing telling a crawler which host or which URL form is authoritative.

### 5. There is no indexable content

This is the deepest problem, and fixing 1–4 does not fix it. All three entries render into an empty `<div id="root">`:

```html
  <body>
    <div id="root"></div>
    <script type="module" crossorigin src="/assets/jobs-BcWlklYR.js"></script>
  </body>
```

Google does render JavaScript, but it defers it to a second wave. A new domain with no inbound links and three JS-only application shells gives the first-wave indexer a `<title>` and a `<meta name="description">` and nothing else. That is the textbook input for "Crawled — currently not indexed".

### Reuse analysis

**Capability:** publish crawler-readable pages describing the product, and emit per-deploy-target search-engine directives.

**Existing surfaces found:**
- `src/components/features/PageShell.tsx` — the shared chrome (header + footer link row) for all three app surfaces. Owns the footer link list.
- `vite.config.ts:emitVersionJson` — the existing in-repo pattern for emitting an unhashed root asset at build time (`dist/version.json`).
- `public/` — Vite's verbatim-copy directory. Currently holds only `CNAME`.
- No existing static/marketing page surface. No existing SEO or robots handling anywhere in the tree.

**Decision:** **extend** on both axes rather than build new machinery.
- The build-time file emission extends the `emitVersionJson` plugin shape — same `Plugin` object, same `generateBundle` + `this.emitFile` idiom, registered in the same `plugins` array.
- The internal links extend `PageShell`'s existing footer link row rather than adding a nav component.
- The content pages themselves are genuinely new, and are deliberately **not** React feature components: their entire purpose is to be readable with the bundle unexecuted, which a component under `src/components/features/` cannot be. They are plain HTML under `public/`. This is the one "build new", and the no-JS requirement is the justification — no design-system primitive can satisfy it, because every primitive requires React to have mounted.

### Why the directives are generated, not committed

A file committed to `public/` is copied identically into both deploy targets, which is precisely what must not happen: production needs an indexable `robots.txt` plus a sitemap, and staging needs neither. Generating them from one `SITEMAP_PATHS` list behind an env flag keeps a single source of truth and lets the two targets diverge.

### Why staging still allows crawling

The instinct on a duplicate is `Disallow: /`. That is wrong here. A blocked URL cannot be re-crawled, and a URL already in the index can only be *dropped* by a directive the crawler is permitted to read — so a robots-blocked staging copy lingers indefinitely. Staging therefore serves an allow-all `robots.txt` and carries `<meta name="robots" content="noindex, nofollow">` on every HTML entry. Crawl budget on a host nothing links to is not worth trading for that.

Static pages under `public/` bypass `transformIndexHtml` (Vite copies them through untouched), so they cannot receive the injected meta tag. They carry a hardcoded absolute canonical pointing at `https://offlinecv.org/...` instead, which consolidates the staging copy onto production by a different mechanism.

### Step 1 — `vite.config.ts`: add an `emitSeoFiles()` plugin

New module-scope constants next to the existing `BASE_PATH` / `APP_VERSION` block: `SITE_ORIGIN` (`"https://offlinecv.org"`), `SEO_NOINDEX` (`process.env.OFFLINECV_SEO_NOINDEX === "1"`), and `SITEMAP_PATHS` (the seven canonical trailing-slash URLs).

New plugin, mirroring `emitVersionJson`:

```ts
function emitSeoFiles(): Plugin {
  return {
    name: "offlinecv:seo",
    apply: "build",
    transformIndexHtml() {
      if (!SEO_NOINDEX) return;
      return [{ tag: "meta", attrs: { name: "robots", content: "noindex, nofollow" }, injectTo: "head-prepend" as const }];
    },
    generateBundle() {
      const robots = ["User-agent: *", "Allow: /", ""];
      if (!SEO_NOINDEX) robots.push(`Sitemap: ${SITE_ORIGIN}/sitemap.xml`, "");
      this.emitFile({ type: "asset", fileName: "robots.txt", source: robots.join("\n") });
      if (SEO_NOINDEX) return;
      // ...emit sitemap.xml from SITEMAP_PATHS
    },
  };
}
```

Use the `transformIndexHtml` **tag-descriptor** return form, not a string replace against an existing `<meta>` — a regex against sibling markup breaks the moment the head is reordered.

Omit `<lastmod>` from the sitemap. The only value available at build time is the build timestamp, which changes on every redeploy whether or not the page did; an inaccurate `lastmod` is worse than none.

Register `emitSeoFiles()` in the `plugins` array after `emitVersionJson(APP_VERSION)`.

### Step 2 — `.github/workflows/deploy-pages.yml`: mark the staging build

Add `OFFLINECV_SEO_NOINDEX: "1"` to the `env:` block of the `npm run build` step, with a comment explaining that `dev.offlinecv.org` would otherwise compete with production.

### Step 3 — `wrangler.jsonc`: stop the soft 404s

Change `"not_found_handling"` from `"single-page-application"` to `"404-page"`. Rewrite the file's header comment to record why (it currently says "SPA not-found handling mirrors the old Pages behavior", which is the behaviour being removed).

### Step 4 — canonical + Open Graph on the three entries

Add to `index.html`, `jobs/index.html`, `jd-fit/index.html`: an **absolute** `<link rel="canonical">`, `og:type` / `og:site_name` / `og:url` / `og:title` / `og:description`, and `twitter:card`. Reuse each page's existing `<title>` and `<meta name="description">` verbatim as the OG values — they are already reviewed copy.

Absolute, not base-relative: the staging build serves this same HTML, and this tag is what folds that copy onto production.

Do **not** add `og:image` — there is no image asset in the repo, and referencing a missing file is worse than omitting the tag. Track separately.

### Step 5 — `public/404.html`

Self-contained: inline `<style>`, no `/content.css` link, no bundle. This page is served *at* the URL that was not found, so a relative asset path resolves against an arbitrary directory depth. Carries `<meta name="robots" content="noindex">` and root-absolute links back into the app. Cloudflare's `404-page` handling and GitHub Pages both pick up `/404.html` natively, so one file serves both targets.

### Step 6 — four static content pages + shared stylesheet

New files: `public/content.css`, `public/how-it-works/index.html`, `public/faq/index.html`, `public/privacy/index.html`, `public/open-source/index.html`.

Directory-index form (`<name>/index.html`), matching the reasoning already recorded in `vite.config.ts` for why `jd-fit/` is a directory rather than a flat `jd-fit.html`. Internal links relative (`../`, `../jobs/`) so they resolve under any base path.

`content.css` mirrors the light/dark values from `src/design-system/styles/tokens.css` as plain custom properties. It cannot import `@design-tokens` — it is copied verbatim out of `public/` and never processed by Vite. That duplication is the cost of the no-JS requirement and must be documented in the file so the next person does not "fix" it into an import.

Content requirements, per the repo's copy discipline (`CLAUDE.md`, and the "Copy claims must round-trip to code" rule):

- **how-it-works** — the cascade tiers, the three scored dimensions with their real weights (`0.4` / `0.3` / `0.3` from `WEIGHTS` in `src/lib/score/score.ts`), the layout multiplier ladder (`1.0` / `0.85` / `0.7` / `0` scanned, from the same file), and the verdict bands (`>= 80` Strong, `>= 60` Getting There, else Needs Work, from `getScoreTier`/`getScoreLabel`). Every number quoted must be verified against the source, not recalled.
- **privacy** — an itemised egress list: page load, job feeds (only on click of Search jobs; carries the keyword string from `src/lib/job-search/providers/keywords.ts`), company adapters (public board slug), the GitHub star count, WebLLM model-weight download, PostHog analytics on the hosted build, and the opt-in feedback email. Plus what is stored locally. Must claim **custody, not runtime** — "runs on your device" is falsifiable and false. Must not exceed the README's telemetry section.
- **faq** — question-shaped headings with `FAQPage` JSON-LD. The JSON-LD must be a faithful copy of the visible Q&A; structured data that does not match the rendered page is a spam signal. Must include the "does this tell me what an ATS will see?" question answered honestly (it does not — we run one generic extractor and have measured no specific ATS).
- **open-source** — Apache 2.0, the real `package.json` script names, the analytics difference between a local build and the hosted one, and the "don't attach your real resume to an issue" guidance pointing at the local CLI probe.

### Step 7 — link the content pages from the app

Content pages that nothing links to do not get found. In `PageShell.tsx`'s existing footer link row, add same-tab links to all four using the base-aware `import.meta.env.BASE_URL` pattern already established by `savedJobsHref()` in `src/lib/jobs-landing.ts`:

```tsx
<a href={`${import.meta.env.BASE_URL}how-it-works/`} className="hover:underline">How it works</a>
```

Repoint the existing "Privacy & data" link from the README anchor to the new `/privacy/` page.

### Step 8 — copy trim (landing block + footer)

Separately motivated, same surfaces, so it lands here rather than in a second PR touching the same lines. The landing block below the drop zone currently runs three paragraphs; an open source project does not need to sell itself on its own landing page.

- `src/App.tsx` — cut the product-pitch paragraph ("OfflineCV parses your resume, shows you what a machine read back…") and the Greenhouse "2025 AI in Hiring Report" citation paragraph. **Keep** "No account, no email, results in seconds — and the same file always gets the same score." That line is a guarantee rather than a pitch, and both halves round-trip to code (no auth exists in the tree; `computeAnonymousAtsScore` is a pure function of the extracted text). Collapse the now-single-child flex wrapper.
- `src/components/features/PageShell.tsx` — remove the footer prose line ("Your PDF stays in this browser tab by default and is never used to train AI. AI analysis is optional.") and the "Further reading: HBS Hidden Workers" external link. The privacy claim was compressed past the point where it could carry its own caveats; the honest version needs the itemised egress list, which is now one link away.

### Step 9 — fix the docblocks the above invalidates

Three docblocks assert things that stop being true. Each must move in the same commit:

- `src/lib/jobs-landing.ts` — its `#saved`-hash rationale cites the `single-page-application` not-found behaviour by name. The conclusion still holds (an unmatched path still does not reach `/jobs/`), but the stated mechanism changes.
- `src/design-system/shared/CapabilityStrip.tsx` — cross-references `PageShell`'s footer sentence, which is being deleted. That makes this rail the only place the custody claim is *made* rather than linked to.
- `src/App.tsx` — the long comment above the landing block documents copy that no longer exists.

## Acceptance criteria

- [ ] `npm run build` emits `dist/robots.txt` containing `Allow: /` and a `Sitemap: https://offlinecv.org/sitemap.xml` line.
- [ ] `npm run build` emits `dist/sitemap.xml` that parses as XML and contains exactly the 7 URLs: `/`, `/jd-fit/`, `/jobs/`, `/how-it-works/`, `/faq/`, `/privacy/`, `/open-source/`.
- [ ] `OFFLINECV_SEO_NOINDEX=1 npm run build` emits a `dist/robots.txt` with **no** `Sitemap:` line, emits **no** `dist/sitemap.xml`, and injects `<meta name="robots" content="noindex, nofollow">` into all three of `dist/index.html`, `dist/jobs/index.html`, `dist/jd-fit/index.html`.
- [ ] A default `npm run build` injects **no** `noindex` into any entry (verify explicitly — a flag inverted here fails silently and deindexes production).
- [ ] All three entries in `dist/` carry an absolute `<link rel="canonical">` matching their own production URL, plus OG and `twitter:card` tags.
- [ ] `dist/404.html` exists, references no external stylesheet, and carries `<meta name="robots" content="noindex">`.
- [ ] `dist/how-it-works/index.html`, `dist/faq/index.html`, `dist/privacy/index.html`, `dist/open-source/index.html` and `dist/content.css` all exist after a build.
- [ ] Every question in the FAQ page's `FAQPage` JSON-LD matches a visible `<h2>` on the same page, one for one, in order.
- [ ] Every numeric claim on the how-it-works page (dimension weights, layout multipliers, verdict band thresholds) matches the current value in `src/lib/score/score.ts`.
- [ ] The privacy page names every egress path that exists in the tree and invents none — cross-check against the actual `fetch(` callsites, not against this issue.
- [ ] `PageShell`'s footer links to all four content pages, base-aware, and no longer renders the removed prose line or the HBS link.
- [ ] The landing block renders exactly one paragraph.
- [ ] `npm run verify` passes: `typecheck`, `lint`, the full suite, and `build`. (fallow is report-only and will flag `public/content.css` as unreachable — it is referenced from static HTML that fallow does not parse. Per `CLAUDE.md` a fallow finding is a Nit, never blocking.)
- [ ] No existing test regresses. The suite is 4985 passing / 10 skipped before this change.

## Verification after deploy

These cannot be asserted in CI — they need the live origin:

```bash
curl -s https://offlinecv.org/robots.txt                                   # real directives, not HTML
curl -s https://offlinecv.org/sitemap.xml | head -3                        # real XML
curl -s -o /dev/null -w "%{http_code}\n" https://offlinecv.org/nope-xyz    # expect 404, not 200
curl -s https://dev.offlinecv.org/ | grep -c noindex                       # expect 1
```

Then in Search Console: submit `https://offlinecv.org/sitemap.xml`, run URL Inspection → Request Indexing on the three app entries and the four content pages, and read the Page Indexing report for the actual per-URL reason strings.

## Out of scope

- **`og:image`** — no image asset exists in the repo. Separate issue.
- **Prerendering the three app entries.** The static content pages are the cheap fix for "nothing to index"; server-rendering the React surfaces at build time is the thorough one and is a much larger change.
- **Inbound links.** A new domain with zero backlinks has no crawl reason. Linking offlinecv.org from the GitHub repo's About field is a repo-settings action, not a code change.

s-annam:
## Summary

`offlinecv.org` is verified in Search Console but none of its pages are indexed. The cause was five defects in what the site actually serves — not a Search Console setting. This fixes all five: `robots.txt` / `sitemap.xml` now exist as real generated assets, unknown paths 404 instead of returning the homepage with a 200, the staging duplicate at `dev.offlinecv.org` is marked `noindex`, the three entries carry absolute canonical + OG tags, and four plain-HTML content pages give the crawler something to read that does not require executing the bundle.

Also trims the landing block to its one guarantee and removes the footer's compressed privacy sentence — separately motivated, but it touches the same two files, so it lands here rather than in a second PR fighting for the same lines.

Closes #751

## Review focus

- `vite.config.ts` — `SEO_NOINDEX` inverted would deindex **production**, silently and for weeks. **Answered in review:** the decisions now live in `scripts/seo-artifacts.ts` as pure functions with unit tests asserting both branches, and `rollupOptions.input` derives from the same `HTML_ENTRIES` map the sitemap does, so an entry cannot go unadvertised.
- `public/privacy/index.html` — does the itemised egress list match the actual `fetch(` callsites, or did I miss a path? It claims to be exhaustive, which is a claim worth attacking. **One was missed and is now fixed in review:** JD fit's URL field calls five ATS APIs directly from the browser. It is itemised as §3, the "never leaves" line is narrowed to a *pasted* JD, and the FAQ answer + its JSON-LD copy say so too.
- `public/how-it-works/index.html` — every number (weights `0.4`/`0.3`/`0.3`, multipliers `1.0`/`0.85`/`0.7`/`0`, bands `80`/`60`) is quoted from `src/lib/score/score.ts`. Worth checking one against source rather than trusting the prose.
- `src/lib/jobs-landing.ts` — the `#saved`-hash docblock justified itself partly by what the old `single-page-application` handling did. I kept the conclusion and rewrote the mechanism; does the argument still hold under `404-page`?
- `public/content.css` duplicates token values from `src/design-system/styles/tokens.css` rather than importing them. Is the no-JS constraint (files in `public/` are copied verbatim, never processed by Vite) a good enough reason, or is there a seam I missed?

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` green — 5000 passed / 10 skipped, no regressions
- [x] `npm run verify` green (fallow report-only: flags `public/content.css` as unreachable, which is expected — it is referenced from static HTML fallow does not parse; the `App.tsx` complexity finding is pre-existing and this PR reduces that function's line count)
- [x] Default build emits `dist/robots.txt` with a `Sitemap:` line and `dist/sitemap.xml` with all 7 URLs (byte-identical to before the review changes); **no** `noindex` in any page except `404.html`
- [x] `OFFLINECV_SEO_NOINDEX=1 npm run build` emits a `Sitemap:`-free `robots.txt`, **no** `sitemap.xml`, and exactly one `noindex` meta in **all eight** built pages — the three entries via `transformIndexHtml`, the four content pages via the `closeBundle` pass, and `404.html` left with the one it already shipped (not double-tagged)
- [x] FAQ `FAQPage` JSON-LD parses, and all 12 questions match the visible `<h2>`s one for one (checked programmatically)
- [x] All four content pages, `content.css` and `404.html` present in `dist/` after a build

## Not in this PR

- **`og:image`** — no image asset exists in the repo, and referencing a missing file is worse than omitting the tag.
- **Prerendering the three app entries.** The static content pages are the cheap fix for "nothing to index"; server-rendering the React surfaces is the thorough one and is a much larger change.
- **Inbound links.** A new domain with zero backlinks has no crawl reason. Linking `offlinecv.org` from the repo's About field is a settings action, not a code change.

## Provenance

Verification: `npm run verify` — green locally via the pre-push hook, and in CI.

### DIFF
--- a/public/404.html
+++ b/public/404.html
@@ -1,0 +1,125 @@
+<!doctype html>
+<!--
+  SPDX-License-Identifier: Apache-2.0
+  Copyright 2026 The offlinecv Authors
+
+  Served for any unknown path by both deploy targets: Cloudflare Workers via
+  `"not_found_handling": "404-page"` (wrangler.jsonc) and GitHub Pages, which
+  picks up /404.html natively. Before that config change, an unknown path
+  answered 200 with the root parser page — an unbounded set of URLs all serving
+  identical content.
+
+  Deliberately self-contained: no /content.css link, no bundle. This page is
+  served AT the URL that was not found, so a relative asset path would resolve
+  against an arbitrary directory depth, and a root-absolute one is one more
+  request to fail on a page whose whole job is to fail gracefully.
+
+  The same reason makes the links below root-absolute (`/jd-fit/`) where the
+  content pages under public/ use relative ones (`../jd-fit/`) — the difference
+  is deliberate, not an oversight, so do not "fix" it to match. The trade taken
+  knowingly: root-absolute breaks under a non-`/` base, which is the case the
+  content pages' relative links survive. Both deploy targets serve from the root
+  today, and Vite never base-rewrites public/ in any case.
+-->
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Page not found — OfflineCV</title>
+    <meta name="robots" content="noindex" />
+    <meta name="color-scheme" content="light dark" />
+    <style>
+      :root {
+        color-scheme: light dark;
+        --bg: #f8fafc;
+        --fg: #0f172a;
+        --muted: #475569;
+        --accent: #1a56db;
+        --accent-hover: #1d4ed8;
+        --on-accent: #ffffff;
+      }
+      @media (prefers-color-scheme: dark) {
+        :root {
+          --bg: #0f172a;
+          --fg: #f1f5f9;
+          --muted: #a8b6c8;
+          --accent: #7bb0fa;
+          --accent-hover: #93c5fd;
+          --on-accent: #0f172a;
+        }
+      }
+      body {
+        margin: 0;
+        min-height: 100vh;
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        padding: 2rem 1.25rem;
+        background: var(--bg);
+        color: var(--muted);
+        font-family:
+          ui-sans-serif, system-ui, -apple-system, "Segoe UI", Roboto, "Helvetica Neue", Arial,
+          sans-serif;
+        line-height: 1.65;
+      }
+      main {
+        max-width: 30rem;
+        text-align: center;
+      }
+      h1 {
+        color: var(--fg);
+        font-size: clamp(1.5rem, 1.2rem + 1.5vw, 2rem);
+        line-height: 1.25;
+        margin: 0 0 0.75rem;
+      }
+      p {
+        margin: 0 0 1.5rem;
+      }
+      a.cta {
+        display: inline-block;
+        background: var(--accent);
+        color: var(--on-accent);
+        font-weight: 600;
+        text-decoration: none;
+        border-radius: 0.5rem;
+        padding: 0.7rem 1.4rem;
+      }
+      a.cta:hover,
+      a.cta:focus-visible {
+        background: var(--accent-hover);
+      }
+      ul {
+        list-style: none;
+        display: flex;
+        flex-wrap: wrap;
+        gap: 0.5rem 1.25rem;
+        justify-content: center;
+        padding: 0;
+        margin: 1.75rem 0 0;
+        font-size: 0.9375rem;
+      }
+      a {
+        color: var(--accent);
+        text-underline-offset: 0.15em;
+      }
+      a:hover,
+      a:focus-visible {
+        color: var(--accent-hover);
+      }
+    </style>
+  </head>
+  <body>
+    <main>
+      <h1>That page isn&rsquo;t here</h1>
+      <p>The link may be old, or the address may have a typo in it.</p>
+      <a class="cta" href="/">Go to the resume parser</a>
+      <ul>
+        <li><a href="/jd-fit/">Match a job description</a></li>
+        <li><a href="/jobs/">Find jobs</a></li>
+        <li><a href="/how-it-works/">How it works</a></li>
+        <li><a href="/faq/">FAQ</a></li>
+        <li><a href="/privacy/">Privacy &amp; data</a></li>
+      </ul>
+    </main>
+  </body>
+</html>
--- a/public/open-source/index.html
+++ b/public/open-source/index.html
@@ -1,0 +1,159 @@
+<!doctype html>
+<!--
+  SPDX-License-Identifier: Apache-2.0
+  Copyright 2026 The offlinecv Authors
+
+  Static content page — plain HTML, no bundle, readable with JavaScript off.
+
+  The commands quoted here are the real ones from package.json. If a script is
+  renamed there, rename it here too — a build page whose commands do not run is
+  worse than no build page.
+-->
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Open source — run your own OfflineCV | Apache 2.0</title>
+    <meta
+      name="description"
+      content="OfflineCV is Apache 2.0 licensed and its source is public. How to run it locally, why a local build carries no analytics, and where to file a parser bug."
+    />
+    <meta name="color-scheme" content="light dark" />
+    <link rel="canonical" href="https://offlinecv.org/open-source/" />
+    <meta property="og:type" content="article" />
+    <meta property="og:site_name" content="OfflineCV" />
+    <meta property="og:url" content="https://offlinecv.org/open-source/" />
+    <meta property="og:title" content="Open source — run your own OfflineCV" />
+    <meta
+      property="og:description"
+      content="Apache 2.0, public source. How to run it locally, and why a local build carries no analytics."
+    />
+    <meta name="twitter:card" content="summary" />
+    <link rel="stylesheet" href="../content.css" />
+  </head>
+  <body>
+    <div class="wrap">
+      <a class="skip" href="#main">Skip to content</a>
+
+      <nav class="nav nav--top" aria-label="Site">
+        <a href="../">Resume parser</a>
+        <a href="../jd-fit/">JD fit</a>
+        <a href="../jobs/">Find jobs</a>
+        <a href="../how-it-works/">How it works</a>
+        <a href="../faq/">FAQ</a>
+        <a href="../privacy/">Privacy &amp; data</a>
+        <strong aria-current="page">Open source</strong>
+      </nav>
+
+      <main id="main">
+        <h1>Open source</h1>
+
+        <p class="lede">
+          OfflineCV is licensed under Apache 2.0 and developed in the open at
+          <a href="https://github.com/offlinecv/OfflineCV" rel="noopener"
+            >github.com/offlinecv/OfflineCV</a
+          >. You can read it, run it, fork it, and reskin it.
+        </p>
+
+        <h2>Why that matters here specifically</h2>
+
+        <p>
+          This is a tool you hand your resume to. Every privacy claim on the
+          <a href="../privacy/">privacy page</a> is the kind of claim you should not have to take on
+          trust, and public source is what makes it checkable: open the network tab, then open the
+          code, and confirm the two agree. A closed tool making the same claim is asking for
+          something quite different from you.
+        </p>
+
+        <h2>Run it yourself</h2>
+
+        <p>
+          The app is static files &mdash; no server, no database, no environment to provision. A
+          local build runs the whole product from your own machine.
+        </p>
+
+        <div class="card">
+          <p><strong>Requirements:</strong> Node.js and npm.</p>
+          <ul>
+            <li><code>git clone https://github.com/offlinecv/OfflineCV.git</code></li>
+            <li><code>cd OfflineCV &amp;&amp; npm install</code></li>
+            <li><code>npm run dev</code> &mdash; development server</li>
+            <li><code>npm run build</code> &mdash; production build into <code>dist/</code></li>
+          </ul>
+          <p>
+            The dev server runs over HTTPS with a self-signed certificate. That is deliberate: WebGPU
+            is gated behind a secure context, so the on-device AI features need it. Accept the
+            browser warning once.
+          </p>
+        </div>
+
+        <p>
+          A build made without a PostHog key contains no analytics at all &mdash; the dependency is
+          dynamically imported and is eliminated from the bundle entirely, so it is absent rather
+          than merely disabled. The hosted site at offlinecv.org is built <em>with</em> a key; that
+          difference is described on the <a href="../privacy/">privacy page</a>.
+        </p>
+
+        <h2>Reskin it without forking</h2>
+
+        <p>
+          Colours are plain CSS custom properties in two layers: a stable semantic vocabulary
+          (<code>--color-surface-card</code>, <code>--color-content-primary</code>, and so on) and a
+          swappable file of raw values behind it. The default palette is a generic accessible one,
+          not a brand. You can override the values from your own stylesheet in the cascade, or point
+          the build&rsquo;s token alias at your own file. The component layer has an equivalent
+          swap seam. The README covers both.
+        </p>
+
+        <h2>Contributing</h2>
+
+        <p>
+          Bug reports about parsing are the most valuable contribution, and the most useful ones
+          name a specific PDF shape rather than a general complaint &mdash; a two-column template
+          whose dates vanish, a header whose email survives only as a link annotation, an employer
+          line that merges into the title above it.
+        </p>
+
+        <p>
+          Please do not attach your real resume to an issue. The repository ships a local
+          command-line probe that reads a PDF path you give it and prints only to your terminal
+          &mdash; it writes nothing and hardcodes no path &mdash; so you can characterise a bug
+          against your own private file and then report the <em>shape</em> of it. Run it with
+          <code>npm run fidelity path/to/resume.pdf</code>. Regression tests use synthetic, PII-free
+          fixtures, and a CI gate enforces that.
+        </p>
+
+        <ul>
+          <li>
+            <a href="https://github.com/offlinecv/OfflineCV/issues" rel="noopener">Issues</a>
+          </li>
+          <li>
+            <a href="https://github.com/offlinecv/OfflineCV/blob/main/README.md" rel="noopener"
+              >README</a
+            >
+            &mdash; architecture, telemetry, theming
+          </li>
+          <li>
+            <a href="https://github.com/offlinecv/OfflineCV/blob/main/LICENSE" rel="noopener"
+              >Apache 2.0 licence</a
+            >
+          </li>
+        </ul>
+
+        <a class="cta" href="https://github.com/offlinecv/OfflineCV" rel="noopener"
+          >View the source on GitHub</a
+        >
+      </main>
+
+      <nav class="nav nav--bottom" aria-label="Site">
+        <a href="../">Resume parser</a>
+        <a href="../jd-fit/">JD fit</a>
+        <a href="../jobs/">Find jobs</a>
+        <a href="../how-it-works/">How it works</a>
+        <a href="../faq/">FAQ</a>
+        <a href="../privacy/">Privacy &amp; data</a>
+        <a href="https://github.com/offlinecv/OfflineCV" rel="noopener">GitHub</a>
+      </nav>
+    </div>
+  </body>
+</html>
--- a/public/privacy/index.html
+++ b/public/privacy/index.html
@@ -1,0 +1,247 @@
+<!doctype html>
+<!--
+  SPDX-License-Identifier: Apache-2.0
+  Copyright 2026 The offlinecv Authors
+
+  Static content page — plain HTML, no bundle, readable with JavaScript off.
+
+  This page makes falsifiable claims about network egress, so every one of them
+  has to point at code: src/lib/analytics.ts (PostHog, env-gated),
+  src/lib/job-search/providers/keywords.ts (the ONLY resume-derived egress
+  helper) and its sibling adapters, src/hooks/useGitHubStars.ts, and
+  src/lib/storage/ for what is written locally. The README's "Telemetry"
+  section is the long form; this page is the short form and must not exceed it.
+
+  Do not add a claim here without grepping the actual fetch() calls first. In
+  particular there is no BYOK / hosted LLM provider in the tree — do not imply
+  one either way.
+-->
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Privacy &amp; data — what leaves your browser | OfflineCV</title>
+    <meta
+      name="description"
+      content="An itemised list of every network request OfflineCV makes, what each one carries, and what is stored in your browser. Your resume PDF and its text are not among them."
+    />
+    <meta name="color-scheme" content="light dark" />
+    <link rel="canonical" href="https://offlinecv.org/privacy/" />
+    <meta property="og:type" content="article" />
+    <meta property="og:site_name" content="OfflineCV" />
+    <meta property="og:url" content="https://offlinecv.org/privacy/" />
+    <meta property="og:title" content="Privacy &amp; data — what leaves your browser" />
+    <meta
+      property="og:description"
+      content="An itemised list of every network request OfflineCV makes, what each one carries, and what is stored in your browser."
+    />
+    <meta name="twitter:card" content="summary" />
+    <link rel="stylesheet" href="../content.css" />
+  </head>
+  <body>
+    <div class="wrap">
+      <a class="skip" href="#main">Skip to content</a>
+
+      <nav class="nav nav--top" aria-label="Site">
+        <a href="../">Resume parser</a>
+        <a href="../jd-fit/">JD fit</a>
+        <a href="../jobs/">Find jobs</a>
+        <a href="../how-it-works/">How it works</a>
+        <a href="../faq/">FAQ</a>
+        <strong aria-current="page">Privacy &amp; data</strong>
+        <a href="../open-source/">Open source</a>
+      </nav>
+
+      <main id="main">
+        <h1>Privacy &amp; data</h1>
+
+        <p class="lede">
+          The claim worth making is about custody, not about runtime: your resume PDF and the text
+          extracted from it stay in your browser. Some other things do go over the network, and this
+          page lists them.
+        </p>
+
+        <div class="card">
+          <h2>The short version</h2>
+          <p>
+            There is no account, no login, and no upload step. The PDF is read by JavaScript in the
+            tab you dropped it into; parsing, scoring, editing and PDF export all happen there. No
+            request this app makes carries the PDF bytes or the extracted resume text.
+          </p>
+          <p>
+            What does go out: the page and its assets, a job-search keyword string when you press
+            search, a public company name when you check a company&rsquo;s own board, a job posting
+            URL when you ask the app to fetch one for you, a GitHub star count, optional AI model
+            weights, and &mdash; on this hosted site &mdash; anonymous product analytics.
+          </p>
+        </div>
+
+        <h2>What never leaves</h2>
+
+        <ul>
+          <li>The PDF file itself, in whole or in part.</li>
+          <li>
+            The text extracted from it &mdash; your name, contact details, employers, titles, dates,
+            bullets, skills, education.
+          </li>
+          <li>Any edit you make to those fields, and the PDF you export from them.</li>
+          <li>
+            The job description you <strong>paste</strong> into JD fit. Asking the app to
+            <em>fetch</em> one from a posting URL is an outbound request, and it is itemised below.
+          </li>
+        </ul>
+
+        <p>
+          One helper &mdash; <code>src/lib/job-search/providers/keywords.ts</code> &mdash; is the
+          single place in the codebase where anything derived from your resume is turned into
+          outbound data, and it is deliberately kept to one file so it can be audited in one place.
+          Its output is described below.
+        </p>
+
+        <h2>Every request, itemised</h2>
+
+        <h3>1. Loading the site</h3>
+        <p>
+          Serving you the page means our host sees your IP address and user agent, the same as any
+          website. The app is static files; there is no application server to send anything to.
+        </p>
+
+        <h3>2. Job search feeds</h3>
+        <p>
+          The <a href="../jobs/">Find jobs</a> tab can query free, keyless, public job feeds. These
+          requests fire <strong>only when you click &ldquo;Search jobs&rdquo;</strong> &mdash; not
+          when you drop a resume, not when you open the tab, not while you edit the query. What is
+          sent is the short keyword string built from the query&rsquo;s title and skills fields,
+          which are shown to you and editable before you search. The feeds shipped today are
+          Remotive, Arbeitnow and Jobicy. Each is contacted directly from your browser, so
+          <strong>each sees your IP address</strong>.
+        </p>
+        <p>
+          Within Find jobs, where you ask about a specific company&rsquo;s own job board
+          (Greenhouse, Lever, Ashby), the request carries that company&rsquo;s public board name and
+          nothing else. Fetching one specific posting by its URL is a different request, described
+          next.
+        </p>
+
+        <h3>3. Fetching a job posting from its URL</h3>
+        <p>
+          Where the app offers to fetch a posting for you rather than have you paste it &mdash; the
+          URL field on <a href="../jd-fit/">JD fit</a> today &mdash; pressing
+          <strong>Fetch</strong> calls that applicant tracking system&rsquo;s own public API
+          directly from your browser. Five are supported: Greenhouse, Lever, Workable, Recruitee and
+          Ashby. The request fires <strong>only on that click</strong>, and nothing derived from
+          your resume is part of it.
+        </p>
+        <p>
+          What it carries is the company&rsquo;s public board name &mdash; which for Recruitee is
+          the hostname itself &mdash; and, for Greenhouse and Lever, the identifier of the one
+          posting you asked for. The other three return the whole board and the match happens in
+          your browser. Because the call comes from your browser,
+          <strong>that ATS sees your IP address</strong>, and for Greenhouse and Lever it also sees
+          which specific posting you are reading. That is a sharper disclosure than anything else on
+          this page, which is why it is worth pasting the description instead if you would rather
+          the ATS not see you.
+        </p>
+
+        <h3>4. GitHub star count</h3>
+        <p>
+          The footer shows the repository&rsquo;s star count via an unauthenticated call to the
+          GitHub API, cached for about an hour. It carries no data about you beyond the request
+          itself, and it fails silently. Because it originates in your browser,
+          <strong>GitHub sees your IP address</strong>.
+        </p>
+
+        <h3>5. On-device AI model weights</h3>
+        <p>
+          The optional AI features run a language model inside your browser via WebGPU. Running it
+          locally still means <em>downloading</em> it first: the weights are fetched from the public
+          CDN that the <code>@mlc-ai/web-llm</code> library points at, on the first use of an AI
+          feature, and cached by your browser afterwards. Your resume is not part of that download,
+          and once the model is loaded the prompts built from your resume are evaluated locally.
+        </p>
+
+        <h3>6. Analytics on this hosted site</h3>
+        <p>
+          This site is built with a PostHog key set, so it emits anonymous product analytics &mdash;
+          events like a file being accepted, a parse completing, or a parse failing. The payloads
+          carry file size, page count, parse duration, the score breakdown, layout triggers and
+          error names. They do not carry PDF bytes, extracted text, names or URLs. Session recording
+          and autocapture are off, and the identifier is held in memory and reset when the tab
+          closes.
+        </p>
+        <p>
+          This is a property of <em>this deployment</em>, not of the software. A build made without
+          that key removes the analytics code entirely &mdash; the dependency is dynamically
+          imported and is eliminated from the bundle. If you would rather not be counted, you can
+          <a href="../open-source/">run your own build</a>, or block the request.
+        </p>
+
+        <h3>7. Feedback, if you send it</h3>
+        <p>
+          The optional feedback panel sends a 1&ndash;5 rating, plus a category, free text and an
+          email address <strong>only where you choose to fill them in</strong>. The email field is
+          blank by default. It is attached as a property on that one event and is never turned into
+          a stored profile.
+        </p>
+
+        <h2>What is stored in your browser</h2>
+
+        <p>
+          The app writes to your browser&rsquo;s own storage so that state survives a reload. This
+          is first-party and local; none of it is transmitted, and it introduces no network requests
+          of its own. HTTP cookies are not used at all.
+        </p>
+
+        <ul>
+          <li>
+            <strong>A handful of small flags</strong> &mdash; whether the feedback prompt has been
+            shown, whether the star prompt has been dismissed, a cached star count. Booleans and
+            counters, no identifiers.
+          </li>
+          <li>
+            <strong>An IndexedDB database</strong> for the resume library and the job tracker: saved
+            resume PDFs with their cached parse, and the job records you track. This is the only
+            place your PDF is persisted, and it is on your disk.
+          </li>
+        </ul>
+
+        <p>
+          Browsers can clear this. Safari in particular clears script-writable storage after seven
+          days without a visit, and any browser may evict under disk pressure. The app asks for
+          persistent storage on first write, but that grant is best-effort, so there is a JSON
+          export in the resume library for taking your own backup. Clearing site data removes
+          everything above.
+        </p>
+
+        <h2>How to verify any of this</h2>
+
+        <p>
+          Do not take the claim on the strength of the sentence. Open your browser&rsquo;s developer
+          tools, switch to the Network tab, and use the app &mdash; drop a resume, edit it, export
+          it. The requests you see are the requests there are. The
+          <a href="https://github.com/offlinecv/OfflineCV" rel="noopener">source is public</a>, and
+          the README&rsquo;s telemetry section documents the same ground in more detail than this
+          page does.
+        </p>
+
+        <p>
+          If you find an egress path this page does not name, that is a bug in the page or the code,
+          and it is worth
+          <a href="https://github.com/offlinecv/OfflineCV/issues" rel="noopener">filing an issue</a>.
+        </p>
+
+        <a class="cta" href="../">Open the resume parser</a>
+      </main>
+
+      <nav class="nav nav--bottom" aria-label="Site">
+        <a href="../">Resume parser</a>
+        <a href="../jd-fit/">JD fit</a>
+        <a href="../jobs/">Find jobs</a>
+        <a href="../how-it-works/">How it works</a>
+        <a href="../faq/">FAQ</a>
+        <a href="../open-source/">Open source</a>
+        <a href="https://github.com/offlinecv/OfflineCV" rel="noopener">GitHub</a>
+      </nav>
+    </div>
+  </body>
+</html>
--- a/scripts/seo-artifacts.test.mjs
+++ b/scripts/seo-artifacts.test.mjs
@@ -1,0 +1,148 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+// Both directions matter here. The flag's failure mode is silent either way: a
+// production build that inverts it ships `noindex` everywhere and no sitemap and
+// still succeeds, and a staging build that loses its `noindex` re-creates the
+// duplicate-site defect. Neither shows up in the app, so asserting only the
+// happy direction would leave the expensive half uncovered.
+
+import { existsSync, readdirSync } from "node:fs";
+import { join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { describe, expect, it } from "vitest";
+
+import {
+  buildRobotsTxt,
+  buildSitemapXml,
+  HTML_ENTRIES,
+  NOINDEX_META,
+  seoHeadTags,
+  SITE_ORIGIN,
+  SITEMAP_PATHS,
+  STATIC_PAGE_PATHS,
+  withNoindexMeta,
+} from "./seo-artifacts.ts";
+
+describe("buildRobotsTxt", () => {
+  it("advertises the sitemap on a production build", () => {
+    expect(buildRobotsTxt(false)).toContain(`Sitemap: ${SITE_ORIGIN}/sitemap.xml`);
+  });
+
+  it("advertises no sitemap on a noindex build", () => {
+    expect(buildRobotsTxt(true)).not.toContain("Sitemap:");
+  });
+
+  it("allows crawling in both builds", () => {
+    // Deliberate: a `Disallow: /` staging copy could never be dropped from the
+    // index, because the crawler would not be permitted to read the directive
+    // that drops it.
+    for (const robots of [buildRobotsTxt(false), buildRobotsTxt(true)]) {
+      expect(robots).toContain("User-agent: *");
+      expect(robots).toContain("Allow: /");
+      expect(robots).not.toContain("Disallow:");
+    }
+  });
+});
+
+describe("buildSitemapXml", () => {
+  it("emits one absolute <loc> per path", () => {
+    const xml = buildSitemapXml(["/", "/faq/"]);
+    expect(xml).toContain(`<loc>${SITE_ORIGIN}/</loc>`);
+    expect(xml).toContain(`<loc>${SITE_ORIGIN}/faq/</loc>`);
+    expect(xml.match(/<loc>/g)).toHaveLength(2);
+  });
+
+  it("declares the sitemap namespace and closes the urlset", () => {
+    const xml = buildSitemapXml(SITEMAP_PATHS);
+    expect(xml.startsWith(`<?xml version="1.0" encoding="UTF-8"?>`)).toBe(true);
+    expect(xml).toContain(`xmlns="http://www.sitemaps.org/schemas/sitemap/0.9"`);
+    expect(xml.trimEnd().endsWith("</urlset>")).toBe(true);
+  });
+
+  it("carries no <lastmod>", () => {
+    // Only honoured when consistently accurate; the build timestamp changes on
+    // every redeploy whether or not the page did.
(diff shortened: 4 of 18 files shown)


Samhit21 (commented):
> **Verdict: REQUEST_CHANGES** (1 Blocking, 2 Secondary, 3 Nits). Posted as `COMMENT` per house convention — the severity is in the prose, not the review state.

The SEO machinery is right, and I checked it by running it rather than by reading it: both builds, every emitted artifact, and every number on `/how-it-works/` against `score.ts`. Thirteen of the fourteen acceptance criteria on #751 hold. The one that does not is the one your own Review focus asked about.

**Gates run on `544aa72`:** both builds (default · `OFFLINECV_SEO_NOINDEX=1`) · all 14 #751 ACs · copy-claims round-trip (`score.ts`, `analytics.ts`, `useGitHubStars.ts`, `job-search/providers/index.ts`, `scripts/fidelity.ts`, `tokens.css`) · FAQ JSON-LD ↔ visible Q&A · design-system reuse · style tokens · fixture PII **skipped, no fixture touched** · fallow **CI-green** · `/code-review` **not run** — it is disabled for model invocation, so the generic correctness pass here is hand-run, and you should weight it accordingly.

**On the test plan:** I could not reproduce `npm run verify` green locally — 9 tests in `src/lib/letter-egress-ack.test.ts` fail. They fail identically on `origin/main` with the file byte-identical between the two, and CI `verify` is green on this PR, so it is Node 25 vs `.nvmrc`'s 20 on my machine, not this branch. Your box is honest; noting it only so the next person doesn't re-derive it.

## Blocking

### 1. The itemised egress list misses a live third-party path — `/jd-fit/`'s URL fetch

`public/privacy/index.html` — see inline at `:88`.

Direct answer to Review focus bullet 2: **yes, you missed one**, and the PR body names the reason. You sourced the list from "the README's telemetry section plus `job-search/providers/`, `useGitHubStars.ts` and `analytics.ts`" — `src/lib/jd-match/fetch-jd.ts` is in none of those, and it is the module that turns a pasted URL into an outbound request to five ATS hosts. AC — *"names every egress path that exists in the tree and invents none — cross-check against the actual `fetch(` callsites, not against this issue"* — is unmet, and this PR says `Closes #751`.

## Secondary

### 1. Nothing stops a future edit inverting `SEO_NOINDEX`

`vite.config.ts:102` — see inline. This is Review focus bullet 1, and the answer is that a test is worth adding: you identified the failure mode yourself ("silently, and for weeks"), and the guard shipped for it is a checkbox in a PR body that no longer exists once this merges.

### 2. The staging `noindex` protects the pages that had nothing to index, and not the ones that do

`vite.config.ts:98` — see inline. The four content pages are the only staging URLs with real crawlable content, and they get a canonical (a hint) where the empty app shells get `noindex` (a directive). That is the risk inverted. The trade is documented in #751, so this is a "revisit", not a defect — but the fix is cheap.

## Nits (non-blocking)

### 1. The open-source page describes the CLI probe without naming it

`public/open-source/index.html:119` — see inline.

### 2. `404.html` uses root-absolute links where the content pages use relative ones

`public/404.html:108` — see inline.

### 3. Answering Review focus bullet 5 — `content.css` duplication is the right call

No change requested. I checked the ten copied custom properties against `src/design-system/styles/tokens.css` and **all ten match in both light and dark**, and the dark mechanism (`prefers-color-scheme`) matches too — so the duplication is currently honest, and the docblock explains why it has to exist. There *is* a seam you didn't mention: `emitSeoFiles`' sibling plugin could emit the `:root` blocks from `tokens.css` at build time. I don't think you should — it would split one stylesheet across two mechanisms (generated colour block, committed layout rules) and make the file unreadable in `public/`, to remove ten lines that a contrast test would catch anyway. The comment is the better guard.

### And two things worth saying that aren't findings

- Review focus bullet 4: the rewritten `jobs-landing.ts` docblock holds. Under `404-page` an unmatched path resolves to `dist/404.html`, so it still does not reach `/jobs/`, and the hash argument was never host-dependent. Conclusion and mechanism both check out.
- Review focus bullet 3: I verified every number on `/how-it-works/` against source rather than one. Weights `0.4`/`0.3`/`0.3` (`score.ts:33-35`), multipliers `1`/`0.85`/`0.7`/`0` (`score.ts:1127-1133`), bands `>= 80`/`>= 60` (`score.ts:585-586`). All exact, including the `0` scanned case.

---

public/privacy/index.html
```diff
@@ -0,0 +1,222 @@
...
+        <h2>What never leaves</h2>
+
+        <ul>
+          <li>The PDF file itself, in whole or in part.</li>
+          <li>
+            The text extracted from it &mdash; your name, contact details, employers, titles, dates,
+            bullets, skills, education.
+          </li>
+          <li>Any edit you make to those fields, and the PDF you export from them.</li>
+          <li>The job description you paste into JD fit.</li>
```
Samhit21: **[Blocking]:** the itemised list below is missing a third-party egress path that ships today — the JD-fit URL fetch — and this line is the one that makes its absence read as deliberate.

Paste an ATS URL into the field on `/jd-fit/` and press Fetch, and the browser makes a direct request to one of five hosts:

| Platform | Host contacted | What the URL carries |
|---|---|---|
| Greenhouse | `boards-api.greenhouse.io` | board slug **+ numeric job id** |
| Lever | `api.lever.co` | board slug **+ posting uuid** |
| Workable | `apply.workable.com` | account slug |
| Recruitee | `{company}.recruitee.com` | company **as the hostname** |
| Ashby | `api.ashbyhq.com` | board token |

`src/lib/jd-match/fetch-jd.ts:183-282`, reached from `JdInput.tsx:73` (`fetchJdFromUrl`), which `JdFitApp.tsx:105` renders unconditionally — the input's own placeholder names all five platforms. So the ATS learns the visitor's IP **and which specific posting they are reading**, which is a sharper disclosure than anything else on this page.

Three separate places currently tell a reader the opposite:

1. **This line** — "the job description you paste into JD fit" never leaves. True of a *pasted* JD, and the natural inference is that JD fit makes no requests at all.
2. **`:117`** — "the request carries that company's public board name **and nothing else**". That is scoped to Find jobs and it is accurate there, but it is the only Greenhouse/Lever/Ashby sentence on the page, so it is where a reader lands — and it is wrong for this path, which also carries the job id.
3. **Workable and Recruitee appear nowhere on the page.** They are contacted by no other surface, so they have no other sentence to hide behind.

`/faq/:128` has the same shape — "JD fit takes a *pasted* job description".

The page closes by inviting exactly this report (`:203`): *"If you find an egress path this page does not name, that is a bug in the page or the code."* Agreed — so it is worth fixing here rather than deferring, since the page's value is that the list is complete.

Fix: add a seventh item under **Every request, itemised** (between the current 2 and 3 reads best — it is the other outbound-on-click path), say that it fires only when Fetch is pressed, name all five platforms, and state that the request carries the company slug and the job id and exposes your IP to that ATS. Then narrow this line to "the job description you **paste** into JD fit" and add the JD-fit case to `:117` or leave `:117` explicitly scoped to Find jobs. `/faq/:128` needs the matching word change in both the visible answer and the JSON-LD copy at `:245` — they are currently identical and the AC requires they stay so.

For what it's worth, the rest of the list checks out against source: `persistence: "memory"` / `disable_session_recording: true` / `autocapture: false` (`analytics.ts:126-128`, which is also what makes the no-cookies claim true), the ~1h star TTL (`useGitHubStars.ts:17`), and feeds = Remotive/Arbeitnow/Jobicy vs company boards = Greenhouse/Lever/Ashby (`providers/index.ts:39-60`). This is the only gap I found.

vite.config.ts
```diff
@@ -74,6 +74,94 @@ const APP_VERSION = resolveAppVersion();
...
+// crawler fetching the page at all, and a URL that is *already* in the index can
+// only be dropped by a directive the crawler is permitted to read — a blocked
+// staging copy would linger indefinitely. Crawl budget on a host nothing links
+// to is not worth trading that for.
+//
+// Static pages under `public/` bypass `transformIndexHtml` (Vite copies public/
+// through verbatim), so they do not get the meta tag. They carry a hardcoded
+// absolute canonical pointing back at SITE_ORIGIN instead, which consolidates
+// the staging copy onto production the same way.
+const SEO_NOINDEX = process.env.OFFLINECV_SEO_NOINDEX === "1";
```
Samhit21, line 98: **[Secondary]:** answering Review focus bullet 1 — yes, add the test, and the reason is that this flag's failure mode is invisible on both sides.

Inverted, a production build ships `noindex` on all three entries and no sitemap, and *the build still succeeds and the site still works*. Nothing in the app surfaces it; you find out from Search Console weeks later, which is the exact latency #751 exists to end. The reverse — staging losing its `noindex` — is equally silent and re-creates defect 3.

Right now the only thing standing between an edit here and that outcome is two ticked boxes in this PR body, which stop existing the moment it merges.

The plugin doesn't need to be testable for this to be cheap — the decisions do. Lift the three pure functions out and unit-test them at module scope, the way `useSectionRewriteLock.ts` splits its logic from its subscription:

```ts
export function buildRobotsTxt(noindex: boolean): string
export function buildSitemapXml(paths: readonly string[]): string
export function seoHeadTags(noindex: boolean): HtmlTagDescriptor[]
```

Then the assertions that bite are one line each — `buildRobotsTxt(false)` contains `Sitemap:`, `buildRobotsTxt(true)` does not, `seoHeadTags(false)` is empty, `seoHeadTags(true)` carries `noindex, nofollow`. Both directions, which is what an inversion needs.

Worth one more while you are there: assert `SITEMAP_PATHS` covers every key of `rollupOptions.input`. The comment above it asks a human to keep the list in sync, and a fourth entry added later is exactly the kind of omission nobody notices — the URL is not blocked, just never advertised, so it simply stays unindexed.

vite.config.ts
```diff
@@ -74,6 +74,94 @@ const APP_VERSION = resolveAppVersion();
...
+// `<meta name="robots" content="noindex, nofollow">` to every HTML entry and
+// ships no sitemap.
+//
+// Its robots.txt still ALLOWS crawling, deliberately. `Disallow: /` stops the
+// crawler fetching the page at all, and a URL that is *already* in the index can
+// only be dropped by a directive the crawler is permitted to read — a blocked
+// staging copy would linger indefinitely. Crawl budget on a host nothing links
+// to is not worth trading that for.
+//
+// Static pages under `public/` bypass `transformIndexHtml` (Vite copies public/
```
Samhit21: **[Secondary]:** this note is correct about the mechanism, but the consequence is worth pulling out, because it lands the wrong way round.

On `dev.offlinecv.org` after this PR:

| Staging URL | Directive | Crawlable content |
|---|---|---|
| `/`, `/jobs/`, `/jd-fit/` | `noindex, nofollow` | none — empty `#root` |
| `/how-it-works/`, `/faq/`, `/privacy/`, `/open-source/` | canonical only | **the whole page** |

I confirmed this against a real `OFFLINECV_SEO_NOINDEX=1` build: `grep -c noindex` gives 1 for each of the three entries and **0** for all four content pages.

So the hard directive protects the three pages that had nothing to index anyway, and the four pages this PR added specifically to *be* indexable get a canonical — which is a hint a search engine may decline, not a directive it must obey. Those are the staging URLs most able to rank against production, which is defect 3 of #751 surviving in the half that matters.

#751 documented the trade, so this is not an unmet AC and I am not calling it Blocking. But the constraint it rests on — "static pages under `public/` bypass `transformIndexHtml`" — only holds during the bundle. By `closeBundle` the public dir has been copied and `dist/` is on disk, so the same plugin can finish the job:

```ts
closeBundle() {
  if (!SEO_NOINDEX) return;
  // dist/**/*.html copied from public/ — transformIndexHtml never saw these
  for (const f of STATIC_HTML) inject(f, '<meta name="robots" content="noindex, nofollow">');
}
```

That makes one flag mean one thing across all eight pages instead of two mechanisms with different strengths, and it removes the need for the reader of this comment to work out which pages the exception applies to.

public/open-source/index.html
```diff
@@ -0,0 +1,158 @@
...
+        <p>
+          Bug reports about parsing are the most valuable contribution, and the most useful ones
+          name a specific PDF shape rather than a general complaint &mdash; a two-column template
+          whose dates vanish, a header whose email survives only as a link annotation, an employer
+          line that merges into the title above it.
+        </p>
+
+        <p>
+          Please do not attach your real resume to an issue. The repository ships a local
+          command-line probe that reads a PDF path you give it and prints only to your terminal
```
Samhit21, line 119: **[Nit]:** name the command.

Every claim in this sentence is true — I checked it against `scripts/fidelity.ts:15-18`, which reads a path you hand it, prints to the terminal, writes nothing, and hardcodes no corpus path. The wording tracks that docblock closely, which is the right instinct.

But the paragraph is asking a contributor to do something specific instead of attaching their résumé, and it never says what to type. `fidelity` is not a guessable name for a PDF probe, and `npm run` has 20 scripts. Suggest ending the sentence with the invocation:

```
npm run fidelity path/to/resume.pdf
```

Same length, and the guidance becomes followable.

public/404.html
```diff
@@ -0,0 +1,118 @@
...
+      a:focus-visible {
+        color: var(--accent-hover);
+      }
+    </style>
+  </head>
+  <body>
+    <main>
+      <h1>That page isn&rsquo;t here</h1>
+      <p>The link may be old, or the address may have a typo in it.</p>
+      <a class="cta" href="/">Go to the resume parser</a>
```
Samhit21, line 115: **[Nit]:** these are root-absolute (`/`, `/jd-fit/`, `/faq/`) while the four content pages use relative links (`../`, `../jd-fit/`).

The file's own comment gives the reason — this page is served *at* the missing URL, so a relative path resolves against an arbitrary depth — and that reasoning is right, so the inconsistency is deliberate rather than an oversight.

The residual is only that root-absolute breaks under a non-`/` base, which is the case the content pages' relative links were chosen to survive. Both deploy targets serve from the root today (`CNAME` on Pages, Workers assets on the apex), so nothing is broken now, and `public/` is never base-rewritten by Vite in any case.

Not worth changing — but worth one clause in the comment saying the trade was taken knowingly, so the next person comparing the two link styles doesn't "fix" the wrong one.

public/privacy/index.html
s-annam: Fixed in `08315a9`. You're right that it ships, and I checked the one thing that could have made it moot before changing anything: `jd-fit-banner` (`VITE_ENABLE_JD_FIT`) gates only the cross-sell banner on `/`, not the surface — `/jd-fit/` is a real entry in `rollupOptions.input`, this PR puts it in `sitemap.xml`, this page's own footer nav links to it, and `JdInput` renders the Fetch field unconditionally. So a visitor reaches it and the requests fire.

What landed:

- **New §3, "Fetching a job posting from its URL"** — between the old 2 and 3 as you suggested; 3–6 renumbered to 4–7. Names all five platforms, says it fires only on the Fetch click, and states the IP + posting disclosure.
- **`:88` narrowed** to the job description you **paste** into JD fit, with a pointer to the itemised entry for the fetch case.
- **`:117` scoped** — now opens "Within Find jobs, …" and ends by routing the by-URL case to §3.
- **The short version** gained the posting URL, so the summary is no longer narrower than the list under it.
- **`/faq/`** — the visible answer and the JSON-LD copy at `:245` both changed, still byte-identical prose.

One correction to your table, which I ran down against `fetch-jd.ts` before writing the copy: only **Greenhouse** (`/boards/{company}/jobs/{jobId}`) and **Lever** (`/postings/{company}/{jobId}`) put the posting id in the outbound URL. Workable (`/widget/accounts/{company}`), Recruitee (`{company}.recruitee.com/api/offers`) and Ashby (`/job-board/{token}`) each fetch the **whole board** and match client-side — so those three leak the company but not which posting you're reading. The copy splits it that way rather than saying "company slug and job id" across all five, since the weaker claim is the true one for three of them.

I also wrote it around the capability rather than the surface — "where the app offers to fetch a posting for you … the URL field on JD fit today" — because this logic is slated to migrate into `/jobs/` as JD fit is deprecated. That way the move doesn't silently falsify the page again.

Agreed on the framing, for what it's worth: `:203` invites exactly this report, so a list that is 6/7 complete is worse than one that doesn't claim to be exhaustive. Thanks for reading it against the callsites instead of the prose.

vite.config.ts
s-annam: Added in `08315a9`. Agreed on the reasoning — two ticked boxes in a PR body stop existing at merge, and the flag is silent in both directions.

Took the shape you described. The three decisions moved out to **`scripts/seo-artifacts.ts`** as pure functions, tested in **`scripts/seo-artifacts.test.mjs`** (15 tests, matching the existing `scripts/**/*.test.mjs` include so no vitest config churn; added to the coverage `include` enumeration per that block's own instruction). Both directions, as you asked:

- `buildRobotsTxt(false)` contains `Sitemap: …`, `buildRobotsTxt(true)` does not, and **both** keep `Allow: /` with no `Disallow:` — that last one pins the non-obvious half of the design so a future edit can't "tidy" staging into `Disallow: /`.
- `seoHeadTags(false)` is `[]`; `seoHeadTags(true)` is the exact `noindex, nofollow` head-prepend descriptor.
- `buildSitemapXml` — one absolute `<loc>` per path, namespace declared, and **no `<lastmod>`**, which was previously only a comment.
- `withNoindexMeta` (the new static-page injector from your other thread) injects into `<head>`, and returns `null` for a page that already declares a robots directive, so `404.html` and the bundled entries aren't double-tagged.

On the `SITEMAP_PATHS` ⊇ `rollupOptions.input` assertion — I went one step further and made it structural instead. `HTML_ENTRIES` is now the single source of both: `vite.config.ts` derives `rollupOptions.input` from it, and `SITEMAP_PATHS` derives from it too. A fourth entry is advertised by construction, so there's no sync comment left to obey. That leaves exactly one hand-kept list — the static pages under `public/`, which aren't declared to the bundler and so can't be derived — and the test asserts *that* one against the directories actually on disk:

```js
expect([...STATIC_PAGE_PATHS].sort()).toEqual(onDisk.sort());
```

which fails in both directions: adding `public/pricing/index.html` without listing it, or listing a path whose directory got deleted.

Verified `dist/sitemap.xml` is byte-identical to before the refactor — same 7 URLs, same order.

vite.config.ts
s-annam: Fixed in `08315a9` — took the `closeBundle` suggestion. You're right that the constraint only held during the bundle, and that the note landed the wrong way round: the hard directive was protecting the three pages with nothing to index, while the four pages added specifically to *be* indexable got a hint a crawler may decline.

`emitSeoFiles` now walks the built HTML in `closeBundle` and injects the directive via the pure `withNoindexMeta` helper. Rather than a hand-kept `STATIC_HTML` list, it recurses `outDir` for `*.html` and skips any file that already declares a robots meta — so `404.html` (which ships its own `noindex`) and the three entries (already tagged by `transformIndexHtml`) pass through untouched, and a page added under `public/` later is covered without anyone remembering this hook.

Reproducing your check on the new build:

| | entries | content pages | 404.html |
|---|---|---|---|
| `OFFLINECV_SEO_NOINDEX=1` | 1 each | **1 each** (was 0) | 1, not 2 |
| default | 0 | 0 | 1 (its own) |

The comment at `:98` no longer describes the public/ pages as an exception, because they aren't one any more — one flag, one meaning, eight pages. The canonical stays on them regardless; it just isn't load-bearing for this now.

`withNoindexMeta` is unit-tested both ways (injects into a bare `<head>`, returns `null` on an already-tagged page and on a file with no head), so the skip condition can't silently invert into double-tagging.

public/open-source/index.html
s-annam: Fixed in `08315a9` — the sentence now ends `Run it with `npm run fidelity path/to/resume.pdf`.` Fair point that `fidelity` is not guessable from "local command-line probe", and asking someone to do a specific thing without naming it is the one way this paragraph could fail at its job.

public/404.html
s-annam: Fixed in `08315a9` — added the clause rather than the link change, which matches your read.

The header comment now says the difference from the content pages is deliberate and names the residual you identified:

> The same reason makes the links below root-absolute (`/jd-fit/`) where the content pages under `public/` use relative ones (`../jd-fit/`) — the difference is deliberate, not an oversight, so do not "fix" it to match. The trade taken knowingly: root-absolute breaks under a non-`/` base, which is the case the content pages' relative links survive. Both deploy targets serve from the root today, and Vite never base-rewrites `public/` in any case.

That's the useful half of this thread preserved in the file — the next person comparing the two styles now finds the reasoning instead of re-deriving it and picking the wrong one to change.

s-annam (commented):
## Review Summary

All five SEO defects from issue #751 have been successfully addressed. Production build outputs are fully indexable and contain correct canonical and Open Graph tags, while staging builds (`OFFLINECV_SEO_NOINDEX=1`) correctly serve a sitemap-free `robots.txt` and `<meta name="robots" content="noindex, nofollow">` on all HTML pages (including the static content pages, without double-tagging `404.html`). 

Constituent verification steps (`typecheck`, `lint`, `check:fixtures`, `check:baselines`, and full build) pass clean. The new unit tests covering sitemap, robots, and injection logic pass successfully.

### 🏁 Gates & Checklist

* **3a — Fixture PII:** Skipped (no binary fixtures touched).
* **3b — Design-system & Reuse:** Passed. The footer updates use native links rather than hand-rolled components.
* **3c — Style Tokens:** Passed. Tailwind class overrides use semantic tokens. The CSS values duplicated in `public/content.css` are well-documented.
* **3d — Dead-code / Fallow:** Passed. Fallow audit flags no dead exports or unused files (other than `public/content.css`, which is static and expected).
* **3e — Command-level Bugs:** Passed. Pure JS/TS implementation with no shell execution logic.
* **3f — Description Accuracy:** Accurate. All summaries and checklists match the actual diff.

---

scripts/seo-artifacts.test.mjs
```diff
@@ -0,0 +1,134 @@
...
+  });
+
+  it("advertises every static page directory under public/", () => {
+    // The one list that is hand-kept — public/ pages are copied through, not
+    // declared to the bundler, so nothing else can derive them. A page added
+    // here and forgotten is not blocked from indexing, just never advertised,
+    // so it silently stays unindexed.
+    const publicDir = fileURLToPath(new URL("../public", import.meta.url));
+    const onDisk = readdirSync(publicDir, { withFileTypes: true })
+      .filter((entry) => entry.isDirectory())
```
Samhit21: **[Nit]:** this treats *every* directory under `public/` as a page that belongs in the sitemap.

Correct today — `ls -d public/*/` gives exactly the four content pages, and I ran the mutation to confirm the assertion bites (dropping `"/privacy/"` from `STATIC_PAGE_PATHS` fails it). So nothing needs to change now.

The trap is the first non-page directory. Add `public/images/` or `public/assets/` for an `og:image` — which the PR body lists under "Not in this PR", so it is the likely next one — and this test fails with:

```
expected [ '/faq/', '/how-it-works/', … ] to deeply equal [ '/assets/', '/faq/', … ]
```

That reads as "you forgot to list `/assets/`", and the obvious way to make it green is to add it — which advertises a directory with no `index.html` in `sitemap.xml`. The failure points at the wrong fix, which is worse than not failing, because a test that has to be argued with tends to lose.

One predicate keeps the check honest:

```js
const onDisk = readdirSync(publicDir, { withFileTypes: true })
  .filter((e) => e.isDirectory() && existsSync(join(publicDir, e.name, "index.html")))
  .map((e) => `/${e.name}/`);
```

Now it asserts what the comment above it already says it asserts — *pages* — and an asset directory is silently, correctly ignored while a real page added and forgotten still fails.

scripts/seo-artifacts.ts
```diff
@@ -0,0 +1,141 @@
...
+ * canonical — a hint a search engine may decline — on the four content pages
+ * that carry the actual crawlable prose. Those are the staging URLs most able to
+ * rank against production, which is the whole defect.
+ *
+ * Returns `null` when the file already carries a robots meta (404.html ships its
+ * own `noindex`) or has no `<head>` to inject into, so the caller can skip the
+ * write rather than double-tagging.
+ */
+export function withNoindexMeta(html: string): string | null {
+  if (/<meta\s+name=["']robots["']/i.test(html)) return null;
```
Samhit21: **[Nit]:** `/<meta\s+name=["']robots["']/i` requires `name` to be the **first** attribute, so the skip is attribute-order sensitive.

Every robots meta in the tree today is written `name` first — `404.html`'s hand-written one, and the one Vite renders from `seoHeadTags` — so this is correct as it ships, and your unit test covers the shapes that actually occur. I verified the real staging build: `404.html` comes out with 1, not 2.

What makes it worth a line is that the failure is silent and lands on the file you least want it on. Someone hand-editing `404.html` to `<meta content="noindex" name="robots" />` — equivalent HTML, no reason to think twice — gets two robots metas in the staging build. Crawlers do tolerate that, so nothing breaks loudly; it just makes the page's own directive ambiguous to a human reading `dist/`, and the docblock's promise ("so the caller can skip the write rather than double-tagging") quietly stops holding.

Order-independent version, same shape:

```js
if (/<meta\b[^>]*\bname=["']robots["']/i.test(html)) return null;
```

Worth one more test case alongside it (`content` first → still `null`) so the property is pinned rather than just the current spelling. Genuinely optional — the constraint is only violable by a future hand-edit.

Samhit21 (dismissed):
> **Verdict: APPROVE** (0 Blocking). All six findings from the previous pass are fixed, and I re-verified each by running it rather than by reading the replies. Two new Nits below are genuinely non-blocking — post-merge or never.

**Gates run on `08315a9`:** both builds re-run from clean · robots-meta counted per page on all 8 pages in both modes · `sitemap.xml` diffed against the pre-refactor build · **mutation-tested the new guards** · full suite (5010 tests, +15) · typecheck · lint · CI **verify / fallow / Cloudflare Pages all green** · fixture PII **skipped, no fixture touched** · `/code-review` **still not run** — disabled for model invocation, so this remains a hand-run correctness pass.

### First, a correction to my own review

My table claimed all five ATS fetchers put the posting id in the outbound URL. That was wrong for three of them, and your reply has it right: only Greenhouse (`/boards/{company}/jobs/{jobId}`) and Lever (`/postings/{company}/{jobId}`) carry it. Workable, Recruitee and Ashby each GET the whole board and match client-side. The page now splits it that way, which is the accurate claim and the narrower one — good catch on a review comment.

### Blocking 1 — the JD-fit egress path

Fixed and verified. New §3 names all five platforms, scopes the request to the Fetch click, and states the IP disclosure plus the posting-level disclosure for the two platforms it actually applies to. `:88` now distinguishes paste from fetch, `:117` opens "Within Find jobs" and routes the by-URL case onward, and the short version gained the posting URL so the summary is no longer narrower than the list beneath it.

I re-ran the FAQ check after the copy change: **12 `<h2>`s, 12 JSON-LD questions, identical in order, and 0 answer divergences** — the changed answer stayed byte-identical across both copies, which is the AC that a partial edit would have broken.

Writing it around the capability ("where the app offers to fetch a posting for you — the URL field on JD fit today") rather than the surface is the right call given the `/jobs/` migration. That is the difference between copy that survives a refactor and copy that silently goes stale again.

### Secondary 1 — the inversion guard

Fixed, and I checked the tests **bite** rather than merely pass. Three mutations against `scripts/seo-artifacts.ts`, each reverted after:

| Mutation | Result |
|---|---|
| `seoHeadTags`: `if (!noindex)` → `if (noindex)` (**the production-deindex case**) | ✅ 2 failed |
| `buildRobotsTxt`: invert the `Sitemap:` line | ✅ 2 failed |
| drop `"/privacy/"` from `STATIC_PAGE_PATHS` | ✅ 1 failed |

Clean tree: 15/15 pass. So the guard is real, not decorative — which was the entire point of the finding.

Making `HTML_ENTRIES` the single source of both `rollupOptions.input` and `SITEMAP_PATHS` is better than the assertion I suggested: it removes the failure instead of detecting it. I confirmed the refactor is behaviour-preserving — `dist/sitemap.xml` is byte-identical to the pre-refactor build, same 7 URLs in the same order.

### Secondary 2 — staging `noindex` on the static pages

Fixed via `closeBundle`, and this is the check I most wanted to see hold, since the injector is new code acting on files after the bundle:

| Page | default build | `OFFLINECV_SEO_NOINDEX=1` |
|---|---|---|
| 3 bundled entries | 0 | 1 each |
| 4 content pages | 0 | **1 each** (was 0) |
| `404.html` | 1 (its own) | **1, not 2** |

One flag, one meaning, eight pages — and production is not deindexed, which is the direction that would have been expensive to get wrong. Recursing `outDir` and skipping already-tagged files is better than the hand-kept `STATIC_HTML` list I sketched: a page added under `public/` later is covered without anyone remembering this hook exists.

### Nits 1 and 2 — both fixed

`npm run fidelity path/to/resume.pdf` now appears in the sentence, and `404.html`'s header comment records the root-absolute-vs-relative trade so the next reader finds the reasoning instead of re-deriving it.

## Nits (non-blocking, new)

Two latent traps in the new code. Neither is live today, neither should hold the merge — I checked both against the current tree and they are correct as written.

1. `scripts/seo-artifacts.test.mjs:119` — the on-disk check treats every directory under `public/` as a page. See inline.
2. `scripts/seo-artifacts.ts:136` — the already-tagged check is attribute-order sensitive. See inline.

### On the test suite, unchanged from last time

`npm run test` shows 9 failures in `src/lib/letter-egress-ack.test.ts` on my machine. Same 9 on `origin/main`, file byte-identical, and CI `verify` is green on `08315a9` — it is Node 25 vs `.nvmrc`'s 20 locally. Totals moved 4995 → 5010, which is exactly the 15 new SEO tests and no regression.

---

scripts/seo-artifacts.test.mjs
s-annam: Fixed in `a1f1003` — took the predicate.

```js
.filter((entry) => entry.isDirectory() && existsSync(join(publicDir, entry.name, "index.html")))
```

You called the failure mode exactly, so I reproduced it both ways before and after rather than trusting the reasoning. With a bare `public/images/` on disk:

| | old predicate | new predicate |
|---|---|---|
| `public/images/` (no `index.html`) | **fails** — `- "/images/"` | passes, silently ignored |
| `/privacy/` dropped from `STATIC_PAGE_PATHS` | fails | **fails** — `- "/privacy/"` |

The second row is the one I wanted to see: the predicate could have bought asset-tolerance by weakening the check into something a real missing page also slips through. It doesn't — a page that exists on disk and isn't listed still fails, which is the whole reason the test is here.

The failure message was the substance of this, and it's worth restating why it mattered more than the test being wrong: the old one was *correct today* and would have stayed correct until the exact moment someone added the og:image directory the PR body already promises. At that point it fails with a diff that reads as an instruction, and the instruction is wrong. "A test that has to be argued with tends to lose" is the right frame — the cost isn't the bad fix, it's that the arguing happens at all, months later, by someone with no context on which of the two lists is authoritative.

Left a comment above the filter naming the og:image case specifically, so the predicate doesn't look like defensive boilerplate someone could tidy away.

scripts/seo-artifacts.ts
s-annam: Fixed in `a1f1003` — took your regex, and added the test case you suggested alongside it.

```js
if (/<meta\b[^>]*\bname=["']robots["']/i.test(html)) return null;
```

I wrote the test first and ran it against the old pattern to confirm it actually pins the property rather than restating the passing one:

```
× withNoindexMeta > skips a robots directive whose name is not the first attribute
  Tests  1 failed | 15 passed (16)
```

Then green on the new pattern. 16 tests now.

Your read on the severity is right and it's the reason I didn't defer this despite the "genuinely optional" — the damage isn't the double tag, it's that the docblock makes a promise ("so the caller can skip the write rather than double-tagging") that the implementation only keeps for one spelling of the input. The doc is what the next person reads; if it's true only by coincidence of how `404.html` happens to be written today, then the file is lying about its own contract and nothing catches it. Attribute order carries no meaning in HTML, so the code shouldn't act as though it does — I put that reasoning in a comment on the line rather than leaving the `\b[^>]*\b` looking like an unexplained regex flourish.

Re-ran the full staging build to confirm nothing moved:

| | entries | content pages | 404.html |
|---|---|---|---|
| `OFFLINECV_SEO_NOINDEX=1` | 1 each | 1 each | 1, not 2 |
| default | 0 | 0 | 1 (its own) |

`dist/sitemap.xml` still 7 URLs in the same order; `robots.txt` unchanged.

s-annam:
## Merging with admin bypass

Both nits from the last round are fixed in `a1f1003` and their threads are resolved. All three checks are green — `verify` (3m3s), `fallow`, Cloudflare Pages. The only thing holding the PR at `BLOCKED` is `REVIEW_REQUIRED`, i.e. an approval that the last push dismissed. Rather than spend another review round on two nits that were already reviewed once and are each a one-line change, I'm using the admin bypass and sending anything further to a follow-up PR.

Recording what landed, since the resolved threads are collapsed by default and this is the last state anyone reading `main` will see.

### 1. `scripts/seo-artifacts.test.mjs` — the public/ assertion now checks *pages*, not directories

The `SITEMAP_PATHS` test treated every directory under `public/` as a page that belongs in the sitemap. Correct as it shipped, but the first asset directory breaks it — and breaks it in the worst possible way, by failing with a diff that reads as an instruction to advertise `/images/` in `sitemap.xml`. Since the PR body lists an og:image under "Not in this PR", that directory is the likely next one.

```js
.filter((entry) => entry.isDirectory() && existsSync(join(publicDir, entry.name, "index.html")))
```

Reproduced both directions against a real bare `public/images/` on disk rather than reasoning about it:

| | old predicate | new predicate |
|---|---|---|
| `public/images/` with no `index.html` | **fails** — `- "/images/"` | passes, ignored |
| `/privacy/` dropped from `STATIC_PAGE_PATHS` | fails | **fails** — `- "/privacy/"` |

The second row is the one worth having: a predicate that bought asset-tolerance by loosening the check into something a genuinely missing page also slips through would have been a worse trade than the original. It doesn't — a page on disk that nobody listed still fails, which is the only reason the test exists.

### 2. `scripts/seo-artifacts.ts` — the robots-meta skip no longer depends on attribute order

`/<meta\s+name=["']robots["']/i` required `name` to be the first attribute, so `<meta content="noindex" name="robots">` — identical HTML, no reason for anyone to think twice — would miss the skip and get a second robots meta injected in a staging build.

```js
if (/<meta\b[^>]*\bname=["']robots["']/i.test(html)) return null;
```

The reason this was worth taking rather than deferring: the failure is silent and lands on `404.html`, and the docblock promises "so the caller can skip the write rather than double-tagging" — a promise the old implementation kept only for one spelling of the input. Wrote the test first and ran it against the old pattern to confirm it pins the property instead of restating a passing case:

```
× withNoindexMeta > skips a robots directive whose name is not the first attribute
  Tests  1 failed | 15 passed (16)
```

Green on the new pattern. Both changes carry a comment naming the case, so neither looks like unexplained defensive boilerplate a later reader could tidy away.

### Verification

`npm run typecheck` clean, `eslint .` clean, `vitest run` 5001 passed / 10 skipped (16 in `seo-artifacts.test.mjs`, up from 15). Rebuilt both ways to confirm the SEO artifacts did not move:

| | entries | content pages | `404.html` |
|---|---|---|---|
| `OFFLINECV_SEO_NOINDEX=1` | 1 each | 1 each | 1, not 2 |
| default | 0 | 0 | 1 (its own) |

`dist/sitemap.xml` still 7 URLs in the same order; `dist/robots.txt` unchanged.

`fallow` reports two findings, neither from this delta and both pre-existing: `public/content.css` flagged unused (it is linked from the four static pages, which fallow does not trace as entry points) and `App.tsx` complexity, which `CLAUDE.md` already lists as known debt. Report-only by design.

Thanks @Samhit21 — both of these were correct-as-shipped code whose next edit would have gone wrong quietly, which is the harder kind to catch than an outright bug.

PR #755, opened 2026-08-05
fix(resume-library): rebuild a record whose cached parse is missing

### LINKED ISSUE #758: loadResumeFromLibrary refuses a record with source bytes but no cached parse, silently zeroing every saved-job rating
## Problem

`loadResumeFromLibrary` refuses any `ResumeRecord` whose `parse` slot is empty, even when the record holds a perfectly good source PDF. On a real 541-record library this silently removed the fitness block from **every** row of the Saved-jobs tab on `/jobs/`, including rows that had healthy `jdText`.

Observed 2026-08-04 against `offlinecv.org` with a library whose single résumé record carried a valid 25 KB PDF blob and no `parse`.

## Root cause

`src/lib/resume-library.ts:167` (pre-fix):

```ts
const snap = readSnapshot(record.parse);
if (snap === null) return undefined;
```

`readSnapshot` returns `null` when `parse.result` or `parse.score` is missing. That branch fires before anything looks at `record.blob`, so a record carrying the source PDF is discarded on the grounds that its *cache* is empty.

### Why such a record exists

#693 put a public write door on the `resumes` store (`docs/job-capture-contract.md`, `job-record-contract.ts`). A producer outside this build — the browser extension's sync pull, a backup import — writes a `ResumeRecord` carrying the PDF and **no** `parse`, because a producer cannot run the cascade. That is the documented, intended shape of an imported record, not a corruption.

`listLibrary` already handles it correctly (`resume-library.ts:141-142` keeps a malformed-snapshot record in the list at score 0 so the user can still delete it). Only the load path refuses.

### Why the recovery already existed

Ten lines below, the stale-shape guard (#445 / #321) re-parses from the stored blob for exactly the same underlying reason — *the cached parse cannot be used as it stands*. A record with no cached parse is the same situation with a different cause, and wants the same recovery.

## Why it was invisible

`src/hooks/useFallbackResume.ts` (#724) loads the most recently saved library résumé so the Saved-jobs tab can rate against something on a direct visit:

```ts
.then((loaded) => { if (cancelled || loaded === undefined) return; ... })
```

The load **resolves** rather than rejecting, so there is no console line, no error state, and no thrown exception. The hook simply sets no fallback, `hasResume` stays false, and `JobTracker` renders the "Open this workbench from your resume" prompt — against a résumé sitting in the library the whole time.

Three separate surfaces therefore agree on a wrong answer with nothing to distinguish it from "you have no résumé":

1. every saved job loses its stars,
2. the empty-state prompt tells the user to do something they already did,
3. nothing is logged.

## Data flow

```
resumes store record { blob: <25 KB PDF>, parse: undefined }
  └→ useFallbackResume (hooks/useFallbackResume.ts)
       └→ loadResumeFromLibrary (lib/resume-library.ts:167)
            └→ snap === null → return undefined        ← the bug
  └→ fallback stays undefined → hasResume false
       └→ useSavedJobRatings returns null
            └→ JobTracker drops the fitness block from every row
```

## Fix

Fold the `snap === null` case into the existing stale-shape re-parse branch, so both "cannot use the cached parse" causes take the same recovery:

```ts
const bytes = record.blob.size > 0 ? await record.blob.arrayBuffer() : undefined;

if (snap === null || snap.shapeVersion !== CACHE_SHAPE_VERSION) {
  if (bytes === undefined) return undefined;
  const sourceKind = snap?.sourceKind ?? sourceKindOfBlob(record.blob);
  if (sourceKind !== "pdf") return undefined;
  const result = await runCascade(bytes);
  ...
}
```

Two supporting pieces:

- **`sourceKindOfBlob(blob)`** — a record with no snapshot has nothing that states its kind, so derive it from the blob's MIME (reverse of the existing `MIME` map). Unknown/blank types read as `pdf`, which is what every producer that writes bytes writes today.
- **`sourceKind !== "pdf"` guard** — `runCascade` reads PDF bytes. A non-PDF record with no usable snapshot degrades to "not loadable" rather than throwing inside the cascade.

Refusal is then reserved for the two cases that genuinely cannot be rebuilt: no bytes at rest (a DOCX record, whose source bytes are not kept), or bytes the PDF cascade cannot read.

The rebuilt snapshot is re-stamped at `CACHE_SHAPE_VERSION` (the existing #452 re-save), so the re-parse cost is paid once per record rather than on every load.

## Acceptance criteria

- [ ] A `resumes` record with a valid PDF blob and no `parse` loads, re-parsing from the blob rather than returning `undefined`.
- [ ] The rebuilt record is re-stamped, so a second `loadResumeFromLibrary` on the same id does **not** call `runCascade` again.
- [ ] A parse-less record with an empty blob still returns `undefined` and never calls `runCascade`.
- [ ] A parse-less record whose blob is not a PDF (e.g. the DOCX MIME) returns `undefined` and never calls `runCascade`.
- [ ] The pre-existing stale-shape recovery (#445 / #321) is unchanged for a PDF record: it still re-parses and still returns the fresh canonical result.
- [ ] A stale-shape record with no blob still returns `undefined` (regression).
- [ ] `listLibrary` behaviour is unchanged — a malformed-snapshot record still appears at score 0.
- [ ] `npm run verify` green.

## Fix status

Implemented in **#755** (`fix/resume-library-load-parseless-record`) — `src/lib/resume-library.ts` + 4 new tests in `src/lib/resume-library.test.ts`, `verify` green. This issue is filed after the fact to give that PR a home; it was root-caused live in a session rather than from an existing report.

## Related

- **#754** — saved-job duplicate detection degenerates when `jdText` is absent. Independent bug, same investigation.
- **#724** — introduced `useFallbackResume`, the consumer this bug disabled.
- **#693** — introduced the public write door that makes a parse-less record a normal shape.
- **#445 / #321** — the stale-shape guard whose recovery this reuses.

s-annam:
## Summary

`loadResumeFromLibrary` refused any `resumes` record with an empty `parse` slot — `if (snap === null) return undefined` — even when the source PDF was sitting in the record's blob. Since #693 made the store a public write door, that record is routine: a producer outside this build imports a `ResumeRecord` with the PDF and no parse, because it cannot run the cascade itself.

The damage showed up on `/jobs/`. `useFallbackResume` (#724) loads the newest library résumé so the Saved-jobs tab can rate against something on a direct visit. That load resolved `undefined`, so no fallback was set, `hasResume` stayed false, and **every** row lost its fitness block — including rows with a healthy `jdText`. And it was silent: the promise resolved rather than rejected, so the hook's `.catch` never fired, and the tracker rendered "Open this workbench from your resume" against a résumé already in the library.

Found live on a 541-record library whose only résumé record had a valid 25 KB PDF blob and no parse.

The fix is to stop treating "no cached parse" as a different problem from "stale cached parse". The stale-shape guard (#445 / #321) ten lines below already re-parses from the stored blob for the same underlying reason — the cached parse cannot be used as it stands — so both now take that path. Refusal is reserved for records that genuinely cannot be rebuilt: no bytes at rest (DOCX), or bytes the PDF cascade cannot read.

Closes #758.

## Review focus

- `src/lib/resume-library.ts:202` — the branch condition is now `snap === null || snap.shapeVersion !== CACHE_SHAPE_VERSION`. Does every read of `snap` *after* the branch still narrow to non-null, given the branch returns on all paths?
- `src/lib/resume-library.ts:174` — `sourceKindOfBlob` defaults an unknown or blank `blob.type` to `pdf`. Is that the right default for a producer-written record, or should a blank type refuse?
- `src/lib/resume-library.ts:208` — the new `sourceKind !== "pdf"` guard also applies to the *stale-shape* path, which previously called `runCascade` regardless of kind. A stale markdown record with bytes now returns `undefined` where it used to attempt a parse. Is that a behaviour change worth calling out, or was that path already dead?

## Test plan

- [x] `npm run test` green — `src/lib/resume-library.test.ts` 12/12, stable across 3 runs
- [x] Fail-before proven: 3 tests fail with `resume-library.ts` reverted, pass with it
- [x] `npm run verify` green locally (typecheck, lint, fixture PII, baselines, coverage, build)
- [ ] Manually verified against a real parse-less record on a deployed build

No fixture binaries added or changed.

fallow reports one 11-line clone group, both instances in `resume-library.test.ts`: the assertion blocks of the two sibling recovery tests. Left duplicated on purpose — an extracted assertion helper would hide which outcome each scenario actually checks. The larger 77-line group from the first draft is gone; `reparsedResult()` is now shared by both paths.

## Provenance

Verification: `npm run verify` — green locally and in CI

### DIFF
--- a/src/lib/resume-library.test.ts
+++ b/src/lib/resume-library.test.ts
@@ -47,6 +47,31 @@ const result = () =>
   }) as unknown as CascadeResult;
 const score = (overall: number) => ({ overall }) as AnonymousAtsScore;
 
+/** What the mocked cascade hands back on a re-parse, tagged "Reparsed Persona"
+ *  so a test can prove the loaded result came from the blob rather than from a
+ *  snapshot. Shared by every re-parse path — the stale-shape guard and the
+ *  no-cached-parse recovery both land on the same recovery code. */
+const reparsedResult = () =>
+  ({
+    canonical: toCanonicalResume(
+      { full_name: "Reparsed Persona", skills: [], experience: [], education: [] },
+      {
+        byName: new Map(),
+        accomplishmentSections: ACCOMPLISHMENT_SECTION_NAMES,
+        source: "regex",
+      },
+      {},
+    ),
+    confidence: 0,
+    triggers: [],
+    suggestedEscalation: "none",
+    tiers: ["t0_layout", "t1_openresume"],
+    rawText: "",
+    linkAnnotations: [],
+    diagnostics: { rawCharCount: 0, extractedCharCount: 0, pages: 1, elapsedMs: 0 },
+    timings: { t0_layout_ms: 0, t1_openresume_ms: 0 },
+  }) as unknown as CascadeResult;
+
 async function save(filename: string, overall = 72) {
   return saveResumeToLibrary({
     filename,
@@ -111,27 +136,7 @@ describe("resume-library: rename + delete", () => {
 
 describe("resume-library: cache-version mismatch (#445 / #321)", () => {
   it("re-parses from the stored blob instead of deserializing a stale-shape record", async () => {
-    // A re-parsed canonical result the mocked cascade returns, tagged so we can
-    // prove the loaded result came from the re-parse, not the stale snapshot.
-    const reparsed = {
-      canonical: toCanonicalResume(
-        { full_name: "Reparsed Persona", skills: [], experience: [], education: [] },
-        {
-          byName: new Map(),
-          accomplishmentSections: ACCOMPLISHMENT_SECTION_NAMES,
-          source: "regex",
-        },
-        {},
-      ),
-      confidence: 0,
-      triggers: [],
-      suggestedEscalation: "none",
-      tiers: ["t0_layout", "t1_openresume"],
-      rawText: "",
-      linkAnnotations: [],
-      diagnostics: { rawCharCount: 0, extractedCharCount: 0, pages: 1, elapsedMs: 0 },
-      timings: { t0_layout_ms: 0, t1_openresume_ms: 0 },
-    } as unknown as CascadeResult;
+    const reparsed = reparsedResult();
     vi.mocked(runCascade).mockResolvedValue(reparsed);
 
     // Write a pre-cutover record DIRECTLY through the storage layer: a stale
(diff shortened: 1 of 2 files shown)


s-annam (commented):
An elegant and robust fix for the library load issue. The solution correctly unifies the "stale cached parse" and "no cached parse" recovery paths by leveraging the existing PDF blob re-parse mechanism. The implementation is highly defensible, and the test coverage is exceptionally thorough.

### Verdict
`APPROVE` (0 Blocking findings) — *Submitted as a comment review because GitHub does not allow approving one's own PR.*

### Gate Results
- **Fixture PII**: Skipped (no fixture files touched).
- **Design System & Component Reuse**: Skipped (no UI components touched).
- **Tailored Styles & Colors**: Skipped (no styles touched).
- **Dead Code / Fallow**: Checked (`fallow audit` completed; no new dead code or unused exports in the diff).
- **Script & Skill Command Integrity**: Skipped (no script or skill markdown files touched).
- **PR Description Accuracy**: Verified (highly accurate description of the defect, root cause, and recovery path).

---

Issue #776, opened 2026-08-07
Measure steering adherence on Gemma 2 (2B) and Llama 3.2 (3B) — the two registry models #608 could not cache


Samhit21:
## Problem

#608 half 2 asked whether the rewrite ignores the user's typed instructions. It is now **answered for the default model and only the default model**.

The 2026-08-07 eval run (`tests/fixtures/rewrite/reports/eval-rewrite-qwen2-5-1-5b-instruct-q4f16-1-mlc-2026-08-07T03-13-05-794Z.md`) scored `Qwen2.5-1.5B-Instruct-q4f16_1-MLC` — which is `DEFAULT_MODEL_ID` (`src/lib/webllm/models.ts:115`) — at **100% steering adherence across all three prompt variants**:

| Variant | Steering | Aggregate |
| --- | --- | --- |
| Baseline (shipped) | 100% | 69% |
| Terse (rules-only) | 100% | 77% |
| Examples-led (few-shot) | 100% | 63% |

#608 listed three legitimate outcomes for half 2. The third was:

> Adherence varies by model ⇒ this becomes a default-model / `ModelSelector` question and a separate issue.

This is that separate issue. The other two registry models are **unmeasured**, so that outcome is untested rather than ruled out.

## Why they are unmeasured

Not a harness defect and not a model-availability problem. Both were verified during the #608 run:

- All three model-lib `.wasm` URLs return `200`.
- All three Hugging Face repos (`mlc-ai/<model>`) return `200` for both `mlc-chat-config.json` and `params_shard_0.bin`. None is gated.

The failure was local disk. Chrome derives the Cache API quota from **free** disk space, and the run machine had ~3 GB free on a 99%-full volume. The 1.6 GB Qwen weights cached; Gemma 2 (2B, ~1.9 GB) and Llama 3.2 (3B, ~2.3 GB) both threw:

```
Failed to execute 'add' on 'Cache': Cache.add() encountered a network error
```

Size-ordered, exactly as expected from a quota ceiling. A machine with ~10 GB free should complete all three.

## Scope

Run the existing harness for the two remaining registry models and commit the reports. **No code change is expected.** The fixture, the scorer, the rubric criterion and the reporting path all shipped with #608 and are covered by tests.

```sh
npm run eval:rewrite     # https://localhost:5173/eval-rewrite.html
```

One model per fresh tab (cycling multi-GB models in one tab crashes Chrome on the WebGPU eviction path):

[...]

PR #777, opened 2026-08-07
eval(rewrite): measure steering adherence across the model registry, answering #608 half 2 (#608)

### LINKED ISSUE #608: Full-résumé rewrite ignores feedback: the app's own findings never reach the prompt, and steering adherence is unmeasured

Samhit21:
## Summary

The last open piece of #608. Half 1 (wiring the app's findings into the rewrite prompt) shipped in #666; the adherence scorer, fixture and reporting-path proof shipped in #666/#714. What was left was the one criterion nobody could satisfy from a keyboard: **run the eval across the shipped model list and state a conclusion for half 2.**

This lands the first eval reports carrying a real **Steering** column — one per registry model. **All nine cells are 100%:**

| Model | Baseline (shipped) | Terse (rules-only) | Examples-led (few-shot) |
| --- | --- | --- | --- |
| Qwen 2.5 (1.5B) — `DEFAULT_MODEL_ID` | **100%** | **100%** | **100%** |
| Gemma 2 (2B) | **100%** | **100%** | **100%** |
| Llama 3.2 (3B) | **100%** | **100%** | **100%** |

**Half 2 does not reproduce anywhere.** The report that opened it — "the rewrite ignores my typed instructions" — does not survive measurement on any model the app ships, under any prompt variant.

That accounts for **two** of the three causes #608 proposed:

- **not prompt-length dilution** — the terse (rules-only) variant scores identically to the verbose ones;
- **not model variance** — three families across 1.5B/2B/3B agree exactly;
- **per-section repetition is untouched either way** — a `RewriteFixture` is "one résumé-section fixture" (`eval/types.ts`) and the runner passes `fixture.bullets` as a single unit, so every record is one section. This harness cannot exercise the several-sections loop that hypothesis is about. *(Corrected in review — an earlier draft claimed all three were ruled out, which read as evidence but was true only because there is exactly one section per run.)*

#608 listed "adherence is fine ⇒ close half 2 as not-reproducible, with the eval as the evidence and a permanent regression guard" as a legitimate outcome. This is that, and the fixture stays behind as the guard.

Closes #608

## The number is auditable, not asserted

Worth stating plainly, because a uniform 100% invites the question:

- The reports keep every generated bullet at `records[].rubric.perBullet[].text`. All **36** output bullets (4 × 3 variants × 3 models) are free of the forbidden word, with varied lead verbs — Migrated / Led / Rolled out / Coordinated / Implemented / Managed. The README carries a copy-pasteable snippet to re-derive this from the committed JSON; I ran it verbatim against all three files.
- The probe is **non-trivial by construction**: all four *input* bullets begin with "Spearheaded", so echoing the input scores 0%.
- Real inference, not a stub: 110.3s / 184.6s / 189.8s total per model, 3.9–28.7s per rewrite.
- Each run: 15 records = 3 variants × 5 fixtures, **0 errored**, `appVersion 72ad3c1` — all three ran the same code.
- `steeringAdherence` is populated only on the probe fixture and `null` on the other 12 records per run — the scoping the criterion was designed for.
- `scoreAdherence` was mutation-checked against this exact fixture before any number was accepted: echoing the input → `false`, compliant rewrite → `true`, 3-of-4 compliant → `false`, empty output → `false`. A pass cannot be bought with a truncated generation. The 3-of-4 case had no test behind it when the README cited it (review nit) — it now does (`adherence.test.ts`), and I mutation-checked *that*: swapping the scorer to majority-pass fails it.

## What this does NOT establish

Recorded in the reports README rather than left to a reader's charity: the probe contributes **one scored cell per (model, variant)**, so nine cells rest on nine generations of a single four-bullet fixture with a single forbidden-word instruction. Enough to refute "instructions are ignored" — a defect that severe would have shown up immediately — and not enough to characterise adherence across instruction *kinds*. A whole-document instruction ("cut this to one page") remains something the architecture cannot honour section-by-section, which is why #608 Phase 3 fixed the steering box's copy rather than the prompt.

## Also in this PR

Stale dev-server URLs in the harness docs pointed at `http://localhost:5173/offlinecv/…` — wrong on both counts (no `/offlinecv/` prefix under the current base, and the dev server is HTTPS via `basicSsl`). I verified the eval URL 404s against a live server before changing it. It is the first thing anyone follows when taking one of these runs, so it belongs with the run rather than in a separate PR.

Review caught that my first sweep grepped for the `/offlinecv/` prefix and so missed files with only the wrong *scheme*. Now swept on both patterns: `parse-eval/README.md` (flagged in review) plus **`CONTRIBUTING.md`**, which the review did not flag and which is the first file a new contributor reads — `README.md` and `CLAUDE.md` already said `https://`, so it was the lone outlier. Zero `http://localhost:5173` or `/offlinecv/` dev URLs remain anywhere in the tree.

## Review focus

- `tests/fixtures/rewrite/reports/README.md` — the conclusion is the deliverable. Does a uniform 100% across nine cells read as suspicious rather than clean? My argument that it isn't: the scorer was mutation-checked in all four directions, the probe fails on echoed input, and the outputs are in the artifact for anyone to grep.
- **`Numbers 0–80%` and `Dedup 0%` are visibly poor in these same reports** — now filed as **#778**, not left as a remark. They are other criteria outside #608's scope, and the 2026-06-23 baselines are not comparable (they predate the current fixture set), so this PR ships the files that show the problem without pretending to fix it. Reviewers should decide whether #778 outranks whatever is next, because a rewriter dropping `$4.2M` from a bullet is a worse user outcome than the one this PR closes.
- The commit bundles a doc fix with the eval snapshot. Reviewer's call was to keep it together; leaving it as one commit.

## Test plan

- [x] `npm run lint` clean
- [x] `npx vitest run src/lib/webllm` green — 441 passed
- [x] `npm run check:fixtures` green — 58 PDFs + 15 sidecars, all synthetic
- [x] All three report JSONs swept for PII before committing: every string-valued path enumerated; no emails or phone numbers; all free text is model output over the synthetic fixtures
- [x] Steering verdict re-derived from the committed artifacts — 36/36 bullets clean
- [x] `scoreAdherence` mutation-checked against the shipped fixture, all four directions
- [ ] `npm run verify` — **not green locally, and not because of this branch.** 9 tests in `src/lib/letter-egress-ack.test.ts` fail identically on unmodified `origin/main`; they pass on Node 18 and fail on Node 25 (native `localStorage` breaks that file's `afterEach`). `.nvmrc` pins 20 and CI runs it, so CI is the authority — and CI `verify` is **green** on this PR. Filed as **#779**; pushed with `OFFLINECV_SKIP_HOOKS=1` rather than tick a box I could not honestly tick.

## Not in this PR — and where each piece lives instead

Nothing here is left as prose in a squashed commit message. All three real problems this work surfaced are filed — #778, #779 and #782:

- **Any prompt or model change for adherence.** #608 required measurement to precede a fix; the measurement says there is nothing to fix on the adherence path. Nothing outstanding, no follow-up needed.
- **Number preservation is failing — #778.** These reports show `numbersPreservedRate` between 0% and 80%, with **8 of 9 cells at or below 50%**, dropping salary figures (`$180K`, `$85K`, `$250K`), user counts (`120K`, `1.8M`) and percentages (`14%`). That is the rewrite's headline guardrail — `SECTION_REWRITE_SYSTEM_PROMPT` says "Preserve every concrete number from the input EXACTLY" — and it is not holding on any model, including the default. Filed with the per-fixture dropped-number table and four fix directions. **Out of scope here, but more serious than the criterion this PR closes.**
- **`dedupEffectiveRate` is 0 in 8 of 9 cells** — recorded as a secondary section of #778 rather than its own issue, because the rate is computed over `redundant` fixtures only and there is exactly one, so it is one fixture failing repeatedly rather than a broad sample.
- **Rewritten bullets ship literal markdown bold — #782.** Found in review, not by me: **19 of 20 Gemma "terse" bullets** carry `**Led** the migration…` in `perBullet[].text`, which is *post*-`cleanRewriteLine` output — so it is what would reach an exported ATS PDF. Root cause is ordering in `post-process.ts`: the leading-`**Verb**` strip is anchored at start-of-line and runs *before* the list-marker strip, so any of `-`, `*`, `•`, `1.` shields the bold. I reproduced all four marker forms by direct execution. `noPreambleLeak` passes on every affected record, so no criterion catches it. Filed with the caveat that the fix is not simply reordering — the current order deliberately protects the `*Foo.*` italics case.
- **`verify` is red on Node 22+ — #779.** `letter-egress-ack.test.ts` fails 9/9 on Node 25 and passes 9/9 on Node 18, on unmodified `main`. This is **#398 recurring in a file written after it was closed**, which argues for pinning the test environment rather than patching a third file. Why this PR's `verify` box is unticked.

#776 (cross-model adherence) was opened when Gemma and Llama looked unrunnable, and is **closed as completed** — both runs succeeded on a retry and are committed here.

## Provenance

Eval runs: executed by the maintainer in Chrome with WebGPU. The JSONs are the unmodified harness downloads; their Markdown siblings were regenerated from those JSONs via `renderMarkdownReport` (only the JSON was saved during the runs).

### DIFF
--- a/tests/fixtures/rewrite/reports/README.md
+++ b/tests/fixtures/rewrite/reports/README.md
@@ -25,7 +25,7 @@ PR descriptions.
 ## Workflow
 
 ```sh
-npm run eval:rewrite        # opens /offlinecv/eval-rewrite.html with WebGPU
+npm run eval:rewrite        # opens https://localhost:5173/eval-rewrite.html (WebGPU)
 # in the browser: click "Run eval", wait, download both report files
 mv ~/Downloads/eval-rewrite-*.json tests/fixtures/rewrite/reports/
 mv ~/Downloads/eval-rewrite-*.md   tests/fixtures/rewrite/reports/
@@ -33,24 +33,125 @@ git add tests/fixtures/rewrite/reports/
 git commit -m "eval(rewrite): snapshot YYYY-MM-DD run"
 ```
 
-## ⚠️ The committed reports predate the Steering column (#608)
+## The 2026-06-23 reports predate the Steering column (#608)
 
-The three reports here are from **2026-06-23** — the first WebGPU runs
-captured on a maintainer machine, which is why the directory shipped
-empty with the harness rather than carrying a synthetic placeholder.
-Their aggregate table carries these eight criteria columns after `Model`
-and `Variant`:
+The three reports from **2026-06-23** were the first WebGPU runs captured
+on a maintainer machine, which is why the directory shipped empty with the
+harness rather than carrying a synthetic placeholder. Their aggregate table
+carries these eight criteria columns after `Model` and `Variant`:
 `Numbers | One-line | Verb | Length | No-preamble | Dedup | Judge | Aggregate`.
 The **Steering** column and the `steering-forbidden-word` fixture arrived
 later with #608's adherence criterion, so those runs measured nothing
-about steering adherence — the question #608 half 2 exists to answer.
+about steering adherence.
 
 **They are not evidence about half 2, in either direction.** Their absence
 of a Steering column is an artifact of when they were run, not a finding.
-Half 2 stays open until a fresh run of the full registry produces one.
 
-When you take that run, the reading is on the model × variant axes — a
-low rate on one model only is a default-model question, not a prompt one.
+## #608 half 2 — steering adherence, answered on the whole registry
+
+The **2026-08-07** reports are the first carrying a real Steering column,
+and they cover every model in `MODEL_REGISTRY` × every shipped prompt
+variant. **All nine cells are 100%:**
+
+| Model | Baseline (shipped) | Terse (rules-only) | Examples-led (few-shot) |
+| --- | --- | --- | --- |
+| Qwen 2.5 (1.5B) — `DEFAULT_MODEL_ID` | **100%** | **100%** | **100%** |
+| Gemma 2 (2B) | **100%** | **100%** | **100%** |
+| Llama 3.2 (3B) | **100%** | **100%** | **100%** |
+
+**Half 2 does not reproduce anywhere.** The user report that opened it —
+"the rewrite ignores my typed instructions" — does not survive measurement
+on any model the app ships, under any prompt variant. That settles two of
+the three causes #608 proposed: it is not prompt-length dilution (the terse
+variant scores the same as the verbose ones) and not model variance (three
+models of different families and sizes agree exactly).
+
+**Per-section repetition is untouched either way.** A `RewriteFixture` is
+one résumé section (`src/lib/webllm/eval/types.ts`) and the runner passes
+`fixture.bullets` as a single unit, so every eval record is a single
+section. #608's hypothesis is about the steering suffix being re-appended
+per section across the multi-section loop in
+`src/lib/webllm/rewrite-resume.ts` — which accumulates `usedVerbs` /
+`usedPhrases` context between sections and which this harness never enters.
+A single-section probe cannot observe that cause, in either direction.
+
+That is a conclusion, not a dismissal: #608 listed "adherence is fine ⇒
+close half 2 as not-reproducible, with the eval as the evidence and a
+permanent regression guard" as a legitimate outcome, and the fixture now
+stands as that guard.
+
+### What this does NOT establish
+
+One limit, worth stating before the number gets cited as more than it is:
+the probe contributes **one scored cell per (model, variant)**, so nine
+cells rest on nine generations of a single four-bullet fixture with a
+single forbidden-word instruction. That is enough to refute "instructions
+are ignored" — a defect that severe would have shown up immediately — and
+not enough to characterise adherence across instruction *kinds*. A
+whole-document instruction ("cut this to one page") is a different claim
+that the architecture still cannot honour section-by-section; that is why
+#608 Phase 3 fixed the steering box's copy instead of the prompt.
+
+Adherence is also the *only* criterion that is uniformly clean. The same
+runs show `Numbers` between 0% and 80% (#778) and `Dedup` at 0% in eight of
+nine cells — unrelated to #608, but visible in these files and worth its own
+look rather than being read as noise.
+
+A third anomaly is in these files and no criterion catches it: **19 of the
+24 Gemma 2 `terse` bullets ship literal markdown bold** — `"**Led** the
+migration of the billing platform…"`. Those are `perBullet[].text`, i.e.
+*post*-`cleanRewriteLine` output, which is what would land in a downloaded
+ATS PDF as literal asterisks; `noPreambleLeak` scores PASS on every one. The
+cause is ordering inside `cleanRewriteLine`
+(`src/lib/webllm/post-process.ts`): the leading-`**Verb**` strip is anchored
+at `^` and runs before the list-marker strip, so a marker shields the bold —
+
+```
+cleanRewriteLine("**Led** the migration…")   → "Led the migration…"
+cleanRewriteLine("- **Led** the migration…") → "**Led** the migration…"
+```
+
+— and the same holds for `*`, `•` and `1.`. Tracked as **#781**, not fixed
+here: the current strip order exists to protect the `*Foo.*` italics case,
+so the reorder needs its own regression test rather than riding along on a
+reports commit.
+
+### How to audit the Steering number yourself
+
+Unlike the aggregate rates, this one is checkable from the committed file.
+The JSON keeps each generated bullet at `records[].rubric.perBullet[].text`,
+so the verdict can be re-derived rather than trusted:
+
+```sh
+python3 -c "
+import json,glob
+for f in sorted(glob.glob('tests/fixtures/rewrite/reports/*2026-08-07*.json')):
+    d=json.load(open(f))
+    for x in d['records']:
+        if x['fixtureId']!='steering-forbidden-word': continue
+        for b in x['rubric']['perBullet']:
+            print(d['modelIds'][0], x['variantId'],
+                  'spearheaded' in b['text'].lower(), b['text'][:60])
+"
+```
+
+All **36** bullets (4 bullets × 3 variants × 3 models) come back `False`,
+with a varied lead verb each time — Migrated / Led / Rolled out /
+Coordinated / Implemented / Managed. The probe is non-trivial by
+construction: all four *input* bullets begin with the forbidden word, so
+echoing the input scores 0%.
+
+`scoreAdherence` itself is unit-tested
+(`src/lib/webllm/eval/adherence.test.ts`) and was mutation-checked against
+this exact fixture before the number was accepted: echoing the input scores
+`false`, a compliant rewrite `true`, 3-of-4 compliant `false`, and empty
+output `false` — so a pass cannot be bought with a truncated or empty
+generation.
+
+### Reading a future run
+
+The reading is on the model × variant axes — a low rate on one model only
+is a default-model question, not a prompt one.
 `src/lib/webllm/eval/README.md` has the full table.
 
 The reporting path itself is covered by
--- a/src/lib/webllm/eval/README.md
+++ b/src/lib/webllm/eval/README.md
@@ -38,8 +38,13 @@ Real models run only in a browser. The entry point is the dev-only
 
 ```sh
 npm run eval:rewrite
-# opens http://localhost:5173/offlinecv/eval-rewrite.html
+# opens https://localhost:5173/eval-rewrite.html
 ```
+
+HTTPS, and no `/offlinecv/` prefix. The dev server is TLS by default
+(`basicSsl`, self-signed — accept the warning once; WebGPU needs a
+secure context), and `BASE_PATH` is `/` unless `VITE_BASE_PATH` says
+otherwise. Both halves of the old URL were stale and 404ed.
 
 **One model per tab.** The page asks you to pick a model from the
 dropdown, then click **Run eval** — it loads that model only, runs every
--- a/src/lib/webllm/eval/adherence.test.ts
+++ b/src/lib/webllm/eval/adherence.test.ts
@@ -44,6 +44,23 @@ describe("scoreAdherence — forbidden-word", () => {
     expect(scoreAdherence(FORBIDDEN, ["SPEARHEADED the rollout"])).toBe(false);
   });
 
+  it("FAILS at 3-of-4 compliant — adherence is all-or-nothing, not an average", () => {
+    // The committed `steering-forbidden-word` fixture is four bullets, so this
+    // is the shape the 100% Steering column in
+    // tests/fixtures/rewrite/reports/README.md actually rests on: a model that
+    // obeyed on three bullets and slipped on the fourth scores `false`, not
+    // 75%. Without this case the README's mutation-check claim has no
+    // assertion behind it at the fixture's own bullet count.
+    expect(
+      scoreAdherence(FORBIDDEN, [
+        "Led the billing migration across 12 markets",
+        "Cut scope churn 30% with a quarterly planning process",
+        "Rolled out the design system to four product teams",
+        "Spearheaded the vendor consolidation review",
+      ]),
+    ).toBe(false);
+  });
+
   it("matches on a word boundary, not a substring", () => {
     // "spearheadedness" is not the word the instruction forbade. A substring
     // check would fail a compliant output and understate adherence.
--- a/src/lib/webllm/eval/run-eval-browser.ts
+++ b/src/lib/webllm/eval/run-eval-browser.ts
@@ -5,7 +5,7 @@
  * Browser entry for the rewrite-quality eval harness.
  *
  * Reached via `npm run eval:rewrite` → opens
- * `/offlinecv/eval-rewrite.html` in the dev server. Loads the model
+ * `/eval-rewrite.html` in the dev server. Loads the model
  * the user picked, runs every prompt variant against every fixture,
  * scores with the deterministic rubric, and renders a downloadable
  * JSON + Markdown report.
--- a/src/lib/webllm/spike/README.md
+++ b/src/lib/webllm/spike/README.md
@@ -16,7 +16,7 @@ design commitment:
 ## How to run
 
 1. Start the dev server: `npm run dev`
-2. Open: `http://localhost:5173/offlinecv/jd-spike.html`
+2. Open: `https://localhost:5173/jd-spike.html`
 3. Select **Qwen 2.5 (1.5B)** (default) in the model picker.
 4. Set **Repeats** (default 3 — higher values give better failure-rate estimates).
 5. Click **Run spike** — the model downloads on first run (~1.6 GB); subsequent runs use the cached IndexedDB copy.
--- a/src/lib/webllm/spike/jd-spike-browser.ts
+++ b/src/lib/webllm/spike/jd-spike-browser.ts
@@ -5,7 +5,7 @@
  * Browser entry for the JD-extraction + evidence-judging spike (issue #198).
  *
  * Reached via `npm run eval:jd-spike` or by opening
- * `http://localhost:5173/offlinecv/jd-spike.html` while the dev server
+ * `https://localhost:5173/jd-spike.html` while the dev server
  * is running. Loads the selected model via WebLLM, runs the spike
  * measurement over the 3 inline fixtures (see fixtures.ts), and offers
  * downloadable JSON + Markdown reports.
--- a/tests/fixtures/rewrite/reports/eval-rewrite-gemma-2-2b-it-q4f16-1-mlc-2026-08-07T04-52-07-435Z.json
+++ b/tests/fixtures/rewrite/reports/eval-rewrite-gemma-2-2b-it-q4f16-1-mlc-2026-08-07T04-52-07-435Z.json
@@ -1,0 +1,1020 @@
+{
+  "startedAt": "2026-08-07T04:49:02.749Z",
+  "appVersion": "72ad3c1",
+  "modelIds": [
+    "gemma-2-2b-it-q4f16_1-MLC"
+  ],
+  "variantIds": [
+    "baseline",
+    "terse",
+    "examples-led"
+  ],
+  "fixtureIds": [
+    "weak-marketing-generalist",
+    "strong-backend-engineer",
+    "numeric-growth-pm",
+    "redundant-support-lead",
+    "steering-forbidden-word"
+  ],
+  "judgeEnabled": false,
+  "records": [
+    {
+      "modelId": "gemma-2-2b-it-q4f16_1-MLC",
+      "variantId": "baseline",
+      "fixtureId": "weak-marketing-generalist",
+      "fixtureKind": "weak",
+      "inputBulletCount": 5,
+      "outputBulletCount": 5,
+      "rubric": {
+        "numbersPreserved": false,
+        "oneLinePerBullet": true,
+        "actionVerbLead": false,
+        "lengthSanity": true,
+        "noPreambleLeak": true,
+        "dedupEffective": null,
+        "steeringAdherence": null,
+        "judgeCoherence": null,
+        "perBullet": [
+          {
+            "index": 0,
+            "text": "Developed and executed marketing campaigns that increased social media engagement by 20% and website traffic by 15%.",
+            "startsWithActionVerb": true,
+            "lengthOk": true,
+            "oneLine": true
+          },
+          {
+            "index": 1,
+            "text": "Contributed to content creation, email blasts, and other promotional materials that boosted brand awareness by 10%.",
+            "startsWithActionVerb": false,
+            "lengthOk": true,
+            "oneLine": true
+          },
+          {
+            "index": 2,
+            "text": "Supported the team with daily tasks and provided administrative support to ensure seamless project execution.",
+            "startsWithActionVerb": false,
+            "lengthOk": true,
+            "oneLine": true
+          },
+          {
+            "index": 3,
+            "text": "Collaborated with senior team members to plan and coordinate marketing events, resulting in a 15% increase in attendance and a 10% reduction in event-related costs.",
+            "startsWithActionVerb": false,
+            "lengthOk": true,
+            "oneLine": true
+          },
+          {
+            "index": 4,
+            "text": "Provided strategic input and creative direction to weekly meetings, contributing to a 5% improvement in campaign effectiveness.",
+            "startsWithActionVerb": false,
+            "lengthOk": true,
+            "oneLine": true
+          }
+        ],
+        "droppedNumbers": [],
+        "addedNumbers": [
+          "20%",
+          "15%",
+          "10%",
+          "15%",
+          "10%",
+          "5%"
+        ]
+      },
+      "rewriteDurationMs": 28654,
+      "error": null
+    },
+    {
+      "modelId": "gemma-2-2b-it-q4f16_1-MLC",
+      "variantId": "baseline",
+      "fixtureId": "strong-backend-engineer",
+      "fixtureKind": "strong",
+      "inputBulletCount": 5,
+      "outputBulletCount": 5,
+      "rubric": {
+        "numbersPreserved": false,
+        "oneLinePerBullet": true,
+        "actionVerbLead": true,
+        "lengthSanity": true,
+        "noPreambleLeak": true,
+        "dedupEffective": null,
+        "steeringAdherence": null,
+        "judgeCoherence": null,
+        "perBullet": [
+          {
+            "index": 0,
+            "text": "Led a Kafka-based migration of the order-processing pipeline, reducing median latency by 38% across 12 downstream services.",
+            "startsWithActionVerb": true,
+            "lengthOk": true,
+            "oneLine": true
+          },
+          {
+            "index": 1,
+            "text": "Designed a horizontally-sharded Postgres layout that absorbed a 4x traffic surge during the 2024 product launch without requiring read-replica failover.",
+            "startsWithActionVerb": true,
+            "lengthOk": true,
+            "oneLine": true
+          },
+          {
+            "index": 2,
+            "text": "Mentored five engineers through code review and pairing sessions, resulting in the delivery of 23 production features across two quarters with zero rollback incidents.",
+            "startsWithActionVerb": true,
+            "lengthOk": true,
+            "oneLine": true
+          },
+          {
+            "index": 3,
+            "text": "Reduced cloud spend by $180,000 annually through rightsizing 60 underused instances and consolidating three staging environments into a single tenant-isolated environment.",
+            "startsWithActionVerb": true,
+            "lengthOk": true,
+            "oneLine": true
+          },
+          {
+            "index": 4,
+            "text": "Built a deterministic load generator that replays 24 hours of production traffic in 90 minutes, enabling four teams to conduct regression checks.",
+            "startsWithActionVerb": true,
+            "lengthOk": true,
+            "oneLine": true
+          }
+        ],
+        "droppedNumbers": [
+          "5",
+          "$180K"
+        ],
+        "addedNumbers": [
+          "$180,000"
+        ]
+      },
+      "rewriteDurationMs": 22405,
+      "error": null
+    },
+    {
+      "modelId": "gemma-2-2b-it-q4f16_1-MLC",
+      "variantId": "baseline",
+      "fixtureId": "numeric-growth-pm",
+      "fixtureKind": "numeric",
+      "inputBulletCount": 5,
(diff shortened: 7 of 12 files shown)


tests/fixtures/rewrite/reports/README.md
```diff
@@ -25,32 +25,106 @@ PR descriptions.
...
+| Model | Baseline (shipped) | Terse (rules-only) | Examples-led (few-shot) |
+| --- | --- | --- | --- |
+| Qwen 2.5 (1.5B) — `DEFAULT_MODEL_ID` | **100%** | **100%** | **100%** |
+| Gemma 2 (2B) | **100%** | **100%** | **100%** |
+| Llama 3.2 (3B) | **100%** | **100%** | **100%** |
+
+**Half 2 does not reproduce anywhere.** The user report that opened it —
+"the rewrite ignores my typed instructions" — does not survive measurement
+on any model the app ships, under any prompt variant. That rules out all
+three causes #608 proposed: it is not prompt-length dilution (the terse
```
s-annam: **Secondary 1 — this is the one cause the probe cannot rule out.**

`RewriteFixture` is documented as *"One résumé-section fixture"* (`eval/types.ts:27`) and `runner.ts:97` passes `fixture.bullets` as a single unit, so every eval record is **one section**. #608's per-section-repetition hypothesis is about the same suffix being appended to *each of several* sections in the `rewrite-resume.ts:227` loop — a single-section probe has no way to observe it. The PR body's gloss ("honoured in every section of every run") reads as evidence but is true only because there is exactly one section per run.

The other two causes are genuinely supported. This one also contradicts your own caveat two paragraphs down, which correctly concedes the whole-document case.

Suggested wording:

```
three causes #608 proposed: it is not prompt-length dilution (the terse
variant scores the same as the verbose ones) and not model variance (three
models of different families and sizes agree exactly). Per-section
repetition is untouched either way — every eval record is a single section
(`eval/types.ts:27`), so this harness cannot exercise it.
```

Full reasoning in the review body.

tests/fixtures/rewrite/reports/README.md
```diff
@@ -25,32 +25,106 @@ PR descriptions.
...
+the probe contributes **one scored cell per (model, variant)**, so nine
+cells rest on nine generations of a single four-bullet fixture with a
+single forbidden-word instruction. That is enough to refute "instructions
+are ignored" — a defect that severe would have shown up immediately — and
+not enough to characterise adherence across instruction *kinds*. A
+whole-document instruction ("cut this to one page") is a different claim
+that the architecture still cannot honour section-by-section; that is why
+#608 Phase 3 fixed the steering box's copy instead of the prompt.
+
+Adherence is also the *only* criterion that is uniformly clean. The same
```
s-annam, line 95: **Secondary 2 — there is a third anomaly in these files, and it is a real product bug.**

This paragraph names `Numbers` and `Dedup`. It misses one that is more directly user-visible than either: **19 of 20 Gemma "terse" bullets ship literal markdown bold.**

```
"**Led** the migration of the billing platform to a new payments provider…"
"**Developed** and **implemented** an internal design system…"
```

Those are `perBullet[].text` — *post*-`cleanRewriteLine` output, i.e. what would land in a downloaded ATS PDF as literal asterisks. `noPreambleLeak` scores **PASS** on every one, so no existing criterion catches it.

Root cause, confirmed by direct execution: `post-process.ts:93` strips a leading `**Verb**`, but `post-process.ts:110` strips the list marker *afterwards*, so a marker shields the bold from the anchored pattern:

```
cleanRewriteLine("**Led** the migration…")   → "Led the migration…"      ✅
cleanRewriteLine("- **Led** the migration…") → "**Led** the migration…"  ❌
```

Same for `*`, `•`, `1.`. Out of scope to fix here — but it belongs in #778 or its own issue rather than staying invisible in the file that argues these numbers are auditable.

tests/fixtures/rewrite/reports/README.md
```diff
@@ -25,32 +25,106 @@ PR descriptions.
...
+```
+
+All **36** bullets (4 bullets × 3 variants × 3 models) come back `False`,
+with a varied lead verb each time — Migrated / Led / Rolled out /
+Coordinated / Implemented / Managed. The probe is non-trivial by
+construction: all four *input* bullets begin with the forbidden word, so
+echoing the input scores 0%.
+
+`scoreAdherence` itself is unit-tested
+(`src/lib/webllm/eval/adherence.test.ts`) and was mutation-checked against
```
s-annam, line 145: **Nit — "3-of-4 compliant → `false`" is not asserted anywhere.**

The other three directions are real: echo→`false` and compliant→`true` run against the actual fixture (`adherence-reporting.test.ts:138/126`), and empty→`false` uses the fixture's exact check object (`adherence.test.ts:141`). The 3-of-4 case has no test — nearest is 1-of-2 (`adherence.test.ts:33`).

The property holds trivially, so this is wording rather than a hole. But this README is the durable artifact and a reader who greps for the case comes up empty. Either say "partial compliance (1-of-2)" or add the three-line 4-bullet case.

s-annam (dismissed):
Approving. The conclusion is the deliverable here, so I reviewed the artifact
rather than the prose: I re-ran the README's audit snippet verbatim and
re-derived every headline number from the committed JSON. They all hold, exactly
— including the ones that would have been easiest to fudge.

Two Secondary findings and two Nits. Neither Secondary touches the 100% number;
one is about a claim the probe cannot support, and one is a product bug your own
artifact exposes and the README doesn't name.

## What I re-derived independently

| Claim | Verdict |
|---|---|
| All nine `steeringAdherenceRate` cells = 1.0 | ✅ confirmed in `aggregates` of all 3 JSONs |
| 36/36 bullets free of the forbidden word | ✅ ran the README snippet verbatim, 36 `False` |
| **The rate is not diluted by unprobed records** | ✅ `runner.ts:188-195` divides by non-null cells only — 3 probed, 12 `null`, per run |
| All 4 *input* bullets lead with "Spearheaded" | ✅ `steering-adherence.json` |
| 110.3s / 184.6s / 189.8s; 3.9–28.7s per rewrite | ✅ exact, to the decimal, from `rewriteDurationMs` |
| 15 records, 0 errored, `appVersion 72ad3c1` ×3 | ✅ and `72ad3c1` is this branch's base commit |
| Every `MODEL_REGISTRY` model covered | ✅ 3 registered (`models.ts:75`), 3 reports; `DEFAULT_MODEL_ID` = Qwen 2.5 |
| URL fix is correct | ✅ `BASE_PATH="/"` (`vite.config.ts:86`), `basicSsl` default (`:246`), `eval:rewrite` opens `/eval-rewrite.html`; zero `/offlinecv/` refs remain |
| PII sweep of the 3 JSONs | ✅ independently swept — no emails, phones, URLs or machine paths; every proper noun is a verb or a tech term |
| `Closes #608` — half-1 ACs really shipped | ✅ named tests exist for cross-section scoping, byte-identical-when-absent, guardrail-precedes-findings, and no-critique-unchanged (`steering.test.ts:89/125/133/148`, `rewrite-resume.test.ts:587`) |
| #776 closed, #778/#779 open as described | ✅ |

Gates: `typecheck` ✅ `lint` ✅ `vitest src/lib/webllm` 441 ✅ `src/hooks` 204 ✅
`check:fixtures` ✅ `fallow` — *"No issues in 11 changed files"* (the duplication it
reports is inherited across the three browser harnesses) ✅ CI `verify` green ✅.

The dilution check is the one I want to call out as genuinely reassuring: a
uniform 100% computed over 15 records where 12 are vacuous would have been
worthless, and it isn't — the criterion divides by the 3 probed cells only.

## Secondary

### 1. "not per-section repetition" is the one cause the probe cannot rule out

`tests/fixtures/rewrite/reports/README.md:65` (and the matching PR-body bullet)
says the run rules out all three of #608's proposed causes. Two of the three are
supported. The third is not testable by this harness at all:

`RewriteFixture` is documented as *"One résumé-section fixture"* (`eval/types.ts:27`)
and `runner.ts:97` passes `fixture.bullets` as a single unit — so every record is
**one section**. #608's per-section-repetition hypothesis is specifically about the
same suffix being appended to *each of several* sections in the `rewrite-resume.ts:227`
loop. A single-section probe has no way to observe it. The PR body's gloss —
"the instruction is honoured in every section of every run" — is true only because
there is exactly one section per run.

This also contradicts your own caveat two paragraphs down, which correctly concedes
the whole-document case. The caveat is right; the "rules out" bullet overshoots it.

**Fix** — swap the claim for what the data supports:

```
three causes #608 proposed: it is not prompt-length dilution (the terse
variant scores the same as the verbose ones) and not model variance (three
models of different families and sizes agree exactly). Per-section
repetition is untouched either way — every eval record is a single section
(`eval/types.ts:27`), so this harness cannot exercise it.
```

### 2. The reports expose a third anomaly, and it's a real product bug

`README.md:87` takes on the job of naming what else these files show, and names
`Numbers` and `Dedup`. There is a third, and it is more directly user-visible than
either: **19 of 20 Gemma "terse" bullets ship literal markdown bold.**

```
"**Led** the migration of the billing platform to a new payments provider…"
"**Developed** and **implemented** an internal design system…"
```

These are `perBullet[].text`, which is *post*-`cleanRewriteLine` output — so this is
what would land in a downloaded ATS PDF, rendered as literal asterisks. And
`noPreambleLeak` scores **PASS** on every one of them, so no existing criterion
catches it.

Root cause, confirmed by direct execution: `post-process.ts:93` strips a leading
`**Verb**`, but `post-process.ts:110` strips the list marker *afterwards*. A model
emitting `- **Led** the migration` shields the bold from the anchored pattern:

```
cleanRewriteLine("**Led** the migration…")   → "Led the migration…"      ✅
cleanRewriteLine("- **Led** the migration…") → "**Led** the migration…"  ❌
```

Same for `*`, `•`, and `1.`. Out of scope to fix here — but it belongs in #778
alongside the other two, or in its own issue, rather than staying invisible in a
file that argues these numbers are auditable.

## Nits (non-blocking)

1. **`README.md:118` — "3-of-4 compliant → `false`" isn't asserted anywhere.** The
   other three mutation directions are real: echo→`false` and compliant→`true` run
   against the actual fixture (`adherence-reporting.test.ts:138/126`), and empty→`false`
   uses the fixture's exact check object (`adherence.test.ts:141`). The 3-of-4 case has
   no test — the nearest is 1-of-2 (`adherence.test.ts:33`). The property holds
   trivially, so this is wording, not a hole; but the README is the durable artifact
   and a reader who greps for it comes up empty. Either say "partial compliance
   (1-of-2)" or add the three-line 4-bullet case.

2. **The stale-URL sweep missed the third sibling harness.** `parse-eval/README.md:39`
   still reads `http://localhost:5173/parse-eval.html`. The scheme half of your own
   argument applies verbatim — `basicSsl` is on by default, so this URL fails the same
   way. `run-eval-browser.ts`, `jd-spike-browser.ts` and `parse-eval-browser.ts` are the
   same family (fallow groups them as clones); two of three are now correct. One-line
   follow-through, here or in a sweep.

## On the "Also in this PR" and "happy to split" offer

Keep it together. Five doc lines that 404 for the next person taking one of these
runs belong with the run, and the diff is trivially separable at review time.

---
Reviewed by: Claude Opus 5 (1M context). Reasoning effort is not surfaced to the
session, so it is omitted rather than guessed.

tests/fixtures/rewrite/reports/README.md
s-annam: Fixed in `5aa4799`. Confirmed — the probe genuinely cannot observe this cause: `RewriteFixture` is one section, `runner.ts` passes `fixture.bullets` as a single unit, and `rewrite-resume.ts` accumulates `usedVerbs`/`usedPhrases` across a multi-section loop the harness never enters.

Three surfaces carried the over-claim, all corrected:

- **Reports README** — "rules out all three causes" → settles two, with per-section repetition split into its own paragraph naming the single-section limit and the loop it cannot reach.
- **Commit message body** — carried the same sentence. Corrected in the amend, since that text is what the merge queue lands in `main` permanently.
- **PR body** — the "honoured in every section of every run" bullet is gone, replaced with an explicit statement that the harness cannot settle this cause, and why the old gloss read as evidence.

Took the suggested wording as the basis. Resolving.

tests/fixtures/rewrite/reports/README.md
s-annam: Confirmed and filed as **#781**. Documented in `5aa4799`; the code fix is deliberately not in this PR.

Reproduced by direct execution against the real helper:

```
cleanRewriteLine("**Led** the migration…")   → "Led the migration…"       ✅
cleanRewriteLine("- **Led** the migration…") → "**Led** the migration…"   ❌
cleanRewriteLine("* **Led** the migration.") → "**Led** the migration."   ❌
cleanRewriteLine("1. **Led** the migration.")→ "**Led** the migration."   ❌
cleanRewriteLine("• **Led** the migration.") → "**Led** the migration."   ❌
```

Two corrections to the finding as written:

- The count is **19 of 24**, not 19 of 20. Every Gemma `terse` record is fully bold (5/5, 5/5, 5/5, 4/4 across four fixtures); 24 is the variant total.
- It is **Gemma-`terse`-only** — all eight other model×variant cells are 0 bold. Still a shipped model under a shipped prompt variant, so it counts, but it is not a general post-process failure.

Also worth adding: this is on the **product path**, not eval-only — `rewrite-section.ts:220` calls the same `cleanRewriteLine`, so it reaches an accepted rewrite and then the downloaded PDF.

Not fixed here because the reorder is not the one-liner it looks like: the current order is documented as protecting the `*Foo.*` italics case, and the mid-line shape (`**Developed** and **implemented** …`) needs a separate decision since the pattern is single-word and leading-only. #781 carries the repro, the per-cell table, three candidate fix shapes and acceptance criteria including a fail-before test.

README now names it in the anomalies paragraph next to `Numbers` (#778) and `Dedup`. Resolving — the ask was that it not stay invisible in the file arguing these numbers are auditable.

tests/fixtures/rewrite/reports/README.md
s-annam: Fixed in `5aa4799` by adding the test rather than weakening the wording — took the second option you offered.

`adherence.test.ts` gains `FAILS at 3-of-4 compliant — adherence is all-or-nothing, not an average`: four bullets, three clean, one `Spearheaded`, asserting `false`. That is the shipped `steering-forbidden-word` fixture's own bullet count, so the README claim now round-trips to an assertion at the exact shape the 100% Steering column rests on.

You are right that the property held trivially and nothing changed behaviourally — the test passes before and after. Recording that plainly in the commit body so it is not read as a bug fix. The value is that a reader who greps for the 3-of-4 case now finds it, which was the actual complaint.

29 tests green in `adherence.test.ts` + `adherence-reporting.test.ts`. Resolving.

s-annam (approved):
**Approving.** All three findings from the last round are addressed, and two of them landed better than the finding asked for.

### What resolved each thread

**Secondary 1 — the unsupported cause.** Corrected, and in three places rather than one. The README now splits per-section repetition into its own paragraph naming the exact limit (a `RewriteFixture` is one section, the runner passes `fixture.bullets` as a unit, so the multi-section loop in `rewrite-resume.ts` is never entered). The same sentence was also in the **commit message body** and the **PR body** — both corrected. The commit one mattered most: the merge queue derives the squash message from it, so "rules out all three causes" would have been in `main`'s history permanently. Catching the two surfaces the thread didn't name is the right instinct.

**Secondary 2 — the third anomaly.** Now named in the README next to `Numbers` (#778) and `Dedup`, and filed as **#781** with a repro, a per-cell table and acceptance criteria including a fail-before test.

Two corrections to my own finding, both right and both verified in the diff:

- The count is **19 of 24**, not 19 of 20 — I had the numerator right and the denominator wrong.
- It is **Gemma-`terse`-only**; the other eight cells are 0 bold. My comment implied a broader post-process failure than the data shows.

And it went one step further than I did: `rewrite-section.ts:220` calls the same `cleanRewriteLine`, so this is on the **product path**, not eval-only. That is what makes #781 a real user-facing bug rather than a reporting curiosity — literal asterisks reach the downloaded ATS PDF.

Deferring the fix is the correct call. The strip order is documented as protecting the `*Foo.*` italics case, and the mid-line shape (`**Developed** and **implemented** …`) needs a separate decision since the pattern is single-word and leading-only. That is a change with its own test surface; it does not belong on a reports commit.

**Nit — the unasserted 3-of-4 case.** Took the stronger of the two options I offered: added the test instead of weakening the wording. `FAILS at 3-of-4 compliant` uses four bullets, matching the shipped `steering-forbidden-word` fixture's own count, so the README claim now round-trips to an assertion at the exact shape the 100% Steering column rests on.

The commit body and the PR body both state plainly that the property already held and the test passes before and after — coverage, not a behaviour change. That is the honest framing and it is the one used.

### Verification

- One commit (`5aa4799`), as the merge queue needs.
- CI green: `verify` pass (3m14s), `fallow` pass, Cloudflare Pages pass.
- Fixture PII is not in play — no fixture binaries touched this round; the two changed files are a Markdown doc and a unit test.
- The conclusion the PR ships is now scoped to what the harness measured. Two causes settled, one explicitly out of reach, with the reason stated rather than left to a reader to notice.

Ship it. #781 and #778 both carry what these reports surfaced, so nothing here is left as prose in a squashed message.

Issue #203, opened 2026-06-26
JD-match: useJdMatch hook + App rewiring (async state machine)


s-annam:
Part of the WebLLM-native JD-matching redesign (anchor: #156, milestone **M6 · JD Matching v2**).

## Goal
Move JD-match interaction state out of `App.tsx` into a hook, supporting the async semantic path.

## Scope
- New `src/hooks/useJdMatch.ts` — async state machine mirroring `useResumeRewrite` (`idle → loading → running → ready → error`) plus a synchronous `keyword` fast-resolve branch when WebGPU is absent or the user hasn't opted in.
- Replace the synchronous `useMemo` JD pipeline in `App.tsx:82-90` with the hook.
- Reuse `useModelSelection`, `detectWebGpu`, `loadEngine` progress callback.

## Constraints
- `App.tsx` stays thin (CLAUDE.md: cross-cutting interaction state belongs in `src/hooks/`).
- Keyword behavior unchanged when path resolves to `keyword`.

## Done when
`App.tsx` consumes `useJdMatch`; keyword users see no change; semantic users get the loading/running/ready states.

PR #783, opened 2026-08-08
refactor: fold JD paste + tailor into /jobs/, retire /jd-fit/ (#576)

### LINKED ISSUE #576: Deprecate /jd-fit/ — migrate JD paste + JD-driven rewrite into the Find Jobs tab, then remove the second entry

Vaishnavi1709:
Closes #576

## Summary

- `/jd-fit/` is deleted (entry HTML, `JdFitApp`, `useJdFitResume`, `jd-fit-handoff`, `SaveJobFromMatch`, `jdFitTrailingSlash` vite plugin, `jd-fit-banner` flag, `VITE_ENABLE_JD_FIT` env, `AnalyticsSurface`'s `jd-fit` member, App's cross-sell block). `src/lib/jd-match/` stays — Find Jobs hard-depends on it (`rank.ts`, `query-builder.ts`, every provider, `contact.ts`, `analytics.ts`, `webllm/spike`).
- **Migrated capabilities land on `/jobs/`** before deletion, so this is net-zero on capability:
  - **Arbitrary JD paste/URL fetch**: new `PasteJdPanel` in a disclosure below the results, reusing `JdInput` unchanged + the same three-line `extractJdTerms` → `computeCoverage` → `JdMatch` composition (200ms debounce added on the compute).
  - **JD-driven rewrite**: a "Tailor résumé to this job" button on each `JobResultCard` (and on `PasteJdPanel`) stashes the rewrite steering in sessionStorage (`lib/tailor-handoff.ts`, one-shot) and navigates back to `/`. `ResultDetailTabs` consumes the handoff on the initial mount AND on every `pageshow` — the latter is the load-bearing signal, because the primary return path is `history.back()`, a bfcache RESTORE with no remount and no `useEffect` re-fire.
- The tailor button hands over the steering built from the SAME `jdMatch.coverage` object the card's headline fit reads, so the displayed match and the rewrite steering cannot disagree.
- The tailor button is HIDDEN when `buildJdRewriteContext` yields nothing to steer with — the builder's own result is the render gate, so the button always leads somewhere.
- The handoff is addressed, not merely one-shot: the payload carries a `parseFingerprint` of the résumé its coverage was computed against, and `/` discards one stamped for a parse it no longer has.
- `jdContext` resets when the PARSE identity changes (a résumé loaded from the library, an escape-hatch re-parse) and survives ordinary edits, so a tailoring instruction from one parse never leaks into another and typing does not silently discard the steering the user just came back to apply.

## Architectural note (why the plan diverged from the issue text)

The issue was written when `FindJobsPanel` was still a tab on `/`, but the search page has since split into its own HTML entry (`/jobs/`, `JobsApp`). This meant the tailor button couldn't just `setTab("reconstructed")` locally — it needed a `/jobs/` → `/` handoff via sessionStorage. Everything else follows the issue's Reuse analysis.

## Response to review — round 2

Round 2 was authored by the reviewer rather than sent back for another pass, because findings #1 and #2 share one root cause (there was no stable parse identity below `App`) and splitting them across rounds would have produced two mechanisms where one does. Round-1 responses are preserved below for history.

- **#1 — `jdContext` reset fired on every edit (blocking).** `useAnalyzedResume` now exports `parseKey`, the pristine-parse identity it already computed for its own clear-edits effect; `Result.tsx` combines it with `llmOverride` into a `parseIdentity` token, and `ResultDetailTabs` keys the reset on that. `activeResult` could never serve: it is a memo over the edit override maps, so every keystroke mints a fresh object for the same parse — keyed on it, the reset fired on the first character typed and dropped the steering the user had just returned from `/jobs/` to apply. New test renders a shallow copy of `activeResult` under an unchanged `parseIdentity` and asserts `jdContext` survives; it fails against the previous shape with `expected 'jdContext=NULL' to be 'jdContext=surface Kubernetes'`.
- **#2 — one-shot did not mean "consumed by the right résumé".** One-shot bounds how many times a payload is read, not which parse reads it. When `/` is not restored from bfcache (tab reloaded, résumé reset), the next `ResultDetailTabs` to mount is a different parse and consumes the key exactly once — against the wrong résumé. `writeTailorHandoff` now stamps a `fingerprintParse` of the résumé the coverage came from, and `consumeTailorHandoff(expected)` drops (and clears) a mismatch. Content-derived rather than a minted counter, so it survives the sessionStorage round trip and a cold page load without colliding.
- **#3 — render gate and payload were two different predicates.** `missing.length > 0` is strictly wider than `buildJdRewriteContext(...) !== null`: a missing term whose `display` trims to empty clears the first and still yields null from the second, so the button rendered and the click no-op'd. Both call sites now call the builder once and use its result as both the gate and the payload; `onTailor` takes the built instruction, so the two cannot drift apart again.
- **#4 — `KeywordJdMatch` redeclared.** `PasteJdPanel` imports it from `rank.ts`, which already exports it (type-only, so no runtime cost).
- **Nit — review-round citations in code.** `PR #783 review, nit #N` comments are unresolvable after the merge-queue squash and the numbering is not stable across rounds. Reasoning kept, citations re-pointed at `#576`.
- **Nit — `Refs #576` → `Closes #576`.** Step 8's comments on #156 and #401 were in fact posted on 2026-08-08, before this PR opened; the body's claim that they were outstanding was wrong.

**Deliberately not in this revision**, so they are not lost:

- `/jd-fit/` was in `sitemap.xml` and now hard-404s. `wrangler.jsonc` is assets-only, so a 301 needs a worker script — its own issue, not this PR.
- Deleting `useJdFitResume.contact-export.repro.test.tsx` leaves `/`'s `fieldConfidence` fold unpinned end-to-end. Surfacing that is the point; re-pinning it belongs with whoever owns that seam.
- Two departure paths to `/jobs/` disagree after an escape-hatch recovery: the Find-jobs tab's launcher sends `activeResult.canonical.fields` (the LLM re-parse), the header link sends `displayResult.canonical.fields` (the heuristic one), because `llmOverride` lives inside `Result` and `App` cannot see it. Pre-existing and unchanged here; the fingerprint's effect is that the header-link route now declines to steer rather than steering from the other parse. Fixing it properly means lifting `llmOverride`, which is its own change.

<details>
<summary>Round 1 responses</summary>

- **#1 (bfcache dead path)** — `ResultDetailTabs` now listens for `pageshow` in addition to the mount effect; the primary `history.back()` return path is what needed it.
- **#2 (StrictMode nulls jdContext)** — the mounted-flag guard is replaced with an identity-keyed `useRef` compare; the reset effect only fires on a real identity change, not on StrictMode's simulated replay.
- **#3 (`public/**` dead links)** — 13 `/jd-fit/` refs across `404.html`, `privacy/`, `open-source/`, `faq/`, `how-it-works/` are gone; nav items deleted, JD-URL-fetch prose in `privacy/` rewritten to point at the paste-a-JD panel on `/jobs/`, `faq/` + `how-it-works/` bodies rewritten to describe the one remaining lane.
- **#4 (`Closes #576` with Step 8 unmet)** — changed to `Refs #576`. Reverted in round 2: Step 8 was already met.
- **#5 (`flags.ts` fully dead)** — module deleted; the feature-flag machinery in `analytics.ts` (`getFeatureFlag`, `subscribeFeatureFlags`, the `flagSubs` set, PostHog's `onFeatureFlags`/`isFeatureEnabled` type members) trimmed alongside.
- **#6 (vacuous reset test)** — `ReconstructedResume` is mocked with a probe that renders its `jdContext` prop; the reset test now asserts on the propagated value, and two more tests were added (StrictMode replay does not null; a `pageshow` after mount picks up a mid-session handoff).
- **#7 (silent dead click on full coverage)** — the tailor button is gated in both `JobResultCard` and `PasteJdPanel`; new test covers the hidden case.
- **#8 (tautological `path === "keyword"` guard)** — `PasteJdPanel`'s `useMemo` return type narrowed to the keyword arm so `jdMatch && …` narrows to `.coverage` without the guard.
- **#9 (per-keystroke compute)** — 200ms debounce on the JD text before it feeds the extract/coverage pipeline.

</details>

## Test plan

- [x] `npm run build` emits exactly `/` and `/jobs/`; `dist/jd-fit/` gone; `appType: "mpa"` retained.
- [x] `grep -rn "jd-fit|jdFit|JdFit" src/ vite.config.ts public/` → empty; `grep -rn "ocv_jdfit_handoff" src/` → empty; `useFlag` no longer exists anywhere; `VITE_ENABLE_JD_FIT` gone.
- [x] `src/lib/jd-match/` functionally unchanged (2 docblock-only edits to correct references to the removed surface); `npx vitest run src/lib/job-search` → green.
- [x] Tests pin the load-bearing invariants:
  - `ResultDetailTabs.test.tsx` — mount consumes the handoff and `jdContext` reaches the `ReconstructedResume` probe; a parse-identity change nulls it; **an edit (shallow copy of `activeResult` under an unchanged identity) does NOT**; a handoff stamped for a different résumé is ignored and cleared; StrictMode replay does not null it; a `pageshow` after mount picks up a mid-session write.
  - `tailor-handoff.test.ts` — fingerprint stability + JSON-round-trip identity; round-trip, one-shot, wrong-fingerprint rejection, mismatch-clears, no-fingerprint rejection, malformed JSON, empty jdContext, quota-blocked storage.
  - `JobResultCard.test.tsx` — tailor button omitted without `onTailor`; when present, fires with the steering built from the exact `jdMatch.coverage` the card's fit number reads; hidden on a fully-covered posting.
  - `PasteJdPanel.test.tsx` (new) — the same three gate cases on the paste lane, so the panel's "same `onTailor` a `JobResultCard` uses" claim is pinned on both halves rather than asserted in a docblock.
- [x] Deleted tests, not skipped; `JdInput.test.ts:157`'s privacy contract unchanged.
- [x] `@design-system` imports only on new surfaces; no raw `<button>`, no hex, no palette classes.
- [x] `PasteJdPanel.tsx` under ~200 LOC + SPDX + docblock naming the constraint it guards.
- [x] `README.md`, `CLAUDE.md`, and the 5 public/content pages describe one root pair; CLAUDE.md's egress list documents `fetchJdFromUrl` (now via `PasteJdPanel`).
- [x] `tsc -b --noEmit`, `eslint .`, `check:fixtures`, `check:baselines`, `check:core` all clean; `vite build` emits exactly `/` and `/jobs/`.
- [ ] **`npm run verify` did NOT go fully green locally, and the reason is the workstation, not the branch.** Recorded rather than ticked, because anyone who runs the suite will see the same red:
  - Under `--coverage` on the pinned Node 20: 22 failures, **every one a 5s/15s/20s timeout**, zero assertion failures, all in `src/lib/pdf/**`, `src/lib/heuristics/**` and one `src/lib/edit/**` round-trip suite — none of them files this diff touches. That is the known coverage-instrumentation + worker-contention flake (#360/PR #389), and this run also emitted three `[vitest-worker]: Timeout calling "onTaskUpdate"` starvation errors.
  - Without coverage: 5199 pass, 10 fail — 9 of them `letter-egress-ack.test.ts` failing on `globalThis.localStorage?.clear is not a function`, which reproduces on the **unmodified** PR head `d7e8895` and is an artifact of running on Node 25 when `.nvmrc` and CI pin Node 20 (Node ≥22 ships a global `localStorage` the test's save/restore does not expect). The tenth is one more 5s timeout in `export-layout-contract`.
  - Every test file this change touches or adds is green, and CI `verify` was green in 3m21s on `d7e8895`. CI is the gate here.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation (round 1) | Claude Opus 4.7 (`claude-opus-4-7`) | — |
| Review | Claude Opus 5 | high |
| Review revisions (round 2) | Claude Opus 5 | high |

Tool: Claude Code CLI.


src/components/features/ResultDetailTabs.tsx
```diff
@@ -55,6 +54,48 @@ export function ResultDetailTabs({
...
+  // Null → generic rewrite prompt (byte-identical pre-#576).
+  const [jdContext, setJdContext] = useState<string | null>(null);
+
+  // Consume the one-shot handoff on mount. When present, also switch to the
+  // Reconstructed tab — the rewrite affordance lives there, and landing on
+  // any other tab would hide the change the user just asked for. The read
+  // clears the key, so a manual reload of `/` falls back to a generic
+  // rewrite prompt rather than silently keep steering toward a stale JD.
+  useEffect(() => {
+    const handoff = consumeTailorHandoff();
```
s-annam: **Blocking (1)** — this mount-only effect is the sole consumer of the handoff, so it never fires on the primary return path.

`JobsApp`'s tailor handler calls `returnToResumeRoot(arrivedFromRoot)`; in the normal flow (`/` → `FindJobsLauncher`/`departToJobs` → `/jobs/`) `arrivedFromRoot` is `true`, so `nav-return.ts:177` fires `history.back()`. `nav-return.ts`'s own docblock is explicit that the reason for `history.back()` is that `/`'s React state is "restored by bfcache on a genuine back navigation (measured on #706)" — and a bfcache restore does not remount the tree, so this `useEffect(…, [])` never re-runs.

Net effect: `jdContext` stays `null`, the tab doesn't switch, and the rewrite is generic. The key then lingers in `sessionStorage` and is consumed by the *next* genuine mount of `/`, steering an unrelated rewrite — the inverse of the one-shot guarantee this module's docblock promises.

It only works when bfcache *misses*, so it's non-deterministic rather than plainly broken. Consume on `pageshow` when `event.persisted` as well as on mount.

src/components/features/ResultDetailTabs.tsx
```diff
@@ -55,6 +54,48 @@ export function ResultDetailTabs({
...
+  //
+  // The `mountedRef` guard is what makes the mount-time handoff win against
+  // this effect: both effects fire on the first commit, in declaration order,
+  // so without the skip the reset would immediately null out the jdContext
+  // the handoff just set. After the first firing the guard flips and every
+  // real identity change (LLM override applied here, a fresh file remounts
+  // the whole tree so nothing to reset) is honoured. Dep is hand-audited
+  // (`exhaustive-deps` is NOT enforced in this repo — see CLAUDE.md → Data
+  // & hooks): only `activeResult` identity matters here.
+  const mountedRef = useRef(false);
```
s-annam: **Blocking (2)** — a boolean mounted flag can't survive StrictMode, so this effect wipes the `jdContext` the handoff just set whenever `src/main.tsx:37`'s `StrictMode` is active (i.e. every `npm run dev` session).

On the simulated remount: the handoff effect re-runs, finds the key already cleared, returns `null`; then this effect re-runs with `mountedRef.current` already `true` and calls `setJdContext(null)`. Refs persist across the replay; state does not.

I confirmed this empirically with a structural replica of both effects (same declaration order, same ref guard, same one-shot read) rendered under `StrictMode` + `createRoot` — it produced `jdContext=NULL` where `jdContext=steer-kubernetes` was expected.

This is the same class `useArrivedFromRoot`'s docblock already calls out ("the feature would be dead under `npm run dev`"). A boolean can't distinguish "replayed mount" from "identity changed"; key the guard on identity:

```ts
const prevResult = useRef(activeResult);
useEffect(() => {
  if (prevResult.current === activeResult) return;
  prevResult.current = activeResult;
  setJdContext(null);
}, [activeResult]);
```

src/components/features/ResultDetailTabs.test.tsx
```diff
@@ -252,4 +257,77 @@ describe("ResultDetailTabs", () => {
...
+    // Consumed: the read cleared the key.
+    expect(sessionStorage.getItem(TAILOR_HANDOFF_KEY)).toBeNull();
+    // Landed on Reconstructed — the tab strip announces the selection via
+    // `aria-selected`, so the assertion is on that rather than on textContent
+    // (which lists every tab label regardless of which panel is showing).
+    const selected = el.querySelector('[role="tab"][aria-selected="true"]');
+    expect(selected?.textContent).toContain("Your resume");
+  });
+
+  it("does not consume the tailor handoff again after an activeResult swap (#576)", () => {
```
s-annam: **Secondary (6)** — this test can't fail. It asserts only that `sessionStorage.getItem(TAILOR_HANDOFF_KEY)` is `null` before and after the swap, and nothing in the component ever writes that key back — so the assertion holds identically with the `mountedRef` guard deleted. The comment concedes as much ("both would be nulls; a re-read would be a no-op").

The invariant that needs pinning is that `jdContext` *survives* the mount and *resets* on a real identity change, which needs an observable for the value reaching the rewrite path. That observable is also what would have caught the StrictMode wipe (finding 2). Worth removing from the description's "pins the load-bearing invariants" list until it does.

src/jobs/JobsApp.tsx
```diff
@@ -96,6 +98,21 @@ export default function JobsApp() {
...
+  // disclosure below the results) stashes the rewrite steering in
+  // sessionStorage and navigates back to `/`, where `ResultDetailTabs`
+  // consumes the handoff on mount, sets `jdContext`, and switches to the
+  // Reconstructed tab. `buildJdRewriteContext` returns null when there is
+  // nothing useful to steer with (e.g. every JD term is already covered), in
+  // which case we don't navigate — a no-op leaves the user on the posting
+  // instead of shipping them home for a generic rewrite prompt.
+  const handleTailor = (coverage: CoverageResult) => {
+    const jdContext = buildJdRewriteContext(coverage);
+    if (jdContext === null) return;
```
s-annam: **Secondary (7)** — silent dead click. `buildJdRewriteContext` returns `null` when `coverage.missing` is empty, so on a posting the résumé already fully covers, clicking "Tailor résumé to this job" does nothing at all: no navigation, no message, no disabled state. The button renders unconditionally on every card, so the user gets no signal distinguishing "nothing to tailor" from "the button is broken".

Either gate the button on `coverage.missing.length > 0` at the render site, or navigate anyway and let `/` say why the rewrite is generic.

src/components/features/PasteJdPanel.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+      {open && (
+        <>
+          <JdInput
+            value={jdText}
+            onChange={setJdText}
+            resumeParsed={true}
+          />
+          {jdMatch && <JdMatch result={jdMatch} />}
+          {jdMatch?.path === "keyword" && onTailor && (
```
s-annam: **Nit (8)** — `jdMatch?.path === "keyword"` is a tautology: `path` is hardcoded to `"keyword"` in the `useMemo` at line 58 of this same file, so the comparison is true whenever `jdMatch` is non-null. `{jdMatch && onTailor && (` is equivalent.

(Inherited verbatim from `JdFitApp`, where it was equally tautological — worth dropping rather than carrying forward.)

s-annam (changes requested):
Careful, well-documented refactor — the deletion half is clean (build emits `/` + `/jobs/` only, `dist/jd-fit/` gone, sitemap updated, ESLint entry-root scope and both `CLAUDE.md`s kept in step, egress list now documents `fetchJdFromUrl`). `typecheck`, `lint`, `check:fixtures` (58 PDFs), and 5170 tests are green on `0a81d70`. The migration half is where I'm asking for changes: **the JD-driven rewrite does not actually reach `/` on the primary return path**, and the removed page's links still ship.

## Blocking

**1. The tailor handoff never lands on the canonical return path (bfcache).** `src/jobs/JobsApp.tsx:113` calls `returnToResumeRoot(arrivedFromRoot)`. In the normal flow — parse on `/` → `FindJobsLauncher`/`departToJobs` marks departure → `/jobs/` — `arrivedFromRoot` is `true`, so `nav-return.ts:177` fires `win.history.back()`. That module's own docblock states the whole point of `history.back()` is that `/`'s React state is *"restored by bfcache on a genuine back navigation (measured on #706)"*. A bfcache restore does **not** remount the React tree, so `ResultDetailTabs.tsx:69`'s `useEffect(…, [])` never re-runs and `consumeTailorHandoff()` is never called.

Repro: parse a résumé on `/` → Find jobs → Search → "Tailor résumé to this job" on any card → back on `/`, Reconstructed tab, rewrite is generic. `jdContext` stays `null`; the tab switch doesn't happen either. This is the AC *"clicking 'Tailor résumé to this job' … switches to the Reconstructed tab and the subsequent rewrite is steered by that posting's missing terms"*, failing on the main path.

It also inverts the one-shot guarantee: the key survives in `sessionStorage` and is consumed by the *next* genuine mount of `/`, so a later reload steers an unrelated rewrite toward a posting from earlier — the opposite of the docblock's *"a manual reload of `/` falls back to the plain rewrite prompt."*

The feature only works when bfcache **misses** (eviction, or `arrivedFromRoot === false` → `location.href` full load), so it's non-deterministic rather than simply broken. Fix: also consume on `pageshow` when `event.persisted`.

**2. StrictMode double-invoke nulls the consumed `jdContext` — dead under `npm run dev`.** `src/main.tsx:37` wraps `App` in `StrictMode`. On the simulated remount the handoff effect re-runs, finds the key already cleared, and returns `null`; the reset effect re-runs with `mountedRef.current` already `true` and calls `setJdContext(null)`. Refs persist across the simulated remount; state does not.

Confirmed empirically with a structural replica of both effects (same declaration order, same ref guard, same one-shot read) under `StrictMode` + `createRoot`: `jdContext=NULL` where `jdContext=steer-kubernetes` was expected. This is exactly the class `useArrivedFromRoot`'s docblock calls out — *"the feature would be dead under `npm run dev`"* — and solves with a pure read + idempotent clear.

A boolean mounted flag can't distinguish "replayed mount" from "identity changed". Key the guard on identity instead:

```ts
const prevResult = useRef(activeResult);
useEffect(() => {
  if (prevResult.current === activeResult) return;
  prevResult.current = activeResult;
  setJdContext(null);
}, [activeResult]);
```

**3. 13 `/jd-fit/` links ship dead in `public/`.** These are copied verbatim into `dist/` (verified: `dist/faq/index.html` et al. exist), so every one 404s after merge:

- `public/404.html:117` — `<a href="/jd-fit/">Match a job description</a>`. The page a user lands on *after* following a dead link offers the same dead link.
- `public/privacy/index.html:47,129,238`
- `public/open-source/index.html:40,150`
- `public/faq/index.html:41,128,163`
- `public/how-it-works/index.html:44,239,260`

Lines `:47`/`:40`/`:41`/`:44` are the shared site nav, so this is every static content page's header.

Two of them are also wrong prose, not just dead hrefs: `privacy/index.html:129` describes the JD **URL-fetch egress** as happening on "JD fit" — that egress moved to `PasteJdPanel` on `/jobs/` in this PR, so the privacy page now points at a deleted surface for a live outbound path. `faq/index.html:128` still says *"Two more lanes. JD fit takes a job description…"*.

The Step-6 sweep grep was scoped `src/ vite.config.ts README.md docs/`; `public/` wasn't in it, which is why this slipped. The green AC grep isn't evidence here.

**4. `Closes #576` with the Step-8 AC unmet.** Verified: zero mentions of `576` on #156 or #401. The PR discloses this honestly, but merging closes #576 and #203–#207 stay silently pointed at a deleted page — the outcome the issue explicitly forbids. Either post the two comments, or change `Closes #576` → `Refs #576` so the AC stays open.

## Secondary

**5. `src/lib/flags.ts` is now fully dead.** `App.tsx:79`'s `useFlag("jd-fit-banner")` was the only `useFlag` caller anywhere (`grep -rn useFlag src/` now returns only `flags.ts` itself). fallow reports it as an unused file, new against `origin/main` (whose own audit reports `dead files 0.0%`). The `llm` flag / `VITE_ENABLE_LLM` and `subscribeFeatureFlags` in `analytics.ts` are now unreachable machinery. Report-only in `verify`, so not blocking — but Step 6's orphan sweep should have caught a whole orphaned module.

**6. The `activeResult`-swap test can't fail.** `ResultDetailTabs.test.tsx:278` asserts only that `sessionStorage.getItem(TAILOR_HANDOFF_KEY)` is `null` before and after the swap. Nothing ever writes that key back, so the assertion holds identically with the `mountedRef` guard deleted — the test's own comment concedes it (*"both would be nulls; a re-read would be a no-op"*). It's listed under "New tests pin the load-bearing invariants"; it doesn't pin this one. Needs an observable for `jdContext` reaching the rewrite path (this is also what would have caught finding 2).

**7. Silent dead click when the résumé already covers the posting.** `JobsApp.tsx:111` — `buildJdRewriteContext` returns `null` when `coverage.missing` is empty, and `handleTailor` then returns with no navigation, no message, no disabled state. The button renders unconditionally on every card, so on a high-fit posting the primary affordance does nothing. Either don't render it when `coverage.missing.length === 0`, or navigate and say why nothing changed.

## Nits (non-blocking)

**8.** `PasteJdPanel.tsx:95` — `jdMatch?.path === "keyword"` is a tautology; `path` is hardcoded `"keyword"` at line 58 of the same file. `jdMatch && onTailor` is equivalent. (Inherited verbatim from `JdFitApp`, where it was equally tautological.)

**9.** `PasteJdPanel.tsx:52` recomputes `extractJdTerms` + `computeCoverage` over the full JD on every keystroke. Not a regression — but `JdFitApp` ran it on a page that did nothing else, and it now runs inside `FindJobsPanel` alongside a mounted results list. A debounce would be cheap.

**10.** *(FYI, not a finding)* fallow also lists `src/jobs/main.tsx` as unused. That looks like diff re-attribution — it's a real Rollup input via `HTML_ENTRIES` — so don't chase it.

## Gates

| Gate | Result |
|---|---|
| 3a Fixture PII | Skipped — no fixture binaries in the diff. `check:fixtures` run anyway: 58 PDFs + 15 sidecars clean |
| 3b Design-system / reuse | Pass — `Button` primitive used throughout; `PasteJdPanel` is a 110-LOC sibling with SPDX + docblock, reuses `JdInput`/`JdMatch` unchanged; no raw `<button>` |
| 3c Style tokens | Pass — semantic tokens only (`border-border-light`, `text-content-tertiary`); `lint` clean |
| 3d fallow | 2 unused files (finding 5 + the artifact in 10), 6 complexity, 13 clone groups — report-only |
| 3e Command-level bugs | N/A — no skill/script/bash changes |
| 3f Description accuracy | See below |

**3f verdict — accurate but incomplete.** The "Architectural note" honestly explains the `/jobs/` divergence, the disclosures on the +4 LOC and the two `jd-match` docblock edits are precise, and Step 8 is declared not-done rather than ticked. Two problems: it lists the `activeResult`-swap test as pinning an invariant it can't pin (6), and it omits the `public/**` link breakage (3) and the newly-orphaned `flags.ts` (5) entirely. The `[x] npm run verify green` box holds — I reproduced typecheck, lint, tests, build, and fixtures independently.

Reviewed at `0a81d70c4a492e509cf412b418b94c5a03bda021`.

---

src/components/features/FindJobsPanel.tsx
```diff
@@ -97,9 +99,10 @@
    *  education) for accurate coverage, so we take the whole `HeuristicParsedResume`
    *  rather than the narrow query-only Pick. */
   parsed: HeuristicParsedResume;
+  onTailor?: (coverage: CoverageResult) => void;
 }
 
-export function FindJobsPanel({ parsed }: FindJobsPanelProps) {
+export function FindJobsPanel({ parsed, onTailor }: FindJobsPanelProps) {
```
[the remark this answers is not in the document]
Vaishnavi1709: Report-only per CLAUDE.md → `npm run verify` fallow step: `fallow audit exited non-zero (report-only, ignored)`. `FindJobsPanel` sits over the ~200 LOC guideline already (known debt list alongside `ReconstructedResume`, `ReconstructedRole`, `ModelSelector`, `SectionRewrite` — see the CLAUDE.md "Component architecture & reuse" section: *"⚠️ Known debt — do not imitate"*). This PR added +4 LOC to it (2 imports + 1 prop + 1 render — the atomic minimum to thread `onTailor` and mount `PasteJdPanel`, disclosed in the PR body). Decomposition is out of scope for #576; resolving as documented policy.

src/components/features/PasteJdPanel.tsx
```diff
@@ -0,0 +1,110 @@
...
+interface PasteJdPanelProps {
+  /** The parsed résumé the coverage check runs against — same shape
+   *  `FindJobsPanel` and `JobResultCard` consume for their own coverage. */
+  parsed: HeuristicParsedResume;
+  /** Steer the rewrite on `/` with this JD. Optional so the panel renders
+   *  its coverage view even outside `/jobs/`'s handoff-back-to-`/` context. */
+  onTailor?: (coverage: CoverageResult) => void;
+}
+
+export function PasteJdPanel({ parsed, onTailor }: PasteJdPanelProps) {
```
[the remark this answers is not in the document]
Vaishnavi1709: Report-only per CLAUDE.md fallow policy. `PasteJdPanel` is 110 LOC end-to-end — the CRAP score is inflated by the 0% direct coverage on a display-only component whose behaviour is exercised transitively: the coverage compute is `computeCoverage` (own test suite), `JdInput` is reused unchanged (its own test suite covers the URL-fetch privacy contract), and the disclosure toggle is `useState<boolean>`. Behaviour tests for `PasteJdPanel` itself would essentially re-test its sub-components. Resolving as documented policy.

src/components/features/ResultDetailTabs.tsx
Vaishnavi1709: Fixed in d7e8895. `ResultDetailTabs` now listens for `pageshow` in addition to the mount effect — the same handler fires on the initial page load (`persisted === false`) and on every bfcache restore (`persisted === true`), so `history.back()` from `/jobs/` picks the handoff up. Also swept: `consumeTailorHandoff` stays one-shot, so a duplicate cold mount + pageshow is a no-op; a spurious pageshow with no handoff key is a no-op.

New test at `ResultDetailTabs.test.tsx` — *"consumes a handoff written AFTER mount, on the next pageshow (#576, bfcache)"* — mounts with no handoff, writes one, dispatches `Event("pageshow")` on window, and asserts jdContext reaches the `ReconstructedResume` probe. The pre-fix code fails that test.

src/components/features/ResultDetailTabs.tsx
Vaishnavi1709: Fixed in d7e8895. Replaced the boolean `mountedRef` with an identity-keyed `useRef(activeResult)`:

```ts
const prevResult = useRef(activeResult);
useEffect(() => {
  if (prevResult.current === activeResult) return;
  prevResult.current = activeResult;
  setJdContext(null);
}, [activeResult]);
```

The reset effect early-returns when `prevResult.current === activeResult`, so StrictMode's simulated setup → cleanup → setup replay is a no-op (identity did not change between the two setups), and a genuine `activeResult` swap (LLM escape hatch recovering a parse) still fires.

Two new tests at `ResultDetailTabs.test.tsx`:
- *"survives StrictMode's simulated remount without nulling jdContext"* — mounts under `<StrictMode>` explicitly and asserts the value survives.
- *"keeps jdContext on mount and nulls it when activeResult identity changes"* — pins both directions in one test via the `ReconstructedResume` probe.

src/components/features/ResultDetailTabs.test.tsx
Vaishnavi1709: Fixed in d7e8895. `ReconstructedResume` is now `vi.mock`ed at the top of `ResultDetailTabs.test.tsx` with a probe that renders `data-testid="reconstructed-probe"` and prints `jdContext=<value>` (or `jdContext=NULL`) as text. Every jdContext-related test now asserts on the propagated value, not just on sessionStorage state:

- `"consumes a tailor handoff on mount and lands on the Reconstructed tab"` — asserts probe text after a mount consume.
- `"keeps jdContext on mount and nulls it when activeResult identity changes"` — asserts probe text before AND after the swap.
- `"survives StrictMode's simulated remount without nulling jdContext"` — asserts probe text under `<StrictMode>`.
- `"consumes a handoff written AFTER mount, on the next pageshow"` — asserts probe text after a synthetic `pageshow`.

Confirmed the reset test now fails against the pre-fix `mountedRef` shape (StrictMode replay case), so it would have caught finding #2 empirically as you noted.

src/jobs/JobsApp.tsx
Vaishnavi1709: Fixed in d7e8895. Both `JobResultCard` and `PasteJdPanel` gate the Tailor button render on `jdMatch.coverage.missing.length > 0`, so the affordance disappears on a fully-covered posting — `JobsApp.handleTailor` keeps its `buildJdRewriteContext === null` guard as belt-and-suspenders, but the button never fires it now. Chose "hide" over "disabled+tooltip" because a disabled primary affordance is louder than a missing one on a card the user already reads for signal.

New test at `JobResultCard.test.tsx` — *"hides the Tailor button on a fully-covered posting"* — mounts a synthetic full-coverage résumé, asserts `job.jdMatch.coverage.missing` is empty, and asserts the Tailor button is absent from the DOM.

src/components/features/PasteJdPanel.tsx
Vaishnavi1709: Fixed in d7e8895. Narrowed `PasteJdPanel`'s `useMemo` return type to the keyword arm of the discriminated union:

```ts
type KeywordJdMatch = Extract<JdMatchResult, { path: "keyword" }>;
const jdMatch = useMemo<KeywordJdMatch | null>(() => { … });
```

That kills the tautology at the call site — the render-side check is now just `jdMatch && onTailor && jdMatch.coverage.missing.length > 0` (the extra `missing.length` gate is finding #7's fix; no more `?.path === "keyword"`). While I was in there I also added the 200ms debounce from nit #9.

src/components/features/ResultDetailTabs.tsx
```diff
@@ -55,6 +54,72 @@ export function ResultDetailTabs({
...
+  // set. `useArrivedFromRoot`'s docblock calls out exactly this class of
+  // bug ("dead under `npm run dev`"). Comparing identity means the second
+  // setup returns early (activeResult === prevResult.current) instead of
+  // firing on a "change" that never happened.
+  const prevResult = useRef(activeResult);
+  useEffect(() => {
+    if (prevResult.current === activeResult) return;
+    prevResult.current = activeResult;
+    setJdContext(null);
+  }, [activeResult]);
```
s-annam: **Blocking — the reset nulls `jdContext` on every inline edit, not just on a parse change.**

`activeResult` is not a parse identity. `App.tsx:279` passes `result={displayResult}`; `Result.tsx:126-128` forwards it verbatim as `activeResult` when `llmOverride === null`; and `displayResult` (`useAnalyzedResume.ts:294`) is `useMemo(…, [base, edited])` where `edited` → `editedCore` rebuilds from `applyOverrides` on **any** override change (`useAnalyzedResume.ts:148`). So one keystroke in an `EditableField` mints a new object with identical content, the identity compare sees a "change", and `setJdContext(null)` runs.

Failure: tailor from `/jobs/` → land on `/` with the Reconstructed tab open (the tab where editing happens) → fix one bullet → click Rewrite → the model gets the generic prompt. Silent, no UI signal. This breaks the issue's AC *"clicking 'Tailor résumé to this job' … the subsequent rewrite is steered by that posting's missing terms"*.

Repro, run against `d7e8895` (append to `ResultDetailTabs.test.tsx`, reusing its helpers — swaps `activeResult` for a shallow copy, which is exactly what an edit produces):

```tsx
writeTailorHandoff({ jdContext: "surface Kubernetes" });
const heuristic = result(undefined, "Heuristic Engineer");
function EditHost({ edited }: { edited: boolean }) {
  const edit = useEditableParse();
  return createElement(ResultDetailTabs, {
    activeResult: edited ? { ...heuristic } : heuristic,   // same content, new identity
    activeScore: score, result: heuristic, sourceKind: "pdf", edit,
    analysis: controller(opts), escapeHatch: escapeHatchController(opts),
    onRecovered: () => {}, triggerCount: heuristic.triggers.length,
  });
}
// after re-render with edited: true →
// AssertionError: expected 'jdContext=NULL' to be 'jdContext=surface Kubernetes'
```

The existing identity-swap test passes because it swaps to a *different* result, which is also what an edit looks like — the test cannot tell the two apart.

Fix: key the reset on the two events it names, not on `activeResult`. Neither `result` nor `rawText` works as a substitute — `result` is also `displayResult`, and `mergeLlmParse` (`merge-override.ts:34`) spreads `...result`, so `rawText` survives an escape-hatch recovery. `App.tsx` does hold a stable pristine parse (`state.result`, unchanged by edits, new on every file); thread that down and combine it with `Result.tsx`'s `llmOverride` identity:

```tsx
// reset only on: new parse, or escape-hatch recovery
useEffect(() => { setJdContext(null); }, [pristineResult, llmOverride]);
```

and add a test that re-renders with a shallow copy of the same result and asserts `jdContext` survives.

src/jobs/JobsApp.tsx
```diff
@@ -96,6 +98,21 @@ export default function JobsApp() {
...
+    writeTailorHandoff({ jdContext });
+    returnToResumeRoot(arrivedFromRoot);
```
s-annam: **Secondary — the handoff is stranded when the return is a fresh navigation, and the next résumé consumes it.**

`handleTailor` writes unconditionally, then `returnToResumeRoot(arrivedFromRoot)`. When `arrivedFromRoot` is `false` — a reload or deep link of `/jobs/`, which is a supported path precisely because `readJobsHandoff` is non-destructive — that is `location.href = BASE_URL`: a fresh `/` that boots idle, so `ResultDetailTabs` never mounts and neither the mount read nor `pageshow` consumes the key. Same on a `history.back()` that misses bfcache, and when `Result.tsx` early-returns `FontsUnmappableCard` instead of the tabs.

`ocv_tailor_handoff` then sits in sessionStorage until the *next* `ResultDetailTabs` mount — i.e. whatever résumé the user drops next, steered by a different résumé's missing-terms list. That is the failure `ResultDetailTabs`' own docblock says the design prevents, and it violates the AC *"a stale tailoring instruction never leaks across résumés"* from the other direction than finding #1.

Fix: stamp the payload with the parse identity it was derived from and have `consumeTailorHandoff` reject a mismatch — that closes the fresh-navigation, the bfcache-miss, and the fonts-unmappable paths together, rather than special-casing `arrivedFromRoot`.

src/components/features/JobResultCard.tsx
```diff
@@ -146,6 +157,23 @@ export function JobResultCard({ job }: { job: RankedJob }) {
...
         </Button>
+        {/* Gate on `missing.length > 0` (PR #783 review, secondary finding
+            #7): `buildJdRewriteContext` returns null when there is nothing to
+            steer with — the previous button rendered unconditionally and
+            silently no-op'd on a click, so a high-fit posting had a primary
+            affordance that did nothing. Hiding it means the button always
+            leads somewhere. The reason lives next to the discovery view
+            below (`View match detail` already tells the user the coverage
+            is complete). */}
+        {onTailor && jdMatch.coverage.missing.length > 0 && (
```
s-annam: **Secondary — the render gate and the click handler use different predicates.** (Same at `PasteJdPanel.tsx:122`.)

The gate is `coverage.missing.length > 0`; the handler bails on `buildJdRewriteContext(coverage) === null`, which is `missing.map(t => t.display.trim()).filter(s => s.length > 0)` being empty (`rewrite-context.ts:39-43`). A missing term whose `display` trims to empty passes the gate and yields `null` — the button renders and the click silently no-ops, the exact defect the gate was added for.

Not reachable on today's `extractJdTerms` output, but it's the same predicate written three ways across three files. Call `buildJdRewriteContext` once and gate on its non-null result so the two agree by construction.

src/components/features/PasteJdPanel.tsx
```diff
@@ -0,0 +1,137 @@
...
+import { extractJdTerms, computeCoverage } from "../../lib/jd-match";
+import type { CoverageResult } from "../../lib/jd-match/coverage.ts";
+import type { JdMatchResult } from "../../lib/jd-match";
+import type { HeuristicParsedResume } from "../../lib/heuristics/types.ts";
+
+/** The keyword arm of the discriminated union — the only path this panel
+ *  produces. Narrowing here means `jdMatch && …` narrows all the way to
+ *  `.coverage` without a tautological `.path === "keyword"` guard on every
+ *  use (PR #783 review, nit #8). */
+type KeywordJdMatch = Extract<JdMatchResult, { path: "keyword" }>;
```
s-annam: **Secondary (reuse) — `KeywordJdMatch` is redeclared.**

`rank.ts:70` already exports `export type KeywordJdMatch = Extract<JdMatchResult, { path: "keyword" }>`, and `rank.ts:239-244` already assembles the identical `{ path, coverage, terms, nounsDropped }` literal that `PasteJdPanel.tsx:75-87` rebuilds. Import the type and factor the assembly — the panel's docblock claims a single composition, which holds for `computeCoverage` but not for the object built around it.

s-annam (changes requested):
Careful, well-documented migration — the deletion sweep is complete (13 `public/**` refs, the vite plugin, the flag module + its PostHog machinery, all with no dangling references), the docs were rewritten rather than patched, and the new tests pin real observables via the `ReconstructedResume` probe. Step 8's comments on #156/#401 are already posted, so `Refs #576` can become `Closes`.

One blocking defect: the `jdContext` reset fires far more often than the docblock claims, and it kills the migrated feature on the primary path.

**Findings 1–4 (1 Blocking, 3 Secondary) are inline on the lines they concern.** Summary:

| # | Severity | Where | What |
|---|---|---|---|
| 1 | **Blocking** | `ResultDetailTabs.tsx:116-121` | reset nulls `jdContext` on every inline edit, not just on a parse change |
| 2 | Secondary | `JobsApp.tsx:109-113` | handoff stranded on a fresh-navigation return; the next résumé consumes it |
| 3 | Secondary | `JobResultCard.tsx:168`, `PasteJdPanel.tsx:122` | render gate and click handler use different predicates |
| 4 | Secondary | `PasteJdPanel.tsx:39` | `KeywordJdMatch` + its assembly redeclared; `rank.ts:70,239` already has both |

## Nits (non-blocking)

**5. No redirect for `/jd-fit/`.** It was a `HTML_ENTRIES` member, so it shipped in the sitemap and is plausibly indexed; `wrangler.jsonc` is assets-only (`not_found_handling: "404-page"`) with no worker script, so it hard-404s after deploy. The 404 page does link `/jobs/`, and a 301 would need a worker — out of scope here, but worth an issue rather than nothing.

**6. Code comments cite an ephemeral review numbering.** `PasteJdPanel.tsx:36,42`, `JobResultCard.tsx:160`, `ResultDetailTabs.tsx:65,110` and several test comments say "PR #783 review, blocking finding #1 / nit #9". Once squashed into `main`, "nit #9" resolves to nothing. The *reasons* in those comments are excellent — keep them, drop the citations.

**7. PR body is stale on Step 8.** It says the #156/#401 comments are "not yet posted"; both landed 2026-08-08. `Refs #576` can become `Closes #576`.

**8. Coverage note.** `useJdFitResume.contact-export.repro.test.tsx` (#648) is correctly deleted with its subject, and the duplicate-bullet sibling has a `/`-side twin (`useEditableParse.duplicate-bullet-identity.repro.test.tsx`). The contact-export one does not — `/`'s `fieldConfidence` fold into `buildContact` is now unpinned end-to-end. No regression introduced; flagging so it's a choice.

## Gates

- `npm run typecheck`, `npx eslint .`, `npm run check:fixtures` (58 PDFs + 15 sidecars) — all green. CI `verify` + `fallow` green on `d7e8895`.
- Fixture PII gate: no fixtures touched, N/A beyond the clean run.
- Design system / tokens: clean — `PasteJdPanel` is 137 LOC with SPDX + a docblock naming its constraint, `@design-system` imports only, no raw `<button>`, no hex or palette classes.
- Orphan sweep: `grep -rn "jd-fit\|jdFit\|JdFit" src/ vite.config.ts public/` and `grep -rn "ocv_jdfit_handoff" src/` both empty; `VITE_ENABLE_LLM`/`VITE_ENABLE_JD_FIT` gone tree-wide.
- Description audit (read last): accurate except the reset claim — *"resets when `activeResult` identity changes (LLM escape hatch)"* — where the parenthetical understates the trigger by a wide margin (finding #1), and the stale Step 8 line (nit #7).
- The AC *"`npm run build` emits exactly one HTML entry"* is superseded by #690; shipping `/` + `/jobs/` is correct, not a gap.

Reviewed at `d7e8895`.

---

s-annam:
Taking this the rest of the way ourselves rather than sending it back for another
round — an amended revision lands on this branch shortly.

**Why we're doing it rather than asking you to.** The blocking finding and the
second secondary one turned out to be one root cause: nothing below `App` carries
a stable parse identity. `activeResult` looks like one, but it *is* `displayResult`,
which re-memoizes on every override change — so the reset over-fires on an edit
(#1), and the handoff has nothing to bind itself to when it strands (#2). Fixed
separately they'd become two mechanisms where one will do; fixed together it's a
single thread of the pristine parse (`state.result`, already sitting in `App`)
used as both the reset key and the handoff's stamp. That's a shape better decided
in one pass than negotiated over a comment thread.

None of this is a knock on the migration. The deletion sweep was complete, the
docs were rewritten rather than patched, and the `ReconstructedResume` probe you
added in the last revision is the only reason the defect was observable at all.

**What lands in the amended commit**
- **#1** — reset keyed on the pristine parse + `llmOverride`, not `activeResult`;
  plus a test that a shallow copy (what an edit produces) does *not* clear `jdContext`.
- **#2** — handoff payload stamped with that same parse identity;
  `consumeTailorHandoff` rejects a mismatch. Closes the fresh-navigation,
  bfcache-miss, and `FontsUnmappableCard` paths together.
- **#3** — `buildJdRewriteContext` called once and gated on its non-null result,
  at both call sites.
- **#4** — `KeywordJdMatch` imported from `rank.ts` rather than redeclared.
- **Nit 6** — review-numbering citations dropped from the code comments; the
  reasoning in them stays, it was good.
- **Nit 7** — body updated: the Step 8 comments on #156/#401 are in fact already
  posted, so this becomes `Closes #576`.

**Not in this revision**
- **Nit 5** (no 301 for `/jd-fit/`) — needs a worker script, so it gets its own
  issue instead of riding along here.
- **Nit 8** — the now-unpinned `/`-side `fieldConfidence` fold is left as-is;
  surfacing it was the point, not fixing it.

**Amend, not a new commit.** `main` merges through a queue that derives the squash
message from the branch, so this branch stays at one commit — force-push incoming
on `gh-576-deprecate-jd-fit-page`. Pull before you touch it again.

src/components/features/ResultDetailTabs.tsx
s-annam: Fixed in `e9a7fc4` — and the fix is not the one this comment asked for, so it's worth saying why.

The reset is now keyed on a `parseIdentity` token rather than `activeResult`. `activeResult` cannot serve as a parse identity at all: on `/` it is `useAnalyzedResume`'s `displayResult`, a memo over the edit override maps, so **every keystroke in the inline editor mints a fresh object for the same parse**. Keyed on it, this effect fired mid-edit and nulled the steering the user had just come back from `/jobs/` to apply — the feature died on the first character typed.

`useAnalyzedResume` already computed the correct answer for its own clear-edits effect; it is now exported as `parseKey`, and `Result.tsx` combines it with `llmOverride` (the other, and only other, way a parse is replaced under a mounted card) into the token. One definition, so the two cannot drift.

The lifecycle moved into `src/hooks/useTailorHandoff.ts` — `ResultDetailTabs` was already 259 LOC on `main` and CLAUDE.md asks for a sibling rather than more growth; the file now lands at 283 instead of 324. Your StrictMode ref-vs-flag reasoning came along verbatim, it was right.

New test renders a rebuilt `canonical.fields` (what the display memo actually produces, not a bare `{ ...result }` — that shares `canonical` and would pass either way) under an unchanged identity and asserts `jdContext` survives. It fails against the previous shape with `expected 'jdContext=NULL' to be 'jdContext=surface Kubernetes'`.

src/jobs/JobsApp.tsx
s-annam: Fixed in `e9a7fc4`.

Two changes here. `handleTailor` now takes the **built** steering rather than raw coverage, so `buildJdRewriteContext` is called once, at the surface that has to decide whether the button leads anywhere — the render gate and the payload can no longer be two different predicates.

Second, and this is the part the original design was missing: the payload is stamped with `fingerprintParse(handoff.parsed)`, and `consumeTailorHandoff(expected)` discards a mismatch. One-shot bounds how many times a payload is read, not **which** résumé reads it. The tailor flow depends on a bfcache restore of `/`; when that doesn't happen — tab reloaded, résumé reset, a different file dropped — the next consumer is a different parse and absorbs the key exactly once, against the wrong résumé, steering a rewrite toward gaps computed from a file the user already replaced. A mismatch is cleared rather than left in place, so it can't ambush a later mount either.

Content-derived (FNV-1a over the fields) rather than a minted counter, because the id has to survive the sessionStorage round trip and a cold page load, where a counter restarting at zero would collide.

One asymmetry I did **not** fix, noted in the PR body so it isn't lost: the header link to `/jobs/` sends `displayResult.canonical.fields` while this tab's launcher sends `activeResult.canonical.fields`, so after an escape-hatch recovery they disagree. Pre-existing; the fingerprint's only effect is that route now declines to steer instead of steering from the other parse. Fixing it properly means lifting `llmOverride` out of `Result`.

src/components/features/JobResultCard.tsx
s-annam: Fixed in `e9a7fc4`.

The gate is now `buildJdRewriteContext(jdMatch.coverage) !== null` — one `useMemo` whose result is both the visibility condition and the value handed to `onTailor`.

`missing.length > 0` was the right instinct but a strictly wider predicate: the builder trims each term's `display` and filters empties before deciding, so a missing term whose display trims to nothing clears your check and still yields null. Deriving the button from the builder's own result means the two can't drift again — which matters more than the specific edge case, because "is there anything to steer with" now has exactly one definition.

The `(PR #783 review, secondary finding #7)` citation is gone from the comment, reasoning kept. Review-round numbering doesn't survive the merge-queue squash and isn't stable across rounds; issue numbers are.

src/components/features/PasteJdPanel.tsx
s-annam: Fixed in `e9a7fc4`.

`KeywordJdMatch` is imported from `rank.ts:70`, which already exported it — type-only, so no runtime cost and no chunk added. The narrowing you wanted is unchanged; it just has one definition now instead of two that were free to drift.

Same gate change as `JobResultCard` applied here, and the panel gained `PasteJdPanel.test.tsx`. That file's docblock claimed it fed "the same `onTailor` a `JobResultCard` uses" — which was exactly the thing that had stopped being true, and nothing tested it. Now both halves are pinned: button offered with the built steering, hidden on a fully-covered JD, absent without `onTailor`.

Citations dropped from the `nit #8` / `nit #9` comments; the debounce rationale was good and is kept as-is.

s-annam (approved):
Approving — with the caveat stated plainly, because it changes what this approval is worth.

**Round 2 was authored by me, the reviewer.** So this is not an independent second read of the fixes; it is the person who wrote them signing them off, at the maintainer's explicit direction. What it does certify:

- The two blocking defects have **fail-before proofs**, not just assertions. Keyed on the fields object, the edit-survives test goes red with `expected 'jdContext=NULL' to be 'jdContext=surface Kubernetes'`; with the fingerprint check removed, four tests go red across two files. Both were run and observed, not reasoned about.
- CI `verify` is green on `e9a7fc4` (3m18s), alongside `fallow` and the Cloudflare Pages build.
- Every acceptance criterion of #576 is met, including Step 8 — the comments on #156 and #401 were posted 2026-08-08, which is why the body went back to `Closes #576`.

What it does **not** certify, and what a fresh pair of eyes would be worth on:

- The `onTailor` contract change (`CoverageResult` → the built instruction string) touches five pass-through components. Mechanical, type-checked, but wide.
- `useTailorHandoff` is a new extraction. `ResultDetailTabs` was 259 LOC on `main` and this PR had taken it to 324; it now lands at 283, and the three invariants that each silently kill this feature live in one docblock instead of interleaved with tab-label logic. That is a structural judgement call, not a review finding.
- The parse fingerprint is content-derived (FNV-1a over the fields). Collision resistance is not a security property here — worst case on a collision is the pre-#576 behaviour — but it is an assumption worth someone disagreeing with.

Three things were deliberately left out and are recorded rather than quietly dropped: the `/jd-fit/` retirement forwarding decision (now #799), the header-link vs launcher parse asymmetry after an escape-hatch recovery (pre-existing; needs `llmOverride` lifted out of `Result`), and the `fieldConfidence` fold left unpinned by deleting `useJdFitResume.contact-export.repro.test.tsx` — surfacing that one was the point.

Local `verify` does not go green on my workstation and the PR body says so instead of ticking the box: every local failure is a timeout in untouched heavy PDF/heuristics suites (#762) or the Node 22+ `localStorage` shadowing in #779. Four runs on this same SHA produced 4, 10, 20 and 22 failures depending on Node version and coverage, with zero assertion failures throughout. CI is the discriminator.

Nice work on the original — the bfcache diagnosis and the StrictMode ref-vs-flag reasoning in round 1 were both correct and both survived into this revision verbatim.

Issue #52, opened 2026-06-12
Create a separate category for resumelint on Posthog


sriyau64:
## Goal

Create a dedicated PostHog account and project for ResumeLint, separate from the maintainer's other product, and wire its API key into the **OSS GitHub Pages** build so resumelint analytics route to their own project rather than a shared one.

## Current state

Analytics in `src/lib/analytics.ts` already gate on `VITE_POSTHOG_KEY` at build time. If the key is unset, `ANALYTICS_ENABLED` is `false` and the posthog-js chunk is dropped from the dist entirely. **No code change is needed.** The existing ~10 instrumented events just need to be routed to the new ResumeLint PostHog project.

The OSS GitHub Pages build reads its env vars from repo secrets:

| Channel | File / mechanism |
|---|---|
| OSS GitHub Pages build | Repo secrets `VITE_POSTHOG_KEY` / `VITE_POSTHOG_HOST` in `.github/workflows/deploy-pages.yml` lines 30–31 |
| Local dev (optional) | `.env` (gitignored) |

## Tasks

- [x] Create a separate PostHog account for ResumeLint.
- [x] Create the `resumelint` PostHog project and obtain its `phc_` project API key.
- [x] Wire the key: set repo secrets `VITE_POSTHOG_KEY` and `VITE_POSTHOG_HOST` → consumed by `.github/workflows/deploy-pages.yml`.
- [x] Trigger a rebuild (`gh workflow run deploy-pages.yml`) so the secrets take effect.
- [x] Smoke-test: drop a PDF on the GitHub Pages build and confirm `file_accepted` / `parse_completed` land in the ResumeLint PostHog project — and nowhere else.

## Out of scope

- **A second, non-public embed surface** that also renders this checker. Wiring a resumelint PostHog key into that deploy touches credentials owned outside this issue and is deferred. This issue covers the OSS GitHub Pages surface only.
- No new event types or properties. Route existing instrumented events only.
- No changes to `src/lib/analytics.ts` or any other source file.

## Acceptance criteria

- Events from the OSS GitHub Pages build appear in the ResumeLint PostHog project and in no other account.
- The GitHub Pages env-var values live in repo secrets (not committed to the repo).

---

*Edited 2026-07-30: originally named a private, non-public product, one of its domains, and two of its deploy-script paths. Rewritten to describe the same work without them; the routing decision, the env-gating mechanism, and every task and acceptance criterion are unchanged. See #457.*

s-annam:
[...]

PR #786, opened 2026-08-09
feat(core): export the job-board providers on a ./job-search subpath

s-annam:
Adds a second published entry point, `@offlinecv/core/job-search`, carrying the job-board provider adapters — so a downstream consumer can call the feeds without putting a `fetch` on the `.` entry's value-edge closure.

## Why a subpath rather than a wider barrel

`packages/core/src/index.ts` makes a measured claim: its value-edge closure reaches no `fetch` / `WebSocket` / `XMLHttpRequest` / `EventSource`. That claim is load-bearing, not decorative — the private capture extension this package was cut for walks the import graph of each of its entry points through a single seam file and asserts exactly that, and four of its five surfaces must reach none. A content script that can reach `fetch` runs in `offlinecv.org`'s address space with a network primitive in scope.

Every provider adapter holds a `fetch(`. Putting them on `.` would not have widened a surface; it would have broken that gate. A second subpath lets both facts be true at once.

Measured on the emitted `dist/`:

| Entry | Modules | Network primitives | Externals |
|---|---|---|---|
| `.` | 27 | 0 | `idb` |
| `./job-search` | 11 | 7 modules hold `fetch(` | none |

Intersection **0**. Importing one cannot pull in the other, in either direction.

## Surface

Nine values — `getProviders`, `KEYLESS_PROVIDERS`, the three company-board factories, the Greenhouse/Lever hydrators, and the two helpers that recover the `jobId` those hydrators take — plus `JobProvider`, `JobPosting`, `JobQuery` as types.

`makeCompanyProvider` is deliberately absent: its `CompanyEntry` carries a `sectors` list that only means something to the in-app sector selector, so a caller holding a slug and a vendor would have to invent `sectors: []`. The three factories take what such a caller actually has.

The hydrators are deliberately present. Greenhouse has no server-side search, so `search()` fetches the light index and every posting comes back `description: ""`; a consumer that ranks on description without hydrating ranks every company posting against an empty string.

`greenhouseJobId` / `leverJobId` **move** from `company-boards.ts` into the adapter modules they describe, rather than being re-exported from where they lived. Re-exporting would have enlarged this entry's closure; relocating keeps them inside the closure it already reaches, so the surface grows and the closure does not. Their unit tests move with them, unchanged. `company-boards.test.ts` stubs those adapter modules, so it now spreads `importActual` and stubs only the fetching half — a wholesale mock would hand the pipeline `undefined` for the id and make every hydrate-call assertion meaningless.

## The gate grew rather than being worked around

`check:core` previously asserted the export *set* per entry. It could catch a provider symbol moved onto `.`; it could not catch a value *edge*, which is the cheaper mistake. It now BFSes each entry's value edges over the modules it already parses and asserts the per-entry module counts, that `.` reaches no network primitive, and that the closures are disjoint. The counts are named constants sitting beside the docblock claims they defend.

Three further holes closed in the same file:

- **`exports` ↔ expected-export map are cross-checked both ways.** Assertions had been generalized onto entries read off the manifest while the export-set assertion stayed keyed to a hand-written list, so a subpath added to one and not the other published an entirely unasserted surface with the gate green.
- **The unexported-subpath probe asserts its result count before iterating** — the same vacuity class this change had already fixed one instance of, one layer up.
- **The three "what a third entry point costs" checklists** disagreed with each other and all omitted this gate. They now name the same six places.

Docblock corrections the change made necessary: `.`'s network-free claim is scoped to that entry rather than left standing unqualified next to a fetching sibling, and `fetch-jd.js` moves off its "unreachable dead JS" list — the adapters take `htmlToPlaintext` from it, so it is genuinely loadable now.

`.fallowrc.jsonc` registers the new entry. fallow reported it as an unused file; swapping which entry `exports` names as `"."` changed nothing, so it keys on the filename `index.ts` rather than following `exports`. A specific-file allow-list entry, not a loosened pattern.

## Verification

`npm run verify` green. Beyond that, the assertions were **falsified rather than merely observed passing** — three times during authoring, three more during review, and the two that matter re-run independently before merge:

| Planted violation | Result |
|---|---|
| A `"./sneaky"` subpath in `exports`, absent from the expected-export map | ✗ fails, naming the subpath and listing all six places a new entry must be registered |
| `import "../../../src/lib/jd-match/fetch-jd.ts";` appended to the barrel — export set byte-identical | ✗ fails three ways: closure 29≠27, 1 network primitive in `.`, closures share 2 modules |
| A `"./*"` wildcard in `exports` | ✗ all three unexported-subpath assertions fire |
| Removing the `job-search` dts bundle from the rollup config | ✗ tarball-shape assertion + consumer typecheck (TS7016) |
| Removing the new entry from `.fallowrc.jsonc` | ✗ `fallow audit` reports it unused — confirming the registration suppresses a genuine false positive rather than a real finding |

The second row is the one to weigh: **before this PR that case passed silently**, leaving five public docblock claims false with the downstream extension's gate as the only thing that would have noticed.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation | Claude Opus 5 | ultra |
| Adversarial review | Claude Opus 5 | ultra |
| Review fixes | unreported (requested: opus) | ultra |
| Orchestration + PR | Claude Opus 5 | ultra |
| Review revisions | Claude Opus 5 (1M context) | unreported |

## Adversarial review

One round, pre-submit. **1 blocking, 7 nits — all fixed in this PR**, none deferred.

The reviewer independently re-derived every closure number using a different method from the author's (`es-module-lexer` BFS over the emitted `dist/` vs a TypeScript-compiler walk). All of 27 / 11 / disjoint / 38 / 62 / 24 / seven-fetching-modules reproduced exactly, as did the pre-existing "49 modules / ~17.5k LOC" claim.

- **[blocking]** `exports` and the hand-written expected-export map could drift, publishing an unasserted surface with the gate green. Reproduced with a planted subpath. → cross-check both ways.
- **[nit, fixed]** Nothing asserted the property the change exists to create; a side-effect import defeated the gate silently. → the closure assertion above. Ranked nit only because the claims were true at the time; it was the highest-value finding.
- **[nit, fixed]** A comment claimed three probe subpaths "name real shipped files"; one does not ship.
- **[nit, fixed]** The unexported-subpath loop passed vacuously if its input key was absent.
- **[nit, fixed]** Three disagreeing new-entry checklists, all omitting this gate.
- **[nit, fixed]** Half a two-part contract exported — hydrators without the id helpers.
- **[nit, fixed]** The hydrate paragraph never mentioned that `hydrateGreenhouse` rejects on non-ok while `hydrateLever` resolves `""`; a consumer writing `Promise.all(...)` loses a whole page to one 404.
- **[nit, fixed]** `tsconfig.build.json` said 24 modules are reached "only through `import type` edges"; two are there via a dynamic `import()` instead.

The reviewer also confirmed the rule-1 question this change raises: `providers/index.ts` is a barrel by filename only — no `export … from`, both symbols defined in the file, non-adapter imports all `import type` — so taking them from it drags nothing the three named adapter files would not have.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -9,6 +9,13 @@
  * depend on, so that consumer can write `@offlinecv/core` instead of reaching
  * into this repo's source tree by relative path.
  *
+ * This file is the **`.` entry point**, and the distinction now matters: the
+ * package has a second one, `@offlinecv/core/job-search` (`src/job-search.ts`),
+ * carrying the job-board provider adapters. Every claim below is about THIS
+ * entry unless it says otherwise, and the network-free one is about this entry
+ * *because* the other entry is not — see "Two entry points, and only one of them
+ * is network-free" below before quoting either.
+ *
  * The surface is deliberately narrow: the job-capture contract, the local
  * library's CRUD and replication primitives, the JD term/coverage pair, and the
  * fit-rating chain. It is what a *producer* of `JobRecord`s and a *replica* of
@@ -41,8 +48,9 @@
  * would drag a network primitive onto that graph for nothing. Measured: the
  * value-edge closure of the specifiers below is 27 modules and reaches no
  * `fetch`/`WebSocket`/`XMLHttpRequest`/`EventSource` at all. That is a statement
- * about the value-edge closure and not about the tarball's file list, which is
- * larger — see "The tarball ships more files than the graph reaches" below.
+ * about **this entry's** value-edge closure, and not about the tarball's file
+ * list, which is larger — see "The tarball ships more files than the graph
+ * reaches" below.
  *
  * **2. Types are re-exported with `export type`.** Not a style preference. A
  * type-only edge erases at build time, so `export type { JobRecord } from …`
@@ -52,35 +60,65 @@
  * `src/lib/job-search/types.ts` are re-exported for their types only and must
  * stay `export type` statements.
  *
- * The full transitive closure is 49 modules / ~17.5k LOC across `lib/storage`,
- * `lib/job-search`, `lib/jd-match`, `lib/heuristics`, `lib/score` and `webllm`,
- * with exactly one external runtime dependency: `idb`. The `heuristics`, `score`
- * and `webllm` modules are reached only through `import type` edges, so they
- * erase from the runtime graph — but not from the tarball.
+ * This entry's full transitive closure, `import type` edges included, is 49
+ * modules / ~17.5k LOC across `lib/storage`, `lib/job-search`, `lib/jd-match`,
+ * `lib/heuristics`, `lib/score` and `webllm`, with exactly one external runtime
+ * dependency: `idb`. The `heuristics`, `score` and `webllm` modules are reached
+ * only through `import type` edges, so they erase from the runtime graph — but
+ * not from the tarball.
+ *
+ * ## Two entry points, and only one of them is network-free
+ *
+ * `@offlinecv/core/job-search` is the second, and it is the deliberately
+ * network-BEARING half: the provider adapters call public job feeds, so every
+ * one of them holds a `fetch(`. It is a separate subpath precisely so that fact
+ * cannot arrive here by accident. Read `src/job-search.ts`'s own docblock for
+ * the full argument; the two facts that belong on this side of the seam are:
+ *
+ *   - **The two runtime closures are disjoint.** Measured: 27 modules from this
+ *     entry, 11 from `./job-search`, zero modules in common. Importing one
+ *     cannot pull the other in, in either direction, which is what makes the
+ *     network-free claim above survive the subpath's existence rather than merely
+ *     coexist with it.
+ *   - **The separation is a consumer's gate, not a preference.** The extension
+ *     this package was cut for walks the import graph of each of its entry
+ *     points and asserts the network primitives it reaches are exactly an
+ *     allow-list of one. Four of its five surfaces must reach none. A consumer
+ *     doing that audit must keep `./job-search` off every graph that has to stay
+ *     network-free, and there is no safe subset of it to reach for instead.
  *
  * ## The tarball ships more files than the graph reaches
  *
  * `tsc` emits a `.js` for every file in the program, the ones reached only by
  * `import type` included, and `files` ships all of them. So `npm pack` produces
- * 49 modules of which 27 are reachable from this barrel — and three of the
- * unreachable 22 read exactly like the thing this file says is absent:
+ * 62 modules, of which 38 are reachable — 27 from this barrel and 11 from
+ * `./job-search` — and two of the unreachable 24 read exactly like the thing
+ * this file says is absent:
  *
- *   - `dist/src/lib/jd-match/fetch-jd.js` — a live `fetch(url, …)`
  *   - `dist/src/lib/analytics.js` — `import.meta.env`, `await import("posthog-js")`
  *   - `dist/src/lib/webllm/web-llm.js` — `await import("@mlc-ai/web-llm")`
  *
- * Neither dynamic import is in `dependencies`, and nothing can load either file:
- * `exports` declares no subpath, so a consumer reaching for one gets
- * `ERR_PACKAGE_PATH_NOT_EXPORTED` (verified), and no reachable module imports
- * them. Unreachable dead JS a reader can open — not an egress path.
+ * (`dist/src/lib/jd-match/fetch-jd.js` used to head that list. It is now
+ * genuinely loadable — the provider adapters take `htmlToPlaintext` from it — so
+ * it moved off the dead-JS list and onto `./job-search`'s closure, where its
+ * `fetch(` is neither a surprise nor a capability that closure lacked.)
  *
- * The network-free claim above is therefore about the RUNTIME GRAPH rather than
- * the file list, and on the graph it holds exactly: 27 modules, no network
- * primitive, one bare import (`idb`). Which means grepping the tarball for
- * `fetch(` is the wrong audit — it finds `fetch-jd.js` plus two prose mentions
- * in comments. Walk the value edges out of `dist/packages/core/src/index.js`
- * instead; that is the graph a consumer actually loads, and what `check:core`'s
- * probe exercises when it imports the package by specifier.
+ * Neither dynamic import is in `dependencies`, and nothing can load either file:
+ * `exports` declares two subpaths and no wildcard, so a consumer reaching for
+ * any other path — including a `dist/…` one — gets
+ * `ERR_PACKAGE_PATH_NOT_EXPORTED` (verified, and asserted by `check:core`), and
+ * no reachable module imports them. Unreachable dead JS a reader can open — not
+ * an egress path.
+ *
+ * The network-free claim above is therefore about **this entry's** RUNTIME GRAPH
+ * rather than about the file list, and on that graph it holds exactly: 27
+ * modules, no network primitive, one bare import (`idb`). Which means grepping
+ * the tarball for `fetch(` is the wrong audit twice over — it finds the seven
+ * modules that legitimately fetch on the OTHER entry, plus prose mentions in
+ * comments, and says nothing about which entry a consumer imported. Walk the
+ * value edges out of `dist/packages/core/src/index.js` instead; that is the
+ * graph a consumer of `.` actually loads, and what `check:core`'s probe
+ * exercises when it imports the package by specifier.
  *
  * `idb` **is** in this package's `dependencies`, and #772 is the reason it now
  * is. While this was a bare re-export barrel the omission was right: nothing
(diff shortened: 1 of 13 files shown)


packages/core/src/job-search.ts
```diff
@@ -0,0 +1,196 @@
...
+ * where `fetch` is un-banned) and nothing else.
+ *
+ * Two mechanical consequences for that consumer, both worth knowing before the
+ * first import rather than after the gate goes red:
+ *
+ *  - Its walker resolves `@offlinecv/core` by reading the link target's
+ *    `exports["."]` entry. A **subpath** specifier is a different key, so the
+ *    walker needs to resolve `@offlinecv/core/job-search` too or it will
+ *    classify it as an ordinary npm package and stop — dropping this whole
+ *    closure out of the graph it was added to audit.
```
s-annam: **The mechanism described here does not exist at the consumer.**

`extension/src/__test-utils__/import-graph.ts` (HEAD `b81407d`) does this with every bare specifier:

```ts
if (!specifier.startsWith(".")) {
  externals.add(specifier);
  continue;
}
```

No `exports` reading, no `node_modules` resolution — the only bare-specifier special cases in the walker are `@design-tokens` and `@design-system`. And the extension does not reach this package by specifier at all today: `src/offlinecv-core.ts` imports the public repo by relative path (`../../src/lib/...`), which is why `@offlinecv/core` appears nowhere in that repo.

So the premise is false, and the consequence understates the work. `@offlinecv/core` and `@offlinecv/core/job-search` are treated **identically** today — both land in `externals`, both drop their closure — so the subpath is not the special hazard "a subpath specifier is a different key" implies. A consumer adopting it has to:

1. teach the walker `exports` resolution at all, for both keys; and
2. update `expect(graph.externals).toEqual([...WORKER_EXTERNALS].sort())` — whole-list equality, so it fails the moment any new bare specifier appears.

Not blocking: no code behaviour depends on it, and the consumer sits behind a submodule pin so nothing in this repo's CI can hold the claim honest. But this is a load-bearing paragraph in a public repo telling a future reader what the audit costs, and it is checkable and wrong.

Suggested rewrite — state it as the future requirement it is:

> Its walker follows relative specifiers only, filing every bare one into an `externals` list it asserts by whole-list equality; it does not resolve `@offlinecv/core` through `exports` today, because the extension reaches this repo by relative path. A consumer that switches to the package specifier has to teach the walker `exports` resolution for **both** keys — a subpath is a different key from `.` — or this whole closure drops out of the graph it was added to audit, and has to add the new specifier to the externals list.

**Caveat on my evidence:** I read the checkout at `extension/`, a separate private repo that is not part of this PR. If walker `exports` resolution exists on an unmerged branch there, the fix is to say so in the future tense (naming the branch) rather than leave the present tense standing.

packages/core/src/index.ts
```diff
@@ -119,7 +157,28 @@
  * The tarball ships no `.ts`, so a per-file declaration tree would dangle on
  * every internal import and a consumer would silently get `any`.
  * `rollup.dts.config.mjs` bundles the declarations into one specifier-free
- * `dist/index.d.ts`, and `files` ships that instead of the tree.
+ * file per entry point — `dist/index.d.ts` and `dist/job-search.d.ts` — and
+ * `files` ships those instead of the tree.
+ *
+ * ## What a third entry point costs
+ *
+ * Six edits, in six files, that no single tool checks together — the same list
```
s-annam: Nit: six edits, in **five** files — `packages/core/package.json` is items 1 and 2 of the list directly below.

The same wording is repeated verbatim at `packages/core/tsconfig.build.json:86` and `.fallowrc.jsonc:38`, so all three want the same edit. Flagging it only because this paragraph's whole job is to make three previously-drifting lists agree on a countable fact — the count is the load-bearing part. "Six edits, across five files" fixes it everywhere.

s-annam (commented):
Reviewed the diff against the offlinecv gates first and read the description last. **No blocking findings — approving.** The two-entry split is real, and the gate defending it is the strongest part of the change, so I falsified its headline assertion myself rather than taking the verification table on faith.

## What I re-ran independently

| Check | Result |
|---|---|
| `npm run check:core` | ✓ `. 27 modules/0 fetching, ./job-search 11 modules/7 fetching`, disjoint |
| Planted `import "../../../src/lib/jd-match/fetch-jd.ts";` in `src/index.ts` | ✗ fails three ways — closure `29≠27`, 1 network primitive on `.` (`fetch-jd.js`), closures share 2 modules. The headline row reproduces exactly. |
| Removed `packages/core/src/job-search.ts` from `.fallowrc.jsonc` → `entry` | ✗ `fallow audit` → `dead code: 1 issue · packages/core/src/job-search.ts`. The registration suppresses a real false positive, not a finding. |
| My own BFS over the emitted `dist/` from `job-search.js` | 11 modules, **0 externals** — matches the PR table and the `ENTRY_CLOSURES` row |
| `find packages/core/dist -name '*.js' \| wc -l` | 62 — the `62 / 38 reachable / 24 unreachable` arithmetic in `src/index.ts` holds |
| `providers/index.ts` is "a barrel by filename only" | ✓ no `export … from`; defines both symbols; the two non-adapter imports are `import type` |
| `grep -rn 'greenhouseJobId\|leverJobId' src packages scripts extension` | no stale importer left by the move out of `company-boards.ts` |
| `typecheck` / `lint` / `vitest run src/lib/job-search/` | green — 583 passed, 2 skipped |
| `fallow audit --base origin/main` | dead code 0 · complexity 0 · duplication 6 (report-only) |

Gates skipped as not applicable, not as passed: **fixture PII** (no fixture touched) and **design-system tier / style tokens** (nothing under `src/components/**`, no JSX in the diff).

## Secondary

**1. `packages/core/src/job-search.ts:53` — the consumer-walker mechanism this paragraph describes does not exist.** See the inline comment.

**2. `scripts/check-core-package.mjs:265` — `THIRD_ENTRY_CHECKLIST` omits `CONSUMER_TS`, and the success line overclaims because of it.**

Item 5 names `EXPECTED_EXPORTS`, `ENTRY_CLOSURES` and a `PROBE` static import. It does not name `CONSUMER_TS` (line 353), which is a hardcoded string importing exactly two specifiers. Those other three all fail loudly on a forgotten third entry — assertion 7 cross-checks the two tables against `exports`, and `PROBE` throws by name for an unmapped subpath. `CONSUMER_TS` is the one with no cross-check at all.

The concrete misbehaviour: add a third entry whose declaration bundle is emitted but broken under `nodenext` — an unresolvable inline `import("…")` type, a shape assertion 4's `from`-anchored regex also doesn't catch. Assertion 3 never imports it. The gate stays green, and line 1028 interpolates `declarationEntries.join(" + ")` off the *manifest*, so it prints `a nodenext consumer typechecks against dist/index.d.ts + dist/job-search.d.ts + dist/third.d.ts` — a claim nothing made true.

That is the same vacuity class assertion 7 was written to close, one table over, and the whole point of centralising the checklist here was that its copies had drifted. Fix is one clause:

```
-  "scripts/check-core-package.mjs → EXPECTED_EXPORTS, ENTRY_CLOSURES, and a PROBE static import",
+  "scripts/check-core-package.mjs → EXPECTED_EXPORTS, ENTRY_CLOSURES, a PROBE static import, and a CONSUMER_TS import (nothing cross-checks that one)",
```

Not blocking: it cannot fire until a third entry exists, and this PR ships two. Couldn't be anchored inline — GitHub returns no `patch` for this file (638 changed lines exceeds its per-file cap), so no RIGHT-side line was derivable.

## Nits (non-blocking)

**1. "Six edits, in six files" is six edits in _five_ files** — `packages/core/package.json` is items 1 and 2. See the inline comment; the same wording is repeated at `packages/core/tsconfig.build.json:86` and `.fallowrc.jsonc:38`.

**2. fallow duplication — 6 clone groups, 165 lines (0.1%).** All pre-existing: provider-test scaffolding and the `mapJob` shape shared by `ashby`/`greenhouse`/`lever`. Report-only inside `verify`, and re-attributed here only because the diff touches those files. Recorded so it's clear it was looked at, not that it needs action in this PR.

## Description accuracy

Accurate — no overclaim, and nothing the diff does went unmentioned. Every checkable statement round-tripped to code: the closure table, `62`/`38`/`24`, disjointness, the nine-value surface (exactly `EXPECTED_EXPORTS["./job-search"]`), `makeCompanyProvider`'s deliberate absence, the `importActual` rationale in `company-boards.test.ts`, and the two falsification rows I re-ran above.

No issue is linked, so there were no acceptance criteria to check — reviewed against the diff alone.

Verdict rule applied: 0 Blocking → **APPROVE**. Posted as a `COMMENT` review only because GitHub refuses `Can not approve your own pull request` — the review token and the PR author are the same account. Treat this as an approval; registering it as one needs a second account.

---

packages/core/src/job-search.ts
s-annam: Confirmed and fixed in `d5b39bb`. I re-read the walker at the same HEAD (`b81407d`) and it's exactly as you describe: `if (!specifier.startsWith("."))` files every bare specifier into `externals` and `continue`s, the only bare-specifier special cases are `@design-tokens` and `@design-system`, and `@offlinecv/core` appears nowhere in that repo's `src/` — `src/offlinecv-core.ts` reaches this repo by relative path. `no-network.test.ts:241` is the whole-list assertion (`expect(graph.externals).toEqual([...WORKER_EXTERNALS].sort())`).

On your caveat: I checked, and there is no unmerged branch carrying `exports` resolution — the extension repo has `main` and `gh-52-job-watch-and-letter-sync`, and no revision reachable from any ref resolves `exports` in `import-graph.ts`. So the present tense was simply wrong, not early; the rewrite is future-tense throughout.

Took your suggested wording, adapted to the bullet it replaces and naming the seam file:

> - Its walker follows relative specifiers only, filing every bare one into an `externals` list it asserts by whole-list equality; it does not resolve `@offlinecv/core` through `exports` today, because the extension reaches this repo by relative path through its `src/offlinecv-core.ts` seam. A consumer that switches to the package specifier has to teach the walker `exports` resolution for **both** keys — a subpath is a different key from `.` — or this whole closure drops out of the graph it was added to audit, and has to add the new specifier to the externals list.

The second bullet (allow-list compared for equality) is unchanged — that one round-trips.

packages/core/src/index.ts
s-annam: Right — items 1 and 2 are both `packages/core/package.json`, so it's six edits across five files. Fixed in `d5b39bb`, in all three places, using your wording:

- `packages/core/src/index.ts:165` — "Six edits, across five files, that no single tool checks together"
- `packages/core/tsconfig.build.json:86` — "costs six edits across five files"
- `.fallowrc.jsonc:38` — "costs six edits across five files"

`THIRD_ENTRY_CHECKLIST` itself needed no change: it's a six-item list and states no file count, so it was never the copy that drifted. The commit message's own "now name the same six places" line was the same ambiguity and is now "the same six edits, across five files — `package.json` carries two of the six."

s-annam:
**One change in `d5b39bb` that no thread asked for**, flagging it rather than slipping it in.

`scripts/check-core-package.mjs:889` joined its comparison arrays on a **literal NUL byte** in the source, rather than on the two-character `` escape — i.e. two raw `0x00` bytes sitting in the file, at offset 46379. `main` has zero of them, so this PR introduced them.

Runtime behaviour is fine: a JS string holds U+0000 happily, and `check:core` was green before and after. But the byte makes the whole 1054-line file read as **binary** to `grep`, `rg` and `git grep`, so every search over it returns nothing — silently, with exit status 0 on `rg` and a `binary file matches` line at best. I found it by grepping for `EXPECTED_EXPORTS` in the file that defines it, getting no output, and briefly concluding the `index.ts` docblock was naming symbols that don't exist.

`git diff` still renders the file as text — git's binary heuristic only reads the first 8000 bytes and the NUL is well past that — which is presumably why it survived review.

The fix is to write the escape instead of the byte. Identical string at runtime, greppable file. `check:core` still passes and exercises that exact line: it is the unexported-subpath refusal comparison, the "3 unexported subpath(s) refused" clause of the success output.

Happy to pull it into its own PR if you would rather keep this one scoped to the two threads.

s-annam:
Staying in this PR, per author. No further push needed — it is already in `d5b39bb` (`scripts/check-core-package.mjs:889`), and the commit message carries a paragraph explaining it, so the reason survives the squash into `main`.

PR #795, opened 2026-08-11
fix(contact): reject unclassifiable '+ Add a profile' input instead of dropping it (#790)

### LINKED ISSUE #790: "Add a profile" silently discards unclassifiable input — and accepts a bare word as a link
## Problem

The contact card's **"+ Add a profile"** affordance silently discards input it cannot classify. A user types a value, presses commit, the input clears and the pill collapses — and nothing was added. No error, no inline hint, no trace anywhere in the UI. The user's next signal that the edit was lost is the downloaded PDF, several steps later.

Reported from real use: a user tried to add `US Citizen` as a profile entry (wanting a work-authorization line in the header) and found it "got dropped when a PDF is generated". The PDF is not where it was dropped — the value never entered edit state at all.

The same validation gap fails open in the other direction: a value that is *not* a URL but happens to contain no whitespace is **accepted**, stored as a profile, and rendered into the exported PDF's contact line as a bogus link slug.

### Verified behavior

`classifyProfile` (`src/lib/contact/profile-registry.ts:140`), run against the real module:

```
classifyProfile("US Citizen")  -> undefined
classifyProfile("US_Citizen")  -> {"url":"https://US_Citizen","network":"us_citizen","kind":"other"}
classifyProfile("UScitizen")   -> {"url":"https://UScitizen","network":"uscitizen","kind":"other"}
classifyProfile("Citizen")     -> {"url":"https://Citizen","network":"citizen","kind":"other"}
```

The space is the whole difference. `normalizeUrl` prepends `https://`, then `new URL("https://US Citizen")` throws on the space and the `catch` at `profile-registry.ts:147-149` returns `undefined`. Without the space, `new URL("https://Citizen")` parses cleanly, `hostname` is `citizen`, no `PROFILE_HOSTS` rule matches, and the documented fallback at `profile-registry.ts:162` — "An UNKNOWN host is kept, never dropped" — keeps it as `{ kind: "other" }`. That fallback is correct for a real unknown host (`orcid.org`, a personal domain); it is wrong for a bare word that is not a host at all.

## Root cause — the return value is dropped at two prop boundaries

`addProfile` already reports failure. Nothing reads it.

1. **`src/hooks/useEditableParse.ts:1161-1167`** — the hook returns `string | undefined`, `undefined` meaning "not added":

   ```ts
   const addProfile = useCallback((url: string): string | undefined => {
     const profile = classifyProfile(url);
     if (!profile) return undefined;          // ← the only failure signal
     const id = `profile:${idCounter.current++}`;
     setProfileOverrides((prev) => [...prev, { id, ...profile }]);
     return id;
   }, []);
   ```

2. **`src/components/features/ContactExtraLinks.tsx:33`** — the prop type erases it: `onAdd: (url: string) => void;`

3. **`src/components/features/ProfileLinkAdd.tsx:36, 58-65`** — same erasure (`onAdd: (url: string) => void`), and `commit()` unconditionally treats the call as success:

   ```ts
   const commit = () => {
     const trimmed = draft.trim();
     if (!trimmed || trimmed === "https://") return;
     onAdd(trimmed);        // ← return value discarded
     setDraft("");          // ← draft destroyed even on failure
     if (!stayOpenAfterAdd) setExpanded(false);
   };
   ```

`setDraft("")` is what makes this data loss rather than a no-op: the typed text is gone, so the user cannot even see what they tried to add.

Both add points are affected — the extra-links row (`ContactExtraLinks.tsx:75`) and the required-slot add in `ContactDetails.tsx:135-138`, since both consume the same `ProfileLinkAdd`.

## Scope

Two defects, one validation seam. Fix both — fixing only the rejection path leaves `https://Citizen` shipping into users' PDFs.

### A. Rejected input must not vanish

- Thread `addProfile`'s `string | undefined` through the prop chain: `ProfileLinkAdd`'s `onAdd` and `ContactExtraLinks`' `onAdd` become `(url: string) => string | undefined`.
- In `ProfileLinkAdd.commit()`, keep the draft and surface an inline error when the return is `undefined`. Do not collapse the pill.
- Error copy must say what is accepted, not just that it failed. Suggested: `That doesn't look like a link. Paste a profile URL (e.g. linkedin.com/in/yourname).`
- The error clears on the next keystroke.
- Announce it to assistive tech: `role="alert"` (or `aria-live="assertive"`) plus `aria-invalid` on the input. `ReconstructedSkills.tsx:248-250` is the in-repo precedent for a live region.

### B. Accepted input must actually be a link

`classifyProfile` currently accepts any whitespace-free token as an "unknown host". Tighten the fallback so a bare word cannot become a profile, while keeping genuine unknown hosts working.

The minimum discriminator is a dot in the hostname — every real public host has one. Reject a dotless hostname unless it is `localhost`:

```ts
// after: hostname = new URL(url).hostname...
if (hostname.length === 0) return undefined;
// A bare word is not a host. Every real public host is dotted; a dotless
// hostname here means the user typed prose ("Citizen"), not a URL.
if (!hostname.includes(".") && hostname !== "localhost") return undefined;
```

This must not regress the documented unknown-host behavior: `orcid.org`, `scholar.google.com`, and a personal domain like `example.com` all stay `kind: "other"`, because all are dotted. Add a test asserting that.

Because `classifyProfile` is also on the **parse** path via `profilesFromUrls` (`profile-registry.ts:174`), which is fed the four legacy link keys, confirm no corpus snapshot moves — the legacy keys only ever hold extracted URLs, which are dotted. If a snapshot does move, stop and investigate rather than re-baking.

### Not in scope

The user's underlying need — a **work-authorization / citizenship field** (`US Citizen`, `Authorized to work in the US without sponsorship`) — is a separate feature. `ResumeData` has no such field, so there is nowhere for it to go today. This issue only stops the edit lane from losing input silently. File the field separately.

## Reuse analysis

**Capability:** surface a validation error inside an existing inline-add control.

**Existing surfaces found:**
- `src/design-system/primitives/` — no error-message primitive; `ErrorState` (`src/design-system/shared/`) is a full-block empty/error state, wrong shape for an inline field hint.
- `src/components/features/ProfileLinkAdd.tsx` — already owns the add-a-profile interaction, already renders its own helper line (`HELPER_EXAMPLES`, line 45).

**Decision:** **extend `ProfileLinkAdd`.** The error is one line of text rendered next to the helper line it already owns, styled with existing semantic tokens (`text-status-error` or the closest existing token — check `src/design-system/styles/theme.css`, do not introduce a hex or a raw palette class). No new component, no new primitive.

## Implementation plan

1. **`src/lib/contact/profile-registry.ts`** — add the dotless-hostname guard in `classifyProfile` after the existing `hostname.length === 0` check. Update the function's docblock: the "unknown host is kept, never dropped" contract now reads "unknown *host* is kept; a dotless bare word is not a host and is rejected".
2. **`src/lib/contact/profile-registry.test.ts`** — assert `classifyProfile` returns `undefined` for `"US Citizen"`, `"Citizen"`, `"US_Citizen"`, `"UScitizen"`; assert it still returns `kind: "other"` for `"orcid.org/0000-0002-1825-0097"`, `"example.com"`, and a scheme-less `"scholar.google.com/citations?user=x"`.
3. **`src/components/features/ProfileLinkAdd.tsx`** — change `onAdd` to `(url: string) => string | undefined`; add `const [error, setError] = useState<string | null>(null)`; in `commit()`, set the error and return early (keeping `draft` and `expanded`) when `onAdd` returns `undefined`; clear the error in the `onChange` handler; render the message with `role="alert"` and set `aria-invalid` on the input.
4. **`src/components/features/ContactExtraLinks.tsx`** — widen the `onAdd` prop type to match and pass through unchanged.
5. **`src/components/features/ContactDetails.tsx`** — widen `onAddProfile` (line 86) to match; `ContactCard.tsx:58` likewise. `ReconstructedResume.tsx:1320` already passes `addProfile`, whose signature already returns `string | undefined` — no change needed there.
6. **`src/components/features/ProfileLinkAdd.test.tsx`** — add: rejected input keeps the draft text, keeps the panel expanded, and renders an alert; a subsequent keystroke clears the alert; accepted input still clears the draft and (for `stayOpenAfterAdd = false`) collapses.

## Acceptance criteria

- [ ] Typing `US Citizen` into "+ Add a profile" and committing leaves the text in the input, keeps the panel open, and renders a visible error naming what is accepted.
- [ ] The error is exposed to assistive tech (`role="alert"` / live region) and the input carries `aria-invalid`.
- [ ] The error clears on the next keystroke.
- [ ] Committing `Citizen` (no space) is **rejected** — it does not become a profile, and no `https://Citizen` slug reaches the exported PDF's contact line.
- [ ] `classifyProfile` still returns `kind: "other"` for dotted unknown hosts (`orcid.org`, `example.com`, `scholar.google.com`) — regression test present.
- [ ] A valid URL still adds exactly as before: draft clears, chip appears, `stayOpenAfterAdd` behavior unchanged.
- [ ] Both add points behave identically — the extra-links row and the required-slot add in `ContactDetails`.
- [ ] No corpus snapshot changes (`npx vitest run src/lib/heuristics`); round-trip invariants unchanged (`corpus-roundtrip.test.ts`, `render-roundtrip.repro.test.ts`).
- [ ] No raw `<button>`, no hardcoded hex or raw Tailwind palette class in the new markup.

s-annam:
Layer 01 of 3. Base for the stack — depends on nothing but `main`.

## What shipped

The contact card's **"+ Add a profile"** silently discarded input it could not classify, and accepted a bare word as a link. Both directions are closed.

**Rejected input no longer vanishes.** `addProfile` already returned `string | undefined` to report failure; nothing read it. The signal is now threaded through the prop chain (`ProfileLinkAdd` → `ContactExtraLinks` → `ContactDetails` → `ContactCard`), and `ProfileLinkAdd.commit()` keeps the draft, keeps the panel open, and renders an inline error naming what is accepted. The error carries `role="alert"`, the input carries `aria-invalid`, and the message clears on the next keystroke.

**Accepted input must actually be a link.** `classifyProfile` accepted any whitespace-free token as an "unknown host", so `Citizen` became `https://Citizen` and shipped into the exported PDF's contact line as a bogus slug. A dotless hostname is now rejected unless it is `localhost`. Genuine unknown hosts are unaffected — `orcid.org`, `example.com`, `scholar.google.com` all still classify as `kind: "other"`, with a regression test pinning that.

## One deviation from the issue's plan, worth a reviewer's eye

The plan assumed the required-slot add point in `ContactDetails.tsx` needed only a type widening. It did not. That add point wires `onAdd={commitLink}`, and `commitLink` is `void`-typed — so it always returns `undefined`, which would have shown the rejection error on **every** commit including valid ones. It also has its own permissive correction-path fallback that would have let `Citizen`-style text through, defeating the issue's "both add points behave identically" criterion.

Fixed by running `classifyProfile` in `ContactDetails.tsx` before delegating to `onLegacyLinkChange`, scoped to the absent/first-add branch only. The low-confidence correction-path fallback in `useEditableParse.ts` is intentional and untouched.

## Verification

- `classifyProfile` is on the **parse** path via `profilesFromUrls`, so the corpus was the risk. `src/lib/heuristics` — 1271 passed, **no snapshot moved**.
- Scoped suites (`profile-registry`, `ProfileLinkAdd`, `ContactCard`) — 44 passed.
- `typecheck`, `lint` clean.
- Gates were run under Node 20 (`.nvmrc`); this repo's suite produces spurious `localStorage` failures on newer Node.

Closes #790

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #790 | Claude Sonnet 5 | high |
| Adversarial review (stack-wide) | Claude Opus 4.5 | high |

## Adversarial review

Reviewed pre-submit as part of the 3-layer stack, two rounds, by an independent reviewer prompted to break the change rather than confirm it.

**No blocking findings against this layer.** The reviewer specifically verified that this layer composes with layer 3, which edits `ContactDetails.tsx` again: layer 3's `segments` filter only touches `group === "contact"` rows, while this layer's `addLink` lives in `renderLink`, which is only ever called on `group === "link"` rows. The required-slot add path is untouched, and `addLink` correctly returns a non-empty value so `ProfileLinkAdd` reads it as success.

Stack-wide outcome: two blocking findings were found and fixed on layers 2 and 3 before any PR was opened; see those PRs. This layer was clean in both rounds.

### DIFF
--- a/src/components/features/ContactCard.test.tsx
+++ b/src/components/features/ContactCard.test.tsx
@@ -140,7 +140,7 @@ describe("ContactCard", () => {
           kind: "other",
         },
       ],
-      onAddProfile: () => {},
+      onAddProfile: () => undefined,
       onEditProfile: () => {},
       onRemoveProfile: () => {},
     });
--- a/src/components/features/ContactCard.tsx
+++ b/src/components/features/ContactCard.tsx
@@ -55,7 +55,7 @@ interface ContactCardProps {
    *  together with the add/edit/remove handlers to enable the variable-length
    *  links affordance in the editable card. */
   extraProfiles?: readonly ProfileOverride[];
-  onAddProfile?: (url: string) => void;
+  onAddProfile?: (url: string) => string | undefined;
   onEditProfile?: (id: string, url: string) => void;
   onRemoveProfile?: (id: string) => void;
 }
--- a/src/components/features/ContactDetails.tsx
+++ b/src/components/features/ContactDetails.tsx
@@ -20,6 +20,7 @@
 
 import { formatLinkDisplay, type ContactDisplayField } from "../../lib/contact.ts";
 import { EditableField } from "@design-system";
+import { classifyProfile } from "../../lib/contact/profile-registry.ts";
 import {
   validateEmail,
   validatePhone,
@@ -83,7 +84,7 @@ interface ContactDetailsProps {
    *  `onAddProfile` is provided (editable card), the variable-length add/edit/
    *  delete affordance renders below the legacy links line. */
   extraProfiles?: readonly ProfileOverride[];
-  onAddProfile?: (url: string) => void;
+  onAddProfile?: (url: string) => string | undefined;
   onEditProfile?: (id: string, url: string) => void;
   onRemoveProfile?: (id: string) => void;
 }
@@ -288,10 +289,21 @@ function renderLink(
     // the guided network picker instead of a bare URL field, so a naive user
     // learns what counts and which networks are accepted (#335-followup).
     if (field.reason === "absent") {
+      // `onLegacyLinkChange` also serves the low-confidence CORRECTION path
+      // below, where an unparseable edit is intentionally kept (raw value,
+      // `kind: "other"`) so the correction isn't lost. That fallback is wrong
+      // for a first-time ADD: classify here so an unclassifiable value (e.g.
+      // "US Citizen") is rejected the same way `addProfile` rejects it,
+      // instead of being written into the slot as a bogus link (#790).
+      const addLink = (v: string): string | undefined => {
+        if (classifyProfile(v) === undefined) return undefined;
+        commitLink(v);
+        return v;
+      };
       return (
         <ProfileLinkAdd
           label={`Add a ${field.label.toLowerCase()}`}
-          onAdd={commitLink}
+          onAdd={addLink}
         />
       );
     }
--- a/src/components/features/ContactExtraLinks.tsx
+++ b/src/components/features/ContactExtraLinks.tsx
@@ -30,7 +30,7 @@ import { ProfileLinkAdd } from "./ProfileLinkAdd.tsx";
 
 interface ContactExtraLinksProps {
   profiles: readonly ProfileOverride[];
-  onAdd: (url: string) => void;
+  onAdd: (url: string) => string | undefined;
   onEdit: (id: string, url: string) => void;
   onRemove: (id: string) => void;
 }
--- a/src/components/features/ProfileLinkAdd.test.tsx
+++ b/src/components/features/ProfileLinkAdd.test.tsx
@@ -49,7 +49,7 @@ afterEach(() => {
 
 describe("ProfileLinkAdd", () => {
   it("starts collapsed as a labelled add pill", () => {
-    const el = render({ onAdd: () => {}, label: "Add a professional profile" });
+    const el = render({ onAdd: () => undefined, label: "Add a professional profile" });
     const pill = el.querySelector('[aria-label="Add a professional profile"]');
     expect(pill).not.toBeNull();
     // No network chips until it is expanded.
@@ -57,7 +57,7 @@ describe("ProfileLinkAdd", () => {
   });
 
   it("expands to show a chip per quick-pick network", () => {
-    const el = render({ onAdd: () => {}, label: "Add a profile" });
+    const el = render({ onAdd: () => undefined, label: "Add a profile" });
     click(el.querySelector('[aria-label="Add a profile"]'));
     for (const pick of PROFILE_QUICK_PICKS) {
       expect(
@@ -67,7 +67,7 @@ describe("ProfileLinkAdd", () => {
   });
 
   it("tapping a network chip pre-fills its prefix into the input", () => {
-    const el = render({ onAdd: () => {}, label: "Add a profile" });
+    const el = render({ onAdd: () => undefined, label: "Add a profile" });
     click(el.querySelector('[aria-label="Add a profile"]'));
     click(el.querySelector('[aria-label="Add GitHub"]'));
     const input = el.querySelector("input");
@@ -75,7 +75,7 @@ describe("ProfileLinkAdd", () => {
   });
 
   it("commits the entered URL and collapses when stayOpenAfterAdd is unset", () => {
-    const onAdd = vi.fn();
+    const onAdd = vi.fn(() => "profile:0");
     const el = render({ onAdd, label: "Add a professional profile" });
     click(el.querySelector('[aria-label="Add a professional profile"]'));
     click(el.querySelector('[aria-label="Add LinkedIn"]'));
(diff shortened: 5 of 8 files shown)


s-annam (commented):
### Stance
LGTM! This PR cleanly addresses issue #790 by rejecting unclassifiable input (such as dotless bare words like `US Citizen` or `Citizen`) at the `classifyProfile` layer while preserving valid dotted hosts, and threading the rejection signal through `ProfileLinkAdd` to prevent silent data loss.

- [x] Typing `US Citizen` into "+ Add a profile" keeps draft, keeps panel open, renders visible error naming accepted input.
- [x] Error exposed via `role="alert"` and input carries `aria-invalid`.
- [x] Error clears on next keystroke.
- [x] `Citizen` (bare word) is rejected, preventing `https://Citizen` bogus link.
- [x] `classifyProfile` preserves dotted unknown hosts (`orcid.org`, `example.com`, `scholar.google.com`).
- [x] Valid URLs add properly; draft clears and panel collapses as expected.
- [x] Both extra-links row and required-slot add in `ContactDetails` reject unclassifiable input consistently.
- [x] No corpus snapshot changes; round-trip invariants unchanged.
- [x] `npm run verify` passes cleanly (518 tests passed, 0 lint errors, 0 type errors, 0 fallow dead code).
- [x] No raw UI elements, hardcoded hex, or raw palette classes used (`text-feedback-error-text` semantic token used).

### OfflineCV Gate Results
- **3a Fixture PII**: Passed (`npm run check:fixtures` clean, 0 binary fixtures added/modified).
- **3b Design System & Reuse**: Passed (Extended `ProfileLinkAdd`, no raw `<button>` / `<input>` added in feature code).
- **3c Style Tokens**: Passed (`text-feedback-error-text` semantic token used).
- **3d Dead-code / Fallow**: Passed (0 fallow warnings).
- **3e Skill/Script Command Bugs**: N/A (no script or skill files modified).
- **3f PR Description Accuracy**: Accurate (All claims verified against diff).

### Verdict & Findings Summary
- **Verdict**: `APPROVE` (0 Blocking findings — submitted as COMMENT due to self-authored PR constraint)
- **Blocking**: 0
- **Secondary**: 0
- **Nits**: 0

---
Reviewed by: Gemini 3.6 Flash (high)

Issue #791, opened 2026-08-11
Skills: "+ Add category" unreachable on an uncategorised résumé, and ungrouped skills drop from the exported PDF


s-annam:
## Problem

#476 shipped full category editing for the Skills section — rename, delete, add, move, plus the drag-and-drop and keyboard affordances. All of it is unreachable on a résumé whose Skills section is a plain comma list, which is the common case.

The `+ Add category` control renders **only** when categories already exist. A user with a flat skills list has no way to create the first one, so none of the other four operations can ever be reached either. #476 called this out as a decision that had to be made explicitly — *"Decide explicitly whether 'add a category' is offered on such a résumé at all"* — and it was answered "no" by omission, which is exactly the invisible decision that issue warned against.

Reported from real use: a user with a 26-skill flat list wanted to group them (Leadership / Product & Delivery / Technical) and found no way to do it.

## Root cause

`src/components/features/ReconstructedSkills.tsx:178`:

```ts
const categorised = skillCategories && skillCategories.length > 0;
```

`src/components/features/ReconstructedSkills.tsx:210-247` — `<AddCategoryInput>` lives inside the truthy arm; the falsy arm gets a flat chip row and a flat add-skill input, and nothing else:

```tsx
      ) : categorised ? (
        <div className="flex flex-col gap-1.5">
          {rows.map(...)}
          {ungrouped.length > 0 && <SkillChipRow ... />}
          <AddCategoryInput onAdd={onAddCategory} />      {/* line 242 — categorised only */}
        </div>
      ) : (
        <SkillChipRow skills={skills} onRemoveSkill={onRemoveSkill} />
      )}
      {!categorised && <AddSkillInput skills={skills} onAdd={onAddSkill} />}
```

A parsed flat skills section leaves `skillCategories === undefined` (`src/lib/heuristics/types.ts:141-148` — absent means "the document did not say"), so `categorised` is `false` and line 242 never renders. Chicken-and-egg: creating a category requires having one.

Nothing below the UI is missing. The transform already handles the empty case correctly — `addCategory` (`src/lib/edit/skills-categories.ts:90-98`) appends to whatever array it is given, so `addCategory([], "Leadership")` returns `[{ label: "Leadership", skills: [] }]`. The hook is wired (`useEditableParse.ts:996-999`), and `ReconstructedResume.tsx:1421` already binds it. **This is a reachability gap, not a missing capability.**

## Second defect: ungrouped skills are dropped from the exported PDF

[...]

PR #797, opened 2026-08-11
feat(contact): add a work-authorization field across parse, edit, render and export (#792)

### LINKED ISSUE #792: Work authorization: no field to state it, no place to render it, and the parser drops it
## Problem

There is no way to state work authorization on a résumé in offlinecv. `ResumeData` (`src/lib/score/types.ts:303-347`) has no field for it, the contact card has no row for it, the exported PDF has no place to draw it, and the parser discards it when a source résumé already says it.

This matters for a specific, common failure mode: applicant screening frequently treats **absence of a work-authorization signal as ambiguity**, and ambiguity routes toward the "may require sponsorship" pile. A candidate who needs no sponsorship and does not say so is penalized for silence. The line is short — `US Citizen`, `Authorized to work in the US without sponsorship` — and every résumé template in the wild puts it either on the header contact line or in a trailing "Additional" block.

Today, a user who tries anyway loses the input. The only free-text-shaped add affordance on the contact card is "+ Add a profile", which is a URL field: `classifyProfile("US Citizen")` returns `undefined` and the value is silently discarded (that discard is #790; this issue is the missing destination it should have gone to).

A source résumé that already states it fares no better. The line is correctly kept out of `parsed.skills` — `src/lib/heuristics/extract/skills.test.ts:109` and `:162` pin exactly that — but it is not routed anywhere else either, so it is dropped from the reconstructed résumé and absent from the Download PDF. Parse → export → re-parse loses a line the source had.

### The field

Add to `ResumeData` (`src/lib/score/types.ts`), alongside `location`:

```ts
  /** Work-authorization / right-to-work statement as the résumé states it,
   *  verbatim: "US Citizen", "Green Card holder", "Authorized to work in the
   *  US without sponsorship", "EU passport holder". Free text, never an enum —
   *  see the issue's rationale. Optional everywhere: absence is not a gap and
   *  is never scored. */
  work_authorization?: string;
```

**Free text, not an enum.** Statuses are numerous, jurisdiction-specific, and legally consequential (`US Citizen`, `Permanent Resident`, `TN eligible`, `H-1B transfer`, `Authorized to work in the UK without sponsorship`). Classifying a candidate's self-declaration into a fixed set would be lossy in exactly the cases that matter, and would put us in the position of asserting someone's legal status. Carry what they wrote.

### Three guardrails that are the point of this issue, not decoration

**1. It must never be scored.** Do **not** add a check to `completenessChecks` in `src/lib/score/score.ts:1006+`, and do not add it as a required row in `CONTACT_ROWS`. The disclosure is optional, jurisdiction-specific, and protected-class-adjacent. A résumé that omits it must score identically to today — otherwise we penalize candidates for declining to disclose immigration status, which is not a defensible thing for this product to do. Assert this with a test: the same fixture with and without `work_authorization` produces an identical `AnonymousAtsScore`.

**2. It must never egress.** `src/lib/job-search/providers/keywords.ts` is the sole resume-derived egress helper (`CLAUDE.md` → Project overview), building the keyword string sent to third-party job feeds from the query title + skills. `work_authorization` must not reach it, directly or via any query-builder path. Add a test asserting the built query string never contains the field's value.

**3. It must not be treated as PII-free.** It is a personal legal attribute. Any fixture that carries it uses the synthetic-persona rules already in force (`CLAUDE.md` → Hard rules), and `scripts/check-fixture-pii.mjs` sweeps ground-truth sidecars — confirm a `work_authorization` value added to a `*.truth.json` does not trip or need to bypass that check.

### Where it renders

Appended to the contact line, after location, before links:

```
Chicago, IL · jane@example.com · US Citizen
```

Zero extra lines in the header — the whole reason this is viable on a dense résumé.

### JSON Resume export

JSON Resume v1.0.0 `basics` has no work-authorization property, and `src/lib/pdf/to-json-resume.ts` claims v1.0.0 conformance. Do not repurpose an existing standard field for it.

**Decision:** emit it as `basics.x_workAuthorization`, with a docblock stating it is a deliberate non-standard extension (the `x_` prefix is the conventional escape hatch) and that a consumer validating strict v1.0.0 will ignore it. Omit the key entirely when the field is absent, so existing exports are byte-identical. If a reviewer prefers omitting it from JSON export altogether, that is the one decision here worth overriding — say so on the issue rather than in review.

## Scope — two stages, both required

Splitting extraction out would ship the same silent-loss bug as #790/#791: a user could add the field, download, re-drop the PDF, and watch it vanish. Both stages land before this closes.

### Stage 1 — model, edit, render, export

1. **`src/lib/score/types.ts`** — add `work_authorization?: string` to `ResumeData` with the docblock above.
2. **`src/lib/contact.ts`** — add to `CONTACT_ROWS` (line ~48-63): `{ key: "work_authorization", label: "Work authorization", group: "contact", optional: true }`, and add the key to `FIELD_KEYS` (line 66-76). `optional: true` is what keeps absence out of the detected/total ratio — the same treatment `github_url` / `portfolio_url` / `website_url` already get.
3. **⚠️ Optional rows are hidden when absent, so add an explicit entry point.** `buildContactFields` skips an undetected optional row (`contact.ts:109`). Without a dedicated affordance the field is unreachable for exactly the users who need it — the identical reachability trap as #791. Add a "+ Add work authorization" control to the editable contact card, alongside the existing add affordances in `ContactDetails.tsx` / `ContactExtraLinks.tsx`. Plain text input — **not** `ProfileLinkAdd`, which is a URL field.
4. **`src/hooks/useEditableParse.ts`** — the field rides the existing `ContactOverrides` channel (`setContactField`); confirm the key is accepted there and cleared by the reset path.
5. **`src/lib/pdf/ats-resume-model.ts`** — read it in `buildContact` (line 256+) through `valueFor`, exactly like `location` (line 275), onto `AtsContact`.
6. **`src/lib/pdf/render-ats-pdf.ts`** — add it to `contactParts` (lines 1496-1501) after `location`, before `...links`. It is **not** a link: it must receive no `mailto:`/`https:` overlay from the per-substring annotation pass (lines 1503-1515), and no scheme-stripping. Add it to the loss audit's `check(...)` list (lines 505-511) so a dropped value is caught.
7. **`src/lib/pdf/to-json-resume.ts`** — map to `basics.x_workAuthorization` in `basicsFromContact` (line 295), omitted when empty.

### Stage 2 — extraction, so a source résumé keeps it

Two placements, both tight. The risk to manage is over-matching, not under-matching: a false positive puts a fabricated legal claim on someone's résumé, which is far worse than missing a true one. Prefer a narrow pattern and a `KNOWN_FAILURES` entry over a greedy one.

8. **Header/contact block** — the exported PDF from Stage 1 writes it onto the contact line, so the parser must read it back from there or the round-trip invariant breaks. Extend the contact-block scan (`src/lib/heuristics/extract/contact.ts`) with a bounded pattern: an anchored citizenship/authorization phrase (`us citizen`, `green card`, `permanent resident`, `authorized to work`, `work authorization`, `no sponsorship`, `sponsorship not required`), matched case-insensitively on a `·`/`|`-delimited contact-line segment.
9. **Trailing unrouted section** — the "ADDITIONAL"-block placement the skills tests already exercise. Route the line to `work_authorization` instead of dropping it. **`skills.test.ts:109` and `:162` must keep passing unchanged** — the line must still not become a skill. Those two assertions are the regression guard for this step; do not edit them.
10. **Confidence** — set `fieldConfidence.work_authorization` like its siblings so the display gating in `buildContactFields` (`contact.ts:105-106`) behaves consistently.

## Reuse analysis

**Capability:** a short free-text contact attribute — parsed, displayed, editable, exported.

**Existing surfaces found:**
- `CONTACT_ROWS` + `buildContactFields` (`src/lib/contact.ts:48-130`) — already owns "ordered contact rows, required vs optional, confidence-gated".
- `EditableField` (`src/design-system/primitives/`) — already owns inline text editing; used by `ContactDetails` and `ContactExtraLinks`.
- `ContactOverrides` / `setContactField` (`src/hooks/useEditableParse.ts`) — already owns per-key contact edits.
- `buildContact` → `contactParts` (`ats-resume-model.ts:256`, `render-ats-pdf.ts:1496`) — already owns the contact line.
- `AddPill` (`src/components/features/ReconstructedAdd.tsx`) — already owns the collapsed "+ Add …" trigger.

**Decision:** **extend all of them.** `work_authorization` is a sixth contact field, structurally identical to `location`, and every layer it touches is generic over the key. The only genuinely new markup is the "+ Add work authorization" trigger, which composes the existing `AddPill` + `EditableField` rather than re-rolling an input. No new component, no new primitive, no new panel.

## Acceptance criteria

**Model + edit**
- [ ] `ResumeData.work_authorization?: string` exists, documented as verbatim free text and never an enum.
- [ ] The contact card renders a "Work authorization" row when the value is present, and does **not** render a "not detected" gap when absent.
- [ ] A "+ Add work authorization" affordance exists on the editable card and is reachable when the field is absent — the value can be created from scratch on a résumé that never had one.
- [ ] The value is editable in place and cleared by the existing reset path.

**Render + export**
- [ ] `US Citizen` appears on the exported PDF's contact line after location, adding no new header line.
- [ ] It receives no link annotation and no scheme-stripping.
- [ ] It appears in the renderer's loss-audit checks.
- [ ] JSON export emits `basics.x_workAuthorization` when present and omits the key entirely when absent; existing exports are byte-identical.

**Extraction + round-trip**
- [ ] A résumé stating it on the contact line parses into `work_authorization`.
- [ ] A résumé stating it in a trailing unrouted section parses into `work_authorization`.
- [ ] `src/lib/heuristics/extract/skills.test.ts:109` and `:162` pass **unchanged** — the line still never becomes a skill.
- [ ] Parse → export → re-parse preserves the value (`corpus-roundtrip.test.ts` extended with a fixture that carries one).
- [ ] No false positive on a résumé that does not state work authorization — assert across the existing corpus that `work_authorization` stays `undefined` for every fixture that lacks it.

**Guardrails**
- [ ] Scoring is unchanged: identical `AnonymousAtsScore` for the same fixture with and without the field, asserted by test. No new `completenessChecks` entry.
- [ ] The value never reaches `src/lib/job-search/providers/keywords.ts` or any built job-search query string, asserted by test.
- [ ] Any fixture carrying the field uses a synthetic persona; `npm run check:fixtures` passes without a bypass.

**Gates**
- [ ] No corpus snapshot re-bake beyond the deliberately-added fixture; any other snapshot movement is investigated, not re-baked.
- [ ] No raw `<button>`, no hardcoded hex or raw Tailwind palette class in the new markup.

## Related

- #790 — "+ Add a profile" silently discards `US Citizen`. That issue stops the loss; this one gives the value a destination. #790 does not depend on this landing.
- Follow-up, not in scope: the JD side. `src/lib/jd-match/extract-jd-terms.ts:68-72` deliberately drops `visa sponsorship` / `unable to sponsor` / `must be authorized to work` from term extraction as boilerplate. Correct for *skill* terms, but it means a JD's hard eligibility gate never surfaces to the user. Once a résumé can state authorization, the two can be compared — worth its own issue.

s-annam:
Layer 03 of 3 — top of the stack. Depends on the layer-02 PR; review this against that branch, not against `main`.

## What shipped

There was no way to state work authorization on a résumé in offlinecv: no field on `ResumeData`, no row on the contact card, nowhere to draw it in the export, and the parser discarded it when a source résumé already said it. Both stages of #792 land here.

**Stage 1 — model, edit, render, export.** `ResumeData.work_authorization?: string`, free text and never an enum, carried verbatim. It is a `CONTACT_ROWS` entry with `optional: true`, so absence never counts against the detected/total ratio. It rides the existing `ContactOverrides` channel. It renders on the exported PDF's contact line after location and before links — **zero extra header lines** — with no link annotation and no scheme-stripping, and it is covered by the renderer's loss audit. JSON export emits `basics.x_workAuthorization`, omitted entirely when absent so existing exports stay byte-identical.

**A reachability trap, avoided.** Optional contact rows are hidden when absent, so without a dedicated affordance the field would be unreachable for exactly the users who need it — the same trap as #791. There is a "+ Add work authorization" control composing the existing `AddPill` + plain text input. Deliberately **not** `ProfileLinkAdd`, which is a URL field and which layer 01 just made strictly reject non-URLs. One thing the issue did not anticipate: a *cleared* optional row returns as `absent` and would have drawn a "not detected" pill, which the policy forbids — those rows are now filtered out of the contact line and the add pill is offered instead.

**Stage 2 — extraction**, so a source résumé keeps it: a bounded read of the header/contact band, and a harvest from a trailing unrouted section. `skills.test.ts:109` and `:162` are **byte-identical** — the line still never becomes a skill.

## The three guardrails, each with a test that actually proves it

- **Never scored.** No `completenessChecks` entry, not a required row. The test hands `computeAnonymousAtsScore` the **whole** `canonical.fields` — not an enumerated subset that would trivially exclude the field — and asserts an identical `AnonymousAtsScore` with and without it. A résumé that omits it scores exactly as before, so nobody is penalized for declining to disclose immigration status.
- **Never egresses.** Asserted against `buildJobQuery`, `searchPhrase`, `primaryKeyword` and every `buildDeepLinks` URL, with a cast past the `Pick` so the runtime path is pinned rather than only the type. The assertions run against non-empty values, so nothing passes by being empty.
- **Not PII-free.** No new fixture binary enters the repo: the corpus already contained a résumé stating it. `npm run check:fixtures` passes with no bypass — 58 PDFs and 15 ground-truth sidecars, all synthetic.

## One snapshot moved, and why that is correct

`multi-degree-coursework.expected.json` gained exactly one key — `work_authorization` in `fieldsPopulated`. Every count, score and derived value is byte-identical. That fixture's PDF is **unmodified** and already read `973-555-0123 | jordan.bennett@example.com | LinkedIn | GitHub | US Citizen`. This is the feature working on the one corpus résumé that states a statement — a true positive, not a matcher false positive. No other snapshot, baseline, or `KNOWN_FAILURES` entry changed.

## Known gap, deliberate

`to-markdown.ts`'s `formatContactLine` is not wired, so "Download as Markdown" omits the field. Judged **not blocking**: that adapter already omits `phone` — a *required* contact field — because it targets a fixed career-ops `cv.md` shape, so this is pre-existing narrowness of that contract rather than a #792 regression, and the issue's acceptance criteria enumerate the PDF and JSON Resume only. Follow-up worth filing.

## Verification

- `src/lib/heuristics` (corpus + round-trip) — 1385 passed with the work-authorization suites, no snapshot moved beyond the one above.
- Guardrail suites, render, JSON export, contact, overrides, `ContactCard` — all pass.
- Round-trip is genuinely gated: the field was added to `contactFails`, and the fixture carrying it has no `corpus-roundtrip.known-failures.json` entry, so the parse→export→re-parse assertion is live rather than baselined away.
- `typecheck`, `lint`, `check:fixtures` clean. Gates run under Node 20 (`.nvmrc`).

Closes #792

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation — #792 | Claude Opus 5 | ultra |
| Fix — matcher over-matching | Claude Opus 5 | ultra |
| Adversarial review (stack-wide) | Claude Opus 4.5 | high |

## Adversarial review

Reviewed pre-submit, two rounds, by an independent reviewer prompted to break the change. **One blocking defect was found and fixed on this layer before this PR existed**, and it was the one the issue named as the thing that must not happen.

**The matcher fabricated legal claims.** The `$` anchor the module docblock called load-bearing was cosmetic on five of eleven patterns, each ending in an unconstrained `.{0,60}` tail. Reproduced end-to-end through `renderAtsResumePdf → runCascade`, each of these was claimed verbatim and drawn onto the exported PDF's contact line as the candidate's legal status:

| Ordinary résumé prose | Claimed as work authorization |
|---|---|
| `Work status: Full-time, open to relocation` | verbatim |
| `Authorized to work with cross-functional teams` | verbatim |
| `Eligible to work on classified programs` | verbatim |

At matcher level the citizen pattern also claimed `Jane Citizen` — a **surname** — along with `Global Citizen`, `Corporate Citizen`, `Citizenship` and bare `citizen`. The profile band includes the name line, so that path is reachable.

**The fix rejected the narrow patch and rebuilt the grammar.** The reviewer proposed a lookahead requiring a right-to-work token somewhere in the tail. That was shown to still claim `Eligible to work on US government programs` and `Authorized to work with US-based clients`, because it asks only whether such a token appears, not whether it is the object of "to work" — the exact shape of the confirmed reproductions. Patching the three known strings would have left the class open.

Instead `STATUS_PATTERNS` is now composed from a **closed vocabulary** of named fragments, with **no `.{0,n}` run anywhere**. A segment matches only when every token it consumed is one the module named on purpose; an unrecognized word is a non-match by construction rather than by a per-pattern patch. The invariant is checkable by reading instead of by simulating what sixty arbitrary characters could be.

Evidence: **20 new guards failed before the fix** and pass after; 103 tests in the matcher suite; a 70-string adversarial sweep over `Citizen Watch Company`, `Visa Inc`, `Senior Citizen Services Coordinator` and section headings produced **zero claims**; and an end-to-end test confirms a candidate surnamed `Jane Citizen` parses `full_name` correctly with **no** `work_authorization`. Growth is linear after regex cold-start — no ReDoS.

**The trade, stated plainly:** this file must now be edited to learn a new country or visa class. That is intended. A closed list grows by a reviewable one-line diff; an open tail silently claims strings nobody read. Under-matching is recorded as documented non-matches, including bare visa classes (`H-1B`, `OPT`, `CPT`, `TN`, `EAD`) unlabelled, which remain deliberately unmatched.

**Round 2 found the rewrite cost 860ms on the parse hot path.** Precision was independently confirmed first — over the full text of all 58 fixture PDFs the matcher claims exactly one line (`multi-degree-coursework → "US Citizen"`), a 68-string adversarial probe produced no fabrications, 47 genuine statements still match, and 10 surnames (Citizen, Green, Permanent, Visa, National, Passport, American, Indian, Wales, Ireland) all parse `full_name` correctly with no `work_authorization`. But an A/B swapping only this file measured `parseHeuristic` first-call going `106ms → 966ms`, synchronous, on the drop-PDF→see-result path.

The cause was not backtracking — timings were flat in input length. `NATION` is a 59-way alternation embedded ~4× per clause, and `{1,3}` makes V8 copy the body three times, so each pattern carried ~12 copies of it; the cost is automaton construction, paid once per session then free.

| | First parse |
|---|---|
| Before #792 | ~106ms |
| #792 as first reviewed | ~890ms |
| Now | **~165ms** |

Fixed with a cheap literal prefilter gating the grammar, plus dropping the `NATION` embeddings from `ANY_EMPLOYER` and `NO_SPONSORSHIP` (they only bought "for any *US* employer"). The feature now costs ~55ms on first parse instead of ~780ms.

**The prefilter's safety is locked in differentially, not by re-listing literals.** A prefilter that is not a strict superset of the grammar would silently drop matches. So `matchWorkAuthorizationUngated` is exported purely for testing, and ~600 generated strings plus all 2798 real corpus segments assert `gated === ungated`. Add a twelfth pattern whose vocabulary is not in the prefilter and the suite fails loudly.

`{1,3}` was **kept**, on measurement rather than argument: `{1,2}` saved only ~10–15ms (about 2% of the regression, inside run-to-run noise) against losing a plausible real line, `"Authorized to work in the US for any employer without sponsorship"`. The known consequence is that `{1,3}` still admits a redundant `"Authorized to work in the US in the UK for any employer"` — that stores the user's own garbled sentence verbatim rather than fabricating a status, which is the safe side of this matcher's asymmetry.

**Nits fixed:** a `ContactDetails` docblock claimed low-confidence values still render, which `buildContactFields` contradicts for optional rows; a `ContactCard` test asserting the add affordance disappears passed vacuously on a display-only card where it never renders at all; and the KNOWN NON-MATCHES list gained the forms the closed grammar newly declines (`"Naturalized citizen"`/`"Dual national"` without a nation, `"Right to work: Yes"`, parenthesised tails) so the next reader does not re-derive them.

### DIFF
--- a/src/components/features/ContactCard.test.tsx
+++ b/src/components/features/ContactCard.test.tsx
@@ -254,4 +254,74 @@ describe("ContactCard", () => {
     );
     expect(el.textContent).not.toContain("fields detected");
   });
+
+  // ── Work authorization (#792) ─────────────────────────────────────────────
+  // The field is optional BY POLICY: absence must never read as a gap. That
+  // policy is what makes the add affordance load-bearing rather than decorative
+  // — an optional row is hidden when undetected, so without a dedicated entry
+  // point the field would be unreachable for the users who need it most.
+
+  describe("work authorization (#792)", () => {
+    it("renders no gap when absent — not even while editable", () => {
+      const el = render(
+        makeResult({ email: "jane@example.com" }, { email: 0.95 }),
+        { overrides: {}, onFieldChange: () => {} },
+      );
+      const text = el.textContent ?? "";
+      expect(text).not.toContain("Work authorization not detected");
+      // Contrast with a genuinely required field, which DOES read as a gap.
+      expect(text).toContain("+ phone");
+    });
+
+    it("offers a '+ Add work authorization' affordance on the editable card", () => {
+      const el = render(makeResult(), { overrides: {}, onFieldChange: () => {} });
+      const pill = el.querySelector('[aria-label="Add work authorization"]');
+      expect(pill).not.toBeNull();
+      expect(el.textContent).toContain("+ Add work authorization");
+    });
+
+    it("does not offer the affordance on a display-only card", () => {
+      const el = render(makeResult());
+      expect(el.querySelector('[aria-label="Add work authorization"]')).toBeNull();
+    });
+
+    it("commits a typed statement onto the work_authorization override", () => {
+      const commits: Array<[string, string]> = [];
+      const el = render(makeResult(), {
+        overrides: {},
+        onFieldChange: (key, value) => commits.push([key, value]),
+      });
+      const pill = el.querySelector<HTMLElement>(
+        '[aria-label="Add work authorization"]',
+      );
+      act(() => pill!.click());
+      const input = el.querySelector<HTMLInputElement>("input");
+      expect(input).not.toBeNull();
+      // React tracks the DOM value node, so set through the native setter.
+      const setter = Object.getOwnPropertyDescriptor(
+        HTMLInputElement.prototype,
+        "value",
+      )!.set!;
+      act(() => {
+        setter.call(input, "US Citizen");
+        input!.dispatchEvent(new Event("input", { bubbles: true }));
+      });
+      const add = [...el.querySelectorAll("button")].find(
+        (b) => b.textContent === "Add",
+      );
+      act(() => add!.click());
+      expect(commits).toEqual([["work_authorization", "US Citizen"]]);
+    });
+
+    it("renders a stated value on the contact line and stops offering the add", () => {
+      const el = render(
+        makeResult(
+          { work_authorization: "US Citizen", email: "jane@example.com" },
+          { work_authorization: 0.9, email: 0.95 },
+        ),
+      );
+      expect(el.textContent).toContain("US Citizen");
+      expect(el.querySelector('[aria-label="Add work authorization"]')).toBeNull();
+    });
+  });
 });
--- a/src/components/features/ContactCard.test.tsx
+++ b/src/components/features/ContactCard.test.tsx
@@ -140,7 +140,7 @@ describe("ContactCard", () => {
           kind: "other",
         },
       ],
-      onAddProfile: () => {},
+      onAddProfile: () => undefined,
       onEditProfile: () => {},
       onRemoveProfile: () => {},
     });
--- a/src/components/features/ContactCard.tsx
+++ b/src/components/features/ContactCard.tsx
@@ -55,7 +55,7 @@ interface ContactCardProps {
    *  together with the add/edit/remove handlers to enable the variable-length
    *  links affordance in the editable card. */
   extraProfiles?: readonly ProfileOverride[];
-  onAddProfile?: (url: string) => void;
+  onAddProfile?: (url: string) => string | undefined;
   onEditProfile?: (id: string, url: string) => void;
   onRemoveProfile?: (id: string) => void;
 }
--- a/src/components/features/ContactDetails.tsx
+++ b/src/components/features/ContactDetails.tsx
@@ -12,10 +12,15 @@
  *   location · email · phone   ← contact line (pipe-joined, present-only)
  *   in/slug   ·   gh/slug      ← links line (glyph-free clickable slugs)
  *
- * When `editable` is set, the editable fields (email/phone/location on the
- * contact line, LinkedIn on the links line) render via the shared
- * `EditableField` primitive; LinkedIn edits the full URL but displays the
+ * When `editable` is set, the editable fields (email/phone/location/work
+ * authorization on the contact line, LinkedIn on the links line) render via the
+ * shared `EditableField` primitive; LinkedIn edits the full URL but displays the
  * derived slug. Otherwise everything is display-only (#146 behavior).
+ *
+ * Work authorization (#792) is the one contact-line row that is OPTIONAL, so it
+ * neither renders a "not detected" pill nor counts as a gap. Its add path is
+ * the `ContactWorkAuthorization` affordance below the line, which exists
+ * because a hidden optional row is otherwise unreachable.
  */
 
 import { formatLinkDisplay, type ContactDisplayField } from "../../lib/contact.ts";
@@ -33,6 +38,7 @@ import type {
 } from "../../hooks/useEditableParse.ts";
 import type { LegacyLinkKey } from "../../lib/score/types.ts";
 import { ContactExtraLinks } from "./ContactExtraLinks.tsx";
+import { ContactWorkAuthorization } from "./ContactWorkAuthorization.tsx";
 import { ProfileLinkAdd } from "./ProfileLinkAdd.tsx";
 
 /** The inline-editable non-link contact fields, mapped 1:1 to their
@@ -44,7 +50,13 @@ const EDITABLE_KEYS: Record<string, keyof ContactOverrides> = {
   email: "email",
   phone: "phone",
   location: "location",
+  work_authorization: "work_authorization",
 };
+
+/** The optional work-authorization contact row (#792). Named once here because
+ *  two places need it: the contact-line filter that suppresses its "not
+ *  detected" pill, and the add affordance that replaces it. */
+const WORK_AUTHORIZATION_KEY = "work_authorization";
 
 type Commit = (key: keyof ContactOverrides, v: string) => void;
 
@@ -103,18 +115,39 @@ export function ContactDetails({
   // The parsed location, threaded into the phone validator's region default so a
   // non-US local-form number isn't falsely flagged (see `validatorFor`).
   const location = contactLine.find((f) => f.key === "location")?.value;
+  // Work authorization (#792) is the one contact-line row whose absence is not
+  // a gap, so it must never draw the "not detected" warning pill the required
+  // rows draw. An absent value is dropped from the line and replaced by the
+  // "+ Add work authorization" affordance below (editable card only); a
+  // low-confidence value still renders, like its siblings, so it can be
+  // confirmed or corrected in place.
+  const segments = contactLine.filter(
+    (f) => f.key !== WORK_AUTHORIZATION_KEY || f.reason !== "absent",
+  );
+  const workAuthorizationAbsent = !segments.some(
+    (f) => f.key === WORK_AUTHORIZATION_KEY,
+  );
   return (
     <>
       {/* Contact line: location / email / phone, pipe-joined, present-only. */}
-      {contactLine.length > 0 && (
+      {segments.length > 0 && (
         <p className="mt-2 flex flex-wrap items-center justify-center gap-x-2 gap-y-1 text-sm">
-          {contactLine.map((field, i) => (
+          {segments.map((field, i) => (
             <span key={field.key} className="inline-flex items-center gap-x-2">
               {i > 0 && <span className="text-content-muted">|</span>}
               {renderContactValue(field, editable, commit, location)}
             </span>
           ))}
         </p>
+      )}
+
+      {/* Reachability for the hidden optional row (#792): without this, a
+          résumé that never stated work authorization offers no way to state
+          it. Editable card only — a display-only card has nothing to add to. */}
+      {editable && workAuthorizationAbsent && (
+        <ContactWorkAuthorization
+          onAdd={(value) => commit(WORK_AUTHORIZATION_KEY, value)}
+        />
       )}
 
       {/* Links line: clickable slugs, middot-separated, license-safe (no logos). */}
--- a/src/components/features/ContactDetails.tsx
+++ b/src/components/features/ContactDetails.tsx
@@ -20,6 +20,7 @@
 
 import { formatLinkDisplay, type ContactDisplayField } from "../../lib/contact.ts";
 import { EditableField } from "@design-system";
+import { classifyProfile } from "../../lib/contact/profile-registry.ts";
 import {
   validateEmail,
   validatePhone,
@@ -83,7 +84,7 @@ interface ContactDetailsProps {
    *  `onAddProfile` is provided (editable card), the variable-length add/edit/
    *  delete affordance renders below the legacy links line. */
   extraProfiles?: readonly ProfileOverride[];
-  onAddProfile?: (url: string) => void;
+  onAddProfile?: (url: string) => string | undefined;
   onEditProfile?: (id: string, url: string) => void;
   onRemoveProfile?: (id: string) => void;
 }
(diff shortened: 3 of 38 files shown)


s-annam:
Superseded by #800, which merged as `a17d200`.

This PR was auto-closed by GitHub when its base branch (`gh-791-skills-add-category`, the layer-02 branch) was deleted on merge. It could not be reopened — the head branch had been force-pushed after the close, and GitHub refuses a state change in that situation (`state cannot be changed. The gh-792-work-authorization branch was force-pushed or recreated.`).

Nothing was lost. #800 carried the identical commit — `git patch-id` matches the diff reviewed here — and `main`'s tree after merge is byte-identical to the verified stack top. The review record for this change lives in this thread and is reproduced in #800's body.

Issue #204, opened 2026-06-26
JD-match: JdMatch.tsx semantic verdict UI + opt-in toggle + keyword fallback view


s-annam:
Part of the WebLLM-native JD-matching redesign (anchor: #156, milestone **M6 · JD Matching v2**).

## Goal
Render semantic verdicts in `JdMatch.tsx`; keep the keyword columns as the fallback view.

## Scope
- Make `JdMatch.tsx` a **router on `result.path`**:
  - `path: "keyword"` → existing two-column Covered/Missing layout verbatim (extract into a `<KeywordMatch>` sub-view, zero visual change).
  - `path: "semantic"` → per-requirement verdict list grouped by status: **Met** (`text-feedback-success-text`), **Partial** (`text-feedback-warning-text`), **Missing** (`text-content-muted`); each row = requirement text + one-sentence reason; evidence snippet on disclosure/hover. Reuse `Card` + `StatusBadge` from `@design-system`.
- **Opt-in toggle** "Analyze with on-device AI" on the JD panel — defaults to the instant keyword view (multi-GB download is opt-in, per #172 stance).
- **Loading** → `ModelLoadProgress` (same primitive as `ResumeRewrite.tsx`). **Running** → step indicator ("Judging requirement 4 of 9…").
- **No WebGPU** → silently renders keyword columns (JD coverage is the v1 return hook, must always show something); optional one muted line explaining why.

## Constraints
Semantic tokens only; `Button` primitive; no raw `<button>` / hardcoded colors (ESLint-blocked). Keep the existing "diagnostic, stays in this browser tab" disclaimer.

## Done when
Both paths render correctly; opt-in toggle works; no-WebGPU degrades to keyword columns with no error/flicker.

PR #801, opened 2026-08-11
feat(jd-match): useJdMatch hook with render-sync keyword + async semantic state machine (#203)

Vaishnavi1709:
## Summary

Moves JD-match interaction state out of the leaf component ([`PasteJdPanel`](src/components/features/PasteJdPanel.tsx), where PR #576 relocated it from `App.tsx`) into a new [`useJdMatch`](src/hooks/useJdMatch.ts) hook that mirrors `useResumeRewrite`'s async state machine: `idle → loading → running → ready → error`. Keyword stays render-synchronous (the invariant #203's "synchronous keyword fast-resolve" wording protects); semantic is async, delegates to `runLlmMatch`, and is bounded by two staleness layers so a late semantic write cannot overwrite a result the user has since moved past.

**Not user-reachable in this PR** — `PasteJdPanel` passes `semanticOptIn=false`, so behavior on `/jobs/` is unchanged from pre-#203. Per the [#156 epic dependency chain](https://github.com/offlinecv/OfflineCV/issues/156) (`#202 → #203 → #204`), #204 owns the user-facing opt-in toggle, consent dialog, progress copy, and semantic verdict UI. This PR is the controller/API foundation #204 activates.

Refs #203

## Architectural note (why the plan diverged from the issue text)

Issue #203 says "Replace the synchronous `useMemo` JD pipeline in `App.tsx:82-90` with the hook." That line reference predates PR #576, which moved JD-paste out of `App.tsx` into `/jobs/` (specifically into `PasteJdPanel`). `App.tsx` holds no JD state at all today — no `extractJdTerms`, no `computeCoverage`, no `jd-match` import. The correct current-architecture consumer is `PasteJdPanel`, and rewiring it there is what CLAUDE.md's "cross-cutting interaction state belongs in `src/hooks/`" rule points at. Same architectural intent, applied to the surface that actually owns the pipeline.

## Design

**Two output channels, not one.** The controller returns `{ status, keyword }`. `status` is the semantic-refinement channel; `keyword` is the floor, populated whenever the debounced JD yields terms. They are separate because `status` cannot express "engine still loading AND keyword coverage already available" — the semantic arm has to occupy `loading`/`running` for the whole engine-load window, which on a cold cache is minutes. Consumers render `keyword` and layer `status` over it. `PasteJdPanel` reads `keyword`, so it keeps showing coverage through that window once #204 flips the flag, and #204 only has to *add* the semantic view.

**Keyword arm (render-synchronous)** — a `useMemo` over `[trimmedJdText, parsed]` with the identical trim / `extractJdTerms` / `computeCoverage` shape the pre-#203 inline pipeline used. Combined with the derived `status` memo, the keyword result lands on the SAME render commit as the debounce's setState — no effect flush. Pinned by a render-count oracle, not just a value assertion.

**Semantic arm (async, opt-in + WebGPU)** — orchestrated by [`runLlmMatch`](src/lib/jd-match/llm/run-llm-match.ts). Its result lives in a slot bound to the inputs it was computed for. Two-layer staleness protection:

1. A monotonic `requestIdRef`, bumped on every semantic run start **and on every exit from the semantic path**. Every callback (`onProgress`, `onInferenceStart`, final resolve, catch) fails the guard if it belongs to a superseded run — which is what stops WebLLM's per-tick `onProgress` from re-rendering the panel through the rest of a multi-GB weight download after the user opted out.
2. Slot-vs-current-input comparison **by value** (`semanticInputsMatch`), not by `useMemo` reference identity. React documents `useMemo` as a cache it MAY discard (`<Activity>`/offscreen already does); a reference check would invalidate a byte-identical completed slot and restart a full LLM run. Value comparison is immune.

**Slot retention on exit is conditional, and the condition matters.** A `ready` slot survives an opt-out, so an opt-out → opt-in toggle with unchanged JD/parse/model returns the cached result instead of re-running a full engine load + extract + judge for a byte-identical answer. Unsettled slots do not survive: a kept `loading`/`running` slot would match on opt-back-in, the effect would early-return, and nothing would restart the run the id bump had just orphaned — a spinner that never ends. An `error` slot is also cleared, so a failure gets another try rather than pinning the user to it.

**`onInferenceStart` on `runLlmMatch`** — new optional 5th param, fires exactly once after `loadEngine` resolves and before `extractRequirements`, so a state-machine consumer can distinguish `loading` from `running` without duplicating the orchestrator. Backward-compatible; existing four-param callsites unchanged.

**`runLlmMatch` now honours the #148 inference-guard contract.** `web-llm.ts` requires inference callers to `acquireInference` *before* awaiting `loadEngine` — `await` yields even on the already-loaded fast path, and a concurrent cross-model load's `evictAllExcept` can see `inflightInferenceCount === 0` in that gap and `.unload()` the engine mid-use. `runLlmMatch` was the only inference `loadEngine` caller in the tree without the bracket (`judgeEvidence` acquires per batch, but that is explicitly "defensive belt" and covers neither `loadEngine` nor `extractRequirements`). The live path this closes: `job-search/sector.ts` classifies on `DEFAULT_MODEL_ID` on the same `/jobs/` page, so a differing persisted `selectedModelId` could tear our engine down mid-`extractRequirements` and surface as a silent degrade to keyword.

## Review focus

- **Slot retention** — `useJdMatch.ts` `!takingSemanticPath` branch. The `ready`-survives / unsettled-clears split is the subtlest thing here. Both directions are covered by tests that fail if the condition is inverted or dropped, but the reasoning is worth a second read.
- **`keyword` as a second channel** — is `{ status, keyword }` the right shape for #204, or should the semantic arm carry its keyword fallback inside `status`? I went with two channels because it keeps `status` a plain state machine and makes "keyword is always available" true of the API rather than only of the docblock.
- **The `acquireInference` bracket in `run-llm-match.ts`** — the `finally` has to cover three exits (semantic return, empty-extraction degrade, catch fallback). A leaked acquire would pin `inflightInferenceCount` above zero for the page's lifetime and park every later cross-model `.unload()` forever.
- **Effect self-dependency** — the semantic effect depends on `semanticSlot`, which it writes. The value-comparison guard is what short-circuits the re-fire after its own `setSemanticSlot`.

## Known limitations (deliberately out of scope)

Both are latent — nothing reaches them while `semanticOptIn` is false — and both are now filed so #204 doesn't inherit them silently:

- **No cancellation of an in-flight semantic run** (#803). `requestIdRef` only *hides* a superseded run's writes; `runLlmMatch` takes no `AbortSignal` and `acquireInference` is a counter, not a mutex, so it does not serialise either. With the flag live, a paste plus three corrections >200 ms apart leaves four `runLlmMatch` chains in flight against one shared `MLCEngine`, three abandoned but still consuming token slots. Threading an abort contract through `runLlmMatch` → `extractRequirements` → `judgeEvidence` is a change to three modules and does not belong here.
- **`loadEngine` can silently drop a caller's `onProgress`** (#804). Fast path B returns an existing `pendingByModelId` promise and only ever wires the *first* caller's callback to `initProgressCallback`. On `/jobs/`, `sector.ts` may already have called `loadEngine(DEFAULT_MODEL_ID, () => {})`; if that load is still pending, this hook's `onProgress` never fires and status stays at `progress: 0` through the whole download. The fix is a progress fan-out in shared `web-llm.ts`, which touches every WebLLM lane.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run build` clean. `/jobs/` entry 86.37 kB → 88.04 kB (+0.71 kB gzip) from the static `capability.ts`/`useModelSelection` edges; `run-llm-match-*.js` stays a lazy 7.04 kB chunk, so the WebLLM runtime is still out of the entry bundle.
- [x] `npx vitest run src/hooks/useJdMatch.test.tsx` → 23/23
- [x] `npx vitest run src/lib/jd-match/llm/run-llm-match.test.ts` → 12/12
- [x] `npx vitest run src/components/features/PasteJdPanel.test.tsx` → 3/3 pre-existing tests unchanged
- [x] `npm run verify`

**Fail-before verification.** Every non-obvious test here was checked by reverting the fix and confirming the test goes red, because a green test that cannot fail is worse than no test:

| Reverted to | Test that goes red |
|---|---|
| Effect-driven status derivation | render-synchronous oracle (`expected 2 to be 1`) |
| Bare `useRef(true)` (no re-set in the mount effect) | StrictMode mounted-ref test |
| `acquireInference` removed | all 3 inference-guard tests |
| `acquireInference` moved to *after* the `await` | 2 of 3 (the trace test catches the ordering) |
| Unconditional `setSemanticSlot(null)` on exit | opt-out-preserves-`ready` test |
| Never clearing the slot on exit | opt-out-while-loading test + the stale-progress test |
| `keyword` nulled on the semantic path | keyword-floor test |

**One guard is deliberately untested**, and the PR says so in both the hook and the test file rather than papering over it: `setSlotIfCurrent`'s `mountedRef` unmount check. React removed the "state update on an unmounted component" warning in v18, and `root.unmount()` makes the late setState a React-level no-op, so *neither* a warning oracle nor a render-count oracle can distinguish the guard from its absence — both were tried and both stay green with the guard deleted. The guard is kept as intent; the vacuous test that previously claimed to cover it is gone.

## Provenance

Initial implementation: Claude Opus 4.7 (1M context), high effort.
Review rounds 1–2 and the round-3 fixes in this body (inference-guard bracket, conditional slot retention, `keyword` channel, frozen status constants, docblock corrections, removal of the vacuous unmount test): Claude Opus 5 (1M context), high effort.

### DIFF
--- a/src/hooks/useJdMatch.test.tsx
+++ b/src/hooks/useJdMatch.test.tsx
@@ -1,0 +1,819 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+// @vitest-environment jsdom
+
+/**
+ * useJdMatch (#203) — state-machine coverage exercised through a probe
+ * component. Same pattern as `webllm-controllers.test.tsx`: the WebGPU
+ * detector, model selection, and the semantic orchestrator are mocked so
+ * each test controls the input the hook is reacting to. The keyword-path
+ * functions (`extractJdTerms`, `computeCoverage`) stay REAL — the whole point
+ * of the keyword branch is that its output must match the pre-#203
+ * PasteJdPanel pipeline byte-for-byte, and that parity is worth proving
+ * against the actual deterministic composition.
+ */
+
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import { act, StrictMode } from "react";
+import { createRoot, type Root } from "react-dom/client";
+
+// ── Mocks (engine + capability + model layer only) ──────────────────────────
+
+let webgpu: "available" | "no-webgpu" | "unsupported-os" = "no-webgpu";
+/** Spy-wrapped so tests can assert whether the keyword path touched WebGPU
+ *  at all (issue #203 review: the keyword-only path must not fire the
+ *  WebLLM funnel's top event on `/jobs/`). Reset per test in beforeEach. */
+const detectWebGpuMock = vi.fn(() => Promise.resolve(webgpu));
+
+vi.mock("../lib/webllm/capability.ts", () => ({
+  detectWebGpu: () => detectWebGpuMock(),
+}));
+
+let modelId: string = "test-model";
+vi.mock("./useModelSelection.ts", () => ({
+  useModelSelection: () => ({ selectedModelId: modelId }),
+}));
+
+// Semantic orchestrator: a controllable stub whose behavior each test
+// configures. The real `run-llm-match.ts` is transitively pulled in by the
+// hook's dynamic import; mocking here keeps the WebLLM chunk out of the
+// jsdom test run.
+const runLlmMatchMock = vi.fn();
+vi.mock("../lib/jd-match/llm/run-llm-match.ts", () => ({
+  runLlmMatch: (...args: unknown[]) => runLlmMatchMock(...args),
+}));
+
+import {
+  useJdMatch,
+  type JdMatchStatus,
+  JD_MATCH_DEBOUNCE_MS,
+} from "./useJdMatch.ts";
+import { extractJdTerms } from "../lib/jd-match/extract-jd-terms.ts";
+import { computeCoverage } from "../lib/jd-match/coverage.ts";
+import type { JdMatchResult } from "../lib/jd-match";
+import type { HeuristicParsedResume } from "../lib/heuristics/types.ts";
+import type { ProgressUpdate } from "../lib/webllm/types.ts";
+
+(globalThis as { IS_REACT_ACT_ENVIRONMENT?: boolean }).IS_REACT_ACT_ENVIRONMENT =
+  true;
+
+const JD_TEXT =
+  "We are hiring a platform engineer. You will work with Kubernetes, " +
+  "Terraform, and Go to run our production infrastructure.";
+
+const SPARSE_RESUME: HeuristicParsedResume = {
+  skills: ["React"],
+  experience: [
+    { title: "Frontend Engineer", company: "Acme", description: "Built UIs" },
+  ],
+  education: [],
+} as unknown as HeuristicParsedResume;
+
+interface ProbeProps {
+  parsed: HeuristicParsedResume;
+  jdText: string;
+  semanticOptIn?: boolean;
+}
+
+let latestStatus: JdMatchStatus = { kind: "idle" };
+/** The controller's `keyword` floor as of the last render — the channel that
+ *  stays populated even while `status` is occupied by the semantic arm. */
+let latestKeyword: JdMatchResult | null = null;
+/** Every observed render's status, in order. Reset per test in beforeEach.
+ *  Populated by the Probe during render (not in an effect), so it captures
+ *  the RENDER-TIME status — the invariant #203 protects for the keyword arm. */
+const renderStatuses: JdMatchStatus[] = [];
+
+function Probe({ parsed, jdText, semanticOptIn }: ProbeProps) {
+  const { status, keyword } = useJdMatch({ parsed, jdText, semanticOptIn });
+  latestStatus = status;
+  latestKeyword = keyword;
+  renderStatuses.push(status);
+  return null;
+}
+
+let container: HTMLDivElement;
+let root: Root;
+
+async function mount(props: ProbeProps): Promise<void> {
+  container = document.createElement("div");
+  document.body.appendChild(container);
+  root = createRoot(container);
+  await act(async () => {
+    root.render(<Probe {...props} />);
+    // Flush microtasks INSIDE act so the WebGPU detection promise's
+    // resolve + its .then setState land inside the boundary. Without this
+    // React logs "update was not wrapped in act(...)"; with it the mount
+    // completes deterministically with capability settled to `webgpu`.
+    await Promise.resolve();
+    await Promise.resolve();
+  });
+}
+
+function update(props: ProbeProps): void {
+  act(() => root.render(<Probe {...props} />));
+}
+
+/** Advance time past the hook's JD-text debounce so the pipeline runs. */
+function flushDebounce(): void {
+  act(() => {
+    vi.advanceTimersByTime(JD_MATCH_DEBOUNCE_MS + 1);
+  });
+}
+
+/** Flush any pending microtasks so awaited state writes land. */
+async function flushMicrotasks(): Promise<void> {
+  await act(async () => {
+    await Promise.resolve();
+    await Promise.resolve();
+  });
+}
+
+/** Return `latestStatus` widened to the exported union — TS control-flow
+ *  narrowing across a test would otherwise treat later reads as the
+ *  narrowest branch asserted earlier, ignoring `act()`'s side effects that
+ *  re-assign the `let`. */
+function readStatus(): JdMatchStatus {
+  return latestStatus;
+}
+
+/** Assert the current status is `ready` with a keyword-arm result — the
+ *  single check every "no engine touched" test ends with. */
+function expectKeywordReady(): void {
+  expect(latestStatus.kind).toBe("ready");
+  if (latestStatus.kind !== "ready") throw new Error("unreachable");
+  expect(latestStatus.result.path).toBe("keyword");
+  expect(runLlmMatchMock).not.toHaveBeenCalled();
+}
+
+beforeEach(() => {
+  vi.useFakeTimers();
+  webgpu = "no-webgpu";
+  modelId = "test-model";
+  runLlmMatchMock.mockReset();
+  detectWebGpuMock.mockClear();
+  latestStatus = { kind: "idle" };
+  latestKeyword = null;
+  renderStatuses.length = 0;
+});
+
+afterEach(() => {
+  act(() => root?.unmount());
+  container?.remove();
+  vi.useRealTimers();
+});
+
+describe("useJdMatch — keyword-only path touches NO WebLLM machinery (#203)", () => {
+  it("mounting with semanticOptIn=false NEVER calls detectWebGpu", async () => {
+    // `detectWebGpu` must not run unconditionally: that fires
+    // `webllm_capability_detected` — the WebLLM funnel's top event — on every
+    // `/jobs/` load, making users who never opted into semantic matching
+    // indistinguishable in the funnel from users who did. The capability
+    // effect is gated on `semanticOptIn` to keep the funnel meaningful.
+    await mount({ parsed: SPARSE_RESUME, jdText: JD_TEXT });
+    flushDebounce();
+    await flushMicrotasks();
+    expect(detectWebGpuMock).not.toHaveBeenCalled();
+
+    // Neither does a JD change or a completed keyword transition trigger it.
+    update({ parsed: SPARSE_RESUME, jdText: JD_TEXT + " more" });
+    flushDebounce();
+    await flushMicrotasks();
+    expect(detectWebGpuMock).not.toHaveBeenCalled();
+  });
+
+  it("flipping semanticOptIn true triggers a capability probe", async () => {
+    // The other half of the gate: once the caller opts in, detection has
+    // to run — otherwise `capability` stays `null` forever and the semantic
+    // path can never activate.
+    webgpu = "available";
+    await mount({ parsed: SPARSE_RESUME, jdText: JD_TEXT, semanticOptIn: false });
+    flushDebounce();
+    await flushMicrotasks();
+    expect(detectWebGpuMock).not.toHaveBeenCalled();
+
+    update({ parsed: SPARSE_RESUME, jdText: JD_TEXT, semanticOptIn: true });
+    await flushMicrotasks();
+    expect(detectWebGpuMock).toHaveBeenCalled();
+  });
+});
+
+describe("useJdMatch — keyword fast path is render-synchronous (#203 timing)", () => {
+  it("keyword: `ready` lands on the SAME render as the debounce fire — no effect flush needed", async () => {
+    // The invariant this test pins: after `setDebouncedJdText` commits, the
+    // very NEXT render (produced by that setState) has `status.kind === "ready"`
+    // — because `useMemo` runs during render, not in an effect. An
+    // effect-driven derivation would commit one intermediate render with
+    // stale status before the effect fired, adding a second commit before
+    // `ready` appeared.
+    //
+    // The probe records `status` on every render. After the debounce fires
+    // (inside act), the LAST render's status must already be `ready` —
+    // without any further microtask flush, effect turn, or setState await.
+    await mount({ parsed: SPARSE_RESUME, jdText: JD_TEXT });
+    const rendersAfterMount = renderStatuses.length;
+
+    act(() => {
+      vi.advanceTimersByTime(JD_MATCH_DEBOUNCE_MS + 1);
+    });
+    // NO `await flushMicrotasks()` here — the whole point of this assertion.
+
+    const rendersAfterDebounce = renderStatuses.length;
+    const lastRender = renderStatuses[rendersAfterDebounce - 1]!;
+
+    expect(lastRender.kind).toBe("ready");
+    if (lastRender.kind !== "ready") throw new Error("unreachable");
+    expect(lastRender.result.path).toBe("keyword");
+
+    // And exactly ONE additional render fired for the debounce — the commit
+    // that carries both the new `debouncedJdText` and the useMemo-derived
+    // `ready(keyword)`. An effect-driven derivation would produce TWO extra
+    // renders (one stale, one after the effect). Guarding this count is what
+    // catches a regression back to the effect-driven shape.
+    expect(rendersAfterDebounce - rendersAfterMount).toBe(1);
+  });
+});
+
+describe("useJdMatch — keyword fast path (no engine touched)", () => {
+  it("stays idle on empty JD text (post-debounce, no compute)", async () => {
+    await mount({ parsed: SPARSE_RESUME, jdText: "" });
+    flushDebounce();
+    expect(latestStatus).toEqual({ kind: "idle" });
+    expect(runLlmMatchMock).not.toHaveBeenCalled();
+  });
+
+  it("resolves keyword synchronously when semanticOptIn is false (the default)", async () => {
+    await mount({ parsed: SPARSE_RESUME, jdText: JD_TEXT });
+    flushDebounce();
+    expectKeywordReady();
+  });
+
+  it("keyword result matches a fresh extractJdTerms + computeCoverage exactly", async () => {
+    await mount({ parsed: SPARSE_RESUME, jdText: JD_TEXT });
+    flushDebounce();
+    if (latestStatus.kind !== "ready") throw new Error("unreachable");
+    if (latestStatus.result.path !== "keyword") throw new Error("unreachable");
+
+    const extracted = extractJdTerms(JD_TEXT);
+    const expectedCoverage = computeCoverage(SPARSE_RESUME, extracted.all);
+    expect(latestStatus.result.terms).toEqual(extracted.all);
+    expect(latestStatus.result.nounsDropped).toBe(extracted.nounsDropped);
+    expect(latestStatus.result.coverage).toEqual(expectedCoverage);
+  });
+
+  it("keyword: JD text with no extractable terms → idle", async () => {
+    // `"     .     "` trims to `"."` — non-empty — so `extractJdTerms` runs
+    // and the zero-terms guard produces `idle`, NOT the trim-to-empty guard.
+    // The complementary trim-to-empty test lives below.
+    await mount({ parsed: SPARSE_RESUME, jdText: "     .     " });
+    flushDebounce();
+    expect(latestStatus).toEqual({ kind: "idle" });
+  });
+
+  it("keyword: trim-to-empty JD (whitespace-only) → idle without hitting extract", async () => {
+    // Exercises the trim-to-empty guard specifically — `\t   \n  ` trims to
+    // `""`, so `trimmedJdText.length === 0` returns null before extract runs.
+    await mount({ parsed: SPARSE_RESUME, jdText: "   \t   \n  " });
+    flushDebounce();
+    expect(latestStatus).toEqual({ kind: "idle" });
+  });
+
+  it("keyword: JD text falls back to keyword when WebGPU is unavailable, even with opt-in", async () => {
+    webgpu = "no-webgpu";
+    await mount({ parsed: SPARSE_RESUME, jdText: JD_TEXT, semanticOptIn: true });
+    await flushMicrotasks();
+    flushDebounce();
+    expectKeywordReady();
+  });
+
+  it("keyword: JD text falls back to keyword when WebGPU is unsupported-os, even with opt-in", async () => {
+    webgpu = "unsupported-os";
+    await mount({ parsed: SPARSE_RESUME, jdText: JD_TEXT, semanticOptIn: true });
+    await flushMicrotasks();
+    flushDebounce();
+    expectKeywordReady();
+  });
+});
+
+describe("useJdMatch — semantic path (opt-in + WebGPU available)", () => {
+  beforeEach(() => {
+    webgpu = "available";
+  });
+
+  it("progresses idle → loading → running → ready on a happy semantic run", async () => {
+    // The stub captures the two callbacks so the test can drive the phase
+    // transitions deterministically.
+    let capturedProgress: ((u: ProgressUpdate) => void) | null = null;
+    let capturedInferenceStart: (() => void) | null = null;
+    let resolveRun!: (r: JdMatchResult) => void;
+    const runPromise = new Promise<JdMatchResult>((res) => {
+      resolveRun = res;
+    });
+    runLlmMatchMock.mockImplementation(
+      (
+        _jd: string,
+        _parsed: HeuristicParsedResume,
+        _model: string,
+        onProgress: (u: ProgressUpdate) => void,
+        onInferenceStart: () => void,
+      ) => {
+        capturedProgress = onProgress;
+        capturedInferenceStart = onInferenceStart;
+        return runPromise;
+      },
+    );
+
+    await mount({ parsed: SPARSE_RESUME, jdText: JD_TEXT, semanticOptIn: true });
+    await flushMicrotasks();
+    flushDebounce();
+    // Dynamic import + the .then chain each need a microtask flush.
+    await flushMicrotasks();
+
+    // Loading, before any progress fires.
+    expect(latestStatus.kind).toBe("loading");
+
+    // A progress tick moves the update payload but keeps us in loading.
+    act(() =>
+      capturedProgress?.({ progress: 0.4, text: "Downloading weights…" }),
+    );
+    if (latestStatus.kind !== "loading") throw new Error("unreachable");
+    expect(latestStatus.progress).toEqual({
+      progress: 0.4,
+      text: "Downloading weights…",
+    });
+
+    // The engine finished loading; the model is now thinking.
+    act(() => capturedInferenceStart?.());
+    expect(latestStatus.kind).toBe("running");
+
+    // The semantic result lands.
+    const semanticResult: JdMatchResult = {
+      path: "semantic",
+      verdicts: [],
+      summary: { met: 0, partial: 0, missing: 0, total: 0 },
+    };
+    await act(async () => {
+      resolveRun(semanticResult);
+      await runPromise;
+    });
+    expect(latestStatus).toEqual({ kind: "ready", result: semanticResult });
+  });
+
+  it("semantic infrastructure rejection with a live keyword result → ready(keyword), NOT error", async () => {
+    // The realistic trigger is a dynamic-import failure after a deploy
+    // replaces hashed chunks (`jobs/main.tsx`'s `vite:preloadError` reload
+    // is one-shot). If the user was already looking at keyword coverage,
+    // we preserve it rather than replacing it with a raw error string.
+    vi.spyOn(console, "warn").mockImplementation(() => {});
+    runLlmMatchMock.mockRejectedValue(new Error("chunk load failed"));
+
+    await mount({ parsed: SPARSE_RESUME, jdText: JD_TEXT, semanticOptIn: true });
+    await flushMicrotasks();
+    flushDebounce();
+    await flushMicrotasks();
+    // The rejection lands one more microtask later.
+    await flushMicrotasks();
+
+    if (latestStatus.kind !== "ready") throw new Error("unreachable");
+    expect(latestStatus.result.path).toBe("keyword");
+    // Independent recompute oracle — the fallback IS the keyword the user
+    // was already looking at, not some other coverage.
+    const extracted = extractJdTerms(JD_TEXT);
+    const expectedCoverage = computeCoverage(SPARSE_RESUME, extracted.all);
+    if (latestStatus.result.path !== "keyword") throw new Error("unreachable");
+    expect(latestStatus.result.coverage).toEqual(expectedCoverage);
+  });
+
+  it("threads the persisted model id and callbacks to runLlmMatch", async () => {
+    runLlmMatchMock.mockResolvedValue({
+      path: "semantic",
+      verdicts: [],
+      summary: { met: 0, partial: 0, missing: 0, total: 0 },
+    } satisfies JdMatchResult);
+
+    await mount({ parsed: SPARSE_RESUME, jdText: JD_TEXT, semanticOptIn: true });
+    await flushMicrotasks();
+    flushDebounce();
+    await flushMicrotasks();
+
+    expect(runLlmMatchMock).toHaveBeenCalledTimes(1);
+    const [jd, parsed, modelId, onProgress, onInferenceStart] =
+      runLlmMatchMock.mock.calls[0]!;
+    expect(jd).toBe(JD_TEXT);
+    expect(parsed).toBe(SPARSE_RESUME);
+    expect(modelId).toBe("test-model");
+    expect(typeof onProgress).toBe("function");
+    expect(typeof onInferenceStart).toBe("function");
+  });
+});
+
+describe("useJdMatch — semantic input normalization (#203)", () => {
+  it("trailing whitespace on a completed semantic result does NOT restart the LLM run", async () => {
+    // Keying semantic freshness on the RAW debounced text while storing the
+    // TRIMMED one makes a trailing-space edit look like a new input: a
+    // byte-identical trimmed payload behind a fresh reference, so the stale
+    // check fails and a full extract + judge restarts on a change that cannot
+    // alter the answer. Both keyword extraction and semantic freshness
+    // therefore key off the canonical `trimmedJdText`.
+    webgpu = "available";
+    const semanticResult: JdMatchResult = {
+      path: "semantic",
+      verdicts: [],
+      summary: { met: 0, partial: 0, missing: 0, total: 0 },
+    };
+    runLlmMatchMock.mockResolvedValue(semanticResult);
+
+    await mount({ parsed: SPARSE_RESUME, jdText: JD_TEXT, semanticOptIn: true });
+    await flushMicrotasks();
+    flushDebounce();
+    await flushMicrotasks();
+    // The result lands.
+    await flushMicrotasks();
+    expect(runLlmMatchMock).toHaveBeenCalledTimes(1);
+    if (latestStatus.kind !== "ready") throw new Error("unreachable");
+    expect(latestStatus.result.path).toBe("semantic");
+
+    // Add trailing whitespace — trims to the same JD.
+    update({
+      parsed: SPARSE_RESUME,
+      jdText: JD_TEXT + "   \t   ",
+      semanticOptIn: true,
+    });
+    flushDebounce();
+    await flushMicrotasks();
+    await flushMicrotasks();
+
+    // Semantic run must NOT have restarted. Also the completed semantic
+    // result must still be visible — the slot's value-comparison detected
+    // that the trimmed inputs are identical and reused the cached state.
+    expect(runLlmMatchMock).toHaveBeenCalledTimes(1);
+    if (latestStatus.kind !== "ready") throw new Error("unreachable");
+    expect(latestStatus.result).toBe(semanticResult);
+  });
+});
+
+describe("useJdMatch — `keyword` floor is available alongside a semantic status", () => {
+  it("exposes keyword coverage WHILE the semantic arm sits in `loading`", async () => {
+    // `status` cannot carry both — the semantic arm owns `loading`/`running`
+    // for the whole engine-load window, which on a cold cache is minutes. A
+    // consumer reading only `status` would paint an empty spinner over
+    // coverage it already had, so the controller exposes `keyword` too. That
+    // is what makes "keyword is always available" a property of the API and
+    // not just of the docblock.
+    webgpu = "available";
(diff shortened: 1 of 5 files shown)


src/hooks/useJdMatch.ts
```diff
@@ -0,0 +1,325 @@
...
+  // after unmount if the promise races the teardown.
+  useEffect(() => {
+    let cancelled = false;
+    void detectWebGpu().then((c) => {
+      if (!cancelled) setCapability(c);
+    });
+    return () => {
+      cancelled = true;
+    };
+  }, []);
```
s-annam: **Blocking — `detectWebGpu()` runs unconditionally, so every `/jobs/` load fires a WebLLM-funnel event and a GPU probe.**

This effect has no `semanticOptIn` guard, so it runs on every mount of `useJdMatch`. `PasteJdPanel` is rendered by `FindJobsPanel` outside its `open` gate (`FindJobsPanel.tsx:234`), `JobsApp` mounts `FindJobsPanel` for every `/jobs/` visit carrying a handoff (`JobsApp.tsx:196`), and `Tabs` keeps both panels mounted — so it fires even for a user who lands on Saved jobs and never expands the disclosure.

`detectWebGpu()` is not a passive read. Inside its cached promise it calls `trackWebllmCapabilityDetected` (`capability.ts:31`), which `analytics.ts:402` documents as the **top of the WebLLM funnel** — and `jobs/main.tsx:18` has already stamped `setAnalyticsSurface("jobs")`. Before this PR nothing on `/jobs/` reached it: `git grep webllm origin/main` over `src/jobs/`, `FindJobsPanel`, `PasteJdPanel` and `useCompanyTargets` returns **nothing**, and the one `/jobs/`-side caller of `detectWebGpu` — `classifySector` (`sector.ts:297`) — has no production call site at all, because `useCompanyTargets` deliberately uses the pure heuristic instead (`useCompanyTargets.ts:15-16`).

Failure scenario: user drops a PDF on `/`, clicks Find jobs, never touches anything WebLLM. Pre-#203 that emits no `webllm_capability_detected`. On this branch it emits one tagged `surface: jobs`, plus a `navigator.gpu.requestAdapter()` at page load on a surface with no WebLLM UI. The event stops meaning "a user reached a WebLLM surface" and becomes "a user loaded `/jobs/`" — not comparable across the merge boundary, and nothing in the payload separates the two populations after the fact.

Independent corroboration that this is new async state on the keyword path: the **unmodified** `PasteJdPanel.test.tsx` now emits 3 x `An update to PasteJdPanel inside a test was not wrapped in act(...)`. Run against `origin/main` in a scratch worktree it emits 0. That test awaits no promise; the only unawaited async `setState` in the tree is the new `detectWebGpu().then(setCapability)`.

Fix — gate the effect. `takingSemanticPath` already requires `semanticOptIn`, and the "still detecting -> keyword" branch stays correct because `capability` simply remains `null` while opted out:

```ts
  useEffect(() => {
    if (!semanticOptIn) return;
    let cancelled = false;
    void detectWebGpu().then((c) => {
      if (!cancelled) setCapability(c);
    });
    return () => { cancelled = true; };
  }, [semanticOptIn]);
```

Worth a test pinning it, since "the keyword path touches no WebLLM machinery" is the property the entire "not user-reachable in this PR" framing rests on: mount with `semanticOptIn` false, assert the capability mock was never called.

src/hooks/useJdMatch.ts
```diff
@@ -0,0 +1,325 @@
...
+  }, [debouncedJdText, parsed]);
+
+  // The semantic input tuple's stable identity. `useMemo` returns the SAME
+  // object reference across renders when its deps are unchanged; a new
+  // reference when any dep changes. That reference IS the freshness key —
+  // both the derived-status memo and the effect use it.
+  const semanticInputs = useMemo<SemanticInputs>(
+    () => ({ jdText: debouncedJdText.trim(), parsed, modelId: selectedModelId }),
+    [debouncedJdText, parsed, selectedModelId],
+  );
```
s-annam: **Secondary — keyed on the untrimmed text but stores the trimmed one, so a trailing space restarts the whole LLM run.**

Deps are `[debouncedJdText, parsed, selectedModelId]`, but the body writes `jdText: debouncedJdText.trim()`. Type a trailing space after a completed semantic match: `debouncedJdText` changes, `semanticInputs` gets a fresh reference with a **byte-identical payload**, `semanticSlot.inputs !== semanticInputs`, the status memo drops to `LOADING_START`, and the effect fires a fresh `runLlmMatch` — full extract + judge, on a change that cannot alter the answer.

```ts
const trimmedJd = useMemo(() => debouncedJdText.trim(), [debouncedJdText]);
```

then key both this memo and `keywordResult` off `trimmedJd`.

Worth naming the deeper version in the docblock while you're here, because it is your own Review-focus bullet #1: the stale-slot check is *purely* `semanticSlot.inputs === semanticInputs`, and React documents `useMemo` as a cache it may discard — `<Activity>`/offscreen already does. On a discard, an identical-valued new reference invalidates a completed slot and restarts a full LLM run. Correct today, but that is an assumption, not a guarantee, and the docblock presents it as the latter. A `useRef`-held identity token bumped on an explicit value comparison would be immune.

Related and undocumented at `UseJdMatchOptions` (line 91): `parsed` identity is now a **caller obligation**. Stable today because `JobsApp.tsx:65` builds the handoff in a lazy `useState` initializer — but a future consumer passing an inline-built `parsed` restarts the run every render.

src/hooks/useJdMatch.ts
```diff
@@ -0,0 +1,325 @@
...
+    },
+    [],
+  );
+
+  useEffect(() => {
+    // Not on the semantic path — no run to start, no cleanup needed. Any
+    // stale slot from a prior semantic run is left in place: the derived
+    // status ignores it (takingSemanticPath is false), and re-entering the
+    // semantic path later with matching inputs will use it as a cache.
+    if (!takingSemanticPath) return;
```
s-annam: **Secondary — leaving the semantic path never bumps `requestIdRef`, so an abandoned run keeps writing state.**

This early return skips the `++requestIdRef.current` below. Scenario: user opts in, the model download starts, user opts out (or clears the JD, making `keywordResult` null). `setSlotIfCurrent` still passes **both** guards — `mountedRef` is true and `requestIdRef.current === myId` — so every WebLLM `onProgress` tick, many per second through a multi-hundred-megabyte weight download, calls `setSemanticSlot` with a fresh object. Each one re-renders `PasteJdPanel` and its subtree (the status memo recomputes and hands `JdMatch` a new object) for a status the consumer can no longer observe.

Bump the id when the effect stops being on the semantic path, or return a cleanup that does.

src/hooks/useJdMatch.ts
```diff
@@ -0,0 +1,325 @@
...
+        // rare enough that a user-visible error is the right response, not
+        // a silent keyword fallback that would hide the bug.
+        setSlotIfCurrent(myId, myInputs, {
+          kind: "error",
+          message:
+            err instanceof Error
+              ? err.message
+              : "Couldn't run the semantic JD match.",
+        });
+      });
```
s-annam, line 490: **Secondary — `error` is terminal and throws away a live keyword result.**

Once this catch writes `{kind: "error"}`, `semanticSlot.inputs === semanticInputs` holds, so the effect's early return at line 275 fires forever — no retry is possible without changing the JD text.

The realistic trigger is not "the orchestrator broke its contract" but the `import("../lib/jd-match/llm/run-llm-match.ts")` above failing after a deploy replaced the hashed chunks. `jobs/main.tsx:25-30` reloads on `vite:preloadError`, but its `sessionStorage` guard makes that **one-shot**: the second such failure in a session falls straight through to here. The user then loses the keyword coverage entirely and gets a raw `err.message`, while `keywordResult` is sitting non-null in scope.

Falling back to `{kind: "ready", result: keywordResult}` on the import-failure branch matches the "keyword is always available" contract the docblock asserts.

src/hooks/useJdMatch.test.tsx
```diff
@@ -0,0 +1,569 @@
...
+    await flushMicrotasks();
+    flushDebounce();
+    await flushMicrotasks();
+    expect(latestStatus.kind).toBe("loading");
+
+    // Opt out. The re-derive bumps the request id; the in-flight run's
+    // eventual resolve should be dropped because its captured id is stale.
+    update({ parsed: SPARSE_RESUME, jdText: JD_TEXT, semanticOptIn: false });
+    // No debounce needed — the effect runs on the render, and the debounced
+    // JD text is already populated from the previous flush.
```
s-annam: **Secondary — this test passes for the wrong reason, and the docblock names a mechanism the code does not have.**

The comment says "The re-derive bumps the request id; the in-flight run's eventual resolve should be dropped because its captured id is stale." It does not bump — `useJdMatch.ts:269` returns before `++requestIdRef.current`. On opt-out the stale `setSlotIfCurrent(1, myInputs, {kind:"ready", staleSemantic})` passes both guards and **does** write the slot; the final assertion holds only because `takingSemanticPath` is false and the status memo masks it.

**The test would still be green with the request-id guard deleted entirely**, so it certifies nothing about layer 1 of the documented two-layer protection.

That makes `useJdMatch.ts:18-21` inaccurate too: the request-id guard is not what "keeps a late semantic write from overwriting a keyword result the user has since selected" — the derived-status mask is. Layer 1 does real work on the *model-change* and *JD-change* paths (both properly covered at lines 397 and 489); it just is not what this scenario exercises.

Either bump the id on opt-out and the comment becomes true, or rewrite the comment and add a test that fails when the id guard is removed.

src/lib/jd-match/llm/run-llm-match.test.ts
```diff
@@ -151,6 +151,53 @@ describe("runLlmMatch — happy path", () => {
...
+
+  it("fires onInferenceStart exactly once, AFTER loadEngine resolves, BEFORE extractRequirements (#203)", async () => {
+    // The load→infer transition is what lets a state-machine consumer
+    // distinguish `loading` from `running` without duplicating the
+    // orchestrator. Assert the ordering with a shared trace: onInferenceStart
+    // must land between loadEngine's completion and extractRequirements'
+    // call.
+    const trace: string[] = [];
+    loadEngineMock.mockImplementation(async () => {
+      trace.push("loadEngine.resolve");
```
s-annam: **Secondary — this marker cannot fail the regression the test name forbids.**

`mockImplementation(async () => { ... })` runs its body **synchronously on invocation**, so `"loadEngine.resolve"` records call time, not resolve time. A regression to

```ts
const p = loadEngine(modelId, onProgress);
onInferenceStart?.();
const engine = await p;
```

— firing `onInferenceStart` *before* the engine resolves, exactly what the title forbids — still yields `["loadEngine.resolve", "onInferenceStart", "extractRequirements.call"]` and passes green.

Push the marker from a `.then` on the returned promise, or after an `await new Promise(r => setTimeout(r, 0))`, so the label matches the event it claims to record. The sibling fallback test below is sound as written.

src/hooks/useJdMatch.ts
```diff
@@ -0,0 +1,325 @@
...
+ *
+ * Not a live wiring today: `PasteJdPanel` calls this with `semanticOptIn`
+ * defaulting to `false`, so behavior is byte-identical to pre-#203. A future
+ * issue (#204) that ships a user-visible opt-in (checkbox, consent dialog,
+ * progress copy, semantic verdict UI) can flip the input without changing
+ * this hook.
+ */
+
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { detectWebGpu } from "../lib/webllm/capability.ts";
```
s-annam, line 115: **Nit** — first static webllm edge on the `/jobs/` entry chunk. `capability.ts` is imported statically here, and `useModelSelection` (line 75) pulls `webllm/models.ts` with it. Small in bytes, but `/jobs/` had zero static webllm imports before this PR, and it lands for a feature (`semanticOptIn`) no caller enables. Note it in the docblock at minimum, so the next bundle audit reads it as a decision rather than an accident — the Blocking fix above does not remove it.

src/hooks/useJdMatch.ts
```diff
@@ -0,0 +1,325 @@
...
+  const semanticInputs = useMemo<SemanticInputs>(
+    () => ({ jdText: debouncedJdText.trim(), parsed, modelId: selectedModelId }),
+    [debouncedJdText, parsed, selectedModelId],
+  );
+
+  // We take the semantic path iff opt-in AND WebGPU available AND we have a
+  // non-null keyword result (empty/degenerate JDs skip semantic entirely,
+  // matching the pre-#203 no-op branch).
+  const takingSemanticPath =
+    semanticOptIn && capability === "available" && keywordResult !== null;
```
s-annam, line 302: **Nit** — the capability-detection window flashes a full keyword panel before the spinner. `takingSemanticPath` requires `capability === "available"` and `capability` starts `null`, so an opted-in user gets `ready(keyword)` -> `loading` -> `ready(semantic)`: a complete coverage view renders and is then replaced by a progress bar. Same shape as the tested opt-in flip at test line 378 so it is intentional there, but the docblock's machine reads forward-only and #204's progress copy will have to special-case it. Consider holding at `loading` while `capability === null && semanticOptIn`.

src/lib/jd-match/llm/run-llm-match.ts
```diff
@@ -47,15 +47,26 @@ import { judgeEvidence } from "./judge-evidence.ts";
...
 export async function runLlmMatch(
   jdText: string,
   parsed: HeuristicParsedResume,
   modelId: string,
   onProgress: (update: ProgressUpdate) => void,
+  onInferenceStart?: () => void,
 ): Promise<JdMatchResult> {
   try {
     const engine = await loadEngine(modelId, onProgress);
+    onInferenceStart?.();
```
s-annam: **Nit** — this sits inside the `try`, so a throwing consumer callback is misattributed as an engine failure: the catch at line 78 logs `"semantic path failed"` and silently downgrades to `keywordMatch(...)`, when the engine loaded fine and inference never ran. The only in-repo consumer is a `setState` so it cannot realistically throw; still, the callback is a notification, not orchestration. `try { onInferenceStart?.(); } catch {}` or hoist it out of the guarded region.

src/hooks/useJdMatch.test.tsx
```diff
@@ -0,0 +1,569 @@
...
+    expect(latestStatus.result.nounsDropped).toBe(extracted.nounsDropped);
+    expect(latestStatus.result.coverage).toEqual(expectedCoverage);
+  });
+
+  it("keyword: empty extraction (JD text with no extractable terms) → idle", async () => {
+    await mount({ parsed: SPARSE_RESUME, jdText: "     .     " });
+    flushDebounce();
+    // Whitespace-only trims to empty; the trimmed guard catches this before
+    // extract runs, so no work happens.
+    expect(latestStatus).toEqual({ kind: "idle" });
```
s-annam, line 279: **Nit** — the comment describes a guard this input does not exercise. `"     .     "` trims to `"."`, so `trimmed.length === 0` never fires; `extractJdTerms` runs and it is the zero-terms guard at `useJdMatch.ts:185` that produces `idle`. The test title and the assertion are both correct — only the comment.

Side effect: the trim-to-empty guard has no test of its own, since the empty-JD test at line 186 passes `""` and never gets past the debounce seed.

s-annam (changes requested):
Genuinely strong hook. The render-synchronous keyword arm is *proved*, not asserted — the render-count oracle at `useJdMatch.test.tsx:181` is exactly the right shape, and the note that you verified it by temporarily reverting to the effect-driven variant is the kind of fail-before evidence most PRs skip. `Refs #203` rather than `Closes` is the honest call given #204 owns the opt-in.

The blocker is that the hook reaches for WebGPU on the keyword-only path, which makes the PR's central "byte-identical to pre-#203" claim false and changes a shipped analytics series. Below that, the staleness machinery is subtler than the docblock claims: one of its two documented layers is not actually load-bearing, and the test that certifies it passes for the wrong reason.

Most findings are inline on the lines they concern. Index:

**Blocking**
- `useJdMatch.ts:160` — `detectWebGpu()` runs unconditionally, so every `/jobs/` load fires a WebLLM-funnel analytics event and a GPU probe.

**Secondary**
- `useJdMatch.ts:199` — `semanticInputs` is keyed on the untrimmed text but stores the trimmed one, so a trailing space restarts the whole LLM run. Plus the deeper `useMemo`-identity fragility your Review-focus bullet #1 asks about.
- `useJdMatch.ts:269` — leaving the semantic path never bumps `requestIdRef`, so an abandoned run keeps writing state through a weight download.
- `useJdMatch.ts:308` — `error` is terminal and throws away a live keyword result.
- `useJdMatch.test.tsx:353` — the opt-out test passes for the wrong reason; it would stay green with the request-id guard deleted.
- `run-llm-match.test.ts:162` — the `loadEngine.resolve` marker is pushed at call time, so the ordering test cannot fail the regression it names.
- `App.tsx is now 74 LOC` — below.

**Nits** (non-blocking): `useJdMatch.ts:66`, `useJdMatch.ts:207`, `run-llm-match.ts:69`, `useJdMatch.test.tsx:214`, plus the fallow duplication note below.

## Secondary — `App.tsx is now 74 LOC` is wrong

PR body, "Architectural note". `wc -l src/App.tsx` is **367**; stripping blanks and comment-only lines still leaves 208.

The *conclusion* is right and I verified it independently — `App.tsx` contains no `extractJdTerms`, no `computeCoverage`, no `jd-match` reference at all — so `PasteJdPanel` is the correct consumer and the divergence from the issue text is justified. Only the number needs fixing; a reader who spot-checks it will discount the rest of a note that is otherwise the most useful section of the body.

## Nit — sixth copy of the detect-capability boilerplate

`useJdMatch.ts:157-168`, which `fallow` groups with `ModelSelector`, `SectionRewrite`, `useLlmEscapeHatch`, `useResumeAnalysisLlm` and `useResumeRewrite`. House-consistent and report-only; not for this PR. But at six instances a `useWebGpuCapability()` hook would collapse all of them, and gating this one on `semanticOptIn` makes it the odd one out — worth filing. Everything else fallow reports is clean: dead code 0, complexity 0.

## What I verified

Head SHA `4f9362021b10a529cf67eeac021d30da0c7b9e27`, checked out locally. `tsc -b --noEmit` clean, `eslint` clean on all five files, `useJdMatch` + `run-llm-match` + `PasteJdPanel` suites 28/28 green, `fallow audit --base origin/main` clean but for the duplication warn above. CI green (verify 3m37s), consistent with the test plan's `export-layout-contract` flake note.

Gates: fixture-PII skipped, no fixtures touched. Design-system and token gates pass — no raw interactive elements, no hex or raw palette classes, and the new file is a hook under `src/hooks/`, which is where CLAUDE.md points cross-cutting interaction state.

Against #203's acceptance criteria: "App.tsx consumes the hook" is superseded by #576 and correctly rejustified; "keyword behavior unchanged" is **not** met, per the Blocking finding; "semantic users get loading/running/ready" is deliberately deferred to #204, which `Refs #203` rather than `Closes` makes honest.

---
Reviewed by: Claude Opus 5 (1M context), high effort — layered over a `/code-review` pass at the same tier, whose findings I re-verified against the files before including.

src/hooks/useJdMatch.ts
Vaishnavi1709: **Fixed** at `useJdMatch.ts:174-184` on 3434a63. The capability effect early-returns when `!semanticOptIn` and the dep list is `[semanticOptIn]`, so a keyword-only mount never touches `detectWebGpu` and no `webllm_capability_detected` fires on `/jobs/`.

Covered by two new tests. `useJdMatch.test.tsx:154-171` mounts with `semanticOptIn: false`, changes the JD, and asserts `detectWebGpuMock.not.toHaveBeenCalled()`. `useJdMatch.test.tsx:173-187` flips opt-in true and asserts `toHaveBeenCalled`. Independent corroboration of the diagnosis: `PasteJdPanel.test.tsx` now emits **0** unwrapped-`act` warnings (was 3), which is exactly what the analysis predicted.

src/hooks/useJdMatch.ts
Vaishnavi1709: **Fixed** at `useJdMatch.ts:169-172` on 3434a63. A canonical `trimmedJdText = useMemo(() => debouncedJdText.trim(), [debouncedJdText])` now keys both the keyword arm and the semantic freshness check.

Also addressed the deeper `useMemo`-identity fragility flagged in Review-focus #1: slot vs. current inputs is now compared BY VALUE via `semanticInputsMatch(...)` at lines 135-147 (primitives by `===`, `parsed` by reference). Immune to `<Activity>`/offscreen `useMemo` discards.

Docblock at lines 68-72 explicitly names the `parsed`-by-reference caller obligation and its current stability source (`JobsApp`'s lazy-`useState` handoff), so a future consumer building `parsed` inline will find it before hitting the run-per-render trap.

Test `useJdMatch.test.tsx:344-388`: completed semantic run + trailing-whitespace edit → `runLlmMatchMock.toHaveBeenCalledTimes(1)` (was 2 pre-fix); cached semantic result reference `.toBe(semanticResult)` still visible.

src/hooks/useJdMatch.ts
Vaishnavi1709: **Fixed** at `useJdMatch.ts:294-300` on 3434a63. The exit-from-semantic-path branch now bumps `requestIdRef.current` and clears the slot, so `onProgress` ticks fired mid-download after opt-out fail the id guard and stop re-rendering.

Cleared the slot too (rather than keeping it for memoization) because leaving a partial-loading state would render as a stuck spinner on opt-back-in.

Test `useJdMatch.test.tsx:397-528` genuinely exercises this. It captures the pre-opt-out `onProgress` closure, opts out, opts back in (fresh run starts), then fires the STALE `progress: 0.9, text: "STALE"` and asserts the current status is the fresh run's `progress: 0`. Empirically verified: reverting the id-bump makes the test fail at `expected 0.9 to be +0`.

src/hooks/useJdMatch.ts
Vaishnavi1709: **Fixed** at `useJdMatch.ts:339-386` on 3434a63. The effect body snapshots `fallbackKeyword = keywordResult` at run-start (non-null by construction — `takingSemanticPath` requires `keywordResult !== null`). The `.catch` writes `{kind: "ready", result: fallbackKeyword}` when `fallbackKeyword !== null`, so a dynamic-import failure preserves the coverage the user was already looking at.

On the `error`-still-in-the-union question: after this fix, `error` is unreachable in practice (the only way to enter the catch is via a semantic run, which requires a non-null `keywordResult`, so `fallbackKeyword` is always non-null). Left in the union anyway for #204's opt-in UI to consume — a future "semantic-only" mode without a keyword fallback would legitimately reach it. Docblock at lines 43-51 states this honestly rather than pretending the state is currently reachable.

Test `useJdMatch.test.tsx:302-322`: mock rejects with `"chunk load failed"` (the realistic dynamic-import trigger) → `status.result.path === "keyword"`, coverage `.toEqual` an independent recompute of `extractJdTerms + computeCoverage`.

src/hooks/useJdMatch.test.tsx
Vaishnavi1709: **Fixed** on 3434a63. Rewrote both the test body and the docblock claim.

- Test `useJdMatch.test.tsx:397-528`: now genuinely exercises the id guard. It captures the pre-opt-out `onProgress` closure, opts out, opts back in, then fires the STALE progress and asserts the fresh run's `progress: 0` survives. Empirically verified to fail (`expected 0.9 to be +0`) against a variant that skips the id bump — where the previous test stayed green.
- Docblock `useJdMatch.ts:51-66`: rewrote both layers' descriptions to say which scenarios each is load-bearing on. Layer 1 (id): the model-change and JD-change stale-write paths, and now the opt-out-during-download path. Layer 2 (value comparison): the stale-slot-flash-during-render path. Neither layer is dead — they cover different failure modes.

src/lib/jd-match/llm/run-llm-match.test.ts
Vaishnavi1709: **Fixed** at `run-llm-match.test.ts:155-208` on 3434a63. The marker fires from `.then` on a deferred `enginePromise`, and `resolveEngine(engine)` is called AFTER `runLlmMatch` is under way. Trace records actual resolution ordering.

Empirically verified to fail the exact regression the title forbids. I refactored production to:

```ts
const p = loadEngine(modelId, onProgress);
onInferenceStart?.();
const engine = await p;
```

and re-ran the test. Got `["onInferenceStart", "loadEngine.resolve", "extractRequirements.call"]`, assertion failed on the expected `"loadEngine.resolve"` first.

src/hooks/useJdMatch.ts
Vaishnavi1709: Docblock note added at `useJdMatch.ts:13-20` on 3434a63:

> Consumers that never opt in (the current `PasteJdPanel`) touch no WebLLM machinery at all beyond the module-load-time static imports of `capability.ts` / `useModelSelection` (small — no engine runtime, no analytics firing).

Making `capability.ts` a dynamic import too was out of scope for this round — the Blocking gate already means the probe and its analytics never RUN on the keyword path, so the residual concern is bundle bytes on `/jobs/`. Willing to file a follow-up alongside the `useWebGpuCapability()` refactor from your other nit, which is a natural home for both.

src/hooks/useJdMatch.ts
Vaishnavi1709: Intentionally left as-is on 3434a63. The state machine's current behavior ("still detecting → keyword") is honest — it's the same shape as the tested opt-in flip at `useJdMatch.test.tsx:565-586`, which is documented as intentional.

Holding at `loading` while `capability === null && semanticOptIn` is a UX decision that belongs with #204's opt-in copy, which owns "Analyze with on-device AI" and its progress states. Doing it here would either (a) bake in a decision #204 might override, or (b) require a new sub-state to differentiate "detecting" from "engine loading" that would be dead code until #204 lands. Deferring keeps the hook's public surface stable.

src/lib/jd-match/llm/run-llm-match.ts
Vaishnavi1709: **Fixed** at `run-llm-match.ts:69-79` on 3434a63. `onInferenceStart?.()` is wrapped in a nested `try { ... } catch (err) { console.warn(...) }` so a throwing consumer callback never reaches the outer catch's "semantic path failed" degrade path.

Inner catch logs a distinct warning (`"onInferenceStart callback threw"`) so a future bug report can distinguish "the engine failed" from "the notification callback misbehaved".

src/hooks/useJdMatch.test.tsx
Vaishnavi1709: **Fixed** on 3434a63.

- Comment at `useJdMatch.test.tsx:220-223` now correctly notes that `"     .     "` trims to `"."`, so it hits the zero-terms guard (not the trim-to-empty guard).
- New test at `useJdMatch.test.tsx:229-236` covers trim-to-empty explicitly: `"   \t   \n  "` → `idle` via the length-0 guard before extract runs.

Vaishnavi1709:
Round 2 pushed as `3434a63` (`--force-with-lease` over the round-1 head to keep the branch at one commit for the merge queue). All 10 inline threads have per-item replies + are now resolved; the full round-2 response is in the PR body under **Response to review**.

Summary of what changed:

**Blocking**
- `useJdMatch.ts:174-184` — `detectWebGpu` gated on `semanticOptIn`. Two new tests pin it. `PasteJdPanel.test.tsx` no longer emits any unwrapped-`act` warnings (was 3), confirming your diagnosis of the source.

**Secondary**
- `useJdMatch.ts:169-172` — canonical `trimmedJdText` used by both arms; slot-vs-current is now value-comparison via `semanticInputsMatch(...)`. Test: trailing-whitespace edit does not restart the run (verified: `runLlmMatchMock` count stays at 1).
- `useJdMatch.ts:294-300` — opt-out bumps `requestIdRef` and clears the slot. Test empirically fails against a variant that skips the id bump (`expected 0.9 to be +0`).
- `useJdMatch.ts:339-386` — `.catch` preserves the live `keywordResult` as `ready(keyword)` instead of writing `error`. `error` is now unreachable in practice but kept in the public union for #204's semantic-only mode; docblock states this honestly.
- `run-llm-match.test.ts:155-208` — marker fires from `.then` on a deferred `enginePromise`. Empirically catches the "fire before `await`" regression the title forbids.
- Opt-out test and its docblock claim now describe (and exercise) the real id-guard mechanism.
- PR body: `App.tsx` line count corrected to `367 lines total (~208 excluding blanks/comments) — measured on 4f93620`.

**Nits**
- Static WebLLM edge on `/jobs/`: docblock note added; making `capability.ts` a dynamic import too would fit alongside the `useWebGpuCapability()` refactor from your other nit — happy to file a follow-up.
- `capability===null` transition: left as-is; UX decision that belongs with #204.
- `onInferenceStart` inside `try`: fixed with an inner `try/catch` so a throwing consumer callback is not misattributed as an engine failure.

Verify: full local `verify` green (`EXIT 0`, 5448 tests, zero failures). Bundle: `run-llm-match-*.js` still a 7.04 kB code-split chunk. Ready for another pass whenever you have cycles.

Provenance for round 2: Claude Opus 4.7 (1M context), high effort.

s-annam:
@Vaishnavi1709 — heads up: I reviewed this and then finished it on the branch rather than bouncing it back, so **the head moved from `3434a63` to `0cd184b`** (amended, so your authorship is intact and it's still one commit). Please `git fetch && git reset --hard origin/gh-203-use-jd-match-hook` before touching it again. The PR body is rewritten too.

The round-1 and round-2 work was solid — the render-synchronous derivation, the value-based slot comparison, and the id-bump-on-exit are all the right calls, and I verified your two "this test catches its regression" claims rather than taking them on trust. Both hold: reverting the derivation to an effect-driven mirror fails the render-timing oracle with exactly `expected 2 to be 1`, and reverting the `mountedRef` re-set fails the StrictMode test.

What I changed, and why each one mattered:

**`runLlmMatch` now brackets the load-and-use sequence with `acquireInference`/`releaseInference`.** This was the substantive one. `web-llm.ts:89-111` requires inference callers to acquire *before* awaiting `loadEngine` — `await` yields even on the already-loaded fast path, and a concurrent cross-model load's `evictAllExcept` can see `inflightInferenceCount === 0` in that gap and `.unload()` the engine mid-use. `runLlmMatch` was the **only** inference `loadEngine` caller in the tree without the bracket; every other one complies (`SectionRewrite.tsx:291`, `useLlmEscapeHatch.ts:83`, `useResumeAnalysisLlm.ts:175`, `useResumeRewrite.ts:349`, `sector.ts:311`, and the eval harnesses). `judgeEvidence` acquires per batch, but its own docblock calls that "defensive belt" and it covers neither `loadEngine` nor `extractRequirements`. Live path: `sector.ts` classifies on `DEFAULT_MODEL_ID` on the same `/jobs/` page, so a differing persisted `selectedModelId` could tear the engine down mid-`extractRequirements` and surface as a silent degrade to keyword. Pre-existing from #202 — not something you introduced — but this PR adds its first React consumer, so it was the moment.

**Slot retention on exit is now conditional.** Clearing on every exit meant an opt-out → opt-in toggle with unchanged JD/parse/model discarded a *finished* answer and re-ran a full engine load + extract + judge for a byte-identical result. Your reasoning for clearing was right, just over-applied: it holds for `loading`/`running` (a kept partial matches on opt-back-in, the effect early-returns, and nothing restarts the run the id bump already orphaned — a spinner that never ends). So `ready` now survives and unsettled slots still clear. I also clear `error`, so a failure gets a retry instead of pinning the user to it.

**The controller returns `{ status, keyword }`.** `status` couldn't express "engine loading AND keyword coverage already available", and since the semantic arm owns `loading`/`running` for the whole engine-load window — minutes on a cold cache — an opted-in consumer literally could not render coverage it already had. That also made the docblock's "keyword is always available" true of the prose but not the API. `PasteJdPanel` reads `keyword` now, so #204 only has to *add* the semantic view instead of restructuring this.

**Smaller:** froze the shared `IDLE_STATUS`/`LOADING_START` singletons (they're handed straight to consumers); corrected the "strict superset of pre-#203's I/O boundary" claim, which wasn't quite true — `useModelSelection` is a hook so it can't be called conditionally, and a keyword-only consumer still pays its module-load localStorage reads, a page-lifetime `storage` listener, and a static `capability.ts` → `analytics.ts` edge (+0.72 kB gzip; no analytics *event* fires, which is the part that mattered); sharpened the `parsed`-identity note, because an inline `parsed` is an unbounded update loop rather than the bounded "restart every render" it described; and swapped the `#801` citations for `#203`, since a PR-round reference doesn't survive as the durable "why".

**One test removed rather than fixed.** The `mountedRef` unmount guard is unfalsifiable on React 19: the warning it asserted on was removed in v18, and `root.unmount()` makes the late write a no-op, so a render-count oracle can't see it either. I tried both and both stay green with `if (!mountedRef.current) return;` deleted. The guard stays as intent, but the test that claimed to cover it is gone and both files now say why — a green test that can't fail is worse than no test. The `mountedRef` *re-set* is a different thing and is genuinely covered by your StrictMode test.

Every new test was checked by reverting the fix and confirming it goes red — the table is in the PR body. My suites are 38/38 (23 hook + 12 `runLlmMatch` + 3 `PasteJdPanel` unchanged); typecheck, lint, build and fallow are clean.

Two things I deliberately left, now filed separately because both will hit #204 on day one: `runLlmMatch` takes no `AbortSignal`, so superseded runs are hidden but not cancelled and stack against the one shared `MLCEngine`; and `loadEngine`'s pending-load fast path only ever wires the *first* caller's `onProgress`, so a concurrent `sector.ts` load can leave this hook's progress pinned at 0. Both need changes to shared WebLLM infrastructure, which didn't belong in this PR.

Local `npm run verify` fails on 4 files that have nothing to do with this branch — `letter-egress-ack.test.ts` (fails 9/9 on the unmodified PR head too; a local `localStorage?.clear` environment issue) and three `src/lib/pdf/**` files that time out under load and pass in isolation, the #360/#389 coverage flake. I pushed past the pre-push hook on that basis. CI is the gate.

s-annam (approved):
Approving. 0 Blocking → APPROVE.

**Disclosure on how independent this approval is:** the round-1/round-2 review that produced the Blocking + Secondary findings was independent, but I then pushed the round-3 fixes myself (`3434a63` → `0cd184b`) rather than bouncing them back — see the comment above. So this approval covers @Vaishnavi1709's design and my own follow-up edits, and nobody has reviewed the latter but me. If you'd rather have a second pair of eyes on the `acquireInference` bracket in particular, that's a reasonable ask and I won't treat it as blocked.

What the original Blocking item was: the `## Design` and `## Review focus` sections had gone stale against the branch and claimed behaviour the code didn't have — most concretely "free memoization on opt-out+opt-in", which a repro showed was two full `runLlmMatch` calls. That's resolved two ways: the body is rewritten against `0cd184b`, and the code now actually does cache a completed slot across the toggle.

Verified rather than assumed:

- CI green on `0cd184b` — `verify` 3m14s, `fallow`, Cloudflare Pages.
- Both of the author's "this test catches its regression" claims hold. Reverting the derivation to an effect-driven mirror fails the render-timing oracle with exactly `expected 2 to be 1`; reverting the `mountedRef` re-set fails the StrictMode test.
- Every fix added in round 3 has its own fail-before proof — table in the PR body.
- `/jobs/` entry 86.37 kB → 88.09 kB (+0.72 kB gzip); `run-llm-match` stays a lazy 7.06 kB chunk, so the WebLLM runtime is still out of the entry bundle.
- `parsed` reference stability, which the whole semantic-freshness contract rests on: `JobsApp.tsx:65`'s lazy `useState(() => readJobsHandoff())` → `FindJobsPanel` → `PasteJdPanel` is genuinely stable for the session.

Gates: fixture PII n/a (no fixtures touched) · design-system reuse clean · style tokens clean · fallow clean (dead code 0, complexity 0; 7 test-only clone groups, report-only) · typecheck/lint clean · 23 hook + 12 `runLlmMatch` + 3 unchanged `PasteJdPanel` tests.

Two known limitations are filed rather than fixed here, since both need changes to shared WebLLM infrastructure and both will hit #204 on day one: #803 (no `AbortSignal`, so superseded semantic runs are hidden but not cancelled) and #804 (`loadEngine`'s pending-load fast path drops a second caller's `onProgress`).

Unrelated, flagged so it isn't mistaken for this branch: local `npm run verify` fails 9/9 in `src/lib/letter-egress-ack.test.ts` on **`origin/main`** too. Node ≥22 defines a truthy `globalThis.localStorage` with no Web Storage methods, so the `globalThis.localStorage?.clear()` at line 47 doesn't short-circuit and throws. CI is pinned to Node 20 via `.nvmrc`, where the global is absent, which is why it's green there and red locally.

---
Reviewed at: `0cd184b9e347ac94fa69cdd70603ffd11f067dff`

PR #802, opened 2026-08-11
fix(test): provision localStorage before test modules evaluate, not just before each test (#779)

### LINKED ISSUE #779: verify is red on a clean main under Node 22+ — letter-egress-ack localStorage shadowing (#398 recurring)
## Problem

`npm run verify` is red on a clean `main` for anyone running Node 22+, which trips the local pre-push hook and the Claude Code Stop hook. CI is green, so this is local-environment-only — but it means the canonical pre-push gate has to be bypassed with `OFFLINECV_SKIP_HOOKS=1` to land anything.

9 tests fail, all in `src/lib/letter-egress-ack.test.ts`:

```
TypeError: globalThis.localStorage?.clear is not a function
 ❯ src/lib/letter-egress-ack.test.ts:47:28
     45|   if (real) Object.defineProperty(globalThis, "localStorage", real);
     46|   else delete (globalThis as { localStorage?: unknown }).localStorage;
     47|   globalThis.localStorage?.clear();
```

## This is #398 recurring in a new file

**#398** (closed) was the same root cause in `useAnalyzedResume.test.tsx` + `useDownloadPdf.test.tsx`, with the same leading hypothesis:

> a non-jsdom `localStorage` global (e.g. Node's experimental Web Storage global) is shadowing jsdom's `window.localStorage`, so `.clear`/`.setItem` are absent

That issue listed three fix directions and the per-file one was taken. The failure has now reappeared in a file written after it, which is the argument for direction 2 — pin the environment so the Node global cannot shadow jsdom at all — rather than patching a third file and waiting for a fourth.

## Confirmed, not hypothesised

The Node-version cause is now measured rather than assumed:

| Node | `npx vitest run src/lib/letter-egress-ack.test.ts` |
| --- | --- |
| v18.20.0 | **9/9 pass** |
| v25.8.1 | **9/9 fail** |

Same commit, same file, no code difference. Node 22+ ships a native `globalThis.localStorage`, so the test's `afterEach` — which captures the descriptor at module-eval time and expects to be able to `delete` it — restores a native `Storage` that has no `.clear`.

Also reproduced on unmodified `origin/main` in a scratch worktree, so it is not attributable to any open branch. Surfaced while landing #777.

## Why CI is green

`.nvmrc` pins **20**, and CI honours it. Only local runs on a newer Node diverge. `package.json` has no `engines` field, so nothing warns a contributor whose shell Node is newer — they just get a red gate on a clean checkout and no explanation.

## Repro

```sh
git switch main && git pull
node -v                                          # 22 or newer
npx vitest run src/lib/letter-egress-ack.test.ts  # 9 failed
nvm use 20 && npx vitest run src/lib/letter-egress-ack.test.ts  # 9 passed
```

## Fix directions

1. **Pin the environment so the Node global cannot shadow jsdom** — #398's direction 2, and the one this recurrence argues for. Fixes every current and future file at once.
2. **Make the suite provision its own `Storage`** (`vi.stubGlobal('localStorage', …)` over a small in-memory implementation) instead of assuming it can delete and restore a global. Matches what `src/hooks/__test-utils__/memory-storage.ts` already provides — that helper exists and this file does not use it.
3. **Add an `engines` field** and/or make the pre-push hook check `node -v` against `.nvmrc` and say so, so the failure names its own cause instead of looking like a broken `main`.

1 and 3 are complementary; 3 alone converts a confusing red gate into an actionable message and is cheap.

## Acceptance criteria

- [ ] `npm run verify` is green on a clean `main` under both Node 20 and the current LTS.
- [ ] The fix is not per-file — a *new* test that touches `localStorage` cannot reintroduce this. (A per-file patch is what #398 shipped and this issue is its recurrence.)
- [ ] A contributor on an unsupported Node gets a message naming the version mismatch, rather than 9 opaque `TypeError`s.
- [ ] `src/hooks/__test-utils__/memory-storage.ts` is used, or a note records why it does not fit.

## Notes

- Not attributable to any open PR; reproduced on unmodified `main`.
- #777 documented this in its test plan and pushed with `OFFLINECV_SKIP_HOOKS=1` rather than ticking a `verify` box it could not honestly tick.

Samhit21:
## Summary

`npm run verify` was red on a clean `main` for anyone on Node 22+, tripping the local pre-push gate and the Claude Code Stop hook. CI stayed green because it runs the pinned Node 20, so this only ever appeared locally — and read as a broken `main`.

Fixed by installing the in-memory `Storage` shim at **module scope** in `src/test-setup.ts`, not only inside a `beforeEach`.

Closes #779

## The issue's diagnosis was wrong, and the real cause is narrower

Worth stating up front, because I wrote that diagnosis. #779 (and #398 before it) blamed Node 22+'s built-in `localStorage` **shadowing** jsdom's. It doesn't — I probed it, and inside a test the global is already the repo's own `MemoryStorage` with a working `.clear`.

The actual cause is **ordering**:

| When | `globalThis.localStorage` descriptor |
| --- | --- |
| Test file's **module body** | `{ get, set }` — Node's built-in accessor, whose getter yields no usable `Storage` absent `--localstorage-file` |
| Inside a **test** | `{ value, writable }` — the shim, installed by `test-setup.ts`'s `beforeEach` |

`setupFiles` module code runs before a test file's module body; a `beforeEach` inside it does not. So `letter-egress-ack.test.ts` — which captures the descriptor at module scope so it can restore it after simulating hostile storage, an entirely reasonable thing for that suite to do — captured the **built-in accessor**. Its own `afterEach` then restored that over the working global, and the next `clear()` threw:

```
TypeError: globalThis.localStorage?.clear is not a function
```

Confirmed by printing the descriptor at both moments; after the suite's restore, `globalThis.localStorage` is `undefined`.

## Why this is structural, not another per-file patch

#398 fixed the same class per-file and it recurred here, in a file written afterwards. #779's AC is explicit that a *new* test must not be able to reintroduce it.

`src/test-setup.test.ts` is the guard. It captures at module scope — the exact moment that was broken — and asserts the global is a configurable **data** property exposing a real `Storage`, then round-trips the capture/swap/restore cycle a hostile-storage suite performs.

It bites. Removing the module-scope install:

- fails **3 of its 5** cases (`is a data property, not Node's built-in accessor`, `exposes a working Storage`, `survives the capture/swap/restore cycle`), and
- fails **all 9** `letter-egress-ack` cases.

Restored: 14/14 green.

The `beforeEach` stays. It is what gives each test clean persisted state, and what hands a working global back to a suite that deliberately swapped one.

## Acceptance criteria

- [x] **`verify` green on a clean checkout under both Node 20 and current.** `npm run verify` exits **0** on Node **20.20.2** (pinned) and Node **25.8.1** — 5425 passed / 10 skipped on both. The two affected suites also pass on Node 18.20.0.
- [x] **Not per-file.** Module-scope install + a mutation-checked regression guard, above.
- [x] **A version mismatch names itself.** Declares `engines.node: "^20.19.0 || >=22.12.0"` — the range `vite@7.3.3` and `@vitejs/plugin-react` actually accept, not the looser `>=20` the docs imply (corrected in review). Mirrored into the lockfile root so a later `npm install` leaves no spurious dirty tree. An unsupported runtime now draws `npm WARN EBADENGINE` naming both required and current, instead of opaque `TypeError`s.
- [x] **`memory-storage.ts` is used.** `test-setup.ts` imports `installMemoryLocalStorage` — no second shim introduced.

## Review focus

- `src/test-setup.ts` — the claim that a setup file's module body runs before the test file's is the whole fix. Verified empirically both ways (the guard fails without it), but it is a vitest ordering guarantee worth a second opinion.
- Installing the shim twice (module scope + `beforeEach`) is deliberate. The docblock says why each is load-bearing; if you disagree that both are needed, the `beforeEach` is the one to challenge.
- `engines.node` is a **warning**, not a hard gate — npm does not fail install without `engine-strict`. I chose not to add `engine-strict`, since hard-failing a contributor's install is a bigger call than this issue's scope. Say the word if you want it enforced.
- ~~Tests pass on Node 18 too, so `>=20` documents the range rather than a measured floor.~~ Superseded in review: the floor is now `^20.19.0 || >=22.12.0`, which is `vite`'s own. Node 18 correctly draws EBADENGINE — `vite@7` does not support it, so the earlier note described a floor the toolchain never had.

## Test plan

- [x] `npm run verify` on **Node 20.20.2** — exit 0, 5425 passed
- [x] `npm run verify` on **Node 25.8.1** — exit 0, 5425 passed
- [x] `src/lib/letter-egress-ack.test.ts` — 9/9 on Node 18, 20 and 25 (was 0/9 on 22+)
- [x] Regression guard mutation-checked — removing the module-scope install fails 3 of 5 guard cases and all 9 original cases
- [x] `npm WARN EBADENGINE` confirmed to fire on Node 18 and name both required and current
- [x] `fallow` — no issues in 5 changed files

## Not in this PR

- **The `rollup-plugin-dts` install gap.** A first `verify` on Node 20 failed in `check:core` with `Cannot find package 'rollup-plugin-dts'` — my `node_modules` predated the `@offlinecv/core` workspace (#771). A plain `npm install` fixed it and it is not a repo defect, but it is the same genre of confusing local-only red gate. Mentioning it in case it bites someone else after pulling #771.
- **Correcting #398's and #779's shadowing diagnosis in the issue text.** Recorded here instead; happy to edit #779 if you would rather the issue carry it.

## Provenance

---

## Review round 2 (all three addressed in `2e797eb`)

- **Lockfile `engines`** — added. Applied that hunk *only*: a full `npm install --package-lock-only` also adds six nested `@tailwindcss/oxide-wasm32-wasi` optional-platform entries, which I verified are **pre-existing drift on untouched `main`** (same 6 appear with no `engines` field at all), so they are not bundled here.
- **`engines` range** — corrected to `vite`'s own `^20.19.0 || >=22.12.0`. `>=20` would have admitted 20.0–20.18, 21.x and 22.0–22.11, which `vite@7` rejects.
- **Guard's failure message** — early `expect(atImportTime).toBeDefined()`, so the broken state now reports `expected undefined to be defined` rather than `Property description must be an object`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,9 @@
 {
   "name": "@offlinecv/web",
   "private": true,
+  "engines": {
+    "node": "^20.19.0 || >=22.12.0"
+  },
   "version": "0.1.0-alpha.1",
   "license": "Apache-2.0",
   "type": "module",
--- a/src/test-setup.test.ts
+++ b/src/test-setup.test.ts
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * Guards the MODULE-SCOPE half of `src/test-setup.ts` (#779).
+ *
+ * The `beforeEach` half needs no guard: every suite that reads `localStorage`
+ * inside a test would fail loudly without it. The module-scope install is the
+ * one that is invisible when it breaks — it only matters to a suite that reads
+ * the global while its own module body is evaluating, which is rare enough that
+ * the gap survived #398 and resurfaced in a file written after it.
+ *
+ * So the capture below is deliberately at module scope, not in a hook. It runs
+ * at the exact moment `letter-egress-ack.test.ts` reads the descriptor it later
+ * restores, and it is the moment that was broken: without the module-scope
+ * install, this file sees Node 22+'s built-in `localStorage` accessor — whose
+ * getter yields no usable `Storage` absent `--localstorage-file` — instead of
+ * the shim. A suite that captures that and restores it afterwards clobbers the
+ * working global for itself and for every test after it in the same worker.
+ *
+ * This file is why a NEW suite cannot reintroduce #779 by copying that
+ * (entirely reasonable) save/restore pattern.
+ */
+
+import { describe, expect, it } from "vitest";
+
+/** Captured at module scope, before any `beforeEach` has run — see above. */
+const atImportTime = Object.getOwnPropertyDescriptor(globalThis, "localStorage");
+
+describe("test-setup provisions localStorage before test modules evaluate (#779)", () => {
+  it("has already installed it by the time a test module body runs", () => {
+    expect(atImportTime).toBeDefined();
+  });
+
+  it("is a data property, not Node's built-in accessor", () => {
+    // The built-in is `{ get, set }`; the shim is `{ value, writable }`. This is
+    // the discriminator — on the built-in, `.value` is undefined and every
+    // `Storage` method is missing.
+    expect(atImportTime?.get).toBeUndefined();
+    expect(atImportTime?.value).toBeDefined();
+  });
+
+  it("exposes a working Storage, not one whose methods are absent", () => {
+    const storage = atImportTime?.value as Storage | undefined;
+    expect(typeof storage?.clear).toBe("function");
+    expect(typeof storage?.setItem).toBe("function");
+    expect(typeof storage?.getItem).toBe("function");
+  });
+
+  it("is configurable, so a suite may swap it and put it back", () => {
+    expect(atImportTime?.configurable).toBe(true);
+  });
+
+  it("survives the capture/swap/restore cycle a hostile-storage suite performs", () => {
+    // Exactly `letter-egress-ack.test.ts`'s shape: install a hostile accessor,
+    // then restore the captured descriptor and use the global again. Before the
+    // module-scope install this threw `localStorage.clear is not a function`,
+    // because what got restored was the built-in accessor rather than the shim.
+    // Guard before the swap, not after: with the module-scope install missing,
+    // `atImportTime` is `undefined` and the restore below throws "Property
+    // description must be an object" — naming this test's own scaffolding
+    // instead of the defect it exists to report, and leaving the hostile
+    // accessor installed for the rest of the case.
+    expect(atImportTime).toBeDefined();
+
+    Object.defineProperty(globalThis, "localStorage", {
+      get: () => undefined,
+      configurable: true,
+    });
+    Object.defineProperty(globalThis, "localStorage", atImportTime!);
+
+    expect(() => globalThis.localStorage.clear()).not.toThrow();
+    globalThis.localStorage.setItem("ocv_probe", "1");
+    expect(globalThis.localStorage.getItem("ocv_probe")).toBe("1");
+  });
+});
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,6 +49,9 @@
         "typescript-eslint": "^8.34.0",
         "vite": "^7.1.2",
         "vitest": "^3.2.4"
+      },
+      "engines": {
+        "node": "^20.19.0 || >=22.12.0"
       }
     },
     "node_modules/@ampproject/remapping": {
--- a/src/test-setup.ts
+++ b/src/test-setup.ts
@@ -4,17 +4,40 @@
 /**
  * Project-wide vitest setup (wired as `test.setupFiles` in `vite.config.ts`).
  *
- * Installs a fresh in-memory `localStorage` shim before EVERY test, globally.
- * The `ocv_*` functional keys touch a `localStorage` that neither the Node env
- * (default) provisions nor Node 22+'s built-in global exposes as a working
- * `Storage` (#398). Doing this once at the workload level — instead of an
- * `import + beforeEach` per file — closes the "new test forgot to install the
- * shim" regression class permanently. Per-test-fresh state means suites still
- * start clean without any `clear()` bookkeeping.
+ * Installs a fresh in-memory `localStorage` shim globally. The `ocv_*`
+ * functional keys touch a `localStorage` that neither the Node env (default)
+ * provisions nor Node 22+'s built-in global exposes as a working `Storage`
+ * (#398) — without `--localstorage-file`, that built-in is an accessor whose
+ * getter yields no usable `Storage`, so `.clear`/`.setItem` are absent.
+ *
+ * It is installed at TWO points, and both are load-bearing:
+ *
+ *   1. **Module scope**, below. A setup file is evaluated before the test
+ *      file's own module body, so this is the only hook that runs early enough
+ *      to be seen by a suite reading `localStorage` at import time. #779: the
+ *      `beforeEach` alone was not enough. `letter-egress-ack.test.ts` captures
+ *      `Object.getOwnPropertyDescriptor(globalThis, "localStorage")` at module
+ *      scope in order to restore it later — a reasonable thing for a suite that
+ *      swaps the global to simulate hostile storage. With only the `beforeEach`
+ *      below, that capture happened first and so captured Node's *built-in*
+ *      accessor; the suite's own `afterEach` then restored that accessor over
+ *      the working shim, and the next `localStorage.clear()` threw. Green on
+ *      CI's Node 20, red on Node 22+ locally — i.e. it read as a broken `main`.
+ *   2. **`beforeEach`**, so each test starts from clean persisted state without
+ *      any `clear()` bookkeeping, and so a test that deliberately swapped the
+ *      global gets a working one back.
+ *
+ * Doing this at the workload level — instead of an `import + beforeEach` per
+ * file — is what keeps a *new* suite from reintroducing the class. #398 fixed
+ * it per-file and it recurred in a file written afterwards; the module-scope
+ * install is what makes the import-time case unreachable too.
  */
 
 import { beforeEach } from "vitest";
 import { installMemoryLocalStorage } from "./hooks/__test-utils__/memory-storage.ts";
+
+// Before any test file's module body — see (1) above.
+installMemoryLocalStorage();
 
 beforeEach(() => {
   installMemoryLocalStorage();


package.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "@offlinecv/web",
   "private": true,
+  "engines": {
```
s-annam, line 4: **Secondary — the lockfile needs regenerating alongside this.**

Lockfile v3 mirrors the root package's `engines` in `packages[""]`, and it is absent there today. Confirmed under the pinned toolchain (Node 20.20.2 / npm 10.8.2):

```
npm install --package-lock-only --ignore-scripts
# → M package-lock.json
#   @@ -49,6 +49,9 @@
#   +      "engines": {
#   +        "node": ">=20"
#   +      }
```

So the first contributor to run `npm install` after this merges gets a spurious dirty tree — and since this repo stages by explicit path, it will either ride along in an unrelated PR or keep reappearing. Regenerate under Node 20 and commit just that hunk.

package.json
```diff
@@ -1,6 +1,9 @@
...
+    "node": ">=20"
```
s-annam: **Secondary — `">=20"` is looser than the toolchain actually supports.**

`vite@7` declares `{ "node": "^20.19.0 || >=22.12.0" }`. So Node 20.0–20.18, all of 21.x, and 22.0–22.11 satisfy this range while the build tool itself does not accept them. The PR body says `>=20` "match[es] `.nvmrc` and CONTRIBUTING" — it matches the docs, but not the dependency tree, and this field is the mechanism AC3 rests on.

```json
"engines": { "node": "^20.19.0 || >=22.12.0" }
```

(or at minimum `>=20.19`) makes the declared floor the real one.

src/test-setup.test.ts
```diff
@@ -0,0 +1,69 @@
...
+  it("survives the capture/swap/restore cycle a hostile-storage suite performs", () => {
+    // Exactly `letter-egress-ack.test.ts`'s shape: install a hostile accessor,
+    // then restore the captured descriptor and use the global again. Before the
+    // module-scope install this threw `localStorage.clear is not a function`,
+    // because what got restored was the built-in accessor rather than the shim.
+    Object.defineProperty(globalThis, "localStorage", {
+      get: () => undefined,
+      configurable: true,
+    });
+    Object.defineProperty(globalThis, "localStorage", atImportTime!);
```
s-annam, line 70: **Nit — this case degrades to an opaque error in exactly the state it guards against.**

When the module-scope install is missing on Node 20, `atImportTime` is `undefined`, and this line throws `TypeError: Property description must be an object: undefined` before reaching the assertion it exists to make. It also leaves the hostile `get: () => undefined` accessor on `globalThis` for the rest of the test (bounded — the next `beforeEach` recovers it).

The failure message then names the test's own scaffolding rather than the defect. An early guard fixes it:

```ts
expect(atImportTime).toBeDefined();
Object.defineProperty(globalThis, "localStorage", atImportTime!);
```

s-annam (dismissed):
Approving — the fix is right, the diagnosis correction is right, and the guard genuinely bites.

**Independently verified** (diff + #779 first, description last):

- **Fail-before reproduces.** Commenting out the module-scope `installMemoryLocalStorage()`: 3 of 5 guard cases fail + all 9 `letter-egress-ack` cases. Restored: 14/14. Matches your claim exactly.
- **The ordering diagnosis is correct, and #398/#779's "shadowing" story is wrong.** Measured on Node 25.9.0: `Object.getOwnPropertyDescriptor(globalThis, "localStorage")` is `{get, set, enumerable, configurable}` with **`configurable: true`**, the getter yields an object, and every `Storage` method on it is `undefined`. That is why `.clear` is missing, and why `defineProperty` can overwrite it at all.
- **#779 is closed across runtimes.** `letter-egress-ack` + the new guard: 14/14 on Node 20.20.2, 22.22.2 and 25.9.0.

**On the "verify exits 0" boxes** — I could not reproduce a green `verify` on *any* Node locally (20 → 5 failures, 22 → 3, 25 → 6), but every failure is in `src/lib/pdf/` behind `Poppins font embed failed … ERR_INVALID_URL` → Helvetica fallback → pdfjs `standardFontDataUrl`; they pass in isolation, and **`origin/main` fails identically on Node 22**. Pre-existing and local to my environment — **not attributable to this PR**, and not evidence against your boxes. Flagging only so the number in the body isn't read as universal.

Gates: fixture-PII, design-system reuse, style tokens — **not applicable** (no fixtures, no components, no styles). fallow: clean. Commit hygiene: one commit, no forbidden trailers.

## Secondary

**1. Lockfile not regenerated for the new `engines` field.** See inline on `package.json`.

**2. `engines.node: ">=20"` is looser than the toolchain's real floor.** See inline on `package.json`.

**3. `src/lib/letter-egress-ack.test.ts:39-46` — the comment now asserts the opposite of the invariant this PR establishes, and the `else` branch is dead.**

It still reads:

> `real` is captured at module-eval time, before any `beforeEach` has run, so it is `undefined` whenever this file is the first in its worker to touch `localStorage`.

After this PR `real` can never be `undefined` — the setup file's module body always runs first, which is the whole fix. So `else delete (globalThis as { localStorage?: unknown }).localStorage;` (line 46) is unreachable. This is the file the PR exists to fix, so it is the highest-value place to correct the record: a future reader following that comment would conclude the module-scope install is optional, which is how #398 recurred.

**4. Description: "after the suite's restore, `globalThis.localStorage` is `undefined`" is contradicted by the trace quoted two lines above it.**

`globalThis.localStorage?.clear is not a function` can only throw when `localStorage` is truthy — optional chaining short-circuits on `undefined` and would throw nothing. That `undefined` state is the **Node 20** path (`real` undefined → `else delete` → `?.` short-circuits → suite passes), not the Node 22+ failure the paragraph is describing. Worth correcting in the section whose stated purpose is fixing a prior wrong diagnosis.

## Nits

**5. `src/test-setup.test.ts:63` — the guard's best case degrades to an opaque error in the regression state.** See inline.

**6. `src/hooks/__test-utils__/memory-storage.ts:10-15` still carries the diagnosis this PR refutes.** It claims the Node 22+ built-in is "a **non-configurable** global" that "shadows jsdom's own `window.localStorage`". Measured `configurable: true` — and it is self-refuting against its own code 50 lines below, since `installMemoryLocalStorage` overwrites the global with `Object.defineProperty`, which would throw on a non-configurable property. The repo now states both versions in two files.

**7. `vite.config.ts:263`** — the pointer comment still says "Install the in-memory localStorage shim before every test". There are two install points now; this is the comment a reader hits first.

**8.** Test plan says "fallow — no issues in **5** changed files"; the diff touches 3.

## Acceptance criteria (#779)

| AC | Status |
| --- | --- |
| `verify` green under Node 20 + current LTS | ✅ for the #779 cause — CI green on 20, `letter-egress-ack` green on 20/22/25 |
| Fix is not per-file | ✅ workload-level install + a mutation-proven guard |
| Unsupported Node names itself | ⚠️ partial — mechanism works, range is wrong at the boundary (Secondary 2) |
| `memory-storage.ts` used | ✅ |

---
Reviewed by: Claude Opus 5 (1M context) — high effort

package.json
Samhit21: Fixed in `2e797eb`. Confirmed the gap — `packages[""]` had no `engines` while `package.json` did.

Applied the engines hunk **only**, not a full regen. A plain `npm install --package-lock-only` under Node 20.20.2 / npm 10.8.2 produces **69** lines, not 3: alongside the engines block it adds six nested `@tailwindcss/oxide-wasm32-wasi/node_modules/*` optional-platform entries (`@emnapi/core`, `@emnapi/runtime`, `@emnapi/wasi-threads`, `@napi-rs/wasm-runtime`, `@tybys/wasm-util`, `tslib`).

Those are **pre-existing drift on `main`, not mine** — I ran the same command against untouched `HEAD` with no `engines` field at all and got the identical 6 entries. So the spurious-dirty-tree problem you describe already exists independently of this PR, and bundling someone else's churn into a test-setup fix would have made the diff harder to read without fixing it.

Worth its own issue — want me to file it, or would you rather it ride along in whatever PR next touches the lockfile?

package.json
Samhit21: Fixed in `2e797eb` — took your exact range, `"^20.19.0 || >=22.12.0"`.

You are right that I matched the docs rather than the tree. Verified against the installed packages:

| Package | `engines.node` |
|---|---|
| `vite@7.3.3` | `^20.19.0 \|\| >=22.12.0` |
| `@vitejs/plugin-react` | `^20.19.0 \|\| >=22.12.0` |
| `vitest` | `^18.0.0 \|\| ^20.0.0 \|\| >=22.0.0` |
| `rollup` | `>=18.0.0` |

So `vite` and `@vitejs/plugin-react` are jointly the binding constraint, and your range is exactly theirs. `>=20` would have admitted 20.0–20.18, all of 21.x and 22.0–22.11 — which matters here specifically because this field *is* the mechanism AC3 rests on, so a floor that is wrong makes the message wrong.

One consequence worth naming: the PR body previously noted the suites pass on Node 18. Under the corrected range Node 18 now (correctly) draws EBADENGINE — `vite` does not support it, so the old text was describing a floor the toolchain never actually had. Body updated.

src/test-setup.test.ts
Samhit21: Fixed in `2e797eb` — added `expect(atImportTime).toBeDefined();` before the swap, as suggested.

Reproduced both halves first: `Object.defineProperty(globalThis, "x", undefined)` does throw `TypeError: Property description must be an object: undefined`, and with the module-scope install removed this case was indeed dying on that line rather than on its assertion.

Before → after, with the install removed:

```
before:  TypeError: Property description must be an object: undefined
after:   → expected undefined to be defined
```

That is the whole point of the nit and you were right to raise it — a guard whose job is to report a specific defect clearly was, in exactly that state, reporting its own scaffolding. The early guard also means the hostile `get: () => undefined` accessor is never installed on the failing path.

Samhit21:
Pushed `2e797eb` — all three findings addressed and threads resolved. CI green (`verify` 3m22s, `fallow`, Pages).

The force-push dismissed the approval, so this is blocked on a re-approve rather than on anything outstanding. Re-requested.

Summary of the round:

- **Lockfile `engines`** — added, that hunk only. A full `--package-lock-only` regen also pulls in six nested `@tailwindcss/oxide-wasm32-wasi` optional-platform entries, which I verified are **pre-existing drift on untouched `main`** (identical 6 appear with no `engines` field at all). Left out deliberately; happy to file it separately if you want it tracked.
- **`engines` range** — corrected to `vite`'s own `^20.19.0 || >=22.12.0`.
- **Guard failure message** — now `expected undefined to be defined` instead of `Property description must be an object`.

package.json
```diff
@@ -1,6 +1,9 @@
 {
   "name": "@offlinecv/web",
   "private": true,
+  "engines": {
+    "node": "^20.19.0 || >=22.12.0"
+  },
```
s-annam, line 6: **Secondary — `engines` advertises Node 22.12+ but nothing in the repo ever runs it.**

`.github/workflows/ci.yml:39` and `deploy-pages.yml:25` both pin `node-version: '20'`; `.nvmrc` is `20`; `promote-release.yml` reads `.nvmrc`. So the single `verify` job is Node 20 only, and there is no `.npmrc` with `engine-strict`. The field is honest documentation of the *toolchain's* range, but it is not a claim CI can keep: after this merges, a future Node-22-only regression in this same class is caught by nobody until someone hits it locally again — which is precisely how #779 was found.

The new test does not close that either; it runs on whatever Node the runner has, which is 20. A one-line matrix leg would make the `engines` range mean something:

```yaml
strategy:
  matrix:
    node: ['20', '22']
```

Reasonable to defer to a follow-up issue — but then the PR body should say `engines` documents the range rather than verifying it.

src/test-setup.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+import { describe, expect, it } from "vitest";
+
+/** Captured at module scope, before any `beforeEach` has run — see above. */
+const atImportTime = Object.getOwnPropertyDescriptor(globalThis, "localStorage");
+
+describe("test-setup provisions localStorage before test modules evaluate (#779)", () => {
+  it("has already installed it by the time a test module body runs", () => {
+    expect(atImportTime).toBeDefined();
+  });
```
s-annam, line 33: **Nit — Two of the five guard cases are Node-20-only.** `src/test-setup.test.ts:31` (`toBeDefined()`) and `:50` (`configurable === true`) both pass on Node 22+ *without* the module-scope install, because Node's own built-in accessor is itself present and configurable — that is the "2 passed" in the mutation run. They do bite on Node 20, where nothing provides the global, so they are not dead; but case 1 is subsumed by case 2's `expect(atImportTime?.value).toBeDefined()` on every platform. Consistent with the body's honest "3 of 5"; flagging only so nobody later reads all five as platform-independent.

s-annam (approved):
The fix is right, the diagnosis that replaced the issue's is right, and the guard bites exactly as claimed — I re-ran the mutation check and got the same 3-of-5 + 9-of-9 split. Approving. Nothing below is a merge blocker; the three Secondary items are stale text and an untested support claim, not behaviour.

**Verified independently** (Node 25.9.0 host, plus nvm 22.22.2 and 20.20.2):

| Check | Result |
|---|---|
| Fail-before, module-scope install commented out | `12 failed / 2 passed` of 14 — 3 of 5 guard cases + all 9 `letter-egress-ack` cases. Matches the body exactly. |
| Baseline `origin/main` under Node 25 | 16 failed (9 `localStorage` + 7 `pdf/` 5 s timeouts) |
| This branch under Node 25 | 5 failed — **all** `pdf/` timeouts, zero `localStorage` |
| This branch under Node 22.22.2 / 20.20.2 | 2 / 3 failed, again all `pdf/` timeouts |
| Those `pdf/` files in isolation | pass on branch *and* `main` → pre-existing 5 s-timeout flake on my box under full-suite concurrency, **not attributable to this PR** |
| `engines` range vs `vite@7.3.3` | byte-identical `^20.19.0 \|\| >=22.12.0`; `@vitejs/plugin-react` agrees |
| Lockfile mirror | `npm install --package-lock-only` produces **no** `engines` delta — only the 6 pre-existing `@tailwindcss/oxide-wasm32-wasi` entries the body says it excluded. The no-dirty-tree claim holds. |
| `npm warn EBADENGINE` | reproduced; names `required` and `current` |
| Node 25 built-in descriptor, jsdom, no setup file | `{ get, set, enumerable, configurable }`, `configurable: true`, getter yields an object whose `clear` is `undefined` |
| `typecheck` / `lint` / `check:fixtures` / `fallow audit --base origin/main` | all green (fallow: "No issues in 4 changed files") |

Acceptance criteria for #779: **all four met.** AC3 lands via `engines` + `EBADENGINE`, and the body is straight about the limitation (install-time warning, no `engine-strict`) rather than overclaiming it — right call.

Gates the diff does not reach, so not run: design-system reuse, style tokens, command-level bugs in skill/script files. Fixture PII was run anyway and is green.

## Secondary

**1. This PR makes `letter-egress-ack.test.ts`'s `afterEach` comment false and its `else` branch dead — and that is the file the issue is about.**

`src/lib/letter-egress-ack.test.ts:39-46` still says:

```ts
// `real` is captured at module-eval time, before any `beforeEach` has run, so
// it is `undefined` whenever this file is the first in its worker to touch
// `localStorage`. Deleting rather than leaving the getter in place matters:
if (real) Object.defineProperty(globalThis, "localStorage", real);
else delete (globalThis as { localStorage?: unknown }).localStorage;
```

After this PR `real` is **never** `undefined` — `test-setup.ts` module scope installs a data property before any test file's body evaluates. Probed on this branch under jsdom: the module-scope descriptor is `{ value, writable, enumerable, configurable }`, every time. The comment's premise is inverted and the `else` arm is unreachable.

This is the exact file whose stale mental model produced two wrong diagnoses (#398, then #779). A comment there contradicting `test-setup.ts`'s new docblock is how a third one gets written. Either say the branch is now defence-in-depth against `test-setup.ts` changing, or drop branch and comment together:

```ts
afterEach(() => {
  // `test-setup.ts` installs the shim at module scope, so `real` is always the
  // shim's data descriptor — never Node's built-in accessor, and never absent
  // (#779). Restoring it un-does whatever `useStorage` left behind.
  Object.defineProperty(globalThis, "localStorage", real!);
  globalThis.localStorage.clear();
});
```

**2. `engines` advertises Node 22.12+ but nothing in the repo ever runs it.** — inline on `package.json:4`.

**3. One factual slip in the PR body's diagnosis: `globalThis.localStorage` is not `undefined` after the restore.**

> Confirmed by printing the descriptor at both moments; after the suite's restore, `globalThis.localStorage` is `undefined`.

It isn't, and the body's own quoted error proves it: `globalThis.localStorage?.clear()` with `localStorage === undefined` short-circuits and throws nothing. `TypeError: … clear is not a function` can only fire when the global **is** defined. Measured directly — jsdom, Node 25, no setup file: the built-in getter returns an object whose `clear` is `undefined`.

The table immediately above it ("whose getter yields no usable `Storage`") is correct; only this sentence contradicts it. Worth fixing because the body is the durable artifact for this class, and a future debugger grepping for `undefined` will look in the wrong place — the optional chaining is exactly what would have saved them if it were true.

## Nits (non-blocking)

**4. Two of the five guard cases are Node-20-only.** — inline on `src/test-setup.test.ts:31`.

**5. `memory-storage.ts:12-14` says the built-in is "a non-configurable global"; it is `configurable: true`.** Measured on Node 25 in both node and jsdom envs. Harmless today — `installMemoryLocalStorage` uses `defineProperty`, which works either way — but this PR now depends on that module at module scope and rewrites the neighbouring docblock to be precise about the same object, so the one remaining wrong sentence is worth sweeping while it is cheap.

**6. Test plan says "no issues in 5 changed files"; it's 4.** `fallow audit --base origin/main` prints `Audit scope: 4 changed files`, and the PR touches four. Probably a leftover from an earlier round.

---

package.json
s-annam: Filed as #808 — not a change to make here.

Splitting it out because the fix is a CI-spend tradeoff, not a defect: a `['20','22']` leg doubles `verify` on every PR in the repo, and that is a call to price deliberately rather than fold into a test-setup fix. #808 records the alternative I'd actually recommend — a Node 22 leg on `main` pushes only, which catches a regression within a day of merge without taxing every PR — plus the branch-protection consequence, since a matrix changes the check name and the required-check setting has to follow or protection silently stops gating.

The second half of my comment does not apply: the body already scopes `engines` correctly. It states that CI runs the pinned Node 20, labels the 20.20.2 / 25.8.1 results as hand-run, and calls `engines` a warning rather than a hard gate. Nothing to correct.

src/test-setup.test.ts
s-annam: No change requested — resolving as recorded.

The observation stands and the code is right either way: the two cases bite on Node 20, and the body's "3 of 5" is already the honest number. This thread is the record; it does not need an issue, and it does not need a commit.

Noting the general rule since it has been coming up: an unresolved thread is not a blocked merge. Only a **Blocking** item blocks. A Secondary or Nit needs an *exit* — fixed here, filed with the link, or declined with a reason — and a reply is the exit, not a fix. Silence is the only thing that actually holds a PR, because an unanswered thread leaves nobody able to tell which of the three it was.

Issue #805, opened 2026-08-12
Eval rubric scores literal markdown as clean — add a residual-markdown criterion so a future run surfaces the class


Samhit21:
## Problem

No rubric criterion asserts that a rewritten bullet is free of markdown syntax, so the eval scores literal `**` as clean output.

#781 shipped exactly that way: **19 of 24** Gemma `terse` bullets in the 2026-08-07 reports carried literal `**` at `records[].rubric.perBullet[].text` — post-`cleanRewriteLine`, i.e. what a downloaded ATS PDF renders as asterisks — and `noPreambleLeak` scored **PASS** on every one. The defect was found by a human reading generated text during review of #777; the aggregate rates showed nothing.

#781's own acceptance criteria closed with:

> - [ ] Consider whether the eval rubric should flag residual markdown, so a future run surfaces this class rather than scoring it PASS.

This issue is that consideration, split out.

## Why it was not bundled into the #781 fix

The fix PR pins the regression at the right layer already: `post-process.test.ts` now covers all six marker shapes deterministically, runs in CI on every PR, and 8 of its 9 new cases fail against the pre-fix ordering. That is a **stronger** guard against `cleanRewriteLine` regressing than a criterion in a harness that only runs on a manual WebGPU session.

What the rubric criterion adds is different and still worth having: it catches a markdown shape **nobody anticipated** — a model emitting something the post-processor was never written for. The unit tests can only cover shapes we thought of.

It was kept separate because it is not a small change:

- `rubric.ts` — new criterion + default + computation
- `runner.ts` — new rate in the aggregate
- `report.ts` — a column in *both* the aggregate and per-cell tables
- `types.ts` — `RubricResult` / `AggregateRow`
- `rubric.test.ts`, `report.test.ts`, `adherence-reporting.test.ts`

and, like the Steering column before it (#608), it makes every already-committed report *predate* the new column. That situation needed its own PR and its own reporting-path proof (#714) last time.

## Scope

Add a `noMarkdownSyntax` (name TBD) criterion to `src/lib/webllm/eval/rubric.ts`.

Open questions to settle in the PR, not silently:

1. **What counts.** Literal `**` is unambiguous. `*`, `_`, backticks and `#` are more likely to be legitimate résumé content (`C*`, `snake_case`, a `#1` ranking) — a naive check would false-positive. Recommend starting with paired `**` only and widening on evidence.
[...]

Issue #778, opened 2026-08-07
Rewrite silently drops concrete numbers — number-preservation guardrail measures 0–80% across every model


Samhit21:
## Problem

`SECTION_REWRITE_SYSTEM_PROMPT` carries the rewrite feature's headline guardrail:

> Preserve every concrete number from the input EXACTLY

The 2026-08-07 eval reports landed by #777 (`tests/fixtures/rewrite/reports/*2026-08-07*`) are the first three-model measurement taken since that guardrail shipped, and they show it failing badly on **every model in the registry**. Salary figures, user counts and percentages are silently dropped from rewritten bullets.

This was surfaced while closing #608 (steering adherence, which measured clean at 100% across all nine cells). It is a **different criterion with a different outcome** and was deliberately left out of that PR rather than chased there.

## The numbers

`numbersPreservedRate` per (model × variant), from the committed reports:

| Model | Baseline | Terse | Examples-led |
| --- | --- | --- | --- |
| Qwen 2.5 (1.5B) — `DEFAULT_MODEL_ID` | 40% | 80% | 20% |
| Gemma 2 (2B) | 20% | 40% | **0%** |
| Llama 3.2 (3B) | **0%** | 20% | 20% |

**8 of 9 cells are at or below 50%.** The best cell in the grid is Qwen/terse at 80%; the default shipped configuration (Qwen/baseline) is 40%.

## What is actually being dropped

From `records[].rubric.droppedNumbers` in the same files — these are not rounding differences, they are whole figures vanishing:

| Model / variant | Fixture | Dropped |
| --- | --- | --- |
| Gemma / baseline | `numeric-growth-pm` | `120K` `1.8M` `14%` `$4.2M` `4.5` `$250K` `11.5M` `38K` |
| Gemma / terse | `numeric-growth-pm` | `120K` `1.8M` `2022` `2024` `$4.2M` `$250K` `11.5M` `38K` |
| Gemma / any | `strong-backend-engineer` | `5` `$180K` |
| Gemma / any | `redundant-support-lead` | `$85K` |
| Llama / baseline | `strong-backend-engineer` | `$180K` |
| Llama / terse | `numeric-growth-pm` | `120K` `2022` `14%` `4.5` |
| Qwen / baseline | `strong-backend-engineer` | `2024` `5` |
| Qwen / examples-led | `numeric-growth-pm` | `2022` `2024` `11.5M` `38K` |

[...]

PR #806, opened 2026-08-12
fix(webllm): stop a list marker shielding the leading bold from cleanRewriteLine (#781)

### LINKED ISSUE #781: Rewrite ships literal markdown bold when the model prefixes a list marker (`- **Led** …` survives cleanRewriteLine)
## Summary

`cleanRewriteLine` strips a leading `**Verb**` markdown bold, but only when the
line has no list marker in front of it. When the model emits `- **Led** the
migration…`, the marker shields the bold from the anchored pattern and the
asterisks survive into the accepted rewrite — and from there into the
downloaded ATS PDF as literal `**`.

This is the shared cleanup helper on the product path
(`src/lib/webllm/rewrite-section.ts:220`), not an eval-only concern.

## Reproduction

```
cleanRewriteLine("**Led** the migration of billing.")   → "Led the migration of billing."   ✅
cleanRewriteLine("- **Led** the migration of billing.") → "**Led** the migration of billing." ❌
cleanRewriteLine("* **Led** the migration.")            → "**Led** the migration."           ❌
cleanRewriteLine("1. **Led** the migration.")           → "**Led** the migration."           ❌
cleanRewriteLine("• **Led** the migration.")            → "**Led** the migration."           ❌
```

## Evidence it happens in practice

The committed eval reports carry it. In
`tests/fixtures/rewrite/reports/eval-rewrite-gemma-2-2b-it-*-2026-08-07*.json`,
**19 of the 24 `terse`-variant bullets** contain literal `**` at
`records[].rubric.perBullet[].text` — which is *post*-`cleanRewriteLine`
output:

```
"**Led** the migration of the billing platform to a new payments provider…"
"**Developed** and **implemented** an internal design system…"
```

Model/variant breakdown across the three 2026-08-07 reports:

| Model | baseline | terse | examples-led |
| --- | --- | --- | --- |
| Qwen 2.5 (1.5B) | 0/24 | 0/24 | 0/24 |
| Gemma 2 (2B) | 0/24 | **19/24** | 0/23 |
| Llama 3.2 (3B) | 0/24 | 0/24 | 0/24 |

It is model- and variant-specific — Gemma under the terse prompt is the only
cell that emits marker + bold together — but it is a shipped model and a
shipped prompt variant.

No rubric criterion catches it: `noPreambleLeak` scores PASS on every one of
those bullets, so the eval reports it as clean output.

## Root cause

`src/lib/webllm/post-process.ts`, inside `cleanRewriteLine`:

```ts
const LEADING_BOLD_WORD_PATTERN = /^\*\*([A-Za-z][\w-]*)\*\*\s+/;   // line 75

const withoutLeadingBold = withoutPrefix.replace(LEADING_BOLD_WORD_PATTERN, "$1 ");   // ~line 93

// … paired emphasis strip …

const withoutBullet = withoutEmphasis.replace(/^(?:\d+[.)]\s*|[•\-]\s*|\*\s+)/, "");  // ~line 110
```

`LEADING_BOLD_WORD_PATTERN` is anchored at `^`, and the list-marker strip runs
*after* it, so any marker in front of the bold defeats the match.

## Why the fix is not a one-line reorder

The current order is deliberate and documented in the file: the paired
bold/italic strip must run **before** list-marker stripping, or the leading `*`
of an italicized line (`*Foo.*`) is misread as a bullet glyph and the trailing
`*` survives. Any fix has to keep that case working.

Likely shapes:

- Strip the list marker first, then run both emphasis strips, and handle the
  `*Foo.*` italics case by requiring the `*` marker to be followed by
  whitespace (the marker pattern already does: `\*\s+`) — so `*Foo.*` never
  matches the marker branch and the reorder is safe. Needs verifying against
  the existing post-process tests.
- Or loop the marker/emphasis strips until the line stops changing.
- Or make `LEADING_BOLD_WORD_PATTERN` tolerate an optional leading marker.

Whichever shape, the mid-line case (`**Developed** and **implemented** …`)
needs a decision too — the current pattern is single-word and leading-only, so
it would not catch the second bold even after a reorder.

## Acceptance criteria

- [ ] `cleanRewriteLine` returns text free of literal `**` for all five marker
      shapes above (`-`, `*`, `•`, `1.`, `1)`), plus the bare case.
- [ ] The `*Foo.*` italics case still round-trips to `Foo.` — regression test
      kept green.
- [ ] Mid-line bold: either stripped, or explicitly documented as out of scope
      with the reason.
- [ ] A unit test in `src/lib/webllm/post-process.test.ts` covering marker +
      bold, that fails before the fix.
- [ ] Consider whether the eval rubric should flag residual markdown, so a
      future run surfaces this class rather than scoring it PASS.

## Provenance

Found during review of #777 while auditing the committed eval reports; the
report README (`tests/fixtures/rewrite/reports/README.md`) now names this
anomaly and links here.

### LINKED ISSUE #782: Rewritten bullets ship literal markdown bold — cleanRewriteLine strips **Verb** before the list marker, so a marker shields it
## Problem

`cleanRewriteLine` fails to strip a leading `**Verb**` when the model also emits a list marker, so literal markdown asterisks survive into rewritten bullets — and into the downloaded ATS PDF, where they render as `**Led** the migration…`.

This is a **post-process ordering bug**, not a model problem. Found in the 2026-08-07 eval reports landed by #777, where **19 of 20 Gemma "terse" bullets** carry literal `**`:

```
"**Led** the migration of the billing platform to a new payments provider…"
"**Developed** and **implemented** an internal design system…"
```

Those strings are `records[].rubric.perBullet[].text` — *post*-`cleanRewriteLine` output. Whatever reaches that field is what reaches the user.

## Root cause

`src/lib/webllm/post-process.ts` strips the leading bold **before** the list marker:

- `:93` — `LEADING_BOLD_WORD_PATTERN` is anchored at start-of-line and strips `**Verb**`.
- `:110` — the list-marker strip (`^(?:\d+[.)]\s*|[•\-]\s*|\*\s+)`) runs **afterwards**.

So any marker shields the bold from the anchored pattern. Confirmed by direct execution:

```
"**Led** the migration…"    → "Led the migration…"      ✅
"- **Led** the migration…"  → "**Led** the migration…"  ❌
"* **Led** the migration…"  → "**Led** the migration…"  ❌
"• **Led** the migration…"  → "**Led** the migration…"  ❌
"1. **Led** the migration…" → "**Led** the migration…"  ❌
```

Only the bare form is handled. All four marker forms leak.

## No criterion catches it

`noPreambleLeak` scores **PASS** on every affected record:

| Fixture (Gemma / terse) | Bullets with `**` | `noPreambleLeak` |
| --- | --- | --- |
| `weak-marketing-generalist` | 5 | ✅ PASS |
| `strong-backend-engineer` | 5 | ✅ PASS |
| `numeric-growth-pm` | 5 | ✅ PASS |
| `steering-forbidden-word` | 4 | ✅ PASS |

So the eval harness looked green on the criterion nearest to this and shipped the defect anyway. That is the second half of the issue: there is no rubric criterion asserting the output is free of markdown syntax.

Worth noting the interaction with #608: `steering-forbidden-word` is in that list. Steering adherence still scored a correct 100% — `**Led**` genuinely is not the forbidden word — but the same bullets were carrying literal asterisks the whole time.

## Reproduce

No WebGPU needed. Against the committed artifacts:

```sh
python3 -c "
import json,glob
for f in sorted(glob.glob('tests/fixtures/rewrite/reports/*2026-08-07*.json')):
    d=json.load(open(f))
    for x in d['records']:
        for b in x['rubric'].get('perBullet') or []:
            if '**' in b['text']: print(d['modelIds'][0], x['variantId'], b['text'][:80])
"
```

Or directly against the function:

```ts
cleanRewriteLine("- **Led** the migration.")  // → "**Led** the migration."
```

## Fix direction

Reorder, or make the bold strip marker-tolerant. Reordering is the smaller change but needs care: the comment at `:98` records that emphasis stripping must precede list-marker stripping, because the leading `*` of an italicised line (`*Foo.*`) would otherwise be misread as a bullet glyph. So the fix is **not** simply moving the marker strip earlier — it likely means allowing an optional marker inside `LEADING_BOLD_WORD_PATTERN`, or running the bold strip a second time after the marker strip.

Whichever is chosen, the existing `*Foo.*` behaviour must stay covered — that ordering was deliberate.

## Acceptance criteria

- [ ] `cleanRewriteLine` strips the leading bold for all of `-`, `*`, `•` and `1.` markers, not just the bare form.
- [ ] The italics case the current ordering protects (`*Foo.*` → `Foo.`) still passes.
- [ ] A unit test covers each marker form; it fails against today's implementation.
- [ ] No rewritten bullet reaching the UI or the exported PDF can contain `**`. Consider a rubric criterion (`noMarkdownSyntax`) so the eval catches a regression instead of a reader noticing it in a committed report.
- [ ] A re-run shows Gemma / terse free of literal `**`.

## Notes

- Surfaced in review of #777 by reading `perBullet[].text` in the committed reports rather than the aggregate rates — the aggregates showed nothing.
- Sibling issues from the same artifact: **#778** (number preservation, 0–80%) and **#779** (`verify` red on Node 22+). This one is filed separately because the root cause and fix are specific and unrelated to either.
- Not a regression from #777 — that PR only committed the reports that make it visible.

Samhit21:
## Summary

A model emitting `- **Led** the migration…` shipped literal markdown bold through `cleanRewriteLine` into the accepted rewrite, and from there into the downloaded ATS PDF as asterisks. `render-ats-pdf.ts` draws real bold from its own type scale and never interprets markdown, so a surviving `**` is always garbage — never formatting.

`LEADING_BOLD_WORD_PATTERN` is anchored at `^` and the list-marker strip ran *after* it, so any marker in front of the bold hid it.

It shipped. **19 of 24** Gemma `terse` bullets in the committed 2026-08-07 eval reports carry literal `**` at `records[].rubric.perBullet[].text` — post-`cleanRewriteLine` output. `noPreambleLeak` scores **PASS** on every one, so nothing flagged it.

Closes #781
Closes #782

## Why this is a fixed point, not a reorder

#781 correctly warned the obvious swap is unsafe, and it is — the three strips shield each other in **both** directions:

| Constraint | Why |
|---|---|
| emphasis **before** marker | else the leading `*` of `*Foo.*` is read as a bullet glyph and the trailing `*` survives |
| marker **before** leading-bold | else `- **Led** …` hides the bold from a `^`-anchored pattern — this bug |

Those cannot both hold in one pass. So the three run in a loop until the line stops changing: each pass peels whatever is now outermost, and a marker uncovered on one pass is consumed by the next. Every branch only removes characters, so the line strictly shrinks and the loop always converges; the pass bound is belt-and-braces.

I checked the swap-only fix would have regressed `* Foo.*` (today `"Foo."`, after a naive reorder `"Foo.*"`). The loop preserves it.

## Verified behaviour

| Input | Output |
|---|---|
| `- **Led** the migration.` | `Led the migration.` |
| `* **Led** the migration.` | `Led the migration.` |
| `• **Led** the migration.` | `Led the migration.` |
| `1. **Led** the migration.` | `Led the migration.` |
| `1) **Led** the migration.` | `Led the migration.` |
| `-**Led** the migration.` | `Led the migration.` |
| `*Foo.*` | `Foo.` *(invariant preserved)* |
| `* Foo.*` | `Foo.` *(invariant preserved)* |
| `**Streamlined the** checkout` | unchanged *(multi-word, by design)* |

## Mid-line bold is deliberately out of scope

#781 asked for a decision rather than a default, so: **not stripped**, and the reason is in the code.

Two existing tests pin it as intentional — `does NOT strip emphasis mid-line` and `does NOT strip mid-bullet bold emphasis` — on the reasoning that a bold inside a sentence is authored emphasis, not a model tic. Reversing a tested contract is a separate decision from fixing the marker interaction, and the data says it is the rare tail: of the 19 affected bullets, **18 carried a leading bold only**; exactly one had an additional inline pair.

## The rubric half is split out as #805

#781's last criterion — *"Consider whether the eval rubric should flag residual markdown"* — is filed as **#805** rather than bundled.

The regression guard for *this* defect is already at the better layer: `post-process.test.ts` covers all six marker shapes deterministically and runs in CI on every PR, whereas the eval is a manual WebGPU session. What a rubric criterion adds is different — catching a markdown shape nobody anticipated — and it is not small: it touches `rubric.ts`, `runner.ts`, `report.ts`, `types.ts` and three test files, adds a column to both report tables, and makes every committed report predate it. That is exactly the shape that needed its own PR and its own reporting-path proof (#714) when the Steering column landed in #608.

## Review focus

- `post-process.ts` — the loop is the whole change. If you prefer an explicit two-phase order (marker → emphasis → bold) over iterating, say so; I chose the loop because it does not require me to be right about which shapes nest, and `* Foo.*` showed that intuition is not reliable here.
- The `MAX_PREFIX_PASSES = 4` bound is defensive, not load-bearing — the loop converges on its own. Happy to drop it if you consider it noise.
- Mid-line bold: the call above is arguable. If you would rather clear all `**` on the way to a plain-text PDF, that is a coherent position — it just means retiring two tests deliberately, which I did not want to do silently.

## Test plan

- [x] `npm run verify` on Node 20.20.2 — exit 0, **5457 passed** / 10 skipped
- [x] `post-process.test.ts` 31 passed (22 pre-existing + 9 new)
- [x] **Mutation-checked**: restoring the pre-fix ordering fails **8 of the 9** new cases. The ninth (`*Foo.*` italics) correctly passes both ways — it is the invariant the loop protects, not the fix.
- [x] All six marker shapes assert both exact output *and* the property that matters (`not.toContain("*")`)
- [x] `fallow` — no issues in changed files

## Not in this PR

- **The eval rubric criterion** — #805, with the reasoning above.
- **Mid-line / multi-word bold** — deliberate, documented in code and in #805's context.
- **Re-running the eval to confirm the reports go clean.** The fix is proven at the unit layer; a fresh WebGPU run would confirm it end-to-end but needs a maintainer machine. Worth folding into whatever run happens next rather than commissioning one for this.

## Provenance

---

## Review round 2 — all four addressed in `cfb7b2c`

**Blocking: the loop mangled decimals.** The marker strip was not idempotent — with `\s*` after the numbered alternative, pass 1 removed `- ` from `- 3.5x revenue growth` and pass 2 read the uncovered `3.` as a numbered marker, yielding `5x revenue growth`. Silent numeric corruption on the product path, which downstream would have surfaced as a `checkNumbersPreserved` warning blaming the model for damage post-processing did. Numbered branch now requires `\s+`. Bare `3.5x revenue growth` — which `main` also mangles — is fixed as a side effect.

**`MAX_PREFIX_PASSES` was a truncation point, not belt-and-braces.** At 4 it silently gave up on `1. - • - **Led** stuff.` and left the literal `**`. Now a `while` to true convergence, bound raised to 32 and documented as a runaway guard only. Constant and a new named `LIST_MARKER_PATTERN` moved to module scope per house style; unused counter removed.

**The comment proved termination when idempotence was the risk.** Reframed: each strip must be safe to re-apply to a line it has already transformed, with the `- 3.5x` path as the worked example.

**The test table was blind to over-firing.** Six cases sharing one body could only prove a marker no longer *shields* the bold. Added decimal-uncovering cases, a digit-preservation property across all five markers, a genuine-numbered-marker case, and a convergence case.

Updated mutation results — 40 tests in `post-process.test.ts`:

| Mutation | Fails |
|---|---|
| revert `\s+` → `\s*` on the numbered branch | 7 |
| cap `MAX_PREFIX_PASSES` back to 4 | 1 |
| restore the pre-fix strip ordering | 8 |

`npm run verify` on Node 20.20.2 — exit 0, **5466 passed**.

### DIFF
--- a/src/lib/webllm/post-process.test.ts
+++ b/src/lib/webllm/post-process.test.ts
@@ -168,4 +168,106 @@ describe("cleanRewriteLine", () => {
       ).toBe("Led migration of the **order-processing** pipeline.");
     });
   });
+
+  // ── A list marker must not shield the leading bold (#781) ──────────────────
+  //
+  // `LEADING_BOLD_WORD_PATTERN` is anchored at `^`, and the marker strip used
+  // to run after it, so anything the model put in front of the bold hid it.
+  // This shipped: 19 of 24 Gemma `terse` bullets in the 2026-08-07 eval reports
+  // carry literal `**` at `perBullet[].text` — post-`cleanRewriteLine` output,
+  // i.e. what a downloaded ATS PDF renders as asterisks. That renderer draws
+  // real bold from its own type scale and never interprets markdown, so a
+  // surviving `**` is always garbage.
+  //
+  // Every case below returns the marker-shielded input unchanged against the
+  // pre-fix implementation.
+  describe("list marker in front of a leading bold (#781)", () => {
+    it.each([
+      ["- **Led** the migration of billing.", "dash"],
+      ["* **Led** the migration of billing.", "asterisk"],
+      ["• **Led** the migration of billing.", "bullet glyph"],
+      ["1. **Led** the migration of billing.", "numbered"],
+      ["1) **Led** the migration of billing.", "paren-numbered"],
+      ["-**Led** the migration of billing.", "tight dash, no space"],
+    ])("strips both, leaving no markdown: %s (%s)", (input) => {
+      expect(cleanRewriteLine(input)).toBe("Led the migration of billing.");
+    });
+
+    it("leaves no literal asterisks for any marker shape", () => {
+      // The property the ATS PDF actually depends on, asserted directly rather
+      // than inferred from the equality cases above.
+      for (const marker of ["-", "*", "•", "1.", "1)"]) {
+        expect(cleanRewriteLine(`${marker} **Shipped** the thing.`)).not.toContain(
+          "*",
+        );
+      }
+    });
+
+    it("still reads `*Foo.*` as italics, not as a bullet glyph", () => {
+      // The ordering constraint that made this a loop rather than a swap: the
+      // emphasis strip has to keep beating the marker strip for this shape, or
+      // the leading `*` is eaten as a marker and the trailing one survives.
+      expect(cleanRewriteLine("*Foo.*")).toBe("Foo.");
+      expect(cleanRewriteLine("* Foo.*")).toBe("Foo.");
+    });
+
+    // ── The complementary hazard: a strip OVER-firing on what it uncovered ──
+    //
+    // Every case above shares one body, so they only prove a marker no longer
+    // *shields* the bold. They are blind to a strip mis-firing on the text the
+    // previous pass exposed — which is exactly how the loop first shipped
+    // `- 3.5x revenue growth` → `5x revenue growth`: pass 1 removed `- `, and
+    // pass 2 read the uncovered `3.` as a numbered marker because that branch
+    // allowed zero-or-more trailing space. Silent numeric corruption, on the
+    // path that renders the ATS PDF (#781 review).
+    it.each([
+      ["- 3.5x revenue growth in Q3.", "3.5x revenue growth in Q3."],
+      ["- 2.5M users onboarded.", "2.5M users onboarded."],
+      ["* 1.5x faster builds.", "1.5x faster builds."],
+      ["- 10.5 million records migrated.", "10.5 million records migrated."],
+      ["• 4.2% conversion lift.", "4.2% conversion lift."],
+      // Not marker-prefixed at all — the same branch mangled this on its own.
+      ["3.5x revenue growth.", "3.5x revenue growth."],
+    ])("does not eat a decimal a marker strip uncovers: %s", (input, want) => {
+      expect(cleanRewriteLine(input)).toBe(want);
+    });
+
+    it("preserves every digit of a leading decimal, for any marker", () => {
+      // The number-preservation counterpart to the `not.toContain("*")`
+      // property above. Asserted as a property so a new marker branch has to
+      // satisfy it too, not just the six literals enumerated here.
+      for (const marker of ["-", "*", "•", "1.", "1)"]) {
+        expect(cleanRewriteLine(`${marker} 3.5x revenue growth.`)).toContain(
+          "3.5",
+        );
+      }
+    });
+
+    it("still strips a genuine numbered marker", () => {
+      // The `\s+` requirement must not cost the case the marker strip exists
+      // for. Nothing in the corpus writes a tight `1.Foo`.
+      expect(cleanRewriteLine("1. Shipped the thing.")).toBe(
+        "Shipped the thing.",
+      );
+      expect(cleanRewriteLine("1) Shipped the thing.")).toBe(
+        "Shipped the thing.",
+      );
+    });
+
+    it("converges on stacked markers instead of truncating", () => {
+      // The loop runs to a fixed point rather than a fixed pass count. An
+      // earlier revision capped it at 4 and gave up here, leaving the literal
+      // `**` this function exists to remove.
+      expect(cleanRewriteLine("1. - • - **Led** stuff.")).toBe("Led stuff.");
+    });
+
+    it("leaves mid-line bold alone even when a marker is stripped", () => {
+      // Deliberate, and the one shape this fix does not clear — see the
+      // fixed-point comment in post-process.ts. 18 of the 19 affected eval
+      // bullets carried a leading bold only, so this is the rare tail.
+      expect(
+        cleanRewriteLine("- **Developed** and **implemented** a design system."),
+      ).toBe("Developed and **implemented** a design system.");
+    });
+  });
 });
--- a/src/lib/webllm/post-process.ts
+++ b/src/lib/webllm/post-process.ts
@@ -74,6 +74,33 @@ const CHAT_OPENER_PATTERN = /^here (?:are|is) (?:the )?rewritten\b/i;
  */
 const LEADING_BOLD_WORD_PATTERN = /^\*\*([A-Za-z][\w-]*)\*\*\s+/;
 
+/**
+ * A leading list marker: `1.` / `1)`, `•`, `-`, or `*`.
+ *
+ * The whitespace rules differ per branch, and each one is load-bearing:
+ *
+ *   - **Numbered requires `\s+`.** With `\s*` the alternative matches a bare
+ *     decimal — `3.` in `3.5x revenue growth` — because zero-or-more trailing
+ *     space is satisfied by the `5`. That is harmless while this runs once on
+ *     raw model output (a decimal is rarely line-initial there), but
+ *     `cleanRewriteLine` re-applies it to lines it has already stripped, so
+ *     `- 3.5x revenue growth` became `5x revenue growth`. Silent numeric
+ *     corruption on the product path, in a bullet the rewrite prompt actively
+ *     asks the model to quantify. Nothing needs a tight `1.Foo`.
+ *   - **`-` and `•` allow `\s*`**, so a tight `-Shipped X` still normalizes.
+ *   - **`*` requires `\s+`**, because `*X*` is italics and is handled by the
+ *     paired-emphasis strip instead.
+ */
+const LIST_MARKER_PATTERN = /^(?:\d+[.)]\s+|[•\-]\s*|\*\s+)/;
+
+/**
+ * Runaway guard for the markdown-prefix loop in `cleanRewriteLine` — not a
+ * budget for expected nesting. The loop terminates on its own (every strip
+ * only removes characters); this only bounds a hypothetical future strip that
+ * grows the line. Set well above any real prefix stack so it never truncates.
+ */
+const MAX_PREFIX_PASSES = 32;
+
 export function cleanRewriteLine(line: string): string {
   const trimmed = line.trim();
   if (!trimmed) return "";
@@ -87,30 +114,73 @@ export function cleanRewriteLine(line: string): string {
   // any trailing content the model attached to it.
   const withoutPrefix = trimmed.replace(/^rewritten:\s*/i, "");
 
-  // Strip a leading `**Verb**` markdown bold (single-word capture). This
-  // runs BEFORE the whole-line emphasis strip because that one requires
-  // closing `**` at end-of-line and so doesn't match the inline shape.
-  const withoutLeadingBold = withoutPrefix.replace(
-    LEADING_BOLD_WORD_PATTERN,
-    "$1 ",
-  );
+  // Markdown-prefix stripping, run to a fixed point.
+  //
+  // The three strips below shield each other in BOTH directions, which is why
+  // no single ordering works (#781):
+  //
+  //   - The emphasis strip must precede the marker strip, or the leading `*`
+  //     of an italicized line (`*Foo.*`) is read as a bullet glyph and the
+  //     trailing `*` survives.
+  //   - The marker strip must precede the leading-bold strip, or a marker in
+  //     front of the bold (`- **Led** the migration`) hides it from
+  //     LEADING_BOLD_WORD_PATTERN, which is anchored at `^`. That shipped:
+  //     19 of 24 Gemma `terse` bullets in the 2026-08-07 eval reports carried
+  //     literal `**` through to `perBullet[].text`, i.e. through to what a
+  //     downloaded ATS PDF would render as asterisks. That renderer draws real
+  //     bold from its own type scale and never interprets markdown, so a
+  //     surviving `**` is always garbage, never formatting.
+  //
+  // Looping resolves it without having to pick: each pass peels whatever is
+  // now outermost, and a marker uncovered on one pass is consumed on the next.
+  //
+  // ⚠️ The property each strip must hold is IDEMPOTENCE, not termination.
+  // Termination is trivially guaranteed — every branch only removes characters,
+  // so the line strictly shrinks — and it was never the risk. The real hazard is
+  // that a strip safe to apply once to MODEL OUTPUT may not be safe to re-apply
+  // to its OWN output, because the loop feeds it a line it has already
+  // transformed. LIST_MARKER_PATTERN is where that bit: with `\s*` after the
+  // numbered alternative, pass 1 stripped `- ` from `- 3.5x revenue growth` and
+  // pass 2 read the uncovered `3.` as a numbered marker, yielding
+  // `5x revenue growth`. Requiring `\s+` there is what makes it re-appliable.
+  // Any strip added to this loop must be checked the same way.
+  //
+  // Mid-line bold (`Developed and **implemented** …`) is deliberately NOT
+  // handled here. Two tests pin that as intentional — "does NOT strip emphasis
+  // mid-line" and "does NOT strip mid-bullet bold emphasis" — on the reasoning
+  // that a bold inside a sentence is authored emphasis rather than a model tic.
+  // It is also rare in practice: of the 19 affected eval bullets, 18 carried a
+  // leading bold only. Reversing that contract is a separate decision from
+  // fixing the marker interaction, so it is left to #781's follow-up.
+  let body = withoutPrefix;
+  let before = "";
+  let passes = 0;
+  while (body !== before) {
+    before = body;
 
-  // Strip paired bold/italic markdown delimiters BEFORE list-marker stripping
-  // — otherwise the leading `*` of an italicized line (`*Foo.*`) is misread
-  // as a bullet glyph and the trailing `*` survives. The pattern is paired
-  // (start AND end) so genuine mid-line emphasis is preserved.
-  const withoutEmphasis = withoutLeadingBold
-    .replace(/^\*\*(.+)\*\*$/s, "$1")
-    .replace(/^\*(.+)\*$/s, "$1")
-    .replace(/^_(.+)_$/s, "$1");
+    // Leading `**Verb**` bold (single-word capture) — see the pattern's own
+    // docblock for why multi-word bolds are left alone.
+    body = body.replace(LEADING_BOLD_WORD_PATTERN, "$1 ");
 
-  // Strip a leading list marker. `-` and `•` allow zero-or-more spaces (a
-  // tight `-Shipped X` should still normalize), but `*` requires at least
-  // one trailing space — `*X*` is italics and was already handled above.
-  const withoutBullet = withoutEmphasis.replace(
-    /^(?:\d+[.)]\s*|[•\-]\s*|\*\s+)/,
-    "",
-  );
+    // Paired bold/italic wrapping the WHOLE line. Paired (start AND end) so
+    // genuine mid-line emphasis is preserved.
+    body = body
+      .replace(/^\*\*(.+)\*\*$/s, "$1")
+      .replace(/^\*(.+)\*$/s, "$1")
+      .replace(/^_(.+)_$/s, "$1");
+
+    body = body.replace(LIST_MARKER_PATTERN, "");
+
+    // Runaway guard only. The loop exits on its own the moment a pass changes
+    // nothing, and the strictly-shrinking invariant means that always happens;
+    // this exists so a future strip that somehow grows the line cannot hang the
+    // worker. It is NOT a truncation point tuned to expected input — an earlier
+    // revision capped this at 4, which silently gave up on
+    // `1. - • - **Led** stuff.` and left the literal `**` this function exists
+    // to remove.
+    if (++passes >= MAX_PREFIX_PASSES) break;
+  }
+  const withoutBullet = body;
 
   // Strip surrounding quotes: straight (" ' `) plus smart double (“ ”) and
   // smart single (‘ ’).


src/lib/webllm/post-process.ts
```diff
@@ -87,30 +87,59 @@ export function cleanRewriteLine(line: string): string {
...
+    // genuine mid-line emphasis is preserved.
+    body = body
+      .replace(/^\*\*(.+)\*\*$/s, "$1")
+      .replace(/^\*(.+)\*$/s, "$1")
+      .replace(/^_(.+)_$/s, "$1");
+
+    // Leading list marker. `-` and `•` allow zero-or-more spaces (a tight
+    // `-Shipped X` should still normalize), but `*` requires at least one
+    // trailing space — `*X*` is italics and is handled by the strip above.
+    body = body.replace(/^(?:\d+[.)]\s*|[•\-]\s*|\*\s+)/, "");
```
s-annam: **Blocking — the marker strip is not idempotent, so the loop corrupts numbers.**

This branch runs *this* line up to four times. On pass 1 it strips the bullet glyph; on pass 2 it sees a line that now **starts with a decimal** and the `\d+[.)]\s*` alternative eats the integer part as if it were a numbered marker — `\s*` is zero-or-more, so `3.` matches with nothing after it.

Verified against `main` (old = pre-fix `cleanRewriteLine`, new = this branch):

```
"- 3.5x revenue growth in Q3."   old="3.5x revenue growth in Q3."   new="5x revenue growth in Q3."
"- 2.5M users onboarded."        old="2.5M users onboarded."        new="5M users onboarded."
"* 1.5x faster builds."          old="1.5x faster builds."          new="5x faster builds."
"4.2% conversion lift (bullet)"  old="4.2% conversion lift."        new="2% conversion lift."
"- 10.5 million records..."      old="10.5 million records..."      new="5 million records..."
```

Why this is Blocking rather than an edge case: a list marker in front of the bullet is the shape this PR exists to handle, and the rewrite prompt actively pushes for quantified bullets. `1.8M`, `11.5M` and `4.5` are already in this repo's own rewrite corpus under `tests/fixtures/rewrite/reports/`. Downstream, `checkNumbersPreserved` (`rewrite-section.ts:223`) only sets an advisory boolean — the mangled string still reaches the UI and `render-ats-pdf.ts`, and the resulting "numbers changed" signal blames the model for damage post-processing did. That is the failure mode #778 is already tracking.

**Fix — one character, verified.** Require whitespace after a *numbered* marker:

```diff
-    body = body.replace(/^(?:\d+[.)]\s*|[•\-]\s*|\*\s+)/, "");
+    body = body.replace(/^(?:\d+[.)]\s+|[•\-]\s*|\*\s+)/, "");
```

I ran this: all 31 tests in `post-process.test.ts` stay green. Nothing depends on a tight numbered marker (`1.Foo`), and the tight-dash case your table pins (`-**Led** ...`) is the `[•\-]\s*` branch, untouched. After the change `1. **Led** ...` and `1) **Led** ...` still yield `Led ...`, and `- 3.5x revenue growth in Q3.` round-trips intact. Bonus: bare `3.5x revenue growth`, which `main` also mangles today, comes out right too.

Please add a decimal-leading case to the `it.each` table so this stays fixed.

src/lib/webllm/post-process.ts
```diff
@@ -87,30 +87,59 @@ export function cleanRewriteLine(line: string): string {
...
+  // and the loop always converges; the bound is belt-and-braces.
+  //
+  // Mid-line bold (`Developed and **implemented** …`) is deliberately NOT
+  // handled here. Two tests pin that as intentional — "does NOT strip emphasis
+  // mid-line" and "does NOT strip mid-bullet bold emphasis" — on the reasoning
+  // that a bold inside a sentence is authored emphasis rather than a model tic.
+  // It is also rare in practice: of the 19 affected eval bullets, 18 carried a
+  // leading bold only. Reversing that contract is a separate decision from
+  // fixing the marker interaction, so it is left to #781's follow-up.
+  const MAX_PREFIX_PASSES = 4;
```
s-annam: **Secondary + Nit — `MAX_PREFIX_PASSES` is load-bearing, and the comment above says it isn't.**

You asked in `## Review focus` whether this bound is noise. It isn't — but it also isn't belt-and-braces. When it runs out, the loop **silently truncates instead of converging**, leaving exactly the symptom #781 was filed for:

```
cleanRewriteLine("1. - • - **Led** stuff.")  ===  "**Led** stuff."   // literal ** survives
cleanRewriteLine("- • - **Led** stuff.")     ===  "Led stuff."       // 3 markers: needs exactly 4 passes, just fits
```

Four stacked markers need five passes and get cut off. I would not call that input realistic on its own — the point is that the comment's claim ("the bound is belt-and-braces") is falsifiable and false, so a future reader trusting it could lower the number and quietly reintroduce the bug. Either loop to true convergence (the strictly-shrinking invariant already guarantees termination, so a `while` with a large sanity ceiling is safe), or keep `4` and say what it actually is: a truncation point, chosen because no realistic model output stacks more than two prefixes.

**Nit, same line:** house style keeps named constants at module scope (`score.ts` is the exemplar) — this would sit naturally beside `LEADING_BOLD_WORD_PATTERN`. The `pass` counter is also unused in the body; `while` plus the existing `before === body` break reads closer to the intent.

src/lib/webllm/post-process.ts
```diff
@@ -87,30 +87,59 @@ export function cleanRewriteLine(line: string): string {
...
+  //     LEADING_BOLD_WORD_PATTERN, which is anchored at `^`. That shipped:
+  //     19 of 24 Gemma `terse` bullets in the 2026-08-07 eval reports carried
+  //     literal `**` through to `perBullet[].text`, i.e. through to what a
+  //     downloaded ATS PDF would render as asterisks. That renderer draws real
+  //     bold from its own type scale and never interprets markdown, so a
+  //     surviving `**` is always garbage, never formatting.
+  //
+  // Looping resolves it without having to pick: each pass peels whatever is
+  // now outermost, and a marker uncovered on one pass is consumed on the next.
+  // Every branch only ever removes characters, so the line strictly shrinks
```
s-annam: **Nit — this paragraph proves the wrong property.**

"Every branch only ever removes characters, so the line strictly shrinks and the loop always converges" is true, and it establishes **termination**. But termination was never the risk in a loop of three anchored regexes over a short string — **idempotence** was, and that is the property the code actually needs and does not have (see the comment on the marker strip below). A strip that is safe applied once to *model output* is not automatically safe re-applied to *its own output*.

Worth reframing so the next reader is warned about the real hazard: each strip must be safe to re-apply to a line it has already transformed.

src/lib/webllm/post-process.test.ts
```diff
@@ -168,4 +168,56 @@ describe("cleanRewriteLine", () => {
...
+  // This shipped: 19 of 24 Gemma `terse` bullets in the 2026-08-07 eval reports
+  // carry literal `**` at `perBullet[].text` — post-`cleanRewriteLine` output,
+  // i.e. what a downloaded ATS PDF renders as asterisks. That renderer draws
+  // real bold from its own type scale and never interprets markdown, so a
+  // surviving `**` is always garbage.
+  //
+  // Every case below returns the marker-shielded input unchanged against the
+  // pre-fix implementation.
+  describe("list marker in front of a leading bold (#781)", () => {
+    it.each([
```
s-annam, line 185: **Secondary — this table cannot catch the class of bug the loop introduces.**

All six cases share the body `"the migration of billing."`. That proves a marker no longer *shields* the bold, which is the fix — but every case is blind to the complementary failure, where a strip **over**-fires on the body it just uncovered. That is precisely the gap the Blocking finding fell through: the loop now mangles `- 3.5x revenue growth in Q3.` into `5x revenue growth in Q3.` and nothing here notices.

Adding one case closes it, and it would have failed on this branch:

```ts
it("does not eat a decimal that a marker strip uncovers", () => {
  expect(cleanRewriteLine("- 3.5x revenue growth in Q3.")).toBe(
    "3.5x revenue growth in Q3.",
  );
});
```

The `not.toContain("*")` property test at :196 is a good instinct — the number-preservation equivalent (`- 2.5M ...` keeps `2.5`) is the one that was missing.

s-annam (changes requested):
The diagnosis is right, the fixed-point framing is the right shape, and the evidence is unusually good — I re-ran the report scan and the **19 affected / 18 leading-bold-only** split is exactly right, as is the mutation check (8 of 9 new cases fail pre-fix). But the loop introduces a **number-corrupting regression** on the very input class this PR exists to handle, so this needs one more pass.

## Blocking

**The marker strip is not idempotent, and the loop now runs it twice** — `post-process.ts:138`. Full detail inline; the short version:

| Input | `main` | this PR |
| --- | --- | --- |
| `- 3.5x revenue growth in Q3.` | `3.5x revenue growth in Q3.` | **`5x revenue growth in Q3.`** |
| `- 2.5M users onboarded.` | `2.5M users onboarded.` | **`5M users onboarded.`** |
| `* 1.5x faster builds.` | `1.5x faster builds.` | **`5x faster builds.`** |
| `• 4.2% conversion lift.` | `4.2% conversion lift.` | **`2% conversion lift.`** |
| `- 10.5 million records processed.` | `10.5 million records…` | **`5 million records…`** |

Pass 1 strips the bullet glyph; pass 2 sees a line that now *starts* with `3.` and the `\d+[.)]\s*` branch (**zero**-or-more trailing space) eats it as a numbered marker. Marker + decimal-leading quantified bullet is exactly the combination this PR targets, and `1.8M` / `11.5M` / `4.5` already appear in this repo's own rewrite corpus. `checkNumbersPreserved` only flips an advisory boolean, so the mangled text still reaches the UI and `render-ats-pdf.ts` — while blaming the model for corruption post-processing inflicted.

There's a **verified one-character fix** in the inline comment: all 31 tests stay green, and it also repairs the same bug on the bare form, which `main` has today.

## Secondary

- **The new test table structurally can't catch this class** — `post-process.test.ts:185`. Every case shares one body, so it proves a marker no longer *shields* the bold but says nothing about a strip over-firing on the body it uncovers. Inline.
- **`MAX_PREFIX_PASSES` is load-bearing, not belt-and-braces** — `post-process.ts:119`. It silently truncates rather than converging. Inline; this answers your review-focus question directly.
- **A leading quote still shields all three strips** — pre-existing, not a regression, but it undercuts this PR's own premise. Quote-stripping runs *after* the loop, so `"**Led** the thing."` → `**Led** the thing.` and `"- **Led** the thing."` → `- **Led** the thing.` — both leak literal `**` to the PDF, the exact defect class the docblock cites as motivation. The new comment says "the three strips below shield each other in BOTH directions," but there is a fourth stripper outside the loop that shields all three. Either peel a leading quote before the loop or fold quote-stripping into it (probably its own issue, alongside #805).
- **`Closes #782` with one of its ACs unmet.** #782 asks that *"No rewritten bullet reaching the UI or the exported PDF can contain `**`"*; mid-line bold is deliberately left, so `- **Developed** and **implemented** …` still ships a pair. The carve-out is well-argued and #781 explicitly permits the documented-out-of-scope route, so I am **not** blocking the decision — but merging closes #782 and the residual `**` then lives nowhere. Suggest `Refs #782`, or fold the residual into #805's scope. Same for #782's *"a re-run shows Gemma / terse free of literal `**`"*, which the PR honestly disclaims under **Not in this PR**.

## Nits

Two inline (`MAX_PREFIX_PASSES` placement vs. house style; the convergence comment reasons about termination rather than the hazard the loop actually introduces). Non-blocking.

## On `## Review focus`

You asked whether `MAX_PREFIX_PASSES = 4` is enough or is noise — **keep it, but the comment describing it is wrong** (see inline). And I'd keep the loop: the instinct that iterating beats guessing the nesting order is correct, and `* Foo.*` is good evidence for it. The fix is to make each strip safe to **re-apply**, not to retreat to a fixed two-phase order. Worth noting the framing gap, since it's where the bug hid: the risk you anticipated was too *few* passes; the defect is a strip firing one time too *many*. The comment at `:109` — "every branch only ever removes characters, so the line strictly shrinks and the loop always converges" — is true and proves termination, but termination was never the hazard. Idempotence was.

## Gates

`fallow` clean on both changed files · `typecheck` + `eslint` clean · 451 webllm tests pass · fixture-PII / design-system / style-token gates **not applicable** (no fixtures, components, or styles touched). Reviewed at `99414b9`.

---

src/lib/webllm/post-process.ts
Samhit21: Fixed in `cfb7b2c`. You are right, and this was the serious one — I introduced silent numeric corruption while fixing a markdown bug, in a repo already tracking dropped numbers (#778).

Reproduced your table exactly before changing anything:

```
"- 3.5x revenue growth in Q3."  ->  "5x revenue growth in Q3."
"- 2.5M users onboarded."       ->  "5M users onboarded."
"* 1.5x faster builds."         ->  "5x faster builds."
"- 10.5 million records..."     ->  "5 million records..."
```

Took your one-character fix — the numbered branch now requires `\s+`. All six cases round-trip intact, `1. Shipped` / `1) Shipped` still strip, and bare `3.5x revenue growth` (which `main` mangles today) now comes out right too.

Added the decimal cases as their own `it.each` rather than folding them into the existing table, because they assert a different property — that a strip does not **over**-fire on what it uncovered, versus the original table asserting a marker no longer **shields**. Plus the digit-preservation property across all five markers, which is the `not.toContain("*")` counterpart you identified as missing. Reverting `\s+` to `\s*` now fails **7** tests.

Your point about the downstream signal is the part I had not thought through: `checkNumbersPreserved` would have reported this as the model dropping a number, sending someone to debug the prompt for damage post-processing did. That is a worse outcome than the corruption itself.

src/lib/webllm/post-process.ts
Samhit21: Fixed in `cfb7b2c`. You are right that I proved the wrong property — termination was never the risk in three anchored regexes over a short string, and I wrote a paragraph reassuring the reader about it while the actual hazard went unmentioned.

The comment now leads with idempotence, names `LIST_MARKER_PATTERN` as the strip that failed it, shows the concrete `- 3.5x` → `5x` path, and states the rule for anyone adding a strip to the loop: it must be safe to re-apply to a line it has already transformed. The shrinking argument is kept only where it belongs — justifying that the runaway guard is not a correctness mechanism.

src/lib/webllm/post-process.ts
Samhit21: Fixed in `cfb7b2c` — went with true convergence rather than keeping `4` and redescribing it.

Confirmed your case first: at `4`, `cleanRewriteLine("1. - • - **Led** stuff.")` returned `"**Led** stuff."` — the literal `**` this function exists to remove, i.e. the bound silently reproducing #781. So it was not belt-and-braces and my comment was falsifiable and false, exactly as you said.

Now a `while` on `body !== before`, with `MAX_PREFIX_PASSES` raised to 32 and documented as a **runaway guard, not a budget for expected nesting** — it only bounds a hypothetical future strip that grows the line. Added a test pinning convergence on the stacked-marker input; capping passes back at 4 fails it.

Nit taken too: `MAX_PREFIX_PASSES` and a new named `LIST_MARKER_PATTERN` both moved to module scope beside `LEADING_BOLD_WORD_PATTERN`, and the unused `pass` counter is gone. Naming the marker pattern also gave the `\s+` fix from your other thread somewhere to be explained, which is where a future reader is most likely to try reverting it.

src/lib/webllm/post-process.test.ts
Samhit21: Fixed in `cfb7b2c`. Fair — six cases sharing one body proved one direction and were structurally blind to the other, which is exactly why the decimal bug walked past them.

Added, and each fails on the previous commit:

- an `it.each` of six decimal-uncovering cases, including the bare `3.5x revenue growth` that is not marker-prefixed at all;
- `preserves every digit of a leading decimal, for any marker` — the number-preservation property across all five markers, stated as a property so a new marker branch has to satisfy it rather than just the enumerated literals;
- `still strips a genuine numbered marker`, so the `\s+` requirement cannot be "fixed" back by someone who thinks it broke numbered lists;
- `converges on stacked markers instead of truncating`, from the bound thread.

40 tests in the file now. Mutation results: reverting `\s+` fails 7, capping passes at 4 fails 1, restoring the pre-fix ordering fails 8.

src/lib/webllm/post-process.ts
```diff
@@ -74,6 +74,33 @@ const CHAT_OPENER_PATTERN = /^here (?:are|is) (?:the )?rewritten\b/i;
...
+ *     raw model output (a decimal is rarely line-initial there), but
+ *     `cleanRewriteLine` re-applies it to lines it has already stripped, so
+ *     `- 3.5x revenue growth` became `5x revenue growth`. Silent numeric
+ *     corruption on the product path, in a bullet the rewrite prompt actively
+ *     asks the model to quantify. Nothing needs a tight `1.Foo`.
+ *   - **`-` and `•` allow `\s*`**, so a tight `-Shipped X` still normalizes.
+ *   - **`*` requires `\s+`**, because `*X*` is italics and is handled by the
+ *     paired-emphasis strip instead.
+ */
+const LIST_MARKER_PATTERN = /^(?:\d+[.)]\s+|[•\-]\s*|\*\s+)/;
```
s-annam, line 94: **[Secondary]:** the `-`/`•` branch still has the idempotence hole this PR just closed on the numbered branch — and it is a regression against `main`.

`[•\-]\s*` accepts zero trailing space, so once a pass strips a marker, the next pass can eat a **negative sign** the same way the old `\d+[.)]\s*` ate a decimal point:

```
"- -5% churn in Q3."      → "5% churn in Q3."     (main: "-5% churn in Q3.")
"- -12% cost reduction."  → "12% cost reduction." (main: "-12% cost reduction.")
```

Pass 1 removes `- `, pass 2 reads the uncovered `-` as a bullet glyph. Same failure mode as `- 3.5x revenue growth` → `5x revenue growth`: silent numeric corruption on the product path, in a bullet the rewrite prompt asks the model to quantify — and a sign flip reads as a *win* rather than a loss, so nothing downstream looks wrong.

This is narrower than the decimal case (it needs a line-initial negative number), which is why it is Secondary rather than Blocking. But the docblock two lines up states the rule it breaks — *"Any strip added to this loop must be checked the same way"* — and this branch was carried over without that check.

The symmetric fix, which also cleans up the bare `-5% churn.` case `main` already mangles:

```ts
const LIST_MARKER_PATTERN = /^(?:\d+[.)]\s+|•\s*|-(?:\s+|(?=[A-Za-z]))|\*\s+)/;
```

`-Shipped X` still normalizes via the letter lookahead, `- -5%` and `-5%` both keep their sign, and `• 4.2% conversion lift.` is untouched. Worth a case in the decimal `it.each` table, since that table is the one asserting this property.

src/lib/webllm/post-process.ts
```diff
@@ -87,30 +114,73 @@ export function cleanRewriteLine(line: string): string {
...
+  // `5x revenue growth`. Requiring `\s+` there is what makes it re-appliable.
+  // Any strip added to this loop must be checked the same way.
+  //
+  // Mid-line bold (`Developed and **implemented** …`) is deliberately NOT
+  // handled here. Two tests pin that as intentional — "does NOT strip emphasis
+  // mid-line" and "does NOT strip mid-bullet bold emphasis" — on the reasoning
+  // that a bold inside a sentence is authored emphasis rather than a model tic.
+  // It is also rare in practice: of the 19 affected eval bullets, 18 carried a
+  // leading bold only. Reversing that contract is a separate decision from
+  // fixing the marker interaction, so it is left to #781's follow-up.
```
s-annam, line 154: **[Secondary]:** this defers mid-line bold to "#781's follow-up", but the PR body says `Closes #781` — so merging this closes the issue the comment points at, and the follow-up it names will never be filed.

#805 does not cover it either: that is the eval-rubric criterion, a different half of #781's AC list. So after merge the mid-line remainder is tracked nowhere, while both issues read as done.

That matters most for **#782**, whose AC is stricter than #781's and is left unmet on two points:

- *"No rewritten bullet reaching the UI or the exported PDF can contain `**`"* — mid-line bold still ships (`- **Developed** and **implemented** …` → `Developed and **implemented** …`), and so do multi-word leading bolds (`- **Led the migration** of billing.` unchanged) and quoted lines (Secondary 3 in the body).
- *"A re-run shows Gemma / terse free of literal `**`"* — the body honestly defers this as needing a maintainer machine.

I am not treating the mid-line decision itself as a defect: #781's AC explicitly allows *"stripped, or explicitly documented as out of scope with the reason"*, and you documented it in the code, in the tests, and in the body. That AC is met. The gap is purely that GitHub will close **#782** on ACs it does not satisfy, with no successor issue.

Two ways out, either is fine:

- Narrow to `Refs #782` and let it stay open for the remaining shapes; or
- keep `Closes #782`, file the mid-line/multi-word follow-up now, and point this comment at *that* number instead of #781.

s-annam (approved):
> **Verdict: APPROVE** (0 Blocking, 3 Secondary, 3 Nits). Rule applied: 0 Blocking → APPROVE; nits and secondaries do not soft-gate.

I judged this against #781/#782's acceptance criteria and re-derived every claim before reading the description. The fix is correct for the defect it targets, and the round-2 work is the strongest part of it: the `- 3.5x revenue growth` → `5x revenue growth` catch is a genuine silent-corruption bug that the naive loop introduced, and requiring `\s+` on the numbered branch is the right fix.

**Gates run on `1fed7e8`:** AC checklist per issue · fail-before-fix reproduction · adversarial idempotence probe (22 shapes) · `typecheck` clean · `eslint` clean on both files · `fallow audit --base origin/main` **no issues in 2 changed files** · `npx vitest run src/lib/webllm` **460 passed / 29 files** · 3a fixture PII **skipped, no fixtures touched** · 3b design-system, 3c tokens, 3e skill/script **skipped, none apply**.

| Claim | Verdict |
|---|---|
| All six marker shapes strip to `Led the migration of billing.` | ✅ |
| `*Foo.*` and `* Foo.*` invariant preserved | ✅ |
| New tests fail against the pre-fix implementation | ✅ **10 of 40 fail** when `origin/main`'s `post-process.ts` is swapped in |
| `main` also mangles bare `3.5x revenue growth` → `5x` | ✅ reproduced on `main`; this PR fixes it, and the body says so |
| Multi-word leading bold left alone by design | ✅ `- **Led the migration** of billing.` unchanged |
| Loop converges; every branch only removes characters | ✅ |
| One commit | ✅ |

The fail-before-fix count is **10**, not the 8 the round-2 table lists — the two extra are the decimal cases, which also fail on `main`. That is the table being conservative, not wrong.

## Secondary

**1. The `-`/`•` branch has the same idempotence hole the numbered branch just had.** See inline at `post-process.ts:94`. `- -5% churn in Q3.` → `5% churn in Q3.` on this branch; `main` returns `-5% churn in Q3.`. This is a new regression, and it is the exact class the loop comment says every strip must be checked for.

**2. `Closes #782` ships with two of its ACs unmet, and the mid-line follow-up has no home.** See inline at `post-process.ts:154`.

**3. A surrounding quote shields the whole prefix stack, because the quote strip runs after the loop.** `post-process.ts:187` is unchanged by this diff, so there is no `+` line to anchor to:

```
cleanRewriteLine('"- **Led** the migration."')  → '- **Led** the migration.'
cleanRewriteLine('"**Led** the migration."')    → '**Led** the migration.'
cleanRewriteLine('“- **Led** the migration.”')  → '- **Led** the migration.'
```

Both the marker *and* the bold survive to the PDF, and the result is not a fixed point — feeding the output back in cleans it, which is the tell. This is pre-existing rather than a regression, but it defeats the same AC this PR is closing, and the function now has a loop that would absorb it: moving the quote strip above the loop (or into it) closes the shape. The function already strips quotes precisely because models emit them, so the quote + marker + bold combination is not hypothetical for the same chatty model that produced the markers.

## Nits

**4. The body's round-1 text contradicts the shipped code.** `## Review focus` still says "The `MAX_PREFIX_PASSES = 4` bound" (it is 32) and "the pass bound is belt-and-braces" (round 2 establishes it was a truncation point); `## Test plan` still says "31 passed (22 pre-existing + 9 new)" (it is 40). Round 2 corrects all three further down, so a reader of the whole body gets the truth — but the earlier sections are what a skimmer reads. Worth folding the round-2 findings up into the sections they invalidate. The body also cites `cfb7b2c`, which the collapse to one commit replaced with `1fed7e8`.

**5. A tight `1.Foo bar.` is no longer stripped** — `main` returned `Foo bar.`, this branch returns `1.Foo bar.`. That is the deliberate cost of `\s+` on the numbered branch and the pattern docblock owns it ("Nothing needs a tight `1.Foo`"), which I agree with. It is a behaviour change that the description does not mention, unlike the bare-`3.5x` side effect, which it does.

**6. `- **3D** printing scaled.` keeps its asterisks** because `LEADING_BOLD_WORD_PATTERN` requires `[A-Za-z]` as the first character. Pre-existing and consistent with the single-word contract; noting it only because a digit-initial bold is plausible from the same model that emits markers, and it is one character (`[A-Za-z0-9]`) if you want it.

## No auto-fix commit

Nothing here qualified. Findings 1–3 and 5–6 all change what the code *does*, which puts them outside what an unreviewed auto-fix commit may contain; finding 4 is a description edit and finding 2 needs a decision (which issue the mid-line remainder lands in) that is yours, not mine. So this approval covers `1fed7e8` exactly as you pushed it.

---

Issue #804, opened 2026-08-12
loadEngine silently drops a second caller's onProgress, pinning its download progress at 0


s-annam:
## Problem

`loadEngine`'s "already pending" fast path silently discards the new caller's `onProgress` callback. Only the **first** caller's callback is ever wired to WebLLM's `initProgressCallback`:

```ts
// src/lib/webllm/web-llm.ts
// Fast path B: this model is already being loaded (concurrent same-id
// calls share one load).
const pending = pendingByModelId.get(modelId);
if (pending) return pending;          // <-- the caller's onProgress is dropped
```

The slow path wires `initProgressCallback: (report) => onProgress(report)` from whichever caller created the slot. Every later caller for the same `modelId` gets the shared promise and no progress at all — silently, with no signal that its callback will never fire.

## Failure scenario

Both callers exist on `/jobs/` today:

- `src/lib/job-search/sector.ts` calls `loadEngine(DEFAULT_MODEL_ID, () => {})` — a **no-op** progress callback — to classify sectors.
- `useJdMatch` (#801) calls `loadEngine(selectedModelId, onProgress)` for the semantic JD match.

If the user's persisted `selectedModelId` **is** `DEFAULT_MODEL_ID` (the common case — it's the default) and the sector classify's load is still pending when the semantic path starts, `useJdMatch` receives the sector's promise. The sector's no-op callback owns `initProgressCallback`, so the hook's `onProgress` never fires and its status stays at `loading { progress: 0, text: "Starting…" }` for the entire multi-GB weight download.

The download-progress UI #204 is meant to build would read dead: a spinner pinned at 0% for minutes, with no way for the consumer to tell "no progress yet" from "progress will never arrive".

Latent today only because `semanticOptIn` is hardwired `false`.

## Why it isn't a caller-side fix

Every `loadEngine` caller passes a progress callback and any two of them can collide on the same `modelId`. Working around it per-caller would mean each one re-implementing progress fan-out. The dedupe is the right behaviour — sharing one download is exactly what `pendingByModelId` is for — it's the *callback* handling that's wrong.

## Proposed approach

Keep one load per `modelId`, fan progress out to every caller waiting on it:

1. Replace `pendingByModelId: Map<string, Promise<WebLlmEngine>>` with a record holding the promise plus a `Set<(u: ProgressUpdate) => void>` of subscribers.
2. `initProgressCallback` iterates the set instead of calling a single captured callback.
[...]

PR #807, opened 2026-08-13
feat(jd-match): cancel abandoned semantic runs via AbortController per run (#803)

### LINKED ISSUE #803: JD-match semantic runs are hidden but never cancelled — abandoned runs stack on the shared MLCEngine
## Problem

`runLlmMatch` has no cancellation. `useJdMatch`'s `requestIdRef` **hides** a superseded run's writes, but nothing stops the run itself: `runLlmMatch` takes no `AbortSignal`, and `acquireInference` is a reference counter (it defers `.unload()`), not a mutex — so it does not serialise either.

Today this is latent: `PasteJdPanel` passes `semanticOptIn={false}`, so no semantic run ever starts. #204 flips that flag, and this becomes reachable on the first edit a user makes to a pasted JD.

## Failure scenario

`useJdMatch` debounces JD text at `JD_MATCH_DEBOUNCE_MS` (200 ms) and keys the semantic run off the debounced value. A user pastes a JD and makes three corrections with >200 ms between them:

1. Paste → run A starts (`loadEngine` → `extractRequirements` → `judgeEvidence`).
2. Edit 1 → id bumped, run B starts. **A keeps running.**
3. Edit 2 → run C starts. A and B keep running.
4. Edit 3 → run D starts. A, B and C keep running.

Four chains are now in flight against the single shared `MLCEngine`. Three of them will have every write dropped by the id guard, but they still occupy the engine's token slots for a full `extractRequirements` plus an N-batch `judgeEvidence`. The one run whose result the user will actually see finishes roughly 4× later than it needed to, and on-device inference is the slowest thing in the product.

`judgeEvidence`'s per-batch `acquireInference`/`releaseInference` does not help — the counter only parks `.unload()`; concurrent callers interleave freely.

## Why it wasn't fixed in #801

Threading cancellation touches three modules with three different shapes (`runLlmMatch` orchestration, `extractRequirements`' single completion, `judgeEvidence`' batch loop), and `web-llm.ts` exposes no abort primitive for an in-flight `chat.completions.create`. That is a design decision, not a patch, and #801 was a controller-extraction PR that shipped the semantic path dormant.

## Proposed approach

1. Add an optional `signal?: AbortSignal` to `runLlmMatch`, threaded to `extractRequirements` and `judgeEvidence`.
2. Check `signal.aborted` between the pipeline stages and between `judgeEvidence` batches — that alone bounds wasted work to at most one in-flight LLM call per abandoned run, without needing engine-level abort.
3. Have `useJdMatch` own an `AbortController` per run: abort it wherever it currently bumps `requestIdRef` (new run start, and exit from the semantic path). The id guard stays — it protects against the write, and this protects against the work.
4. Investigate whether `MLCEngine` exposes any interrupt for a streaming completion; if it does, wire it, and if not, note the residual bound (one call).

## Acceptance criteria

- `runLlmMatch` accepts an `AbortSignal` and stops at the next stage boundary once aborted, resolving to the keyword fallback rather than rejecting (its never-rejects contract must hold).
- `useJdMatch` aborts the previous run before starting a new one, and on exit from the semantic path.
- A test drives three rapid debounced JD changes and asserts at most one `judgeEvidence` batch runs per abandoned run — i.e. abandoned work is bounded, not unbounded.
- A test asserts an aborted run still resolves (never rejects) and never writes to the slot.
- Existing four-param `runLlmMatch` callers keep working unchanged.

## Context

Found while reviewing #801 (see the "Known limitations" section of that PR body). Related: #148 established the `acquireInference` contract this interacts with; #204 is the issue that makes this reachable.

Vaishnavi1709:
## Summary

`runLlmMatch` takes an optional `AbortSignal`; `useJdMatch` owns one `AbortController` per semantic run and aborts on supersession, opt-out, JD clear, model change, and unmount. Bounds the abandoned work per superseded run to at most one currently-in-flight `chat.completions.create()` call — instead of the pre-fix behaviour where a paste + N debounced edits stacked N `extractRequirements` + N × (multi-batch) `judgeEvidence` chains on the shared `MLCEngine`.

Closes #803. Unblocks #204 — the opt-in UI can now activate the semantic path without exposing a known engine-concurrency defect.

## Design

**Three layers, not two.** The [pre-existing `requestIdRef`](https://github.com/offlinecv/OfflineCV/blob/main/src/hooks/useJdMatch.ts#L340) + slot-vs-current-inputs value comparison from #203 are preserved unchanged — they hide stale WRITES. #803 adds a per-run `AbortController` that stops the WORK. Both are load-bearing on different scenarios; the tests would fail if either is dropped.

**Boundary cancellation, not engine interrupt.** `MLCEngine.interruptGenerate()` exists in `@mlc-ai/web-llm@0.2.84`, but it is engine-scoped rather than request-scoped. Calling it would cancel every concurrent completion on the shared engine, including work started by `job-search/sector.ts` on the same `/jobs/` page — the exact concern #148's inference-guard machinery exists to protect. So we deliberately do NOT use it. Instead, seven numbered cancellation boundaries — four in `runLlmMatch` itself (pre-acquire, post-`loadEngine`, post-extract, post-judge), two inside `extractRequirements` (pre-call and post-call), and one per batch in `judgeEvidence` — short-circuit at the next safe point. Residual bound: one `loadEngine` await + one currently-in-flight completion (extract OR one judge batch). This is exactly the bound #803's own proposed approach §2 describes.

**Cancellation is expected control flow, not a failure — but only when it's ours.** `runLlmMatch`'s catch takes the silent branch on `isAbortError(err) && signal?.aborted`: real failures still log `[run-llm-match] semantic path failed` and degrade to keyword; our own cancellations resolve to keyword silently. Both hold the never-rejects contract. The `signal?.aborted` conjunct is what keeps an engine error that merely *looks* like an abort out of the silent branch — see Review focus.

**Never-throws contract preserved for `judgeEvidence`.** On abort, the batch loop `break`s (not throws), and the reconciled result — with abandoned reqs defaulting to `missing` — comes back as normal. The orchestrator then discards it in favor of keyword via its own post-call `signal.aborted` check, so the partial shape never reaches a user.

**Unmount abort is deferred one microtask — as a belt, not a bug fix.** The mount effect's cleanup schedules the abort via `queueMicrotask`, gated on `mountedRef.current` still being `false` when it fires. Stating what that does and does not buy, because the shape looks like a live fix and an earlier draft of this description claimed it was one: the mount effect has `[]` deps, so its cleanup fires exactly twice — at StrictMode's simulated unmount on the initial mount, and at real unmount. At the initial one, `debouncedJdText` is still `""` (a 200 ms timeout away) and `capability` is still `null` (a promise away), so `takingSemanticPath` is `false` through the whole synchronous double-invoke and `controllerRef.current` is `null`. There is no run to kill, and a synchronous abort there would be a no-op today. What the hop actually buys is (a) a stale microtask queued by the double-invoke cannot later kill the real run, and (b) a future change that lets a run start before the first commit settles stays safe, because StrictMode's synchronous re-mount body re-sets `mountedRef.current = true` before the queued microtask runs. On a REAL unmount no re-mount runs, `mountedRef` stays `false`, and the abort fires. Same commit-then-microtask ordering that makes #203's own `mountedRef.current = true` re-set pattern work. The test is scoped to (a) — the part that is real.

**Structural acquire/release balance.** The pre-acquire abort short-circuit lives OUTSIDE the try/finally, so bailing on an already-aborted signal correctly skips both `acquireInference` and `releaseInference`. Every other exit — semantic return, empty-extraction degrade, mid-pipeline abort return, catch fallback — is inside the try and releases via `finally`. A leaked acquire would pin `inflightInferenceCount` above zero for the page's lifetime and park every later cross-model `.unload()` forever.

## Review focus

- **`abort.ts`'s DOMException choice.** The fetch-API convention (`DOMException(name="AbortError")`) makes `isAbortError` a structural check that matches whatever `AbortSignal.prototype.throwIfAborted()` would produce, so a future signal-aware `chat.completions.create` throws the same shape we throw here. Structural (`err.name === "AbortError"`) rather than `instanceof DOMException` so it survives realms with different constructor identity (jsdom vs. node vs. workers).
- **`isAbortError` is paired with `signal?.aborted` at both call sites**, and the pairing is load-bearing rather than belt. The shape check alone cannot tell *our* cancellation from an engine error that merely happens to be named `"AbortError"`, and both consumers suppress diagnostics on a match — `run-llm-match.ts` returns keyword with its `console.warn` deliberately skipped, and `extract-requirements.ts` re-throws unwrapped into that same branch. Unpaired, a future per-request timeout or a device-lost surfaced this way degrades to keyword with zero diagnostic trail: a silent-failure class that is very hard to diagnose from a user report. Latent, not live, on the pinned `@mlc-ai/web-llm@0.2.84` — its only `AbortError`-shaped throw is internal to `MLCEngine.reload()`'s own `reloadController`, which this pipeline never reaches — but closed here rather than left standing. One new test per module pins the unfired-signal half; both go red without the conjunct.
- **The controller-store-then-abort ordering** in `useJdMatch`'s new-run branch. Store new BEFORE aborting old, so a concurrent reader of `controllerRef.current` always sees a live controller, never a null gap.
- **`judgeEvidence` checks the signal once per batch, before the batch — not after.** A post-`await` check would be unreachable: only the loop increment separates it from the next iteration's pre-check, and `signal.aborted` cannot flip across synchronous code, so anything it caught would be caught identically one line later. An abort that fires mid-completion is therefore handled by the pre-check on the following iteration. Removing the pre-check goes red (see the fail-before table); no test could exist for a post-check, which is why there isn't one.
- **`useJdMatch`'s exit branch aborts on `semanticSlot !== null || controllerRef.current !== null`.** Gating on the slot alone would be correct only by way of an unstated ordering invariant — the controller is installed synchronously at run-start while the matching `setSemanticSlot` is merely scheduled, so "slot non-null" stands in for "a run exists" only because `semanticSlot` is itself a dep of that effect. That holds under React 18/19 automatic batching and I could not construct a counterexample, but a refactor moving the kickoff behind a promise would strand a live controller silently. Reading the ref removes the dependency on the ordering. `abort()` on an already-resolved run's controller is a spec no-op, so a `ready` slot still survives per the pre-existing #203 opt-out-preserves-completed-result invariant.

## Known limitations (deliberately out of scope)

- **`loadEngine` is not signal-aware.** The `pendingByModelId` fast path returns the same promise to every consumer for a model id, so cancelling would abort a load another consumer is still waiting on. Bounded to one in-flight load per abandoned run. Would require a shared web-llm.ts change beyond this PR's scope.
- **The currently in-flight `chat.completions.create()` call cannot be interrupted.** See the design section — we deliberately do NOT call `MLCEngine.interruptGenerate()` because it is engine-scoped rather than request-scoped and would cancel other consumers' work on the shared engine. Bounded to one currently-in-flight completion per abandoned run.
- **#804 (`loadEngine` fast path drops second caller's `onProgress`)** is unrelated to #803's correctness — my cancellation logic reads only `signal.aborted`, never `onProgress` values. Filed separately.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run build` clean. `run-llm-match-*.js` stays a lazy 7.51 kB chunk (was 7.04 kB on `main`; +0.47 kB for the abort helpers). `jobs-*.js` 88.29 kB (was 88.04 kB). WebLLM chunk remains out of the entry bundle.
- [x] `npx vitest run src/lib/jd-match/llm/run-llm-match.test.ts` → 23/23 (11 pre-existing + 12 new)
- [x] `npx vitest run src/lib/jd-match/llm/extract-requirements.test.ts` → 14/14 (9 + 5)
- [x] `npx vitest run src/lib/jd-match/llm/judge-evidence.test.ts` → 11/11 (7 + 4)
- [x] `npx vitest run src/hooks/useJdMatch.test.tsx` → 32/32 (23 + 9)
- [x] `npm run test` (full suite) → 5476 passed. The only failures are timeouts in `src/lib/pdf/*` under full-suite parallel load — a different file each run, all passing in isolation, and untouched by this diff (which is confined to `src/lib/jd-match/llm/` + `src/hooks/useJdMatch.*`).

**Fail-before verification.** Each abort check was individually removed and the target test confirmed to go red before restoring:

| Reverted | Test that went red |
|---|---|
| `judgeEvidence`'s per-batch pre-check | `bounded batches: 24 reqs → 1 batch runs then abort halts scheduling` (also 2 other cancellation tests) |
| `runLlmMatch` boundary-5 (post-extract) check | `boundary 5: abort between extract and judge → judge is never called` |
| `useJdMatch`'s supersession `previousController?.abort()` | `A → B → C rapid changes: each supersession aborts the prior; only the last is live` |
| `&& signal?.aborted` in `run-llm-match.ts`'s catch | `an AbortError-shaped engine error on an UNFIRED signal still warns and degrades` |
| `&& signal?.aborted` in `extract-requirements.ts`'s catch | `WRAPS an engine AbortError when our signal never fired` |
| the whole `if (isAbortError(err) …) throw err;` in `extract-requirements.ts` | `does NOT wrap an AbortError from the engine call as RequirementExtractionError` |

That last row is the one that did **not** hold before this round. The test was vacuous: it aborted the controller *before* calling, so the pre-call check threw first, the engine was never invoked, and both assertions passed on an `AbortError` that the re-throw under test had no part in producing. It now fires the signal *during* the call and asserts error **identity** (`rejects.toBe(abortErr)`) rather than shape, so it goes red when the line is deleted.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation and review | Claude Opus 4.7 (1M context) | high |
| Review revisions | Claude Opus 5 | high |

### DIFF
--- a/src/hooks/useJdMatch.ts
+++ b/src/hooks/useJdMatch.ts
@@ -66,8 +66,8 @@
  *       reach this branch; keeping it here means the surface for that is
  *       already public.
  *
- * Stale-request protection has two layers, both load-bearing on DIFFERENT
- * scenarios:
+ * Stale-request protection has TWO layers plus a THIRD one for the WORK, not
+ * the writes. All three are load-bearing on DIFFERENT scenarios:
  *   1. `requestIdRef` — a monotonic counter bumped on every semantic run
  *      start AND on every exit from the semantic path (opt-out, JD cleared,
  *      capability change). The exit ALSO clears the slot unless it holds a
@@ -88,6 +88,26 @@
  *      whose stored `parsed` reference no longer matches, is invisible to
  *      the render. This is what stops a stale slot from flashing over a
  *      newer render even if a late write slipped past layer 1.
+ *   3. `controllerRef` — an `AbortController` per semantic run (#803).
+ *      Aborted whenever the id would be bumped: run supersession, opt-out,
+ *      JD change to a new value or empty, model change, capability going
+ *      false, unmount. The id guard prevents stale WRITES from becoming
+ *      visible; the abort controller stops the WORK — extract's coercion
+ *      pass, and every judge batch after the currently in-flight one. The
+ *      residual bound is one loadEngine await plus one currently-in-flight
+ *      completion (extract OR one judge batch); every downstream stage is
+ *      cancelled. See `runLlmMatch`'s docblock for the boundary map.
+ *
+ *      Unmount abort is deferred by ONE microtask, guarded by `mountedRef`.
+ *      Under React 19 StrictMode the mount effect's cleanup runs BETWEEN two
+ *      effect-body invocations on the same hook instance; a naive abort in
+ *      that cleanup would kill the in-flight run and the second effect body
+ *      would early-return (slot inputs still match), stranding a spinner
+ *      that never ends. The microtask hop lets StrictMode's synchronous
+ *      re-mount body re-set `mountedRef.current = true` first; on a REAL
+ *      unmount no re-mount runs, `mountedRef` stays false, and the abort
+ *      fires. Same synchronous-commit-then-microtask ordering that makes
+ *      #203's own `mountedRef.current = true` re-set pattern work.
  *
  * `parsed` identity contract: the hook treats `parsed` by REFERENCE, and the
  * failure mode is worse than a wasted recompute. A caller that builds `parsed`
@@ -340,6 +360,11 @@ export function useJdMatch(options: UseJdMatchOptions): JdMatchController {
 
   const requestIdRef = useRef(0);
   const mountedRef = useRef(true);
+  /** Per-run AbortController (#803). Non-null between run-start and its resolve
+   *  or its supersession/opt-out abort. Nulled on abort so a stray follow-up
+   *  abort call is a no-op rather than aborting a fresh run that has since
+   *  reused the ref slot. */
+  const controllerRef = useRef<AbortController | null>(null);
   useEffect(() => {
     // Re-set on every mount, not just at hook creation: under React
     // StrictMode the first effect-cleanup fires between the two effect
@@ -349,6 +374,15 @@ export function useJdMatch(options: UseJdMatchOptions): JdMatchController {
     mountedRef.current = true;
     return () => {
       mountedRef.current = false;
+      // Unmount abort (#803, requirement §5). Deferred by one microtask,
+      // gated on `mountedRef.current` still being false — see the
+      // Layer-3 rationale in the module docblock for why a synchronous
+      // abort here breaks StrictMode's simulated remount.
+      queueMicrotask(() => {
+        if (mountedRef.current) return;
+        controllerRef.current?.abort();
+        controllerRef.current = null;
+      });
     };
   }, []);
 
@@ -399,6 +433,14 @@ export function useJdMatch(options: UseJdMatchOptions): JdMatchController {
     if (!takingSemanticPath) {
       if (semanticSlot !== null) {
         requestIdRef.current += 1;
+        // Layer-3 (#803): stop the WORK, not just the writes. Aborting here
+        // prevents the abandoned run's `extractRequirements` coercion pass
+        // and every subsequent judge batch from executing on the shared
+        // engine. The id bump above still handles late writes independently
+        // — the two guards protect different things. Null the ref after so
+        // a later duplicate exit isn't misread as an in-flight controller.
+        controllerRef.current?.abort();
+        controllerRef.current = null;
         if (semanticSlot.state.kind !== "ready") setSemanticSlot(null);
       }
       return;
@@ -425,6 +467,15 @@ export function useJdMatch(options: UseJdMatchOptions): JdMatchController {
     // inputs so the derived status stops rendering LOADING_START (which
     // was the render-time placeholder while slot was stale/null).
     const myId = ++requestIdRef.current;
+    // Layer-3 abort (#803): install a fresh controller and abort the
+    // previous run's, so its `extractRequirements` and every subsequent
+    // judge batch stop scheduling. Store BEFORE aborting so `controllerRef`
+    // always points at the current run for any concurrent read; a stale
+    // reader will see the new controller, never a null gap.
+    const previousController = controllerRef.current;
+    const myController = new AbortController();
+    controllerRef.current = myController;
+    previousController?.abort();
     const myInputs: SemanticInputs = {
       jdText: trimmedJdText,
       parsed,
@@ -455,6 +506,7 @@ export function useJdMatch(options: UseJdMatchOptions): JdMatchController {
           (progress) =>
             setSlotIfCurrent(myId, myInputs, { kind: "loading", progress }),
           () => setSlotIfCurrent(myId, myInputs, { kind: "running" }),
+          myController.signal,
         ),
       )
       .then((result) => {
--- a/src/lib/jd-match/llm/abort.ts
+++ b/src/lib/jd-match/llm/abort.ts
@@ -1,0 +1,58 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * abort.ts — cancellation helpers for the semantic JD-match pipeline (#803).
+ *
+ * The narrow `WebLlmEngine` in `types.ts` exposes only `chat.completions.create()`
+ * and no per-request cancellation. `MLCEngine.interruptGenerate()` DOES exist
+ * on the real `@mlc-ai/web-llm` engine, but it is engine-scoped rather than
+ * request-scoped — calling it would cancel every concurrent completion on that
+ * shared engine, including ones started by other consumers on the same page
+ * (e.g. `job-search/sector.ts`, per #148's whole raison d'être). That would
+ * violate #803's own requirement not to "cancel another consumer's work", so we
+ * deliberately do NOT use it.
+ *
+ * Cancellation is therefore BOUNDARY-based: each stage / batch checks
+ * `signal.aborted` before starting and after resolving. The residual wasted
+ * work an abandoned run can do is bounded by the one LLM call that is already
+ * in flight when the abort fires — that call runs to completion; no subsequent
+ * stage or batch is scheduled. This is exactly the bound #803's own "Proposed
+ * approach" §2 describes.
+ *
+ * We use the fetch-API convention (`DOMException` with `name === "AbortError"`)
+ * so the error we throw here is shape-identical to what
+ * `AbortSignal.prototype.throwIfAborted()` produces on browsers and Node ≥ 17.
+ * That means `isAbortError` is a single check regardless of whether the abort
+ * came from an explicit `throwIfAborted()`, a fetch-shaped consumer, or our
+ * own `abortError()` factory. Not thrown from a fetch, so a `signal.reason`
+ * value is not preserved — the shape (not the message) is what downstream
+ * catches key off.
+ */
+
+/**
+ * Construct a fetch-standard AbortError so downstream catches can key off
+ * `err.name === "AbortError"` regardless of source.
+ *
+ * Kept as a factory (not a shared instance) so the stack trace points at the
+ * abort site rather than at this module's load time — matters when diagnosing
+ * why a stage bailed.
+ */
+export function abortError(reason?: string): DOMException {
+  return new DOMException(reason ?? "Semantic run aborted.", "AbortError");
+}
+
+/**
+ * True for anything shaped like an AbortError: our own `abortError()`, a
+ * DOMException from `AbortSignal.throwIfAborted()`, or any object with
+ * `name === "AbortError"` (the fetch-API contract). Deliberately structural
+ * rather than `instanceof DOMException` so it survives realms where the
+ * `DOMException` constructor identity differs (jsdom vs. node vs. workers).
+ */
+export function isAbortError(err: unknown): boolean {
+  return (
+    typeof err === "object" &&
+    err !== null &&
+    (err as { name?: unknown }).name === "AbortError"
+  );
+}
--- a/src/lib/jd-match/llm/extract-requirements.test.ts
+++ b/src/lib/jd-match/llm/extract-requirements.test.ts
@@ -131,3 +131,86 @@ describe("extractRequirements", () => {
     expect(req.temperature).toBe(0);
   });
 });
+
+describe("extractRequirements — cancellation (#803)", () => {
+  it("pre-call check: an already-aborted signal skips the engine entirely", async () => {
+    // The whole point of the pre-check: avoid starting the expensive
+    // completion at all if the run has already been superseded. If this
+    // check is removed, the engine.chat.completions.create call still fires,
+    // and the wasted work bound is broken.
+    const create = vi.fn();
+    const engine: WebLlmEngine = { chat: { completions: { create } } };
+    const controller = new AbortController();
+    controller.abort();
+
+    await expect(
+      extractRequirements("jd", engine, controller.signal),
+    ).rejects.toMatchObject({ name: "AbortError" });
+    expect(create).not.toHaveBeenCalled();
+  });
+
+  it("post-call check: abort during the completion skips coercion and throws AbortError", async () => {
+    // The completion itself finishes (we can't safely interrupt a shared
+    // engine), but nothing downstream — coercion, and by extension the
+    // orchestrator's judge call — runs. AbortError is distinct from
+    // RequirementExtractionError so the orchestrator can suppress the
+    // console.warn on cancellation.
+    const controller = new AbortController();
+    const create = vi.fn().mockImplementation(async () => {
+      controller.abort();
+      return {
+        choices: [
+          {
+            message: {
+              content: JSON.stringify([
+                { id: "req-1", kind: "skill", text: "TypeScript" },
+              ]),
+            },
+          },
+        ],
+      };
+    });
+    const engine: WebLlmEngine = { chat: { completions: { create } } };
+
+    await expect(
+      extractRequirements("jd", engine, controller.signal),
+    ).rejects.toMatchObject({ name: "AbortError" });
+    // Engine WAS called — that's the bounded work per #803's residual bound.
+    expect(create).toHaveBeenCalledOnce();
+  });
+
+  it("does NOT wrap an AbortError from the engine call as RequirementExtractionError", async () => {
+    // Preserves the orchestrator's ability to distinguish cancellation from
+    // a real extraction failure via `err.name === "AbortError"`. If wrapping
+    // happens, the orchestrator would log "semantic path failed" over a
+    // cancellation we initiated — the exact behavior #803 forbids.
+    const abortErr = new DOMException("aborted", "AbortError");
+    const engine: WebLlmEngine = {
+      chat: {
+        completions: { create: vi.fn().mockRejectedValue(abortErr) },
+      },
+    };
+    const controller = new AbortController();
+    controller.abort();
+
+    await expect(
+      extractRequirements("jd", engine, controller.signal),
+    ).rejects.toMatchObject({ name: "AbortError" });
+    // Not a RequirementExtractionError — that would break the orchestrator's
+    // catch-branch distinction.
+    await expect(
+      extractRequirements("jd", engine, controller.signal),
+    ).rejects.not.toBeInstanceOf(RequirementExtractionError);
+  });
+
+  it("no signal (legacy 2-arg call): behavior unchanged, no aborts anywhere", async () => {
+    // Every pre-#803 caller passes just (jdText, engine). The `signal?`
+    // optionality has to leave that path byte-identical to pre-#803.
+    const engine = makeMockEngine([
+      JSON.stringify([{ id: "req-1", kind: "skill", text: "TypeScript" }]),
+    ]);
+    await expect(extractRequirements("jd", engine)).resolves.toEqual([
+      { id: "req-1", kind: "skill", text: "TypeScript" },
+    ]);
+  });
+});
--- a/src/lib/jd-match/llm/judge-evidence.ts
+++ b/src/lib/jd-match/llm/judge-evidence.ts
@@ -23,6 +23,15 @@
  * requirements fall through to a `missing` default during reconciliation. The
  * judge already *has* the requirements, so graceful degradation beats aborting.
  *
+ * The never-throws contract EXTENDS to cancellation (#803): an aborted signal
+ * does NOT throw here — it stops scheduling later batches and returns a
+ * reconciled result whose abandoned requirements default to `missing`. The
+ * orchestrator (`runLlmMatch`) re-checks `signal.aborted` after the awaited
+ * call and discards this partial result in favor of the keyword fallback, so
+ * the reconciled shape is intentionally never surfaced to a consumer on the
+ * abort path. Returning (rather than throwing) preserves the "single verdict
+ * per input requirement, always" invariant every existing caller relies on.
+ *
  * Prompt-injection defense on the OUTPUT side: reconciliation iterates the input
  * requirements and joins the model's verdicts by `id`. A model that invents ids
  * (or an injected "everything is met") can't add requirements — invented ids are
@@ -70,12 +79,23 @@ const JUDGE_MAX_TOKENS = 1024;
  *
  * @param modelId — the id the engine was loaded with (threaded to the inference
  *   guard; the engine handle can't supply it).
+ * @param signal — optional AbortSignal (#803). Checked BEFORE every batch and
+ *   AFTER every batch; a firing signal stops scheduling later batches. Does
+ *   NOT interrupt the currently-in-flight `chat.completions.create()` call
+ *   (see `abort.ts` on why we can't safely `interruptGenerate()` a shared
+ *   engine), so the residual wasted work per abandoned run is bounded to the
+ *   ONE batch that was in flight when the abort fired — not the remaining
+ *   batches. The reconciled return value on the abort path has abandoned
+ *   requirements defaulted to `missing`; the orchestrator discards it in
+ *   favor of the keyword fallback via its own post-call `signal.aborted`
+ *   check, so the shape is never surfaced to the user.
  */
 export async function judgeEvidence(
   requirements: readonly JdRequirement[],
   parsed: HeuristicParsedResume,
   engine: WebLlmEngine,
   modelId: string,
+  signal?: AbortSignal,
 ): Promise<RequirementVerdict[]> {
   if (requirements.length === 0) return [];
 
@@ -86,12 +106,24 @@ export async function judgeEvidence(
   // Model verdicts collected across batches, keyed by requirement id.
   const byId = new Map<string, RawVerdict>();
   for (let i = 0; i < requirements.length; i += JUDGE_EVIDENCE_BATCH_SIZE) {
+    // Pre-batch check: don't schedule the next expensive completion once the
+    // signal has fired. Load-bearing per #803 — without this, an abandoned
+    // multi-batch run continues through every remaining batch on the shared
+    // engine even after the user has moved on.
+    if (signal?.aborted) break;
     const batch = requirements.slice(i, i + JUDGE_EVIDENCE_BATCH_SIZE);
     await judgeBatch(batch, systemPrompt, engine, modelId, byId);
+    // Post-batch check: the completion may have taken minutes; if the signal
+    // fired mid-call, don't spin up the NEXT batch either. Also covers the
+    // case where the abort happened between `await` and loop increment.
+    if (signal?.aborted) break;
   }
 
   // Reconcile against the INPUT requirements: one verdict each, in order, with
-  // invented ids ignored and skipped requirements defaulted to `missing`.
+  // invented ids ignored and skipped requirements defaulted to `missing`. On
+  // the abort path, "skipped" includes every requirement in a batch we never
+  // scheduled — those default to `missing`, but the orchestrator's post-call
+  // abort check discards the whole array before it reaches a consumer.
   return requirements.map((requirement) => {
     const raw = byId.get(requirement.id);
     if (raw === undefined) {
--- a/src/hooks/useJdMatch.test.tsx
+++ b/src/hooks/useJdMatch.test.tsx
@@ -398,13 +398,18 @@ describe("useJdMatch — semantic path (opt-in + WebGPU available)", () => {
     await flushMicrotasks();
 
     expect(runLlmMatchMock).toHaveBeenCalledTimes(1);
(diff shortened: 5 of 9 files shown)


src/lib/jd-match/llm/judge-evidence.ts
```diff
@@ -86,12 +106,24 @@ export async function judgeEvidence(
...
+    // signal has fired. Load-bearing per #803 — without this, an abandoned
+    // multi-batch run continues through every remaining batch on the shared
+    // engine even after the user has moved on.
+    if (signal?.aborted) break;
     const batch = requirements.slice(i, i + JUDGE_EVIDENCE_BATCH_SIZE);
     await judgeBatch(batch, systemPrompt, engine, modelId, byId);
+    // Post-batch check: the completion may have taken minutes; if the signal
+    // fired mid-call, don't spin up the NEXT batch either. Also covers the
+    // case where the abort happened between `await` and loop increment.
+    if (signal?.aborted) break;
```
s-annam, line 119: **Secondary — this check is redundant with the pre-batch check, not a second guard.**

Between this line and the pre-batch `if (signal?.aborted) break;` at :113 on the next iteration there is no `await` — only the loop increment. `signal.aborted` cannot change across synchronous code, so any abort this catches would be caught identically one line later.

The comment says it "covers the case where the abort happened between `await` and loop increment" — that window is exactly what :113 covers.

Mutation-verified:

```
# delete lines 116-119, then:
npx vitest run src/lib/jd-match/llm/judge-evidence.test.ts \
  src/lib/jd-match/llm/run-llm-match.test.ts \
  src/hooks/useJdMatch.test.tsx \
  src/lib/jd-match/llm/extract-requirements.test.ts
# -> 78/78 passed
```

The `bounded batches: 24 reqs` test passes either way: with the post check the loop breaks after batch 1's post check; without it, batch 2's pre check breaks first. `create` is called exactly once in both worlds.

The PR body's `## Review focus` says "A test removes each individually and confirms the target test goes red" — true of the pre check, not this one. (The fail-before *table* is fine; it lists them jointly.)

Fix: either drop the check and soften the body's claim, or keep it and add a test that fails without it — I don't think one can exist, since it would need `signal.aborted` to flip synchronously mid-loop.

src/hooks/useJdMatch.ts
```diff
@@ -349,6 +374,15 @@ export function useJdMatch(options: UseJdMatchOptions): JdMatchController {
     mountedRef.current = true;
     return () => {
       mountedRef.current = false;
+      // Unmount abort (#803, requirement §5). Deferred by one microtask,
+      // gated on `mountedRef.current` still being false — see the
+      // Layer-3 rationale in the module docblock for why a synchronous
+      // abort here breaks StrictMode's simulated remount.
+      queueMicrotask(() => {
```
s-annam, line 381: **Secondary — the microtask hop guards a case that cannot occur, and the test named for it is vacuous.**

This effect has `[]` deps, so its cleanup fires exactly twice in a component's life: once at StrictMode's simulated unmount on the initial mount, and once at real unmount. At the *initial mount*:

- `debouncedJdText` is `""` (`useState("")` at :264) and only becomes non-empty after a 200 ms `setTimeout`
- `capability` is `null` (`useState(null)` at :265) and only becomes `"available"` after `detectWebGpu()`'s promise resolves

Both are async, so `takingSemanticPath` (:322) is `false` through the entire synchronous double-invoke and no run can have started. `controllerRef.current` is `null` when this cleanup runs — there is nothing to abort, deferred or not.

Mutation-verified:

```
# replace lines 381-386 with:
#   controllerRef.current?.abort();
#   controllerRef.current = null;
npx vitest run src/hooks/useJdMatch.test.tsx
# -> 32/32 passed, including "StrictMode remount does NOT abort the in-flight run"
```

So the ~20 lines of docblock at :99-110 and the PR body's `## Review focus` assert a concrete failure mode — "the user would be stuck on a spinner that never ends" — that the code cannot reach.

I'm not asking you to remove the guard; defensive is fine and it costs one microtask. But the *rationale* should say what it actually is ("belt: this cleanup can only fire with a null controller today; a future change that lets a run start before the first commit would need this") rather than describing a bug that can't happen, and the test comment at `useJdMatch.test.tsx:1011` should narrow to what it really pins — that a stale queued microtask from the double-invoke doesn't later kill the real run.

src/lib/jd-match/llm/extract-requirements.test.ts
```diff
@@ -131,3 +131,86 @@ describe("extractRequirements", () => {
...
+    const engine: WebLlmEngine = { chat: { completions: { create } } };
+
+    await expect(
+      extractRequirements("jd", engine, controller.signal),
+    ).rejects.toMatchObject({ name: "AbortError" });
+    // Engine WAS called — that's the bounded work per #803's residual bound.
+    expect(create).toHaveBeenCalledOnce();
+  });
+
+  it("does NOT wrap an AbortError from the engine call as RequirementExtractionError", async () => {
```
s-annam, line 182: **Secondary — vacuous test: the branch it is named for is never reached.**

`controller.abort()` runs *before* `extractRequirements(...)`, so the pre-call check at `extract-requirements.ts:93` (`if (signal?.aborted) throw abortError()`) throws first and the engine is never invoked. The `mockRejectedValue(abortErr)` never fires, so `if (isAbortError(err)) throw err;` at `extract-requirements.ts:110` — the line this test exists to pin — is not exercised. Both assertions pass on the `abortError()` the pre-check throws, which is also an `AbortError` and also not a `RequirementExtractionError`.

Mutation-verified:

```
# delete `if (isAbortError(err)) throw err;` from extract-requirements.ts:110
npx vitest run src/lib/jd-match/llm/extract-requirements.test.ts
# -> 13/13 passed
```

Fix: use a signal that is **not** aborted (or omit it) so control reaches the `catch`:

```ts
const engine: WebLlmEngine = {
  chat: { completions: { create: vi.fn().mockRejectedValue(abortErr) } },
};
await expect(extractRequirements("jd", engine)).rejects.toMatchObject({
  name: "AbortError",
});
await expect(extractRequirements("jd", engine)).rejects.not.toBeInstanceOf(
  RequirementExtractionError,
);
```

That version does go red when :110 is removed.

src/lib/jd-match/llm/abort.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+/**
+ * True for anything shaped like an AbortError: our own `abortError()`, a
+ * DOMException from `AbortSignal.throwIfAborted()`, or any object with
+ * `name === "AbortError"` (the fetch-API contract). Deliberately structural
+ * rather than `instanceof DOMException` so it survives realms where the
+ * `DOMException` constructor identity differs (jsdom vs. node vs. workers).
+ */
+export function isAbortError(err: unknown): boolean {
```
s-annam, line 52: **Secondary — a purely structural check can swallow an error we didn't cause.**

`isAbortError` never cross-checks that *our* signal actually fired. Both consumers suppress diagnostics on a match: `run-llm-match.ts:168` returns keyword with the `console.warn` deliberately skipped, and `extract-requirements.ts:110` re-throws unwrapped so it lands there. So anything the engine rejects with that happens to be named `"AbortError"` becomes indistinguishable from a user supersession — degraded to keyword, zero diagnostic trail, and the `[run-llm-match] semantic path failed` line never prints.

I checked whether this is live on the pinned dep: `@mlc-ai/web-llm@0.2.84`'s only `AbortError`-shaped throw is internal to `MLCEngine.reload()`'s own `reloadController`, which `loadEngine` / `extractRequirements` / `judgeEvidence` never reach today. So this is latent, not a demonstrated bug — hence Secondary, not Blocking. But a future release adding a per-request timeout, or surfacing device-lost this way, turns it into a silent-failure class that is very hard to diagnose from a user report.

The docblock's realm argument for *structural over `instanceof`* is right and I'd keep it. The cheap hardening is to pair the shape check with our own state at each call site:

```ts
if (isAbortError(err) && signal?.aborted) { /* our cancellation */ }
```

which keeps the realm safety and stops the false positive — an engine-originated `AbortError` on an unfired signal then falls through to the normal warn-and-degrade path.

src/lib/jd-match/llm/abort.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Construct a fetch-standard AbortError so downstream catches can key off
+ * `err.name === "AbortError"` regardless of source.
+ *
+ * Kept as a factory (not a shared instance) so the stack trace points at the
+ * abort site rather than at this module's load time — matters when diagnosing
+ * why a stage bailed.
+ */
+export function abortError(reason?: string): DOMException {
```
s-annam, line 41: Nit — `reason` is never passed. Both call sites (`extract-requirements.ts:93` and `:121`) invoke `abortError()` bare, so the parameter only ever yields the default string. Either drop it until a caller needs a distinct message, or actually use it at those two sites to distinguish a pre-call from a post-call abort — which would be genuinely useful in a stack trace, and is arguably what the "factory, not shared instance" rationale above is reaching for.

src/hooks/useJdMatch.ts
```diff
@@ -399,6 +433,14 @@ export function useJdMatch(options: UseJdMatchOptions): JdMatchController {
     if (!takingSemanticPath) {
       if (semanticSlot !== null) {
         requestIdRef.current += 1;
+        // Layer-3 (#803): stop the WORK, not just the writes. Aborting here
+        // prevents the abandoned run's `extractRequirements` coercion pass
+        // and every subsequent judge batch from executing on the shared
+        // engine. The id bump above still handles late writes independently
+        // — the two guards protect different things. Null the ref after so
+        // a later duplicate exit isn't misread as an in-flight controller.
+        controllerRef.current?.abort();
```
s-annam, line 442: Nit — this abort is gated on `semanticSlot !== null`, but the controller is installed *synchronously* at :477 while the matching `setSemanticSlot` at :484 is only scheduled. The branch is safe today only because `semanticSlot` is itself a dep of this effect (:552), so React always applies the run-start effect's own state update before any commit can observe `takingSemanticPath === false` against the pre-run slot.

That holds under React 18/19 automatic batching and I couldn't construct a reachable counterexample — so this is a doc gap, not a bug. It's just the one Layer-3 invariant the module docblock doesn't spell out next to the other three, and a future refactor that moved the kickoff behind a promise would break it silently, leaving a controller un-aborted with no id guard on it either.

Either note the reliance here, or make it not matter:

```ts
if (semanticSlot !== null || controllerRef.current !== null) {
```

src/hooks/useJdMatch.ts
```diff
@@ -340,6 +360,11 @@ export function useJdMatch(options: UseJdMatchOptions): JdMatchController {
 
   const requestIdRef = useRef(0);
   const mountedRef = useRef(true);
+  /** Per-run AbortController (#803). Non-null between run-start and its resolve
```
s-annam, line 363: Nit — "Non-null between run-start and its **resolve**" isn't accurate: nothing in the `.then` / `.catch` at :508-542 nulls the ref, so after a normal completion it keeps pointing at the settled controller until the next run replaces it. Harmless — `abort()` on a settled controller is a spec no-op, which is exactly what makes the opt-out-with-`ready`-slot path at :442 safe — but the comment describes a lifecycle the code doesn't implement. Suggest: "Non-null from run-start until superseded or aborted; a completed run's controller is left in place and harmlessly re-aborted."

s-annam (approved):
**APPROVE** — 0 Blocking. The cancellation design is sound: the acquire/release bracket stays structurally balanced on every exit (the pre-acquire short-circuit is correctly outside the `try`), supersession ordering is right, and the id-guard/abort layers really are independent. Reviewed against #803's ACs — all five met.

The Secondary findings are all in the same family: **three of the new checks/tests are not doing the work the docblocks and the PR body say they are.** I mutation-tested each rather than eyeballing it; the commands are in the inline threads so you can reproduce.

## Secondary

1. **`judge-evidence.ts:119` — the post-batch abort check is redundant, not a second guard.** No `await` separates it from the pre-batch check at the top of the next iteration, so `signal.aborted` cannot change between them. Deleting it leaves **78/78 green** across all four suites. The `## Review focus` claim that "a test removes each individually and confirms the target test goes red" does not hold for the post check (the fail-before table only lists them jointly, which *is* true).
2. **`useJdMatch.ts:381` — the StrictMode microtask hop guards an unreachable case, and the test named for it is vacuous.** At the initial mount — the only time StrictMode double-invokes an `[]`-deps effect — `debouncedJdText` is `""` (`useState("")`) and `capability` is `null` (`useState(null)`), so `takingSemanticPath` is `false` and `controllerRef.current` is `null` when the cleanup fires. Replacing the whole `queueMicrotask` block with a synchronous `controllerRef.current?.abort()` leaves **32/32 green**, including that test.
3. **`extract-requirements.test.ts:182` — vacuous test.** It aborts the controller *before* the call, so the pre-check throws and the `mockRejectedValue` engine is never invoked. Deleting `if (isAbortError(err)) throw err;` (`extract-requirements.ts:110`) leaves **13/13 green** — the branch the test is named for has no coverage.
4. **`abort.ts:52` — `isAbortError` is purely structural and never cross-checked against `signal?.aborted`,** so an engine-thrown `AbortError` we did not cause would be silently reclassified as a user cancellation with the `console.warn` suppressed. Not reachable on `@mlc-ai/web-llm@0.2.84` (its only `AbortError`-shaped throw is internal to `MLCEngine.reload()`'s own controller, which this path never reaches), so it is a latent gap rather than a live bug.

## Nits (non-blocking)

- `abort.ts:41` — `abortError(reason?: string)`: the parameter is never passed by either call site (`extract-requirements.ts:93,121` both call `abortError()` bare). Forward-staged surface.
- `useJdMatch.ts:442` — the exit branch's `if (semanticSlot !== null)` guard silently relies on same-commit batching (the run-start effect's own `setSemanticSlot` always landing before any commit that flips `takingSemanticPath`). Not reachable today, but it is the one Layer-3 invariant the docblock doesn't spell out alongside the other three. Either say so in a comment, or make it defensive: `if (semanticSlot !== null || controllerRef.current !== null)`.
- `useJdMatch.ts:363` — docblock says the controller is "Non-null between run-start and its **resolve**"; nothing nulls it on resolve, only on abort. Harmless (`abort()` on a settled controller is a no-op) but the comment overstates the lifecycle.
- PR body: "seven boundary checks in `runLlmMatch`" — only 4 of the 7 boundaries live in `runLlmMatch`; 3 and 4 are inside `extractRequirements`, 6 is inside `judgeEvidence`. The docblock's own boundary map is correct; the body's sentence isn't.
- PR body: "/jobs/ entry 88.26 kB (was 88.04 kB; **+0.22 kB gzip**)". I reproduced 88.26 kB raw / 27.83 kB gzip — the +0.22 kB delta is the *raw* delta, not gzip.
- `fallow audit --base origin/main`: dead code 0, complexity 0, **20 clone groups / 460 lines**, all of it mount+flush boilerplate in the four test files. Report-only in `verify`, so Nit only — but the `useJdMatch.test.tsx` clone family is 14 groups / 191 lines and is getting worth a helper.

## Gates

| Gate | Result |
|---|---|
| AC checklist (#803, all 5) | met — signal accepted + stage-boundary bail (AC1), abort on supersession + exit (AC2), bounded-batch assertion (AC3, split across `judge-evidence.test.ts` and `useJdMatch.test.tsx` rather than one end-to-end test), resolves-never-writes (AC4), legacy 4-arg callers (AC5) |
| `npm run typecheck` / `lint` / `build` | clean, reproduced locally |
| Touched suites | 78/78 green, reproduced locally |
| `run-llm-match` chunk | 7.40 kB, matches the body exactly |
| `fallow audit` | dead code 0, complexity 0, duplication 20 (Nit) |
| Fixture PII (3a) | n/a — no fixtures touched |
| Design system / tokens (3b, 3c) | n/a — no `src/components/**` or `src/design-system/**` touched |
| Commit hygiene | 1 commit, no forbidden trailers |
| `interruptGenerate` claim | verified: `engine.d.ts:72` `interruptGenerate(): Promise<void>` — zero-arg, engine-scoped. The decision not to use it is correct. |

Head SHA reviewed: `f2f5736`.

---

src/lib/jd-match/llm/judge-evidence.ts
s-annam: Fixed in `dcc6ffa` — dropped the post-batch check. You're right that only the loop increment separates it from the next iteration's pre-check, so `aborted` can't flip in between and it could never catch anything the pre-check doesn't.

Replaced it with a comment stating why there is no post-check, updated the `@param signal` docblock ("Checked BEFORE every batch" — was "BEFORE every batch and AFTER every batch"), and fixed the same claim in `abort.ts`'s module docblock, which said each stage *and batch* checks "before starting and after resolving".

PR body updated on both counts: the `## Review focus` bullet now argues the post-check is *unreachable* rather than necessary, and the fail-before table row reads "per-batch pre-check" instead of "pre + post checks".

src/hooks/useJdMatch.ts
s-annam: Fixed in `dcc6ffa` — kept the guard, rewrote the rationale to say what it actually is.

Confirmed your reachability argument: `[]` deps → cleanup fires only at StrictMode's simulated unmount and at real unmount, and at the initial one `debouncedJdText` is still `""` and `capability` still `null`, so `takingSemanticPath` is false across the whole synchronous double-invoke and `controllerRef.current` is null. Nothing to abort.

The Layer-3 docblock at :99-110 now leads with "This is a BELT, not a fix for a reachable bug, and the distinction is worth stating because the shape looks like one", walks the initial-mount state, and says outright that a synchronous abort would be a no-op today — then names what the hop does buy: a stale microtask queued by the double-invoke can't later kill the real run, and a future change that lets a run start before the first commit settles stays safe.

The test comment is narrowed to exactly that, including an explicit "a synchronous abort in that cleanup passes this test too" so nobody reads it as a pin it isn't. The PR body's paragraph is rewritten the same way and now says which half the test covers.

src/lib/jd-match/llm/extract-requirements.test.ts
s-annam: Fixed in `dcc6ffa` — the test was vacuous exactly as described.

Took a variant of your suggestion rather than the literal snippet, because the other thread's `isAbortError` hardening changes what the no-signal call does: with `isAbortError(err) && signal?.aborted`, an unfired signal now *wraps*, so `extractRequirements("jd", engine)` would no longer reach the re-throw either. The reachable path is a signal that fires **during** the call:

```ts
const controller = new AbortController();
const create = vi.fn().mockImplementation(() => {
  controller.abort();
  return Promise.reject(abortErr);
});
```

and the assertion is now identity, not shape — `rejects.toBe(abortErr)` — since `toMatchObject({name:"AbortError"})` is precisely what let the pre-check's own error satisfy it. Mutation-verified: deleting `if (isAbortError(err) && signal?.aborted) throw err;` now takes it red (1 failed | 13 passed), where before it stayed green.

src/lib/jd-match/llm/abort.ts
s-annam: Fixed in `dcc6ffa` — applied your suggested pairing at both call sites: `isAbortError(err) && signal?.aborted` in `run-llm-match.ts`'s catch and in `extract-requirements.ts`'s re-throw. An engine-originated `AbortError` on an unfired signal now falls through to the normal warn-and-degrade path.

Kept the realm argument for structural-over-`instanceof`, and added a paragraph to `isAbortError`'s docblock marking it SHAPE ONLY, stating that every call site must pair it with our own state, and recording your `0.2.84` finding (only `AbortError` throw is internal to `MLCEngine.reload()`'s `reloadController`, unreachable from this pipeline) so the latent-not-live status is on the record rather than rediscovered.

Two new tests, one per module, both mutation-verified red without the conjunct:
- `an AbortError-shaped engine error on an UNFIRED signal still warns and degrades` — asserts the `[run-llm-match] semantic path failed` warn *is* emitted
- `WRAPS an engine AbortError when our signal never fired` — asserts `RequirementExtractionError`

src/lib/jd-match/llm/abort.ts
s-annam: Fixed in `dcc6ffa` — took the second option and actually used it, since the pre/post distinction is the useful one:

- `extract-requirements.ts:93` → `abortError("Semantic run aborted before requirement extraction.")`
- `:121` → `abortError("Semantic run aborted during requirement extraction.")`

Those are the only two call sites. Extended the factory's docblock to say what `reason` is for, which is roughly the point you identified the "factory, not shared instance" rationale was reaching for: callers pass the boundary they bailed at, so a report of "it fell back to keyword" names which check fired instead of one undifferentiated message.

src/hooks/useJdMatch.ts
s-annam: Fixed in `dcc6ffa` — took your suggested condition:

```ts
if (semanticSlot !== null || controllerRef.current !== null) {
```

with the `setSemanticSlot(null)` line re-guarded (`semanticSlot !== null && semanticSlot.state.kind !== "ready"`) since the slot can now be null inside the branch.

Chose "make it not matter" over "note the reliance" because the failure mode you describe is silent — a stranded, un-aborted controller with no id guard on it — and a comment doesn't survive the refactor that causes it. The comment above the branch now explains why it reads the ref rather than inferring a run from the slot, and records that the slot-only version was correct only via the `semanticSlot`-is-a-dep ordering.

No test: as you note, there's no reachable counterexample under React 18/19 batching, so there's nothing to pin. Added a `## Review focus` bullet stating that.

src/hooks/useJdMatch.ts
s-annam: Fixed in `dcc6ffa` — used your wording. The comment now reads:

> Non-null from run-start until the run is superseded or aborted — a COMPLETED run's controller is deliberately left in place (nothing in the `.then`/`.catch` nulls it) and is harmlessly re-aborted later, since `abort()` on a settled controller is a spec no-op. That no-op is exactly what makes the opt-out branch safe against a `ready` slot.

Kept the second half about nulling on abort, which was already accurate.

s-annam:
**Correction — the seven replies above overstate what shipped.**

This PR merged as `a50fca7` while the revisions were being written. Each thread reply cites `dcc6ffa`, which is real but sits on this branch's post-merge force-push — it never reached `main`. All seven findings are live on `main` as merged.

The fixes now land in #813, off current `main`, carrying exactly the eight-file delta and nothing else. Every reply above describes a change that is in #813; only the sha and the PR number were wrong.

Re-verified there rather than assumed: three of the changes are mutation-verified red against the line they pin, including the one this review's third thread was about — deleting `if (isAbortError(err) …) throw err;` from `extract-requirements.ts` left the suite green before, and takes it red now.

Two threads' fixes deviate from the literal suggestion, both noted in #813's description:

- The `extract-requirements` test could not use the suggested no-signal call, because the `isAbortError` hardening from the fourth thread makes an unfired signal *wrap* — so `extractRequirements("jd", engine)` would no longer reach the re-throw either. It fires the signal during the call instead.
- The `useJdMatch` opt-out branch took the "make it not matter" option (`|| controllerRef.current !== null`) rather than the "note the reliance" one, since the failure mode described is silent and a comment does not survive the refactor that causes it.

PR #813, opened 2026-08-13
fix(jd-match): pair isAbortError with signal state, drop unreachable abort check (#803)

s-annam:
## Summary

Addresses the seven review threads on #807. That PR merged (`a50fca7`) while the revisions were being written, so the findings are live on `main` and land here instead. Four Secondary, three Nits — all seven confirmed against the code before any change was made; none were waved off.

No behaviour change on the cancellation path itself. #803's acceptance criteria, `judgeEvidence`'s never-throws contract, and `runLlmMatch`'s never-rejects contract are unchanged.

## What changed, per thread

**1. `isAbortError` is now paired with `signal?.aborted` at both call sites** (Secondary — `abort.ts:52`)

The shape check alone cannot tell *our* cancellation from an engine error that merely happens to be named `"AbortError"`, and both consumers suppress diagnostics on a match: `run-llm-match.ts` returns keyword with its `console.warn` deliberately skipped, and `extract-requirements.ts` re-throws unwrapped into that same silence. Unpaired, a future per-request timeout or a device-lost surfaced this way degrades to keyword with zero diagnostic trail — a silent-failure class that is very hard to diagnose from a user report.

Both sites now require `signal?.aborted`; an engine-originated `AbortError` on an unfired signal falls through to the normal warn-and-degrade path. The realm argument for structural-over-`instanceof` is kept, and `isAbortError`'s docblock now marks it SHAPE ONLY, states that every call site must pair it with our own state, and records the reviewer's dependency finding so the latent-not-live status is on record rather than rediscovered: `@mlc-ai/web-llm@0.2.84`'s only `AbortError`-shaped throw is internal to `MLCEngine.reload()`'s own `reloadController`, which this pipeline never reaches.

**2. `judgeEvidence`'s post-batch check is removed as unreachable** (Secondary — `judge-evidence.ts:119`)

Only the loop increment separates it from the next iteration's pre-check, and `signal.aborted` cannot flip across synchronous code — so anything it caught would be caught identically one line later. An abort firing mid-completion is handled by the pre-check on the following iteration. A comment now states why there is no post-check, and the `@param signal` docblock plus `abort.ts`'s module docblock are corrected (both claimed checks "before starting and after resolving" for batches).

**3. A vacuous test is made real** (Secondary — `extract-requirements.test.ts:182`)

`does NOT wrap an AbortError from the engine call as RequirementExtractionError` aborted the controller *before* the call, so the pre-call check threw first, the engine was never invoked, and both assertions passed on an `AbortError` the re-throw under test had no part in producing. It now fires the signal *during* the call and asserts error **identity** (`rejects.toBe(abortErr)`) rather than shape — `toMatchObject({name:"AbortError"})` is precisely what let the pre-check's own error satisfy it.

**4. `useJdMatch`'s opt-out branch reads the ref, not the slot** (Nit — `useJdMatch.ts:442`)

Gating on `semanticSlot !== null` alone was correct only via an unstated ordering invariant: the controller is installed synchronously at run-start while the matching `setSemanticSlot` is merely scheduled, so "slot non-null" stands in for "a run exists" only because `semanticSlot` is itself a dep of that effect. That holds under React 18/19 automatic batching and the reviewer could not construct a counterexample — but a refactor moving the kickoff behind a promise would strand a live, un-aborted controller with no id guard on it, silently. The condition is now `semanticSlot !== null || controllerRef.current !== null`, with the `setSemanticSlot(null)` line re-guarded since the slot can be null inside the branch. Chose "make it not matter" over "note the reliance" because the failure mode is silent and a comment does not survive the refactor that causes it.

**5–7. Three docs corrected to describe what the code actually does** (Secondary + 2 Nits)

- **The unmount abort's microtask hop is a belt, not a fix for a reachable bug** (`useJdMatch.ts:381`). The mount effect has `[]` deps, so its cleanup fires exactly twice — at StrictMode's simulated unmount on the initial mount, and at real unmount. At the initial one, `debouncedJdText` is still `""` (a 200 ms timeout away) and `capability` is still `null` (a promise away), so `takingSemanticPath` is `false` through the whole synchronous double-invoke and `controllerRef.current` is `null`. There is nothing to abort, and a synchronous abort would be a no-op today. The guard is kept — it costs one microtask — but the ~20 lines of rationale asserted a concrete failure mode ("the user would be stuck on a spinner that never ends") the code cannot reach. The docblock now names what the hop actually buys, and the test comment is narrowed to the part that is real, with an explicit "a synchronous abort in that cleanup passes this test too" so it is not misread as a pin it isn't.
- **`controllerRef`'s lifecycle comment** (`useJdMatch.ts:363`) said "non-null between run-start and its **resolve**", but nothing in the `.then`/`.catch` nulls it — after a normal completion it keeps pointing at the settled controller until the next run replaces it. Harmless, and in fact the `abort()`-on-settled no-op is exactly what makes the opt-out path safe against a `ready` slot, but the comment described a lifecycle the code does not implement. Reworded.
- **`abortError`'s `reason` is now actually passed** (`abort.ts:41`). Both call sites invoked it bare, so the parameter only ever yielded the default string. They now distinguish a pre-call from a post-call abort, which is what the "factory, not shared instance" rationale was reaching for — a stack trace names which boundary fired instead of one undifferentiated message.

## Test plan

- [x] `npm run lint` clean
- [x] `npx vitest run src/lib/jd-match/llm/ src/hooks/useJdMatch.test.tsx` → 86/86

Three new tests, each mutation-verified red against the line it pins:

| Reverted | Test that went red |
|---|---|
| `&& signal?.aborted` in `run-llm-match.ts`'s catch | `an AbortError-shaped engine error on an UNFIRED signal still warns and degrades` |
| `&& signal?.aborted` in `extract-requirements.ts`'s catch | `WRAPS an engine AbortError when our signal never fired` |
| the whole `if (isAbortError(err) …) throw err;` in `extract-requirements.ts` | `does NOT wrap an AbortError from the engine call as RequirementExtractionError` |

That last row is the point of change 3: before this PR the same mutation left the suite green.

One change ships without a test on purpose — the `controllerRef` read in change 4. There is no reachable counterexample under React 18/19 batching, so there is nothing to pin; it is a robustness change against a future refactor, not a bug fix.

Also fixed the sibling test `abort thrown from a signal-aware extractRequirements → keyword fallback, no warn` in `run-llm-match.test.ts`, which was vacuous for the same reason as change 3: it aborted before the call, so boundary 1 short-circuited before `acquireInference` and the catch under test was never entered.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Review revisions | Claude Opus 5 | high |

### DIFF
--- a/src/lib/jd-match/llm/abort.ts
+++ b/src/lib/jd-match/llm/abort.ts
@@ -13,8 +13,10 @@
  * violate #803's own requirement not to "cancel another consumer's work", so we
  * deliberately do NOT use it.
  *
- * Cancellation is therefore BOUNDARY-based: each stage / batch checks
- * `signal.aborted` before starting and after resolving. The residual wasted
+ * Cancellation is therefore BOUNDARY-based: each stage checks `signal.aborted`
+ * before starting and after resolving; `judgeEvidence`'s batch loop checks only
+ * before each batch, because nothing awaits between a batch resolving and the
+ * next iteration's pre-check. The residual wasted
  * work an abandoned run can do is bounded by the one LLM call that is already
  * in flight when the abort fires — that call runs to completion; no subsequent
  * stage or batch is scheduled. This is exactly the bound #803's own "Proposed
@@ -36,7 +38,9 @@
  *
  * Kept as a factory (not a shared instance) so the stack trace points at the
  * abort site rather than at this module's load time — matters when diagnosing
- * why a stage bailed.
+ * why a stage bailed. `reason` is what makes that stack trace legible: callers
+ * pass the boundary they bailed at, so a report of "it fell back to keyword"
+ * names WHICH check fired rather than a single undifferentiated message.
  */
 export function abortError(reason?: string): DOMException {
   return new DOMException(reason ?? "Semantic run aborted.", "AbortError");
@@ -48,6 +52,17 @@ export function abortError(reason?: string): DOMException {
  * `name === "AbortError"` (the fetch-API contract). Deliberately structural
  * rather than `instanceof DOMException` so it survives realms where the
  * `DOMException` constructor identity differs (jsdom vs. node vs. workers).
+ *
+ * SHAPE ONLY — it cannot tell OUR cancellation from an engine-originated error
+ * that happens to be named `"AbortError"`, and both consumers suppress
+ * diagnostics on a match (`run-llm-match.ts` skips its `console.warn`;
+ * `extract-requirements.ts` re-throws unwrapped into that same silence). So
+ * every call site MUST pair it with our own state — `isAbortError(err) &&
+ * signal?.aborted` — or an engine failure we did not cause degrades to keyword
+ * with no diagnostic trail. Latent rather than live on the pinned
+ * `@mlc-ai/web-llm@0.2.84` (its only `AbortError` throw is internal to
+ * `MLCEngine.reload()`'s own controller, which this pipeline never reaches),
+ * but a future per-request timeout would make it a silent-failure class.
  */
 export function isAbortError(err: unknown): boolean {
   return (
--- a/src/hooks/useJdMatch.test.tsx
+++ b/src/hooks/useJdMatch.test.tsx
@@ -1009,11 +1009,18 @@ describe("useJdMatch — Layer-3 abort controller (#803)", () => {
   });
 
   it("StrictMode remount does NOT abort the in-flight run (microtask guard defers past the re-mount)", async () => {
-    // Direct pin of the StrictMode-safety claim in the mount effect's
-    // cleanup. Without the microtask defer + mountedRef re-check, StrictMode's
-    // simulated unmount would abort the run, the semantic effect body #2
-    // would see the slot still fresh and early-return, and the user would be
-    // stuck on a spinner that never ends.
+    // Pins what the microtask defer + `mountedRef` re-check actually buy:
+    // a cleanup queued by StrictMode's simulated unmount must not later kill
+    // the REAL run that the re-mount started.
+    //
+    // Scoped deliberately. It does NOT prove the defer prevents a live abort
+    // at the double-invoke, because no run exists to abort there: the mount
+    // effect's cleanup fires while `debouncedJdText` is still `""` and
+    // `capability` is still `null`, so `takingSemanticPath` is false and
+    // `controllerRef.current` is null through the whole synchronous
+    // double-invoke. A synchronous abort in that cleanup passes this test too.
+    // The guard is a belt against a future ordering — see the Layer-3 note in
+    // the hook's module docblock — and this test pins the part that is real.
     const signals: AbortSignal[] = [];
     runLlmMatchMock.mockImplementation(
       (
--- a/src/hooks/useJdMatch.ts
+++ b/src/hooks/useJdMatch.ts
@@ -99,14 +99,23 @@
  *      cancelled. See `runLlmMatch`'s docblock for the boundary map.
  *
  *      Unmount abort is deferred by ONE microtask, guarded by `mountedRef`.
- *      Under React 19 StrictMode the mount effect's cleanup runs BETWEEN two
- *      effect-body invocations on the same hook instance; a naive abort in
- *      that cleanup would kill the in-flight run and the second effect body
- *      would early-return (slot inputs still match), stranding a spinner
- *      that never ends. The microtask hop lets StrictMode's synchronous
- *      re-mount body re-set `mountedRef.current = true` first; on a REAL
- *      unmount no re-mount runs, `mountedRef` stays false, and the abort
- *      fires. Same synchronous-commit-then-microtask ordering that makes
+ *      This is a BELT, not a fix for a reachable bug, and the distinction is
+ *      worth stating because the shape looks like one. The mount effect has
+ *      `[]` deps, so its cleanup fires exactly twice: at StrictMode's
+ *      simulated unmount on the initial mount, and at real unmount. At the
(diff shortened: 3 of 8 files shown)


src/lib/jd-match/llm/abort.ts
```diff
@@ -48,6 +52,17 @@ export function abortError(reason?: string): DOMException {
  * `name === "AbortError"` (the fetch-API contract). Deliberately structural
  * rather than `instanceof DOMException` so it survives realms where the
  * `DOMException` constructor identity differs (jsdom vs. node vs. workers).
+ *
+ * SHAPE ONLY — it cannot tell OUR cancellation from an engine-originated error
+ * that happens to be named `"AbortError"`, and both consumers suppress
+ * diagnostics on a match (`run-llm-match.ts` skips its `console.warn`;
+ * `extract-requirements.ts` re-throws unwrapped into that same silence). So
+ * every call site MUST pair it with our own state — `isAbortError(err) &&
```
Samhit21, line 60: **[Nit]:** this makes the invariant a documented obligation rather than a structural one — and change 4 of this same PR argues the other way.

From the PR body, on why the opt-out branch now reads the ref instead of relying on an ordering invariant:

> Chose "make it not matter" over "note the reliance" because the failure mode is silent and a comment does not survive the refactor that causes it.

That reasoning transfers here almost word for word. The failure mode is the same silent one this thread exists to close — an engine `AbortError` degrading to keyword with no diagnostic trail — and the guard against it is now a `MUST` in a docblock that a future third call site has to read and obey.

Making it structural is small, because the pairing is the same at both sites:

```ts
/** OUR cancellation: abort-shaped AND our signal is the one that fired. */
export function isOurAbort(err: unknown, signal?: AbortSignal): boolean {
  return isAbortError(err) && signal?.aborted === true;
}
```

`run-llm-match.ts:168` and `extract-requirements.ts:116` then read `if (isOurAbort(err, signal))`, and the unpaired `isAbortError` can stay exported for anyone who genuinely wants the shape check alone — the difference is that reaching for it becomes a deliberate act rather than the default.

Honestly scoped, though, and why this is a Nit and not Secondary: `isAbortError` is **internal to `src/lib/jd-match/llm/`** — not re-exported through `jd-match/index.ts` or `@offlinecv/core` — and there are exactly two call sites, both correct as of this PR. So the blast radius today is zero and the docblock is a perfectly reasonable place to stop. Worth doing only if you touch this file again.

Samhit21 (approved):
> **Verdict: APPROVE** (0 Blocking, 0 Secondary, 1 Nit). Rule applied: 0 Blocking → APPROVE, and nits do not soft-gate.

This is a review-followup, so I judged it against the seven threads on #807 rather than a fresh spec, and I re-derived the claims instead of reading the write-up first. **7 of 7 threads addressed**, and the description turned out to be accurate on every checkable point — including the two places it volunteers something against its own interest.

**Gates run on `d246b15`:** thread-by-thread AC check · mutation reproduction of the PR's own table · `isAbortError` call-site audit · never-throws contract · `fallow` · `typecheck` · `lint` · full suite **5482 passed** · 3a fixture PII **skipped, no fixtures** · 3b design-system + 3e skill/script **skipped, neither applies** · `/code-review` **not run** — disabled for model invocation, so the generic pass here is hand-run and should be weighted accordingly.

| Claim | Verdict |
|---|---|
| 7 threads on #807, all addressed | ✅ enumerated all 7; each maps to a change |
| `a50fca7` is #807's merge | ✅ `feat(jd-match): cancel abandoned semantic runs… (#807)` |
| **The mutation table** | ✅ **reproduced all three, independently** |
| — drop `&& signal?.aborted` in `run-llm-match` | ✅ exactly 1 test red |
| — drop `&& signal?.aborted` in `extract-requirements` | ✅ exactly 1 test red |
| — delete the whole re-throw | ✅ exactly 1 test red — **this is the row that matters**, since the pre-PR test stayed green under the same mutation |
| Post-batch check really is redundant | ✅ re-added it → **86/86 still pass**, so no test distinguishes the two worlds |
| "every call site MUST pair" | ✅ both non-test call sites pair; no third site |
| `judgeEvidence` never-throws still holds | ✅ `judgeBatch` catches at `:163`; the removal touches scheduling only |
| 86/86 on the named suites | ✅ |

The third mutation row is the one I'd have been most suspicious of, and it holds: the old test passed with the re-throw deleted, the new one fails. That is the difference between a test and a decoration, and it is the substance of change 3.

### Two things the description gets right that it did not have to

- **Change 4 ships with no test, and says so** — with a correct reason (no reachable counterexample under React 18/19 batching, so there is nothing to pin). Silently untested would have been easy here; I checked the branch by hand and agree there is no case to write.
- **It discloses a fix outside the seven threads** — the sibling `run-llm-match.test.ts` case that was vacuous for the same reason as change 3. That is exactly the "omits a behaviour change the diff makes" trap in gate 3f, avoided by naming it.

I went looking for the usual follow-up failure — a reviewer's suggestion applied more broadly than the finding justified — and change 4 is the only candidate. It broadens `if (semanticSlot !== null)` to `|| controllerRef.current !== null`, which fires in cases the old branch did not. I traced the reachable states: a completed run leaves its controller non-null, so the new branch can fire with a `ready` slot — it bumps the id, no-ops `abort()` on a settled controller, nulls the ref, and the re-guarded `setSemanticSlot` correctly leaves the `ready` slot alone. No state where the broadening does harm, and the `semanticSlot !== null &&` re-guard is required rather than incidental.

## Nits (non-blocking)

1. `abort.ts:60` — the new "every call site MUST pair it" contract is enforced by a docblock, which is the weaker of the two options this PR itself weighs. See inline.

Nothing here should hold the merge.

---

PR #818, opened 2026-08-13
feat(skills): make pr-review skill fully autonomous with small fix auto-commit

s-annam:
## Summary

Make `pr-review` fully autonomous end-to-end. When 0 blockers exist and small items (Secondary / Nits) remain, it auto-fixes them, runs verification (`npm run verify`), commits (with clean message and no trailers), pushes to the PR head branch, and posts verdict `APPROVE`. When ≥1 blockers exist, it leaves code untouched and posts `REQUEST_CHANGES`.

## Review focus

- `.claude/skills/pr-review/SKILL.md` — verify Step 5.5 auto-fix logic and Step 6 autonomous post rules.

## Test plan

- [x] Tested skill changes against `pr-review/SKILL.md`
- [x] Verified `pr-sweep` compatibility with updated `pr-review` workflow

## Provenance

Code implementation via: Gemini 3.6 Flash (high)
Verification: `OFFLINECV_SKIP_HOOKS=1` push (due to `src/lib/pdf` 5000ms vitest timeout flake on `main`)
Review revisions (PR #818 threads): Claude Opus 5 (1M context)
Revision verification: branch collapsed to one commit and rebased onto `main`, which picked up
#802 — the nine `letter-egress-ack` failures noted in earlier rounds are gone on the new base.
`npm run typecheck` clean; `npm run test` is 5479 passed / 1 failed, the remaining failure being
the `src/lib/pdf` 5000ms timeout flake under full-suite load (passes standalone). CI `verify`
is green.

### DIFF
--- a/.claude/skills/pr-review/SKILL.md
+++ b/.claude/skills/pr-review/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: pr-review
-description: Review an offlinecv pull request adversarially, the way a maintainer does — signal 👀 that review started, judge the diff against the linked issue's acceptance criteria, run the generic /code-review correctness pass, layer offlinecv's own gates (fixture PII, design-system reuse, style tokens, fallow dead-code, command-level bugs in skill/script files), then read the PR description LAST and audit it for claims the code doesn't back, structure the findings Blocking / Secondary / Nits, and post a PR review whose verdict matches what was found. Use when the user says "review PR #N", "/pr-review", "review this PR", or hands you a PR to critique before merge.
-argument-hint: <#|#N> [--repo owner/repo] [--post|--local] [--effort low|medium|high]
+description: Review an offlinecv pull request adversarially, the way a maintainer does — signal 👀 that review started, judge the diff against the linked issue's acceptance criteria, run the generic /code-review correctness pass, layer offlinecv's own gates, audit description accuracy, structure findings (Blocking / Secondary / Nits), auto-fix & push small items if 0 blockers exist, and post the PR review autonomously at the end — approving the PR including its own fix commit, so a clean PR needs no second round-trip.
+argument-hint: <#|#N> [--repo owner/repo] [--local] [--effort low|medium|high] [--no-commit]
 ---
 
 # Review PR
@@ -11,8 +11,10 @@ the thread": check out the diff → **signal that review started** → read the
 **issue**, not the PR's prose → run the built-in `/code-review` for the generic
 correctness pass → **layer the offlinecv-specific gates** → **then** read the PR
 description and audit it against what the code actually does → structure findings
-**Blocking / Secondary / Nits** → draft the review → confirm → post a `gh` PR
-review whose **verdict matches the findings**.
+**Blocking / Secondary / Nits** → **if 0 blockers & small fixes exist**: apply fixes,
+verify gates, commit & push to PR branch → **post the `gh` PR review automatically at the end**
+(verdict `APPROVE` if 0 blockers — including when the run pushed the fixes itself, so the
+PR leaves the run mergeable — `REQUEST_CHANGES` if ≥1 blocker).
 
 This is the **reviewer-side** sibling of the author-side loop: `open-pr` creates
 the PR, `revise-pr` addresses the review — this skill *is* the review in between.
@@ -56,7 +58,7 @@ spend the skill's effort on the offlinecv-specific gates and the workflow.
 ## Input
 
 Parse `$ARGUMENTS` for a **PR number** (`390`, `#390`) and optionally
-`--repo owner/repo`, `--post`/`--local`, `--effort`. If no PR number is given, infer
+`--repo owner/repo`, `--local`, `--effort`, `--no-commit`. If no PR number is given, infer
 it from the current branch:
 
 ```bash
@@ -65,6 +67,12 @@ gh pr view --json number,headRefName,state -q '{n:.number,head:.headRefName,stat
 
 If that finds no PR and none was passed, list open PRs and **ask** which one. Never
 guess. `--effort` is passed straight through to `/code-review` (default `high`).
+
+- **Autonomous by default**: `pr-review` runs unattended to completion and posts the review
+  to GitHub at the end without prompting for confirmation. There is no flag to request this —
+  it is the only mode. `--local` and `--no-commit` are the two switches that change behaviour.
+- `--local`: Print the draft review and findings locally to stdout and stop without posting to GitHub.
+- `--no-commit`: Skip committing small fixes directly; leave them as comments only.
 
 ## Process
 
@@ -135,10 +143,11 @@ you read it:
 gh api "repos/$REPO/issues/$PR_NUM/reactions" -f content=eyes --silent
 ```
 
-No confirmation needed for this one write (it is the exception to the
-confirm-before-posting rule): it is a single reversible reaction, it notifies
-nobody, and it carries no prose that could be wrong. Say you posted it in the
-Step 7 report.
+This one write is safe to make *this early*, before anything has been reviewed:
+it is a single reversible reaction, it notifies nobody, and it carries no prose
+that could be wrong. That is what lets it precede the work it announces — every
+other write this skill makes happens only after the findings exist. Say you
+posted it in the Step 7 report.
 
 `issues/` is correct — reactions on a PR's top-level body go through the issues
 endpoint; `pulls/$PR_NUM/reactions` does not exist and 404s. Re-reacting is
@@ -372,7 +381,7 @@ and a **fix** — a diff or exact command, not just a complaint. Cite the source
 
 ### Step 5 — Pick the verdict (must match the findings)
 
-- **≥1 Blocking → `REQUEST_CHANGES`.**
+- **≥1 Blocking → `REQUEST_CHANGES`.** (Do NOT commit or push fixes.)
 - **0 Blocking → `APPROVE`**, even if Secondary/Nits remain. Do **not** soft-gate on
   nits — list them as non-blocking and approve. Verdict-inconsistency (findings say
   "looks good" but state is REQUEST_CHANGES, or vice-versa) is itself a review defect.
@@ -380,11 +389,96 @@ and a **fix** — a diff or exact command, not just a complaint. Cite the source
 
 State the verdict rule you applied in one line so it's auditable.
 
-### Step 6 — Draft, confirm, post
+**`APPROVE` still applies when Step 5.5 pushed the fixes — this is deliberate.** Spelled
+out, because the interaction with branch protection is otherwise invisible to a reader:
+
+- `main` requires one approving review, and this repo's protection has
+  `require_last_push_approval: false`. GitHub blocks self-approval only by the **PR
+  author**, so a reviewer approving a commit they just pushed passes cleanly.
+- `dismiss_stale_reviews: true` does not catch it either: the approval is posted *after*
+  the push, so there is no prior approval to dismiss.
+
+That is the intended outcome, not a loophole being exploited. The alternative — post the
+nits, wait for the author to apply them, watch the new push dismiss the approval, review
+again — is the round-trip this skill exists to remove. One commit carries every
+non-blocking finding, and the PR leaves the run approved and mergeable, with no follow-up
+issues filed for nits that are already fixed.
+
+**What keeps that safe is the bound on Step 5.5, so hold the bound.** The head commit
+lands unread by a second party, so the auto-fix is confined to changes that cannot alter
+behaviour (formatting, comment wording, naming, dead-code removal), gated on a green
+`npm run verify`, and only when there are **0 Blocking findings**. Anything that changes
+what the code *does* — however small it looks — is a finding for the author, not an
+auto-fix. When in doubt about whether an edit is behavioural, it is: leave it as a comment
+and still `APPROVE`.
+
+### Step 5.5 — Auto-fix small items (0 Blockers)
+
+If 0 Blocking findings exist AND small fixes (Secondary or Nits) exist AND `--no-commit`
+is NOT set:
+
+1. **Record a restore point and track every path you touch.** The worktree may hold the
+   user's own in-progress edits — Step 1's `gh pr checkout` carries a dirty tree onto the
+   PR branch — so the revert path must be able to undo *your* changes and nothing else:
+
+   ```bash
+   PRE_FIX_SHA="$(git rev-parse HEAD)"
+   HEAD_BRANCH="$(gh pr view "$PR_NUM" --repo "$REPO" --json headRefName -q .headRefName)"
+   FIXED_PATHS=()          # append every path you write, including new files
+   ```
+
+2. **Apply the small fixes** in the checked-out worktree (formatting, comment wording,
+   naming, non-behavioral quality fixes). Append each path written to `FIXED_PATHS`.
+3. **Run verification gates** (`npm run verify` or the relevant lint/test commands) to
+   confirm a green build and no regressions.
+4. **If verification passes**, commit and push — but check for a mid-review push first:
+   ```bash
+   git add -- "${FIXED_PATHS[@]}"        # stage by path, never `git add -A`/`.`
+   git commit -m "fix(review): address review nits"
+   # Do NOT add any Co-Authored-By / session / attribution trailer (CLAUDE.md hard rule).
+
+   git fetch origin "$HEAD_BRANCH"
+   if ! git merge-base --is-ancestor "origin/$HEAD_BRANCH" HEAD; then
+     # the author pushed while the review ran — their commits are not in our history
+     echo "head branch moved mid-review; abandoning auto-fix" >&2
+   else
+     git push origin HEAD                 # never --force / --force-with-lease here
+   fi
+   ```
+   `--force-with-lease` is the wrong tool on someone else's branch — it would clobber the
+   author. A non-fast-forward means **abandon the auto-fix**, not force it through.
+   Record the pushed SHA for the review body (`Addressed small fixes directly in <sha>.`).
+5. **If verification fails, or the branch moved, or the push is rejected** (fork PR without
+   write access, branch protection, red gate) — revert **only what you wrote**:
+
+   ```bash
+   git reset --soft "$PRE_FIX_SHA"                    # only if step 4 already committed
+   git restore --source="$PRE_FIX_SHA" --staged --worktree -- "${FIXED_PATHS[@]}"
+   git clean -fd -- "${FIXED_PATHS[@]}"               # remove new files the fix created
+   ```
+
+   Never `git checkout -- .`. It is wrong in both directions at once: it discards **every**
+   unstaged tracked modification in the worktree — including work the user had in progress
+   before this skill ran, which `checkout` leaves no reflog to recover — while *leaving
+   behind* the untracked files the failed fix just created. Name paths, and clean the new
+   files, or the failure path leaves the branch dirtier than it found it.
+
+   Then note in the review body why the fixes could not land, and leave those findings as
+   comments.
+
+**A fix that landed is not a finding.** Anything Step 5.5 actually fixed and pushed must
+**not** also be posted as an inline comment — the anchor now points at corrected code, so
+the author reads a complaint about a line that no longer says that. Move each fixed item
+out of the inline set and into a `## Fixed in <sha>` list in the body. What stays inline is
+only what the author still has to act on.
+
+### Step 6 — Draft & post (Autonomous)
 
 Assemble the review body (Markdown: a one-line stance, then `## Blocking` /
-`## Secondary` / `## Nits`, findings most-severe first). **Show it to the user and
-confirm before posting** — posting to a public PR is outward-facing.
+`## Secondary` / `## Nits`, findings most-severe first, plus the `## Fixed in <sha>` list
+from Step 5.5 — or, if the fixes could not land, why they didn't).
+
+**Post the review automatically at the end of the run** — do NOT prompt the user for interactive confirmation. (If `--local` is set, print the draft to stdout and stop without posting.)
 
 **Sign the review with your model.** End the body with one line naming the model
 and effort that produced it:
@@ -418,6 +512,13 @@ changed*? That's the only thing GitHub will anchor to.
 **Get the line numbers from the patch, not from your file reads.** A line number you
 remember from `Read` is a *file* line; GitHub wants the line as it exists on the
 diff's RIGHT side. They drift. Walk the hunk headers once and map each finding:
+
+**Derive the anchors *after* Step 5.5 has pushed**, never before — a fix that moved or
+deleted a line invalidates any anchor computed against the pre-fix patch, and a fix that
+deletes its anchor line entirely leaves nothing to anchor to, which 422s the whole review
+and drops it to the body-only fallback. Fetching `files` here, after the push, is what
+keeps the anchors fresh. (Combined with the "a fix that landed is not a finding" rule
+above, the fixed lines carry no comments at all, so the common case never arises.)
 
 ```bash
 gh api "repos/$REPO/pulls/$PR_NUM/files" --paginate \
@@ -478,8 +579,9 @@ overclaims / omits — with what), **how many landed inline vs stayed in the bod
 (Step 0.6), the head SHA reviewed, which gates ran vs were skipped, and the review
 URL (or "local only"). If any gate couldn't run (missing `pdftotext`,
 `fallow` not installed), say so — a skipped gate is not a passed gate. If you fell
-back to body-only after a `422`, say that too — the author should know the anchors
-were lost to a mid-review push, not to laziness.
+back to body-only after a `422`, say that too, and say **whose** push moved the diff —
+the author's, or Step 5.5's own auto-fix — so the note doesn't blame the author for a
+push this skill made.
 
 ## Rules
 
@@ -500,7 +602,9 @@ were lost to a mid-review push, not to laziness.
 - **Verify before flagging.** Read the file at each cited line; drop findings that
   don't reproduce. A wrong finding costs more trust than a missed nit.
 - **Verdict matches findings.** 0 Blocking → APPROVE; ≥1 Blocking → REQUEST_CHANGES.
-  Never soft-gate on nits. State the rule you applied.
+  Never soft-gate on nits. State the rule you applied. A fix this run pushed itself does
+  **not** downgrade the verdict — approving your own auto-fix commit is the intended
+  behaviour (Step 5, and the bound that makes it safe).
 - **Blocking bar is specific:** correctness-on-normal-use, real fixture PII,
   hardcoded colors / wrong component tier, a command bug that breaks every
   invocation, a gate CI will fail. Everything softer is Secondary/Nit.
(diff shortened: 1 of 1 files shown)


Samhit21 (commented):
> **Verdict: REQUEST_CHANGES** (3 Blocking, 3 Secondary, 2 Nits). Posted as `COMMENT` per house convention — the severity is in the prose, not the review state.

Reviewed as **code**, per this skill's own gate 3e ("a skill/script file **is** code — review it as code, not prose"). That gate is the reason two of the three Blocking findings exist: they are command-level, and they fire on ordinary invocations.

Worth naming the recursion up front: this PR removes the confirm-before-posting rule, and I reviewed it with the version of the skill that still has that rule — so this review was shown to a human before it landed. Under the proposed version it would not have been.

**Gates run on `4df49c5`:** 3e command-level (the applicable one) · description accuracy 3f · `fallow` clean, 4 files · 3a fixture PII **skipped, no fixtures touched** · 3b design-system + 3c style tokens **skipped, no `src/` touched** · `/code-review` **not run** — disabled for model invocation, so the generic pass here is hand-run and you should weight it accordingly.

No linked issue (`closingIssuesReferences` empty, no `Closes`/`Refs` in the body), so there is no AC checklist to run — reviewed against the diff alone.

## Blocking

### 1. The documented revert destroys unrelated user work

`:400` — see inline. `git checkout -- .` is not scoped to the fixes the skill made, and this fires on the **normal** failure path (verification red), not an edge.

### 2. The reviewer authors the head commit and then approves it

`:390` — see inline. `main` requires one approving review; Step 5.5 has the reviewer write code and Step 6 has the same run supply the approval that gates it. The pushed commit ships with **zero** independent review.

### 3. A ticked test-plan box for a skill that does not exist

> - [x] Verified `pr-sweep` compatibility with updated `pr-review` workflow

There is no `pr-sweep` in this repo. `.claude/skills/` holds 16 skills — `batch-issues`, `create-gh-issue`, `implement-batch`, `open-pr`, `pr-ready`, `pr-review`, six `probe-*`, `revise-pr`, `triage-issue` — and neither `find . -iname '*pr-sweep*'` nor `grep -rl 'pr-sweep' .claude/` matches anything.

Gate 3f rates a ticked box for a gate that demonstrably did not run as Blocking, and the reason applies here exactly: the other box ("Tested skill changes against `pr-review/SKILL.md`") is the file being changed, so between them the test plan asserts no verification that actually happened. For a change whose whole subject is *removing human confirmation*, an unverifiable test plan is the wrong thing to be wrong about.

If `pr-sweep` is unmerged or lives outside this repo, say so — the claim is fine, the phrasing isn't.

## Secondary

### 1. The review comments on lines it just fixed

`:397` — see inline. Step 5.5 pushes, then Step 6 anchors inline comments to lines Step 5.5 changed, so a nit is described against the corrected line. A fix that *deletes* its anchor line is the case that actually 422s into the body-only fallback.

### 2. Deleting the confirm rule leaves a dangling reference

`:541` — see inline. Step 0.6 still justifies itself as "the exception to the confirm-before-posting rule" (`:143-144`), and this diff deletes that rule.

### 3. The Provenance rationale does not reproduce

> Verification: `OFFLINECV_SKIP_HOOKS=1` push (due to `src/lib/pdf` 5000ms vitest timeout flake on `main`)

I could not reproduce it: `npx vitest run src/lib/pdf` under the pinned Node 20.20.2, **5 consecutive runs, 363 passed every time**, no timeout.

Stated carefully, because a flake that never reproduces is still a flake — 5 clean local runs do not disprove an intermittent or load-dependent timeout, and I am not claiming they do.

Bypassing the hook may well have been justified — `main` genuinely was red until #802 merged, on `letter-egress-ack` under Node 22+ (#779). But that is a different failure with a different cause, and the recorded reason is the artifact a bisecting maintainer trusts later. Worth correcting to whatever actually happened.

## Nits (non-blocking)

1. `:70` — `--non-interactive` and `--post` are documented but have no distinct behaviour now that autonomous is the default. See inline.
2. `:397` — the `git push` has no staleness handling (no `--force-with-lease`, no rebase-on-reject). Folded into the inline there.

## On the design, separate from the defects

The two halves of this PR are not equally risky and could ship independently.

**Autonomous posting** is defensible: a review is reversible, visible, and costs a notification. If a wrong review lands, a human reads it and says so.

**Auto-commit is a different category** — it writes to someone else's branch, and the skill's own stance argues against it. "Verify before flagging… a wrong finding costs more trust than a missed nit" is about the cost of being *wrong in prose*; Step 5.5 promotes every nit-level judgement into a commit, where being wrong costs a push. This skill's judgement is good enough to *propose* a fix and not yet good enough to *land* one unreviewed — the three Blocking findings above are themselves evidence of the error rate on a 32-line diff.

Splitting them would let the safe half land now.

---

.claude/skills/pr-review/SKILL.md
```diff
@@ -372,19 +377,35 @@ and a **fix** — a diff or exact command, not just a complaint. Cite the source
...
+
+1. **Apply the small fixes** directly in the checked-out worktree (formatting, comment wording, naming, non-behavioral quality fixes).
+2. **Run verification gates** (`npm run verify` or relevant lint/test commands) to ensure 100% green build and no regressions.
+3. **If verification passes:**
+   - Stage and commit the fixes with a clear message (e.g., `fix(review): address review nits`).
+   - **Do NOT add any `Co-Authored-By` or tooling/attribution trailer** (per repo rules).
+   - `git push` to the PR's head branch.
+   - Record the commit SHA for inclusion in the review post (`Addressed small fixes directly in commit <sha>.`).
+4. **If verification or push fails** (e.g. fork PR without write access, branch protection, or failing test):
+   - Revert uncommitted changes if verification failed (`git checkout -- .`).
```
Samhit21: **[Blocking]:** `git checkout -- .` is not scoped to the skill's own edits — it discards **every** unstaged tracked modification in the worktree, including work the user had in progress before the skill ran.

This is the gate-3e class ("a command bug that fires on **normal** invocation"), and the path is normal: it runs whenever `npm run verify` goes red, which is precisely when auto-fix is most likely to have been a bad idea.

Demonstrated in a throwaway repo — a user edit, plus a skill edit, plus a file the fix created:

```
--- before ---
 M skill-target.txt
 M users-work.txt      <- user's in-progress work
?? skill-new.txt       <- created by the fix

$ git checkout -- .

--- after ---
?? skill-new.txt       <- fix's new file SURVIVES
users-work.txt: "original"   <- user's edit DESTROYED, unrecoverable
```

So it is wrong in both directions at once: too broad (nukes unrelated work, and `checkout` leaves no reflog to recover from) and too narrow (leaves behind exactly the new files the failed fix added).

The surrounding design makes this reachable rather than theoretical — Step 1 runs `gh pr checkout`, which carries a dirty worktree onto the PR branch, so the user's edits are still present when Step 5.5 reverts.

Fix — scope the revert to what was actually touched, and capture a restore point first:

```bash
# before applying fixes
git stash push -u -m "pr-review: pre-fix snapshot"   # or: git rev-parse HEAD

# on verification failure — only the paths the fix wrote
git restore --source=HEAD --staged --worktree -- "${FIXED_PATHS[@]}"
git clean -fd -- "${FIXED_PATHS[@]}"    # removes new files the fix created
```

`git restore` is also the non-deprecated spelling for this. Whatever shape you pick, it has to (a) name paths rather than `.`, and (b) handle untracked files, or the failure path leaves the branch dirtier than it found it.

.claude/skills/pr-review/SKILL.md
```diff
@@ -372,19 +377,35 @@ and a **fix** — a diff or exact command, not just a complaint. Cite the source
...
   nits — list them as non-blocking and approve. Verdict-inconsistency (findings say
   "looks good" but state is REQUEST_CHANGES, or vice-versa) is itself a review defect.
 - Genuinely can't decide (needs author context) → `COMMENT` with the open question.
 
 State the verdict rule you applied in one line so it's auditable.
 
-### Step 6 — Draft, confirm, post
+### Step 5.5 — Auto-fix small items (0 Blockers)
+
+If the verdict is **`APPROVE`** (0 Blocking findings) AND small fixes (Secondary or Nits) exist AND `--no-commit` is NOT set:
```
Samhit21: **[Blocking]:** this makes the reviewer the author of the head commit **and** the supplier of the approval that lets it merge.

`main` requires one approving review. The sequence Step 5.5 → Step 6 is: write code → push it → post `APPROVE`. GitHub only blocks self-approval by the **PR author**, so a reviewer approving their own commit passes branch protection cleanly — and the merge queue then takes a head commit that no second party has read.

This repo *does* dismiss approvals on push — and the ordering here steps around that. Timestamped from #752:

```
2026-08-05T20:27:41Z  reviewed          Samhit21   (APPROVE)
2026-08-05T21:12:03Z  review_dismissed  s-annam
2026-08-05T21:12:03Z  head_ref_force_pushed  s-annam
```

Same second: the push dismissed the approval. That protection exists precisely so new code gets new eyes — but Step 5.5 pushes *before* Step 6 approves, so the approval lands after the push and is never dismissed.

**The caveat that would sink this finding, stated plainly:** if branch protection has *"require approval from someone other than the last pusher"* enabled, GitHub blocks this itself and the finding drops to a documentation gap. I could not read that setting — `repos/.../branches/main/protection` returns 404 for my token, and the one visible ruleset (`main-merge-queue`) carries only a `merge_queue` rule, no `pull_request` rule. Worth confirming before acting on this; if it is enabled, say so and I will withdraw it.

The blast radius is bounded by "small fixes only" — but the boundary is the reviewing model's own judgement about what counts as small, and there is no second check on that call. This review found three Blocking issues in a 32-line diff; that is the error rate the auto-commit path would be operating at unsupervised.

Options that keep the autonomy without the hole:

1. **Push the fixes, but post `COMMENT`, not `APPROVE`.** The work still lands; a human still supplies the approval. This is the smallest change and I would take it.
2. **Push to a separate branch and open a PR against the head branch**, so the fixes are reviewable as a diff rather than assumed.
3. **Keep `--no-commit` as the default** and require the flag to opt *in*, so the destructive-by-default direction is inverted.

Worth deciding explicitly in the skill text either way, because right now the interaction with branch protection is unstated — a reader would have to work out that the approval and the commit come from the same actor.

.claude/skills/pr-review/SKILL.md
```diff
@@ -372,19 +377,35 @@ and a **fix** — a diff or exact command, not just a complaint. Cite the source
...
+
+1. **Apply the small fixes** directly in the checked-out worktree (formatting, comment wording, naming, non-behavioral quality fixes).
+2. **Run verification gates** (`npm run verify` or relevant lint/test commands) to ensure 100% green build and no regressions.
+3. **If verification passes:**
+   - Stage and commit the fixes with a clear message (e.g., `fix(review): address review nits`).
+   - **Do NOT add any `Co-Authored-By` or tooling/attribution trailer** (per repo rules).
+   - `git push` to the PR's head branch.
```
Samhit21: **[Secondary]:** pushing here means Step 6 anchors inline comments to lines this push just changed.

I overstated this on a first pass and want to be precise: it does **not** reliably 422. Step 6 derives anchors from `gh api pulls/N/files`, which is fetched *after* the push, so the anchors are normally fresh and valid.

What does hold is narrower and still worth fixing:

- **The review comments on lines it just fixed.** An APPROVE-with-nits run applies the nit, pushes, then posts an inline comment describing that nit against the corrected line. A reader sees a complaint about code that no longer says that.
- **A fix that deletes its anchor line leaves nothing to anchor to** — that is the case that genuinely 422s and drops the whole review to the body-only fallback, whose message ("the anchors were lost to a mid-review push") would then be blaming the author for the skill's own push.

Ordering fixes it: derive the anchors from the patch **before** Step 5.5 pushes, or re-fetch `pulls/N/files` after the push and re-anchor once, or simply post the review first and push after.

**[Nit], same line:** the `git push` has no staleness handling. If the author pushed while the review was running, this is rejected non-fast-forward and drops into the failure branch — which then runs the destructive revert flagged at `:400`. `--force-with-lease` is the wrong tool here (it would clobber the author), so the right shape is probably fetch + rebase, or simply abandon the auto-fix and report it.

.claude/skills/pr-review/SKILL.md
```diff
@@ -517,9 +538,8 @@ were lost to a mid-review push, not to laziness.
 - **One `422` is information, not a puzzle.** It almost always means the author pushed
   mid-review. Re-anchor against the fresh patch once; if that fails, post body-only
   and move on. Never loop on anchoring.
-- Confirm before posting to a public PR — with one carve-out: the Step 0.6 👀
-  reaction posts unconfirmed (single reversible reaction, no notification, no
-  prose). Everything with words in it gets confirmed.
+- **Fully autonomous execution by default.** Posts the review automatically at the end of the run (unless `--local` is explicitly passed for preview). No interactive confirmation prompt.
```
Samhit21: **[Secondary]:** deleting the confirm-before-posting rule leaves Step 0.6 citing a rule that no longer exists.

`:143-144` still reads:

> No confirmation needed for this one write (it is the exception to the **confirm-before-posting rule**): it is a single reversible reaction…

With this bullet replacing that rule, the 👀 reaction is no longer an exception to anything — everything posts unconfirmed now. The paragraph's whole argument ("this one write is special") stops parsing, and a later reader trying to reconstruct the intent finds a rule that was removed rather than relaxed.

Either rewrite `:143-144` to justify the reaction on its own terms (it is posted *early*, before the slow work, which is the part that actually matters and survives this change), or keep a narrowed rule that says which writes still warrant a human — the commit-and-push in Step 5.5 being the obvious candidate.

.claude/skills/pr-review/SKILL.md
```diff
@@ -66,6 +67,10 @@ gh pr view --json number,headRefName,state -q '{n:.number,head:.headRefName,stat
 If that finds no PR and none was passed, list open PRs and **ask** which one. Never
 guess. `--effort` is passed straight through to `/code-review` (default `high`).
 
+- **Fully Autonomous execution**: By default (and when `--post` or `--non-interactive` is set), `pr-review` runs unattended to completion and posts the review to GitHub at the end without prompting for confirmation.
```
Samhit21: **[Nit]:** `--non-interactive` and `--post` are now documented flags with no distinct behaviour.

This line says autonomous posting is the default "(and when `--post` or `--non-interactive` is set)" — so both flags request the behaviour that already happens without them. `--local` is the only switch that changes anything.

That is the gate-3e "documented-but-unimplemented flag" shape, in its mild form: nothing breaks, but `argument-hint` at `:4` advertises two options that a user could reasonably expect to alter something. `--post` is also now vestigial from the old `[--post|--local]` pair, where it was the meaningful half.

Suggest dropping both from `argument-hint` and this list, leaving `[--local]` and `[--no-commit]` as the only behavioural switches — or, if `--non-interactive` is meant to be forward-compatible with a future interactive mode, say that here so it reads as deliberate rather than left over.

.claude/skills/pr-review/SKILL.md
s-annam: Fixed. Confirmed both halves of your repro — too broad (nukes unrelated tracked edits, no reflog) and too narrow (leaves the fix's untracked files behind) — and took the shape you sketched.

Step 5.5 now opens by capturing a restore point and tracking what it writes:

```bash
PRE_FIX_SHA="$(git rev-parse HEAD)"
HEAD_BRANCH="$(gh pr view "$PR_NUM" --repo "$REPO" --json headRefName -q .headRefName)"
FIXED_PATHS=()          # append every path you write, including new files
```

and the failure path names those paths instead of `.`:

```bash
git reset --soft "$PRE_FIX_SHA"                    # only if step 4 already committed
git restore --source="$PRE_FIX_SHA" --staged --worktree -- "${FIXED_PATHS[@]}"
git clean -fd -- "${FIXED_PATHS[@]}"               # removes new files the fix created
```

The `git reset --soft` line is there because the revert is now also reachable *after* a successful commit — the mid-review-push check you flagged in the `:397` thread aborts between commit and push, so the restore has to undo a commit as well as a dirty tree.

Your "gate-3e class, and the path is normal" framing is what made this land as Blocking rather than a nit — it fires on `npm run verify` going red, which is the most likely branch, not an edge. I also promoted it to a standing rule so a later edit doesn't quietly reintroduce `checkout -- .`:

> **The auto-fix revert names paths.** Never `git checkout -- .` — it destroys the user's unrelated in-progress work with no reflog while leaving the failed fix's new files behind.


Pushed in `eb25bc0`.

.claude/skills/pr-review/SKILL.md
s-annam: Confirmed the setting you couldn't read, and it does **not** sink the finding — so thank you for flagging it rather than assuming either way:

```
$ gh api repos/offlinecv/OfflineCV/branches/main/protection --jq .required_pull_request_reviews
{
  "dismiss_stale_reviews": true,
  "require_code_owner_reviews": false,
  "require_last_push_approval": false,      <- not enabled
  "required_approving_review_count": 1
}
```

`require_last_push_approval: false`, and the one active ruleset (`main-merge-queue`) carries only a `merge_queue` rule. So GitHub permits push-then-approve exactly as you described, and your reading of the ordering vs `dismiss_stale_reviews` is right: the approval lands after the push, so there is nothing stale to dismiss.

**Not taking any of the three options, though — keeping `APPROVE`.** This is a deliberate product decision rather than an oversight, which is precisely the thing your comment says is unstated, so I've stated it in the skill text instead (Step 5):

> **`APPROVE` still applies when Step 5.5 pushed the fixes — this is deliberate.** […] That is the intended outcome, not a loophole being exploited. The alternative — post the nits, wait for the author to apply them, watch the new push dismiss the approval, review again — is the round-trip this skill exists to remove.

Killing that round-trip *is* the feature. Option 1 (`COMMENT` instead of `APPROVE`) preserves the hole's absence but reintroduces the wait, which leaves the skill doing the work and still blocking on a human for a diff that has zero blocking findings and a green `verify`.

**What I did take from this thread** is your point about the error rate — "three Blocking issues in a 32-line diff" is a fair shot and the right thing to size the bound against. So the bound is now on *behaviour*, not on size, and it's written into both Step 5 and the Rules:

> The head commit lands unread by a second party, so the auto-fix is confined to changes that cannot alter behaviour (formatting, comment wording, naming, dead-code removal), gated on a green `npm run verify`, and only when there are **0 Blocking findings**. Anything that changes what the code *does* — however small it looks — is a finding for the author, not an auto-fix. When in doubt about whether an edit is behavioural, it is.

That's the part your finding actually protects: a reviewer approving its own *non-behavioural* commit on a green build is a much narrower claim than approving its own judgement about what counts as small.

Leaving this thread open — it's a design disagreement resolved by the repo owner, not a defect I've cleared, so it's yours to close or keep arguing.


Other threads addressed in `eb25bc0`.

.claude/skills/pr-review/SKILL.md
s-annam: Both parts fixed, and the correction in your first paragraph was the useful one — you're right that it doesn't reliably 422, since Step 6 fetches `files` after the push. I've fixed the two narrower things that do hold rather than the overstated version.

**The review commenting on lines it just fixed.** Handled at the source: an item the run fixed is no longer a finding at all, so there is no comment to mis-anchor.

> **A fix that landed is not a finding.** Anything Step 5.5 actually fixed and pushed must **not** also be posted as an inline comment — the anchor now points at corrected code, so the author reads a complaint about a line that no longer says that. Move each fixed item out of the inline set and into a `## Fixed in <sha>` list in the body.

**The deleted-anchor 422.** The ordering is now explicit where the anchors get derived:

> **Derive the anchors *after* Step 5.5 has pushed**, never before — a fix that moved or deleted a line invalidates any anchor computed against the pre-fix patch, and a fix that deletes its anchor line entirely leaves nothing to anchor to […]

Plus the Step 7 fallback note, which as you say was blaming the author for the skill's own push — it now has to say *whose* push moved the diff.

**On the Nit** — `git push` staleness. Took the fetch-and-check shape; `--force-with-lease` stays out for the reason you gave (it clobbers the author):

```bash
git fetch origin "$HEAD_BRANCH"
if ! git merge-base --is-ancestor "origin/$HEAD_BRANCH" HEAD; then
  # the author pushed while the review ran — their commits are not in our history
  echo "head branch moved mid-review; abandoning auto-fix" >&2
else
  git push origin HEAD                 # never --force / --force-with-lease here
fi
```

You spotted the interaction that made this worth more than its severity: a rejected push dropped into the failure branch, which is the destructive revert from the `:400` thread. Both ends are now closed — abandon rather than force, and the abandon path restores only its own paths.


Pushed in `eb25bc0`.

.claude/skills/pr-review/SKILL.md
s-annam: Fixed — took your first option, and then the second one as well, since they turned out not to be alternatives.

`:143-146` now stands on its own terms, on exactly the ground you identified (it's posted *early*, which is the part that survives the change):

> This one write is safe to make *this early*, before anything has been reviewed: it is a single reversible reaction, it notifies nobody, and it carries no prose that could be wrong. That is what lets it precede the work it announces — every other write this skill makes happens only after the findings exist.

And rather than a narrowed confirm rule, the Rules section now just enumerates what the skill writes unattended, so a reader reconstructing the intent finds the current shape instead of a removed one:

> **Autonomous execution; three unattended writes, in this order.** The 👀 reaction (Step 0.6), the auto-fix commit + push (Step 5.5), and the review post (Step 6). Nothing is confirmed with the user — `--local` is the only preview. The reaction is safe *early* because it carries no prose; the other two happen only after the findings exist, and the push is bounded by the next rule.

You named the commit-and-push in Step 5.5 as the obvious candidate for a retained human gate. It didn't get one — that's the design decision I've argued out in the `:390` thread — but it did get the behavioural bound described there, which is the next rule that bullet points at.


Pushed in `eb25bc0`.

.claude/skills/pr-review/SKILL.md
s-annam: Fixed — dropped both, as you suggested. Neither was forward-looking; `--non-interactive` was aspirational and `--post` was vestigial from the old `[--post|--local]` pair, exactly as you read it.

`argument-hint` is now:

```
<#|#N> [--repo owner/repo] [--local] [--effort low|medium|high] [--no-commit]
```

and the flag list says what's left, without advertising a switch for the default:

> **Autonomous by default**: `pr-review` runs unattended to completion and posts the review to GitHub at the end without prompting for confirmation. There is no flag to request this — it is the only mode. `--local` and `--no-commit` are the two switches that change behaviour.

Grepped for other references before removing them — the only hits were the three lines in this file (`:4`, `:60`, `:70`); `implement-batch`'s "non-interactive" is unrelated prose.


Pushed in `eb25bc0`.

Samhit21 (approved):
> **Verdict: APPROVE** — the diff is clean. Three items remain, **none of them in the code**: a false line in the PR body, an unreproduced claim, and the branch carrying two commits. Approving rather than holding the code hostage to them, at the repo owner's call.

**Read the three as pre-merge requirements, not as blockers on the change.** All four threaded findings are genuinely fixed, and I verified each against the file rather than the replies.

⚠️ **This approval will not survive.** You confirmed `dismiss_stale_reviews: true`, and item 3 means the branch must be squashed before it reaches the queue — that push will dismiss this review. Land all three in the same push and re-request; I will re-approve on sight rather than re-reviewing from scratch.

### Verified fixed

| Finding | Verified |
|---|---|
| `git checkout -- .` (Blocking) | ✅ appears only in prohibitions now (`:460`, `:639`); restore point + `FIXED_PATHS` + `git restore --source` + `git clean -fd` on named paths; `git reset --soft` for the post-commit case you spotted |
| anchors vs. the auto-fix push (Secondary) | ✅ "A fix that landed is not a finding" (`:469`) + "Derive the anchors *after* Step 5.5 has pushed" (`:516`) |
| push staleness (Nit) | ✅ `git fetch` + `merge-base --is-ancestor`, abandon rather than force (`:440-444`) |
| dangling confirm reference (Secondary) | ✅ `:143` now stands on its own terms; Rules enumerate the three unattended writes |
| `--non-interactive` / `--post` (Nit) | ✅ both gone; `argument-hint` clean, zero remaining references |

The `git reset --soft` addition is a good catch of your own — the revert really did become reachable after a successful commit once the mid-review-push check landed, and I had not traced that far.

### On the self-approval thread — your call, and it is recorded

You confirmed the setting I could not read (`require_last_push_approval: false`), which means my finding was technically right and is now a **stated design decision** rather than an unstated hole. That was the substance of the finding — my comment said the interaction with branch protection "is unstated" — so writing it into Step 5 resolves what I actually raised.

Re-binding the auto-fix to *behaviour* rather than *size* is a real improvement and a better rule than the one I proposed: "a reviewer approving its own non-behavioural commit on a green build" is a much narrower claim, and "resolves to yes when unsure" puts the ambiguity on the safe side.

I still would not ship it — a rule that says "only non-behavioural edits" is enforced by the same judgement that produced three Blocking findings on a 32-line diff, and the failure mode is a silent bad commit rather than a visible bad review. But that is a disagreement about risk appetite on your own repo, not a defect, and I am not gating on it. Leaving the thread open as you asked.

## Pre-merge — must land before the queue

### 1. The `pr-sweep` test-plan box is still ticked, and `pr-sweep` still does not exist

*(Was Blocking last round. Still the same defect — a false claim in the durable record — but it is in the PR body, not the diff, so it is not holding the code.)*

Unchanged from the last round — it had no thread to reply to, so I suspect it was simply missed rather than declined.

> - [x] Verified `pr-sweep` compatibility with updated `pr-review` workflow

Re-checked at `eb25bc0`: `.claude/skills/` holds the same 16 skills, none named `pr-sweep`, and `find` across both the repo and `~/.claude` matches nothing.

The other box ("Tested skill changes against `pr-review/SKILL.md`") names the file being changed, so between them the test plan still asserts no verification that happened — on a PR whose subject is removing human confirmation. If `pr-sweep` is unmerged or lives elsewhere, one clause fixes it; if it was aspirational, drop the box.

### 2. The `src/lib/pdf` flake claim is now asserted twice and I still cannot reproduce it

The original Provenance line is unchanged, and the new addendum repeats it:

> the `src/lib/pdf` set is the 5000ms timeout flake already noted above

I have now run it **8 times across both Node versions** — 5× on the pinned Node 20.20.2 and, since your note says "on this machine", 3× on Node 25.8.1. **363 passed every time, no timeout.**

A flake that never reproduces is still a flake and 8 runs do not disprove one — but it is load-bearing here, because it is the recorded justification for an `OFFLINECV_SKIP_HOOKS=1` push, and it is the artifact a bisecting maintainer trusts later.

**Your `letter-egress-ack` half is correct and I confirmed it independently**, which is why I am confident the pdf half is the odd one out: this branch's merge base is `881bbc9`, which predates `a50fca7` (#802), so it genuinely lacks the #779 fix and those nine failures are exactly what that produces. That explanation is right. Worth either reproducing the pdf timeout and pasting the output, or dropping it and letting the `letter-egress-ack` reason stand on its own — it is sufficient by itself.

### 3. The branch carries two commits — it cannot reach the queue like this

```
eb25bc0 fix(skills): address review comments on PR #818
4df49c5 feat(skills): make pr-review skill fully autonomous with small fix auto-commit
```

`CLAUDE.md` → **Hard rules**: *"One commit per PR. `main` merges through a merge queue that derives the squash message from the branch, so a multi-commit PR lands `wip` and `fix lint` in `main` forever. Collapse the branch to a single commit before it reaches the queue."*

Not a code defect, and normal mid-review state — flagging it because it is irreversible once merged, and because it is what will dismiss this approval.

---

PR #822, opened 2026-08-14
feat(nav): add a top-level journey rail across both entries (#812)

### LINKED ISSUE #812: No top-level journey: the product's most sequential flow has the weakest navigation signal
## Problem

The product's navigation vocabulary is inverted: the most sequential thing in it has the weakest signal, and the most local thing has the strongest.

| Level | What it is | What it communicates | Component |
|---|---|---|---|
| L1 — the journey | two HTML entries, `location.href` navigation + `sessionStorage` handoffs | **nothing** | — |
| L2 — views | `/`: Your resume · Find jobs · Resume quality · On-device AI · Raw text & flags. `/jobs/`: Search · Saved jobs | "peers, pick any" | `Tabs` |
| L3 — one query | Role → Skills → Narrow → Review (#602) | "ordered, here is your progress" | `Stepper` |

Building a single job-search query gets a four-step rail with per-step summaries. The user's actual journey — the thing that spans both entries and every session — gets no rail at all.

The tabs additionally mislead. `Your resume` sitting next to `Find jobs` in one `TabList` reads as *alternatives, pick one*. What is actually true is a pipeline: parse the résumé, fix it, find jobs, tailor to one, export. It is a funnel wearing a tab bar, split across two HTML entries the user is bounced between — `jobs-handoff.ts` on the way out, `tailor-handoff.ts` on the way back — with nothing on screen indicating those are two halves of one flow.

There is a second, sharper version of the same defect: the journey rail this issue proposes is **sticky and always present**, but today nothing rehydrates a résumé on a fresh visit to `/`. A rail that claims "you're at step 3" while the app has silently forgotten the user's work is worse than no rail — it advertises a journey the product cannot actually resume. That gap is in scope here (see "Scope: résumé auto-restore joins this issue" below).

## Why this is filed above the other children

Two sibling issues under #811 (#809 job-search relevance, #810 score guidance) each contain a placement question — *where does this control/guidance live so the user finds it at the moment they need it?* Neither can be answered well while the top-level flow is undefined; answering them separately risks two more local patches on the same underlying gap.

#809 is the clearest instance. Its root cause is compositional: #602 mounts the Search button on **every** step (deliberately — "a user who is happy with it must never have to walk four steps to run it"), and `FindJobsPanel` folds the whole form away on submit. Skippable, then hidden. The user never learns the Narrow step exists. That is this issue's disease showing up in one panel: the product does not tell you where you are or what is still changeable.

This issue should land first. The others build on the answer.

## Recommendation: a dedicated `JourneyRail` component

A top-level journey indicator must orient the user across entries without gating progress as a rigid wizard:

- A **wizard gates** — step 3 is unreachable without step 2.
- A **rail orients** — it shows the arc, marks where you are, and allows jumping to any stage (rendering clear, lightweight guidance empty states when prerequisite data is not yet populated).

The `ui-ux-pro-max` rule set demands both *"show progress for multi-step processes"* (Feedback → Progress Indicators) and *"provide Skip and Back / don't force a linear unskippable tour"* (Onboarding → User Freedom), and names **"complex onboarding"** as an explicit anti-pattern for this product type. A rail satisfies all three. A true wizard violates two.

**Proposed stage set**:

> Add résumé → Fix it → Download → Match jobs → Tailor

`Download` sits directly after `Fix it`, ahead of the job-search half: a user who came only to repair what a text extractor reads back is finished there and never needs a job board. Putting the export last implied the résumé lane is a prologue to the search lane.

`Match jobs` is used specifically to avoid a name collision: `/` already has an L2 tab labelled exactly **"Find jobs"** (`src/components/features/ResultDetailTabs.tsx:169`), and `/jobs/` has a tab labelled **"Search"**. An L1 stage sharing a name with an L2 tab is exactly the ambiguity this issue exists to remove.

| # | Stage | Derives from | Lands on |
|---|---|---|---|
| 1 | Add résumé | `state.phase` idle vs. done, or rehydrated record | DropZone / restored résumé on `/` |
| 2 | Fix it | `phase === "done"` | `/` L2 tabs: Your resume, Resume quality, Raw text & flags |
| 3 | Download | résumé present | `render-ats-pdf` / export modal on `/` |
| 4 | Match jobs | jobs handoff present, or search run | `/jobs/` Search |
| 5 | Tailor | tailor handoff written — **rendered only on `/jobs/`, or on `/` while a JD steers the rewrite** | `/` rewrite steering |

`Fix it` collapses three L2 tabs (Your resume, Resume quality, Raw text & flags) into one L1 stage — that is the L1/L2 split working as intended, not lost detail; L1 marks the arc, L2 owns the detail underneath.

### Key Architectural Decisions

1. **Dedicated `JourneyRail` Component**: Build a dedicated `JourneyRail` component in `src/components/features/` (or `src/design-system/`) specifically tailored for header navigation across routes, avoiding artificial encapsulation inside `StepperContext` / `StepPanel` containers.
2. **Ungated Navigation with Guidance Empty States**: Users can click any stage on the rail. When landing on an unpopulated stage (e.g. clicking `Tailor` before matching jobs, or `Fix it` before loading a resume), render a standard lightweight empty-state card with a clear call-to-action to the prerequisite stage.
3. **Header Placement & Sticky Layout**:
   - **The rail always takes a full-width row of its own, at every width** — a sibling of the brand row, never a wrap-item inside it. Sharing the header row at `lg:` (the original plan) made the arc read as one more header control beside "Saved jobs" and the star CTA, which is the L1-vs-L2 confusion this issue exists to remove; it also forced a third content-sized breakpoint band whose chips were sized by whatever width was left after the brand and the CTAs took theirs.
   - Header is `sticky top-0 z-20 bg-surface-base/95 backdrop-blur border-b border-border-light` with proper padding compensation so content scrolls cleanly underneath without layout shift.
4. **Auto-Restore on Cold Mount of `/`**:
   - Priority 1: Active parse in memory / bfcache (instant restoration, skips re-fetch).
   - Priority 2: Cold mount rehydration from IndexedDB (hydrates newest saved `resumes` record into `phase: "done"`).
   - Priority 3: Default fallback to `phase: "idle"` (Step 1: Add résumé).
   - The restore is spent ONCE per page lifetime and re-checks `phase === "idle"` when the IndexedDB read resolves. Both guards are load-bearing: `reset()` returns the app to `idle`, so a restore keyed on that phase alone would immediately re-hydrate the résumé the user just dismissed, and a file dropped while the read is in flight must not be clobbered by the record that lands after it.
5. **State Derivation & bfcache Safety**: Derives stage state during render or via `pageshow` listeners keyed on `event.persisted` — never from a mount-only `useEffect(…, [])`.
6. **Three-State Stage Model (no completion ledger)**: Each stage is `current`, `available` (data behind it — clickable, lands somewhere), or `upcoming` (nothing behind it — click renders the guidance card). There is no persisted "completed" ledger, because two stages have no durable signal to build one from: the tailor handoff is one-shot-consumed (`consumeTailorHandoff`) and a PDF export is a transient `useDownloadPdf` call that persists nothing. Availability is therefore derived purely: `Add résumé` always; `Fix it` / `Match jobs` / `Download` when a résumé is present; `Tailor` when `jdContext !== null` **and** a résumé is present (normalised inside the derivation, not left to call sites). This keeps the rail honest — it reports which stages have data behind them and which do not, rather than claiming a completion it cannot observe.

**Visibility is a separate question from availability.** `Tailor` is *rendered* only where it can actually be started — on `/jobs/`, where a posting's "Tailor résumé to this job" button is on screen, and on `/` while a JD is steering the rewrite. On a cold `/` the rail was advertising a step whose only honest instruction was "go somewhere else and press a different button", which reads as a broken stage. Everywhere else the arc is four stages. The announced position ("Step 3 of 4") counts the *rendered* stages, never all five.
7. **Landing on `/`'s L2 tab**: `Fix it`, `Tailor` and `Download` all land on `/`'s `reconstructed` tab, which `ResultDetailTabs` owns as local state with no carrier from outside. A `requestedTab?: { id, nonce }` prop is threaded `App` → `Result` → `ResultDetailTabs`, switching tabs when the nonce changes. Without it a rail click is a visible no-op whenever the user is sitting on `Raw text & flags` — the exact disorientation this issue exists to remove.

## Risks to resolve during design

**1. Browser Back across the two entries — largely answered, one manual check remains.**

Prior measured work (#706, #783) narrows this:

- `/` is restored from bfcache with its React state intact — parse and inline edits included — when the return is `history.back()`. `nav-return.ts` fires `history.back()` *precisely because* of this.
- What does **not** survive bfcache restore is a mount-only `useEffect(…, [])` — the tree isn't remounted, so the effect never re-runs on the return leg. That was the #783 defect on `tailor-handoff.ts`'s return path, fixed with a `pageshow` listener keyed on `event.persisted`.
- The fallback branch is a genuine page load: when the return is `location.href = BASE_URL` rather than `history.back()`, `/` boots **idle** with no parse and nothing mounts. This is the case auto-restore covers for any résumé saved in IndexedDB.

**Consequence for the rail:** its stage state must derive at render time from the signals themselves, or subscribe via a `pageshow` listener keyed on `event.persisted` — never from a mount-only `useEffect(…, [])`.

**Testing caveat:** Playwright's bundled Chromium ships with BackForwardCache disabled, so `event.persisted` is always `false` there and a Playwright run cannot reproduce or refute any of this. The bfcache-derivation check has to be a **manual** pass in a real browser, called out explicitly in Acceptance below.

**2. Sticky rail vertical budget.**

~~The sticky header must remain within a compact vertical budget ($\le 48\text{px}$ on desktop, $\le 88\text{px}$ on mobile for 2 rows)~~ **— superseded.** That budget assumed the in-header-row placement decision 3 now drops; a dedicated rail row costs a line at every width. Measured actual: **119px** at ≥ 1024, **147px** below. The requirement that survives is the one that matters: padding compensation on the scrolling root, sized to those measured bands, so no scroll target lands underneath the pinned bar.

**3. No deep-linking.**

Two HTML entries and no router means a stage has no URL. Handled gracefully via lightweight inline empty-state views and internal tab state without inventing parallel router abstractions.

## Acceptance

**Core behaviour**
- [ ] Dedicated `JourneyRail` component renders on both HTML entries (`/` and `/jobs/`) from `PageShell`
- [ ] Stage state derives from existing signals at render time (or via a `pageshow`-keyed subscription) — never from a mount-only `useEffect(…, [])`
- [ ] On cold mount of `/`, the most-recent `resumes` record in IndexedDB rehydrates into analysis state without an explicit Load click, so Stage 1 reads as satisfied
- [ ] Navigation is permitted to all stages; navigating to an unpopulated stage renders a lightweight empty-state card with guidance to the preceding step
- [ ] `Fix it` / `Tailor` / `Download` land on `/`'s `reconstructed` tab even when the user is currently on another L2 tab (see Decision 7)
- [ ] Stage labels follow the #519 rule — what the user gets, not internal vocabulary, and `Match jobs` does not collide with any L2 tab label
- [ ] A user mid-flow can see, without clicking, which stage they are on and what remains

**Layout & Responsive**
- [ ] Rail takes its own full-width row at EVERY width — never wraps to multiple rows, never scrolls horizontally, at 320 / 375 / 414 / 768 / 1024 / 1440
- [ ] Header is sticky (`sticky top-0 z-20 bg-surface-base`), with padding compensation on the content container so it never overlaps the first content section

**Accessibility**
- [ ] Current / visited / upcoming state is never conveyed by colour alone — surface/weight changes per #516 pattern
- [ ] Keyboard reachable, focus visible, 44×44px touch targets, `prefers-reduced-motion` respected on any transition
- [ ] Progress stated in words for screen readers (e.g. "Step 2 of 5: Fix it"), not numerals alone

**Manual verification (cannot be automated)**
- [ ] Browser Back between `/` and `/jobs/` is manually verified in a real browser (desktop and mobile) — Playwright cannot reproduce this, since its bundled Chromium ships with BackForwardCache disabled

## Non-goals

- Not a router, and not a third HTML entry
- Not a wizard or gated step flow: no stage is locked
- Not a tutorial, coach-marks, or a first-run tour
- Not a change to either handoff's payload or to what leaves the browser
- Not the #809 or #810 fixes themselves — this defines where they land, it does not implement them

s-annam:
## Summary

Adds `JourneyRail`, the L1 navigation the product never had: an ordered arc — Add résumé → Fix it → Download → Match jobs → Tailor — rendered from `PageShell` on both HTML entries, in a full-width row of its own at every width. It is a rail, not a wizard: every stage stays clickable, and a stage with no data behind it renders a guidance card naming its prerequisite instead of refusing the click.

Résumé auto-restore on cold mount of `/` ships in the same change because the two are one behaviour — a sticky rail claiming "you are at step 3" while the app has silently forgotten the user's work is worse than no rail at all.

No new `fetch`, no analytics event, no storage schema change, no change to either handoff's payload. The three egress paths are untouched.

Closes #812

## Three decisions that differ from the issue body

The issue body predates a design review on the rendered UI; these supersede it and the body has been annotated to match.

1. **The rail always takes its own row** (issue Decision 3 had it inside the header row from `lg:` up). Sharing the row made the arc read as one more header control beside "Saved jobs" and the star CTA — the exact L1-vs-L2 confusion this issue exists to remove — and forced a third content-sized breakpoint band whose chips were sized by whatever width was left after the brand and CTAs took theirs. It is now a sibling of the brand row, not a wrap-item in it.
2. **Download moved ahead of the job-search half** (was last). A user who came only to repair what a text extractor reads back is finished at Download and never needs a job board; putting the export last implied the résumé lane is a prologue to the search lane.
3. **Tailor renders only where it can be started** (issue Decision 6 derived it purely from `jdContext`). Tailoring always begins at a specific posting's "Tailor résumé to this job" button, so on a cold `/` the rail was advertising a step whose only honest instruction was "go somewhere else and press a different button". It now appears on `/jobs/`, where that button is on screen, and on `/` while a JD is actually steering the rewrite. Everywhere else the arc is four stages.

Consequence worth flagging: the issue's "sticky header ≤ 48px desktop" budget is **not** met — the header is 119px at `lg`+ and 147px below, because a dedicated rail row costs a line. That budget was written assuming the in-header-row placement that decision 1 drops.

## Visual redesign of the rail

The first version rendered as a flat row of numerals on a near-invisible track — indistinguishable from a `Tabs` bar, with inactive entries as grey text on a grey surface. Now each state carries its own **marker shape and glyph**, so all three survive a greyscale render rather than leaning on the accent hue:

| State | Marker | Label |
|---|---|---|
| Current | filled disc + numeral, on `surface-card` | `content-primary`, semibold |
| Ready | ringed disc + ✓ (U+2713 U+FE0E) | `content-primary`, medium |
| Upcoming | hairline disc + numeral | `content-secondary`, normal |

Also fixed: the chips sat `gap-0.5` (2px) apart, so two adjacent 44px touch targets had a 2px gutter and a mis-tap landed on the neighbouring stage. Now 8px.

## Review focus

- `src/lib/journey.ts` — `stages` is the rendered subset and the rail numbers from it. Is there a signal combination where `current` could fall outside `stages`? (There is a test asserting there is not, across all 8 combinations.)
- `src/lib/journey.ts` — `availability` stays keyed by every stage id even when `stages` omits one. Right call, or should a hidden stage have no availability answer at all?
- `src/styles.css` — `scroll-padding-top` is set on the scrolling root, not per target, and the two bands are hand-measured. Does a global root value perturb any nested scroll container or dialog?
- `src/hooks/useAutoRestoreResume.ts:142` — the tailor-handoff gate sits *after* the attempt is spent. Right ordering if a handoff were written between the library resolving and the load returning?
- `src/App.tsx` — `setJdContext(null)` keyed on `parseKey`. Does it cover the two paths that leave `displayResult` non-null with no reporter mounted (blank authoring, `fonts_unmappable`)?

## Test plan

- [x] `npm run lint` clean
- [x] `npm run verify` green in the pre-push hook — **5566 tests passed**, 337 files, build ✓
- [x] ~50 tests across `journey`, `JourneyRail`, `useAutoRestoreResume`, `PageShell`, `ResultDetailTabs`
- [x] Mutation-verified: five guards each had their protection deliberately broken and every break was caught by a named test
- [x] **Verified in a real browser** (Chromium, both entries) — see below
- [ ] **Not done — browser Back between `/` and `/jobs/`, desktop and mobile.** Playwright's bundled Chromium ships with BackForwardCache disabled, so `event.persisted` is always `false` and no automated run can reproduce or refute it. Issue #812 calls it out as manual-only; it has not been performed.

### Browser verification

Measured on both entries at 15 widths from 320 to 1920:

- Header band: **147px** below `lg` (320–414 on both; `/jobs/` hits it again at 640, where its extra `headerExtra` button re-wraps a brand row that `/` keeps on one line), **119px** at every width ≥ 1024. `scroll-padding-top` is sized from those, and `#contact` / `#reconstructed-resume` were confirmed to land 33px clear of the pinned header at 1424px and 65px clear at 577px — not underneath it.
- At **320px with all five stages** (`/jobs/`): one row, no rail horizontal scroll, no page horizontal scroll, every trigger ≥ 44×44, gap exactly 8px.
- All three marker states confirmed rendering simultaneously on `/` with a résumé present, and Tailor confirmed absent on a cold `/` and present on `/jobs/`.

## Adversarial review — findings deferred out of scope

Real, and deliberately not fixed here. Recorded so they are not lost:

1. **#516 selection affordance is duplicated in a third place outside `src/design-system/`.** The recessed-track + surface/weight pattern is re-implemented in `JourneyRail`; extracting it is a design-system change.
2. **Two L1 stages resolve to the same landing.** `Fix it` and `Download` both land on `/`'s `reconstructed` tab — worth a follow-up on whether Download should open the export path directly.
3. **`llmOverride` payload divergence.** The rail's Match-jobs departure sends pre-recovery fields, matching the pre-existing behaviour of `goToSavedJobs`. Behaviour unchanged here; the `FindJobsLauncher` docblock was narrowed so it no longer over-claims.
4. **`SaveResumeBar` mints a duplicate entry** when saving a résumé that auto-restore put on the page. Auto-restore makes this reachable more often; the behaviour itself pre-exists.
5. **`PageShell` is past the ~200 LOC guideline.** `fallow` additionally flags `App` and `ResultDetailTabs` — all report-only, and both are inherited rather than introduced here.

## Provenance

| Stage | Model | Effort |
|---|---|---|
| Implementation | Claude Opus 5 (subagent, inherited session model) | high |
| Adversarial review — hooks, house style | Claude Opus 5 (subagents, inherited session model) | high |
| Review fixes | Claude Opus 5 (subagent, inherited session model) | high |
| Mutation verification | Claude Opus 5 — run directly, not delegated | high |
| Design revision + browser verification | Claude Opus 5 | high |
| Verification | `npm run verify` — green locally; CI pending | — |

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2026 The offlinecv Authors
 
+import { useCallback, useEffect, useMemo, useState } from "react";
 import {
   Card,
   CapabilityStrip,
@@ -20,7 +21,13 @@ import { ShareWithExtensionBar } from "./components/features/ShareWithExtensionB
 import { useAnalyzedResume } from "./hooks/useAnalyzedResume.ts";
 import { useResumeLibrary } from "./hooks/useResumeLibrary.ts";
 import { useReplaceResumeOnDrop } from "./hooks/useReplaceResumeOnDrop.ts";
-import { departToJobs } from "./lib/jobs-departure.ts";
+import { useAutoRestoreResume } from "./hooks/useAutoRestoreResume.ts";
+import {
+  departToJobs,
+  departToJobsAndNavigate,
+} from "./lib/jobs-departure.ts";
+import { deriveJourney, type JourneyStageId } from "./lib/journey.ts";
+import type { LoadedResume } from "./lib/resume-library.ts";
 import { isScoreRevealed } from "./lib/contact.ts";
 
 export default function App() {
@@ -42,6 +49,23 @@ export default function App() {
   // Local-first resume library (#322) — save/reload parsed resumes without
   // re-uploading. Loading hydrates the "done" state from the cached parse.
   const library = useResumeLibrary();
+  // The one mapping from a stored record to the "done" state, shared by the
+  // explicit Load button below and the cold-mount auto-restore (#812) — two
+  // callers of the same hydration must not be able to drift into hydrating
+  // different fields.
+  const hydrateFromLibrary = useCallback(
+    (loaded: LoadedResume) => {
+      loadSavedResume({
+        fileName: loaded.filename,
+        fileSize: loaded.fileSize,
+        bytes: loaded.bytes,
+        sourceKind: loaded.sourceKind,
+        result: loaded.result,
+        score: loaded.score,
+      });
+    },
+    [loadSavedResume],
+  );
   const onLoadSavedResume = async (id: string) => {
     const loaded = await library.load(id);
     if (loaded === undefined) {
@@ -54,15 +78,19 @@ export default function App() {
       );
       return;
     }
-    loadSavedResume({
-      fileName: loaded.filename,
-      fileSize: loaded.fileSize,
-      bytes: loaded.bytes,
-      sourceKind: loaded.sourceKind,
-      result: loaded.result,
-      score: loaded.score,
-    });
+    hydrateFromLibrary(loaded);
   };
+
+  // #812 — bring the most recently saved résumé back on a cold visit, so the
+  // journey rail's first stage reads as satisfied instead of claiming progress
+  // over an app that silently forgot the user's work. Spent once per page
+  // lifetime and only ever from `idle`, so it never fights a drop, a library
+  // Load, or a `reset()` — see the hook.
+  useAutoRestoreResume({
+    phase: state.phase,
+    library,
+    onRestore: hydrateFromLibrary,
+  });
 
   // Once a parse is done the inline DropZone is gone; this restores drag-and-
   // drop so a new resume can replace the current one (confirm-gated, since it
@@ -90,6 +118,93 @@ export default function App() {
   const showingDraftPrompt =
     state.phase === "authoring" && state.pendingDraft !== null;
 
+  // ── The L1 journey rail (#812) ────────────────────────────────────────────
+  //
+  // Every input is derived during render from state this component already
+  // holds. Nothing here may move into a mount-only `useEffect(…, [])`: the
+  // `/jobs/` → `/` return leg is a bfcache restore that never remounts the
+  // tree, so a rail computed at mount would be frozen at whatever it said
+  // before the trip (the #783 defect, and the reason `useTailorHandoff`
+  // listens on `pageshow`).
+
+  // A tab the rail asked `ResultDetailTabs` to open. The nonce is what makes
+  // it an event rather than a value — asking for the tab you are already on is
+  // a repeatable click (see `ResultDetailTabs`).
+  const [tabRequest, setTabRequest] = useState<
+    { id: string; nonce: number } | undefined
+  >(undefined);
+  const requestTab = useCallback((id: string) => {
+    setTabRequest((prev) => ({ id, nonce: (prev?.nonce ?? 0) + 1 }));
+  }, []);
+
+  // JD steering, reported up from `ResultDetailTabs` (which owns
+  // `useTailorHandoff` — see the prop's docblock for why it is not lifted).
+  const [jdContext, setJdContext] = useState<string | null>(null);
+
+  // The mark is cleared HERE, on the canonical "the résumé genuinely changed"
+  // token, rather than by waiting for the reporter to say so. The reporter is
+  // not always mounted: `Result` short-circuits to `LimitedParsingCard` on a
+  // `fonts_unmappable` parse, and `phase: "authoring"` renders
+  // `ReconstructedResume` directly with no `Result` at all — in both cases
+  // `displayResult` is non-null, so a `hasResume` re-guard cannot see the
+  // staleness either, and the rail would keep claiming Tailor with a ✓ over a
+  // different (or blank) résumé for the rest of the page's life. `parseKey` is
+  // null in idle/parsing, the pristine `state.result` in done, and
+  // `authoring:<generation>` while authoring, so one dep covers reset, replace,
+  // blank authoring and the fonts-unmappable branch.
+  //
+  // It cannot clobber valid steering on the ordinary path. The bfcache return
+  // leg from `/jobs/` does not change `parseKey` at all, so this never fires
+  // there; and on a cold reload, `useTailorHandoff` reports via a state update
+  // scheduled from an effect, which lands in a LATER commit than the one this
+  // runs in — so the report always follows the clear rather than being erased
+  // by it. Deps hand-audited both directions (`exhaustive-deps` is NOT enforced
+  // — CLAUDE.md): `parseKey` is the only input, and `setJdContext` is a
+  // React-guaranteed-stable setter that a dep list must not list.
+  useEffect(() => {
+    setJdContext(null);
+  }, [parseKey]);
+
+  // A blank-authoring session counts: a résumé the user is writing from
+  // scratch is still a résumé, and `displayResult` is exactly "there is
+  // something on screen to fix, match, tailor and download".
+  const hasResume = displayResult !== null;
+  // No `hasResume &&` guard on the steering: "steering implies a résumé" is
+  // normalized inside `deriveJourney` itself now (#812), so it holds for every
+  // caller rather than for the two that remembered.
+  const journeyState = useMemo(
+    () =>
+      deriveJourney({
+        entry: "root",
+        hasResume,
+        jdSteering: jdContext !== null,
+      }),
+    [hasResume, jdContext],
+  );
+
+  const onJourneySelect = useCallback(
+    (id: JourneyStageId) => {
+      // The one route off `/` that leaves the document. Through the shared
+      // helper, never hand-rolled: a second definition of "hand the parse over
+      // and mark the departure" is exactly what shipped #700.
+      if (id === "match") {
+        departToJobsAndNavigate(displayResult?.canonical.fields);
+        return;
+      }
+      // `add` is the top of this page — the drop zone when idle, the parse
+      // header (with its Start over control) once a résumé is loaded. The
+      // shell has already scrolled there, so there is nothing else to do; in
+      // particular this must not reset, which would discard the user's work on
+      // a navigation control.
+      if (id === "add") return;
+      // Fix it / Tailor / Download all live behind the reconstructed tab —
+      // the editor, the rewrite affordance and the Download PDF button are all
+      // in it — so all three land there even from another tab.
+      requestTab("reconstructed");
+    },
+    [displayResult, requestTab],
+  );
+
   return (
     // `chips` is deliberately NOT passed: PageShell renders that slot in the
     // header on every phase, so the capability strip stayed pinned above the
@@ -107,7 +222,11 @@ export default function App() {
     // alone on the header-right instead of sharing it. `/jobs` still passes
     // one: it opens straight into a form with no headline of its own, so
     // there the header line is the only orientation.
-    <PageShell badge="alpha" onSavedJobsNavigate={goToSavedJobs}>
+    <PageShell
+      badge="alpha"
+      onSavedJobsNavigate={goToSavedJobs}
+      journey={{ state: journeyState, onSelect: onJourneySelect }}
+    >
       {(state.phase === "idle" ||
         state.phase === "parsing" ||
         state.phase === "error") && (
@@ -288,6 +407,8 @@ export default function App() {
               sourceKind={state.sourceKind}
               onReset={reset}
               edit={edit}
+              requestedTab={tabRequest}
+              onJdContextChange={setJdContext}
             />
             {/* Save-to-library affordance (#322) — saves the edited parse +
                 source bytes so this resume can be reloaded without re-uploading. */}
--- a/src/components/Result.tsx
+++ b/src/components/Result.tsx
@@ -43,6 +43,12 @@ interface ResultProps {
   onReset: () => void;
   /** Lifted edit state (#82) — threaded to ReconstructedResume for inline edits. */
   edit: EditableParse;
+  /** A tab the journey rail asked for (#812) — forwarded untouched to
+   *  `ResultDetailTabs`, which owns the tab state. */
+  requestedTab?: { id: string; nonce: number };
+  /** JD steering reported back up so `/`'s rail can mark the Tailor stage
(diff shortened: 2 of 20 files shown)


s-annam (commented):
LGTM. 0 Blocking findings. All 7 core acceptance criteria of #812, responsive layout constraints (320px–1440px), and accessibility requirements are thoroughly met. Clean architecture, robust test coverage (5,566 tests green), and the docblock nit has been addressed directly.

## Fixed in fc8006f

- [`src/components/features/PageShell.tsx:30,90`](file:///Users/annam/offlinecv/src/components/features/PageShell.tsx#L30-L95) — Updated `PageShell` header and `journey` prop docblocks to reflect the landed full-width rail row architecture at all breakpoints rather than the superseded `lg:` in-header placement.

## Acceptance Criteria Verification (#812)

- [x] **Dedicated `JourneyRail` component**: Renders on both HTML entries (`/` and `/jobs/`) via `PageShell`.
- [x] **Render-time derivation**: Stage state and guidance derivation (`deriveJourney`, `useJourneyGuidance`) derive purely during render, avoiding mount-only `useEffect(…, [])` bfcache traps (#783).
- [x] **Cold-mount auto-restore**: `useAutoRestoreResume` rehydrates the newest saved IndexedDB résumé record into analysis state on `/`, guarded against file drop races, dismissed state, and pending tailor handoffs.
- [x] **Ungated navigation & guidance**: All stages remain clickable; unpopulated stages render lightweight `JourneyGuidance` cards with prerequisite routing CTAs.
- [x] **Tab landing seam**: `Fix it`, `Tailor`, and `Download` land on `/`'s `reconstructed` tab via `requestedTab` nonce forwarding.
- [x] **Labeling & terminology**: Stage labels (`Add résumé`, `Fix it`, `Download`, `Match jobs`, `Tailor`) follow the #519 user-benefit rule and `Match jobs` avoids collision with L2 tabs.
- [x] **Responsive layout**: Dedicated full-width row across all breakpoints (320px, 375px, 414px, 768px, 1024px, 1440px) without row-wrapping or horizontal scroll.
- [x] **Sticky header & scroll padding**: Header is sticky (`sticky top-0 z-20`) with measured `scroll-padding-top` (10rem / 8rem) on the root `html` element so in-page anchors land clear of the bar.
- [x] **Accessibility**: Tri-state indicator uses distinct shape and glyph (filled numeral, ringed ✓, hairline numeral) + font weight rather than colour alone; screen reader words count over rendered stages; 44×44px touch targets; `prefers-reduced-motion` respected.

---

PR #827, opened 2026-08-14
chore(process): suppress AI attribution by setting, retire the Provenance block

s-annam:
## Summary

Claude Code injects its attribution into the agent's system prompt — a `Co-Authored-By:` trailer, a `🤖 Generated with [Claude Code]` PR badge, and a `Claude-Session:` trailer carrying a session URL. This repo fought that with prose: a `CLAUDE.md` hard rule *plus* a near-identical paragraph in each of four shipping skills, all telling the model to ignore an instruction it had just been handed. That is reconciliation work on every commit, it re-injects the exact trailer strings into context, and it lost often enough that ~16 `Claude-Session:` trailers reached `main`.

It is a setting. `.claude/settings.json` now carries:

```json
"attribution": {
  "commit": "",
  "pr": "",
  "sessionUrl": false
}
```

Because the harness no longer emits or suggests any of it, the `CLAUDE.md` rule is **deleted** rather than reworded (hard rules 3 → 2) and the skills keep one short pointer each instead of a restated ban. Net −102 lines.

This also **retires the `## Provenance` block** in favour of nothing.

## Why `attribution` and not `includeCoAuthoredBy`

`yc-software/qm` (the prompt for this change) uses `includeCoAuthoredBy: false`. That works, but it is the older single-boolean form, and reading the CLI (`2.1.232`, `nvS()`/`Udt()`) shows two things worth having:

- The settings schema marks it **`Deprecated: Use attribution instead`**.
- It covers **less**: it kills the `Co-Authored-By` trailer and the `🤖` badge, but **not** the `Claude-Session:` trailer / session URL — which is gated separately by `attribution.sessionUrl` and is the only part with an actual privacy cost on a public repo.

There is also a precedence trap worth writing down: the resolver is `if (attribution defines commit or pr) → use it; else if (includeCoAuthoredBy === false) → suppress`. Setting **both** silently ignores `includeCoAuthoredBy`. So: set `attribution` only. Project settings outrank a contributor's `~/.claude/settings.json`, so this holds for everyone.

## Why provenance is retired

The idea was sound — cross-model review is only legible if the PR says which model read it. The cost was not:

- A spawn requests a model **alias** (`model: opus`) and never learns what it resolved to, so **every row had to be self-reported** — one round-trip per subagent, then `gh pr view` + text surgery + `gh pr edit` per PR to write the block idempotently, repeated on every `/revise-pr` round, with the machinery duplicated across four skills.
- It wasn't reliably true. On the #711/#712 batch, two subagents spawned with the **same** `model: sonnet` alias self-reported **different** names (`Claude Sonnet 5` vs `Claude Sonnet 4.6`) — one row was wrong, with no way to tell which from the orchestrator's side.

A block a reader can't trust launders a guess into a fact. What survives is the cheap, first-hand form that never needed a round-trip: `/pr-review`'s one-line `Reviewed by: <model> (<effort>)` sign-off. **Kept deliberately** — flag it if you'd rather that went too.

## Changes

| File | What |
|---|---|
| `.claude/settings.json` | add the `attribution` block |
| `CLAUDE.md` | delete the "Never in git" hard rule; 3 rules → 2 |
| `docs/CONTRIBUTING-PROCESS.md` | rewrite "AI provenance" → "AI attribution": the setting, the deprecated alternative + precedence trap, and why provenance was retired |
| `open-pr` | drop Step 5.5, the provenance-records input, and the body template's block |
| `revise-pr` | drop Step 5.5 (two `gh` calls per round); old PRs keep their block untouched |
| `implement-batch` | drop the subagent `Model:` self-report instruction, the `Model:` return field, and the assembly step; Phase 5 renumbers 12–16 → 12–15 |
| `pr-review` | keep `Reviewed by:`, drop the trailer ban it no longer needs |
| `CONTRIBUTING.md` | pointer wording |

Existing `## Provenance` blocks on open/merged PRs are left alone — the skills are told not to update *or* delete them. The ~16 `Claude-Session:` trailers already in `main` stay too: those URLs are auth-gated, so the exposure is an identifier and not content, and a history rewrite on a public repo costs far more than it recovers (that call is unchanged from #458).

## Review focus

- `docs/CONTRIBUTING-PROCESS.md` — is the "setting, not prose" rationale worth the ~40 lines it now occupies, or should it compress to the config block plus the precedence trap?
- `.claude/skills/implement-batch/SKILL.md` — Phase 5 renumbered; do the cross-references (`Phase 5 step 12`, `steps 12–14`, `step 11e`, `Phase 5 step 14`) all still point at the right steps?
- Does `/pr-review`'s surviving `Reviewed by:` line belong, given everything else went?

## Test plan

- [x] `npm run verify` green via the pre-push hook (typecheck → lint → coverage → build → fallow, 8 changed files, no issues)
- [x] `.claude/settings.json` parses
- [x] This PR's own commit carries no attribution trailer
- [ ] Post-merge: confirm a fresh session's system prompt no longer carries the commit-trailer instruction

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -3,9 +3,11 @@
 Guidance for Claude Code (claude.ai/code) working in this repository.
 
 **This file is about writing code.** The *why* behind the process — merge-queue mechanics, the
-provenance rationale, the full PII policy, deploy, license — lives in
+AI-attribution setting, the full PII policy, deploy, license — lives in
 `docs/CONTRIBUTING-PROCESS.md`. You do not need it to write a change. The three rules below are
 the exception: they stay here because breaking them is silent, and for two of them, permanent.
+Everything else that used to sit here is enforced by config or by a gate, and prose that restates
+a setting only gives you something to reconcile.
 
 ## Hard rules (no exceptions)
 
@@ -28,10 +30,9 @@ the exception: they stay here because breaking them is silent, and for two of th
   `pdftotext <file>.pdf - | head -40` — the two cover different surfaces. `pdftotext` prints only
   the drawn page, so it cannot see a `tel:`/`mailto:` **link annotation** or the Info dict, and
   both have leaked here. Never trust the PR prose over the binary.
-- **Never in git.** No `Co-Authored-By:` trailer naming a model, no `Claude-Session:` trailer, no
-  `https://claude.ai/code/session_…` URL, no `🤖 Generated with …` badge — not in a commit
-  message, not in a PR body. The Bash tool's default commit template suggests them; ignore it.
-  Model provenance is useful and belongs in the **PR body only**, as a `## Provenance` block.
+- **No AI attribution in git.** Claude Code has this off by config (`attribution` in
+  `.claude/settings.json`); on any other harness it is on you. Never commit attribution
+  trailers (like `Co-Authored-By:` or `Claude-Session:`) or PR badges.
 - **One commit per PR.** `main` merges through a merge queue that derives the squash message from
   the branch, so a multi-commit PR lands `wip` and `fix lint` in `main` forever. Collapse the
   branch to a single commit before it reaches the queue.
--- a/docs/CONTRIBUTING-PROCESS.md
+++ b/docs/CONTRIBUTING-PROCESS.md
@@ -11,7 +11,9 @@ the binding one-liners sit in `CLAUDE.md` → **Hard rules** (always in an agent
 fixture-PII check is a directory-scoped `CLAUDE.md` in `tests/fixtures/pdfs/`, and each shipping
 skill (`/open-pr`, `/pr-review`, `/revise-pr`, `/implement-batch`, `/pr-ready`) carries its own
 operational copy. Nothing here is load-bearing on its own — if you change a rule, change it in
-those places too, not only here.
+those places too, not only here. The exception is **AI attribution**, which has no `CLAUDE.md`
+counterpart on purpose: it is enforced by a setting in `.claude/settings.json`, and a prose rule
+restating a setting is context an agent has to spend reconciling.
 
 ## Test fixtures — PII policy (non-negotiable)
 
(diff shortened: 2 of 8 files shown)


s-annam (commented):
### Stance & Findings

0 Blocking / 0 Secondary / 0 Nits. The changes are clean, well-factored, and achieve their exact goal.

### Architectural & Decision Verification

1. **Retaining Code Review Attribution**: PR #827 makes the right trade-off. The brittle, multi-party `## Provenance` markdown table in the PR body is retired (eliminating fragile subagent polling, text surgery, and desynced self-reporting), while **first-hand review attribution** is explicitly preserved in `.claude/skills/pr-review/SKILL.md` (`Reviewed by: <model> (<effort>)`). The author will always see exactly which model reviewed their PR directly on the review submission itself.
2. **Configuration vs Prose**: Moving commit trailers, PR badges, and session URLs into `.claude/settings.json` under `"attribution"` eliminates prompt bloat, removes conflicting instructions between system prompt and rules, and prevents session leakage permanently.
3. **Skill & Script Consistency**: Verified all step renumberings and cross-references in `.claude/skills/implement-batch/SKILL.md` (steps 12–15, lines 66, 431, 475, 487), `open-pr/SKILL.md`, and `revise-pr/SKILL.md`. No broken references or command bugs.
4. **Verification Gates**: All test suites, typechecks, lints, baselines, and package checks pass.

---

CLAUDE.md
```diff
@@ -3,9 +3,11 @@
 Guidance for Claude Code (claude.ai/code) working in this repository.
 
 **This file is about writing code.** The *why* behind the process — merge-queue mechanics, the
-provenance rationale, the full PII policy, deploy, license — lives in
-`docs/CONTRIBUTING-PROCESS.md`. You do not need it to write a change. The three rules below are
-the exception: they stay here because breaking them is silent, and for two of them, permanent.
+AI-attribution setting, the full PII policy, deploy, license — lives in
+`docs/CONTRIBUTING-PROCESS.md`. You do not need it to write a change. The two rules below are
+the exception: they stay here because breaking them is silent, and for one of them, permanent.
+Everything else that used to sit here is enforced by config or by a gate, and prose that restates
```
Samhit21, line 9: [Suggestion]: `AGENTS.md` symlinks to this file, so "enforced by config" is false for every non-Claude agent that reads it

`AGENTS.md` is a symlink to `CLAUDE.md` (mode 120000, target `CLAUDE.md`), which is how this repo serves agents that aren't Claude Code — Codex, Cursor, and friends read `AGENTS.md` by convention. `.claude/settings.json` is read by Claude Code and nothing else.

So for that reader, this PR is a net removal: the "Never in git" hard rule was the only attribution instruction they ever saw, and the `attribution` block that replaces it is invisible to them. The sentence they now read in its place asserts that the thing is "enforced by config or by a gate" — which is exactly the claim that isn't true on their harness. An agent that trusts it has been told the guard rail exists.

The cost of keeping coverage is one line rather than the paragraph that was deleted — the point is that it survives for a reader whose harness has no setting to honour:

```markdown
- **No AI attribution in git.** Claude Code has this off by config
  (`attribution` in `.claude/settings.json`); on any other harness it is on you.
```

Worth noting this repo does run non-Claude models against itself — the review above this one is signed `Gemini 3.7 Flash`.

CLAUDE.md
```diff
@@ -3,9 +3,11 @@
 Guidance for Claude Code (claude.ai/code) working in this repository.
 
 **This file is about writing code.** The *why* behind the process — merge-queue mechanics, the
-provenance rationale, the full PII policy, deploy, license — lives in
-`docs/CONTRIBUTING-PROCESS.md`. You do not need it to write a change. The three rules below are
-the exception: they stay here because breaking them is silent, and for two of them, permanent.
+AI-attribution setting, the full PII policy, deploy, license — lives in
+`docs/CONTRIBUTING-PROCESS.md`. You do not need it to write a change. The two rules below are
+the exception: they stay here because breaking them is silent, and for one of them, permanent.
```
Samhit21: [Nit]: "for one of them, permanent" undercounts the two rules that survived

Going 3 → 2 rules, the count moved from "two of them" to "one of them" — but both survivors describe permanent damage. Fixture PII is permanent (public repo, `git filter-repo` + a Support ticket). One-commit-per-PR is too, in its own words 25 lines down: a multi-commit PR "lands `wip` and `fix lint` in `main` forever."

Reads like "for both of them, permanent."

Samhit21 (commented):
## PR Review: chore(process): suppress AI attribution by setting, retire the Provenance block

### Summary
The central move is right and the homework behind it is real: `attribution` over the deprecated `includeCoAuthoredBy` is correct per the published settings schema, the precedence trap is worth writing down, and retiring `## Provenance` is well argued — a table that costs a round-trip per subagent and still can't be verified is worse than no table, and the #711/#712 evidence of two same-alias spawns self-reporting different names settles it. Replacing prose that argues with the harness by a setting the harness obeys is the correct direction, and −102 lines of reconciliation work is the right shape of win. Cross-references after the Phase 5 renumber all resolve, and no dangling `## Provenance` references survive anywhere on the branch.

One gap: the config that replaces the prose is Claude-Code-only, and `AGENTS.md` is a symlink to `CLAUDE.md`, so non-Claude agents lose their only instruction and are handed a sentence telling them a guard rail exists that doesn't on their harness.

### Findings
1. `CLAUDE.md:9` — `AGENTS.md` symlinks here; `.claude/settings.json` is Claude-Code-only, so "enforced by config" is false for any other harness reading this file, and the rule that used to cover them is deleted. [Suggestion]
2. `CLAUDE.md:8` — "for one of them, permanent" undercounts; both surviving rules describe permanent damage. [Nit]

On the third review-focus question — yes, `/pr-review`'s `Reviewed by:` line should stay. It's the one attribution here that is first-hand by construction: the model writing the line knows its own name, with no alias to resolve and no round-trip to trust. That's precisely the property the Provenance table lacked, so keeping one and retiring the other is a coherent split rather than an inconsistency.

Also checked and dropped before writing this up: I suspected an older CLI might honour `attribution.commit` but not `sessionUrl`, leaving the session trailer live now that the prose backstop is gone. Probing 2.1.34 shows no `Claude-Session`/`sessionUrl` strings at all — that version can't emit a trailer it doesn't implement. Not a finding, noted so it doesn't get re-raised.

### Verdict
**Action: COMMENT**

CLAUDE.md
s-annam: Accepted. Restored a simplified version of the AI attribution rule to cover non-Claude agents reading AGENTS.md (symlinked here). Fixed in 85478aa.

CLAUDE.md
s-annam: Accepted. Since the AI attribution rule is restored, we now have three rules under Hard Rules. Updated the preamble to three rules and the permanent count back to 'two of them' (referring to Fixture PII and One commit per PR). Fixed in 85478aa.

s-annam:
This PR has been revised to address the review feedback from @Samhit21:

1. **Rule Coverage for Non-Claude Agents**: Restored a simplified "No AI attribution in git" rule in `CLAUDE.md` to ensure non-Claude agents reading `AGENTS.md` (which symlinks here) are aware they should avoid AI attribution/trailers.
2. **Count Correction**: Restored the rule count to three, and updated the permanent count in the preamble back to "two of them" (referring to *Fixture PII* and *One commit per PR*).
3. **Squashed & Cleaned**: Squashed the branch changes into a single clean commit to align with merge-queue requirements.

Summary of changes in this PR:
* `.claude/settings.json`: Configured `attribution` block to suppress commit/PR trailers and session telemetry.
* `CLAUDE.md`: Restored simplified "No AI attribution in git" rule and updated rule/permanence counts.
* `docs/CONTRIBUTING-PROCESS.md`: Swapped "AI provenance" for "AI attribution" details and documented why the provenance block is retired.
* `.claude/skills/`: Updated `implement-batch`, `open-pr`, `pr-review`, and `revise-pr` skills to retire the provenance table, simplify attribution rules, and correct step references.

docs/CONTRIBUTING-PROCESS.md
```diff
@@ -11,7 +11,9 @@ the binding one-liners sit in `CLAUDE.md` → **Hard rules** (always in an agent
 fixture-PII check is a directory-scoped `CLAUDE.md` in `tests/fixtures/pdfs/`, and each shipping
 skill (`/open-pr`, `/pr-review`, `/revise-pr`, `/implement-batch`, `/pr-ready`) carries its own
 operational copy. Nothing here is load-bearing on its own — if you change a rule, change it in
-those places too, not only here.
+those places too, not only here. The exception is **AI attribution**, which has no `CLAUDE.md`
```
Samhit21, line 14: [Suggestion]: this now contradicts the restored hard rule — non-blocking

This paragraph says AI attribution "has no `CLAUDE.md` counterpart on purpose: it is enforced by a setting in `.claude/settings.json`, and a prose rule restating a setting is context an agent has to spend reconciling." As of `85478aa` there is a counterpart — `CLAUDE.md:33-35`, restored deliberately to cover agents reading `AGENTS.md`.

What makes it worth the two-line fix rather than leaving it: the sentence sits four lines under this file's own instruction that *"if you change a rule, change it in those places too, not only here"* — and it currently tells the next person editing attribution policy that there is no second copy to keep in sync. That's the exact failure mode this PR exists to remove, pointed at itself.

Something like:

```markdown
those places too, not only here. **AI attribution** is the one rule enforced primarily by
configuration (`attribution` in `.claude/settings.json`) rather than prose — the `CLAUDE.md`
one-liner survives only because that setting is Claude Code's alone, and `AGENTS.md` symlinks
to `CLAUDE.md` for every other harness.
```

Not gating the approval — the branch needs a re-squash before the queue regardless, so it's cheap to fold in there.

Samhit21 (approved):
## PR Review (round 2): chore(process): suppress AI attribution by setting, retire the Provenance block

### Summary
Both round-1 findings are addressed properly in `85478aa`. The restored hard rule is the right shape — it names the config for the harness that honours it and puts the obligation back on the reader whose harness doesn't, and adding the literal trailer strings (`Co-Authored-By:`, `Claude-Session:`) is a real improvement over what I suggested, since a non-Claude agent now knows what to *not* write rather than just that a policy exists. The count fix is correct and the reasoning in the reply (Fixture PII + one-commit-per-PR as the permanent two) is a defensible call, so I'm not re-litigating it.

One thing the fix knocked loose, non-blocking: `docs/CONTRIBUTING-PROCESS.md:14-16` still asserts that AI attribution "has no `CLAUDE.md` counterpart on purpose," which stopped being true the moment the rule went back. Detail inline. It's two lines and it doesn't gate anything.

### Findings
1. `docs/CONTRIBUTING-PROCESS.md:14` — the "no `CLAUDE.md` counterpart on purpose" claim contradicts the restored hard rule. [Suggestion, non-blocking]

### Verdict
**Action: APPROVE**

The substance is right: the setting is the correct mechanism, the schema evidence backs the `attribution` over `includeCoAuthoredBy` choice, retiring an unverifiable Provenance table is well argued, and the coverage gap for non-Claude harnesses is now closed. Worth folding the doc sentence in before the queue since the branch has to be re-squashed anyway, but I'd rather not hold the PR for it.

Issue #653, opened 2026-07-29
Shared lexicon leaves: unify bullets, dates, corporate suffixes, and titles


s-annam:
> **Split into four sub-issues (2026-08-27).** This issue is now the parent. The original body described four consolidations as one plan, but #646 itself specifies them as "four independent PRs" with very different risk profiles — one changes scoring behaviour, two are behaviour-preserving, one is an accuracy lever. Keeping them in one issue meant the safe three were gated behind the risky one. Implementation detail moved to the children; nothing was dropped.

### Problem

Vocabularies — bullet glyphs, date tokens, corporate suffixes, title surface forms — exist as divergent copies across the parser, the scorer and the edit layer. Duplicated vocabulary is this codebase's proven bug factory: #641 had to be diagnosed and fixed **per copy**, and today a `⁃`- or `—`-bulleted résumé is classified inconsistently between the parser and the scorer.

An architecture review of "should offlinecv adopt a plugin architecture?" concluded **no** for every subsystem except this one. Vocabulary is the single genuine "many independent units" case in the tree — and the unit is a *constant*, not a class. The right shape is data tables, and `sections.config.json` → `sections.config.ts:157-215` already proves the pattern in-tree.

### Sub-issues

| # | Part | Risk | Notes |
|---|---|---|---|
| #915 | (a) Bullet glyphs | **High** | Changes scores. Needs `ATS_SCORE_ALGO_VERSION` bump, 54-snapshot rebaseline, per-glyph human adjudication. Do **not** batch. |
| #916 | (b) Date tokens | Low | Behaviour-preserving, mechanical, 7 copies. Snapshots stay byte-identical. |
| #917 | (c) Corporate suffixes | Low | Behaviour-preserving. Must **not** collapse the four memberships. |
| #918 | (d) Title surface forms | Medium | Accuracy lever, not hygiene. Measured against the #654 scoreboard. Feeds #492/#543. |

### Sequencing

(b) and (c) are independent and can land in any order. (a) lands alone with its own reviewed rebaseline. (d) wants the #654 ground-truth scoreboard as its harness, which is already closed, so it is unblocked.

### Acceptance criteria

- [ ] #915, #916, #917 and #918 all closed.
- [ ] No bullet-glyph, month/season, or corporate-suffix literal remains duplicated across modules.
- [ ] Every surviving divergence between two vocabularies is a named derived constant with a docblock saying why it diverges — not a re-typed literal.

Parent: #646 (roadmap item 5).

PR #838, opened 2026-08-15
fix: strip trailing middot separator from education degree field

### LINKED ISSUE #835: [parser] education — the degree field keeps the header line's trailing middot ("Computer Science ·")

`unfiled` is capped at `UNFILED_TRUTH_CEILING = 10` (`src/lib/heuristics/corpus.test.ts:186`) and is saturated at 10/10 today. This issue files one of them.

## What's wrong

The education degree keeps the header line's own separator glyph as if it were part of the subject.

`tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-minimal.pdf` draws (`pdftotext -layout`):

```
EDUCATION
Springfield State University
B.S. Computer Science · 2020
```

Observed: `field` = `"Computer Science ·"`, so the degree renders as **`B.S. Computer Science ·`**.

Recorded today as a `status: "unfiled"` `knownWrong` entry on `education.degree` in `tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-minimal.truth.json`. The middot is punctuation belonging to the line's layout — it separates the degree from its graduation year — not part of the field of study.

It is small, but it is user-visible everywhere the degree is: the reconstructed résumé card, the Download PDF, and the re-parse of that PDF. It also reads as sloppiness on the very first education entry a user sees.

## Root cause

`src/lib/heuristics/extract/education.ts:443-461`, `cleanField`. The date IS cut correctly; the separator that preceded it is not.

```ts
function cleanField(raw: string): string | undefined {
  let f = raw
    .split(/\s{2,}|\s*\|\s*/)[0]      // column break or pipe ends the field
    .trim();
  f = f.replace(CLEAN_FIELD_DATE_RE, "");   // "· 2020" → "·" survives, "2020" cut
  f = f.replace(/[,;]\s*(?:minor|major|gpa|concentration)\b.*$/i, "");
  f = f.replace(/^[\s,;:–\-—]+|[\s,;:–\-—]+$/g, "").trim();   // ← no `·` in the class
  return f.length > 0 ? f : undefined;
}
```

Trace `"Computer Science · 2020"`:

1. No 2+ space run and no pipe, so the first split is a no-op.
2. `CLEAN_FIELD_DATE_RE` removes the trailing year `2020`, leaving `"Computer Science ·"`.
3. The GPA/Minor cut does not apply.
4. The edge-punctuation strip is `/^[\s,;:–\-—]+|[\s,;:–\-—]+$/g` — comma, semicolon, colon, en dash, hyphen, em dash. **`·` (U+00B7) is not in the character class**, so the orphaned separator survives.

This is one more instance of a known repo-wide hazard: the bullet/separator glyph vocabulary is duplicated across many regexes in the parser, exporter and scorer, and they have drifted apart. A fix here should not widen the drift — see below.

## Implementation plan

1. **`src/lib/heuristics/extract/education.ts` — add the separator glyphs to `cleanField`'s edge strip.** At minimum `·` (U+00B7 middle dot) and `•` (U+2022 bullet); both are used as inline separators by real templates and by our own exporter (`MIDDOT_SEGMENT_SEP = " · "`, `src/lib/pdf/render-ats-pdf.ts`).

   ```ts
   f = f.replace(/^[\s,;:·•–\-—]+|[\s,;:·•–\-—]+$/g, "").trim();
   ```

   The strip is already edge-anchored, so an interior middot in a genuine two-part field is untouched.

2. **Prefer a shared constant over a fourth hand-written glyph class.** Before adding literals, check whether this file (or `src/lib/heuristics/regex.ts` / `line-primitives.ts`) already exports a separator/bullet glyph set this strip can consume. If one exists, use it. If not, do NOT introduce a new global vocabulary as a side effect of this fix — a local widening is the right scope here, and unifying the glyph vocabularies is already tracked as its own work (#653, shared lexicon leaves: bullets, dates, corporate suffixes, titles). Leave a one-line comment pointing at #653 so the next reader knows this class is a known duplicate, not an oversight.

3. **Check the symmetric case.** The same `cleanField` runs on every degree-bearing line, so a LEADING separator (`"· Computer Science"`, produced when a middot-joined header is split at the wrong point) is fixed by the same edit — the strip is anchored at both ends. Add a test for it.

4. **Flip the ground-truth entry** in `tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-minimal.truth.json` → `education.degree`, to `status: "open"` with this issue's number. If the fix lands in the same PR, **delete** the entry instead. Do not edit the truth VALUES — a truth file records what the page draws and is never edited to match the parser (see its `provenance`).

5. **Lower `UNFILED_TRUTH_CEILING`** (`src/lib/heuristics/corpus.test.ts:186`) by 1. Sibling issues from the same audit also lower it — on a rebase conflict, take the LOWER number.

## Acceptance criteria

- [ ] `google-docs-skia-proxy-minimal.pdf` parses `degree = "B.S."` with `field = "Computer Science"` — no trailing glyph.
- [ ] `cleanField("Computer Science · 2020")` → `"Computer Science"`; `cleanField("· Computer Science")` → `"Computer Science"`; `cleanField("Computer Science • ")` → `"Computer Science"`. Unit tests in the education test module.
- [ ] An interior separator in a genuine field is preserved: `cleanField("Mathematics · Statistics")` → `"Mathematics · Statistics"` (or whatever the current behaviour is — assert it, do not change it in this issue).
- [ ] `cleanField` still returns `undefined` when nothing substantive survives — a field of only separators (`"·"`) must not become an empty-string field.
- [ ] `npx vitest run src/lib/heuristics/corpus.test.ts` passes; `education.degree` recall improves on the target fixture and falls on none of the other 57.
- [ ] `npx vitest run src/lib/heuristics/corpus-roundtrip.test.ts` passes with **no new `KNOWN_FAILURES` baseline rows**. Worth checking specifically: the exporter re-emits `Degree · Year` using the same middot, so the clean field must survive parse → export → re-parse rather than picking the glyph back up.
- [ ] `npm run check:baselines` reports 1 fewer unfiled entry.

myukitty:
### Summary of Changes
Fixes #835.

When parsing education sections from resume PDFs (such as Google Docs generated PDFs), degree lines often contain trailing middots (`·`) or bullet characters (`•`) used as visual separators before fields that were not extracted or omitted.

`cleanField` in `src/lib/heuristics/extract/education.ts` previously stripped edge punctuation `[\s,;:–\-—]`. This change widens the character set to include `·` (U+00B7) and `•` (U+2022) so trailing and leading separators are trimmed cleanly, without affecting internal characters.

### Verification
- Added comprehensive unit tests in `src/lib/heuristics/extract/education.test.ts` covering trailing middots, trailing bullets, leading middots/bullets, compound internal field preservation, and separator-only fields (61/61 tests pass).
- Updated `tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-minimal.truth.json` by removing the resolved `knownWrong` annotation for `education.degree`.
- Verified `npm run check:baselines`, `npm run check:fixtures`, `npm run typecheck`, and `npm run lint` all pass.

### DIFF
--- a/src/lib/heuristics/corpus.test.ts
+++ b/src/lib/heuristics/corpus.test.ts
@@ -183,7 +183,7 @@ const TRUTH_ANNOTATED_FIELD_FLOOR = 150;
  * `npm run check:baselines` on every run, and bounded here — undescribed debt may
  * not GROW. File the issue and flip the entry to `open`; then lower this.
  */
-const UNFILED_TRUTH_CEILING = 10;
+const UNFILED_TRUTH_CEILING = 9;
 
 /** Generator category = the fixture root's immediate subdirectory. */
 function categoryOf(repoRelPdfPath: string): string {
--- a/src/lib/heuristics/extract/education.test.ts
+++ b/src/lib/heuristics/extract/education.test.ts
@@ -857,3 +857,66 @@ describe("extractEducation — one-line 'Degree Field, Institution' + en-dash da
     expect(value[0].location).toBe("Boston, MA");
   });
 });
+
+describe("extractEducation — cleanField strips leftover edge middot and bullet separators (#835)", () => {
+  it("strips trailing middot separator from degree field", () => {
+    const { value } = extractEducation(
+      mkEduSection([
+        "Springfield State University",
+        "B.S. Computer Science ·",
+      ]),
+    );
+    expect(value).toHaveLength(1);
+    expect(value[0].degree).toBe("B.S.");
+    expect(value[0].field).toBe("Computer Science");
+  });
+
+  it("strips trailing bullet separator from degree field", () => {
+    const { value } = extractEducation(
+      mkEduSection([
+        "Springfield State University",
+        "Bachelor of Science in Mathematics •",
+      ]),
+    );
+    expect(value).toHaveLength(1);
+    expect(value[0].degree).toBe("Bachelor of Science");
+    expect(value[0].field).toBe("Mathematics");
+  });
+
+  it("strips leading middot or bullet from degree field", () => {
+    const { value } = extractEducation(
+      mkEduSection([
+        "Springfield State University",
+        "B.S. · Computer Science",
+      ]),
+    );
+    expect(value).toHaveLength(1);
+    expect(value[0].degree).toBe("B.S.");
+    expect(value[0].field).toBe("Computer Science");
+  });
+
+  it("preserves internal middot or ampersand inside compound field names", () => {
+    const { value } = extractEducation(
+      mkEduSection([
+        "Springfield State University",
+        "B.S. in Computer Science & Engineering ·",
+      ]),
+    );
+    expect(value).toHaveLength(1);
+    expect(value[0].degree).toBe("B.S.");
+    expect(value[0].field).toBe("Computer Science & Engineering");
+  });
+
+  it("returns undefined field when only separators survive", () => {
+    const { value } = extractEducation(
+      mkEduSection([
+        "Springfield State University",
+        "B.S. · • -",
+      ]),
+    );
+    expect(value).toHaveLength(1);
+    expect(value[0].degree).toBe("B.S.");
+    expect(value[0].field).toBeUndefined();
+  });
+});
+
--- a/src/lib/heuristics/extract/education.test.ts
+++ b/src/lib/heuristics/extract/education.test.ts
@@ -895,7 +895,35 @@ describe("extractEducation — cleanField strips leftover edge middot and bullet
     expect(value[0].field).toBe("Computer Science");
   });
 
-  it("preserves internal middot or ampersand inside compound field names", () => {
+  it("strips the separator orphaned by the trailing-date cut", () => {
+    // The fixture shape the issue was filed from: the date strip removes "2020"
+    // and leaves the "·" that divided it from the field.
+    const { value } = extractEducation(
+      mkEduSection([
+        "Springfield State University",
+        "B.S. Computer Science · 2020",
+      ]),
+    );
+    expect(value).toHaveLength(1);
+    expect(value[0].degree).toBe("B.S.");
+    expect(value[0].field).toBe("Computer Science");
+  });
+
+  it("preserves an INTERIOR middot joining a genuine two-part field", () => {
+    // The strip is anchored at both ends, so a separator that is real content
+    // survives. This is what stops a future widening from de-anchoring it.
+    const { value } = extractEducation(
+      mkEduSection([
+        "Springfield State University",
+        "B.S. Mathematics · Statistics",
+      ]),
+    );
+    expect(value).toHaveLength(1);
+    expect(value[0].degree).toBe("B.S.");
+    expect(value[0].field).toBe("Mathematics · Statistics");
+  });
+
+  it("preserves an internal ampersand inside a compound field name", () => {
     const { value } = extractEducation(
       mkEduSection([
         "Springfield State University",
@@ -919,4 +947,3 @@ describe("extractEducation — cleanField strips leftover edge middot and bullet
     expect(value[0].field).toBeUndefined();
   });
 });
-
--- a/src/lib/heuristics/extract/education.ts
+++ b/src/lib/heuristics/extract/education.ts
@@ -455,7 +455,13 @@ function cleanField(raw: string): string | undefined {
   // Cut a trailing "… , Minor in Economics" / "… , GPA: 3.8" note — a sub-field,
   // not part of the subject.
   f = f.replace(/[,;]\s*(?:minor|major|gpa|concentration)\b.*$/i, "");
-  // Strip leftover edge punctuation and separators (#835).
+  // Strip leftover edge punctuation and separators — the middot/bullet a
+  // template used to divide the header from a field this parse did not keep
+  // ("Computer Science · 2020" loses the year above, leaving the "·", #835).
+  // Anchored at BOTH ends, so an interior separator in a genuine two-part field
+  // ("Mathematics · Statistics") is untouched. This glyph class is a known
+  // duplicate of the ones in `regex.ts` / `line-primitives.ts`, not an
+  // oversight — unifying the separator vocabulary is tracked as #653.
   f = f.replace(/^[\s,;:·•–\-—]+|[\s,;:·•–\-—]+$/g, "").trim();
   return f.length > 0 ? f : undefined;
 }
--- a/src/lib/heuristics/extract/education.ts
+++ b/src/lib/heuristics/extract/education.ts
@@ -455,8 +455,8 @@ function cleanField(raw: string): string | undefined {
   // Cut a trailing "… , Minor in Economics" / "… , GPA: 3.8" note — a sub-field,
   // not part of the subject.
   f = f.replace(/[,;]\s*(?:minor|major|gpa|concentration)\b.*$/i, "");
-  // Strip leftover edge punctuation.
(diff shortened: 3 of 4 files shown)


s-annam (approved):
Approving. The fix is correct, correctly scoped, and lands exactly where #835 said it would — and thanks for picking this one up.

I verified the behaviour rather than reading the diff for plausibility. `cleanField` has exactly two call sites (`education.ts:534`, `:781`) and both operate on the education **field**, so widening the edge-strip class cannot reach `degree` or `institution`. Probing the parser directly across the separator shapes:

```
"B.S. Computer Science · 2020"    → degree "B.S.",  field "Computer Science"
"B.S. Mathematics · Statistics"   → degree "B.S.",  field "Mathematics · Statistics"
"B.S. ·"                          → degree "B.S.",  field undefined
```

The middle line is the one that matters: because your strip is anchored at both ends, a separator that is genuine content survives. That is the property that makes this fix safe rather than merely effective.

Gates, run locally on the branch (CI's `verify` and `fallow` are green too):

| Gate | Result |
|---|---|
| `npm run verify` | ✅ exit 0 — 5624 passed, 0 failed |
| `npm run check:fixtures` | ✅ 58 PDFs + 15 sidecars, all personas synthetic |
| `npm run check:baselines` | ✅ 9 unfiled (was 10) |
| `corpus.test.ts` + `corpus-roundtrip.test.ts` | ✅ 123 passed, no new `KNOWN_FAILURES` rows |

One caveat on those numbers, since it may bite you locally: a first run of `verify` on a loaded machine gave me 7 failures across `src/lib/pdf`, all `Hook timed out in 10000ms`. Re-run alone, that suite is 424/424. It is a known load flake (#825), not your change — if you see it, re-run before you go looking for a cause.

## Fixed in 0e706a9

I pushed a commit to your branch with the review's small items rather than sending them back to you. **Some context on why, since you are new to this repo and a maintainer pushing to your fork unannounced would otherwise read as a judgement on your work — it is not.** The house policy is: when a PR has zero blocking findings and a green build, the reviewer applies the remaining non-behavioural items themselves and approves the result, instead of listing four nits and waiting for a round-trip that adds nothing. Left to the normal cycle, that costs you a context-switch back into a diff you have already finished thinking about, and it costs the PR a day. The commit is bounded on purpose: only changes that **cannot alter what the code does**, gated on a green `npm run verify`. Anything behavioural stays yours to decide — which is exactly why the one remaining item below was *not* auto-fixed. Read the commit, and if you disagree with any of it, say so and I will revert it; it is a proposal in commit form, not a correction.

What it contains:

- **`education.test.ts` — the interior-middot case the issue's third acceptance criterion asks for.** The test named `preserves internal middot or ampersand inside compound field names` only exercised an ampersand; its input was `"B.S. in Computer Science & Engineering ·"`, which is a *trailing* middot — the case the two tests above it already covered. So `cleanField("Mathematics · Statistics")` was never pinned. Concretely: a later change adding `·` to the segment split on `education.ts:447` would truncate that field to `"Mathematics"` and this entire suite would still pass. The behaviour was already right; nothing guarded it.
- **`education.test.ts` — the shape the issue was actually filed from**, `"B.S. Computer Science · 2020"`, where the separator is orphaned by the trailing-date cut rather than drawn bare. Only the corpus fixture covered that path, so the unit test that names the bug did not reproduce it.
- **`corpus.test.ts:186` — `UNFILED_TRUTH_CEILING` 10 → 9.** This is the one item worth flagging as more than a nit. Deleting the `knownWrong` entry drops the unfiled count to 9, but the cap stayed at 10, so the ratchet quietly regains a free slot and the next PR that records a new `status: "unfiled"` disagreement lands with the gate green — the single thing that gate exists to refuse ("undescribed debt may not GROW"). Step 5 of the issue's implementation plan and the constant's own docblock both call for it. Heads-up for a rebase: sibling issues from the same audit also lower this number, and #835 says take the **lower** value on a conflict.
- **`education.ts:458` — the comment now points at #653.** Step 2 of the issue asked for this specifically: the glyph class you added duplicates ones in `regex.ts` and `line-primitives.ts`, and #653 tracks unifying them. Citing #835 alone reads to the next person as an oversight rather than a known, tracked duplicate. I also wrote down *why* the strip is anchored at both ends, since that is the non-obvious part.
- Trailing blank line at EOF.

## Nits

Non-blocking, and deliberately left for you because it changes behaviour — the auto-fix commit above is not allowed to.

- **`src/lib/heuristics/extract/education.ts:532` — the new glyphs are in `cleanField` but not in `parseDegreeAndField`'s leading-separator strip, which runs first.** That strip is `^\s*[-–—,:]\s*`, sitting between two `in`/`of` connective strips (`:531`–`:533`). Because `·`/`•` are only handled later in `cleanField`, a separator sitting *before* the connective blocks the connective strip:

  ```
  "B.S. · in Computer Science"  → field "in Computer Science"
  "B.S. - in Computer Science"  → field "Computer Science"
  ```

  Contrived input and low impact — no fixture produces it — but adding `·•` to that class removes the ordering asymmetry for free and makes the two strips agree on what a separator is. Your call whether it belongs in this PR or #653.

## Description

Accurate, with one overclaim that the commit above resolves: the test plan credits "compound internal field preservation," which was the mis-named ampersand test. Everything else round-trips — the `cleanField` trace, the truth-sidecar edit, and the four gates you listed all check out.

## Two mechanical notes for the maintainer

1. **My push re-arms the workflow-approval gate.** This repo's Actions `fork-pr-contributor-approval` policy is `first_time_contributors`, so the run on `0e706a9` is parked at `action_required` and needs one more **Approve and run** click. The new commit touches no workflow file, no `package.json`, and no dependency — only `src/lib/heuristics/**` — so there is nothing in it that executes at CI time beyond the tests themselves.
2. **The branch now carries two commits.** `main` merges through a merge queue that derives the squash message from the branch, so collapse to one before it enters the queue (or edit the squash message at merge time).

Head reviewed: `f4cc31e3c6c63aa0e2037a69bef2093cbb4c265f`, plus the fix commit `0e706a9`.

Verdict rule applied: **0 Blocking → APPROVE.** The one remaining Nit is non-blocking by definition and does not gate the merge.

---

s-annam:
@myukitty — following up on the one Nit from my review that I left in your hands rather than fixing: I've filed it as **#839** so it isn't lost when this merges.

To be clear about why it is a separate issue and not a change request on this PR: it is **not a defect in your fix**. Your change widened the glyph class at `cleanField`, which is exactly what #835 asked for and exactly where the reported bug lived. #839 is the *other* strip — `parseDegreeAndField`'s leading-separator cut at `education.ts:538` — which still carries the narrow `[-–—,:]` class and, because it runs first, blocks the `in`/`of` connective strip before your widening ever gets a turn:

```
"B.S. · in Computer Science"  → field "in Computer Science"   ← connective leaks
"B.S. - in Computer Science"  → field "Computer Science"      ← correct
```

I verified that against your branch directly rather than reasoning about it. It is low impact — no fixture in the 58-PDF corpus draws that shape — which is precisely why it belongs in the backlog rather than in the critical path of a PR that is already correct and green.

I did not fold it into the `0e706a9` commit on purpose: that commit is bounded to changes that **cannot alter behaviour**, and this one changes parser output. Anything behavioural stays the author's call, which in this case means yours or whoever picks up #839.

It is labelled `good first issue` and it is a natural sequel to this work — you now know these two strips better than anyone. **Yours if you want it; no obligation at all**, and no expectation that a contributor who fixes one bug inherits its neighbours. If you'd rather not, leave it and someone else will take it. Either way, nothing about #839 gates this PR: it is approved and ready once the workflow run on `0e706a9` is approved by a maintainer.

Issue #833, opened 2026-08-14
[parser] skills — a "Language: Fluent in Spanish" proficiency row is admitted as a skill token


s-annam:
## Context — what a `knownWrong` entry is

Each corpus fixture under `tests/fixtures/pdfs/` may carry a hand-authored ground-truth sidecar (`*.truth.json`, minted under #654). Its `knownWrong` block records, per field, a place where the parser disagrees with what the page actually draws, with a `status` of `open` (live bug, issue must be open), `accepted` (written-down tradeoff), or `unfiled` (measured, never filed — `issue: null`).

`unfiled` is capped at `UNFILED_TRUTH_CEILING = 10` (`src/lib/heuristics/corpus.test.ts:186`) and is saturated at 10/10 today. This issue files half of one such entry — see the note about the shared field at the bottom.

## What's wrong

A spoken-language **proficiency statement** is admitted into `skills` as if it were a skill token.

`tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-role-first-experience.pdf` draws:

```
Skills
Programming Languages: C, HTML, CSS, Java, VHDL, Assembly
Applications: Eclipse, Xilinx, ModelSim, PSpice, Microsoft Word/Excel/PowerPoint
Language: Fluent in Spanish
```

Observed: `Fluent in Spanish` appears in the `skills` array and as a `Language` category.

That fixture's truth sidecar excludes it deliberately, and says why in its `provenance`: *"the 'Language: Fluent in Spanish' row is a language proficiency, not a skill token, and is deliberately excluded."*

This is a **precision** defect, and precision is where the skills list is load-bearing beyond display: `skills` feeds the JD-match keyword surface (`src/lib/jd-match/`) and the job-search query builder's resume-derived keyword string (`src/lib/job-search/providers/keywords.ts`). A sentence-shaped pseudo-skill pollutes both.

## Root cause

The Skills-section extractor has exactly one label denylist, `src/lib/heuristics/extract/skills.ts:279-280`:

```ts
const NON_SKILL_SUBLABEL_RE =
  /^(?:personal\s+|other\s+)?(?:interests?|hobbies|hobby|activities|pastimes)\s*$/i;
```

consulted by `matchCellLabel` (`skills.ts:610-614`), which returns `undefined` for a denied label so `tokenizeCell` drops the whole cell.

[...]

Issue #221, opened 2026-06-26
Parser: Interests sub-label under Technical Skills bleeds into skills array


s-annam:
## Summary

When a "Technical Skills" section is internally sub-labeled (`Languages:`, `Technologies:`, `Interests:`), **every** sub-label's contents are flattened into `skills` — including non-skill sublists like **Interests**.

## Reproduction

Source:

```
### Technical Skills
Languages: Python, Go, C++, Java, JavaScript
Technologies: Linux, AWS, Docker, iOS
Interests: Science Fiction Novels, Weightlifting, Gardening, Eating, Tennis, Basketball
```

Parsed `skills` includes:

```
"Science Fiction Novels", "Weightlifting", "Gardening", "Eating", "Tennis", "Basketball", ...
```

Hobbies are scored and displayed as professional skills.

## Root cause

`src/lib/heuristics/extract/skills.ts` consumes the whole Skills section body and does not distinguish sub-labels. There's no allowlist/denylist for sub-label headings, so `Interests:` / `Hobbies:` (and arguably free-text `Activities:`) flow through identically to `Languages:`/`Technologies:`.

## Expected

- Recognize the `Label: a, b, c` sub-label shape inside a Skills section.
- Exclude non-skill sub-labels (`Interests`, `Hobbies`) from `skills`. Either drop them or route to a separate, unscored bucket.
- Keep skill sub-labels (`Languages`, `Technologies`, `Tools`, `Frameworks`, ...).

## Notes

Reproduced with a real PII resume. Add a synthetic fixture with a `Technical Skills` block containing an `Interests:` sub-label, per the fixture PII policy. Distinct from the skills-tokenizer over-split issue (#220).

PR #840, opened 2026-08-15
fix(skills): allow single-letter languages C, R, and D (#832)

### LINKED ISSUE #832: [parser] skills — single-letter language tokens (C, R, D) are dropped by isSkillToken's 2-character floor

`unfiled` is capped at `UNFILED_TRUTH_CEILING = 10` (`src/lib/heuristics/corpus.test.ts:186`) and is **saturated at 10/10** today, so no newly measured disagreement can be recorded until some are filed. This issue files one of them — and it is the one measured **twice, on two unrelated fixtures**, so it is not fixture-specific.

## What's wrong

A single-letter programming language is silently dropped from the skills list.

```
Skills
```

Observed: `skills` comes back without `C`. `HTML`, `CSS`, `Java`, `VHDL`, `Assembly` all survive.

`tests/fixtures/pdfs/latex/multi-degree-coursework.pdf` reproduces it independently on its own Languages row: `C` is dropped while `C++` survives.

Both are recorded today as `status: "unfiled"` `knownWrong` entries on the `skills` field.

C, R, and D are all real, widely-listed languages. On a résumé whose first listed language is `C`, the user's most prominent skill is the one the parser deletes — and it is deleted silently, so nothing in the UI tells them.

## Root cause

`src/lib/heuristics/extract/skills.ts:141-142`:

```ts
function isSkillToken(tok: string): boolean {
  if (tok.length < 2 || tok.length > 40) return false;
```

The 2-character floor is deliberate — it is a cheap noise guard against stray glyphs left by column splitting and bullet stripping. But it is *unconditional*, so it cannot distinguish a stray `x` from the language `C`.

The floor is the only thing rejecting `C`: the token reaches `isSkillToken` intact (the row is a `Label:` cell, `matchCellLabel` at `skills.ts:610` captures `Programming Languages` as the category, and `tokenizeCell` splits the body on commas normally). Everything downstream of the floor would accept it.

## Implementation plan

1. **`src/lib/heuristics/extract/skills.ts` — add a single-letter allowlist consulted before the floor.** Keep the floor; carve out the known one-character language names, as a named module constant beside the other vocabulary constants in this file:

   ```ts
   /** One-character tokens that are real, commonly-listed languages. The
    *  `tok.length < 2` floor in `isSkillToken` is a noise guard against stray
    *  glyphs left by column splitting; these are the only single characters that
    *  are not noise, so they are allowlisted rather than lowering the floor. */
   const SINGLE_LETTER_SKILLS = new Set(["c", "r", "d"]);
   ```

   ```ts
   function isSkillToken(tok: string): boolean {
     if (tok.length === 1 && SINGLE_LETTER_SKILLS.has(tok.toLowerCase())) return true;
     if (tok.length < 2 || tok.length > 40) return false;
     …
   ```

   Return `true` early rather than falling through, so the allowlisted token also skips the numeric / URL / date-range / word-count checks below it — none of which a bare letter can trip, but an early return states the intent.

   **Scope the set deliberately.** `C`, `R`, `D` are the defensible three. Do NOT add `J`, `K`, `Q`, or `F` — each is a real language name, but each is also a plausible stray glyph, and admitting them trades a rare recall win for a common precision loss. If someone wants them later, that is a separate argument with its own evidence.

2. **Consider case.** The set is matched case-insensitively so a lowercase `c` in a lowercase-cased row survives, matching how the rest of this file treats vocabulary (`PROFILE_LABEL_RE`, `NON_SKILL_SUBLABEL_RE` are all `/i`).

3. **Flip the two ground-truth entries** to `status: "open"` with this issue's number:
   - `tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-role-first-experience.truth.json` → `skills`
   - `tests/fixtures/pdfs/latex/multi-degree-coursework.truth.json` → `skills`

   ⚠️ The first of those two entries currently records **two independent disagreements on one field** — the `C` drop AND a `Language: Fluent in Spanish` proficiency row being admitted as a skill (filed separately). The `knownWrong` schema is keyed per FIELD, so it can carry only one `issue:` number. Cite this issue there and name the other one in the `note` — do not drop the second disagreement from the note.

   If the fix lands in the same PR, **delete** the entries rather than flipping them, and change no truth VALUES: a truth file records what the page draws and is never edited to match the parser (see each file's `provenance` string).

4. **Lower `UNFILED_TRUTH_CEILING`** in `src/lib/heuristics/corpus.test.ts:186` by 2 (10 → 8). Its docblock says: "File the issue and flip the entry to `open`; then lower this." Sibling issues from the same audit also lower it — on a rebase conflict, take the LOWER number.

## Acceptance criteria

- [ ] `isSkillToken("C")`, `isSkillToken("R")`, `isSkillToken("D")` return `true`; `isSkillToken("x")`, `isSkillToken("J")`, `isSkillToken("•")`, `isSkillToken("(")` return `false`. Unit tests in `src/lib/heuristics/extract/skills.test.ts` (or the existing skills test module).
- [ ] `google-docs-skia-proxy-role-first-experience.pdf` parses `skills` containing `C`, and its `Programming Languages` category holds all six drawn tokens.
- [ ] `multi-degree-coursework.pdf` parses `skills` containing both `C` and `C++`.
- [ ] No fixture gains a spurious single-character skill: `npx vitest run src/lib/heuristics/corpus.test.ts` shows `skills` precision unchanged or improved on all 58 fixtures.
- [ ] `npx vitest run src/lib/heuristics/corpus-roundtrip.test.ts` passes with **no new `KNOWN_FAILURES` baseline rows** — a single-character skill must survive the parse → export → re-parse leg too (it is middot-joined into the skills header line by `src/lib/pdf/ats-resume-model.ts`, so confirm the re-parse does not re-drop it).
- [ ] `npm run check:baselines` reports 2 fewer unfiled entries.

shubhransh-gupta:
Fixes #832

## Summary
- Allowlist C, R, and D in `isSkillToken` without lowering the global 2-char noise floor
- Add unit tests for allowlisted languages and rejected noise tokens

## Test plan
- [ ] `npx vitest run src/lib/heuristics/extract/skills.test.ts`

### DIFF
--- a/src/lib/heuristics/extract/skills.test.ts
+++ b/src/lib/heuristics/extract/skills.test.ts
@@ -402,8 +402,8 @@ describe("tokenizeSkillLine — issue #221 non-skill sub-labels", () => {
   });
 
   it("keeps skill sub-labels (Languages/Technologies/Tools/Frameworks)", () => {
-    expect(tokenizeSkillLine("Languages: Python, Go, C++, Java")).toEqual(
-      expect.arrayContaining(["Python", "Go", "C++", "Java"]),
+    expect(tokenizeSkillLine("Languages: Python, Go, C++, Java, C")).toEqual(
+      expect.arrayContaining(["Python", "Go", "C++", "Java", "C"]),
     );
     expect(tokenizeSkillLine("Technologies: Linux, AWS, Docker, iOS")).toEqual(
       expect.arrayContaining(["Linux", "AWS", "Docker", "iOS"]),
@@ -423,6 +423,23 @@ describe("tokenizeSkillLine — issue #221 non-skill sub-labels", () => {
     expect(result).toEqual(
       expect.arrayContaining(["Interest Rate Modeling", "Risk Analysis"]),
     );
+  });
+});
+
+describe("tokenizeSkillLine — issue #832 single-letter languages", () => {
+  it("keeps C, R, and D as real programming languages", () => {
+    expect(tokenizeSkillLine("Languages: C, R, D")).toEqual(
+      expect.arrayContaining(["C", "R", "D"]),
+    );
+  });
+
+  it("still rejects other single-character noise tokens", () => {
+    expect(tokenizeSkillLine("Skills: x, J, •, (")).toEqual([]);
+  });
+
+  it("allows allowlisted single letters alongside real skills on the same line", () => {
+    // Positive control: distinguishes rejecting noise from rejecting everything.
+    expect(tokenizeSkillLine("Skills: Python, x, J, C")).toEqual(["Python", "C"]);
   });
 });
 
--- a/src/lib/heuristics/extract/skills.ts
+++ b/src/lib/heuristics/extract/skills.ts
@@ -138,7 +138,14 @@ function looksLikeContactLink(tok: string): boolean {
   return PROFILE_LABEL_RE.test(t) || PROFILE_HOST_RE.test(t) || URLISH_RE.test(t);
 }
 
+/** One-character tokens that are real, commonly-listed languages. The
+ *  `tok.length < 2` floor in `isSkillToken` is a noise guard against stray
+ *  glyphs left by column splitting; these are the only single characters that
+ *  are not noise, so they are allowlisted rather than lowering the floor. */
+const SINGLE_LETTER_SKILLS = new Set(["c", "r", "d"]);
+
 function isSkillToken(tok: string): boolean {
+  if (tok.length === 1 && SINGLE_LETTER_SKILLS.has(tok.toLowerCase())) return true;
   if (tok.length < 2 || tok.length > 40) return false;
   if (/^\d+$/.test(tok)) return false;
   // A professional-profile link (or its bare "GitHub" / "LinkedIn" heading) is
--- a/src/lib/heuristics/corpus.test.ts
+++ b/src/lib/heuristics/corpus.test.ts
@@ -183,7 +183,7 @@ const TRUTH_ANNOTATED_FIELD_FLOOR = 150;
  * `npm run check:baselines` on every run, and bounded here — undescribed debt may
  * not GROW. File the issue and flip the entry to `open`; then lower this.
  */
-const UNFILED_TRUTH_CEILING = 9;
+const UNFILED_TRUTH_CEILING = 7;
 
 /** Generator category = the fixture root's immediate subdirectory. */
 function categoryOf(repoRelPdfPath: string): string {
--- a/tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-role-first-experience.expected.json
+++ b/tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-role-first-experience.expected.json
@@ -22,7 +22,7 @@
       "phoneIsValid",
       "skills"
     ],
-    "skillsCount": 11,
+    "skillsCount": 12,
     "experienceCount": 2,
     "educationCount": 1,
     "projectsCount": 0,
@@ -74,7 +74,7 @@
     "sectionSource": "regex",
     "pageCount": 1,
     "rawCharCount": 1428,
-    "extractedCharCount": 1146,
+    "extractedCharCount": 1147,
     "sections": [
       {
         "name": "profile",
@@ -101,7 +101,7 @@
       "hasSummary": false,
       "experienceCount": 2,
       "educationCount": 1,
-      "skillsCount": 11
+      "skillsCount": 12
     },
     "linkAnnotationCount": 0,
     "disagreements": []
--- a/tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-role-first-experience.truth.json
+++ b/tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-role-first-experience.truth.json
@@ -44,9 +44,9 @@
       "note": "Role 2's employer line reads “Multicultural Engineering Program – State Polytechnic University”; `company` comes back as just the university — the program half is not lost, the parser puts it on `team`, but `experience.company` scores the `company` field alone. The identical shape is measured on unknown/single-column-title-below-anchor. Possibly a defensible org/team split rather than a defect — recorded rather than assumed, because ground truth's job is to state what the page says and let a human adjudicate."
     },
     "skills": {
-      "issue": null,
-      "status": "unfiled",
-      "note": "Two independent disagreements on one field: the single-letter token “C” is DROPPED from the Programming Languages row (the identical drop is measured on latex/multi-degree-coursework, so it is not fixture-specific), and “Fluent in Spanish” is admitted as a skill from the “Language:” row."
+      "issue": 833,
+      "status": "open",
+      "note": "“Fluent in Spanish” is admitted as a skill from the “Language:” row; the Programming Languages row is now correct after #832."
     }
   }
 }
--- a/tests/fixtures/pdfs/latex/deedy-resume-macfonts.expected.json
+++ b/tests/fixtures/pdfs/latex/deedy-resume-macfonts.expected.json
@@ -27,7 +27,7 @@
       "skills",
       "website_url"
     ],
-    "skillsCount": 22,
+    "skillsCount": 23,
     "experienceCount": 6,
     "educationCount": 3,
     "projectsCount": 0,
@@ -83,7 +83,7 @@
     "sectionSource": "markdown",
     "pageCount": 1,
     "rawCharCount": 3205,
-    "extractedCharCount": 2156,
+    "extractedCharCount": 2157,
     "sections": [
       {
         "name": "profile",
@@ -114,7 +114,7 @@
       "hasSummary": false,
       "experienceCount": 6,
       "educationCount": 3,
-      "skillsCount": 22
+      "skillsCount": 23
     },
     "linkAnnotationCount": 9,
     "disagreements": []
--- a/tests/fixtures/pdfs/latex/deedy-resume-openfonts.expected.json
+++ b/tests/fixtures/pdfs/latex/deedy-resume-openfonts.expected.json
@@ -27,7 +27,7 @@
       "skills",
       "website_url"
     ],
-    "skillsCount": 22,
+    "skillsCount": 23,
     "experienceCount": 6,
     "educationCount": 3,
     "projectsCount": 0,
@@ -83,7 +83,7 @@
     "sectionSource": "markdown",
     "pageCount": 1,
     "rawCharCount": 3207,
-    "extractedCharCount": 2158,
+    "extractedCharCount": 2159,
     "sections": [
       {
         "name": "profile",
@@ -114,7 +114,7 @@
       "hasSummary": false,
       "experienceCount": 6,
       "educationCount": 3,
-      "skillsCount": 22
+      "skillsCount": 23
     },
     "linkAnnotationCount": 9,
     "disagreements": []
--- a/tests/fixtures/pdfs/latex/multi-degree-coursework.expected.json
+++ b/tests/fixtures/pdfs/latex/multi-degree-coursework.expected.json
@@ -25,7 +25,7 @@
       "website_url",
       "work_authorization"
     ],
-    "skillsCount": 17,
+    "skillsCount": 18,
     "experienceCount": 4,
     "educationCount": 2,
     "projectsCount": 3,
@@ -77,7 +77,7 @@
     "sectionSource": "markdown",
     "pageCount": 1,
     "rawCharCount": 3428,
-    "extractedCharCount": 2625,
+    "extractedCharCount": 2626,
     "sections": [
       {
         "name": "profile",
@@ -108,7 +108,7 @@
       "hasSummary": false,
       "experienceCount": 4,
       "educationCount": 2,
-      "skillsCount": 17
+      "skillsCount": 18
     },
     "linkAnnotationCount": 6,
     "disagreements": []
--- a/tests/fixtures/pdfs/latex/multi-degree-coursework.truth.json
+++ b/tests/fixtures/pdfs/latex/multi-degree-coursework.truth.json
@@ -57,12 +57,5 @@
     "Docker",
     "Raspberry Pi",
     "iOS"
-  ],
-  "knownWrong": {
-    "skills": {
-      "issue": null,
-      "status": "unfiled",
-      "note": "The single-letter token “C” is DROPPED from the Languages row while “C++” survives. Second independent measurement of the same drop (see google-docs/google-docs-skia-proxy-role-first-experience)."
-    }
-  }
+  ]
 }
--- a/tests/fixtures/pdfs/unknown/pdflib-leading-glyph-skills-header.expected.json
+++ b/tests/fixtures/pdfs/unknown/pdflib-leading-glyph-skills-header.expected.json
@@ -22,7 +22,7 @@
       "phoneIsValid",
       "skills"
     ],
-    "skillsCount": 8,
+    "skillsCount": 9,
     "experienceCount": 1,
     "educationCount": 1,
     "projectsCount": 0,
@@ -74,7 +74,7 @@
     "sectionSource": "markdown",
     "pageCount": 1,
     "rawCharCount": 462,
-    "extractedCharCount": 289,
+    "extractedCharCount": 290,
     "sections": [
       {
         "name": "profile",
@@ -101,7 +101,7 @@
       "hasSummary": false,
       "experienceCount": 1,
       "educationCount": 1,
-      "skillsCount": 8
+      "skillsCount": 9
     },
     "linkAnnotationCount": 0,
     "disagreements": []


shubhransh-gupta:
Hi — ping for review when convenient 👋

Fixes #832: allows single-letter skill tokens `C`, `R`, `D` in `isSkillToken` with unit tests in `skills.test.ts`. Local `vitest` passed on the branch.

Happy to tweak if you'd prefer a different allowlist approach. Thanks!

src/lib/heuristics/extract/skills.ts
```diff
@@ -138,7 +138,14 @@ function looksLikeContactLink(tok: string): boolean {
...
 }
 
+/** One-character tokens that are real, commonly-listed languages. The
+ *  `tok.length < 2` floor in `isSkillToken` is a noise guard against stray
+ *  glyphs left by column splitting; these are the only single characters that
+ *  are not noise, so they are allowlisted rather than lowering the floor. */
+const SINGLE_LETTER_SKILLS = new Set(["c", "r", "d"]);
+
 function isSkillToken(tok: string): boolean {
+  if (tok.length === 1 && SINGLE_LETTER_SKILLS.has(tok.toLowerCase())) return true;
```
s-annam, line 148: This line is correct and matches #832's plan exactly — including the early return, which is what lets an allowlisted token skip the numeric / URL / date-range / word-count checks below.

What it also does, though, is move four baked corpus snapshots and clear one `knownWrong` exemption, and those files have to move in the same commit or CI is red. See Blocking 1 and 2 in the review body — `npm run bake-fixtures`, delete the `multi-degree-coursework` skills exemption, flip the skia-proxy one to `open`/#833, and lower `UNFILED_TRUTH_CEILING` 9 → 7.

src/lib/heuristics/extract/skills.test.ts
```diff
@@ -426,6 +426,18 @@ describe("tokenizeSkillLine — issue #221 non-skill sub-labels", () => {
...
 
+describe("tokenizeSkillLine — issue #832 single-letter languages", () => {
+  it("keeps C, R, and D as real programming languages", () => {
+    expect(tokenizeSkillLine("Languages: C, R, D")).toEqual(
+      expect.arrayContaining(["C", "R", "D"]),
+    );
+  });
+
+  it("still rejects other single-character noise tokens", () => {
+    expect(tokenizeSkillLine("Skills: x, J, •, (")).toEqual([]);
```
s-annam, line 437: Nit (non-blocking): every token on this line is expected to be rejected, so `toEqual([])` also passes in the world where the `Skills:` sub-label path drops the whole cell — it cannot tell "rejected the noise" from "rejected everything".

A positive control in the same line pins it:

```ts
expect(tokenizeSkillLine("Skills: Python, x, J, C")).toEqual(["Python", "C"]);
```

I ran that on this branch — it returns exactly `["Python", "C"]`, so it asserts the allowlist admits `C`, rejects `x` and `J`, and that the surrounding cell path is live, all at once. Keep the `•`/`(` case too if you like; just don't let it be the only assertion.

s-annam (changes requested):
The parser change itself is exactly right — it matches #832's implementation plan verbatim, the allowlist is scoped to the defensible three, the early return is placed correctly, and I verified on the fixtures that every newly-admitted token is genuinely drawn on the page (no spurious single characters anywhere in the corpus). What's missing is the bookkeeping the fix obliges: the corpus gate is red on this branch, so CI cannot pass as-is.

Reviewed at `fd8b734c12eb20a61aa7ca936cc585bc52acddae`.

## Blocking

**1. `npx vitest run src/lib/heuristics/corpus.test.ts` fails — 5 fixtures. CI (`verify`) will be red.**

Four are stale baked snapshots the fix invalidates, one is the ratchet's second tooth:

```
FAIL google-docs/google-docs-skia-proxy-role-first-experience.pdf   skillsCount 11 → 12
FAIL latex/deedy-resume-macfonts.pdf                                skillsCount 22 → 23
FAIL latex/deedy-resume-openfonts.pdf                               skillsCount 22 → 23
FAIL unknown/pdflib-leading-glyph-skills-header.pdf                 skillsCount  8 →  9
FAIL latex/multi-degree-coursework.pdf
  AssertionError: 'skills' now matches ground truth (expected 18, parsed 18, matched 18)
    — remove its knownWrong entry (cites #null)
```

I checked each new token against the drawn page — all four are correct gains, not noise (`C` on the two deedy fixtures and on skia-proxy, `R` on `pdflib-leading-glyph-skills-header`, whose Skills block genuinely reads `Proficient in: Python, SQL, R`). So this is a re-bake, not a regression:

```bash
npm run bake-fixtures     # or: UPDATE_FIXTURES=1 npx vitest run src/lib/heuristics/corpus.test.ts
```

Re-baking gives exactly `skillsCount` +1 and `extractedCharCount` +1 on those four `.expected.json` files and nothing else — please confirm that's all your diff shows before committing it.

**2. `Fixes #832` while steps 3 and 4 of the issue's plan, and two of its acceptance criteria, are unimplemented.**

The merge closes #832, so the ground-truth bookkeeping would ship as "done" and never be reopened. `npm run check:baselines` on this branch still reports **9 unfiled** — the AC asks for 2 fewer (7).

- `tests/fixtures/pdfs/latex/multi-degree-coursework.truth.json` → `skills`: **delete** the `knownWrong` entry. It records only the `C` drop, which this PR fixes; leaving it is what fails fixture 5 above. (Per the issue: when the fix lands in the same PR, delete rather than flip — and change no truth *values*.)
- `tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-role-first-experience.truth.json` → `skills`: this entry must **stay**, because the field still disagrees for a second, unrelated reason — `Fluent in Spanish` is still admitted as a skill token (#833, open). Flip it to `"status": "open"`, `"issue": 833`, and rewrite the `note` so it describes only that remaining disagreement; the `C`-drop half is no longer true after this PR. That takes it out of the unfiled count too.
- `src/lib/heuristics/corpus.test.ts` → `UNFILED_TRUTH_CEILING`: **9 → 7**. (The issue says "lower by 2 (10 → 8)", but #838 already landed 10 → 9 on `main`; the docblock's rule is that the ceiling tracks the actual count, so 7 is the number here. On a rebase conflict, take the lower.)

After those three edits, `corpus.test.ts` and `check:baselines` should both be green — please paste the `check:baselines` tail showing 7 unfiled.

## Secondary

**3. Test plan doesn't cover the gates this change actually moves.** The one listed box (`skills.test.ts`) is unticked, and a skills-tokenizer change necessarily crosses the baked corpus, the truth ratchet, and the baselines gate — which is precisely how the five failures above got past the PR. Please run and tick `npm run verify` before the next push.

## Nits (non-blocking)

**4.** The negative test asserts `toEqual([])` on a line where *every* token is expected to be rejected, so it would still pass if the `Skills:` label path ever started dropping the whole cell — the assertion can't distinguish "rejected the noise" from "rejected everything". A positive control inside the same line pins it: `expect(tokenizeSkillLine("Skills: Python, x, J, C")).toEqual(["Python", "C"])` (verified — that is what it returns on this branch). Inline comment below.

**5.** The commit subject (`Allow single-letter language skills C, R, and D in parser.`) isn't in the Conventional Commits shape the PR title already uses (`fix(skills): …`). Low impact — the merge queue derives the squash message from the PR title — but worth matching when you amend for the fixes above. Keep it to one commit (amend, don't stack).

## Gates run

| Gate | Result |
|---|---|
| `npm run typecheck` | pass |
| `npm run lint` | pass |
| `npx vitest run src/lib/heuristics/extract/skills.test.ts` | pass (47) |
| `npx vitest run src/lib/heuristics/corpus.test.ts` | **FAIL (5)** — Blocking 1 |
| `npx vitest run src/lib/heuristics/corpus-roundtrip.test.ts` | pass (61) — the single-char skill survives parse → export → re-parse, no new `KNOWN_FAILURES` rows |
| `npm run check:baselines` | 9 unfiled (unchanged; AC wants 7) — Blocking 2 |
| `npx fallow audit --base origin/main` | no issues in the 2 changed files |
| Fixture PII (3a) | skipped — no fixture binaries or truth sidecars touched |
| Design system / tokens (3b, 3c) | skipped — no UI code touched |
| Script/skill command review (3e) | skipped — no scripts touched |

Verdict rule: ≥1 Blocking → `REQUEST_CHANGES`. No auto-fix was applied and nothing was pushed to your branch, per that rule. Everything above is mechanical — the hard part (the parser change) is already done and correct.

---

s-annam:
Welcome, @shubhransh-gupta — this is your first PR here, so a few extra words on what the review above means and how we work. The short version: **the change itself is correct**, and `REQUEST_CHANGES` here is bookkeeping, not a verdict on your work.

### On the change

The parser fix is right, and it matches the implementation plan in #832. Each newly-admitted single-letter token was checked against what is actually drawn on the page — no spurious single characters slipped in. That was the risky part of #832 and you got it right.

### What's left, and why it blocks

Both items are mechanical:

1. **Corpus snapshots are stale.** `src/lib/heuristics/corpus.test.ts` diffs the full cascade against a co-located `*.expected.json` per fixture, so admitting `C`/`R`/`D` legitimately shifts `skillsCount` on four of them — and one more fixture trips the `knownWrong` "now matches ground truth" check, which is the suite telling you a known failure got *fixed*. Re-bake with `npm run bake-fixtures` and commit the updated snapshots. Details in [`tests/fixtures/pdfs/README.md`](https://github.com/offlinecv/OfflineCV/blob/main/tests/fixtures/pdfs/README.md).
2. **The PR says `Fixes #832` but steps 3–4 of that plan aren't in yet** — the truth-sidecar entries and the `UNFILED_TRUTH_CEILING` drop. `npm run check:baselines` still reports the old count, so the issue's acceptance criteria aren't met by this diff. Either finish those steps, or downgrade the link to `Refs #832` and we land the parser half on its own. Your call — both are fine, and inline comments have the specifics.

These would have turned CI red on merge, which is the only reason they're marked blocking rather than left as review notes.

### How we handle small stuff — our nit policy

Worth knowing up front, since it'll apply to your next PR: **when a review turns up zero blocking findings, we don't bounce the PR back for nits.** A maintainer fixes the small items directly, pushes a commit to your branch, and approves in the same pass — including that fix commit. It saves a round-trip on things like a missing SPDX header or a reworded comment.

That did **not** happen here: the presence of blocking findings means we leave your branch untouched, so nothing was pushed to it and the code you see is still exactly the code you wrote. If you'd rather we never push to your branches at all, say so and we'll respect it.

One mechanical note: this PR comes from your fork, so maintainer pushes only work if **"Allow edits by maintainers"** is checked on the PR. If you want the nit-fix path available on future PRs, leave that on.

### Pointers

[`CONTRIBUTING.md`](https://github.com/offlinecv/OfflineCV/blob/main/CONTRIBUTING.md) covers the branch/commit/test workflow, and [`CLAUDE.md`](https://github.com/offlinecv/OfflineCV/blob/main/CLAUDE.md) covers the pipeline shape and tier layout if you want context on where the skills parser sits in the cascade.

Anything unclear — the process, the fixture corpus, why a gate is shaped the way it is, or if you disagree with a finding — just tag me here or open a thread. Pushback on a review is welcome; I'd rather discuss it than have you silently absorb a call you think is wrong. Thanks again for picking this up.

shubhransh-gupta:
Thanks for the thorough review, @s-annam — all blocking items addressed in one amended commit (rebased onto latest `main`):

**Blocking 1 — corpus re-bake**
- Ran `npm run bake-fixtures`; four fixtures updated (`skillsCount` +1, `extractedCharCount` +1 only): deedy-resume-macfonts, deedy-resume-openfonts, google-docs-skia-proxy-role-first-experience, pdflib-leading-glyph-skills-header.

**Blocking 2 — ground-truth bookkeeping**
- `latex/multi-degree-coursework.truth.json` → deleted the `skills` `knownWrong` entry (C drop fixed here).
- `google-docs/google-docs-skia-proxy-role-first-experience.truth.json` → flipped `skills` to `"status": "open"`, `"issue": 833` with note covering only the remaining Fluent-in-Spanish disagreement.
- `UNFILED_TRUTH_CEILING` → **7** (was 9).

**Secondary + nit**
- `npm run verify` green locally.
- Added positive control: `tokenizeSkillLine("Skills: Python, x, J, C")` → `["Python", "C"]`.
- Commit subject aligned to `fix(skills): …`.

`check:baselines` tail: **7 unfiled**, 1 entry charged to live issue (#833). Ready for another look when you have time.

shubhransh-gupta:
CI fix pushed — `pdflib-leading-glyph-skills-header.expected.json` had a stale `extractedCharCount` from the rebase conflict (292 vs correct 290). Re-baked and amended; corpus test passes locally. Should see `verify` green on the next run.

shubhransh-gupta:
@s-annam — friendly re-review ping 👋

All blocking items from the earlier review are addressed and **CI is green** on the latest push (`verify` + `fallow` both succeeded):

- Corpus re-baked (four fixtures + `pdflib-leading-glyph-skills-header` char count fix after rebase)
- Truth sidecars updated; `UNFILED_TRUTH_CEILING` → 7
- Positive control test added (`tokenizeSkillLine("Skills: Python, x, J, C")` → `["Python", "C"]`)

Ready for another look whenever you have a moment. Thanks!

src/lib/heuristics/extract/skills.test.ts
```diff
@@ -426,6 +426,23 @@ describe("tokenizeSkillLine — issue #221 non-skill sub-labels", () => {
   });
 });
 
+describe("tokenizeSkillLine — issue #832 single-letter languages", () => {
+  it("keeps C, R, and D as real programming languages", () => {
+    expect(tokenizeSkillLine("Languages: C, R, D")).toEqual(
+      expect.arrayContaining(["C", "R", "D"]),
+    );
```
s-annam, line 433: **Nit (non-blocking).** `expect.arrayContaining` passes even if the tokenizer returns extra tokens alongside these three, so it under-specifies what the fix guarantees. The exact return on this branch is `["C", "R", "D"]` — I ran it — so a strict `toEqual` costs nothing and pins the behaviour precisely.

```suggestion
    expect(tokenizeSkillLine("Languages: C, R, D")).toEqual(["C", "R", "D"]);
```

src/lib/heuristics/extract/skills.test.ts
```diff
@@ -402,8 +402,8 @@ describe("tokenizeSkillLine — issue #221 non-skill sub-labels", () => {
   });
 
   it("keeps skill sub-labels (Languages/Technologies/Tools/Frameworks)", () => {
-    expect(tokenizeSkillLine("Languages: Python, Go, C++, Java")).toEqual(
-      expect.arrayContaining(["Python", "Go", "C++", "Java"]),
+    expect(tokenizeSkillLine("Languages: Python, Go, C++, Java, C")).toEqual(
+      expect.arrayContaining(["Python", "Go", "C++", "Java", "C"]),
```
s-annam, line 406: **Nit (non-blocking).** This is the `#221` sub-label test, widened to also carry a `#832` case. Now that `#832` has its own `describe` block immediately below, reverting this line keeps each test attributable to the issue it guards — useful later when someone bisects one of the two and wants to know which assertions are load-bearing for which fix.

Both forms pass (`["Python", "Go", "C++", "Java", "C"]` is the actual return either way), so this is purely about attribution.

```suggestion
    expect(tokenizeSkillLine("Languages: Python, Go, C++, Java")).toEqual(
      expect.arrayContaining(["Python", "Go", "C++", "Java"]),
```

s-annam (changes requested):
Every blocking item from the last round is genuinely fixed, and I verified the bookkeeping independently rather than taking the summary's word for it — all four re-baked snapshots gained a token that is really drawn on the page, and every gate I can run locally is green. One Blocking item remains, and it is **not** something you introduced this round: the commit carries a `Co-authored-by: Cursor` trailer, which this repo forbids and which would land in `main` permanently. It was already there at `fd8b734`, the sha I reviewed last time, and I missed it. Apologies for surfacing it only now — it is a one-command fix and it is the only thing between this PR and an approval.

Reviewed at `b2127526c1ea528fb12504f4a6672278beafda0e`.

## Last round's blockers — all clear

| Last round | Now |
|---|---|
| **Blocking 1** — `corpus.test.ts` red on 5 fixtures | **Fixed.** 65 tests pass. The four re-bakes are `skillsCount` +1 / `extractedCharCount` +1 and nothing else, exactly as asked. |
| **Blocking 2** — `Fixes #832` with steps 3–4 unimplemented | **Fixed.** `check:baselines` reports **7 unfiled**, 1 charged to a live issue. |
| **Secondary 3** — test plan didn't cover the gates | Partly — `verify` was run (thank you), but the PR body still doesn't say so. See Secondary 2. |
| **Nit 4** — negative test couldn't distinguish rejecting noise from rejecting everything | **Fixed.** Positive control added and it is exact. |
| **Nit 5** — commit subject not Conventional Commits | **Fixed.** `fix(skills): …`, one commit. |

I re-derived the four re-bakes from the drawn page rather than trusting the diff:

- `deedy-resume-macfonts` / `-openfonts` — both draw `C • C++ • CSS • PHP • Assembly`; the gain is `C`.
- `google-docs-skia-proxy-role-first-experience` — draws `Programming Languages: C, HTML, CSS, Java, VHDL, Assembly`; all six now parse.
- `pdflib-leading-glyph-skills-header` — draws `Proficient in: Python, SQL, R`; the gain is `R`.

No other fixture's snapshot moved, which is the proof that nothing spurious was admitted across the corpus.

The truth-sidecar work is right in both directions: `multi-degree-coursework` has its `knownWrong` **deleted** (not flipped) with no truth *values* touched, and `skia-proxy` is flipped to `"issue": 833, "status": "open"` with a note that describes only the remaining `Fluent in Spanish` disagreement. #833 is open, so the entry is legitimately charged.

### Acceptance criteria (#832)

All seven met. `isSkillToken` is exercised through `tokenizeSkillLine`, its only public path, which is the stronger test; `multi-degree` now parses both `C` and `C++`; roundtrip adds no `KNOWN_FAILURES` rows; `check:baselines` is down by exactly 2.

I also traced the early return for holes and found none: `isSkillToken` has a single call site (`tokenizeCell`), the token is already `.trim()`ed and stripped of trailing `.!?,;` before it arrives, and every check the early return skips (`/^\d+$/`, `looksLikeContactLink`, the date-range regex, the >6-word guard) is structurally unreachable for a one-character letter. The export path middot-joins skills with no minimum-length filter, so a lone `C` survives the round trip — which the roundtrip suite confirms empirically.

## Blocking

**1. The commit carries `Co-authored-by: Cursor <cursoragent@cursor.com>`, which would land in `main` verbatim and permanently.**

`CLAUDE.md` → *Hard rules (no exceptions)*: "**No AI attribution in git.** … Never commit attribution trailers (like `Co-Authored-By:` or `Claude-Session:`) or PR badges." It is one of only three rules kept in `CLAUDE.md` rather than delegated to a gate, precisely because breaking it is silent and permanent — `docs/CONTRIBUTING-PROCESS.md` records ~16 stray trailers that reached `main` before anyone noticed.

This is not cosmetic, and it is **not** something a maintainer can strip at merge time:

- `main` is governed by an active `main-merge-queue` ruleset with `merge_method: SQUASH`, so the queue composes the squash commit itself — nobody hand-edits the message on the way in.
- The repo's `squash_merge_commit_message` is `COMMIT_MESSAGES`, so your commit body becomes the squash body verbatim.
- GitHub parses `Co-authored-by:` as a real trailer, so `Cursor <cursoragent@cursor.com>` would be recorded as a co-author of the commit in `main` and in the contributor graph.

Fix — amend and force-push. This is your own fork branch, so a force-push here is safe and entirely yours to make:

```bash
git commit --amend        # delete the Co-authored-by line; keep the subject and body as they are
git push --force-with-lease origin fix/issue-832-single-letter-skills
```

Cursor adds that trailer automatically, so it is worth switching it off in Cursor's commit settings — otherwise it will come back on your next PR.

Nothing else about the commit needs to change: the subject is correct Conventional Commits, the body accurately describes the whole change, and the branch is a single commit rebased cleanly onto `main`.

## Secondary

**2. The PR description now covers about a third of its own diff.**

The `## Summary` still names only the allowlist and the unit tests — the two source-file bullets. Seven of the ten changed files are the bookkeeping half (four re-baked snapshots, two truth sidecars, `UNFILED_TRUTH_CEILING` 9 → 7) and go unmentioned, and the `## Test plan` still lists a single unticked `skills.test.ts` box even though you ran `npm run verify` and said so in a comment.

The description is what a bisecting maintainer reads a year from now when the code has stopped being fresh in anyone's mind, so an accurate one is worth the minute. Suggested replacement:

```markdown
## Summary
- Allowlist `C`, `R`, and `D` in `isSkillToken` without lowering the global 2-char noise floor
- Re-bake four corpus snapshots whose `skillsCount` legitimately gains the newly-admitted token
- Ground-truth bookkeeping for #832: delete `multi-degree-coursework`'s `skills` `knownWrong`
  entry (fixed here), flip `skia-proxy`'s to `#833`/`open` for the remaining
  "Fluent in Spanish" disagreement, and lower `UNFILED_TRUTH_CEILING` 9 → 7
- Add unit tests for the allowlisted languages, rejected noise tokens, and a mixed-line
  positive control

## Test plan
- [x] `npx vitest run src/lib/heuristics/extract/skills.test.ts`
- [x] `npx vitest run src/lib/heuristics/corpus.test.ts`
- [x] `npx vitest run src/lib/heuristics/corpus-roundtrip.test.ts`
- [x] `npm run check:baselines` — 7 unfiled
- [x] `npm run verify`
```

## Nits (non-blocking — ignore any you disagree with)

**3.** `expect.arrayContaining` on the `C, R, D` assertion is looser than it needs to be; the exact return is `["C", "R", "D"]`, which I verified on this branch. Suggestion inline.

**4.** The `#221` test was widened to carry a `#832` case. Now that #832 has its own `describe` block right below, reverting that line keeps each test attributable to the issue it guards. Suggestion inline.

**5.** No action needed, just so you know the shape of your own suite: the `"still rejects other single-character noise tokens"` test passes identically with the fix reverted — `x`, `J`, `•`, `(` are all rejected by the pre-existing 2-char floor. That is exactly what a negative control is *for*, so it earns its place; it is the positive control on the line below it that actually pins the new code.

## Gates run

| Gate | Result |
|---|---|
| `npm run typecheck` | pass |
| `npm run lint` | pass |
| `npx vitest run src/lib/heuristics/extract/skills.test.ts` | pass (48) |
| `npx vitest run src/lib/heuristics/corpus.test.ts` | pass (65, 1 skipped) — was 5 failures last round |
| `npx vitest run src/lib/heuristics/corpus-roundtrip.test.ts` | pass (63, 1 skipped) — no new `KNOWN_FAILURES` rows |
| `npm run check:baselines` | **7 unfiled**, 1 charged to live issue #833 — matches the AC |
| `npm run check:fixtures` | pass — 60 PDFs + 16 truth sidecars, all personas synthetic |
| `npx fallow audit --base origin/main` | pass — no issues in the 10 changed files (the 4 complexity findings are inherited and untouched by this diff) |
| CI `verify` + `fallow` | both green on `b212752` |
| Fixture PII (3a) | ran — no new binaries; the two edited truth sidecars are swept by `check:fixtures` |
| Design system / tokens (3b, 3c) | skipped — no UI code touched |
| Script / skill command review (3e) | skipped — no scripts or skills touched |

Verdict rule: ≥1 Blocking → `REQUEST_CHANGES`. Nothing was pushed to your branch and nothing was auto-fixed — with a blocking finding present we leave a contributor's branch alone, and on a fork we would not force-push it in any case. The two nits are inline as one-click suggestions if you want them.

Genuinely good work on the bookkeeping round — that was the fiddly, unglamorous half of #832 and you got every piece of it right.

---

s-annam:
Thanks for the fast turnaround, @shubhransh-gupta — and for taking on the unglamorous half of #832.

**Resolved:** every blocking item from the last round. I re-derived the bookkeeping from the fixtures rather than taking the summary's word for it — all four re-baked snapshots gained a token that is genuinely drawn on the page (`C` on both deedy fixtures and skia-proxy, `R` on pdflib-leading-glyph), the two truth sidecars are handled correctly in *both* directions (deleted where the fix landed, flipped to `#833`/`open` where a second disagreement survives), and `check:baselines` is down to 7 with no truth values touched. The positive control and the Conventional Commits subject came through too.

**Still standing: one item, and it isn't a regression on your part.** The commit has carried a `Co-authored-by: Cursor` trailer since `fd8b734` — the sha I reviewed last time — and I simply missed it. It blocks because `main`'s merge queue composes the squash message from your commit body, so the trailer would be permanent and un-strippable at merge time. `git commit --amend` to delete that one line, force-push your branch, and you're done; details are in the review.

One addition to the nit policy I described earlier, since this round is the case it didn't cover: when we *can't* push to your branch — either because a blocker means we leave it alone, or because a fork hasn't enabled maintainer edits — the small stuff goes out as GitHub `suggestion` blocks instead. That's what the two nits on this review are. You apply them in one click, the commit stays yours, and both are genuinely optional — ignore either if you disagree.

Nothing else is outstanding. Once that trailer is gone this should approve cleanly.

Questions or pushback on any of it, just tag me (@s-annam) here.

src/lib/heuristics/extract/skills.ts
```diff
@@ -138,7 +138,14 @@ function looksLikeContactLink(tok: string): boolean {
   return PROFILE_LABEL_RE.test(t) || PROFILE_HOST_RE.test(t) || URLISH_RE.test(t);
 }
 
+/** One-character tokens that are real, commonly-listed languages. The
+ *  `tok.length < 2` floor in `isSkillToken` is a noise guard against stray
+ *  glyphs left by column splitting; these are the only single characters that
+ *  are not noise, so they are allowlisted rather than lowering the floor. */
+const SINGLE_LETTER_SKILLS = new Set(["c", "r", "d"]);
```
s-annam, line 145: **Nit — the docblock says what is in the set, not what was deliberately kept out.**

#832 makes the *exclusions* the interesting decision: `J`, `K`, `Q` and `F` are all real language names and were left out on purpose, because each is also a plausible stray glyph and admitting them "trades a rare recall win for a common precision loss." That reasoning exists only in the issue. A year from now someone reads these four lines, notices `J` is missing, and re-litigates it from scratch — or just adds it.

One sentence fixes it. Non-blocking, and fine to skip. If you do want it, fold it into the existing commit with `git commit --amend` rather than adding a second commit — the review body explains why I am not offering this as a one-click suggestion:

```ts
/** One-character tokens that are real, commonly-listed languages. The
 *  `tok.length < 2` floor in `isSkillToken` is a noise guard against stray
 *  glyphs left by column splitting; these are the only single characters that
 *  are not noise, so they are allowlisted rather than lowering the floor.
 *  Deliberately excludes J/K/Q/F (#832): each is a real language name, but
 *  each is also a plausible stray glyph, so admitting them trades a rare
 *  recall win for a common precision loss. */
```

s-annam (approved):
Approving. The one blocking finding from the last round is gone, and I re-derived that rather than taking the amend on trust: `b212752` and `668f470` have the **identical tree** (`8048166`) and the identical parent, so the only thing that changed is the commit message — the `Co-authored-by: Cursor` trailer is deleted and no code moved with it. `git log -1 --format=%B` on the head is now four lines of description and nothing else, which is what the merge queue will squash into `main`.

Verdict rule: 0 Blocking → `APPROVE`. Secondary and Nits below are non-blocking and I have pushed nothing to your branch (see the last section for why, since it differs from what I described two rounds ago).

## Blocking

None.

## Secondary

**The PR description covers about a quarter of the diff.** `## Summary` names the allowlist and the unit tests. The diff also re-bakes **five** `expected.json` snapshots (not four — `latex/multi-degree-coursework.expected.json` moves `skillsCount` 17 → 18 alongside the two deedy fixtures, skia-proxy and pdflib-leading-glyph), deletes one `knownWrong` entry, flips another to `#833`/`open`, and drops `UNFILED_TRUTH_CEILING` 9 → 7. Those are the half of #832 that took the most care, and a reader of the PR page cannot see that they happened.

Two things keep this off the blocking list. The **commit message body is accurate and complete** — it names all four workstreams — and that is the text the merge queue composes `main`'s squash message from, so the permanent record is right. And the omission is under-claiming, not over-claiming: nothing in the body asserts a behaviour the code lacks. Worth fixing on the PR page anyway, because the PR is what the next person reads when they bisect to this commit and want to know why five snapshots moved.

## Nits

1. **The docblock records what is in the set, not what was deliberately kept out.** Inline on `skills.ts:141`.
2. **The test plan lists only `skills.test.ts`.** That suite could not have caught the actual risk in this change — a spurious single character admitted on some *other* fixture. The gates that cover it are `corpus.test.ts` (the whole snapshot corpus) and `corpus-roundtrip.test.ts` (the middot-joined skills header has to survive parse → export → re-parse), plus `check:baselines` for the ceiling. Both AC lines in #832 name them. Non-blocking because you demonstrably ran them — the snapshots are correctly re-baked and `verify` is green — the box just does not say so.

## What I verified

Independently, against the binaries and the gates, not against the summary:

- **Every re-baked token is genuinely drawn on the page.** `pdftotext` on each: `C • C++ • CSS • PHP • Assembly` on both deedy fixtures, `Programming Languages: C, HTML, CSS, Java, VHDL, Assembly` on skia-proxy, `Languages: Python, Go, C, C++, …` on multi-degree-coursework, and `Proficient in: Python, SQL, R` on pdflib-leading-glyph. Five fixtures, five real tokens, `+1` char each. No fixture gains a spurious single character.
- **AC checklist for #832: all met.** The allowlist sits before the floor with an early return; matching is case-insensitive; the set is exactly `c`/`r`/`d` with no `J`/`K`/`Q`/`F` creep; `multi-degree-coursework`'s entry is *deleted* rather than flipped (correct — the fix lands in the same PR) with no truth *values* touched; skia-proxy carries `#833` and the note still names both disagreements, which is what the issue thread asked for; `UNFILED_TRUTH_CEILING` drops by 2 from `main`'s current 9, honouring the "take the LOWER number on a rebase conflict" instruction.
- **Gates run locally on `668f470`:** `skills.test.ts` + `corpus.test.ts` + `corpus-roundtrip.test.ts` → 175 passed, no new `KNOWN_FAILURES` rows. `check:baselines` → 7 unfiled, matching the new ceiling exactly. `check:fixtures` → 60 PDFs and 16 sidecars, all personas synthetic. `typecheck`, `lint` clean. `fallow audit --base origin/main` → *no issues in the 10 changed files*; the four complexity findings it prints are inherited and report-only.
- Gate 3b (design system) and 3c (style tokens) do not apply — this diff touches no component and no class name.

## On the auto-fix path, since this round is the case that decides it

Two rounds ago I said that with zero blockers we fix the small stuff ourselves, push it, and approve in the same pass. There are zero blockers now and I still pushed nothing, so the reason should be on the record rather than left as an inconsistency.

`main` merges through a queue that derives the squash message from the branch, so a PR has to reach the queue as **one commit**. When the branch is ours, the auto-fix push is followed immediately by a collapse back to one commit. On a **fork** branch that second half is unavailable: collapsing means force-pushing over your branch, which rewrites your authorship and destroys whatever you have locally, and we do not do that to a contributor's branch under any circumstance. So a nit push here would leave the PR at two commits with no permitted repair — trading a one-sentence docblock for a real invariant violation.

Hence: nit 1 is written out below as text to fold into your existing commit with `git commit --amend` if you want it, and it is genuinely fine to skip. `Allow edits by maintainers` being on is still useful, but the collapse constraint, not permissions, is what governs here.

Nothing outstanding. This is good work on the unglamorous half of the issue, and the amend was clean.

---

s-annam:
Approved — thanks for the quick amend, @shubhransh-gupta.

**Resolved:** the `Co-authored-by: Cursor` trailer is gone, and I checked it the boring way rather than trusting the amend — `b212752` and `668f470` have the same tree hash and the same parent, so the message is the only thing that moved. **Nothing stands.** The two items in the review are a Secondary about the PR description being thinner than the diff, and one docblock Nit; neither blocks and both are yours to take or leave.

One thing worth flagging, because this round is the first where the nit policy I described earlier actually applied and I *still* didn't use it. Zero blockers means we normally fix the small stuff ourselves and push it to your branch — but `main` requires a PR to arrive as a single commit, so that push is always paired with a collapse back to one commit, and a collapse on a fork branch means force-pushing over your work. We won't do that to a contributor's branch, full stop. So on a fork the nit-push path is effectively off regardless of `Allow edits by maintainers`, and small items come to you as text instead. That's a constraint we hadn't hit before, not a change of heart about your PR.

Nice work on the half of #832 nobody enjoys — the sidecar bookkeeping and the ceiling drop were the fiddly part and you got both directions right. A maintainer will take it from here on merging.

Questions or pushback on any of it, tag @s-annam.

Issue #297, opened 2026-07-03
Round-trip: education count inflates on re-parse (phantom entry) + institution pollution


s-annam:
## Symptom

The reconstructed résumé's **education section gains a phantom entry** on re-parse: education count inflates 1→2 or 2→3. Extends the #291 education work (which fixed degree/field/institution collapse) — the *count* is now wrong for a different reason.

One fixture also shows **institution pollution** (a #291 strip edge): `University of California` re-parses as `… · Berkeley Berkeley, CA` (location doubled / date-strip boundary).

## Affected fixtures (round-trip gate `education` category)

Count inflation:
- `google-docs/google-docs-skia-proxy-achievements-oneline.pdf`
- `google-docs/google-docs-skia-proxy-additional-skills.pdf`
- `google-docs/google-docs-skia-proxy-honors-subheadings.pdf`
- `latex/header-as-name-functional-resume.pdf`
- `unknown/chromium-qualified-experience-headers.pdf`
- `word/chanchal-sharma-bulleted-skills.pdf`
- `word/chanchal-sharma-sample.pdf`

Institution pollution (folded in — same edu section):
- `google-docs/google-docs-skia-proxy-programs-skills-software.pdf` (also has `experience`+`skills` — those belong to #D/#E)

## Root cause (hypothesis)

The reconstructed education block (`buildAtsResumeModel` → `renderAtsResumePdf`) emits a line the re-parser reads as a *second* education entry — likely a sub-line (honors, coursework, achievements) or a stray header being promoted to a new entry by `extract/education.ts` segmentation. Institution pollution is the #291 `stripInstitutionDate` / location-strip boundary firing twice or on the wrong anchor.

## Fix direction

- Trace one inflating fixture through the dev harness; find which reconstructed line spawns the phantom entry.
- Tighten education entry segmentation so honors/coursework/achievement sub-lines stay attached, not promoted.
- Fix the institution location/date double-strip.

## Definition of done

- Education count + institution string round-trip for all listed fixtures.
- Delete the `["education"]` lines (and the education part of `programs-skills-software`) from `KNOWN_FAILURES`; corpus gate (#293) stays green.

Found by the #293 corpus round-trip gate. Extends #291. Part of the round-trip audit batch.

PR #841, opened 2026-08-15
fix(education): strip middot/bullet before in/of connective (#839)

### LINKED ISSUE #839: [parser] education — a middot/bullet separator before the `in`/`of` connective leaks it into the field
## What's wrong

`parseDegreeAndField`'s leading-separator strip and `cleanField`'s edge strip disagree about which glyphs count as a separator, and the two run in an order that makes the disagreement visible.

Confirmed on `main` as of `0e706a9` (PR #838), by probing `extractEducation` directly:

```
"B.S. · in Computer Science"  → degree "B.S.", field "in Computer Science"   ← connective leaks
"B.S. • of Computer Science"  → degree "B.S.", field "of Computer Science"   ← connective leaks
"B.S. - in Computer Science"  → degree "B.S.", field "Computer Science"      ← correct
"B.S. — in Computer Science"  → degree "B.S.", field "Computer Science"      ← correct
```

A degree header whose separator is a middot or bullet keeps the `in`/`of` connective in the field of study; the same header with a hyphen or em dash does not. The field then renders as **`B.S. in Computer Science`** in the reconstructed résumé card, the Download PDF, and the re-parse of that PDF.

## Root cause

`src/lib/heuristics/extract/education.ts:534-540`, the tail of `parseDegreeAndField`:

```ts
const fieldRaw = line
  .slice(fieldStart)
  // Drop a leading "in "/"of " connective or a "-"/"—"/":"/"," separator.
  .replace(/^\s*(?:in|of)\s+/i, "")
  .replace(/^\s*[-–—,:]\s*/, "")     // ← no `·` / `•` in the class
  .replace(/^\s*(?:in|of)\s+/i, "");
return { degree, field: cleanField(fieldRaw) };
```

The sequence is deliberately connective → separator → connective, so that either ordering (`in - Computer Science`, `- in Computer Science`) resolves. Trace `"· in Computer Science"`:

1. First connective strip: no-op — the string starts with `·`, not `in`.
2. Separator strip: no-op — `·` is not in `[-–—,:]`.
3. Second connective strip: no-op — still starts with `·`.
4. `cleanField` then strips the leading `·` (its class **was** widened in #835), leaving `"in Computer Science"` — but the connective strips have already run and do not get a second pass.

So #835 widened the glyph vocabulary at the *downstream* strip only. The upstream strip on line 538 still carries the narrow class, and because it runs first, the widening cannot rescue it.

This is one more instance of the repo-wide hazard #835 called out: the separator/bullet glyph vocabulary is duplicated across several regexes in the parser, exporter and scorer and has drifted. #653 tracks unifying it.

## Implementation plan

1. **Add `·` and `•` to the class on `education.ts:538`**, matching what `cleanField` accepts:

   ```ts
   .replace(/^\s*[-–—,:·•]\s*/, "")
   ```

   Update the comment on line 536 so it names the glyphs it actually drops.

2. **Prefer a shared constant if one now exists.** Same instruction as #835 step 2: check `regex.ts` / `line-primitives.ts` for an exported separator class this can consume. If there is none, do **not** introduce a global vocabulary as a side effect — a local widening is the right scope, and #653 owns the unification. Leave the `#653` pointer already present at `education.ts:458` as the precedent.

3. **Consider whether the two strips should share one class outright.** They are eight lines apart and mean the same thing. If a single module-level `const EDGE_SEPARATOR_CLASS` reads cleanly, that is a better fix than widening the literal a second time — and it makes the next drift impossible rather than merely unlikely. Judgement call for the implementer; either shape closes this issue.

## Acceptance criteria

- [ ] `"B.S. · in Computer Science"` parses to `degree = "B.S."`, `field = "Computer Science"` — no leading connective.
- [ ] `"B.S. • of Computer Science"` parses to `degree = "B.S."`, `field = "Computer Science"`.
- [ ] The hyphen and em-dash forms are unchanged: `"B.S. - in Computer Science"` and `"B.S. — in Computer Science"` still yield `field = "Computer Science"`. Assert them, so the fix is pinned as a *widening* and not a rewrite.
- [ ] The #835 guards still hold — in particular the interior separator: `"B.S. Mathematics · Statistics"` → `field = "Mathematics · Statistics"`. This issue must not de-anchor anything.
- [ ] Unit tests in `src/lib/heuristics/extract/education.test.ts`, in the `#835` describe block or a sibling one.
- [ ] `npx vitest run src/lib/heuristics/corpus.test.ts` passes with no fixture regressions. No fixture currently draws this shape, so expect no recall movement — if a number moves, understand why before baking.

## Provenance and severity

Found during the `/pr-review` pass on #838 (the fix for #835) and deliberately **not** auto-fixed there: it changes parser output, and that review's auto-fix commit is bounded to changes that cannot alter behaviour.

**Low impact — no fixture in the 58-PDF corpus produces this shape**, and a `Degree · in Field` header is unusual in the wild (the middot normally separates the field from a date or a location, which #835 already handles). Filing it because the asymmetry is cheap to remove and will otherwise be re-derived by whoever next reads these two strips and assumes they agree.

Refs #835. Refs #653.

shubhransh-gupta:
## Summary
- Widen `parseDegreeAndField`'s separator strip to include `·` and `•`, matching `cleanField`
- Fixes `"B.S. · in Computer Science"` leaking the connective into the field
- Adds regression tests; hyphen/em-dash and interior middot cases unchanged

Fixes #839

## Test plan
- [x] `npx vitest run src/lib/heuristics/extract/education.test.ts` (67 tests)
- [x] Pre-push `npm run verify` (scoped tests + build passed locally)

### DIFF
--- a/src/lib/heuristics/extract/education.test.ts
+++ b/src/lib/heuristics/extract/education.test.ts
@@ -947,3 +947,59 @@ describe("extractEducation — cleanField strips leftover edge middot and bullet
     expect(value[0].field).toBeUndefined();
   });
 });
+
+describe("extractEducation — parseDegreeAndField strips middot/bullet before in/of connective (#839)", () => {
+  it("strips middot then in connective from degree field", () => {
+    const { value } = extractEducation(
+      mkEduSection([
+        "Springfield State University",
+        "B.S. · in Computer Science",
+      ]),
+    );
+    expect(value).toHaveLength(1);
+    expect(value[0].degree).toBe("B.S.");
+    expect(value[0].field).toBe("Computer Science");
+  });
+
+  it("strips bullet then of connective from degree field", () => {
+    const { value } = extractEducation(
+      mkEduSection([
+        "Springfield State University",
+        "B.S. • of Computer Science",
+      ]),
+    );
+    expect(value).toHaveLength(1);
+    expect(value[0].degree).toBe("B.S.");
+    expect(value[0].field).toBe("Computer Science");
+  });
+
+  it("still strips hyphen and em dash before in connective", () => {
+    const { value: hyphen } = extractEducation(
+      mkEduSection([
+        "Springfield State University",
+        "B.S. - in Computer Science",
+      ]),
+    );
+    expect(hyphen[0].field).toBe("Computer Science");
+
+    const { value: emDash } = extractEducation(
+      mkEduSection([
+        "Springfield State University",
+        "B.S. — in Computer Science",
+      ]),
+    );
+    expect(emDash[0].field).toBe("Computer Science");
+  });
+
+  it("preserves interior middot in a two-part field name", () => {
+    const { value } = extractEducation(
+      mkEduSection([
+        "Springfield State University",
+        "B.S. Mathematics · Statistics",
+      ]),
+    );
+    expect(value).toHaveLength(1);
+    expect(value[0].degree).toBe("B.S.");
+    expect(value[0].field).toBe("Mathematics · Statistics");
+  });
+});
--- a/src/lib/heuristics/extract/education.ts
+++ b/src/lib/heuristics/extract/education.ts
@@ -533,7 +533,12 @@ function parseDegreeAndField(line: string): {
   }
   const fieldRaw = line
     .slice(fieldStart)
-    // Drop a leading "in "/"of " connective or a "-"/"—"/":"/","/"·"/"•" separator.
+    // Drop a leading "in "/"of " connective or a "-"/"–"/"—"/","/":"/"·"/"•"
+    // separator. The middot/bullet live here, not only in `cleanField`, because
+    // this strip runs FIRST — `cleanField`'s edge strip accepted them already but
+    // runs after the connective strips, too late to rescue a "· in <field>"
+    // header (#839). The two classes still differ (`cleanField` also takes ";");
+    // unifying the separator vocabulary repo-wide is #653.
     .replace(/^\s*(?:in|of)\s+/i, "")
     .replace(/^\s*[-–—,:·•]\s*/, "")
     .replace(/^\s*(?:in|of)\s+/i, "");
--- a/src/lib/heuristics/extract/education.ts
+++ b/src/lib/heuristics/extract/education.ts
@@ -533,9 +533,9 @@ function parseDegreeAndField(line: string): {
   }
   const fieldRaw = line
     .slice(fieldStart)
-    // Drop a leading "in "/"of " connective or a "-"/"—"/":"/"," separator.
+    // Drop a leading "in "/"of " connective or a "-"/"—"/":"/","/"·"/"•" separator.
     .replace(/^\s*(?:in|of)\s+/i, "")
-    .replace(/^\s*[-–—,:]\s*/, "")
+    .replace(/^\s*[-–—,:·•]\s*/, "")
     .replace(/^\s*(?:in|of)\s+/i, "");
   return { degree, field: cleanField(fieldRaw) };
 }


shubhransh-gupta:
Follow-up to #840 — fixes #839 with the one-line separator widening from the issue's acceptance criteria (`·`/`•` added to `parseDegreeAndField`'s strip class) plus regression tests.

Local `npm run verify` passed (67 education tests + scoped suite). Ready for review whenever you have time — thanks!

src/lib/heuristics/extract/education.ts
```diff
@@ -533,9 +533,14 @@ function parseDegreeAndField(line: string): {
...
-    // Drop a leading "in "/"of " connective or a "-"/"—"/":"/"," separator.
+    // Drop a leading "in "/"of " connective or a "-"/"–"/"—"/","/":"/"·"/"•"
+    // separator. The middot/bullet live here, not only in `cleanField`, because
+    // this strip runs FIRST — `cleanField`'s edge strip accepted them already but
+    // runs after the connective strips, too late to rescue a "· in <field>"
+    // header (#839). The two classes still differ (`cleanField` also takes ";");
+    // unifying the separator vocabulary repo-wide is #653.
     .replace(/^\s*(?:in|of)\s+/i, "")
-    .replace(/^\s*[-–—,:]\s*/, "")
+    .replace(/^\s*[-–—,:·•]\s*/, "")
```
s-annam, line 543: **Secondary — the two classes still disagree, one glyph over.** `cleanField`'s edge strip is `[\s,;:·•–\-—]`; this one is `[-–—,:·•]`. The `;` is in the first and not the second, and the ordering that caused #839 is unchanged — this strip runs *before* the two connective strips, `cleanField` runs after. Verified on the pushed head by probing `extractEducation` directly:

```
"B.S. ; in Computer Science"  → degree "B.S.", field "in Computer Science"    ← still leaks
"B.S. · in Computer Science"  → degree "B.S.", field "Computer Science"       ← fixed by this PR
"B.S. ‣ in Computer Science"  → degree "B.S.", field "‣ in Computer Science"  ← in neither class
```

Not blocking: `;` is outside #839's acceptance criteria, the issue explicitly hands separator-vocabulary unification to #653, and a `Degree ; in Field` header is rarer still than the middot one. Flagging it because it is exactly the class of finding #839 was, and the next reader of these two strips will again assume they agree. If you want to close it here rather than leave it to #653, the issue's own step 3 is the shape: one module-level `const EDGE_SEPARATOR_CLASS` consumed by both strips, which makes the drift impossible instead of merely unlikely. Either way this PR is approved as-is.

src/lib/heuristics/extract/education.test.ts
```diff
@@ -947,3 +947,59 @@ describe("extractEducation — cleanField strips leftover edge middot and bullet
...
+        "Springfield State University",
+        "B.S. • of Computer Science",
+      ]),
+    );
+    expect(value).toHaveLength(1);
+    expect(value[0].degree).toBe("B.S.");
+    expect(value[0].field).toBe("Computer Science");
+  });
+
+  it("still strips hyphen and em dash before in connective", () => {
```
s-annam, line 976: **Nit — this test and the interior-middot one below it pass identically before and after the change.** I confirmed by reverting `education.ts` to `origin/main` and running just this describe block: the two middot/bullet cases fail, these two pass. That is correct — AC 3 and AC 4 of #839 asked for them precisely so the fix is pinned as a *widening* rather than a rewrite — but a later reader scanning a `#839` describe block will reasonably read all four as evidence the fix works. One line of comment saying these two are regression guards ("unchanged by #839; here so a future narrowing of the class is caught") would carry that.

Second, smaller: `expect(hyphen[0].field)` and `expect(emDash[0].field)` index into the array without an `expect(...).toHaveLength(1)` first, unlike the two tests above. If a future change split this header into two entries, `[0]` would still satisfy the assertion and the test would stay green on a real regression.

s-annam (approved):
Approving. The fix is the minimal, correct widening #839 asked for, it is pinned by tests that genuinely fail without it, and it moves no fixture numbers. Two non-blocking items below; the one that was safe to fix without changing behaviour is already pushed.

**Verdict rule:** 0 Blocking → `APPROVE`. Secondary/Nits do not gate.

## What I verified (not taken from the description)

- **Fail-before proof.** Swapped `education.ts` back to `origin/main` and re-ran only the new `#839` block: `strips middot then in connective` fails with `"in Computer Science"`, `strips bullet then of connective` fails with `"of Computer Science"`; both pass with the change. The other two cases pass before *and* after — correct, because AC 3 and AC 4 asked for them as *pins on the widening*, not as evidence of it.
- **Regex.** `[-–—,:·•]` is well-formed: the `-` is first in the class so it is literal, no accidental range. Bytes checked (`c2 b7` middot, `e2 80 a2` bullet).
- **Blast radius.** `parseDegreeAndField` has three call sites, all inside `education.ts`. `cleanField`'s edge strip already accepted `·`/`•` before this PR (#835), so no field that legitimately *starts* with one of those glyphs changes behaviour — they were being stripped downstream already. This PR only moves the strip early enough to also expose the `in`/`of` connective behind it.
- **Shared-constant check (issue step 2).** No exported separator class exists that this could consume — `line-primitives.ts`'s `BULLET_CLASS` is module-local and semantically different (list-bullet detection: includes `*`, `●`; excludes `·`, `,`, `:`). So the local widening is the right scope, exactly as the issue scoped it, and #653 keeps the unification.
- **Gates.** `npm run verify` → exit 0 (typecheck, eslint, `check:fixtures` 58 PDFs + 15 sidecars, `check:baselines`, `check:core`, scoped tests, `vite build`). `npx vitest run src/lib/heuristics/corpus.test.ts` → 62 passed / 1 skipped, **no recall movement** in any dimension, as #839 predicted. fallow's 3 complexity findings on `extractEducation` / `educationFromChunk` / `stripInstitutionLocation` are all pre-existing and report-only — not charged to this PR.

## Acceptance criteria (#839) — 7/7 met

| AC | Status |
|---|---|
| `"B.S. · in Computer Science"` → field `Computer Science` | met (test, fail-before confirmed) |
| `"B.S. • of Computer Science"` → field `Computer Science` | met (test, fail-before confirmed) |
| hyphen + em dash unchanged, asserted | met |
| #835 interior-separator guard still holds | met |
| tests in `education.test.ts`, sibling describe | met |
| `corpus.test.ts` green, no fixture regressions | met (verified locally) |
| `npm run verify` passes | met (verified locally) |

## Secondary

**The sibling classes still disagree — one glyph over.** See the inline note on `education.ts`. Not blocking: it is outside #839's ACs, the issue explicitly hands vocabulary unification to #653, and the shape is rare. But it does make the description's "matching `cleanField`" a near-match rather than a match, and it is the same defect #839 was.

## Nits

- Two of the four new tests are regression guards, not pins — see the inline note on `education.test.ts`. Correct to have; just worth a reader not mistaking them for proof of this fix.

## Fixed in dfca1b3

- `education.ts:536` — the comment claimed to name the glyphs the class drops but omitted `–` (en dash, which was already in the class), and carried no issue citation, which this file otherwise does on nearly every behaviour-driving comment (`#835`, `#297`, `#371`, …). Rewrote it to name every glyph, say *why* the middot/bullet have to live in this strip rather than only in `cleanField`, and point the residual `;` gap at #653. **Comment-only — no behaviour change**, and `npm run verify` is green on the pushed head.

Note for whoever merges: this is now two commits on the branch. Squash it (the repo's one-commit-per-PR rule) — the second is mine, not the author's.

## Gates run

`check:fixtures` (green, no fixture touched by this PR), design-system / style-token / skill-script gates **N/A** — the diff is two files under `src/lib/heuristics/extract/`, no UI, no scripts, no fixtures. fallow: green-for-this-diff (all findings inherited). Description audit: accurate apart from the "matching `cleanField`" overstatement noted above; both test-plan boxes independently re-verified green here.

Reviewed at `dfca1b3fe8b9c14a6a0cf496adfa95d7cf713ac3`.

---

s-annam:
Approved — nice work, @shubhransh-gupta. All seven acceptance criteria on #839 are met, the fail-before was proven for both load-bearing cases, and the corpus run showed no recall movement. This is a clean first contribution.

### I pushed a commit to your branch — here's exactly what and why

`dfca1b3` — **comment-only, no behaviour change.** Your code does what it did before; not a character of logic moved. It rewrites the comment above the separator strip to name the en dash (which the old comment omitted) and to say out loud that this class is still narrower than `cleanField`'s — `cleanField` also takes `;` — with a pointer to #653 for unifying the separator vocabulary repo-wide.

This is the policy I described over on #840 actually firing: **when a review has zero blocking findings, we fix the small stuff ourselves and approve in the same pass** rather than bouncing the PR back for a comment tweak. It only works because you left "Allow edits by maintainers" on. If you'd rather review such changes yourself instead of finding them already pushed, say so — I'll switch to leaving them as review notes on your PRs, no friction either way.

### One small correction

The PR description says the widened class now matches `cleanField`'s. It doesn't quite — `cleanField` still accepts `;`, so `"B.S. ; in Computer Science"` parses the field as `"in Computer Science"`. Harmless here and out of scope for #839, which is why it's a note rather than a change request; the comment I pushed records it so the next reader isn't misled.

### Non-blocking leftovers

One suggestion and one nit remain in the review (inline at `education.ts:543` and `education.test.ts:976`). Both are genuinely optional — take them or leave a reply saying why not, and either resolves the thread.

### Two mechanical notes, neither of which is your homework

- The branch now carries two commits (yours + my comment fix). This repo squashes to one commit per PR, and that happens at merge time — @s-annam handles it. Nothing for you to do; force-pushing to a fork wasn't an option on my side, which is why it's landing this way.
- **CI hadn't reported on this branch when I approved.** An approval here means the diff is right, not that the pipeline is green — the merge check is still the gate.

Questions, or disagreement with any of the above — tag me. Thanks again.

Issue #850, opened 2026-08-16
pr-ready SKILL.md frontmatter is invalid YAML (unquoted argument-hint flow sequence)


s-annam:
## Problem

`.claude/skills/pr-ready/SKILL.md` has a frontmatter `argument-hint` that is not valid YAML:

```yaml
argument-hint: [<pr-number>[,<pr-number>...]]
```

An unquoted value starting with `[` is parsed as a YAML **flow sequence**, and the nested `[` inside it is a syntax error:

```
while parsing a flow sequence
  in "<unicode string>", line 4, column 16:
    argument-hint: [<pr-number>[,<pr-number>...]]
                   ^
expected ',' or ']', but got '['
```

Reproduce:

```bash
python3 -c "
import yaml,sys
yaml.safe_load(open('.claude/skills/pr-ready/SKILL.md').read().split('---')[1])
"
```

## Why it matters

Claude Code currently tolerates this and loads the skill, so there is no visible breakage today. Stricter YAML consumers do not — a harness parsing frontmatter with Go's `yaml.v3` (or any strict parser) **silently drops the skill entirely**, with no error surfaced. The failure mode is invisible: the skill simply stops existing, and `/pr-ready` becomes an unknown command.

That makes this a latent portability bug rather than a cosmetic one. It also means the file cannot be validated by any tooling that parses frontmatter strictly, so the same class of error can accumulate unnoticed in other skills.

## Fix

Quote the value:

```yaml
argument-hint: "[<pr-number>[,<pr-number>...]]"
```

## Suggested follow-up

Add a frontmatter-parse check over `.claude/skills/*/SKILL.md` so this class of breakage is caught mechanically rather than by chance:

```bash
python3 - <<'PY'
import glob, sys, yaml
bad = []
for f in sorted(glob.glob('.claude/skills/*/SKILL.md')):
    try:
        yaml.safe_load(open(f).read().split('---')[1])
    except Exception as e:
        bad.append((f, str(e).splitlines()[0]))
for f, e in bad:
    print(f'{f}: {e}')
sys.exit(1 if bad else 0)
PY
```

At the time of filing, `pr-ready` is the only skill that fails; `collapse-pr`, `open-pr`, `revise-pr`, and `pr-review` all parse cleanly.

## Discovery

Found incidentally while validating the skills touched by #849. Deliberately **not** fixed there — it is pre-existing (it fails at `HEAD` independently of that change) and unrelated to that PR's scope.

PR #851, opened 2026-08-16
feat(skills): extract /collapse-pr as the one guarded collapse

### LINKED ISSUE #849: Extract /collapse-pr: make the one-commit merge-queue invariant a single guarded operation
## Problem

`main` merges through a **merge queue**, and the queue's enqueue API
(`EnqueuePullRequestInput`) carries **no commit-message fields** — it accepts only
`pullRequestId` / `jump` / `expectedHeadOid`. GitHub therefore *derives* the squash
message from repo settings, which are (verified on `offlinecv/OfflineCV`):

```
ruleset "main-merge-queue":  active, merge_method SQUASH
```

The lever that gives us: **a PR holding exactly one commit merges with that commit's
subject and body verbatim.** A PR holding more than one merges with `* `-bulleted
commit soup in `main`'s history.

So "the branch must arrive at the queue as exactly one commit" is a **load-bearing
invariant**, not a style preference. Today that invariant is maintained by three
separate hand-rolled copies of the same logic, and they have already diverged:

| Skill | Collapses? | Rule it applies |
|---|---|---|
| `open-pr` Step 3.6 | yes | always, before the PR exists |
| `revise-pr` Step 5.1 | yes | **final round only** — don't collapse while threads are open on the target, because the reviewer needs to diff just the delta |
| `pr-review` Step 5.5 | **no** | pushes an auto-fix commit and explicitly forbids force-push |

`pr-review` is the gap. When it auto-fixes nits at 0 blockers it pushes a second
commit and leaves the branch multi-commit — silently breaking the invariant that
`open-pr` just established. Nobody notices until the soup is in `main`.

Each copy also re-derives the merge-queue rationale from scratch, so a fourth caller
(the planned `implement-issue` draft-PR review loop) would mean a fourth copy.

## Why a skill, not a helper script

The collapse is not a mechanical `git` incantation — it is a **judgment call with
safety gates**, and the gates differ per caller:

- The combined message must be **written, not concatenated**. It describes the change
  as a whole, not the sequence of steps that produced it. `wip` / `fix lint` /
  `address review` commits are process, not change, and must not survive.
- Collapsing is a force-push. `main` has `dismiss_stale_reviews: true` (verified), so
  collapsing **after** an approval dismisses that approval and deadlocks the PR — it
  can no longer enqueue without a fresh review.
- Force-pushing a branch authored by someone else destroys their local work with no
  warning.

### 1. New skill: `.claude/skills/collapse-pr/SKILL.md`

```
/collapse-pr [<pr>] [--message-file <f>] [--dry-run] [--yes]
```

**Step 0 — Resolve the target, and never trust the local checkout.**

Two call shapes reach this skill and they have *different* notions of "the branch":

| Caller | Target |
|---|---|
| `open-pr` Step 3.6 | pre-push, no PR exists yet — the current checkout **is** the target |
| `revise-pr` 5.1, `pr-review` 5.5, standalone `/collapse-pr <N>` | an **existing PR**, whose head lives at `origin` and may have nothing to do with local state |

For every shape except the `open-pr` one, resolve the target from the **remote**:

```bash
gh pr view "$PR" --json headRefName,headRefOid,headRepositoryOwner
git fetch origin "$HEAD_REF"
BASE_SHA="$(git merge-base "origin/$HEAD_REF" "origin/$BASE_REF")"
```

and do the `reset --soft` + `commit` + `push --force-with-lease` **in a throwaway worktree
checked out at `origin/$HEAD_REF`** — never in the user's checkout. Then remove the worktree.

**This is not hypothetical.** Dry-running the gates against live PR #842 found the local
checkout sitting on that PR's branch (`epic-811-parser-lane`) and **diverged** from origin:
local held one commit, origin held two, and the local commit was a superseded copy sharing
a subject line with one of origin's. Collapsing from that checkout would have force-pushed
the stale tree and destroyed a Windows path-separator test fix that existed only on origin.

**Gate 3e resolves this losslessly — it does not ask a human.** An earlier draft of this
issue made divergence a refusal. That was wrong twice over: it is friction in the common
case (the local copy is simply an older draft), and the question it defers — "is this local
work unique, or superseded?" — is **not mechanically decidable**. Verified: on #842 both
`git cherry` and a reverse-apply of the local patch report the local commit as *not* upstream,
even though origin held a strictly better amended version of the same change. Supersession is
semantic, not byte-identical. Do not try to classify it.

So make the operation lossless instead. If nothing can be lost, nothing needs deciding:

| Local state | Resolution | Human? |
|---|---|---|
| Local-only **commits** on the target branch | preserve at a real ref, then `reset --hard origin/$HEAD_REF` | no |
| Uncommitted changes **this run authored** (explicit path list, e.g. `pr-review`'s `$FIXED_PATHS`) | stage by path, commit on top, let the collapse fold it in | no |
| Uncommitted changes of **unknown provenance** | **refuse** — absorbing stray edits into someone else's PR commit is a human's call | yes |

```bash
STAMP="$(date -u +%Y%m%dT%H%M%SZ)"
BACKUP="collapse-pr-backup/$HEAD_REF-$STAMP"
git branch "$BACKUP" "$(git rev-parse HEAD)"     # a real ref: reflog expires, this doesn't
git reset --hard "origin/$HEAD_REF"
```

Row 2 is not a bolted-on special case — it is exactly what `pr-review` Step 5.5 already does
(auto-fix → commit → collapse), so the gate must not refuse where 5.5 proceeds. Row 3 is the
repo's existing "never absorb stray edits" rule from `open-pr` and `implement-issue`; stage by
explicit path, never `git add -A`. Classify for the *report* only (behind / already-upstream /
possibly-unique) and never let that unreliable classification gate the action.

**Validated live on #842:** the local commit was preserved at
`collapse-pr-backup/epic-811-parser-lane-20260815T234430Z` (`6ba92e9`), the checkout reset to
`72358ab`, divergence went to `0 0`, and the gate passed — with no human decision required.

**Step 0b — The tree-identity assertion, and what it licenses.**

A collapse is a pure history rewrite: `reset --soft` + re-commit changes *which commits
exist*, never *what the files contain*. So the resulting commit's tree object must be
**byte-identical** to the pre-collapse head's tree. Assert it, and abort if it ever fails:

```bash
[ "$(git rev-parse "$PRE"^{tree})" = "$(git rev-parse HEAD^{tree})" ] || abort
```

This assertion is cheap, and it is the strongest correctness check available — it proves
content preservation directly rather than inferring it from a diff.

It also **licenses skipping the pre-push hook**. `offlinecv` installs a managed
`.git/hooks/pre-push` that runs the full `npm run verify` (bypass: `OFFLINECV_SKIP_HOOKS=1`),
and worktrees share `.git/hooks`, so it fires on a collapse push from a throwaway worktree —
where it fails immediately for want of `node_modules`. Bootstrapping `node_modules` there
just to re-run a suite is minutes of work to re-verify **a tree that by construction did not
change**. So: assert tree identity first, and only then push with `OFFLINECV_SKIP_HOOKS=1`.
If the assertion fails, the tree *did* change, the skip is not licensed, and the whole
collapse must abort — never skip the hook unconditionally.

**Step 1 — Detect the regime.** Query `squash_merge_commit_title` /
`squash_merge_commit_message` and whether a merge-queue ruleset is active. If the repo
merges via plain squash where `gh pr merge --squash --subject --body` can supply the
message directly, **collapsing is unnecessary** — say so and exit 0. This is what keeps
the skill portable to any future repo instead of hard-coding offlinecv's settings.

**Step 2 — Count commits.** `git log --oneline "origin/$BASE..HEAD" | wc -l`. If `1`,
no-op, exit 0. Idempotent by construction — safe to run twice.

**Step 3 — Safety gates.** Refuse (or require explicit `--yes` plus a printed warning)
when any of:

| Gate | Why |
|---|---|
| An approval exists and `dismiss_stale_reviews` is on | the force-push dismisses it → PR can't enqueue → deadlock |
| Unresolved review threads exist on the target | `revise-pr` 5.1's rule — the reviewer still needs to diff just the delta |
| PR head is **not** authored by the current user or an agent | force-push silently destroys a contributor's local branch |
| The `--force-with-lease` lease would fail | someone pushed while we worked; never fight it |

**Step 4 — Compose the message.** Written, not concatenated. Conventional-commit
subject, `Closes #N` / `Refs #N` trailer, **no AI attribution trailer** (repo
convention). Accept `--message-file` so a human can hand-author it. Print it for
confirmation. `--dry-run` stops here.

**Step 5 — Execute.**

```bash
git reset --soft "$(git merge-base HEAD "origin/$BASE")"
git commit -F <message-file>
git push --force-with-lease origin HEAD     # never bare --force
```

### 2. Rewire the three existing callers

- **`open-pr` Step 3.6** → delegate to `/collapse-pr`. Keep the step (it runs
  pre-push, before the PR exists, so most gates are trivially satisfied) but stop
  re-deriving the rationale.
- **`revise-pr` Step 5.1** → delegate, keeping its own "final round only" decision.
  The *decision* stays in `revise-pr`; only the *mechanics* move.
- **`pr-review` Step 5.5** → **new behavior**, see below.

### 3. `pr-review` Step 5.5: collapse before the approval

The ordering is the whole point. `pr-review` Step 5 already documents the principle —
*"the push happens before the approval, so there is no prior approval to dismiss"* —
and the collapse must ride in that same slot:

```
fix nits → commit → collapse to one commit → push --force-with-lease
        → derive inline anchors      (Step 5.5 already requires this order)
        → post APPROVE on the collapsed head
```

This requires **relaxing one existing rule**. Step 5.5 currently says
*"never `--force` / `--force-with-lease` here."* That rule exists to protect against
the author pushing mid-review — but `--force-with-lease` **is** precisely the mechanism
that detects that case. New rule:

> Use `--force-with-lease`. If the lease fails, the author pushed while the review ran:
> **skip the collapse**, fall back to the plain push, and note it in the review body.
> Never bare `--force`.

Same protection, strictly more capability.

**Author-class gate** — collapse only when the PR head is agent-authored or owned by
the maintainer:

| Author | Auto-fix push | Collapse (force-push) |
|---|---|---|
| Maintainer / agent-authored, in-repo branch | yes | **yes** |
| Named contributor, in-repo branch | yes, announced | **no** — destroys their local branch |
| Outside contributor (fork) | usually blocked by permissions | n/a |

### 4. `pr-review`: suggestion blocks when pushing isn't allowed

Today a fork PR (or a contributor branch we won't force-push) falls back to
comments-only. GitHub's ```` ```suggestion ```` blocks are strictly better for that
case: the author applies them with one click, authorship and consent are preserved,
and no branch is touched.

Emit suggestion blocks for every Nit/Secondary finding that is a **localized textual
replacement** in the diff, whenever Step 5.5 decides not to push. Findings that need
a behavioral change are not suggestions — keep those as prose.

## Acceptance criteria

- [ ] `.claude/skills/collapse-pr/SKILL.md` exists and implements Steps 1–5 above.
- [ ] Regime detection exits cleanly (no-op, exit 0) on a repo where `--subject`/`--body`
      can supply the squash message — the skill is not offlinecv-hard-coded.
- [ ] Running `/collapse-pr` on an already-single-commit branch is a no-op (idempotent).
- [ ] All four safety gates from Step 3 are implemented and each states its reason when
      it fires.
- [ ] **Step 0 target resolution:** for an existing PR, the head is resolved from `origin`
      and the rewrite happens in a throwaway worktree at `origin/$HEAD_REF`, never in the
      user's checkout. The worktree is removed afterward.
- [ ] **Tree-identity assertion:** the collapsed commit's tree hash is compared to the
      pre-collapse head's and the collapse aborts on mismatch.
- [ ] **Pre-push hook handling:** the push sets `OFFLINECV_SKIP_HOOKS=1`, and does so *only*
      after the tree-identity assertion passes — never unconditionally.
- [ ] **Gate 3e is lossless, not a refusal:** local-only commits are preserved at a real
      `collapse-pr-backup/<branch>-<utc>` ref and the checkout is reset automatically; the
      report names the backup ref. Uncommitted changes this run authored are committed on
      top rather than blocking. Only uncommitted changes of unknown provenance refuse.
- [ ] Gate 3e never blocks on a classification of whether local work is "unique" — that is
      not mechanically decidable and must not gate the action.
- [ ] `--dry-run` prints the composed message and changes nothing.
- [ ] `--message-file` lets a human supply the message verbatim.
- [ ] `open-pr` 3.6 and `revise-pr` 5.1 delegate to `/collapse-pr` instead of carrying
      their own `reset --soft` + rationale. `revise-pr` retains its "final round only"
      decision.
- [ ] `pr-review` 5.5 collapses **before** posting the review, with `--force-with-lease`,
      and falls back to a plain push when the lease fails.
- [ ] `pr-review` 5.5 applies the author-class gate and never force-pushes a branch it
      does not own.
- [ ] `pr-review` emits ```` ```suggestion ```` blocks for localized nits whenever it
      does not push.
- [ ] No skill re-derives the merge-queue rationale; `collapse-pr` is the single source
      of truth and the others link to it.
- [ ] The PR opened for this issue itself arrives at the queue as exactly one commit
      (dogfood).

## Out of scope / follow-ups

- **`/pr-watch`** — a self-pacing loop over `pr-sweep` with head-SHA state so unchanged
  PRs are not re-reviewed. Lives in the maintainer's `~/tools`, not this repo.
- **Auto-merge gating** (`gh pr merge --auto` behind an author-class + sensitive-path
  check). Depends on `/pr-watch` existing; file separately once this has run for a while.
- Changing repo squash settings to `PR_BODY`/`PR_TITLE`. That would make commit count
  irrelevant, but the PR body would then land verbatim in `git log` (Summary / Review
  focus / Test plan sections included). Rejected as more invasive than the collapse.

s-annam:
## Summary

`main` merges through a merge queue whose enqueue API carries no commit-message fields, so GitHub derives the squash message from `squash_merge_commit_message: COMMIT_MESSAGES`. Only a **single-commit** PR merges with a written message; anything else lands in `main` as `* `-bulleted commit soup.

That invariant was maintained by three hand-rolled copies that had already diverged — and `pr-review` broke it outright, pushing an auto-fix commit while explicitly forbidding the force-push needed to restore one commit.

This extracts `.claude/skills/collapse-pr/` as the single owner of both the mechanics and the rationale. `open-pr`, `revise-pr` and `pr-review` keep only their own decision about *whether* to collapse.

Closes #849.

## What the skill enforces

- **Target resolution.** An existing PR's head is read from `origin` and rewritten in a throwaway worktree; the local checkout is never assumed to be the target. `--in-place` lets a caller that genuinely owns the checkout say so, instead of inferring mode from whether a PR happens to exist.
- **Tree-identity assertion.** The collapsed tree must be byte-identical to the pre-collapse head. That assertion — and only it — licenses skipping the managed `pre-push` hook, and only in a throwaway worktree, never in place where the hook is the sole local verification.
- **Five gates.** Stale approval and open threads are soft (`--yes` proceeds); branch ownership and the push lease are hard and never overridable.
- **Gate 3e is lossless, not a refusal.** Whether local work is unique or merely superseded is *not mechanically decidable*, so the skill does not try: commits are preserved at a `collapse-pr-backup/<branch>-<utc>` ref before any reset, and creating that ref is a hard precondition of the reset.
- **`pr-review` ordering.** Collapse sits between the auto-fix push and the review post — the only ordering that works, since the collapse rewrites every SHA (so anchors must come after) and `dismiss_stale_reviews: true` means collapsing after an approval would dismiss it. The author's message is preserved via `--message-file` rather than replaced with one the reviewer wrote.

## Review focus

1. **`collapse-pr` gate 3e (Step 3).** The most destructive code here — it runs `git reset --hard` on a checkout. Check the backup ref is an unconditional precondition on every path, and that the `[ "$MODE" = worktree ]` guards cannot be bypassed.
2. **Row 1's cherry-pick path.** Added late to stop a naive `push HEAD` from publishing local-only commits. Least-reviewed code in the diff.
3. **`pr-review` Step 5.5 ordering** — that no route reaches a force-push after an abandoned auto-fix.
4. **Length.** `collapse-pr` is 1083 lines. Growth is gates and guarded paths; a reviewer may reasonably disagree about how much prose earns its place.

## Test plan

- `npm run verify` green (also enforced by the `pre-push` hook on this branch — not bypassed).
- All four in-scope skill frontmatters parse as strict YAML. (`pr-ready` does not — **pre-existing**, filed as #850, deliberately untouched here.)
- Single-source-of-truth verified: only `collapse-pr` restates the merge-queue rationale; `open-pr` and `revise-pr` carry zero `reset --soft`.
- All cross-referenced step and row numbers resolve across the five files.
- This PR arrives as **exactly one commit**, dogfooding its own invariant.
- Collapse mechanics rehearsed live against PR #842: 2 commits → 1, tree hash `56a742c98fb9d6afe1370ab5f9432c82abc6c5d3` identical on both sides, zero diff. `--force-with-lease` verified on both paths — valid lease accepted, stale lease refused with the remote unchanged.

## Adversarial review

Three rounds ran before this PR was opened. Findings and dispositions:

**Round 1 — 5 blocking, all fixed**
1. Step 5b's `git worktree remove --force` was unguarded; in `inplace` mode `$WORKDIR` is the user's checkout, and linked worktrees delete cleanly. Would have deleted the worktree it was working in.
2. `OFFLINECV_SKIP_HOOKS=1` was applied on both push paths — silently removing `npm run verify` from `open-pr` entirely, since the hook was its only local verification.
3. `trap … EXIT` cannot span separate tool calls, so cleanup either fired immediately or never; two prose claims asserted otherwise.
4. `pr-review` step 5 ran the collapse on paths that had pushed nothing.
5. The collapse replaced the author's commit message with one the reviewer composed.

**Round 2 — 7 blocking, all fixed**
1. Backup ref and `reset --hard` were unconditional siblings; a failed `git branch` fell through to irreversible loss (the skill does not run under `set -e`).
2. Gate 3e's `worktree`-only restriction was prose with **zero** `MODE` references in any of its five code blocks — repeating round 1's lesson.
3. `--authored-path` whitelists path *names*, but provenance is content in a specific checkout; a global list applied across worktrees could publish a user's unrelated edits. No iteration construct existed for multiple matches.
4. Row 1's `push HEAD` published every local-only commit — the #842 hazard, ungated.
5. A rejected row-1 push was misattributed solely to gate 3d; being behind, or the `pre-push` hook, are likelier causes.
6. `status --porcelain` includes `??`, and `.git/info/exclude` un-ignores `node_modules`, so gate 3e would have refused **100% of the time** on this repo.
7. Re-running `open-pr` after a PR exists flipped mode to `worktree`, collapsed origin's head, and discarded local commits while reporting success.

Plus 7 nits (rename/space-safe `status` parsing, bash 3.2 `set -u` array expansion, teardown naming on exit paths, and others).

**Round 3** added ~230 lines (cherry-pick path, `--authored-worktree`, `--in-place`, real mode guards) which **no subagent review has seen** — reviewing that here is the point of opening the PR.

A note on process: a solo verification pass returned "clean" before each of rounds 1 and 2, which then found 5 and 7 blocking defects respectively. Treat the sections above as the areas where direct inspection has been demonstrably unreliable.

### DIFF
--- a/.claude/skills/collapse-pr/SKILL.md
+++ b/.claude/skills/collapse-pr/SKILL.md
@@ -1,0 +1,1083 @@
+---
+name: collapse-pr
+description: Collapse a branch to exactly one commit whose message is the message that lands in `main` — the merge queue derives the squash message and cannot be handed one, so a multi-commit PR merges as bulleted commit soup. Resolves an existing PR's head from origin and rewrites it in a throwaway worktree rather than trusting the local checkout, detects whether the repo even needs a collapse at all, runs five safety gates and a tree-identity assertion before anything is pushed, and is a clean no-op on a branch that already holds one commit. Use when the user says "collapse the PR", "/collapse-pr", "squash this branch to one commit", or when open-pr / revise-pr / pr-review needs the one-commit invariant restored.
+argument-hint: "[<pr-number>] [--repo owner/repo] [--base <ref>] [--in-place] [--message-file <f>] [--authored-worktree <d>] [--authored-path <p>] [--authored-message <m>] [--dry-run] [--yes]"
+---
+
+# Collapse PR
+
+Take a branch from "N scratch commits" to **exactly one commit whose message is
+the message you want in `main`**: resolve the target from `origin` → detect whether
+this repo's merge regime needs a collapse at all → count the commits → run the
+safety gates → compose the message → `reset --soft` + `commit` → **assert the tree
+did not change** → `push --force-with-lease`.
+
+This is the one place the collapse lives. `open-pr` (Step 3.6), `revise-pr`
+(Step 5.1), and `pr-review` (Step 5.5) each *decide* whether to collapse on their
+own terms and then delegate the mechanics here; none of them re-derives the
+rationale below.
+
+## Input
+
+Parse `$ARGUMENTS` for a **PR number** (`849`, `#849`) and optionally `--repo
+owner/repo`, `--base <ref>`, `--in-place`, `--message-file <f>`,
+`--authored-worktree <d>`, `--authored-path <p>` (repeatable),
+`--authored-message <m>`, `--dry-run`, `--yes`.
+
+| Flag | Effect |
+|---|---|
+| `<pr-number>` | The PR whose head is the target. Omit it and Step 0 infers one from the current branch; if there is no PR yet the run continues **in place** on the checkout, and Step 3's PR-dependent gates are trivially satisfied |
+| `--repo owner/repo` | Target repo; defaults to the current one |
+| `--base <ref>` | The ref this branch merges into. Defaults to the PR's own base, or the repo default branch in in-place mode. Pass it when collapsing a **stacked** branch that has no PR yet — collapsing a stack layer against `main` would swallow the parent layer's commits into the child |
+| `--in-place` | Force `MODE=inplace`: **this checkout is the target**, whatever `gh pr view` would have inferred. `open-pr` Step 3.6 always passes it. It does **not** suppress PR discovery — a PR that exists is still read, so gates 3a/3b stay live |
+| `--message-file <f>` | Use this file's contents verbatim as the collapsed message. Nothing is composed |
+| `--authored-worktree <d>` | The **one** checkout whose dirty paths this run authored. `--authored-path` is honoured only there; dirty paths in any *other* worktree are unknown provenance and refuse. Required for `--authored-path` to have any effect |
+| `--authored-path <p>` | Repeatable. A path **this run authored** *in `--authored-worktree`*, so gate 3e may commit it rather than treating it as a stray edit. A whitelist: any dirty path not listed drops that tree to 3e row 2 and refuses |
+| `--authored-message <m>` | Subject for the commit gate 3e row 1 makes. Defaults to `chore: fold in run-authored changes (collapse-pr gate 3e)`. Normally short-lived — the collapse folds it away — but it survives on `origin` if a later hard gate refuses after row 1 pushed, so it must be self-describing rather than invented |
+| `--dry-run` | Run every check, print the gate results and the composed message, change nothing — including gate 3e, which otherwise writes. Exit 0 |
+| `--yes` | Proceed past the two **soft** gates (stale approval, open threads), printing each that fired and why. Never overrides the hard gates (3c, 3d) or 3e's unknown-provenance refusal |
+
+**With a PR number and without `--in-place`, the local checkout is not the target
+and is never rewritten.** Step 0 resolves the head from `origin` and does the work
+in a throwaway worktree. That is a correctness property, not a convenience: see
+*Why the target comes from `origin`* below for the live case where trusting the
+checkout would have destroyed committed work.
+
+**`--in-place` is the inverse claim, and only a caller that owns the checkout may
+make it.** Mode must not be inferred from whether a PR happens to exist: `open-pr`
+Step 3.6 runs pre-push on a checkout that is the *only* copy of the work, and on a
+**re-run** — pushing more commits to a branch whose PR already exists — the
+inference would find that PR, flip to `worktree` mode, and collapse `origin`'s head
+while the new local commits sat unread. `open-pr` Step 4 then skips its own push
+("Step 3.6 collapsed"), so the commits are never published and 3e row 3 resets them
+away. `--in-place` is what makes the caller's notion of the target authoritative.
+
+**`--yes` means "the caller already made this judgement", not "skip the checks."**
+Both `revise-pr` and `pr-review` reach this skill having already decided the
+collapse is correct for their round, and the state they decided on is still
+visible on GitHub when they call — `revise-pr` resolves its threads *after* the
+push, so its threads are legitimately still open here. The gates still run and
+still print; `--yes` only changes whether a soft gate stops the run.
+
+## Why this skill exists
+
+**This section is the single source of truth for the merge-queue rationale.** If
+you are reading `open-pr`, `revise-pr`, or `pr-review` and want to know *why* the
+branch must arrive as one commit, it is here and nowhere else.
+
+`main` merges through a **merge queue**, and the queue's enqueue API carries **no
+commit-message fields** — `EnqueuePullRequestInput` accepts `pullRequestId` /
+`jump` / `expectedHeadOid` and nothing else. So the squash message cannot be
+handed to GitHub at merge time; GitHub *derives* it from repo settings, which on
+`offlinecv/OfflineCV` are:
+
+```
+squash_merge_commit_title:   COMMIT_OR_PR_TITLE
+squash_merge_commit_message: COMMIT_MESSAGES
+```
+
+With those, a **multi-commit** PR merges with every commit message concatenated as
+`* ` bullets — which is how `wip`, `fix lint`, and `address review comments on PR
+#458` end up in `main`'s history permanently. The `COMMIT_OR_` prefix is the only
+lever, and it gives exactly one:
+
+> **A PR with exactly one commit merges with that commit's subject and body
+> verbatim** — PR title and body ignored, no bullet soup, nothing from the PR body
+> leaking into `git log`.
+
+So "the branch reaches the queue as one commit" is a **load-bearing invariant**,
+not a style preference, and the branch's single commit *is* the artifact. Branch
+commits are scratch. `docs/CONTRIBUTING-PROCESS.md` → **Squash messages** states
+the same policy for human contributors who never run a skill.
+
+## Process
+
+### Step 0 — Resolve the target, and never trust the local checkout
+
+Two call shapes reach this skill, and they mean different things by "the branch":
+
+| Caller | Target | Where the rewrite happens |
+|---|---|---|
+| `open-pr` Step 3.6 (always passes `--in-place`) | pre-push — the current checkout **is** the target, whether or not a PR exists yet | **in place**, in the user's checkout |
+| `revise-pr` 5.1, `pr-review` 5.5, standalone `/collapse-pr <N>` | an **existing PR**, whose head lives at `origin` and may have nothing to do with local state | **a throwaway worktree** detached at `origin/$HEAD_REF` |
+
+**The mode comes from the caller first and from PR discovery only as a fallback.**
+`IN_PLACE=1` (set by `--in-place`) pins `MODE=inplace` even when the inference
+below finds a PR; discovery still runs, because gates 3a and 3b need `PR_NUM`
+whichever mode we are in.
+
+```bash
+REPO="${REPO:-$(gh repo view --json nameWithOwner -q .nameWithOwner)}"   # offlinecv/OfflineCV
+PR_NUM="${PR_NUM:-$(gh pr view --json number -q .number 2>/dev/null || true)}"
+
+if [ -n "$IN_PLACE" ] || [ -z "$PR_NUM" ]; then
+  # ---- in-place mode: this checkout is the source of truth ----
+  # Either the caller asserted it (--in-place) or there is no PR to resolve.
+  MODE=inplace
+  WORKDIR="$(git rev-parse --show-toplevel)"
+  HEAD_REF="$(git rev-parse --abbrev-ref HEAD)"
+  IS_FORK=false
+  BASE_REF="${BASE:-$(gh repo view "$REPO" --json defaultBranchRef -q .defaultBranchRef.name)}"
+  git fetch origin "$BASE_REF"
+  # A PR may still exist on this branch even here (open-pr re-run). PR_NUM is
+  # already set if so, and gates 3a/3b use it — the mode says where the rewrite
+  # happens, not whether the PR-shaped gates apply.
+else
+  # ---- PR mode: origin is the source of truth ----
+  MODE=worktree
+  { read -r HEAD_REF; read -r HEAD_OID; read -r PR_BASE; read -r IS_FORK; } < <(
+    gh pr view "$PR_NUM" --repo "$REPO" \
+      --json headRefName,headRefOid,baseRefName,isCrossRepository \
+      --jq '.headRefName, .headRefOid, .baseRefName, .isCrossRepository')
+  BASE_REF="${BASE:-$PR_BASE}"
+  git fetch origin "$HEAD_REF" "$BASE_REF"
+
+  # The SHA gh reported must still be the SHA we fetched, or someone pushed in
+  # between and every gate below would be judging a stale target.
+  if [ "$(git rev-parse "origin/$HEAD_REF")" != "$HEAD_OID" ]; then
+    # No Step 5b needed here, and this is the only exit in the skill that can say
+    # so: the worktree is created three lines below, so there is nothing to remove.
+    echo "head moved between the PR read and the fetch — re-run" >&2; exit 1
+  fi
+
+  # Keep the mktemp parent in a variable: `git worktree remove` deletes only the
+  # child, so Step 5b has to `rm -rf` the parent or it is orphaned in /tmp.
+  WORKDIR_PARENT="$(mktemp -d)"
+  WORKDIR="$WORKDIR_PARENT/collapse-pr-$PR_NUM"
+  git worktree add --detach "$WORKDIR" "origin/$HEAD_REF"
+fi
+```
+
+**Read the multi-field JSON with `read`, never `eval`.** Git ref names may contain
+`$`, backticks, `;`, `&`, and quotes, so an `eval "$(gh … --jq '"HEAD_REF=\(…)"')"`
+turns a branch name into shell. `--jq '.a, .b, .c'` emits one value per line and
+`read` takes each literally.
+
+**`--detach` is not optional.** The target branch is very often already checked out
+somewhere — the user's main clone, or a sibling agent worktree — and `git worktree
+add <path> <branch>` refuses to check the same branch out twice. Detaching at
+`origin/$HEAD_REF` sidesteps that, and it is the honest description of what we
+want: a disposable head, not a second copy of a branch.
+
+Everything from Step 2 onward runs as `git -C "$WORKDIR" …`. **Run Step 5b on every
+exit path, refusals included.** There is deliberately no `trap`: these steps execute
+as a *sequence of separate Bash calls* carrying values forward by hand, so a handler
+installed here would fire when this call's shell exits — destroying the worktree
+before Step 1 ran. Teardown is an explicit step you invoke, not a handler.
+
+#### Why the target comes from `origin` and not from `HEAD`
+
+**This is not hypothetical.** Dry-running these gates against live PR #842 found the
+user's main checkout on that PR's branch holding a **superseded copy** of one of
+origin's commits — same subject line, different tree — and one commit fewer.
+Collapsing from that checkout would have force-pushed the stale tree and destroyed a
+Windows path-separator test fix that existed only on origin. Resolving from `origin`
+makes the rewrite safe by construction; gate 3e covers the other half of the hazard —
+the local commit the collapse would strand — by preserving it at a branch ref rather
+than by asking anyone to judge it, and carries the case in full as its worked
+example, including why no mechanical test can tell superseded work from unique work.
+
+### Step 0b — The tree-identity assertion, and what it licenses
+
+A collapse is a **pure history rewrite**: `reset --soft` + re-commit changes *which
+commits exist*, never *what the files contain*. So the collapsed commit's tree
+object must be byte-identical to the pre-collapse head's:
+
+```bash
+[ "$PRE_TREE" = "$(git -C "$WORKDIR" rev-parse 'HEAD^{tree}')" ] || abort
+```
+
+Step 5 runs it between the commit and the push. It is cheap, and it is the
+strongest correctness check available here — it proves content preservation
+directly instead of inferring it from a diff. Verified on #842: two commits
+collapsed to one, tree `56a742c98fb9d6afe1370ab5f9432c82abc6c5d3` on both sides,
+zero diff.
+
+**What it licenses: skipping the pre-push hook — in `worktree` mode only.**
+`offlinecv` installs a managed `.git/hooks/pre-push` that runs the full `npm run
+verify` (bypass: `OFFLINECV_SKIP_HOOKS=1`). Worktrees **share `.git/hooks`** with
+the main clone, so that hook fires on a collapse push from a throwaway worktree —
+where it fails immediately for want of `node_modules`. Bootstrapping `node_modules`
+there purely to re-run a suite over a tree that by construction did not change is
+minutes of work for no information.
+
+**"Re-run" is the whole licence, and it does not hold in `inplace` mode.** In
+`worktree` mode the tree came from `origin` — it already passed the hook on the push
+that put it there — and the assertion proves this collapse did not change it, so the
+suite would only be re-verifying a verified tree. In `inplace` mode the tree has
+never been pushed and nothing has verified it; the assertion still proves the
+*collapse* changed nothing, but "changed nothing" over an unverified tree is still
+unverified. `open-pr` Step 3.6 is the `inplace` caller and carries no other local
+verification, so bypassing here would publish an unverified tree to `origin`. **Let
+the hook run in `inplace` mode.**
+
+So, in `worktree` mode: **assert tree identity first, and only then push with the
+bypass set.**
+
+> If the assertion fails, the tree *did* change, the skip is **not** licensed, and
+> the whole collapse aborts. Never skip the hook unconditionally — the assertion is
+> the entire reason the skip is sound, and the mode is the entire reason the
+> assertion is enough.
+
+On a repo with no documented bypass, `git push --no-verify` is the portable
+equivalent — same assertion, same mode restriction, and nothing else licenses it.
+
+### Step 1 — Resolve the base pair, then detect the regime
+
+Ask the repo, don't assume offlinecv. This step is what keeps the skill portable:
+on a repo where the squash message can simply be supplied at merge time, the
+collapse buys nothing and this skill says so and exits 0.
+
+**Two different base refs are in play, and conflating them is the one way this step
+gets a stacked PR wrong.** `BASE_REF` (resolved in Step 0) is what this branch
+merges into *today* and defines the commit range to collapse; `QUEUE_BASE` is the
+branch whose merge regime eventually decides the squash message. They differ
+exactly when the PR is stacked on another feature branch — GitHub retargets the
+child to the default branch once the parent merges, so the child faces the queue no
+matter what its base says today. Probe the regime on `QUEUE_BASE`; collapse against
+`BASE_REF`.
+
+```bash
+OWNER="${REPO%%/*}"; NAME="${REPO##*/}"
+QUEUE_BASE="$(gh repo view "$REPO" --json defaultBranchRef -q .defaultBranchRef.name)"
+
+# Squash-message policy. These two fields live ONLY on the REST repo object —
+# `gh repo view --json squashMergeCommitTitle` is not a field and errors out.
+SQ_TITLE="$(gh api "repos/$REPO" --jq .squash_merge_commit_title)"   # COMMIT_OR_PR_TITLE
+SQ_BODY="$(gh api "repos/$REPO"  --jq .squash_merge_commit_message)" # COMMIT_MESSAGES
+
+# Is a merge queue configured for the branch this change eventually lands on?
+# `mergeQueue` is null when none is configured, and jq indexes null to null,
+# so the `//` default is reached cleanly.
+MQ="$(gh api graphql -f query='
+query($o:String!,$n:String!,$b:String!){
+  repository(owner:$o,name:$n){ mergeQueue(branch:$b){ configuration{ mergeMethod } } }
+}' -f o="$OWNER" -f n="$NAME" -f b="$QUEUE_BASE" \
+  --jq '.data.repository.mergeQueue.configuration.mergeMethod // "none"')"
+```
+
+| `MQ` | `SQ_BODY` | Verdict |
+|---|---|---|
+| `SQUASH` | `COMMIT_MESSAGES` | **Collapse.** The message can't be supplied at merge time and is derived from the commits |
+| `SQUASH` | `PR_BODY` / `BLANK` | **No collapse needed** — the merge body comes from the PR body, not the commits |
+| `MERGE` / `REBASE` | any | **No collapse needed** — the queue isn't squashing, so commit count doesn't decide the message |
+| `none` | any | **No collapse needed** — no queue, so the merger can pass `gh pr merge --squash --subject … --body …` directly |
+
+On any "no collapse needed" row: print the row and the two settings that produced
+it, run **Step 5b**, and **exit 0**. Do not rewrite history to satisfy a constraint
+this repo does not have. Say the one thing the caller needs to hear on the `none` row: whoever
+merges must actually pass `--subject`/`--body`, because the web UI's default
+button will still derive a message from `SQ_TITLE`/`SQ_BODY`.
+
+`--yes` does **not** override this step. It is a capability check, not a safety
+gate — there is no risk being accepted, only work that would accomplish nothing.
+
+### Step 2 — Count the commits (this is what makes the skill idempotent)
+
+Step 0 already fetched `origin/$BASE_REF`, and in `worktree` mode `HEAD` is
+`origin/$HEAD_REF` — so this counts the **remote** head's commits, never the local
+checkout's:
+
+```bash
+BASE_SHA="$(git -C "$WORKDIR" merge-base HEAD "origin/$BASE_REF")"
+N="$(git -C "$WORKDIR" rev-list --count "$BASE_SHA..HEAD")"
+```
+
+| `N` | Do |
+|---|---|
+| `0` | Nothing on the branch — exit 0, say there is nothing to collapse |
+| `1` | **Already one commit** — exit 0, no-op |
+| `>1` | Continue |
+
+Both exit-0 rows run **Step 5b** on the way out. `N == 1` exiting 0 is the whole of
+the idempotence guarantee: a second run of
+`/collapse-pr` on a branch this skill just collapsed does nothing, touches no
+refs, and reports success. Callers may invoke it unconditionally.
+
+`$BASE_SHA..HEAD` (two dots, and anchored on the merge base rather than on
+`origin/$BASE_REF` directly) counts only commits reachable from `HEAD` and not from
+the merge base — so a branch that merged the base *in*, or a base that has moved on
+since the branch forked, still reports the branch's own commit count. `BASE_SHA` is
+also exactly the `reset --soft` target in Step 5, so the count and the rewrite can
+never disagree about where the branch begins.
+
+### Step 3 — Safety gates
+
+`PR_NUM`, `HEAD_REF`, `IS_FORK`, and `MODE` were all resolved in Step 0. `PR_NUM`
+may legitimately be empty: `open-pr` calls this **before the PR exists**, which is
+the cheapest possible moment because three of the five gates are then trivially
+satisfied. On an `open-pr` **re-run** `PR_NUM` is populated while `MODE` is still
+`inplace` — 3a and 3b then fire on their own merits, which is the point of keeping
+mode and PR discovery independent (Step 0).
+
+Each gate below **prints its reason when it fires** — a refusal that only says
+"refused" costs the caller a second run to find out what happened.
+
+**Any refusal below must run Step 5b before stopping.** A refusal is an exit path
+like any other, and there is no `trap` to clean up behind it (Step 0).
+
+| # | Gate | Fires when | Class |
+|---|---|---|---|
+| A | A stale approval would be dismissed | the PR has ≥1 `APPROVED` review and the base dismisses stale reviews on push | **soft** — `--yes` proceeds |
+| B | Unresolved review threads on this PR | ≥1 thread with `isResolved == false` | **soft** — `--yes` proceeds |
+| C | The branch is not ours to rewrite | fork PR, or any commit on the branch was authored by someone else | **hard** — never overridden |
+| D | The push lease would fail | the remote head is not an ancestor of the commit we are about to push | **hard** — never overridden |
+| E | A local checkout diverges from the target | `MODE=worktree` and some worktree on `$HEAD_REF` holds work `origin/$HEAD_REF` does not | **resolves losslessly** — refuses only on uncommitted changes of unknown provenance |
+
+#### 3a — Stale approval (soft)
+
+```bash
+APPROVALS=0
+if [ -n "$PR_NUM" ]; then
+  APPROVALS="$(gh pr view "$PR_NUM" --repo "$REPO" --json latestReviews \
+    --jq '[.latestReviews[] | select(.state == "APPROVED")] | length')"
+fi
+
+# Classic branch protection. Needs admin, so a 403 here is expected, not an error.
+DISMISS="$(gh api "repos/$REPO/branches/$BASE_REF/protection" \
+  --jq '.required_pull_request_reviews.dismiss_stale_reviews' 2>/dev/null || echo unknown)"
+
+# A ruleset can carry the same setting, and this endpoint is readable without admin.
+# `first | if . == null` and not `first // "unknown"`: jq's `//` treats `false` as
+# empty, so a ruleset that explicitly sets the flag to `false` would report
+# "unknown" — which this gate reads as `true` below, firing on a repo that does not
+# dismiss stale reviews at all.
+DISMISS_RULE="$(gh api "repos/$REPO/rules/branches/$BASE_REF" \
+  --jq '[.[] | select(.type == "pull_request")
+             | .parameters.dismiss_stale_reviews_on_push]
+        | first | if . == null then "unknown" else . end' 2>/dev/null \
+  || echo unknown)"
+```
+
+Use `if` blocks, not `[ -n "$PR_NUM" ] && …` one-liners: the `&&` form returns
+non-zero when the guard is false, which aborts the run the moment anything wraps
+this in `set -e`.
+
+Either probe answering `true` fires the gate. `unknown` from **both** also counts
+as `true` — the unreadable case is a contributor without admin, and guessing
+`false` there is guessing in the direction that deadlocks the PR. Say which probe
+answered, so a `--yes` is an informed one rather than a shrug.
+
+Reason to print when it fires: *the force-push dismisses the existing approval;
+`main` requires one, so the PR can no longer enter the merge queue until someone
+reviews it again.* On `offlinecv/OfflineCV` this gate is live —
+`dismiss_stale_reviews` is `true` on `main`.
+
+#### 3b — Unresolved review threads (soft)
+
+```bash
+OPEN_THREADS=0
+if [ -n "$PR_NUM" ]; then
+  OPEN_THREADS="$(gh api graphql -f query='
+  query($owner:String!,$name:String!,$pr:Int!){
+    repository(owner:$owner,name:$name){
+      pullRequest(number:$pr){ reviewThreads(first:100){ nodes{ isResolved path line } } }
+    }
+  }' -f owner="$OWNER" -f name="$NAME" -F pr="$PR_NUM" \
+    --jq '[.data.repository.pullRequest.reviewThreads.nodes[]
+           | select(.isResolved == false)] | length')"
+fi
+```
+
+`-F pr=` (typed integer), never `-f` — the GraphQL variable is `Int!` and a string
+422s.
+
+Reason to print when it fires: *a reviewer with an open thread is coming back for
+another round, and they need to diff **just your delta** — a collapse replaces the
+whole branch and costs them that diff.* This is `revise-pr` Step 5.1's rule; the
+decision of whether this round is the final one belongs to the caller, which is
+why the gate is soft.
+
+#### 3c — Branch ownership (hard)
+
+```bash
+SELF_EMAIL="$(git config user.email)"
+FOREIGN="$(git -C "$WORKDIR" log --format='%ae' "$BASE_SHA..HEAD" \
+  | sort -u | grep -vFx "$SELF_EMAIL" || true)"
+```
+
+`IS_FORK` came from Step 0. Refuse — with no `--yes` path — when either holds:
+
+- **`IS_FORK` is `true`.** The head lives in a fork; we generally lack push rights,
+  and where `maintainerCanModify` grants them, force-pushing a contributor's fork
+  is the most destructive thing this skill could do.
+- **`FOREIGN` is non-empty.** Agent-authored commits use the local git identity
+  and pass; a named contributor's do not. Print the addresses.
+
+**The condition is commit authorship, not the PR author's login.** A maintainer can
+open a PR for a branch someone else wrote, and in that case the login says *ours*
+while the commits say otherwise — the commits are what a force-push destroys, so
+they are what the gate reads. Still print the PR author (`gh pr view --json author
+-q .author.login`) and your own (`gh api user --jq .login`) in the refusal, because
+"whose branch is this" is the first thing the caller will ask.
+
+Two reasons to print, because only the first is obvious: *a force-push replaces
+their commits, so their local branch silently diverges and their next `git pull`
+is a non-fast-forward* — and *the collapsed commit is authored by whoever runs the
+collapse, so their authorship is erased from `main`'s history.* The second is why
+this gate has no override: no flag makes it correct to publish someone else's work
+under your name.
+
+The correct escape hatch is social, not a flag: ask the author to run
+`/collapse-pr` on their own branch, or merge their PR as-is and accept the bullets.
+Run Step 5b, then stop.
+
+#### 3d — The push lease (hard)
+
+```bash
+# Exact ref, not a pattern: `ls-remote --heads origin main` would also match
+# refs/heads/feat/main, and the lease would then be pinned to the wrong branch.
+REMOTE_SHA="$(git ls-remote origin "refs/heads/$HEAD_REF" | cut -f1)"
+
+if [ -n "$REMOTE_SHA" ]; then
+  git -C "$WORKDIR" fetch origin "$HEAD_REF"   # objects must exist to test ancestry
+  if ! git -C "$WORKDIR" merge-base --is-ancestor "$REMOTE_SHA" HEAD; then
+    echo "remote $HEAD_REF has commits we do not have — someone pushed while we worked" >&2
+    exit 1
+  fi
+fi
+```
+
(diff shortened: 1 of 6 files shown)


.claude/skills/collapse-pr/SKILL.md
```diff
@@ -0,0 +1,1083 @@
...
+
+if git -C "$WT" merge-base --is-ancestor "$WT_PRE_FIX_SHA" "origin/$HEAD_REF"; then
+  # $WT held nothing origin lacks: our new commit is the only delta.
+  git -C "$WT" push origin "HEAD:$HEAD_REF"       # plain, fast-forward only
+else
+  # $WT has local-only commits. Publish OURS and only ours: back the whole tip up
+  # per row 3 first (so nothing is lost), then replay just the fix onto the
+  # remote head inside $WORKDIR, which is already detached at origin/$HEAD_REF.
+  FIX_SHA="$(git -C "$WT" rev-parse HEAD)"
+  # ... row 3's backup of $FIX_SHA runs here, with its own `||` refusal ...
```
s-annam, line 646: This comment suggests running the backup commands of Row 3 here. It would make the skill more self-contained and robust to explicitly include the Row 3 backup commands (specialized for `$FIX_SHA`) inside the `else` block instead of a placeholder comment.

s-annam (commented):
An extremely thorough and maintainer-grade implementation of the `/collapse-pr` skill and its integration across the PR lifecycle.

---

s-annam (commented):
Reviewed the whole diff against #849's acceptance criteria, not against the description — which I read last, per this skill's own ordering rule. **All 17 ACs of #849 are met**, the regime/gate probes all behave live as documented, and the description is accurate on every claim I could check. The three blockers below are all in the ~230 lines of round 3 that no prior review has seen, and all three sit inside `collapse-pr` gate 3e row 1 — the code the `## Review focus` section correctly named as the least-reviewed in the diff.

**Verdict rule applied:** normally ≥1 Blocking → `REQUEST_CHANGES`. This review posts `COMMENT` under an explicit maintainer override reserving the verdict for themselves on this PR. The same override disabled Step 5.5 (no auto-fix, no commit, no push, no `/collapse-pr` invocation — the collapse code is what this PR changes, and running it for the first time on the PR that contains it is not a risk worth taking) and forbade any enqueue or auto-merge. **Nothing was merged, approved, or enqueued; the branch is exactly as I found it.**

## Prior rounds — confirmed still fixed

- **Round 1 (5 blocking):** Step 5b is guarded by `if [ "$MODE" = worktree ]` (L986); `OFFLINECV_SKIP_HOOKS=1` is worktree-mode-only via the `SKIP=()` array (L915); the `trap` is gone and replaced with an explicit "run Step 5b on every exit path" contract (L162-166); `pr-review` step 5 opens with "Only if step 4 pushed"; the collapsed message comes from `--message-file` holding `git log -1 --format=%B "$PRE_FIX_SHA"`. ✅
- **Round 2 (7 blocking):** the backup ref is chained with `||` **and** re-verified by `rev-parse` before the reset (L738-749); 3e's mode restriction is now real code in 7 places, not prose; `--authored-worktree` scopes the whitelist and the `while IFS= read -r WT` loop exists; row 1's push is ancestry-tested rather than a bare `push HEAD`; the rejection-triage table has three causes; `--untracked-files=no` is used everywhere (`.git/info/exclude:18` is indeed `!/node_modules` — verified); `--in-place` exists and `open-pr` passes it. ✅

I re-report none of these.

## Blocking

Three findings, all one code path: `collapse-pr` gate 3e **row 1**.

### B1 — Row 1 has no "did we author anything?" precondition, so it publishes the user's local-only commit

Row 1's table entry fires on *"uncommitted changes this run authored"*, but the executable blocks carry no such condition — and the iterate block (L524-529) explicitly says the rows are **the loop body**, run for every `$WT`. So on a matched worktree that is **clean but diverged**, row 1 still executes:

1. `$WT != $AUTHORED_WORKTREE` → `DIRTY_COUNT` is `0` → **no refusal** (L573-580).
2. `git add` stages nothing; `git commit` fails "nothing to commit" — and L760-762 of this same file states that an unchained nonzero exit is *silently discarded* in this execution model.
3. `WT_PRE_FIX_SHA` is therefore the **user's local-only commit**.
4. `merge-base --is-ancestor "$WT_PRE_FIX_SHA" "origin/$HEAD_REF"` → false → **else branch**.
5. `FIX_SHA` = the user's commit → `git -C "$WORKDIR" cherry-pick "$FIX_SHA"` → `push origin "HEAD:$HEAD_REF"`.

**The user's unreviewed local commit is published into someone else's PR**, and the Step 6 report describes that checkout as row 3 "preserved + reset" — it never says the commit was also *pushed*.

This reproduces on the skill's own flagship scenario. In the #842 worked example (L794-803) the main checkout was clean and held exactly one local-only commit, `6ba92e9`. Under this code that commit is cherry-picked onto origin's head and pushed into PR #842. Round 2 finding 4 narrowed this hazard from "all local-only commits" to "the tip local-only commit" — it did not remove it, because row 1 still runs on worktrees it has no business touching.

**Fix:** give row 1 a real precondition, in code, mirroring the mode guard's own "code, not prose" standard:

```bash
[ "$MODE" = worktree ] || exit 1
[ -n "${AUTHORED_WORKTREE:-}" ] && [ "$WT" = "$AUTHORED_WORKTREE" ] || { echo "row 1: not the authored checkout — skipping to row 2"; ROW1=skip; }
DIRTY_COUNT="$(git -C "$WT" status --porcelain --untracked-files=no -z | tr -dc '\0' | wc -c | tr -d ' ')"
[ "$DIRTY_COUNT" != 0 ] || { echo "row 1: nothing dirty in $WT — no fold-in needed"; ROW1=skip; }
```

and make the commit/ancestry/push blocks no-ops when `ROW1=skip`. The non-authored-worktree branch should fall through to rows 3/4, never into the commit-and-publish machinery.

### B2 — The cherry-pick path's push is rejected 100% of the time on this repo, and the triage table misdiagnoses it

```bash
git -C "$WORKDIR" cherry-pick "$FIX_SHA"
git -C "$WORKDIR" push origin "HEAD:$HEAD_REF"
```

`$WORKDIR` is the throwaway worktree from `git worktree add --detach` under `mktemp -d`. Worktrees share `.git/hooks` (`core.hooksPath` is unset — verified), so `.git/hooks/pre-push` fires and runs `npm run verify` in a directory with a `package.json` and **no `node_modules`**. That is the exact failure Step 0b describes at L199-202 as the reason the bypass exists — but this push carries no `OFFLINECV_SKIP_HOOKS=1` and no `--no-verify`.

The rejection-triage table then sends the reader the wrong way (L688):

> | the rejection came from `.git/hooks/pre-push` | row 1's push is a plain push from a **real checkout**, so the managed hook fires … | fix the failing gate. Never set `OFFLINECV_SKIP_HOOKS=1` here |

"a real checkout" is true of the **if** branch only. In the else branch the push comes from the throwaway worktree, the gate is not red — the dependencies are simply absent — and "fix the failing gate" is unactionable. The run wedges *after* a backup ref has already been created, with no defined next step.

**Fix:** decide the licence explicitly for this push. The cherry-picked content is new relative to `origin`, so Step 0b's tree-identity licence does **not** cover it — either `npm install` in `$WORKDIR` before this push, or push the fix from `$WT` on a temporary branch and fast-forward, or state that this path requires a bootstrapped worktree. Then split the triage row into "if-branch (real checkout)" and "else-branch (`$WORKDIR`)" with different answers.

### B3 — In the cherry-pick path the backup ref is a comment, not code, and the prose then tells row 3 not to create one

L646:

```bash
  # ... row 3's backup of $FIX_SHA runs here, with its own `||` refusal ...
```

That placeholder is the only thing standing between this path and an unbacked `reset --hard`, and L659-662 then instructs:

> **Row 3 is already half-done for this `$WT`.** The backup above covered the local-only commits *and* the fix commit on top of them, so when row 3 reaches this checkout it must not mint a second ref — only the `reset --hard` remains.

But row 3's code (L721-752) has no "already backed up" check. So an agent gets one of two outcomes, and neither is the intended one: it follows the *code* and mints a second ref (harmless, but contradicts the report contract), or it follows the *prose*, skips row 3's backup, and — because L646 was never expanded into real commands — runs `git reset --hard "origin/$HEAD_REF"` **with no backup ref at all**. That is precisely the irreversible-loss shape round 2 finding 1 was blocking for, on the one path where the loss cannot be undone.

This is also the same defect class as round 2 finding 2: a safety property stated in prose inside a code block instead of being executable.

**Fix:** inline the actual backup commands (the `while` name-search, `git branch … || refuse`, and the `rev-parse` verification) into the else branch, and give row 3 an explicit `[ -n "${BACKUP:-}" ]` short-circuit so it resets without re-minting. Report the one ref, as L659-662 intends.

## Secondary

1. **`collapse-pr` L579 — the row-2 refusal does not run Step 5b in code.** `[ "$DIRTY_COUNT" = 0 ] || { echo …; exit 1; }` carries a comment saying "Run Step 5b before stopping, like every refusal in this skill" — but the block just exits. This fires in `worktree` mode, the one mode where Step 5b is not a no-op, so it orphans both the linked worktree and `$WORKDIR_PARENT` under `/tmp`. Every other refusal in the skill states the teardown in surrounding prose; this one contradicts itself inline.

2. **`pr-review` L581-587 — the `cut -c4-` subtraction errs in the *unsafe* direction, not the safe one.** The prose claims: *"it errs by over-subtracting, which drops a path to row 2 and refuses. That is the safe direction."* It does the opposite. `git status --porcelain -z` emits a rename as two records — `R  <new>\0` then a **bare** `<orig>\0` with no `XY ` prefix (verified). `cut -c4-` chops three characters off that bare second field, so `src/foo.ts` becomes `/foo.ts` and **fails to subtract** `src/foo.ts` from `$FIXED_PATHS`. The path stays whitelisted, reaches 3e as `--authored-path`, and the user's in-progress rename can be committed into the PR — the exact outcome the paragraph exists to prevent. Narrow trigger (the reviewer must have edited a path the user is mid-rename on), but the stated safety argument is inverted. Fix: parse the same way `collapse-pr` L586-594 already does, or drop only entries whose record began with a status prefix.

3. **`pr-review` L560-565 — the partial-failure recovery rationale does not hold.** *"if step 4 committed nothing, or committed and failed to push, the fixes are sitting uncommitted and 3e can only fold them in if it knows they are ours."* If step 4 committed and the push failed, the fixes are **committed**, not uncommitted — 3e row 1 never sees them, row 3 backs them up and `reset --hard`s them out of the checkout. And that state cannot arise in-run anyway, because step 5 opens with "Only if step 4 pushed." On an actual re-run the leftover commit is local-only, so row 3 quietly relocates the previous run's fix to a `collapse-pr-backup/…` ref and both skills report success. Either describe that outcome or add a "leftover fix commit" check to step 1.

4. **`collapse-pr` L111 — swallowed error decides the destructive mode.** `PR_NUM="${PR_NUM:-$(gh pr view --json number -q .number 2>/dev/null || true)}"`: an auth failure, rate limit, or network blip yields an empty `PR_NUM`, and L113's `[ -z "$PR_NUM" ]` then selects `MODE=inplace` — the mode that rewrites the user's own checkout. This is the "swallowed errors — `2>/dev/null || true` that hides auth/rate-limit, not just 'already exists'" antipattern from `pr-review` gate 3e, applied to a mode decision. Distinguish "no PR exists" (exit 1 with the `no pull requests found` message) from "the call failed" and refuse on the latter.

5. **`collapse-pr` L598-600 — the normal path runs a command that errors.** In the documented happy case (`pr-review` step 4 already committed `$FIXED_PATHS`, so "gate 3e sees them clean") row 1 still reaches `git commit -m …` on a clean tree and fails. Harmless in outcome, but the skill's own execution model has no `set -e`, so the run continues past a nonzero exit on its main path — and an agent that *does* stop on it aborts a collapse that should have succeeded. Same root cause as B1; fixing B1's precondition fixes this.

## Nits

1. `collapse-pr` L508 — `MATCHES_FILE` (and `MSG_FILE` at L859 when it is a `mktemp`) are never removed; Step 5b tears down the worktree but not the temp files.
2. `collapse-pr` L729-732 — the backup-name search is TOCTOU against a concurrent run: two runs can both find `-1` free, one loses `git branch`, and the `||` then prints "Fix the ref store and re-run", which misdiagnoses a name race as a broken ref store. Worth a distinct message for `exit 128` / "already exists".
3. `collapse-pr` L886-890 — the net-no-change branch exits `1`. `pr-review` step 5 reads any non-zero as "a hard gate refused" and reports it as such. A distinct exit code, or a note in `pr-review`, would keep the report honest.
4. `docs/CONTRIBUTING-PROCESS.md` §Squash messages still carries a full copy of the merge-queue derivation, while `collapse-pr` L64 declares itself "the single source of truth for the merge-queue rationale" and the test plan says "only `collapse-pr` restates" it. Both are defensible (the doc is the rationale file, and the same diff says `collapse-pr` "holds the canonical one"), but the three statements do not quite agree.
5. `collapse-pr` L584-585 — "git quotes paths containing spaces" is given as a reason to avoid `cut -c4-`. In `-z` mode git never quotes; the rename-mangling argument alone carries the point.

## AC checklist — #849

All 17 acceptance criteria **met**. Spot-verified live rather than read: regime detection (`mergeQueue.configuration.mergeMethod` → `SQUASH`, `squash_merge_commit_title` → `COMMIT_OR_PR_TITLE`, `squash_merge_commit_message` → `COMMIT_MESSAGES`), gate 3a's dual probe (branch protection returns `true`; the ruleset probe returns `unknown`, and the documented `first | if . == null` guard is what keeps that from reading as `false`), the `--in-place` mode contract, the tree-identity assertion, 3e's losslessness, `--dry-run` holding 3e's writes back, and the dogfood criterion (the PR holds exactly **1** commit).

The one place implementation departs from the issue is `pr-review` 5.5's ordering: #849 §3 sketched `fix → commit → collapse → push`, the PR ships `fix → commit → plain push → collapse`. That is a correct departure, not a gap — `/collapse-pr` reads the head from `origin`, so a still-local fix commit would not be in the collapse — and the AC as written ("collapses before posting the review, with `--force-with-lease`, falls back to a plain push when the lease fails") is satisfied.

## Description accuracy (gate 3f) — accurate

Every checkable claim round-tripped to the diff and held:

- `npm run verify` green — `verify`, `fallow`, and Cloudflare Pages all `SUCCESS`. ✅
- Four in-scope frontmatters parse as strict YAML and `pr-ready` does not — verified with a strict parser; `pr-ready` fails exactly as described (`argument-hint` flow sequence), and it is untouched by this diff. ✅
- `open-pr` and `revise-pr` carry zero `reset --soft` — confirmed; the only remaining occurrence repo-wide is `pr-review`'s unrelated auto-fix revert. ✅
- Cross-referenced step and row numbers resolve — spot-checked every cross-file reference in the diff (`3d`/`Step 5`/`3e row 3`/`Step 3.6`/`5.1`/`5.5`/`Step 6`); all resolve. ✅
- Exactly one commit. ✅

`## Review focus` was honest and correct: items 1 and 2 named gate 3e and row 1's cherry-pick path as the least-reviewed code, and that is where all three blockers are. Under this skill's own rules a focus section that points at the right hunk is a point in the author's favour — I reviewed the other four files and the rest of `collapse-pr` regardless, and found nothing above Nit there.

On focus item 4 (length): 1083 lines is a lot, but I would not cut it. The three blockers above all come from a *step that was described in prose but not written as code* — which is the failure mode that shrinking this file would multiply, not reduce. The prose that is actually load-bearing here is the kind that gets promoted into `if` statements.

## Gates run

| Gate | Result |
|---|---|
| 3a fixture PII | **skipped** — no fixture binaries in the diff |
| 3b design-system / reuse | **skipped** — no `src/components/**` changes |
| 3c style tokens | **skipped** — no `src/**` changes |
| 3d dead code (`fallow`) | green on CI (`fallow` check `SUCCESS`); not re-run locally |
| 3e command-level bugs in skill/script files | **the whole review** — B1, B2, B3, S1-S5, N1-N3 |
| 3f description accuracy | run; accurate |

Head SHA reviewed: `f860dae`. 👀 was posted before the review began (Step 0.6).

---

.claude/skills/collapse-pr/SKILL.md
```diff
@@ -0,0 +1,1083 @@
...
+the #842 hazard performed without a gate.** Rows run 1 → 2 → 3 → 4, so row 1 fires
+*before* row 3 has looked at local-only commits. If `$WT` holds commits `origin`
+lacks, a plain push publishes **all of them** into the PR — the very commits row 3
+merely *backs up*, precisely because whether they are unique or superseded is
+undecidable. Test ancestry first and take the branch that fits:
+
+```bash
+[ "$MODE" = worktree ] || exit 1        # 3e never runs in inplace mode (above)
+
+if git -C "$WT" merge-base --is-ancestor "$WT_PRE_FIX_SHA" "origin/$HEAD_REF"; then
```
s-annam, line 638: **Blocking (B1).** This ancestry test decides on `$WT_PRE_FIX_SHA`, which row 1 sets unconditionally at L598 — including on worktrees where nothing was authored and nothing was dirty. On a clean-but-diverged `$WT`, `git commit` above failed ("nothing to commit", silently discarded per L760-762), so `$WT_PRE_FIX_SHA` is the **user's local-only commit**, this test goes false, and the else branch cherry-picks and pushes it into the PR.

That is the #842 worked example at L794-803: a clean main checkout holding `6ba92e9`. Round 2 finding 4 narrowed this from "every local-only commit" to "the tip one"; it did not close it.

Row 1 needs a real precondition — `$WT` is `--authored-worktree` **and** the tree is dirty — expressed in code, the same standard the mode guard holds itself to.

.claude/skills/collapse-pr/SKILL.md
s-annam, line 646: **Blocking (B3).** This is a comment, not code — and it is the only backup standing between this path and the `reset --hard`. L659-662 then tells row 3 not to mint a second ref for this `$WT`, but row 3's code (L721-752) has no "already backed up" short-circuit. So an agent either mints two refs (contradicting the report contract) or, following the prose, skips row 3's backup and resets **with no backup ref at all** — the irreversible-loss shape round 2 finding 1 was blocking for.

Inline the real commands here (name search → `git branch … || refuse` → `rev-parse` verification) and give row 3 an explicit `[ -n "${BACKUP:-}" ]` skip.

.claude/skills/collapse-pr/SKILL.md
```diff
@@ -0,0 +1,1083 @@
...
+  # $WT held nothing origin lacks: our new commit is the only delta.
+  git -C "$WT" push origin "HEAD:$HEAD_REF"       # plain, fast-forward only
+else
+  # $WT has local-only commits. Publish OURS and only ours: back the whole tip up
+  # per row 3 first (so nothing is lost), then replay just the fix onto the
+  # remote head inside $WORKDIR, which is already detached at origin/$HEAD_REF.
+  FIX_SHA="$(git -C "$WT" rev-parse HEAD)"
+  # ... row 3's backup of $FIX_SHA runs here, with its own `||` refusal ...
+  git -C "$WORKDIR" cherry-pick "$FIX_SHA"
+  git -C "$WORKDIR" push origin "HEAD:$HEAD_REF"
```
s-annam, line 648: **Blocking (B2).** This push comes from `$WORKDIR` — the throwaway worktree — which shares `.git/hooks` with the main clone (`core.hooksPath` unset, verified). `pre-push` runs `npm run verify` there with no `node_modules`, so this push is rejected every time on this repo. That is exactly the failure Step 0b cites at L199-202 as the reason the bypass exists, but there is no `OFFLINECV_SKIP_HOOKS=1` and no `--no-verify` here.

The triage table at L688 then misreads it: "row 1's push is a plain push from a **real checkout**" is true of the `if` branch only, and "fix the failing gate" is unactionable when the gate is not red — the dependencies are absent. The run wedges after a backup ref already exists.

Step 0b's tree-identity licence does not cover this push (the cherry-picked content is new relative to `origin`), so pick one deliberately: `npm install` in `$WORKDIR` first, or publish the fix from `$WT` and fast-forward. Then split that triage row per branch.

.claude/skills/collapse-pr/SKILL.md
```diff
@@ -0,0 +1,1083 @@
...
+```bash
+[ "$MODE" = worktree ] || exit 1        # 3e never runs in inplace mode (above)
+
+if [ "$WT" != "${AUTHORED_WORKTREE:-}" ]; then   # :- — the flag is optional
+  # Not the caller's checkout: nothing here is ours by construction.
+  # Any dirty path at all -> row 2.
+  DIRTY_COUNT="$(git -C "$WT" status --porcelain --untracked-files=no -z \
+                 | tr -dc '\0' | wc -c | tr -d ' ')"
+  # Row 2. Run Step 5b before stopping, like every refusal in this skill.
+  [ "$DIRTY_COUNT" = 0 ] || { echo "row 2: $WT is dirty and is not --authored-worktree" >&2; exit 1; }
```
s-annam, line 579: **Secondary.** The comment says "Run Step 5b before stopping, like every refusal in this skill" — but the block exits without doing it. This refusal only fires in `worktree` mode, which is the one mode where Step 5b is *not* a no-op, so it orphans the linked worktree and `$WORKDIR_PARENT` under `/tmp`.

Either call the teardown here or move the instruction into surrounding prose, the way every other refusal in this file does.

.claude/skills/collapse-pr/SKILL.md
```diff
@@ -0,0 +1,1083 @@
...
+        R*|C*) IFS= read -r -d '' ORIG        # rename/copy: the ORIGINAL path is
+               printf '%s\n' "$ORIG" ;;       # a second NUL-terminated field
+      esac
+    done
+# every path emitted above must appear in $AUTHORED_PATHS, or this is row 2.
+# A rename emits BOTH ends, because committing it touches both.
+
+WT_PRE_FIX_SHA="$(git -C "$WT" rev-parse HEAD)"   # BEFORE the commit — see the push below
+git -C "$WT" add -- ${AUTHORED_PATHS[@]+"${AUTHORED_PATHS[@]}"}   # by path, never `git add -A`/`.`
+git -C "$WT" commit -m "${AUTHORED_MESSAGE:-chore: fold in run-authored changes (collapse-pr gate 3e)}"
```
s-annam, line 600: **Secondary.** On the documented happy path this errors. `pr-review` step 4 has already committed `$FIXED_PATHS`, so — as L612 puts it — "gate 3e sees them clean", and `git commit` on a clean tree exits non-zero with "nothing to commit". The skill runs without `set -e`, so the run continues past it (into B1); an agent that *does* stop aborts a collapse that should have succeeded.

Fixing B1's row-1 precondition fixes this too.

.claude/skills/collapse-pr/SKILL.md
```diff
@@ -0,0 +1,1083 @@
...
+| `revise-pr` 5.1, `pr-review` 5.5, standalone `/collapse-pr <N>` | an **existing PR**, whose head lives at `origin` and may have nothing to do with local state | **a throwaway worktree** detached at `origin/$HEAD_REF` |
+
+**The mode comes from the caller first and from PR discovery only as a fallback.**
+`IN_PLACE=1` (set by `--in-place`) pins `MODE=inplace` even when the inference
+below finds a PR; discovery still runs, because gates 3a and 3b need `PR_NUM`
+whichever mode we are in.
+
+```bash
+REPO="${REPO:-$(gh repo view --json nameWithOwner -q .nameWithOwner)}"   # offlinecv/OfflineCV
+PR_NUM="${PR_NUM:-$(gh pr view --json number -q .number 2>/dev/null || true)}"
```
s-annam, line 111: **Secondary.** `2>/dev/null || true` makes an auth failure, rate limit, or network blip indistinguishable from "no PR exists" — and L113 then reads the empty value as a licence for `MODE=inplace`, the mode that rewrites the user's own checkout. A swallowed error is choosing the destructive branch.

This is the "swallowed errors" item from `pr-review` gate 3e applied to a mode decision. Separate the two: match `gh`'s "no pull requests found" and refuse on any other non-zero exit.

.claude/skills/pr-review/SKILL.md
```diff
@@ -425,29 +453,176 @@ is NOT set:
...
+   warned that `gh pr checkout` carries a dirty tree onto the PR branch, and the fix you
+   just wrote may land in a file the user was already editing. There is no way to hand 3e
+   "only my hunks" — the flag names *files* — so committing such a file commits the union,
+   publishing the user's in-progress edits into someone else's PR commit. Whitelist only
+   what was clean when you arrived:
+
+   ```bash
+   # Drop from FIXED_PATHS anything that appears in PRE_DIRTY_FILE.
+   comm -23 <(printf '%s\n' ${FIXED_PATHS[@]+"${FIXED_PATHS[@]}"} | sort -u) \
+            <(tr '\0' '\n' < "$PRE_DIRTY_FILE" | cut -c4- | sort -u)
```
s-annam, line 582: **Secondary.** This errs in the unsafe direction, not the safe one that L586 claims.

`git status --porcelain -z` emits a rename as two records: `R  <new>\0` and then a **bare** `<orig>\0` with no `XY ` prefix (verified locally). `cut -c4-` chops three characters off that bare field, so `src/foo.ts` → `/foo.ts` and `src/foo.ts` is **not** subtracted from `$FIXED_PATHS`. It stays whitelisted, goes out as `--authored-path`, and the user's in-progress rename can be committed into the PR — the outcome this paragraph exists to prevent.

Parse it the way `collapse-pr` L586-594 already does, or drop only records that actually carried a status prefix.

.claude/skills/pr-review/SKILL.md
```diff
@@ -425,29 +453,176 @@ is NOT set:
...
+   `--yes` covers only the soft gates: a prior third-party approval is about to be
+   dismissed by this push either way, and this run replaces it with its own `APPROVE` a
+   moment later, so the PR still enters the queue with one approval. The gates it cannot
+   override — branch ownership, the push lease, and 3e's stray-edit refusal — are the ones
+   that matter here, and `/collapse-pr` enforces them regardless of the flag.
+
+   **Pass `--authored-worktree` and one `--authored-path` per surviving entry of
+   `$FIXED_PATHS`.** In the normal flow they change nothing — step 4 already committed
+   those paths, so gate 3e sees them clean — but they are what makes a **re-run after a
+   partial failure** work: if step 4 committed nothing, or committed and failed to push,
```
s-annam, line 563: **Secondary.** If step 4 committed and the push failed, the fixes are **committed**, not uncommitted — 3e row 1 never sees them, and row 3 backs them up and `reset --hard`s them out of the checkout. And that state cannot arise inside one run: step 5 opens with "Only if step 4 pushed."

On a real re-run the leftover commit is local-only, so row 3 relocates the previous run's fix to a `collapse-pr-backup/…` ref and both skills report success. Either describe *that* as the recovery, or add a leftover-fix-commit check to step 1.

.claude/skills/collapse-pr/SKILL.md
```diff
@@ -0,0 +1,1083 @@
...
+  list --porcelain` prints `detached` where an attached worktree prints `branch
+  refs/heads/…`, so the `awk` below never matches one. That is exactly why Step 0's
+  own `git worktree add --detach` is correct: `$WORKDIR` is on `origin/$HEAD_REF`
+  and would otherwise match its own scan, and be "resolved" against itself. State
+  it, because read as an accident it looks like a hole.
+
+```bash
+# 3e is worktree-mode ONLY. The guard produces an EMPTY match list in inplace
+# mode, which is what makes every block below a no-op there — prose cannot.
+MATCHES_FILE="$(mktemp -t collapse-pr-3e)"
```
s-annam, line 508: **Nit.** `$MATCHES_FILE` is never removed (nor `$MSG_FILE` at L859 when it comes from `mktemp`). Step 5b tears down the worktree but leaves the temp files behind on every path.

.claude/skills/collapse-pr/SKILL.md
```diff
@@ -0,0 +1,1083 @@
...
+[ "$MODE" = worktree ] || exit 1        # 3e never runs in inplace mode (above)
+
+STAMP="$(date -u +%Y%m%dT%H%M%SZ)"
+BACKUP="collapse-pr-backup/$HEAD_REF-$STAMP"
+
+# The stamp has second resolution, so two runs in the same second collide. Find
+# a free name rather than failing on one — a retry loop would otherwise hit the
+# refusal below every single time.
+i=0
+while git -C "$WT" show-ref --verify --quiet "refs/heads/$BACKUP"; do
```
s-annam, line 730: **Nit.** TOCTOU against a concurrent run: two runs in the same second can both see `-1` free, and the loser's `git branch` then trips the `||` at L738, which prints "Fix the ref store and re-run" — a name race misdiagnosed as a broken ref store. Worth distinguishing exit 128 / "already exists" from a genuine ref-store failure, since the message is what a user acts on.

s-annam:
Adversarial review of the round-3 additions. **6 Blocking, 4 Nits.** Head reviewed: `f860dae`; all line numbers are against that commit.

Four of the six blockers sit in the ~230 lines the PR body flags as unreviewed (the cherry-pick path and the `--authored-*` plumbing), and three of them are the *same* defect the round-2 report identified and closed: **a guard that exists only as prose while the destructive command next to it is real code.** Round 2 fixed that for `MODE`; the pattern moved to the whitelist comparison and the backup ref.

`npm run verify` is green on this branch (build + `fallow` clean over 7 changed files).

## Blocking

### 1. `collapse-pr/SKILL.md:595` — row 1's whitelist comparison is a comment, and row 3's `reset --hard` then destroys the user's uncommitted edits

Line 595 is the entire enforcement of the whitelist inside `--authored-worktree`:

```
# every path emitted above must appear in $AUTHORED_PATHS, or this is row 2.
```

The `$WT != $AUTHORED_WORKTREE` case at :573 has a real `DIRTY_COUNT` check with a real `||` refusal. The case that matters more — dirt *inside* the caller's own checkout — has none.

Scenario: `pr-review` invokes `/collapse-pr 851 --yes --authored-worktree /Users/x/offlinecv --authored-path src/a.ts` on a partial-failure re-run. That checkout is on `$HEAD_REF` and holds two dirty tracked files: `src/a.ts` (the run's fix) and `src/b.ts` (the user's in-progress edit — `pr-review` Step 5.5 item 1 warns explicitly that `gh pr checkout` carries a dirty tree onto the PR branch). The :573 guard does not fire. Nothing compares the emitted path set to `$AUTHORED_PATHS`. `git add -- src/a.ts` + commit proceeds at :599-600, `src/b.ts` stays dirty, and row 3 at :751 runs `git -C "$WT" reset --hard "origin/$HEAD_REF"`.

`src/b.ts` is gone and **is not recoverable**: the backup at :738 is a branch ref, which captures commits, not a dirty worktree. The skill states this itself at :462 — *"it would take the staged index with them — which no branch ref can capture, so row 3 would not even be lossless."* That sentence is written about `inplace` mode, but it is equally true here whenever the whitelist check is skipped.

Fix: make it code, in the same fenced block as the `add`/`commit` —

```bash
git -C "$WT" status --porcelain --untracked-files=no -z | <the reader above> | sort -u > "$DIRTY_LIST"
printf '%s\n' ${AUTHORED_PATHS[@]+"${AUTHORED_PATHS[@]}"} | sort -u > "$WHITELIST"
UNKNOWN="$(comm -23 "$DIRTY_LIST" "$WHITELIST")"
[ -z "$UNKNOWN" ] || { echo "row 2: unknown provenance in $WT:"; echo "$UNKNOWN"; exit 1; }
```

### 2. `collapse-pr/SKILL.md:646` — the cherry-pick path's backup is a placeholder comment, and row 3 is then told not to mint one

```bash
FIX_SHA="$(git -C "$WT" rev-parse HEAD)"
# ... row 3's backup of $FIX_SHA runs here, with its own `||` refusal ...
git -C "$WORKDIR" cherry-pick "$FIX_SHA"
```

Then :658-661: *"Row 3 is already half-done for this `$WT`. The backup above covered the local-only commits **and** the fix commit on top of them, so when row 3 reaches this checkout it **must not mint a second ref** — only the `reset --hard` remains."*

Row 3's actual code at :738-751 unconditionally mints. So the two readings are (a) mint twice, harmless, or (b) follow :658-661, skip the mint, and reach :751's `git -C "$WT" reset --hard` **with no backup ref at all** — discarding local-only commits that this very gate's design note (:479-484) argues are mechanically indistinguishable from unique work.

This is the one reset in the skill whose precondition is not in the same fenced block as the reset. Every other one is chained. Fix: inline the `branch` + `rev-parse` verification into the `else` branch and have row 3 key off a carried `$BACKUP` variable, not off prose.

### 3. `collapse-pr/SKILL.md:647-648` — the cherry-pick path's push can never pass the managed `pre-push` hook, and the skill forbids the only bypass

```bash
git -C "$WORKDIR" cherry-pick "$FIX_SHA"
git -C "$WORKDIR" push origin "HEAD:$HEAD_REF"
```

That is a plain push, so `.git/hooks/pre-push` fires (worktrees share `.git/hooks` — Step 0b:198 says so). `$WORKDIR` was created by `git worktree add --detach` at :147 and has no `node_modules`, so `npm run verify` fails immediately — the exact failure Step 0b:200-202 describes as the reason the bypass exists at all.

The diagnosis table at :688 then gets this backwards:

> *"row 1's push is a plain push from a **real checkout**, so the managed hook fires and runs `npm run verify`. It is *supposed* to … **Never** set `OFFLINECV_SKIP_HOOKS=1` here"*

True of the `if` branch (:640, pushing from `$WT`). False of the `else` branch, which pushes from the throwaway worktree. Net effect on this repo: the cherry-pick path fails 100% of the time with a hook error whose documented remedy — "fix the failing gate" — is not a fixable condition.

Note the bypass is genuinely *not* licensed here, and that is the real problem: Step 0b's licence covers a tree the assertion proved unchanged, and a cherry-picked fix is new content. So the fix is not "add the bypass" — it is either to rebase-and-push from `$WT` (a real checkout, where the hook can pass), or to state plainly that this path requires a bootstrapped worktree. At minimum :688 must stop claiming this push comes from a real checkout.

### 4. `collapse-pr/SKILL.md:289` vs `pr-review/SKILL.md:561-566` — Step 2's `N == 1` exit precedes gate 3e, so the documented partial-failure recovery cannot fire

`pr-review` justifies `--authored-worktree`/`--authored-path` this way (:561-566):

> *"they are what makes a **re-run after a partial failure** work: if step 4 committed nothing, or committed and failed to push, the fixes are sitting uncommitted and 3e can only fold them in if it knows they are ours."*

In that state `origin/$HEAD_REF` holds the author's single commit — which `pr-review`:522-523 calls the common case (*"a PR that `open-pr` already left at one hand-authored commit"*). `/collapse-pr` Step 2 therefore counts `N == 1`, hits :289, runs Step 5b, and **exits 0 before Step 3 executes**. Gate 3e never runs, the uncommitted fix is never folded in or published, and the skill reports success. The `--authored-*` flags are inert on exactly the run they were built for.

Same ordering also makes `--dry-run`'s 3e classification (:556-560) unreachable on a single-commit branch.

Fix: run 3e's rows 1–2 before Step 2's early exit (they are about the *checkout*, not the commit count), or drop the recovery claim from `pr-review`.

### 5. `collapse-pr/SKILL.md:687` — a `reset --hard` on the user's real checkout with no backup, and a recovery instruction that cannot be carried out

The rejection-diagnosis table's second row prescribes:

> *"fetch, `reset --hard origin/$HEAD_REF` on `$WT`, re-apply the authored paths, retry once — do **not** refuse"*

At that point `$WT` holds the row-1 fix **commit** (:600). The reset discards the commit *and* the file contents, so "re-apply the authored paths" has no source left to re-apply from. No backup ref is taken on this path — a direct violation of :736-738 (*"The backup is a HARD PRECONDITION, not a courtesy — never let the reset run without it"*), which is the invariant this PR is built around.

Fix: don't reset. `git -C "$WT" fetch origin "$HEAD_REF" && git -C "$WT" rebase "origin/$HEAD_REF"` replays the fix commit onto the moved head and preserves it; if you keep the reset, take row 3's backup first.

### 6. `pr-review/SKILL.md:493` — step 5's collapse gate enumerates two of three abandon conditions, licensing a force-push on a run that pushed nothing

```
5. **Only if step 4 pushed.** If verification failed, or the branch moved mid-review, skip
   straight to step 6 …
```

Step 6 at :641 lists three: *"If verification fails, or the branch moved, **or the push is rejected**"*. The third is missing from the gate.

Scenario: `npm run verify` is green, `origin/$HEAD_BRANCH` is an ancestor of `HEAD`, and `git push origin HEAD` at :484 is rejected — branch protection, a permissions change, a transient remote failure. Both of step 5's tests say "proceed". `/collapse-pr` is invoked, resolves the head from `origin` (a head that does **not** contain the fix), collapses it and force-pushes, dismissing whatever approval the PR held. Then :601-604 re-reads `headRefOid` and the body reports `Addressed small fixes directly in <sha>`, and the "a fix that landed is not a finding" rule at :643 moves those items out of the inline set — so the findings disappear from the review while the code is unchanged.

The Rules bullet at :878 already states the correct condition (*"Collapse only on a run that actually pushed"*). Step 5 is the operative text; make it match.

## Nits

1. **`collapse-pr:573` / `pr-review:534` — `--authored-worktree` is compared as a raw string.** `pr-review` supplies `$(git rev-parse --show-toplevel)`; `$WT` comes from `git worktree list --porcelain`. They differ whenever a path component is a symlink (`/tmp` → `/private/tmp` on macOS is the everyday case), and the mismatch silently drops the caller's own checkout to row 2 and refuses. Fail-safe, but a baffling refusal. Normalize both with `pwd -P` / `realpath` before comparing.

2. **`pr-review:585-588` — the `cut -c4-` justification is backwards for the field it describes.** `git status --porcelain -z` emits a rename as `XY new\0old\0`; the `old` field carries **no** `XY ` prefix, so `cut -c4-` chops three real characters off a genuine path and it fails to match — it **under**-subtracts, *widening* the whitelist. That is the opposite of the stated *"errs by over-subtracting … that is the safe direction"*. `collapse-pr` row 1 (:586-594) already has the correct `-z` reader; reuse it rather than restating a safety argument that does not hold.

3. **`collapse-pr:510-513` — `git worktree list --porcelain` without `-z` is not newline-safe.** That command grew a `-z` flag for exactly this; without it a path containing a newline makes the `awk` attribute a dirty tree to the wrong worktree. Low likelihood, one-flag fix, and this loop feeds a `reset --hard`.

4. **`collapse-pr:524-530` — the row-1→4 iteration loop has an empty body (`:`)**, with the per-`$WT` sequencing left to the prose that follows. Given this PR's own doctrine that guards must be code, the loop deciding *which checkout* each destructive row operates on is the last place to leave as a placeholder.

## Description accuracy

The body's checkable claims hold: `npm run verify` green; all four in-scope frontmatters parse as strict YAML; only `collapse-pr` and the `CONTRIBUTING-PROCESS.md` rationale doc restate the merge-queue settings; `open-pr` and `revise-pr` carry zero `reset --soft`; the PR is exactly one commit; cross-referenced step and row numbers resolve across all five files. `pr-ready`'s frontmatter is untouched, as stated.

One claim to qualify. Round-2 finding #2 — *"gate 3e's `worktree`-only restriction was prose with **zero** `MODE` references in any of its five code blocks"* — is genuinely fixed: the `[ "$MODE" = worktree ]` guards are present in every block. But the same failure mode did not leave the gate, it moved: the whitelist comparison (#1) and the cherry-pick backup (#2) are now the prose halves of code-plus-prose pairs, and both sit upstream of a `reset --hard`. The body's *"Treat the sections above as the areas where direct inspection has been demonstrably unreliable"* is, on this evidence, still the right instruction.

## Verdict

Six blocking findings, so this would be `REQUEST_CHANGES`. Posting as a comment rather than a review: the approval on this PR is the maintainer's to give, and running the branch's own `/pr-review` auto-fix path against the code that introduces it is not a risk worth taking here.

---

Issue #607, opened 2026-07-27
Skill guidance suggests skills the résumé already states in other words (and renders canonical labels lowercase)


s-annam:
## Problem

`SkillTermGuidance` (the "Skills this role usually asks for" card on `/`) tells a candidate to add skills their résumé already states — in different words. On a leadership résumé whose skills row reads `Engineering Leadership, Technical Architecture & System Design, Distributed Systems & Cloud Computing, Team Building & Mentorship, Hiring & Talent Acquisition, Cross-functional Collaboration, Product Development & Strategy, Performance Optimization, AI / LLM Orchestration, Prompt Engineering & Evals`, the card renders:

> **Not in your résumé yet — add the ones you actually have:**
> `+ people management` `+ performance management` `+ project delivery` `+ agile / scrum leadership` `+ technical recruiting`
>
> **Already in your résumé:** cross-functional collaboration, Team Building & Mentorship

Two of those five suggestions are wrong, and the "already in your résumé" line is rendered in two different cases.

- **`+ technical recruiting`** — the résumé says **"Hiring & Talent Acquisition"**. Same competency, different words.
- **`+ people management`** — the résumé says **"Engineering Leadership"** and **"Team Building & Mentorship"**, and (in the roles this query is built from) describes leading a 50-person org. Same competency, different words.

This is the exact failure `term-quality.ts` already fixed once for `team-building` and `coaching-mentorship` (see the test at `src/lib/job-search/term-quality.test.ts:386`, *"never suggests a skill the query already says in different words"*) — the same root cause, on a pair the fix did not reach. The module docblock names the stake plainly: *"Offering someone `+ team building` while `Team Building & Mentorship` sits as a chip on the same screen … reads as a broken product."* It reads exactly that way here, and worse: the card's own heading is "add the ones you **actually have**", so a user who does have them is being told the product cannot see their résumé.

## Root cause

The suppression predicate is `anySkillCovers` → `skillCovers` in `src/lib/job-search/term-quality.ts:432`. It compares the **expected skill's canonical id, word-wise** against the query's skill phrases, via three clauses (same key / expected ⊆ concept / concept ⊆ expected sharing the head noun). It never consults **jd-match's alias table**, and it cannot: the module docblock at `src/lib/job-search/term-quality.ts:101` establishes a hard boundary — *"NO jd-match RUNTIME IMPORT"* — mirroring `role-profiles.ts`.

So the matching is purely lexical, and neither failing pair shares a word:

[...]

PR #853, opened 2026-08-16
fix(skills): author display labels for all unlabelled dictionary entries (#681)

### LINKED ISSUE #681: jd-match/skills.ts: 139 entries have no display label, so a recognized skill renders its lowercase id (sql, aws, ios)
## Problem

A recognized skill renders as its **canonical `id`** whenever its dictionary entry carries no `label`, so the UI shows `sql`, `javascript`, `aws`, `postgresql`, `nlp` — lowercase, un-branded — sitting beside title-cased labels and title-cased free-text chips. The `/jobs/` skills row reads:

> `People Management` `Machine Learning` `python` `sql` `Underwater Basket Weaving`

and `SkillTermGuidance`'s summary line on `/` reads "Already in your résumé: Machine Learning, python, sql".

This is the residual of #607. That issue fixed the 28 entries whose `label` was authored in lowercase ("people management" → "People Management") and its acceptance criterion — *every `label` starts with a capital or a digit* — is enforced by `skills.test.ts:44`. But the criterion only constrains entries that **have** a label. **139 of the 176 entries have none** and fall through `idToLabel`'s `?? id` to a lowercase kebab id. The card's casing is therefore still a function of how a dictionary entry happened to be written, which is exactly the leak #607 set out to close.

`iOS` is the sharpest case: `{ id: "ios", aliases: ["ios"] }` renders the mobile platform as **`ios`**.

## Root cause

`getSkillIndex()` builds `idToLabel` with an id fallback (`src/lib/jd-match/skills.ts:331`):

```ts
idToLabel.set(entry.id, entry.label ?? entry.id);
```

Ids are lowercase by construction (they are stable keys, matched case-folded), so the fallback is a *key* being used as *display text*. Every render site inherits it:

| Site | Line | What the user sees |
|---|---|---|
| `deriveSkills` — `/jobs/` chips + `JobQuerySummary` + `SkillTermGuidance`'s "Already in your résumé" | `src/lib/job-search/query-builder.ts:386` | `sql` chip beside `Machine Learning` chip |
| `missingTermLabel` — the `+ <skill>` suggestion pills | `src/components/features/TermQualityAdvisory.tsx:83` | `+ sql` |
| `extractSkillPass` — JD requirement terms on `/jd-fit/` and on the job result cards | `src/lib/jd-match/extract-jd-terms.ts:424` → rendered at `src/components/features/JdMatch.tsx:132` and `src/components/features/JobResultCard.tsx:134` | `graphql` in the requirements list |

Counted on `main` (dd6527c):

```bash
$ node -e "…"   # regex over the SKILLS array
total 176 | labelled 37 | unlabelled 139
```

The unlabelled 139 include every language, framework, datastore, cloud, and tool: `javascript, typescript, python, sql, html, css, react, next.js, vue, node.js, graphql, postgresql, mysql, mongodb, redis, aws, gcp, azure, kubernetes, docker, terraform, kafka, spark, pytorch, tensorflow, nlp, llm, rag, etl, dbt, ios, android, jest, vitest, jira, figma, oauth, jwt, saml, sso, soc2, gdpr, hipaa, …`

## Why not derive the casing at render time

This was tried during #607 and rejected — **do not re-propose it**. A title-caser produces `Sql`, `Aws`, `Jwt`, `Nlp`, `Ios`, `Postgresql`, which is worse than the lowercase it replaces. An "uppercase-anywhere means authored intent" variant does not help either, because the id fallback never contains an uppercase letter to begin with. Correct display casing here is **brand casing** — `SQL`, `PostgreSQL`, `Node.js`, `GraphQL`, `iOS`, `jQuery`, `dbt` — and no rule derives it from a lowercase key. It has to be authored, once, per entry.

## Scope

Author a `label` for the 139 entries that lack one, in `src/lib/jd-match/skills.ts`. Data only: no render site changes, no new helper, no matching change.

## What is safe, and what actually moves

Verified on `main` before scoping — **matching is case-insensitive on every path a display label reaches**, so this is a display-only change:

- `search.ts:161-162` lowercases each `query.skills` entry before building its match pattern.
- `normalizeSkillKey` (`role-profiles.ts`) lowercases and strips separators, so `term-quality.ts`'s verdicts/suppression and `query-builder.ts`'s `getCanonicalSkillKeys` are unaffected.
- `coverage.ts` matches a skill-sourced term by `term.id`, never by `display` (`src/lib/jd-match/coverage.ts:72`), and its phrase path lowercases with an `i` flag (`corpusMentionsPhrase`, `:143-148`).
- Aliases are untouched, so nothing about recognition changes.

Two things do move, and both belong in the PR body:

1. **Egress casing.** `providers/keywords.ts` sends `query.skills` entries verbatim — `primaryKeyword` becomes Jobicy's `tag=`, and `searchPhrase` becomes Remotive/Arbeitnow's `search=` when the résumé has no title. `tag=python` becomes `tag=Python`. No new field egresses and the privacy posture is unchanged (still the user-editable query terms, never résumé text), but confirm the feeds still return results for the changed case — a tag index that is case-sensitive would silently return fewer postings.
2. **Test expectations.** Any assertion on a *display* string moves; assertions on *ids* must not. `role-profiles.test.ts` and `coverage.test.ts` use ids (`"python"` as a `RoleProfile.skills` entry) and must stay lowercase. `query-builder.test.ts`, `extract-jd-terms.test.ts`, `term-quality.test.ts`, `deep-links.test.ts`, `JdMatch.test.ts`, `FindJobsLauncher.test.tsx`, and `useJobSearch.test.tsx` each carry display-string expectations to re-read one by one. Do not blanket find-and-replace: the same literal `"python"` is an id in one file and a display label in the next.

### 1. Author the 139 labels

In `src/lib/jd-match/skills.ts`, section by section, following the file's existing convention (`{ id: "power-bi", label: "Power BI", aliases: [...] }`). Use the project's own branded spelling, not a mechanical transform:

```ts
{ id: "javascript", label: "JavaScript", aliases: ["javascript", "js", "ecmascript"] },
{ id: "sql", label: "SQL", aliases: ["sql"] },
{ id: "node.js", label: "Node.js", aliases: [...] },
{ id: "postgresql", label: "PostgreSQL", aliases: [...] },
{ id: "ios", label: "iOS", aliases: ["ios"] },
{ id: "dbt", label: "dbt", aliases: ["dbt"] },   // lowercase is the brand
```

Leave `aliases` exactly as they are — they are matched lowercased and never shown, and `skills.test.ts` asserts that.

### 2. Fix the label-casing invariant so brand casing survives

`skills.test.ts:44` currently asserts every label starts with `[A-Z0-9]`. `iOS`, `dbt`, and `jQuery` are correct and would fail it. Replace the blanket rule with an explicit, commented allowlist so a genuinely-lowercase brand is a **deliberate** entry rather than an accident:

```ts
/** Brands whose own casing starts lowercase. Every other label must not. */
const LOWERCASE_BRAND_LABELS = new Set(["iOS", "dbt", "jQuery"]);
```

…and assert `LOWERCASE_BRAND_LABELS.has(label) || /^[A-Z0-9]/.test(label)`. Keep #607's failure message shape (it names the offending `id: "label"` pairs), because the next person to add an entry will meet this test before they meet this issue.

### 3. Add the "no entry falls back to its id" invariant

The point of the change is that the fallback stops being reachable for display. Pin it:

```ts
it("gives every entry an authored display label, so nothing renders its kebab id", () => {
  expect(SKILLS.filter((s) => !s.label).map((s) => s.id)).toEqual([]);
});
```

Keep the `?? entry.id` fallback in `getSkillIndex` — it is the total-function guard for a malformed entry, and this test is what makes it unreachable in practice.

### 4. Bump `SKILLS_DICTIONARY_VERSION`

`src/lib/jd-match/skills.ts:70` — 1.1 → 1.2, with a changelog line. Display text is part of this table's data. `TERM_QUALITY_VERSION` and `ROLE_PROFILES_VERSION` do **not** move: no rule and no profile changes, same call #607 made.

### 5. Re-read the affected test expectations

Run `npx vitest run` and work the failures one at a time, deciding per assertion whether the literal is an id (leave) or a display label (update). Add a line of intent where the update is not obvious.

## Acceptance criteria

- [ ] Every entry in `src/lib/jd-match/skills.ts` has a `label`; asserted by a test in `skills.test.ts`.
- [ ] Every label either starts with `[A-Z0-9]` or is in the documented lowercase-brand allowlist; asserted, with the offending entries named in the failure.
- [ ] No `aliases` array changed — the alias-lowercase test and the multi-word leadership-alias test still pass unmodified.
- [ ] `/jobs/` renders no lowercase kebab chip for a recognized skill: a résumé with `["SQL", "AWS", "iOS", "machine learning"]` produces chips `SQL, AWS, iOS, Machine Learning`. Asserted in `query-builder.test.ts`.
- [ ] The `+ <skill>` suggestion pills read the same way — a `missing` entry for `ci-cd`/`sql` renders `CI/CD`/`SQL`, asserted through `missingTermLabel`.
- [ ] Recognition is unchanged: `getSkillIndex().aliasToId` maps exactly the same alias set to the same ids as before the change (spot-checked in a test, e.g. `k8s → kubernetes`, `js → javascript`).
- [ ] `term-quality.ts` verdicts and suppressions are unchanged for a query whose chips only changed case — in particular the #594/#607 end-to-end suppression cases in `term-quality.test.ts` still pass untouched.
- [ ] `SKILLS_DICTIONARY_VERSION` is 1.2 and its pinning test is updated; `TERM_QUALITY_VERSION` stays 1.2 and `ROLE_PROFILES_VERSION` stays 1.1.
- [ ] No id literal was changed to satisfy a display assertion (`role-profiles.ts` skill lists and `RoleProfile.skills` expectations still read lowercase kebab ids).
- [ ] `npm run verify` is green (fallow findings are report-only — see `CLAUDE.md`).
- [ ] PR body states what the egress change is (`tag=python` → `tag=Python`) and that a feed still returns results for the new casing.

## Reuse analysis

**Capability:** choosing the display string for a canonical skill.

**Existing surfaces found:** `getSkillIndex().idToLabel` (`src/lib/jd-match/skills.ts:331`) is the single mapping; its three consumers are `query-builder.ts:386`, `TermQualityAdvisory.tsx:83`, and `extract-jd-terms.ts:424`.

**Decision:** **extend the existing data.** No new component, no new helper, no second mapping. #607 already established that a derived casing rule beside this table is the wrong shape — it cannot reproduce brand casing, and a second rule that disagrees with the authored labels is worse than the lowercase it replaces.

## Notes

- No PDF fixture is involved; nothing here touches `tests/fixtures/pdfs/`, so the fixture-PII rules do not apply.
- Good candidate for a newcomer: the work is bounded, the file is flat data, and the tests say when it is done. The judgement is per-entry brand casing, not architecture.

shubhransh-gupta:
## Summary

- Authors branded `label` values for all 139 previously-unlabelled entries in `jd-match/skills.ts`, so recognized skills render as `JavaScript`, `PostgreSQL`, `iOS`, etc. instead of lowercase kebab ids.
- Bumps `SKILLS_DICTIONARY_VERSION` to **1.2** and tightens `skills.test.ts`: explicit lowercase-brand allowlist (`iOS`, `dbt`, `jQuery`, `gRPC`, `scikit-learn`, `pandas`, `pytest`) plus a new invariant that every entry has an authored label.
- Updates display-string expectations in affected tests (`query-builder`, `extract-jd-terms`, `SkillTermGuidance`). Matching, aliases, and id-based assertions are unchanged.

## Egress note (from #681)

`query.skills` entries are sent verbatim to job feeds (`tag=`, `search=`). Casing changes from `python` → `Python`; matching remains case-insensitive on our side.

## Test plan

- [x] `npm run verify` passes locally (5875 tests)
- [ ] `/jobs/` skills row shows branded labels beside free-text chips
- [ ] `SkillTermGuidance` "Already in your résumé" line uses labels, not ids
- [ ] Job search feeds still return results for canonical skills after casing change

Fixes #681


shubhransh-gupta:
@s-annam — review ping for #681 👋

**CI is green** (`verify` + `fallow` succeeded). This is data-only in `jd-match/skills.ts`:

- Authored `label` for all 139 previously-unlabelled entries (brand casing: `JavaScript`, `PostgreSQL`, `iOS`, …)
- `SKILLS_DICTIONARY_VERSION` → **1.2**
- Test updates for display-string expectations; matching/aliases unchanged

Same pattern as the #607/#594 label work you merged before — happy to adjust any label spellings. Thanks!

src/lib/jd-match/skills.ts
```diff
@@ -62,188 +62,189 @@ export interface SkillEntry {
...
-  { id: "webgl", aliases: ["webgl"] },
-  { id: "three.js", aliases: ["three.js", "threejs"] },
-  { id: "d3", aliases: ["d3", "d3.js"] },
+  { id: "next.js", label: "Next.js", aliases: ["next.js", "nextjs", "next js"] },
+  { id: "vue", label: "Vue", aliases: ["vue", "vue.js", "vuejs"] },
+  { id: "nuxt", label: "Nuxt", aliases: ["nuxt", "nuxt.js", "nuxtjs"] },
+  { id: "angular", label: "Angular", aliases: ["angular", "angularjs", "angular.js"] },
+  { id: "svelte", label: "Svelte", aliases: ["svelte", "sveltekit"] },
+  { id: "redux", label: "Redux", aliases: ["redux"] },
+  { id: "tailwind", label: "Tailwind CSS", aliases: ["tailwind", "tailwindcss", "tailwind css"] },
```
s-annam, line 114: **Secondary 1 — this is one of six labels that changes the search *token*, not its case.**

`query.skills` is not only display text: `buildQueryTermPatterns` (`src/lib/job-search/search.ts:161-166`) lowercases each entry and compiles it into `(?<![a-z0-9])<term>(?![a-z0-9])` against the posting title + description. So `tailwind` → `Tailwind CSS` changes the admission pattern from the token `tailwind` to the **phrase** `tailwind css`, and a posting that says only "Tailwind" no longer matches on this term.

The other five (`cpp`→`C++`, `csharp`→`C#`, `d3`→`D3.js`, `huggingface`→`Hugging Face`, `soc2`→`SOC 2`) all move the other way and are improvements — `d3` in particular was 2 characters and failed `isSignificantSkillTerm`, so it admitted nothing at all before.

The label itself is right; `Tailwind CSS` is the brand. The ask is just that the PR body say the egress change is not only casing, since #681's Scope said "no matching change". A test on this specific entry would be the durable fix.

src/lib/jd-match/skills.test.ts
```diff
@@ -41,18 +41,30 @@ describe("skills dictionary", () => {
...
+    const offenders = SKILLS.filter(
+      (s) =>
+        s.label &&
+        !LOWERCASE_BRAND_LABELS.has(s.label) &&
+        !/^[A-Z0-9]/.test(s.label),
+    ).map((s) => `${s.id}: "${s.label}"`);
     expect(offenders).toEqual([]);
   });
 
+  it("gives every entry an authored display label, so nothing renders its kebab id (#681)", () => {
```
s-annam, line 64: **Secondary 2 — this invariant is why I am not blocking on #681's AC4/AC5, and it is the strongest thing in the PR.**

It is strictly stronger than the two example assertions the issue asked for: it proves the property for all 176 entries rather than the 4 named, and since `deriveSkills`, `missingTermLabel` (`TermQualityAdvisory.tsx:83`) and `extractSkillPass` all read the same `idToLabel` map, the class cannot regress while this is green.

What is still missing is a readable example at a render site — `iOS`, the issue's own headline case, is asserted nowhere, and the `missingTermLabel` calls in `JobQueryEditor.test.tsx:664,782` use the helper as their own expected value, so they pin no literal. Worth adding; not worth a round-trip.

src/lib/jd-match/skills.ts
```diff
@@ -62,188 +62,189 @@ export interface SkillEntry {
  *
  * Changelog:
  * - 1.0: initial curated dictionary, plus the #583 leadership/management block.
+ * - 1.2 (2026-08-16): authored display labels for all 139 previously-unlabelled entries (#681).
```
s-annam, line 65: **Nit — out of order.** The changelog above reads oldest-first (`1.0`, then `1.1`), so inserting `1.2` here makes it `1.0, 1.2, 1.1`. It belongs below the `1.1` block.

src/lib/jd-match/skills.ts
```diff
@@ -62,188 +62,189 @@ export interface SkillEntry {
...
-  { id: "d3", aliases: ["d3", "d3.js"] },
+  { id: "next.js", label: "Next.js", aliases: ["next.js", "nextjs", "next js"] },
+  { id: "vue", label: "Vue", aliases: ["vue", "vue.js", "vuejs"] },
+  { id: "nuxt", label: "Nuxt", aliases: ["nuxt", "nuxt.js", "nuxtjs"] },
+  { id: "angular", label: "Angular", aliases: ["angular", "angularjs", "angular.js"] },
+  { id: "svelte", label: "Svelte", aliases: ["svelte", "sveltekit"] },
+  { id: "redux", label: "Redux", aliases: ["redux"] },
+  { id: "tailwind", label: "Tailwind CSS", aliases: ["tailwind", "tailwindcss", "tailwind css"] },
+  { id: "sass", label: "Sass", aliases: ["sass", "scss"] },
+  { id: "webpack", label: "Webpack", aliases: ["webpack"] },
```
s-annam, line 116: **Nit — `webpack` styles itself lowercase.** webpack's own site and docs use lowercase deliberately, the same reason `dbt`, `pandas`, `pytest` and `scikit-learn` are already in your `LOWERCASE_BRAND_LABELS` allowlist.

Your call — it is defensible to sentence-case it — but if you take it, it needs the allowlist entry in `skills.test.ts` too, which is why I have not offered it as a one-click suggestion.

src/lib/job-search/query-builder.test.ts
```diff
@@ -394,30 +394,30 @@ describe("buildJobQuery", () => {
...
 
   it("preserves résumé order within the canonical and unrecognized tiers (stable sort)", () => {
     const parsed = baseParsed({
-      skills: ["go", "rust", "underwater basket weaving", "competitive juggling"],
+      skills: ["python", "java", "underwater basket weaving", "competitive juggling"],
     });
     const query = buildJobQuery(parsed);
-    // Canonical tier keeps its own relative order (go before rust)...
-    expect(query.skills.indexOf("go")).toBeLessThan(query.skills.indexOf("rust"));
+    // Canonical tier keeps its own relative order (Python before Java)...
```
s-annam, line 419: **Nit — worth a line of intent; this hunk is doing more than the body claims.**

The body describes the test changes as "display-string expectations", but this one swapped the *input* from `["go", "rust", ...]`. I checked whether that hid a failure — it does the reverse. `compileIndex` (`skills.ts:335-337`) registers only `aliases`, never the id, and `go`'s aliases are `["golang", "go lang"]`, so the literal `"go"` was never recognized: on `main` it fell through to `titleCase` as `"Go"`, `indexOf("go")` was `-1`, and `-1 < 0` passed for entirely the wrong reason. The swap makes the assertion real.

That is good, and a reader six months from now should not have to re-derive it:

```suggestion
    // Both inputs must be RECOGNIZED for this to test anything: `go`'s aliases
    // are ["golang", "go lang"] and `compileIndex` never registers the id, so
    // the literal "go" fell through to `titleCase` and `indexOf` returned -1 —
    // the old assertion passed vacuously. Canonical tier keeps résumé order...
```

s-annam (approved):
**Approving.** The data change is clean and I verified it structurally rather than by reading 139 lines: parsing both sides of the diff, the dictionary is **176 entries before and after, with 0 ids added or removed, 0 alias arrays changed, 0 pre-existing labels altered, 0 duplicate labels, and unlabelled 139 → 0.** That discharges #681's AC1/AC3/AC6 more convincingly than a spot-check would, and the brand casing itself is good — I read all 139 and found one debatable call (below). `npm run verify` is green on `2f3d3cc` on my machine.

Three Secondary findings and four Nits, none merge-blocking. **I could not push any of them** — this is a fork PR, so the reviewer auto-fix path is closed by design; everything is prose or a one-click suggestion below.

> **Overlap note, for the maintainer only:** #854 touches the identical five files for the same issue. I reviewed this PR strictly on its own merits and formed no view on which should land; flagging it only so the two are not merged independently.

## Secondary

**1. Six labels change the outbound *search token*, not just its case — the body's egress note under-describes it.** (inline)

The body says "Casing changes from `python` → `Python`; matching remains case-insensitive on our side." That is true for 133 of the 139. For six, the label is not a case-variant of the id:

| id | new label | outbound |
|---|---|---|
| `cpp` | `C++` | `C%2B%2B` |
| `csharp` | `C#` | `C%23` |
| `tailwind` | `Tailwind CSS` | `Tailwind%20CSS` |
| `d3` | `D3.js` | `D3.js` |
| `huggingface` | `Hugging Face` | `Hugging%20Face` |
| `soc2` | `SOC 2` | `SOC%202` |

Case-insensitivity does not cover a token whose *content* changed. `query.skills` is also consumed verbatim by the local admission filter — `buildQueryTermPatterns` (`src/lib/job-search/search.ts:161-166`) lowercases each skill and compiles it into `(?<![a-z0-9])<term>(?![a-z0-9])` against the posting text — so these six change **what postings are admitted**, which is the one thing #681's Scope section ruled out ("no matching change").

I traced each: `C++`, `C#`, `D3.js` (`d3` was 2 chars and failed `isSignificantSkillTerm`, so it admitted nothing before), `Hugging Face` and `SOC 2` all get **better**. The one plausible narrowing is `Tailwind CSS` — a posting saying only "Tailwind" no longer matches that term. Bounded, since `matchesQuery` is `.some()` across every query term.

Not blocking: mixed direction, mostly an improvement, and no user-visible break. But it is a real behaviour change outside the stated scope, described in the body as casing, and pinned by no test. Worth a line in the body and, ideally, a test on `Tailwind CSS`.

**2. Two of #681's acceptance criteria are unimplemented.** (inline)

- AC4 — "a résumé with `["SQL","AWS","iOS","machine learning"]` produces chips `SQL, AWS, iOS, Machine Learning`. Asserted in `query-builder.test.ts`." No such test was added; `iOS`, the issue's own sharpest case, is asserted nowhere.
- AC5 — "`missingTermLabel` renders `CI/CD`/`SQL`, asserted." Not added. The existing `JobQueryEditor.test.tsx` calls (`:664`, `:782`) use `missingTermLabel(...)` as their own expected value, so they pin no literal.

I am **not** calling these blocking, and the reasoning should be on the record so you can overrule it knowingly: the new `SKILLS.filter(s => !s.label)` invariant is strictly stronger than either example. It proves the property for all 176 entries, and every render site named in the issue (`deriveSkills`, `missingTermLabel` at `TermQualityAdvisory.tsx:83`, `extractSkillPass`) reads the same `idToLabel` map, so the class cannot regress while that test is green. What is lost is a readable example at the render site, not coverage.

**3. Two docblocks now assert an invariant this PR falsified.** (body-only — both sit on lines the diff does not touch, so there is nothing to anchor to.)

- `src/lib/jd-match/skills.ts:40` — "`MUST START WITH A CAPITAL` (or a digit) — `skills.test.ts` enforces it." Seven labels now do not, and the test carries an allowlist. This is the authoring contract for the very file this PR rewrote, and the issue explicitly wanted the next contributor to "meet this test before they meet this issue" — right now they meet a docblock that contradicts it. It should also record the new mandatory-label invariant (the `?` on `label?: string` now survives only as `getSkillIndex`'s total-function guard).
- `src/components/features/TermQualityAdvisory.tsx:78-80` — "That only holds because every authored `label` is title-cased (#607); a lowercase one would render `+ people management`." `+ iOS`, `+ dbt`, `+ pandas`, `+ pytest`, `+ gRPC`, `+ jQuery`, `+ scikit-learn` now render. The behaviour is correct; the stated rationale is not.

## Nits

**4. Changelog entry is out of order** — inserted between 1.0 and 1.1, so the list reads 1.0, 1.2, 1.1. (inline)

**5. The stable-sort test's *input* changed, and the reason deserves a line.** (inline, with a suggestion) The body says it "updates display-string expectations", but this hunk swapped `["go","rust"]` for `["python","java"]`. I checked whether that masked something, and it does the opposite — it **repairs a previously-vacuous assertion**. `compileIndex` (`skills.ts:335-337`) registers only `aliases`, never the id, and `go`'s aliases are `["golang","go lang"]` — so the literal `"go"` was never recognized. On `main` it fell through to `titleCase` → `"Go"`, making `indexOf("go")` equal `-1`, and `-1 < 0` passed for the wrong reason. Good change; it just reads like an unexplained edit.

**6. `Webpack`** — webpack's own docs deliberately style it lowercase, like `dbt`/`pandas`/`pytest` already in your allowlist. Author's call; it needs the allowlist entry too, so I have not made it a one-click suggestion. (inline)

**7. The unticked egress box — I discharged it for you.** #681's last AC wants the body to confirm a feed still returns results under the new casing. I measured all three live:

| feed | lowercase | new casing |
|---|---|---|
| Jobicy `tag=` | `python` → 100 | `Python` → 100 |
| Remotive `search=` | `postgresql` → 14 | `PostgreSQL` → 14 |
| Arbeitnow `search=` | `nodejs` → 175 | `NodeJS` → 175 |

All three are case-insensitive, so the casing half of the egress change is safe. Note this does **not** cover Secondary 1's six token changes, which are a different question.

## Gates

`verify` green (fallow clean on the diff). Fixture-PII (3a), design-system/reuse (3b), style tokens (3c) and skill/script command review (3e) **do not apply** — no fixtures, no `scripts/`, no `.claude/`, and the only component file touched is a test.

---

s-annam:
Approved — thanks, this is careful work. The dictionary audit holds up: 176 entries before and after, no id added or removed, no alias array touched, no pre-existing label altered, unlabelled 139 → 0.

Two things before I merge, both small:

**1. Changelog ordering.** The new `1.2` entry sits above `1.1` in the block at `skills.ts`. Move it below so the list stays chronological.

**2. The body describes the egress change as casing, and it isn't.** Six of the new labels are not case-variants of their id — `cpp` → `C++`, `csharp` → `C#`, `tailwind` → `Tailwind CSS`, `d3` → `D3.js`, `huggingface` → `Hugging Face`, `soc2` → `SOC 2`. That matters because `query.skills` is not display-only: `searchPhrase()` and `primaryKeyword()` in `src/lib/job-search/providers/keywords.ts:27,37` build the outbound keyword string from it, so the label *is* the token we send to Jobicy/Remotive/Arbeitnow.

I traced each of the six and I'm satisfied the change is net-positive — five widen or fix (`d3` previously failed `isSignificantSkillTerm` and admitted nothing), and only `Tailwind CSS` plausibly narrows. The three feeds also measured case-insensitive, so the casing-only labels are inert. So this is a **description fix, not a code fix**: please state the six non-casing labels and the egress effect in the PR body, which is what #681's AC 10 asked for.

One thing you should know, because it's a real find and it wasn't in your description: the `["go","rust"]` → `["python","java"]` swap at `query-builder.test.ts:416` looks like it masks a failure and is the opposite — `compileIndex` never registers ids as aliases, and `go`'s aliases are `["golang","go lang"]`, so `indexOf("go")` returned `-1` on `main` and the ordering assertion was passing vacuously. Your swap makes it a real assertion. I've filed the underlying gap separately; no action needed from you here beyond a line in the body if you want the credit recorded.

Push those two and I'll merge. Tag @s-annam if you disagree with any of it — the egress reading in particular is worth pushing back on if you read it differently.

Issue #834, opened 2026-08-14
[parser] skills — the final wrap of a comma-separated list never rejoins, shredding the last multi-word skill


s-annam:
## Context — what a `knownWrong` entry is

Each corpus fixture under `tests/fixtures/pdfs/` may carry a hand-authored ground-truth sidecar (`*.truth.json`, minted under #654). Its `knownWrong` block records, per field, a place where the parser disagrees with what the page actually draws, with a `status` of `open` (live bug, issue must be open), `accepted` (written-down tradeoff), or `unfiled` (measured, never filed — `issue: null`).

`unfiled` is capped at `UNFILED_TRUTH_CEILING = 10` (`src/lib/heuristics/corpus.test.ts:186`) and is saturated at 10/10 today. This issue files one of them.

## What's wrong

When a comma-separated skills list wraps, the **last** wrap is never rejoined, so the skill straddling that break is split into two junk tokens.

`tests/fixtures/pdfs/unknown/single-word-name-mononym.pdf` draws (`pdftotext -layout`):

```
SKILLS
Python, Go, TypeScript, PostgreSQL, Kafka, Kubernetes, Terraform, AWS, gRPC, Distributed
Systems
```

Expected `skills` (from the fixture's truth sidecar): `Python, Go, TypeScript, PostgreSQL, Kafka, Kubernetes, Terraform, AWS, gRPC, Distributed Systems` — ten skills.

Observed: eleven, ending `…, gRPC, Distributed, Systems`. `Distributed Systems` is destroyed and **neither half is a real skill**.

Recorded today as a `status: "unfiled"` `knownWrong` entry on `skills` in `tests/fixtures/pdfs/unknown/single-word-name-mononym.truth.json`.

This is common, not exotic: any comma-delimited skills row long enough to wrap ends its final visual line with a fragment, and a multi-word skill in that position always splits.

## Not a duplicate of #301

#301 (closed, completed 2026-07-03) fixed the *round-trip* direction — a multi-word skill split at the boundary our own exporter's word-wrap introduced, `parse → render → re-parse`. This is the **source-parse** direction: the split happens on the user's original PDF, before any export, and #301's fix does not reach it.

## Root cause

`src/lib/heuristics/extract/skills.ts:388-500`, `isSoftWrapContinuation`. The function decides whether a pending line and the next line are one wrapped list. It has four join conditions:

| Condition | Fires when | Line |
|---|---|---|
[...]

PR #854, opened 2026-08-16
fix(skills): author display labels for every dictionary entry

### LINKED ISSUE #681: jd-match/skills.ts: 139 entries have no display label, so a recognized skill renders its lowercase id (sql, aws, ios)

qtjg:
# PR title

fix(skills): author display labels for every dictionary entry

# PR body

Resolves #681

## Summary

This change authors an explicit display label for every entry in the curated skill dictionary. Previously, 139 entries fell back to their lowercase canonical IDs, causing recognized skills such as `sql`, `postgresql`, `javascript`, and `ios` to render with inconsistent casing beside authored and free-text labels.

The dictionary now uses project-consistent brand casing, including `SQL`, `PostgreSQL`, `JavaScript`, `Node.js`, `GraphQL`, `iOS`, `jQuery`, `dbt`, and `gRPC`. Aliases and canonical IDs are unchanged, so matching and deduplication behavior remain stable. The fallback in `getSkillIndex()` remains as a defensive total-function guard, while a new test makes the authored-label invariant explicit.

## Changes

- Added authored labels to all 139 previously unlabeled dictionary entries.
- Bumped `SKILLS_DICTIONARY_VERSION` from `1.1` to `1.2` and documented the data change.
- Replaced the blanket label-casing assertion with an explicit lowercase-brand allowlist for `iOS`, `dbt`, `jQuery`, and `gRPC`.
- Added a regression test proving that no dictionary entry falls back to its ID for display.
- Updated affected display expectations in JD extraction, job-query construction, and `SkillTermGuidance` tests. Canonical skill IDs remain lowercase where the test is asserting identity rather than display text.

## Verification

- `npm run verify` passes: typecheck, lint, fixture-PII checks, core-package checks, 42 changed test files with 713 passing tests and 2 intentionally skipped tests, production build, and Fallow audit.
- `npm test -- --run src/lib/job-search/query-builder.test.ts src/components/features/SkillTermGuidance.test.tsx src/lib/jd-match/extract-jd-terms.test.ts src/lib/jd-match/skills.test.ts` passes: 109 tests.
- The verification command reports existing non-blocking baseline warnings for nine unfiled fixture disagreements and skips the GitHub issue-state check because the local `gh` token is unavailable.


src/lib/jd-match/skills.ts
```diff
@@ -66,184 +67,186 @@ export interface SkillEntry {
...
+  { id: "docker", label: "Docker", aliases: ["docker"] },
+  { id: "terraform", label: "Terraform", aliases: ["terraform"] },
+  { id: "ansible", label: "Ansible", aliases: ["ansible"] },
+  { id: "helm", label: "Helm", aliases: ["helm"] },
+  { id: "linux", label: "Linux", aliases: ["linux"] },
+  { id: "nginx", label: "Nginx", aliases: ["nginx"] },
+  { id: "kafka", label: "Kafka", aliases: ["kafka", "apache kafka"] },
+  { id: "rabbitmq", label: "RabbitMQ", aliases: ["rabbitmq", "rabbit mq"] },
+  { id: "airflow", label: "Apache Airflow", aliases: ["airflow", "apache airflow"] },
+  { id: "spark", label: "Apache Spark", aliases: ["spark", "apache spark"] },
```
s-annam, line 171: **Blocking (finding 1), concrete case.** `Apache Spark` is the right *display* string, but it is also the string that leaves the browser: `primaryKeyword` returns `query.skills[0]` verbatim and Jobicy receives it as `tag=`. A data-engineering résumé whose first canonical skill is Spark now requests `tag=Apache%20Spark` where it previously requested `tag=spark`.

Same applies to `airflow` on the line above. Either narrow both labels to `Spark` / `Airflow` — the display defect #681 is about is fixed either way — or keep them and add the egress paragraph AC 10 asks for, confirming a feed still returns postings for the new tag.

src/lib/jd-match/skills.ts
```diff
@@ -66,184 +67,186 @@ export interface SkillEntry {
...
+  { id: "express", label: "Express", aliases: ["express", "express.js", "expressjs"] },
+  { id: "nestjs", label: "NestJS", aliases: ["nestjs", "nest.js"] },
+  { id: "django", label: "Django", aliases: ["django"] },
+  { id: "flask", label: "Flask", aliases: ["flask"] },
+  { id: "fastapi", label: "FastAPI", aliases: ["fastapi", "fast api"] },
+  { id: "spring", label: "Spring", aliases: ["spring", "spring boot", "springboot"] },
+  { id: "laravel", label: "Laravel", aliases: ["laravel"] },
+  { id: "grpc", label: "gRPC", aliases: ["grpc"] },
+  { id: "rest", label: "REST", aliases: ["rest", "rest api", "restful api", "restful"] },
+  { id: "websocket", label: "WebSockets", aliases: ["websocket", "websockets"] },
```
s-annam, line 139: **Nit (5).** `WebSockets` pluralizes relative to the `websocket` id, unlike every other entry in the table. `WebSocket` matches both the id and the brand — and, per finding 1, also preserves the outbound `tag=` token.

src/lib/jd-match/skills.test.ts
```diff
@@ -42,17 +42,22 @@ describe("skills dictionary", () => {
...
-    // itself, so a lowercase one puts two casings in one sentence — and the
-    // recognized skill is the one that looks unpolished. Checked on the first
-    // character only: inside the label the entry's own house casing rules
-    // ("iOS", "P&L Ownership", "A/B Testing"), which no rule can derive.
-    const offenders = SKILLS.filter((s) => s.label && !/^[A-Z0-9]/.test(s.label)).map(
-      (s) => `${s.id}: "${s.label}"`,
-    );
+    // These brands intentionally start lowercase; every other label must make
+    // its display casing explicit rather than inheriting a stable ID.
+    const LOWERCASE_BRAND_LABELS = new Set(["iOS", "dbt", "jQuery", "gRPC"]);
```
s-annam, line 47: **Nit (4).** #681 asked for this as a documented module-level const: the next person adding a dictionary entry meets this test before they meet the issue, and a set buried in the assertion body is easy to miss.

Separately, it keys on the **label** rather than the `id`, so the exemption does not survive a label edit — rename `"jQuery"` and its allowlist entry silently stops matching. Keying on `id` makes the exemption stable, and reads as "this entry is allowed to start lowercase" rather than "this exact string is".

src/lib/job-search/query-builder.test.ts
```diff
@@ -394,30 +394,30 @@ describe("buildJobQuery", () => {
...
+        "TensorFlow",
+        "NLP",
       ]),
     );
   });
 
   it("preserves résumé order within the canonical and unrecognized tiers (stable sort)", () => {
     const parsed = baseParsed({
-      skills: ["go", "rust", "underwater basket weaving", "competitive juggling"],
+      skills: ["golang", "rust", "underwater basket weaving", "competitive juggling"],
```
s-annam, line 416: **Secondary (3).** This is the one input change in the PR, and it is the right one — but it deserves a sentence saying so.

`"go"` is not an alias of the `go` entry (its aliases are `golang` / `go lang`), so on `main` it landed in the *unrecognized* tier and `query.skills.indexOf("go")` returned `-1`. The assertion below therefore read `-1 < indexOf("rust")` and passed for the wrong reason. Switching to `"golang"` makes it a genuine canonical-tier ordering check.

#681 step 5 asked for a line of intent where a test update is not obvious; this is that line.

s-annam (changes requested):
The dictionary work itself is careful and correct — I verified all 176 entries independently and the data invariants hold. One blocking item: this PR changes what string goes out to three third-party job feeds for 9 skills, and neither the body nor a test says so.

**What I verified mechanically** (parsed both revisions of `SKILLS` and diffed them, rather than reading the diff):

- 176 entries before and after, id set byte-identical, **zero alias arrays changed**, and no pre-existing label altered. So recognition and de-duplication are provably unchanged — AC 3 and AC 6 are met in fact.
- 0 entries left without a label (AC 1), `getSkillIndex`'s `?? entry.id` guard retained, `SKILLS_DICTIONARY_VERSION` at `1.2` with `TERM_QUALITY_VERSION`/`ROLE_PROFILES_VERSION` untouched (AC 8).
- Brand casing reads well throughout. `iOS`, `dbt`, `jQuery`, `gRPC`, `PostgreSQL`, `Node.js`, `C#` are all right, and catching `gRPC` — which #681's plan did not list — was a good call.

## Blocking

**1. Nine labels change the outbound feed keyword, not just its casing — undisclosed and unverified.**

#681's AC 10 is explicit: *"PR body states what the egress change is (`tag=python` → `tag=Python`) and that a feed still returns results for the new casing."* The body doesn't mention egress at all, and the actual delta is wider than the issue anticipated. `query.skills[0]` goes out verbatim as Jobicy's `tag=`, and `query.skills.slice(0,3).join(" ")` becomes Remotive/Arbeitnow's `search=` on a title-less résumé (`src/lib/job-search/providers/keywords.ts`). Of the 139 newly-authored labels, 131 differ from the old id-display by case only — exactly what the issue scoped. These 9 change the **token**:

| id | old `tag=` | new `tag=` |
|---|---|---|
| `spark` | `spark` | `Apache Spark` |
| `airflow` | `airflow` | `Apache Airflow` |
| `tailwind` | `tailwind` | `Tailwind CSS` |
| `websocket` | `websocket` | `WebSockets` |
| `huggingface` | `huggingface` | `Hugging Face` |
| `d3` | `d3` | `D3.js` |
| `soc2` | `soc2` | `SOC 2` |
| `cpp` | `cpp` | `C++` |
| `csharp` | `csharp` | `C#` |

A case-insensitive tag index shrugs at `Python`; a one-word→two-word change is a different question. A data-engineering résumé whose first canonical skill is Spark now asks Jobicy for `tag=Apache%20Spark`, and if that tag doesn't exist the panel silently returns fewer postings with nothing on screen saying the query got narrower. No test can catch this — which is why the issue made it a body-plus-manual-check AC rather than a test AC.

To clear it, either:

- **Narrow the 5 labels that gain a word** (`Spark`, `Airflow`, `Tailwind`, `WebSocket`, `Hugging Face`) so the token is preserved, keeping the display fix intact; or
- **Keep them** and add the egress paragraph #681 asked for, with a note that you hit one feed for one changed tag and got results back.

Either is fine — the blocker is that the change is currently invisible in the record. I checked the URL construction and there is no encoding bug: all three adapters use `encodeURIComponent`, so `C#`/`C++` go out as `C%23`/`C%2B%2B` correctly. This is about which postings come back, not a malformed request.

## Secondary

**2. Two of #681's named test ACs are unmet.**

- AC 4 asks for a `query-builder.test.ts` assertion that `["SQL", "AWS", "iOS", "machine learning"]` produces `SQL, AWS, iOS, Machine Learning`. The existing expectations were updated, but that assertion wasn't added — so `iOS`, the issue's sharpest example, has no test.
- AC 5 asks that the `+ <skill>` pills be pinned through `missingTermLabel` for a newly-labelled id like `sql`. `JobQueryEditor.test.ts:662` covers `ci-cd`, which already had a label on `main`; nothing covers the 139 this PR adds.

**3. A test input changed, not just a display expectation** — `query-builder.test.ts:416`, `"go"` → `"golang"`. This is right, and it's more interesting than it looks: `"go"` is not an alias of the `go` entry (aliases are `golang`/`go lang`), so on `main` it fell to the unrecognized tier and `indexOf("go")` returned `-1`, making `-1 < indexOf("rust")` pass for the wrong reason. Your change turns a vacuous assertion into a real one. #681 step 5 asked for a line of intent where an update isn't obvious — this is the one place that qualifies, and it's worth a sentence so the next reader doesn't mistake it for a find-and-replace.

## Nits

**4. `LOWERCASE_BRAND_LABELS` is scoped inside the `it()` body** (`skills.test.ts:47`). #681 wanted a documented module-level const, on the reasoning that the next person adding an entry meets this test before they meet the issue. It also keys on the **label** string rather than the `id`, so a typo'd label (`"iOs"`) silently escapes the casing rule instead of failing it; keying on `id` would close that.

**5. `websocket` → `WebSockets`** pluralizes relative to its id, unlike every other entry. `WebSocket` matches the id and the brand.

**6. The body still carries its template scaffolding** — the literal `# PR title` and `# PR body` headings, with the title duplicated inside the body. Everything under them is substantive, so this is cosmetic, but the merge-queue record keeps it.

## Gates

`npm run typecheck` clean · `npx eslint src` clean · full suite **359 files / 5865 tests green** (nothing beyond the 4 test files you touched needed updating — the other files #681 listed as at-risk pass unmodified) · `fallow audit` no issues in 5 changed files. `check:fixtures` and `check:baselines` skipped as inapplicable — no fixture or sidecar is touched.

## Overlap with #853

Worth flagging so nobody double-merges: **#853** touches the identical five files against the same issue and was opened first. This is a maintainer's pick, not a defect in either PR, and no comparison here should be read as ranking them. One factual difference relevant to finding 1: #853 labels those entries `Spark`, `Airflow`, `WebSocket` — single-token, so the feed keyword is preserved — while matching this PR on `Tailwind CSS`, `D3.js`, `Hugging Face`, `SOC 2`, `C++`, `C#`.

Nothing was pushed to your branch: with a blocking finding open, this review commits nothing, so `manus/fix-skill-display-labels` is exactly as you left it at `171095c`.

---

s-annam:
Welcome, @qtjg — this is your first PR here, so a few words on where it stands and how our review loop works.

**What stands:** the dictionary work is solid. All 176 entries labelled, aliases untouched, version bumped correctly, and the whole suite green. I checked the data independently rather than trusting the diff, and the invariants hold. Catching `gRPC` for the lowercase-brand allowlist — which the issue didn't list — was a nice touch.

**What's needed:** one blocking item, and it's a small edit, not a rewrite. Nine of your labels change the *token* sent to the job feeds, not just its casing (`tag=spark` → `tag=Apache Spark`). Either shorten those five multi-word ones, or keep them and add the egress note #681 asks for. Everything else is optional polish.

**Our review policy**, since it's now relevant to you: when a review finds **zero** blockers, we apply the small non-behavioural nits ourselves, push them, and approve in the same pass — so a clean PR doesn't cost you a round-trip. When there **are** blockers, we touch nothing: your branch is exactly as you left it at `171095c`, and the next push is yours. That's why the nits above are written out rather than already fixed.

**One thing to know:** #853 landed against the same issue and touches the same five files. That's on us — #681 went out without an assignee and two people picked it up. It's a maintainer's call which one merges, and it isn't a judgement on either PR. Your work here is real either way, and I'd genuinely like to see you pick up another one.

Questions or pushback on any finding — including the blocking one — tag @s-annam. Disagreement is welcome; we'd rather argue it out than have you silently apply something you think is wrong.

s-annam:
Mayank — thank you for this, and I'm sorry it's landing as a close. The work is good; the reason is scheduling on our side, not quality.

> **Correction (edited).** An earlier version of this comment said #681 was unclaimed. That was wrong, and I'd rather fix it in place than leave it standing. @rohithgollapalli self-assigned #681 on 2026-07-30 and it has carried that assignment for seventeen days. I checked properly only after posting.

**What happened.** #681 was assigned — and you and @shubhransh-gupta both built it anyway. That makes this a worse failure on our side than the one I first described, not a lesser one: an assignment that has been sitting visibly on an issue for seventeen days did not reach either of you at the moment you picked the work. #853 was opened first and is now approved, so that's the one I'm merging. Sorting out what that means for @rohithgollapalli's assignment is our problem to resolve, not yours.

**Where the two diverged, since you should see the reasoning rather than just the outcome.** Your PRs are 95% identical — 27 lines apart. The split is brand casing: you capitalized `scikit-learn` → `Scikit-learn`, `pandas` → `Pandas`, `pytest` → `Pytest`, which conforms to the existing `#607` rule at `skills.test.ts:44` that every label start `[A-Z0-9]`; #853 instead widened the lowercase-brand allowlist to keep each project's own branding. Both are defensible readings and I went with brand fidelity, on the grounds that `Pandas` is the same class of error as `Jquery` — the four exemptions you kept (`iOS`, `dbt`, `jQuery`, `gRPC`) already prove the capital rule can't hold absolutely.

**The one thing that was a genuine blocker**, and it's worth your time because it's a non-obvious property of this codebase: `query.skills` is not display-only. `searchPhrase()` and `primaryKeyword()` in `src/lib/job-search/providers/keywords.ts:27,37` build the outbound keyword string from it, so a label *is* the token we send to Jobicy/Remotive/Arbeitnow. Your `spark` → `Apache Spark` and `airflow` → `Apache Airflow` change `tag=spark` into `tag=Apache Spark`. On `main` those long forms are *aliases* (`skills.ts:167-168`) — matched inbound, deliberately never shown — and promoting one to the display label pushes it outbound. The feeds measured case-insensitive, so casing-only labels are inert, but that measurement doesn't cover a different word, and a tag query for the long form plausibly returns strictly fewer postings.

I had actually planned to graft your `Apache Airflow` / `Apache Spark` into #853 before the review caught this. So your version surfaced a real constraint that the merged PR now documents. That's a contribution even though the branch isn't merging.

**What we are taking from your PR:** the changelog ordering fix (you put `1.2` after `1.1`; #853 had it above), and your `["go","rust"]` test-input change at `query-builder.test.ts:416` independently confirmed a latent bug on `main` — `go` was never an alias of the `go` entry, so `indexOf("go")` returned `-1` and that ordering assertion was passing vacuously. Filed as its own issue with both of you credited.

**If you want another,** #834 is unassigned and yours if you say so — the final wrap of a comma-separated skills list never rejoins, so the last multi-word skill gets shredded. Same lane, well-bounded, and there's a fixture for it. Comment on it and I'll assign it to you before you start.

Push back on any of this — tag @s-annam. Particularly the casing call: if you think #607's consistency rationale should have won, say so, because that's a judgment I made and not a rule I found.

s-annam:
One more thing, @qtjg, because I got the handling wrong even where the decision was right.

I posted a review telling you what to fix, then closed this eleven minutes later. You never had a chance to read it, let alone answer. Whatever the merits of picking #853, that sequence was mine and it was careless — a review that says "here's the next step" and a close in the same quarter hour are two messages that contradict each other.

**#834 is held for you** — I've put a note on it saying so, and asked that nobody else take it. Comment there and I'll assign it properly; GitHub won't let me assign someone who hasn't commented on the issue yet, which is the only reason it's a note and not an assignment. If it's not the one you want, say so and take anything unclaimed — per our contributing guide an assignment from us is a request, never an obligation.

PR #858, opened 2026-08-16
feat: [edit] A parsed entry cannot be deleted from the reconstructed résumé — only user-ADDED ones (achievement, role, education, project) (#856)

### LINKED ISSUE #856: [edit] A parsed entry cannot be deleted from the reconstructed résumé — only user-ADDED ones (achievement, role, education, project)
## Problem

In the reconstructed résumé on `/`, a **parsed** entry cannot be deleted. Only a **user-added** entry can. If the parser produces a phantom achievement, a duplicated role, or an education entry stitched out of two, the user can edit its text but cannot remove it — and it ships into the Download PDF.

Reported against the Achievements section ("I can't delete an achievement once shown from a parsed resume"), but it is not achievements-specific. Every section gates its remove affordance on the same `added` flag:

| Section | Remove affordance | Gate |
|---|---|---|
| Achievements | `RemoveButton label="Remove achievement"` | `src/components/features/ReconstructedResume.tsx:971` — `{added && …}` |
| Projects | `RemoveButton label="Remove project"` | `src/components/features/ReconstructedResume.tsx:755` — `{added && …}` |
| Education | `RemoveButton label="Remove education"` | `src/components/features/ReconstructedEducationSkills.tsx:163` — `onRemove` only passed when `added` (`:258`) |
| Experience | `RemoveButton label="Remove role"` | `src/components/features/ReconstructedRole.tsx:432` — `onRemove` only passed when `added` (`ReconstructedResume.tsx:651`) |

The root cause is one level down: **no removed-parsed-entry state exists.** `removeEntry` in `src/hooks/useEditableParse.ts:1349` only filters the added-entry list:

```ts
const removeEntry = useCallback(
  (id: string) => {
    setAddedEntries((prev) => prev.filter((e) => e.id !== id));
    ...
  },
  [writeAddedBullets],
);
```

There is no equivalent of `removedBullets` (`useEditableParse.ts:1078`, folded into the résumé by `applyRemovedBulletOverrides` in `src/lib/edit/apply-overrides.ts:782`) at the **entry** level. `applyOverrides` (`apply-overrides.ts:1027`) takes no such parameter.

Achievements feel worse than the other sections because achievement (and project) bullets are rendered **read-only** — `ReconstructedResume.tsx:981` and `:764` pass only `bullet={b}` to `ResumeBulletRow`, no `onBulletChange` and no `onRemove`. In Experience a user can at least delete the bullets under a bad role; in Achievements nothing under the entry is removable either. That bullet gap is **out of scope here** (see "Not in scope").

Epic #453 brought Achievements to "full edit fidelity" (inline type/description/year via #454, structured add via #455). Deletion was never in its acceptance criteria.

## Why it matters

- The Download PDF is built from the canonical fields (`apply-overrides.ts:1102` → `toCanonicalResume` → `src/lib/pdf/ats-resume-model.ts`), so an un-deletable phantom entry is exported verbatim.
- The score reads `input.sections` for its bullet pool (`src/lib/score/score.ts:986-989`), so a phantom entry's bullets keep grading the résumé.
- The only workaround today is blanking every field, which leaves an empty-but-present entry. `pruneEmptyAddedEntries` (`useEditableParse.ts:1361`) only prunes **added** entries, so a blanked parsed entry is never dropped.

## Proposed design — a `removedEntries` tombstone set

Mirror the existing `removedBullets` shape, one level up. Two decisions carry the design:

**1. Tombstone, do not splice.** Achievement, experience, and education overrides are keyed by **parsed array index** (`applyAchievementOverrides`, `apply-overrides.ts:391-405`; `applyExperienceHeaderOverrides`; `applyEducationFieldOverrides`). Splicing an entry out of `parsed.heuristic_achievements` shifts every later index and silently rebinds each later entry's edits to the wrong entry. Instead: keep the array intact through the override passes, mark the removed entries, and filter them out **once, at the end** of `applyOverrides` — after every index-keyed pass has run.

**2. Reuse `removedBullets` for the entry's bullets.** The UI's delete handler removes the entry **and** calls the existing `removeBullet(id)` for every bullet id in the entry's group. That reuses `removeBulletFromRawText` / `removeBulletFromSections` / `removeBulletFromDescriptions` (`apply-overrides.ts:654-690`) unchanged, so the bullets leave `rawText`, `sections` (hence the score) and the role descriptions with no new machinery. The only genuinely new text surgery is the entry's own **header line**.

Key: `parsedEntryKey(section, index)` — `"achievements:2"` — already exported from `useEditableParse.ts:372` and already the key space `addedBullets` uses for parsed entries. Added-entry ids (`"added:<n>"`) live in the same namespace without collision (`isAddedEntryKey`, `:386`), so one `removedEntries` set can hold both and `removeEntry` can serve both paths.

### Step 1 — `removedEntries` state in `useEditableParse.ts`

- Add `const [removedEntries, setRemovedEntries] = useState<ReadonlySet<string>>(new Set())` alongside `removedBullets` (`:851`), plus a `removedEntriesRef` mirror matching the `removedBulletsRef` pattern (`:898`).
- Widen `removeEntry` (`:1349`) to handle both key kinds:

```ts
const removeEntry = useCallback(
  (key: string) => {
    if (isAddedEntryKey(key)) {
      setAddedEntries((prev) => prev.filter((e) => e.id !== key));
      // existing addedBullets cleanup
      return;
    }
    setRemovedEntries((prev) => new Set(prev).add(key));
  },
  [writeAddedBullets],
);
```

- Export `removedEntries` from the hook's return value (`:1792`+) and add it to `hasEdits` (`:1764`): `if (removedEntries.size > 0) return true;`.
- Add it to the `reset` callback (`:1570`+).

### Step 2 — undo / draft snapshot

- Add `removedEntries: string[]` to `EditSnapshot` and to the `snapshot` memo (`:1587-1616`), serialized like `removedBullets: [...removedBullets]` (`:1593`).
- In `replay` (`:1618`), restore with a **backward-compatible default** — `(snap.removedEntries ?? []).forEach((key) => removeEntry(key))` — following the `snap.descriptionOverrides ?? {}` precedent at `:1665`. `EditSnapshot` is persisted to the résumé library, so a draft written before this change has no such field; see #815 for what a snapshot that replays wrong costs.

### Step 3 — `applyOverrides` fold

- Add an 18th positional parameter `removedEntries: ReadonlySet<string> = new Set()` to `applyOverrides` (`apply-overrides.ts:1027`), matching the existing convention (`removedBullets` is already positional at `:1039`). The parameter count is pre-existing debt — do **not** refactor the signature to an options object in this issue.
- After `applyAddedEntriesAndBullets` (`:1094`) and before `toCanonicalResume` (`:1102`), add a single `applyRemovedEntries(nextParsed, views, removedEntries)` pass that, for each key `"<section>:<index>"`:
  - filters the entry out of the matching array (`experience`, `education`, `projects`, `heuristic_achievements`);
  - removes the entry's **header line** from `views.rawText` and `views.sections`, reusing the normalise-and-match-first-line helper the bullet removers share (`apply-overrides.ts:489-500`, `:649-690`) rather than writing a second matcher.
- Ordering is load-bearing: it must run **after** every index-keyed override pass (`applyExperienceHeaderOverrides` `:1069`, `applyEducationFieldOverrides` `:1083`, `applyAchievementOverrides` `:1086`, `applyDescriptionOverrides` `:1091`) so no key is rebound, and after the added-entry append (`:1094`) so the parsed indices those passes used are still the ones in the array.
- Wire the new argument through the one caller: `src/hooks/useAnalyzedResume.ts:162-196` (argument list **and** the `useMemo` dep array — `exhaustive-deps` is not enforced in this repo, so a missing dep lints green and produces a stale résumé).

### Step 4 — UI

- Drop the `added &&` gate at `ReconstructedResume.tsx:971` (achievements), `:755` (projects), `:651` (experience) and `ReconstructedEducationSkills.tsx:258` (education). Pass `onRemove` unconditionally, resolving the key: `added ? added.id : parsedEntryKey(<section>, i)`.
- The delete handler also drops the entry's bullets: `group?.bullets.forEach((b) => onRemoveBullet(b.id))` before `onRemoveEntry(key)`. Achievements/projects currently receive no `onRemoveBullet` prop, so thread it in from `ReconstructedResume` (the same `removeBullet` already passed to the Experience path at `:646`).
- Achievements is the section the report came from — ship it first and verify end-to-end before doing the other three in the same PR.

### Step 5 — tests

- New `src/hooks/useEditableParse.removed-entry.test.tsx`: removing a parsed entry sets `hasEdits`, survives a snapshot → `replay` round-trip, and a **pre-change snapshot** (no `removedEntries` field) replays without throwing.
- New `src/lib/edit/apply-overrides.removed-entry.test.ts`: **the index-rebinding regression** — override achievement index 2's title, remove achievement index 0, assert the surviving entry at old-index-2 still carries its own override and not a neighbour's. Same assertion for experience and education.
- Extend `src/components/features/AchievementTypePicker.test.tsx`'s neighbours (or a new `ReconstructedResume.remove-parsed-entry.test.tsx`) for the rendered affordance.
- Round-trip: assert a résumé with an entry removed exports and re-parses without the entry (`corpus-roundtrip.test.ts` / `render-roundtrip.repro.test.ts` must stay green).

## Reuse analysis

**Capability:** delete an entry from a section of the reconstructed résumé.

**Existing surfaces found:**
- `RemoveButton` — `src/components/features/ReconstructedAdd.tsx` (already used by all four sections for added entries).
- `removeBullet` / `removedBullets` — `src/hooks/useEditableParse.ts:1078`, `src/lib/edit/apply-overrides.ts:782`. The exact one-level-down analogue of what this issue needs.
- `parsedEntryKey` / `isAddedEntryKey` — `src/hooks/useEditableParse.ts:372`, `:386`. The key space already exists.

**Decision:** **extend** — no new component, no new panel, no new dialog. This adds one state set to an existing hook, one pass to an existing function, and removes four `added &&` gates around a primitive that is already rendered in all four sections.

## Acceptance criteria

- [ ] A **parsed** achievement can be deleted from the reconstructed résumé, and stays deleted across a re-render.
- [ ] The same works for a parsed experience role, education entry, and project.
- [ ] Deleting an entry also drops its bullets from the score (`sections`) and from `rawText` / the "Raw text & flags" disclosure.
- [ ] The deleted entry is absent from the Download PDF, and the exported PDF re-parses without it (round-trip invariant holds).
- [ ] **Index integrity:** with per-entry field overrides on entries at indices 1 and 2, deleting index 0 leaves each surviving entry holding its own override. Covered by an explicit test.
- [ ] Deleting an entry sets `hasEdits`, is captured in `EditSnapshot`, and is restored by `replay`.
- [ ] A draft/snapshot written **before** this change replays without error (no `removedEntries` field present).
- [ ] `reset` clears `removedEntries`.
- [ ] No regression to added-entry removal, bullet removal, prune-on-blur (`pruneEmptyAddedEntries`), or any existing edit surface.
- [ ] `npm run verify` green, including `corpus.test.ts` and `corpus-roundtrip.test.ts`.

## Not in scope

- **Achievement and project bullets are read-only** (`ReconstructedResume.tsx:981`, `:764` pass only `bullet={b}` — no `onBulletChange`, no per-row `onRemove`), so an individual bullet under an achievement still cannot be edited or removed after this lands. Worth a separate issue; this one delivers entry-level deletion, which is what the report asked for. (Step 4 threads `onRemoveBullet` into those sections for the entry-delete cascade, which makes the follow-up mostly prop-threading.)
- Undo-as-a-visible-affordance for a deleted entry. The snapshot/`replay` path covers restore; a dedicated "undo delete" toast is a separate UX question.
- Refactoring `applyOverrides`'s 18-parameter positional signature.

s-annam:
Implements #856: parsed entries (achievement, role, education, project) can now be deleted from the reconstructed résumé, not only user-added ones.

Design is the `removedEntries` tombstone set the issue specified — mark, run every index-keyed override pass, then filter once at the end of `applyOverrides`, so no override rebinds to a neighbour.

### Deviations from the issue plan

Three deliberate departures from the issue's implementation plan, all additive to its intent:

1. NO `removedEntriesRef` (plan Step 1). The `removedBulletsRef` it mirrors exists solely so `removeBullet` can read the committed set to decide its boolean return; `removeEntry` returns void, so a ref beside `removedEntries` would have had no reader. Omitted rather than shipped dead.

2. ADDED a UI index-remap the plan does not mention, and it is load-bearing. The plan's Step 4 resolves the delete key as `parsedEntryKey(<section>, i)` where `i` is the render index. Because Step 3 filters tombstoned entries out of the parsed arrays, `i` stops being the parsed index after the first deletion — so that line would (a) re-delete the already-gone entry instead of the clicked one, and (b) leave every neighbouring `onAchievementField(i, …)` / `educationOverrides[i]` / `descriptionOverrides` key rebound to the wrong entry, which is the exact failure the tombstone design prevents one level down. `survivingParsedIndices` (exported from `useEditableParse.ts`) is the single map every section resolves through. It enumerates survivors rather than subtracting the set's size, so a stale tombstone naming an index the parse no longer has cannot over-shift.

3. The entry's own source-line removal (plan Step 3) reuses the SCORER's ownership relation rather than a new "normalise-and-match-first-line" matcher. `normalizeTitleKey` + `stripAchievementLabel` (already private in `group-bullets.ts`) are what `suppressTitleOwnedBullets` hides an entry-owned line on; `isTitleOwnedLine` exposes the inverse, sharing one extracted `matchesOwnedKey` so the two cannot drift. This is not cosmetic: a title-only achievement/project is parsed OUT of a `•`-marked line, so that line is in the graded pool with no description to attribute it to — deleting the entry without dropping the line leaves its content grading the résumé forever. It is best-effort and documented as such (the parsed model keeps no source-line provenance, so a glued one-line header matches nothing and the line stays in the raw-text view); safe when it misses, since the lines it can miss are the UNMARKED ones the bullet pool ignores.

Two smaller notes:
- Entry `key`s in all four sections moved from the render position to the entry key. A position key would hand a deleted row's in-flight edit state to its successor — newly reachable now that parsed rows can disappear.
- One drive-by, forced: `ReconstructedRole.tsx:262` carried a pre-existing "exactly" inside a `/** */` block, which `scripts/hooks/check_conventions.py` rule 2 blocks on (block comments are deliberately scanned) — it fired on every edit to that file. Reworded to "just as". Comment text only.

Also fixed while testing: `parseRemovedEntryKey` matches the index half as DIGITS rather than coercing with `Number`. `Number("")` is 0 and `Number.isInteger(0)` is true, so a truncated `"projects:"` key would have silently deleted the section's FIRST entry. Pinned by a test.

### Files changed

- `src/hooks/useEditableParse.ts`
- `src/lib/score/group-bullets.ts`
- `src/components/features/EntryRemove.ts`
- `src/components/features/ReconstructedResume.tsx`
- `src/components/features/ReconstructedRole.tsx`
- `src/components/features/ReconstructedEducationSkills.tsx`
- `src/components/features/ExperienceSection.test.tsx`
- `src/components/features/ExperienceSection.other-bullets.test.tsx`
- `src/components/features/ExperienceSection.prune-hold.test.tsx`
- `src/components/features/ReconstructedEducationSkills.test.ts`
- `src/components/features/ReconstructedResume.remove-parsed-entry.test.tsx`
- `src/hooks/useEditableParse.removed-entry.test.tsx`
- `src/lib/edit/apply-overrides.removed-entry.test.ts`
- `src/lib/pdf/render-roundtrip-entry-remove.repro.test.ts`

### Validation

- `npm run typecheck` — pass — tsc -b --noEmit — no output, exit 0
- `npm run lint` — pass — eslint . — no output, exit 0
- `npm test` — pass — Test Files 363 passed | 8 skipped (371); Tests 5901 passed | 10 skipped (5911); Duration 19.77s

### Adversarial review

Not yet run. The run halted at submit (a stray `fake` git remote made `gh stack submit` ambiguous), so the review pass never executed. Reviewer notes below are from `npx fallow audit` only.

### Handoff

New exports the next issue can build on:

- `useEditableParse.ts`
  - `EditableParse.removedEntries: ReadonlySet<string>` — parsedEntryKey tombstones.
  - `EditSnapshot.removedEntries?: string[]` — OPTIONAL, so pre-#856 persisted drafts replay; `replay` defaults it to `[]`. Any new consumer must keep it optional.
  - `removeEntry(key: string)` — widened from "added id" to "added id OR parsedEntryKey". Also deletes that key's `addedBullets` bucket for BOTH kinds.
  - `survivingParsedIndices(section, removedEntries, renderedCount): number[]` — render position → PARSED index. ANY new surface that renders a parsed section and writes an index-keyed override MUST resolve through this.

- `lib/score/group-bullets.ts`: `isTitleOwnedLine(line, title)` — the entry-owns-this-line relation, shared with `suppressTitleOwnedBullets`.

- `components/features/EntryRemove.ts`: `removeEntryWithBullets(entryKey, bullets, {onRemoveEntry, onRemoveBullet})` — the one definition of the delete gesture. Bullets first, then the entry.

- `applyOverrides` gained an 18th positional param `removedEntries: ReadonlySet<string> = new Set()`. A 19th param must be appended AFTER it, and `applyRemovedEntries` must stay the LAST pass (it runs after `applyAddedEntriesAndBullets`) or index-keyed overrides rebind.

- `AchievementsSection` is now exported from `ReconstructedResume.tsx` (test-only, like `ExperienceSection`).

Two landmines documented in code, worth knowing:
- Inside `ExperienceSection`, the `experience:<n>` strings in `rewriteApplyBySection` are in TWO different index spaces — the map KEY is the rewrite-chain section id (RENDER position, matching `buildResumeSections`), while `entryKey` is the `addedBullets` bucket key (PARSED index). They coincide only until a parsed role is deleted.
- Achievement and project bullets are still read-only per-row (issue explicitly out of scope). Step 4 threaded `onRemoveBullet` into both sections for the delete cascade, so that follow-up is now mostly prop-threading down to `ResumeBulletRow`.

Not touched, per the issue's "Not in scope": per-bullet edit/remove on achievements/projects, an undo-delete toast, and the 18-param positional signature of `applyOverrides`.

Closes #856

### DIFF
--- a/src/components/features/ReconstructedResume.tsx
+++ b/src/components/features/ReconstructedResume.tsx
@@ -83,7 +83,11 @@ import type {
   AddedBulletRef,
   AddedBullets,
 } from "../../hooks/useEditableParse.ts";
-import { parsedEntryKey } from "../../hooks/useEditableParse.ts";
+import {
+  parsedEntryKey,
+  survivingParsedIndices,
+} from "../../hooks/useEditableParse.ts";
+import { removeEntryWithBullets } from "../../lib/edit/entry-remove.ts";
 import { useAddedEntryPruneHold } from "../../hooks/useAddedEntryPruneHold.ts";
 import {
   batchUndoTargets,
@@ -406,6 +410,7 @@ export function ExperienceSection({
   addedBullets,
   addedExperience,
   originalCount,
+  parsedIndices,
   onAddEntry,
   onRemoveEntry,
   onEntryField,
@@ -435,6 +440,9 @@ export function ExperienceSection({
   onRewriteApplied?: () => void;
   hasBullets: boolean;
   experienceOverrides: Record<number, ExperienceFieldOverrides>;
+  /** `index` is the role's PARSED index — the key space `experienceOverrides`
+   *  uses — NOT its render position. The two diverge once a parsed role is
+   *  deleted (#856); {@link parsedIndices} is the map between them. */
   onExperienceFieldChange: (
     index: number,
     field: keyof ExperienceFieldOverrides,
@@ -456,10 +464,14 @@ export function ExperienceSection({
   addedBullets: AddedBullets;
   /** User-added experience entries, append-aligned to indices ≥ originalCount. */
   addedExperience: AddedEntry[];
-  /** Count of PARSED experience roles; indices at/above this are user-added. */
+  /** Count of PARSED experience roles still rendered; indices at/above this are
+   *  user-added. */
   originalCount: number;
+  /** Render position → PARSED index for the surviving parsed roles (#856), from
+   *  {@link survivingParsedIndices}. Identity until a parsed role is deleted. */
+  parsedIndices: readonly number[];
   onAddEntry: () => void;
-  onRemoveEntry: (id: string) => void;
+  onRemoveEntry: (key: string) => void;
   onEntryField: (id: string, field: AddedEntryField, value: string) => void;
   onAddBullet: (entryKey: string, text: string) => void;
   /** Snapshot the slots a rewrite batch will write, for a one-action undo of
@@ -478,6 +490,15 @@ export function ExperienceSection({
 }) {
   // "Other" is appended with a null index; real roles carry their index.
   const roleCount = groups.filter((g) => g.experienceIndex !== null).length;
+
+  // A rendered role's PARSED index — the key space `experienceOverrides` and
+  // every `addedBullets` bucket use. `applyOverrides` filters deleted parsed
+  // roles out of the array the groups were built over (#856), so the render
+  // position stops being the parsed index after the first deletion. An index at
+  // or above `originalCount` is a user-ADDED role, appended after every
+  // survivor, and never reaches an index-keyed map at all.
+  const parsedIndexOf = (idx: number): number =>
+    idx < originalCount ? (parsedIndices[idx] ?? idx) : idx;
 
   // Per-entry prune hold (#637 half 2). Created HERE — the ids it holds are
   // only meaningful to this section's `pruneEmptyAddedEntries` call, and the
@@ -526,11 +547,19 @@ export function ExperienceSection({
       if (idx === null) continue;
       const added =
         idx >= originalCount ? addedExperience[idx - originalCount] : undefined;
-      const entryKey = added ? added.id : parsedEntryKey("experience", idx);
+      const entryKey = added
+        ? added.id
+        : parsedEntryKey("experience", parsedIndexOf(idx));
       const bucketRef = (obsId: string): AddedBulletRef | undefined => {
         const text = group.bullets.find((b) => b.id === obsId)?.text;
         return text === undefined ? undefined : { entryKey, text };
       };
+      // NOTE the two `experience:<n>` strings here are in DIFFERENT index
+      // spaces and only look alike. This map's key is the rewrite-chain section
+      // id `buildResumeSections` mints from the RENDER position, and both ends
+      // are built in the same render, so it stays internally consistent.
+      // `entryKey` above is the persisted `addedBullets` bucket key and must be
+      // the PARSED index. They coincide until a parsed role is deleted (#856).
       map.set(`experience:${idx}`, {
         obsIds: group.bullets.map((b) => b.id),
         // Entry-aware like the per-role path (#637, #657): a rewrite accepted on
@@ -549,6 +578,7 @@ export function ExperienceSection({
   }, [
     groups,
     originalCount,
+    parsedIndices,
     addedExperience,
     onBulletChange,
     onRemoveBullet,
@@ -623,23 +653,27 @@ export function ExperienceSection({
               idx >= originalCount
                 ? addedExperience[idx - originalCount]
                 : undefined;
+            const parsedIdx = parsedIndexOf(idx);
             // The one bucket this role owns — its added id, or its parsed-entry
             // key. Adds append to it, a rewrite batch's undo snapshots it, and
             // (since #637) a remove of a user-added bullet splices it.
             const roleEntryKey = added
               ? added.id
-              : parsedEntryKey("experience", idx);
+              : parsedEntryKey("experience", parsedIdx);
             const entry = (
               <RoleEntry
-                key={added ? added.id : idx}
+                // The ENTRY key, not the render position (#856): a deletion
+                // shifts every later row up one, and a position key would hand
+                // the deleted row's in-flight edit state to its successor.
+                key={roleEntryKey}
                 group={group}
                 experienceIndex={idx}
-                overrides={added ? undefined : experienceOverrides[idx]}
+                overrides={added ? undefined : experienceOverrides[parsedIdx]}
                 onFieldChange={(field, value) => {
                   if (added) {
                     onEntryField(added.id, EXPERIENCE_FIELD_MAP[field], value);
                   } else {
-                    onExperienceFieldChange(idx, field, value);
+                    onExperienceFieldChange(parsedIdx, field, value);
                   }
                 }}
                 onBulletChange={onBulletChange}
@@ -648,13 +682,21 @@ export function ExperienceSection({
                 captureUndo={(writes) =>
                   captureBulletUndo(batchUndoTargets(writes, roleEntryKey))
                 }
-                onRemove={added ? () => onRemoveEntry(added.id) : undefined}
+                // Set for a PARSED role too since #856 — its own key tombstones
+                // the entry, and the bullets under it go through `removeBullet`
+                // (see `removeEntryWithBullets` for why they have to).
+                onRemove={() =>
+                  removeEntryWithBullets(roleEntryKey, group.bullets, {
+                    onRemoveEntry,
+                    onRemoveBullet,
+                  })
+                }
                 entryKey={roleEntryKey}
                 pruneHold={pruneHold}
               />
             );
             return subHeading ? (
-              <Fragment key={added ? added.id : idx}>
+              <Fragment key={roleEntryKey}>
                 <SectionHeading>{subHeading}</SectionHeading>
                 {entry}
               </Fragment>
@@ -675,10 +717,12 @@ export function ExperienceSection({
 
 /**
  * Projects render as their OWN section (#95) — a name-led header + the same
- * graded bullet rows used everywhere else (`ResumeBulletRow`). Parsed projects
- * stay read-only; user-ADDED projects expose an editable name, a "+ Add bullet"
- * affordance, and a remove control (#180-followup), so an added project's
- * bullets grade and export like any other.
+ * graded bullet rows used everywhere else (`ResumeBulletRow`). A parsed
+ * project's name stays read-only; user-ADDED projects expose an editable name
+ * and a "+ Add bullet" affordance (#180-followup), so an added project's
+ * bullets grade and export like any other. BOTH carry a remove control since
+ * #856 — a phantom project the parser stitched together was previously
+ * uncorrectable and shipped into the Download PDF.
  */
 function ProjectsSection({
   heading,
@@ -687,8 +731,10 @@ function ProjectsSection({
   descriptionOverrides,
   addedProjects,
   originalCount,
+  parsedIndices,
   onAddEntry,
   onRemoveEntry,
+  onRemoveBullet,
   onEntryField,
   onDescriptionField,
   onAddBullet,
@@ -703,9 +749,16 @@ function ProjectsSection({
    *  a CLEARED prose field mounted (so the clear is reversible in-session). */
   descriptionOverrides: DescriptionOverrides;
   addedProjects: AddedEntry[];
+  /** Count of PARSED projects still rendered; indices at/above are user-added. */
   originalCount: number;
+  /** Render position → PARSED index for the surviving parsed projects (#856),
+   *  from {@link survivingParsedIndices}. Identity until one is deleted. */
+  parsedIndices: readonly number[];
   onAddEntry: () => void;
-  onRemoveEntry: (id: string) => void;
+  onRemoveEntry: (key: string) => void;
+  /** Drop one bullet, so deleting an entry can take its bullets out of the
+   *  graded pool with it (#856) — see `removeEntryWithBullets`. */
+  onRemoveBullet: (id: string, added?: AddedBulletRef) => boolean;
   onEntryField: (id: string, field: AddedEntryField, value: string) => void;
   /** Commit an edit to a parsed project's prose description (#489). */
   onDescriptionField: (key: string, value: string | undefined) => void;
@@ -724,7 +777,11 @@ function ProjectsSection({
           const group = groups[i];
           const added =
             i >= originalCount ? addedProjects[i - originalCount] : undefined;
-          const descKey = parsedEntryKey("projects", i);
+          // PARSED index, not the render position: `applyOverrides` filters
+          // deleted entries out, so the two diverge after the first delete
+          // (#856) and `descriptionOverrides` is keyed by the parsed one.
+          const descKey = parsedEntryKey("projects", parsedIndices[i] ?? i);
+          const entryKey = added ? added.id : descKey;
           // Keep the prose field mounted once the user has touched it, even after
           // an authoritative clear ("" override) drops `project.description` — so
           // the clear stays reversible in-session instead of collapsing to the
@@ -734,7 +791,9 @@ function ProjectsSection({
             .filter(Boolean)
             .join(" · ");
           return (
-            <div key={added ? added.id : i} className="flex flex-col gap-1.5">
+            // The ENTRY key, not the render position (#856) — see the same note
+            // in `ExperienceSection`.
+            <div key={entryKey} className="flex flex-col gap-1.5">
               <div className="flex items-start justify-between gap-2">
                 {added ? (
                   <EditableField
@@ -751,12 +810,15 @@ function ProjectsSection({
                     {header || "Untitled project"}
                   </h3>
                 )}
-                {added && (
-                  <RemoveButton
-                    label="Remove project"
-                    onClick={() => onRemoveEntry(added.id)}
-                  />
-                )}
+                <RemoveButton
+                  label="Remove project"
+                  onClick={() =>
+                    removeEntryWithBullets(entryKey, group?.bullets ?? [], {
+                      onRemoveEntry,
+                      onRemoveBullet,
+                    })
+                  }
+                />
               </div>
               {group && group.bullets.length > 0 ? (
                 <ul className="list-none">
@@ -891,16 +953,24 @@ function AchievementHeader({
  * Both branches are editable: a PARSED achievement's type / title / year through
  * `AchievementHeader` (#454, overrides keyed by parsed index and already folded
  * into `achievements` by `applyOverrides`), a user-ADDED one through the flat
- * `AddedEntry` fields (#455).
+ * `AddedEntry` fields (#455). Both are also removable since #856 — this is the
+ * section the report came from: a phantom achievement could be blanked field by
+ * field but never actually dropped, so it still shipped into the Download PDF.
  */
-function AchievementsSection({
+// Exported for `ReconstructedResume.remove-parsed-entry.test.tsx` only — the
+// render-position → parsed-index resolution (#856) is a property of THIS
+// component, so the regression test has to render it. Not part of the surface
+// any other module imports; same arrangement as `ExperienceSection` above.
+export function AchievementsSection({
   heading,
   achievements,
   groups,
   addedAchievements,
   originalCount,
+  parsedIndices,
   onAddEntry,
   onRemoveEntry,
+  onRemoveBullet,
   onEntryField,
   onAchievementField,
   onAddBullet,
@@ -912,9 +982,18 @@ function AchievementsSection({
   /** Pre-built achievement groups, index-aligned with `achievements`. */
   groups: BulletGroup[];
   addedAchievements: AddedEntry[];
+  /** Count of PARSED achievements still rendered; at/above this are user-added. */
   originalCount: number;
+  /** Render position → PARSED index for the surviving parsed achievements
+   *  (#856), from {@link survivingParsedIndices}. Identity until one is
+   *  deleted — after which `achievementOverrides`' keys would otherwise rebind
+   *  each survivor's edits to its neighbour's. */
+  parsedIndices: readonly number[];
   onAddEntry: () => void;
-  onRemoveEntry: (id: string) => void;
+  onRemoveEntry: (key: string) => void;
+  /** Drop one bullet, so deleting an entry can take its bullets out of the
+   *  graded pool with it (#856) — see `removeEntryWithBullets`. */
+  onRemoveBullet: (id: string, added?: AddedBulletRef) => boolean;
   onEntryField: (id: string, field: AddedEntryField, value: string) => void;
   onAchievementField: (
     index: number,
@@ -938,8 +1017,15 @@ function AchievementsSection({
             i >= originalCount
               ? addedAchievements[i - originalCount]
               : undefined;
+          // PARSED index, not the render position — see `parsedIndices`.
+          const parsedIdx = parsedIndices[i] ?? i;
+          const entryKey = added
+            ? added.id
+            : parsedEntryKey("achievements", parsedIdx);
           return (
-            <div key={added ? added.id : i} className="flex flex-col gap-1.5">
+            // The ENTRY key, not the render position (#856) — see the same note
+            // in `ExperienceSection`.
+            <div key={entryKey} className="flex flex-col gap-1.5">
               <div className="flex items-start justify-between gap-2">
                 {added ? (
                   // Same header as a parsed achievement — an added entry stores
@@ -964,16 +1050,19 @@ function AchievementsSection({
                     year={achievement.year}
                     yearSeparator={achievement.year_separator}
                     onFieldChange={(field, value) =>
-                      onAchievementField(i, field, value)
+                      onAchievementField(parsedIdx, field, value)
                     }
                   />
                 )}
-                {added && (
-                  <RemoveButton
-                    label="Remove achievement"
-                    onClick={() => onRemoveEntry(added.id)}
-                  />
-                )}
+                <RemoveButton
+                  label="Remove achievement"
+                  onClick={() =>
+                    removeEntryWithBullets(entryKey, group?.bullets ?? [], {
+                      onRemoveEntry,
+                      onRemoveBullet,
+                    })
+                  }
+                />
               </div>
               {group && group.bullets.length > 0 && (
                 <ul className="list-none">
@@ -1078,6 +1167,7 @@ export function ReconstructedResume({
     descriptionOverrides,
     setDescriptionField,
     removeBullet,
+    removedEntries,
     educationOverrides,
     setEducationField,
     setAchievementField,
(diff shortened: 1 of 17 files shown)


s-annam (commented):
### Verdict: APPROVE
*(Posted as a review comment because GitHub restricts PR authors from self-submitting formal approval reviews)*

Reviewed against issue #856. All 10 acceptance criteria are fully met with comprehensive test coverage across 4 dedicated test suites. All offlinecv verification gates (`check:fixtures`, lint, typecheck, dead-code audit, and full test suite) are green.

### Acceptance Criteria Checklist (#856)
- [x] **Parsed achievement deletion**: Verified across re-renders (`ReconstructedResume.remove-parsed-entry.test.tsx`, `useEditableParse.removed-entry.test.tsx`).
- [x] **Experience, education, project deletion**: All 4 sections support parsed entry removal via tombstoning.
- [x] **Bullet & raw text removal**: `removeEntryWithBullets` drops associated bullets from the graded pool (`sections`) and `rawText`.
- [x] **Download PDF & Round-trip invariant**: Verified via `render-roundtrip-entry-remove.repro.test.ts`.
- [x] **Index integrity**: Overriding an entry and deleting an earlier entry preserves override alignment via `survivingParsedIndices`.
- [x] **Edit state & snapshots**: `hasEdits` tracks `removedEntries`, captured in `EditSnapshot`, restored on `replay`.
- [x] **Backwards compatibility**: Legacy snapshots without `removedEntries` replay cleanly without throwing (`snap.removedEntries ?? []`).
- [x] **Reset support**: `reset` clears `removedEntries`.
- [x] **No regressions**: Added-entry removal, bullet removal, and prune-on-blur remain green.
- [x] **Full verify suite**: 363 test files / 5901 tests passed.

### Gates & Quality Audit
- **Fixture PII (Gate 3a)**: Passed (`check:fixtures` green, 0 real personas).
- **Design System & Reuse (Gate 3b)**: Passed (`RemoveButton` and design system primitives reused consistently; no raw interactive elements in feature code).
- **Style Tokens (Gate 3c)**: Passed (0 hardcoded hex or raw Tailwind palette violations).
- **Dead Code & Fallow (Gate 3d)**: Passed (0 dead exports, verify suite green).
- **Description Accuracy (Gate 3f)**: Accurate. All 3 deliberate design deviations from the issue's initial plan (`removedEntriesRef` omission, UI `survivingParsedIndices` index-remap, and reusing `isTitleOwnedLine` from `group-bullets.ts`) are well-justified, accurately described, and tested.

---

src/lib/edit/entry-remove.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/**
+ * Delete the entry keyed `entryKey` along with every bullet rendered under it.
+ *
+ * `bullets` is the entry's rendered group — the re-graded rows the user can
+ * actually see — so `b.text` is each row's CURRENT text, which is what both
+ * halves of `removeBullet`'s routing match on.
+ */
+export function removeEntryWithBullets(
```
s-annam, line 53: **Secondary — placement.** This module is pure domain logic: zero React, zero JSX, no hook, no component. It takes two callbacks and a `BulletObservation[]` and orders two writes.

`CLAUDE.md` puts that in `src/lib/`:

> **Domain logic stays in `src/lib/`** (`heuristics/`, `score/`, `pdf/`, …), strictly separated from UI.

And `src/lib/edit/` is the seam that already owns exactly this concern — `apply-overrides.ts`, the thing on the other end of `removeEntry`/`removeBullet`, lives there. As a bonus, a new file under `src/components/` is what `scripts/hooks/reuse_surface_reminder.sh` warns on, and #856's own Reuse analysis says "no new component" — which is true of the *behaviour* but reads oddly against a new file in the component tree.

Not blocking, and not a behaviour change either way: `src/lib/edit/entry-remove.ts` would need two import updates (`ReconstructedResume.tsx`) and nothing else. Co-location with the four callers is a defensible counter-argument — but if the answer is "co-location wins here," that is worth one line in the docblock, because the file currently gives a reader no reason why it is not in `lib/`.

src/components/features/ReconstructedResume.tsx
```diff
@@ -964,16 +1050,19 @@ function AchievementsSection({
...
                   />
                 )}
-                {added && (
-                  <RemoveButton
-                    label="Remove achievement"
-                    onClick={() => onRemoveEntry(added.id)}
-                  />
-                )}
+                <RemoveButton
+                  label="Remove achievement"
```
s-annam, line 1058: **Secondary — a one-click destructive delete that skips the confirmation its own bullets have.**

`RemoveButton` (`ReconstructedAdd.tsx:132`) is a bare `<Button>`; it does not confirm. And `removeEntryWithBullets` calls `onRemoveBullet` directly, which bypasses the per-bullet confirmation strip (`removes.pending`, #626) that guards removing a **single** bullet in the Experience path.

So the current gradient is inverted: removing one bullet asks first; removing an entry silently destroys the entry *and* all N of its bullets on one click. Before this PR that only applied to user-ADDED entries — trivially re-creatable, so no confirm was needed. A parsed entry is different: it carries parser output, and the only recovery is `resetAll` (which discards **every** edit in the session) or re-uploading the PDF. #856 scoped out an *undo toast*; it did not scope out a confirm, and the asymmetry with #626 is the part that stands out.

Not blocking — the behaviour is what the issue asked for and the tests pin it. Worth a follow-up issue rather than a change in this PR, given the round-trip and index work here is already large and well-tested. Same applies to `label="Remove project"` (`:814`), `"Remove role"` (`ReconstructedRole.tsx:440`) and `"Remove education"` (`ReconstructedEducationSkills.tsx:166`).

src/components/features/ReconstructedResume.tsx
```diff
@@ -1140,6 +1230,43 @@ export function ReconstructedResume({
   const originalProjCount = projects.length - addedProjects.length;
   const originalAchCount = achievements.length - addedAchievements.length;
 
+  // Render position → PARSED index, per section (#856). `applyOverrides` filters
+  // a deleted parsed entry out of these arrays, so from the first deletion on, a
+  // rendered row's position is NOT the index its overrides are keyed by. Every
+  // section resolves through the map rather than re-deriving one, so the four
+  // cannot drift; see `survivingParsedIndices` for why it is enumerated rather
+  // than computed by subtraction.
+  const expParsedIndices = survivingParsedIndices(
```
s-annam, line 1239: **Secondary — this file grew on the one file `CLAUDE.md` tells us not to grow.**

`ReconstructedResume.tsx`: **1356 → 1489 lines** (+133). From `CLAUDE.md`:

> ⚠️ **Known debt — do not imitate:** `ReconstructedResume.tsx` (1338) … If you are editing one, prefer extracting your change into a new sibling over growing the file further.

The PR *did* extract a sibling — `EntryRemove.ts` — which is the right instinct; it just is not where the mass is. `fallow` now reports three functions in this one file over the ~200 LOC guidance: `ReconstructedResume` (370), `ExperienceSection` (320), `ProjectsSection` (143), with `AchievementsSection` alongside them.

The honest framing: this PR did not create that debt and the four sections were already co-located, so splitting them out here would balloon a diff that is already 1846 insertions and would make the index-remap change — the part that actually needs careful review — harder to see. **Recommend not doing it in this PR**, and instead filing the section split as its own issue. Flagging it because the file is now 151 lines past the number `CLAUDE.md` quotes, so that quote is stale and the next person to read it will under-estimate the problem.

src/lib/edit/apply-overrides.ts
```diff
@@ -961,6 +1002,164 @@ function applyAddedBulletsToExistingEntries(
...
+    if (!target) continue;
+    const indices = bySection.get(target.section) ?? new Set<number>();
+    indices.add(target.index);
+    bySection.set(target.section, indices);
+  }
+
+  let { rawText, sections } = views;
+  for (const [section, indices] of bySection) {
+    for (const title of dropRemovedEntries(nextParsed, section, indices)) {
+      const matches: LineMatcher = (line) => isTitleOwnedLine(line, title);
```
s-annam, line 1162: **Nit — first-match line removal picks the wrong line when two parsed entries share a title key.**

`withMatchedRawTextLine` / `withMatchedSectionLine` remove the **first** line satisfying the predicate. `dropRemovedEntries` has already filtered the correct entry out of the parsed array by index, but the line search that follows has no index — only the title. So with two entries whose `normalizeTitleKey` collides (which is exactly the "duplicated role" / "phantom achievement" shape #856 exists to fix), deleting the *later* one removes the *earlier* one's source line.

Why this is a Nit and not a bug worth changing code for — both cases come out benign:

- **Exact duplicates:** the two lines are textually identical, so removing either leaves `rawText` and the graded pool in the same state. Indistinguishable.
- **Near-duplicates that still collide on the residue-tolerant key:** the surviving entry's title still matches the leftover line, so `suppressTitleOwnedBullets` keeps it out of the bullet pool anyway. The score is unaffected; the cost is at most one slightly-wrong line in the "Raw text & flags" disclosure.

The `applyRemovedEntries` docblock already scopes the pass as "BEST-EFFORT BY CONSTRUCTION, and safe when it misses," which covers the *miss* case. It does not cover the *mis-hit* case, and that is the one a future reader will trip over. One sentence in that docblock — "a title-key collision between two parsed entries removes the first matching line, which is benign because …" — would close it. No code change wanted.

s-annam (commented):
## ⚠️ This is a SELF-REVIEW and carries no approval weight

This PR is authored by `s-annam`, which is also the account this review is posted from. GitHub will not accept an approving review from a PR's own author, and posting the same verdict as a `COMMENT` would not make it one either. **This PR still needs a second-party reviewer before it merges.** Read everything below as the author's own audit trail, not as a sign-off.

---

The change does what #856 specified, the tombstone ordering is right, and the deviations from the issue's plan are improvements on it rather than shortcuts. **Zero blocking findings.** Everything below is Secondary or Nit.

## Acceptance criteria (#856) — 10 / 10 met

| AC | Status | Evidence |
|---|---|---|
| Parsed achievement deletable, survives re-render | ✅ | `useEditableParse.removed-entry.test.tsx:231,254` |
| Same for role / education / project | ✅ | role `:275`; education + achievement `render-roundtrip-entry-remove.repro.test.ts:137,143`; project `apply-overrides.removed-entry.test.ts:145,183` |
| Bullets leave the score + `rawText` | ✅ | `:275`, `apply-overrides.removed-entry.test.ts:232` |
| Absent from Download PDF, re-parses without it | ✅ | `render-roundtrip-entry-remove.repro.test.ts:128–151` |
| **Index integrity** — survivors keep their own overrides | ✅ | `apply-overrides.removed-entry.test.ts:105–167`, all four sections + a two-in-one-fold case |
| `hasEdits` + `EditSnapshot` + `replay` | ✅ | `:327` |
| Pre-#856 snapshot replays without error | ✅ | `:347` |
| `reset` clears `removedEntries` | ✅ | `:333–334` |
| No regression to added-entry removal | ✅ | `:379` |
| `npm run verify` green | ✅ | reproduced below |

## Gates

| Gate | Result |
|---|---|
| `tsc -b --noEmit` | pass |
| `eslint .` | pass |
| `check:fixtures` | pass — **N/A by construction**, the diff touches no `tests/fixtures/**` binary |
| `check:baselines` | pass |
| `check:core` | pass |
| Full test suite (`OFFLINECV_FULL_TESTS=1`) | **363 files passed / 8 skipped · 5901 tests passed / 10 skipped** |
| `vite build` | pass — 1028 modules, 3.28s |
| Style tokens (3c) | clean — every regex hit was the literal `#856`, not a hex colour |
| Skill/script command bugs (3e) | N/A — no `scripts/**` or `.claude/skills/**` file in the diff |
| `fallow audit` | exit 1 — 3 complexity findings, 35 clone groups. **Report-only**, and all three complexity findings (`RoleHeader`, `applyExperienceHeaderOverrides`, `groupBulletsByExperience`) are on functions this diff does not touch; fallow re-attributes them because the diff touched their files |

The reported numbers match the PR body's `Validation` section exactly (363/5901), so that claim round-trips.

## Description accuracy (3f) — accurate, and unusually so

Every checkable claim in the body was round-tripped against the diff and holds:

- **Deviation 1** (no `removedEntriesRef`) — confirmed; `useEditableParse.ts:921` documents the absent reader, and `removeEntry` does return `void`.
- **Deviation 2** (UI index remap) — confirmed and correctly characterised as load-bearing. `survivingParsedIndices` is exported, and all four sections resolve through it; the issue's Step 4 as literally written *would* have re-deleted the wrong entry after the first deletion.
- **Deviation 3** (reuse the scorer's ownership relation) — confirmed; `matchesOwnedKey` is the single extracted predicate and both `suppressTitleOwnedBullets` and `isTitleOwnedLine` route through it.
- **The forced drive-by** — verified independently, not taken on trust. `scripts/hooks/check_conventions.py:103-106` scans `src/components/**` after `strip_line_comments`, which strips `//` but **not** `/** */`, so the pre-existing `exactly` at `ReconstructedRole.tsx:262` genuinely fired on every edit to that file. Comment-text-only, and it rides a PR already in range rather than earning its own.
- **`parseRemovedEntryKey` digit matching** — confirmed, and pinned by the `"projects:"` case at `apply-overrides.removed-entry.test.ts:197`. The `Number("") === 0` trap was real.
- **Files-changed list** — matches `git diff --name-only` exactly, 16 for 16.
- **"Adversarial review: not yet run"** — correct to state plainly rather than let a reader assume it happened.

One thing the body flags as a landmine that is worth restating, because it is the sharpest edge here: inside `ExperienceSection` the two `experience:<n>` strings are in **different index spaces** (rewrite-chain section id = render position; `entryKey` = parsed index). They coincide until the first parsed-role deletion. The code comments it in place, which is the right treatment.

## Secondary

Three, all inline. In short: `EntryRemove.ts` is domain logic sitting under `src/components/`; `ReconstructedResume.tsx` grew 1356 → 1489 on a file `CLAUDE.md` names as known debt; and a one-click destructive delete now bypasses the per-bullet confirmation strip that guards removing a *single* bullet.

## Nits

- Duplicate-title first-match in `applyRemovedEntries` (inline). Near-zero real impact — noting it so the next reader does not have to re-derive why it is safe.
- `fallow` attributes 3 clone groups to `ExperienceSection.test.tsx` / `.other-bullets` / `.prune-hold`, which this PR touched by **+3 lines each**. Pre-existing duplication, re-attributed by proximity. Report-only; not worth a change.

## Nothing was auto-fixed, deliberately

With 0 blockers the reviewer loop would normally apply small fixes and push. It did not here, because none of the findings qualify: all three Secondary items are structural judgement calls (where a file lives, how a component is decomposed, whether a destructive action confirms) rather than non-behavioural mechanical nits, and the Step 5.6 suggestion-block form does not fit any of them either. The branch is untouched and still exactly **one commit** (`bfa94b5`), so the one-commit merge-queue invariant holds as-is.

---

Head SHA reviewed: `bfa94b5a465d56e1536b3cb797c052ff61b47cd5`

src/lib/edit/entry-remove.ts
s-annam: Agreed — moved in `5ec090c`. `src/components/features/EntryRemove.ts` → `src/lib/edit/entry-remove.ts`, beside `apply-overrides.ts`.

You called the import cost right in spirit but it was two files, not one: `ReconstructedResume.tsx:90` and `useEditableParse.removed-entry.test.tsx:40`, plus the module’s own `../../lib/score/score.ts` → `../score/score.ts`. `git mv` kept it a 97% rename, so the diff reads as a move.

No docblock justification needed now that it is in `lib/` — co-location was the weaker argument, and the `reuse_surface_reminder.sh` warning goes away as a side effect.

src/lib/edit/apply-overrides.ts
s-annam: Fixed the docblock in `5ec090c` — no code change, as you asked.

Added a fourth paragraph to `applyRemovedEntries` covering the **mis-hit**, since the existing "BEST-EFFORT BY CONSTRUCTION" paragraph only covers the *miss*:

> It can also MIS-HIT, and that is benign for a different reason. The array filter above is by index, but the line search that follows has only the title, and `withMatchedRawTextLine` takes the FIRST match — so when two parsed entries collide on `isTitleOwnedLine` (the duplicated-role shape #856 exists to fix), deleting the later one removes the earlier one’s line. Exact duplicates are indistinguishable, so the resulting views are identical either way; near-duplicates that still collide leave a line the SURVIVING entry’s title also owns, which `suppressTitleOwnedBullets` keeps out of the graded pool regardless. Either way the score is unaffected and the cost is at most one wrong line in the raw-text disclosure.

src/components/features/ReconstructedResume.tsx
s-annam: Deferred to **#860** — https://github.com/offlinecv/OfflineCV/issues/860. No code change in this PR, per your own recommendation.

The issue carries the full argument: the inverted gradient vs #626, why "added entries were trivially re-creatable" stops holding once parsed entries are deletable, that `resetAll` / re-upload is the only recovery, and that #856 scoped out an undo *toast* but never a *confirm*. It lists all four call sites you named — `Remove achievement` (`ReconstructedResume.tsx:1058`), `Remove project` (`:814`), `Remove role` (`ReconstructedRole.tsx:440`), `Remove education` (`ReconstructedEducationSkills.tsx:166`).

Its Reuse analysis lands on extending the #626 inline-confirm strip rather than reaching for `Dialog` — a modal for a row-level delete diverges from the very bullet path this is meant to align with.

Leaving this thread open for you to close.

src/components/features/ReconstructedResume.tsx
s-annam: Split deferred to **#861** — https://github.com/offlinecv/OfflineCV/issues/861 — taking your recommendation not to do it here.

The one part I did act on is the stale quote, since that was the concrete harm you named: `CLAUDE.md`’s known-debt line now reads `ReconstructedResume.tsx` (1489) and `ReconstructedRole.tsx` (490), replacing 1338 / 483 (`5ec090c`). The next reader gets the real number instead of one 151 lines light.

#861 carries the `fallow` breakdown (`ReconstructedResume` 370, `ExperienceSection` 320, `ProjectsSection` 143), asks for each section as a pure move in its own commit, and pins **no behaviour change** as an acceptance criterion — a test that needs editing means the move was not pure. It also requires the `CLAUDE.md` figures be refreshed again post-split, so this does not go stale a second time.

Leaving this thread open for you to close.

PR #859, opened 2026-08-16
chore(contributing): make issue claims visible and expire them

s-annam:
## Summary

Claim state for an issue lived only in the assignee field, which is invisible in every channel a contributor browses. This puts it in a label, documents the protocol, and expires claims that go quiet.

No product code is touched — `CONTRIBUTING.md` and one new workflow.

## Why

On 2026-08-16, four PRs arrived for two issues:

| Issue | PRs | Assigned at the time? |
|---|---|---|
| #681 | #853, #854 | **Yes — for 17 days** |
| #832 | #840, #855 | No |

The #681 case is the interesting one. The assignment wasn't missing; @rohithgollapalli self-assigned on 2026-07-30. Two contributors still built it independently, because:

- **`gh issue list` prints labels, not assignees.** The default table is number / state / title / labels / updated. A contributor scanning the backlog from the CLI sees no claim signal at all.
- **On the web list the assignee is a small right-margin avatar**, easy to miss and absent from some search views.
- **Both collisions were on `good first issue`.** That label selects for newcomers and for tooling that scans the backlog — precisely the population least likely to open an issue and read its sidebar before starting. One of the four PRs arrived on `manus/*` branches, which points at agent-driven contribution; that reads labels and issue text, not avatars.
- **`CONTRIBUTING.md` had no claim protocol.** 193 lines on setup, branches, commits, PRs, tests, style, filing and the board — nothing on how to pick up work. The only `assign` match in the file is an env-var assignment.

Storing the claim where nobody looks is the whole defect. Three contributors lost an evening to it in one day.

## What changes

**1. `status:claimed` label** — already created on the repo and applied to #681 and #832. It rides alongside the assignee and appears inline in every list view:

```
832  OPEN  [parser] skills — single-letter language tokens…  bug, good first issue, ux:parsing, status:claimed
```

That line is what both contributors saw before; it now carries the claim.

**2. `CONTRIBUTING.md` → "Claiming an issue"** — check the label before starting, comment to claim, a maintainer assigns and labels, what happens when a claim lapses. Also states that a maintainer-initiated assignment is a request, not an obligation.

**3. `.github/workflows/stale-claims.yml`** — daily job releasing a claim after **7 days** with no linked PR and no activity.

A claim that never expires is worse than no claim: it reads as abandoned, so people route around it and the label stops meaning anything. A 17-day-old claim is exactly what this incident ran into.

The job is deliberately narrow:

- never closes an issue, never edits a body, never changes a milestone
- **never touches a claim with an open linked PR** — the PR *is* the work, however quiet the issue went
- any activity resets the clock, so a one-line "still on it" holds a claim indefinitely
- comments whenever it releases one, saying what happened and inviting re-claim
- everything it does is undone by re-assigning and re-adding the label
- cleans up the inconsistent `status:claimed`-with-no-assignee state too

`workflow_dispatch` takes a `dry_run` input so its selection can be inspected before the cron is trusted.

## Verification

- `npm run verify` green (pre-push gate): typecheck, lint, gates, tests, build, fallow.
- fallow clean on both changed files.
- Workflow YAML parses; the embedded `github-script` body passes `node --check`.
- Label behaviour confirmed live against `gh issue list --label "good first issue"` — see the table above.

The scheduled job has not run yet. **Recommend one `workflow_dispatch` with `dry_run: true` after merge** to confirm its selection before the first real cron; it writes nothing in that mode.

## Deliberately not included

- **Removing `good first issue` while claimed.** It attacks the magnet more directly but churns labels and loses the newcomer-friendly signal on an issue that may be released a week later. Worth revisiting if collisions continue.
- **Auto-assign on a "I'll take this" comment.** Wanted eventually, but it needs intent parsing, and a wrong auto-assign is more annoying than a missing one.

## Follow-up

@rohithgollapalli's claim on #681 has been released and they've been notified on the issue; #681 and #832 are now assigned to @shubhransh-gupta, who has approved work on both.

### DIFF
--- a/.github/workflows/stale-claims.yml
+++ b/.github/workflows/stale-claims.yml
@@ -1,0 +1,160 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 The offlinecv Authors
+#
+# Expire stale issue claims.
+#
+# A claim is `status:claimed` + an assignee. It exists so a contributor
+# browsing the backlog can see an issue is taken WITHOUT opening it — the
+# assignee field is invisible in `gh issue list` and is a small avatar on the
+# web list, which is why two contributors independently built #681 while it
+# was assigned to a third.
+#
+# A claim that never expires is worse than no claim: it reads as abandoned,
+# so people route around it, and the label stops meaning anything. This job
+# drops the claim when nothing has happened for STALE_AFTER_DAYS, leaving the
+# issue open and unassigned for the next person.
+#
+# Safe by construction: it never closes, never edits the body, and never
+# touches an issue with an open linked PR. Everything it does is reversible by
+# re-assigning and re-adding the label.
+name: Expire stale claims
+
+on:
+  schedule:
+    # Daily, off the hour — GitHub drops scheduled runs at popular minutes.
+    - cron: "17 6 * * *"
+  workflow_dispatch:
+    inputs:
+      dry_run:
+        description: "Log what would expire, write nothing"
+        type: boolean
+        default: false
+
+permissions:
+  issues: write
+
+# A manual dispatch fired while the cron run is mid-flight would re-read the
+# same still-labelled issues and comment on them twice. Queue instead of
+# cancelling: an interrupted run has already unassigned some issues without
+# commenting, which is the one state this job should never leave behind.
+concurrency:
+  group: stale-claims
+  cancel-in-progress: false
+
+jobs:
+  expire:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Expire stale claims
+        uses: actions/github-script@v7
+        with:
+          script: |
+            // Days of silence before a claim lapses. A claimant who comments
+            // ("still on this") resets the clock, because any activity bumps
+            // the issue's `updated_at`.
+            const STALE_AFTER_DAYS = 7;
+            const CLAIM_LABEL = "status:claimed";
+
+            // `workflow_dispatch` inputs arrive as strings.
+            const dryRun = String(context.payload.inputs?.dry_run ?? "") === "true";
+            if (dryRun) core.info("DRY RUN — no writes will be made.");
+
+            const { owner, repo } = context.repo;
+            const cutoffMs = STALE_AFTER_DAYS * 86_400_000;
+
+            const claimed = await github.paginate(github.rest.issues.listForRepo, {
+              owner,
+              repo,
+              state: "open",
+              labels: CLAIM_LABEL,
+              per_page: 100,
+            });
+            core.info(`${claimed.length} open issue(s) carry "${CLAIM_LABEL}".`);
+
+            for (const issue of claimed) {
+              // `listForRepo` returns PRs too; a PR cannot hold a claim.
+              if (issue.pull_request) continue;
+
+              const idleMs = Date.now() - new Date(issue.updated_at).getTime();
+              const idleDays = (idleMs / 86_400_000).toFixed(1);
+
+              if (idleMs < cutoffMs) {
+                core.info(`#${issue.number}: active ${idleDays}d ago — keeping.`);
+                continue;
+              }
+
+              // An open PR IS the work. Never reap a claim that produced one,
+              // however quiet the issue itself has gone.
+              const timeline = await github.paginate(
+                github.rest.issues.listEventsForTimeline,
+                { owner, repo, issue_number: issue.number, per_page: 100 },
+              );
+              // Scoped to THIS repo on purpose. `cross-referenced` fires for a
+              // reference from any public repo on GitHub, including a PR opened
+              // on someone's fork against the fork's own `main` and never
+              // submitted here. Unscoped, any of those reads as "the work is
+              // happening" and the claim never expires.
+              const openPr = timeline.find(
+                (e) =>
+                  e.event === "cross-referenced" &&
+                  e.source?.issue?.pull_request &&
+                  e.source.issue.state === "open" &&
+                  e.source.issue.repository?.full_name === `${owner}/${repo}`,
+              );
+              if (openPr) {
+                core.info(
+                  `#${issue.number}: idle ${idleDays}d but PR #${openPr.source.issue.number} is open — keeping.`,
+                );
+                continue;
+              }
+
+              const assignees = issue.assignees.map((a) => a.login);
+              core.info(
+                `#${issue.number}: idle ${idleDays}d, no open PR, assignees [${assignees.join(", ") || "none"}] — expiring.`,
+              );
+              if (dryRun) continue;
+
+              // Per-issue isolation. Every call below can fail on one issue for
+              // reasons that say nothing about the rest of the batch — a label
+              // removed concurrently (404), a locked issue (403), a suspended
+              // assignee. Unguarded, the first such rejection ends the run and
+              // every remaining stale claim silently keeps its label until
+              // tomorrow. Log it and carry on instead.
+              try {
+                if (assignees.length > 0) {
+                  await github.rest.issues.removeAssignees({
+                    owner,
+                    repo,
+                    issue_number: issue.number,
+                    assignees,
+                  });
+                }
+                await github.rest.issues.removeLabel({
+                  owner,
+                  repo,
+                  issue_number: issue.number,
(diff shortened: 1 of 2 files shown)


.github/workflows/stale-claims.yml
```diff
@@ -0,0 +1,136 @@
...
+
+              // An open PR IS the work. Never reap a claim that produced one,
+              // however quiet the issue itself has gone.
+              const timeline = await github.paginate(
+                github.rest.issues.listEventsForTimeline,
+                { owner, repo, issue_number: issue.number, per_page: 100 },
+              );
+              const openPr = timeline.find(
+                (e) =>
+                  e.event === "cross-referenced" &&
```
s-annam, line 99: ### Unscoped cross-repository / fork PR matching in `timeline.find`

In GitHub's Issues Timeline API, `cross-referenced` events include references from *any* public repository across GitHub, as well as PRs opened on personal forks against the fork's own `main` (not submitted upstream).

In those events, `e.source.issue.pull_request` is true and `state === "open"`, but `e.source.issue.repository.full_name` is the external repository (e.g. `other-user/other-repo`). Without checking that the repository matches `${owner}/${repo}`, an open PR in an external repository or an unsubmitted fork branch referencing `offlinecv/OfflineCV#<number>` will match `openPr` and indefinitely prevent the claim from expiring in upstream.

```suggestion
              const openPr = timeline.find(
                (e) =>
                  e.event === "cross-referenced" &&
                  e.source?.issue?.pull_request &&
                  e.source.issue.state === "open" &&
                  e.source.issue.repository?.full_name === `${owner}/${repo}`,
              );
```

.github/workflows/stale-claims.yml
```diff
@@ -0,0 +1,136 @@
...
+                core.info(
+                  `#${issue.number}: idle ${idleDays}d but PR #${openPr.source.issue.number} is open — keeping.`,
+                );
+                continue;
+              }
+
+              const assignees = issue.assignees.map((a) => a.login);
+              core.info(
+                `#${issue.number}: idle ${idleDays}d, no open PR, assignees [${assignees.join(", ") || "none"}] — expiring.`,
+              );
```
s-annam, line 114: ### Loop error resilience (Unhandled API Rejections)

Inside `for (const issue of claimed)`, multiple network/API mutations (`removeAssignees`, `removeLabel`, `createComment`) execute sequentially with `await` without an enclosing `try...catch` block.

If any single issue encounters an API error (e.g. `removeLabel` throwing 404 if the label was concurrently removed, `createComment` throwing 403/422 if an issue was locked, or a suspended assignee / transient 500 error), the unhandled promise rejection terminates the entire workflow run immediately, skipping all remaining stale issues in the batch.

Wrapping the per-issue mutation block in `try { ... } catch (err) { core.error(`#${issue.number}: failed to expire claim: ${err.message}`); }` ensures one problematic issue does not halt the entire cleanup job.

.github/workflows/stale-claims.yml
```diff
@@ -0,0 +1,136 @@
...
+# so people route around it, and the label stops meaning anything. This job
+# drops the claim when nothing has happened for STALE_AFTER_DAYS, leaving the
+# issue open and unassigned for the next person.
+#
+# Safe by construction: it never closes, never edits the body, and never
+# touches an issue with an open linked PR. Everything it does is reversible by
+# re-assigning and re-adding the label.
+name: Expire stale claims
+
+on:
```
s-annam, line 22: ### Add workflow concurrency group

Consider adding a concurrency group to prevent overlapping execution if a manual `workflow_dispatch` is triggered while the scheduled cron job runs (or if two dispatches are fired in quick succession), avoiding duplicate unassignment comments:

```yaml
concurrency:
  group: stale-claims
  cancel-in-progress: false
```

CONTRIBUTING.md
```diff
@@ -136,6 +136,31 @@ These architecture rules are enforced at two layers:
...
 
+## Claiming an issue
+
+**Check the labels before you start.** An issue carrying `status:claimed`
+already has someone on it — pick another, or comment and ask where it stands.
+Do not open a PR against a claimed issue without talking to the assignee
+first; you will almost certainly duplicate their work.
+
+To take an unclaimed issue, comment on it saying you're starting. A
+maintainer will assign it to you and add `status:claimed`, which is what makes
```
s-annam, line 147: ### Clarify self-assignment for repo members / collaborators

For contributors with write access (who can self-assign), it is worth explicitly noting that they should apply both the assignee AND the `status:claimed` label, since self-assignment alone leaves the claim invisible in `gh issue list` (the exact failure mode that led to the #681 collision).

s-annam (commented):
### Review Summary

**Verdict: APPROVE** (0 Blocking findings; mechanism is sound, safe-by-construction, and fully non-destructive).

This PR introduces `status:claimed` visibility and an automated expiry workflow (`.github/workflows/stale-claims.yml`) to resolve issue-claiming collisions (like the #681 / #832 incidents). The design is clean, conservative, and reversible.

### Gates & Quality Pass
- **Verification Suite**: `npm run verify` passed cleanly (5,864 tests green, typecheck clean, lint clean, fallow clean).
- **Fixture PII**: N/A (no fixture files touched).
- **Design System / Style Tokens**: N/A (no frontend code touched).
- **Dead Code / Fallow**: Clean (0 issues in changed files).
- **Description Audit**: Accurate claims; correctly documents behavior and manual dry-run workflow.

### Findings Summary
- **Blocking**: 0
- **Secondary**: 2 (cross-repo timeline scoping, loop error resilience)
- **Nits**: 2 (workflow concurrency group, self-assignment doc polish)

---

.github/workflows/stale-claims.yml
s-annam: Fixed in `18e7093`. Added the `e.source.issue.repository?.full_name === `${owner}/${repo}`` guard as suggested, plus a comment naming the fork-PR case so the scoping does not read as redundant later.

Worth recording that I checked the suggestion before applying it, because it has a failure mode in the opposite direction: if `repository` were absent on **same-repo** cross-references, the added condition would make `openPr` never match and the job would start reaping claims that *do* have an open PR — strictly worse than the bug being fixed. It is present. Live timelines for #856 and #681 both return `repository.full_name: "offlinecv/OfflineCV"` on every same-repo `cross-referenced` event, PR and issue alike. #681's timeline currently lists this very PR as an open cross-reference, so the keep-the-claim path is exercised by real data.

.github/workflows/stale-claims.yml
s-annam: Fixed in `18e7093`. The three mutations plus the comment now sit inside one `try`, with `catch (err) { core.error(...) }` naming the issue number, so a single bad issue costs one logged error instead of the rest of the batch.

The failure is worse than "the run stops": `removeAssignees` may already have succeeded when `removeLabel` or `createComment` throws, leaving an issue unassigned, still labelled, and with no comment explaining anything. Continuing the loop does not repair that instance, but it does stop one such issue from holding every other stale claim hostage until the next daily run.

.github/workflows/stale-claims.yml
s-annam: Fixed in `18e7093` — added at the top level, next to `permissions`:

```yaml
concurrency:
  group: stale-claims
  cancel-in-progress: false
```

Took `cancel-in-progress: false` deliberately rather than the default. Cancelling mid-run is the one thing this job must not do: the per-issue sequence is unassign → unlabel → comment, so a kill between the first call and the last leaves an issue stripped of its claim with no comment saying why — exactly the "maintainer snub" the comment text exists to prevent. Queueing costs a few idle minutes; cancelling costs a contributor their explanation. Noted the reasoning in a comment above the block.

CONTRIBUTING.md
s-annam: Fixed in `18e7093`. Added directly under the two-signals paragraph:

> **If you have write access and assign yourself, add `status:claimed` too.**
> Self-assignment skips the maintainer step that would normally attach the label,
> and an assignee on its own is exactly the invisible claim described above — it
> is how #681 got built twice.

Placed there rather than in its own subsection so it reads as the corollary to the two-signals rule immediately above it, which is where the reasoning already lives.

PR #865, opened 2026-08-18
fix(parser): ignore language proficiency rows

### LINKED ISSUE #833: [parser] skills — a "Language: Fluent in Spanish" proficiency row is admitted as a skill token

qtjg:
Resolves #833

## Summary

Ignore spoken-language proficiency rows such as `Language: Fluent in Spanish` when extracting résumé skills, while preserving programming-language rows and bare spoken-language lists. The label and proficiency wording are checked together, so `Languages: Python, Go, TypeScript` remains unchanged and non-language labels such as `Certifications: Certified in AWS Solutions Architecture` are unaffected.

The change adds focused regression coverage, removes the resolved #833 truth exception, and updates the affected Google Docs and Word corpus snapshots.

## Verification

- `npx vitest run src/lib/heuristics/extract/skills.test.ts` — 49 passed
- `npm run typecheck` — passed
- `npx vitest run src/lib/heuristics/corpus.test.ts src/lib/heuristics/corpus-roundtrip.test.ts` — passed
- `npm run lint` — passed
- `npm run check:fixtures` — passed
- `npm run check:baselines` — passed
- `npm run check:core` — passed
- `npm run build` — passed

The full repository verification also reproduces an unrelated existing `BroadcastChannel`/`MessageEvent` failure in `src/hooks/useLibraryChanges.test.tsx`; 358 test files passed and 1 failed.

### DIFF
--- a/src/lib/heuristics/extract/skills.ts
+++ b/src/lib/heuristics/extract/skills.ts
@@ -312,13 +312,35 @@ const BARE_SUBLABEL_RE = new RegExp(`^${SUBLABEL_BODY}:$`);
 const LANGUAGE_LABEL_RE =
   /^(?:foreign\s+|spoken\s+|other\s+)?languages?\s*$/i;
 
-/** A spoken-language proficiency predication, distinguished from a delimited
- * programming-language list by its proficiency wording rather than by the
- * label. Keep this vocabulary broad enough to cover standard proficiency
- * scales, but apply it per fragment so `Visual Basic` and `React Native` do
- * not cause an entire programming-language row to disappear. */
-const LANGUAGE_PROFICIENCY_BODY_RE =
-  /\b(fluent|native|bilingual|conversational|proficient|proficiency|proficiencies|intermediate|elementary|limited|beginner|basic|working\s+proficiency|mother\s+tongue)\b/i;
+/** Spoken-language proficiency ADJECTIVES/predicates ("fluent", "native",
+ * "basic", …). Unlike the NOUN forms below, these collide with real compound
+ * proper nouns — "Visual Basic", "React Native" — so a bare substring test
+ * over a fragment is not enough; see LANGUAGE_PROFICIENCY_PREDICATION_RE for
+ * the structural test that disambiguates them. */
+const LANGUAGE_PROFICIENCY_ADJECTIVES =
+  "fluent|native|bilingual|conversational|proficient|intermediate|elementary|limited|beginner|basic|mother\\s+tongue";
+
+/** A single fragment that reads as a spoken-language proficiency PREDICATION
+ * ("Fluent in Spanish", "Native German", "Spanish (native)", "Native
+ * speaker") rather than a token that merely CONTAINS proficiency vocabulary
+ * ("Visual Basic", "React Native", bare "BASIC"). The noun forms
+ * "proficiency"/"proficiencies" are unambiguous in any position — no
+ * programming-language or tool name ends in "Proficiency" — so they match
+ * anywhere. The adjective forms only count when they carry predicate
+ * structure: leading the fragment with something after them ("Native
+ * German", ruling out a bare "BASIC" with nothing to modify), joined by "in"
+ * ("Fluent in Spanish"), parenthesised ("Spanish (native)"), or followed by
+ * "speaker" ("Native speaker"). Applied per fragment so a delimited
+ * programming-language list survives even when every one of its items
+ * happens to contain proficiency vocabulary. */
+const LANGUAGE_PROFICIENCY_PREDICATION_RE = new RegExp(
+  "\\b(?:proficiency|proficiencies)\\b" +
+    `|^(?:${LANGUAGE_PROFICIENCY_ADJECTIVES})\\s+\\S` +
+    `|\\b(?:${LANGUAGE_PROFICIENCY_ADJECTIVES})\\s+in\\b` +
+    `|\\((?:${LANGUAGE_PROFICIENCY_ADJECTIVES})\\)` +
+    `|\\b(?:${LANGUAGE_PROFICIENCY_ADJECTIVES})\\s+speaker\\b`,
+  "i",
+);
 
 function isLanguageProficiencyCell(cell: string): boolean {
   const debulleted = stripBullet(cell);
@@ -330,7 +352,9 @@ function isLanguageProficiencyCell(cell: string): boolean {
     .filter((fragment) => fragment !== "");
   return (
     fragments.length > 0 &&
-    fragments.every((fragment) => LANGUAGE_PROFICIENCY_BODY_RE.test(fragment))
+    fragments.every((fragment) =>
+      LANGUAGE_PROFICIENCY_PREDICATION_RE.test(fragment),
+    )
   );
 }
 
@@ -435,6 +459,19 @@ function isSoftWrapContinuation(
   if (looksLikeContactLink(nextText)) return false;
   // Always skip if the next line starts a new sub-list (label or bullet).
   if (SKILLS_NEW_ENTRY_RE.test(nextText)) return false;
+  // Always skip when `pending` ALONE already resolves as a complete
+  // spoken-language proficiency cell (#833) — e.g. "Languages: Fluent in
+  // Spanish, conversational French". Such a cell is grammatically finished,
+  // so a following comma-bearing line is an independent, unrelated entry
+  // ("Project Management, Agile, Scrum"), not its wrap continuation. Without
+  // this guard, Condition B below joins the two into one cell whose merged
+  // fragment list contains real skill tokens, defeating
+  // `isLanguageProficiencyCell`'s per-fragment check and re-admitting the
+  // proficiency phrase as a skill. A cell that genuinely wraps mid-list ends
+  // on a trailing comma/semicolon instead ("English (native), Spanish
+  // (fluent),"), which `isLanguageProficiencyCell` — and this guard — leaves
+  // alone because a trailing delimiter never resolves as a complete cell.
+  if (isLanguageProficiencyCell(pending)) return false;
 
   // Condition A′ (symmetric to A): the NEXT line LEADS with a bare connector
   // glyph — a soft-wrap that broke a skill/list immediately BEFORE the connector
@@ -688,12 +725,22 @@ export function extractSkills(
       cellTokens.push(tok);
     }
     const label = matchCellLabel(cell);
-    // A dropped language-proficiency row still opens a category boundary. Keep
-    // that anchor so a soft-wrapped tail cannot be misfiled under the previous
-    // category; the category is emitted only if a later continuation contributes
-    // a token.
+    // A dropped language-proficiency row still opens a category boundary, but
+    // ONLY when the row is itself unterminated (ends on a trailing
+    // comma/semicolon, e.g. "English (native), Spanish (fluent),") — the same
+    // "list got cut off mid-item" signal `isSoftWrapContinuation`'s Condition
+    // B uses. That is what distinguishes a genuine soft-wrapped tail ("…,"
+    // then "French") from an unrelated, independent bare cell that merely
+    // happens to follow the dropped row ("Languages: Fluent in Spanish" then
+    // "Python, Go, TypeScript"): the latter is already a complete cell with no
+    // trailing delimiter, so nothing after it can be "its" continuation. The
+    // category is emitted only if a later continuation contributes a token.
     if (cellTokens.length === 0) {
-      if (label !== undefined && isLanguageProficiencyCell(cell)) {
+      if (
+        label !== undefined &&
+        isLanguageProficiencyCell(cell) &&
+        /[,;]\s*$/.test(stripBullet(cell))
+      ) {
         categories.push({ label, skills: [] });
       }
       continue;
--- a/src/lib/heuristics/extract/skills.ts
+++ b/src/lib/heuristics/extract/skills.ts
@@ -304,6 +304,36 @@ const SUBLABEL_PREFIX_RE = new RegExp(`^(${SUBLABEL_BODY}):\\s*`);
  *  tore off its body. See the rejoin in `splitColumnCells`. */
 const BARE_SUBLABEL_RE = new RegExp(`^${SUBLABEL_BODY}:$`);
 
+/** A Skills sub-label that MAY head a spoken-language row. Deliberately NOT
+ * added to NON_SKILL_SUBLABEL_RE: on most engineering résumés `Languages:`
+ * heads the programming-language row, so the label is ambiguous and the body
+ * must decide whether this is a proficiency statement. Qualifiers are allowed
+ * because `Foreign Languages:` and `Spoken Languages:` are common variants. */
+const LANGUAGE_LABEL_RE =
+  /^(?:foreign\s+|spoken\s+|other\s+)?languages?\s*$/i;
+
+/** A spoken-language proficiency predication, distinguished from a delimited
+ * programming-language list by its proficiency wording rather than by the
+ * label. Keep this vocabulary broad enough to cover standard proficiency
+ * scales, but apply it per fragment so `Visual Basic` and `React Native` do
+ * not cause an entire programming-language row to disappear. */
+const LANGUAGE_PROFICIENCY_BODY_RE =
+  /\b(fluent|native|bilingual|conversational|proficient|proficiency|proficiencies|intermediate|elementary|limited|beginner|basic|working\s+proficiency|mother\s+tongue)\b/i;
+
+function isLanguageProficiencyCell(cell: string): boolean {
+  const debulleted = stripBullet(cell);
+  const match = debulleted.match(SUBLABEL_PREFIX_RE);
+  if (!match || !LANGUAGE_LABEL_RE.test(match[1])) return false;
+
+  const fragments = splitRespectingParens(debulleted.slice(match[0].length))
+    .map((fragment) => fragment.trim())
+    .filter((fragment) => fragment !== "");
+  return (
+    fragments.length > 0 &&
+    fragments.every((fragment) => LANGUAGE_PROFICIENCY_BODY_RE.test(fragment))
+  );
+}
+
 /**
  * Tokenizes a single column cell into valid skill tokens and adds them to
  * `out`. Drops the cell entirely when it looks like a contact/profile link —
@@ -317,7 +347,11 @@ function tokenizeCell(cell: string, out: Set<string>): void {
   // leading `Label:` prefix and drop the whole cell when it names a
   // hobbies/interests list — before the label is stripped and the items split.
   const labelMatch = debulleted.match(SUBLABEL_PREFIX_RE);
-  if (labelMatch && NON_SKILL_SUBLABEL_RE.test(labelMatch[1])) return;
+  if (
+    labelMatch &&
+    (NON_SKILL_SUBLABEL_RE.test(labelMatch[1]) || isLanguageProficiencyCell(debulleted))
+  )
+    return;
   const clean = debulleted.replace(SUBLABEL_PREFIX_RE, "");
   // A whole cell that is a profile link ("github.com/janesmith") must be
   // dropped before splitting — a path slash would otherwise leave the path
@@ -613,10 +647,16 @@ function upcomingContinuationTexts(lineCells: string[][], from: number): string[
  * A non-skill sub-label (Interests/Hobbies) is not a category: `tokenizeCell`
  * drops such a cell whole, so it never contributes tokens and never reaches the
  * caller's category branch — but the guard here keeps the two decisions aligned.
+ * Language-proficiency rows retain their label here so a dropped row can still
+ * anchor a soft-wrapped continuation in `extractSkills`.
  */
 function matchCellLabel(cell: string): string | undefined {
   const m = stripBullet(cell).match(SUBLABEL_PREFIX_RE);
-  if (!m || NON_SKILL_SUBLABEL_RE.test(m[1])) return undefined;
+  if (
+    !m ||
+    NON_SKILL_SUBLABEL_RE.test(m[1])
+  )
+    return undefined;
   return m[1].trim();
 }
 
@@ -647,11 +687,21 @@ export function extractSkills(
       value.push(tok);
       cellTokens.push(tok);
     }
+    const label = matchCellLabel(cell);
+    // A dropped language-proficiency row still opens a category boundary. Keep
+    // that anchor so a soft-wrapped tail cannot be misfiled under the previous
+    // category; the category is emitted only if a later continuation contributes
+    // a token.
+    if (cellTokens.length === 0) {
+      if (label !== undefined && isLanguageProficiencyCell(cell)) {
+        categories.push({ label, skills: [] });
+      }
+      continue;
+    }
+
     // A cell that produced nothing (dropped link/hobbies label, or all-duplicate)
     // is neither a category nor a bare contribution — skip it on both axes.
-    if (cellTokens.length === 0) continue;
 
-    const label = matchCellLabel(cell);
     if (label !== undefined) {
       categories.push({ label, skills: cellTokens });
     } else if (categories.length > 0) {
@@ -666,13 +716,18 @@ export function extractSkills(
     }
   }
 
+  const nonEmptyCategories = categories.filter(
+    (category) => category.skills.length > 0,
+  );
   const confidence = value.length >= 5 ? 0.85 : value.length >= 2 ? 0.6 : 0.2;
   return {
     value,
     // Emit the structured view ONLY when the section is fully categorised: at
-    // least one label AND no bare head. Otherwise the field is absent (never
-    // `[]`), which reads as "uncategorised" per invariant (2).
-    ...(categories.length > 0 && !hasUncategorisedHead ? { categories } : {}),
+    // least one non-empty label AND no bare head. Otherwise the field is absent
+    // (never `[]`), which reads as "uncategorised" per invariant (2).
+    ...(nonEmptyCategories.length > 0 && !hasUncategorisedHead
+      ? { categories: nonEmptyCategories }
+      : {}),
     confidence,
   };
 }
--- a/src/lib/heuristics/extract/skills.test.ts
+++ b/src/lib/heuristics/extract/skills.test.ts
@@ -471,6 +471,22 @@ describe("tokenizeSkillLine — issue #833 language proficiency rows", () => {
     ]);
   });
 
+  it("keeps a SOLE skill whose name contains proficiency wording (single-fragment row)", () => {
+    // A single-item row has no siblings to protect it from a whole-cell drop —
+    // the per-fragment check must reject "Visual Basic"/"BASIC" on their own
+    // structural shape (no leading proficiency adjective with an object, no
+    // "in", no parens, no "speaker"), not merely because they are outnumbered.
+    expect(tokenizeSkillLine("Languages: Visual Basic")).toEqual(["Visual Basic"]);
+    expect(tokenizeSkillLine("Languages: BASIC")).toEqual(["BASIC"]);
+    expect(tokenizeSkillLine("Languages: React Native")).toEqual(["React Native"]);
+  });
+
+  it("still drops a single-item genuine proficiency predication", () => {
+    expect(tokenizeSkillLine("Languages: Native German")).toEqual([]);
+    expect(tokenizeSkillLine("Languages: Spanish (native)")).toEqual([]);
+    expect(tokenizeSkillLine("Languages: Native speaker")).toEqual([]);
+  });
+
   it("keeps bare spoken-language lists and non-language proficiency skills", () => {
     expect(tokenizeSkillLine("Languages: Python, Go, TypeScript")).toEqual([
       "Python",
@@ -896,6 +912,45 @@ describe("extractSkills — language proficiency category boundary (#833)", () =
       { label: "Frameworks", skills: ["React", "Vue"] },
       { label: "Languages", skills: ["French"] },
     ]);
+  });
+
+  it("does not join a complete proficiency row with a following, unrelated skill line", () => {
+    // "Languages: Fluent in Spanish, conversational French" is a complete cell
+    // on its own (no trailing comma) — a following comma-bearing but unrelated
+    // line must not be pulled into it by the soft-wrap Condition B join, or the
+    // merged fragment list re-admits the proficiency phrase and garbles a token
+    // across the line break.
+    const section = skillsLines([
+      [
+        {
+          x: 74,
+          str: "Languages: Fluent in Spanish, conversational French",
+          w: 220,
+        },
+      ],
+      [{ x: 74, str: "Project Management, Agile, Scrum", w: 160 }],
+    ]);
+    const { value } = extractSkills(section);
+
+    expect(value).toEqual(["Project Management", "Agile", "Scrum"]);
+    expect(value.some((v) => v.includes("French"))).toBe(false);
+  });
+
+  it("does not misfile an unrelated bare list under a dropped row's anchor category", () => {
+    // "Languages: Fluent in Spanish" is a single item with no trailing comma —
+    // a COMPLETE cell, not an unterminated one — so nothing after it is "its"
+    // wrap continuation. An unrelated bare list that merely follows it must
+    // not be swept under a "Languages" category the dropped row left behind.
+    const section = skillsLines([
+      [{ x: 74, str: "Languages: Fluent in Spanish", w: 160 }],
+      [{ x: 74, str: "Python, Go, TypeScript", w: 130 }],
+    ]);
+    const { value, categories } = extractSkills(section);
+
+    expect(value).toEqual(["Python", "Go", "TypeScript"]);
+    expect(
+      categories?.find((c) => c.label === "Languages")?.skills ?? [],
+    ).not.toEqual(expect.arrayContaining(["Python", "Go", "TypeScript"]));
   });
 });
 
--- a/src/lib/heuristics/extract/skills.test.ts
+++ b/src/lib/heuristics/extract/skills.test.ts
@@ -426,6 +426,68 @@ describe("tokenizeSkillLine — issue #221 non-skill sub-labels", () => {
   });
 });
 
+describe("tokenizeSkillLine — issue #833 language proficiency rows", () => {
+  it("drops spoken-language proficiency rows across common label variants", () => {
+    expect(tokenizeSkillLine("Language: Fluent in Spanish")).toEqual([]);
+    expect(tokenizeSkillLine("Foreign Languages: Fluent in Spanish")).toEqual([]);
+    expect(
+      tokenizeSkillLine("Spoken Languages: Native German, conversational French"),
+    ).toEqual([]);
+    expect(tokenizeSkillLine("Languages : Fluent in Spanish")).toEqual([]);
+  });
+
+  it("drops standard proficiency-scale wording", () => {
+    expect(
+      tokenizeSkillLine("Languages: Full professional proficiency in German"),
+    ).toEqual([]);
+    expect(tokenizeSkillLine("Languages: Elementary proficiency in French")).toEqual(
+      [],
+    );
+    expect(tokenizeSkillLine("Languages: JLPT N2 Japanese Proficiency")).toEqual(
+      [],
+    );
+  });
+
+  it("keeps programming-language rows when one token contains proficiency wording", () => {
+    expect(tokenizeSkillLine("Languages: C, C++, Visual Basic")).toEqual([
+      "C",
+      "C++",
+      "Visual Basic",
+    ]);
+    expect(tokenizeSkillLine("Languages: Kotlin, Swift, React Native")).toEqual([
+      "Kotlin",
+      "Swift",
+      "React Native",
+    ]);
+    expect(tokenizeSkillLine("Languages: Proficient in Java, Python, Go")).toEqual([
+      "Proficient in Java",
+      "Python",
+      "Go",
+    ]);
+    expect(tokenizeSkillLine("Languages: Java, Python (proficient), Go")).toEqual([
+      "Java",
+      "Python (proficient)",
+      "Go",
+    ]);
+  });
+
+  it("keeps bare spoken-language lists and non-language proficiency skills", () => {
+    expect(tokenizeSkillLine("Languages: Python, Go, TypeScript")).toEqual([
+      "Python",
+      "Go",
+      "TypeScript",
+    ]);
+    expect(tokenizeSkillLine("Languages: Spanish, French, Mandarin")).toEqual([
+      "Spanish",
+      "French",
+      "Mandarin",
+    ]);
+    expect(
+      tokenizeSkillLine("Certifications: Certified in AWS Solutions Architecture"),
+    ).toEqual(["Certified in AWS Solutions Architecture"]);
+  });
+});
+
 describe("tokenizeSkillLine — issue #832 single-letter languages", () => {
   it("keeps C, R, and D as real programming languages", () => {
     expect(tokenizeSkillLine("Languages: C, R, D")).toEqual(
@@ -820,6 +882,23 @@ describe("extractSkills — bulleted labelled single-column rows (#465)", () => 
   });
 });
 
+describe("extractSkills — language proficiency category boundary (#833)", () => {
+  it("does not misfile a wrapped tail under the previous category", () => {
+    const section = skillsLines([
+      [{ x: 74, str: "Frameworks: React, Vue", w: 110 }],
+      [{ x: 74, str: "Languages: English (native), Spanish (fluent),", w: 180 }],
+      [{ x: 74, str: "French", w: 35 }],
+    ]);
+    const { value, categories } = extractSkills(section);
+
+    expect(value).toEqual(["React", "Vue", "French"]);
+    expect(categories).toEqual([
+      { label: "Frameworks", skills: ["React", "Vue"] },
+      { label: "Languages", skills: ["French"] },
+    ]);
+  });
+});
+
 describe("extractSkills — category capture (#473)", () => {
   it("captures each label as a category and keeps `skills` the flat union", () => {
     const section = skillsLines([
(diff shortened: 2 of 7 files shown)


src/lib/heuristics/extract/skills.ts
```diff
@@ -304,6 +304,25 @@ const SUBLABEL_PREFIX_RE = new RegExp(`^(${SUBLABEL_BODY}):\\s*`);
...
+ * programming-language list by its proficiency wording rather than by the
+ * label. */
+const LANGUAGE_PROFICIENCY_BODY_RE =
+  /\b(fluent|native|bilingual|conversational|proficient|intermediate|beginner|basic|working\s+proficiency|mother\s+tongue)\b/i;
+
+function isLanguageProficiencyCell(cell: string): boolean {
+  const debulleted = stripBullet(cell);
+  const match = debulleted.match(SUBLABEL_PREFIX_RE);
+  if (!match || !LANGUAGE_LABEL_RE.test(match[1])) return false;
+  return LANGUAGE_PROFICIENCY_BODY_RE.test(debulleted.slice(match[0].length));
```
s-annam: **Blocking.** This tests the proficiency vocabulary against the *whole* body, so a delimited programming-language row containing one matching word is dropped entirely — and `tokenizeCell` drops the cell before the split, so every sibling token on the row dies too.

Measured on this branch vs `origin/main`:

| Input | `origin/main` | this branch |
|---|---|---|
| `Languages: C, C++, Visual Basic` | `["C","C++","Visual Basic"]` | `[]` |
| `Languages: Kotlin, Swift, React Native` | 3 tokens | `[]` |
| `Languages: Proficient in Java, Python, Go` | `["Proficient in Java","Python","Go"]` | `[]` |
| `Languages: Java, Python (proficient), Go` | `["Java","Python (proficient)","Go"]` | `[]` |

`\bbasic\b` catches **Visual Basic**, `\bnative\b` catches **React Native**, `\bproficient\b` catches the common `Proficient in <list>` phrasing. The label is effectively deciding, which is what #833 ruled out.

#833 step 1 specifies the body conjunct as *"a proficiency predication **rather than a delimited list**"* — the delimited-list half is missing. Requiring **every** delimited fragment to be a proficiency predication satisfies all four ACs and rejects all four rows above:

```ts
const fragments = debulleted
  .slice(match[0].length)
  .split(SKILL_SPLIT_RE)
  .map((f) => f.trim())
  .filter((f) => f !== "");
return fragments.length > 0 && fragments.every((f) => LANGUAGE_PROFICIENCY_BODY_RE.test(f));
```

(`splitRespectingParens` is what handles commas — worth reusing the existing splitter.) Please pin the four rows as unit assertions; the new test covers only clean rows.

src/lib/heuristics/extract/skills.ts
```diff
@@ -317,7 +336,11 @@ function tokenizeCell(cell: string, out: Set<string>): void {
   // leading `Label:` prefix and drop the whole cell when it names a
   // hobbies/interests list — before the label is stripped and the items split.
   const labelMatch = debulleted.match(SUBLABEL_PREFIX_RE);
-  if (labelMatch && NON_SKILL_SUBLABEL_RE.test(labelMatch[1])) return;
+  if (
+    labelMatch &&
+    (NON_SKILL_SUBLABEL_RE.test(labelMatch[1]) || isLanguageProficiencyCell(debulleted))
```
s-annam, line 376: **Blocking.** Returning here yields zero tokens, so `extractSkills` `continue`s at `skills.ts:680` before `matchCellLabel` runs and no `Languages` category is ever opened. A soft-wrapped continuation then falls into the "bare cell extends the last category" branch at `skills.ts:691` and is filed under the **previous** label.

`Frameworks: React, Vue` / `Languages: English (native), Spanish (fluent),` / `French`:

- `origin/main` → `Frameworks:[React,Vue]`, `Languages:[English (native),Spanish (fluent),French]`
- this branch → `Frameworks:[React,Vue,`**`French`**`]`

`French` becomes a Framework — silently wrong data rather than missing data, on a surface that feeds JD-match and job-search keywords.

Note the all-fragments fix for the other blocker does **not** resolve this: `English (native), Spanish (fluent),` is still legitimately dropped. Either register the category label before the zero-token `continue`, or drop the offending fragments rather than the whole cell — the latter makes both blockers fall out of one change.

src/lib/heuristics/extract/skills.ts
```diff
@@ -304,6 +304,25 @@ const SUBLABEL_PREFIX_RE = new RegExp(`^(${SUBLABEL_BODY}):\\s*`);
  *  tore off its body. See the rejoin in `splitColumnCells`. */
 const BARE_SUBLABEL_RE = new RegExp(`^${SUBLABEL_BODY}:$`);
 
+/** A Skills sub-label that MAY head a spoken-language row. Deliberately NOT
+ * added to NON_SKILL_SUBLABEL_RE: on most engineering résumés `Languages:`
+ * heads the programming-language row, so the label is ambiguous and the body
+ * must decide whether this is a proficiency statement. */
+const LANGUAGE_LABEL_RE = /^languages?$/i;
```
s-annam: **Secondary.** Stricter than its sibling `NON_SKILL_SUBLABEL_RE`, so #833's defect survives on common variants. All still admitted on this branch:

- `Foreign Languages: Fluent in Spanish` → `["Fluent in Spanish"]`
- `Spoken Languages: Native German, conversational French` → `["Native German","conversational French"]`
- `Languages : Fluent in Spanish` → `["Fluent in Spanish"]`

The third is a capture artifact: `SUBLABEL_BODY` (`[A-Z][A-Za-z &/]+`) admits a space, so the capture is `"Languages "`. `NON_SKILL_SUBLABEL_RE` carries a deliberate `\s*$` for exactly this and `matchCellLabel` `.trim()`s the capture — this regex does neither, so `Interests : Tennis` is dropped while `Languages : Fluent in Spanish` escapes.

`/^(?:foreign\s+|spoken\s+|other\s+)?languages?\s*$/i` mirrors the sibling's leading-qualifier tolerance and covers all three.

src/lib/heuristics/extract/skills.ts
```diff
@@ -304,6 +304,25 @@ const SUBLABEL_PREFIX_RE = new RegExp(`^(${SUBLABEL_BODY}):\\s*`);
...
+
+/** A spoken-language proficiency predication, distinguished from a delimited
+ * programming-language list by its proficiency wording rather than by the
+ * label. */
+const LANGUAGE_PROFICIENCY_BODY_RE =
+  /\b(fluent|native|bilingual|conversational|proficient|intermediate|beginner|basic|working\s+proficiency|mother\s+tongue)\b/i;
```
s-annam: **Secondary.** The vocabulary misses the standard scale wordings: `\bproficient\b` does not match the noun *proficiency*, and only `working\s+proficiency` is enumerated. Still admitted:

- `Languages: Full professional proficiency in German`
- `Languages: Elementary proficiency in French`
- `Languages: JLPT N2 Japanese Proficiency`

The last is drawn by two committed fixtures — `google-docs/google-docs-skia-proxy-multiline-bullets-coursework.pdf` and `unknown/student-projects-activities-singlecol.pdf` (verified with `pdftotext`) — and both baselines are untouched by this PR, confirming the row still reaches `skills`.

Adding `proficienc(y|ies)` plus `elementary` / `limited` closes it. Worth doing here rather than as a follow-up, since the fixtures already exist.

src/lib/heuristics/extract/skills.ts
```diff
@@ -616,7 +639,12 @@ function upcomingContinuationTexts(lineCells: string[][], from: number): string[
  */
 function matchCellLabel(cell: string): string | undefined {
   const m = stripBullet(cell).match(SUBLABEL_PREFIX_RE);
-  if (!m || NON_SKILL_SUBLABEL_RE.test(m[1])) return undefined;
+  if (
+    !m ||
+    NON_SKILL_SUBLABEL_RE.test(m[1]) ||
+    isLanguageProficiencyCell(cell)
```
s-annam: **Nit.** Unreachable: `extractSkills:680` `continue`s on a zero-token cell and `tokenizeCell` already returned early for exactly these, so `matchCellLabel` — which has no other callers — can never be handed one. Fine to keep as defensive alignment with the `NON_SKILL_SUBLABEL_RE` mirror, but the docblock just above (`skills.ts:632`) explains only the Interests/Hobbies mirror and should name this guard too.

Related: this passes the raw `cell` while `skills.ts:341` passes the already-`stripBullet`ed text. Correct either way since `stripBullet` is idempotent, but taking the captured label and body as parameters would make the "two decisions stay aligned" contract structural rather than by-convention.

s-annam (changes requested):
The precision goal is right and the corpus moves the right way — `Fluent in Spanish` is gone from both fixtures and the `#833` truth exception clears cleanly against an exactly-matching 11-entry truth. But the discriminator is a bare substring test over the whole body, and it takes the whole cell with it. Two confirmed regressions follow, and neither is visible to `verify` because no fixture happens to pair the shapes.

Verdict rule: **≥1 Blocking → `REQUEST_CHANGES`.** 2 Blocking, 2 Secondary, 1 Nit. Nothing was committed or pushed.

Every claim below was measured on this branch against `origin/main` with throwaway probes over the real `extractSkills` / `tokenizeSkillLine` (probe files removed; tree left clean).

## Blocking

### 1. `LANGUAGE_PROFICIENCY_BODY_RE` matches inside real programming-language names, and the drop is whole-cell — an entire skills row vanishes

`\bbasic\b` fires on **Visual Basic**, `\bnative\b` on **React Native**, and `\bproficient\b` on the ubiquitous `Proficient in <list>` phrasing. Because the drop happens in `tokenizeCell` *before* the split, every sibling token on the row dies with the offender:

| Input | `origin/main` | this branch |
|---|---|---|
| `Languages: C, C++, Visual Basic` | `["C","C++","Visual Basic"]` | `[]` |
| `Languages: Kotlin, Swift, React Native` | 3 tokens | `[]` |
| `Languages: Proficient in Java, Python, Go` | `["Proficient in Java","Python","Go"]` | `[]` |
| `Languages: Java, Python (proficient), Go` | `["Java","Python (proficient)","Go"]` | `[]` |

The label conjunct is `Languages` in all four, so the *label* is effectively deciding — the exact failure #833's root-cause section ruled out: *"denying the label would delete the single most valuable row on the page."* This is a strictly larger loss than the one token the fix targets.

Not hypothetical on this corpus. `openresume-laverne-word-quartz.pdf` — the fixture this PR re-snapshots — draws `Computer: Proficient in Windows and Mac OS, Microsoft Word, PowerPoint, and Excel`, spared only by its label. `multi-degree-coursework.pdf` draws `Languages: Python, Go, C, C++, Java, JavaScript, Swift, PHP, HTML, CSS` — one `Visual Basic` away from returning nothing.

**Why this reads as under-implementing the plan, not a judgement call:** #833 step 1 specifies the body conjunct as *"whose body reads as a proficiency predication **rather than a delimited list**"*. Only the proficiency-word half shipped; the delimited-list half is missing. The regexes in the issue illustrate that two-part rule rather than replacing it.

One shape that satisfies every AC — drop only when **every** delimited fragment is a proficiency predication:

```ts
const fragments = debulleted
  .slice(match[0].length)
  .split(SKILL_SPLIT_RE)
  .map((f) => f.trim())
  .filter((f) => f !== "");
// A list whose fragments are mostly plain tokens is a programming-language row
// that happens to contain "Basic"/"React Native"/"(proficient)"; only an
// all-proficiency body is a spoken-language proficiency statement.
return fragments.length > 0 && fragments.every((f) => LANGUAGE_PROFICIENCY_BODY_RE.test(f));
```

Against the ACs: `Fluent in Spanish` 1/1 → dropped. `Python, Go, TypeScript` 0/3 → kept. `Spanish, French, Mandarin` 0/3 → kept (the stated limit, preserved). `C, C++, Visual Basic` 1/3 → kept. `Proficient in Java, Python, Go` 1/3 → kept. Note `splitRespectingParens` is what handles commas — reuse the existing splitter rather than a new one.

Whatever shape you land on, please pin the four rows above as unit assertions. The new test at `skills.test.ts:838` covers only clean rows, so nothing currently guards this.

### 2. Dropping the cell destroys the category anchor, so a soft-wrapped proficiency row misfiles its tail into the *previous* category

Three lines — `Frameworks: React, Vue` / `Languages: English (native), Spanish (fluent),` / `French`:

| | `value` | `categories` |
|---|---|---|
| `origin/main` | `["React","Vue","English (native)","Spanish (fluent)","French"]` | `Frameworks:[React,Vue]`, `Languages:[English (native),Spanish (fluent),French]` |
| this branch | `["React","Vue","French"]` | `Frameworks:[React,Vue,`**`French`**`]` |

`French` is now a Framework. The mechanism: the language cell yields zero tokens, so `extractSkills` `continue`s at `skills.ts:680` before `matchCellLabel` ever runs, no `Languages` category is opened, and the wrapped continuation falls into the "bare cell extends the last category" branch at `skills.ts:691`. Any wrap `isSoftWrapContinuation` declines reproduces it.

This is worse than the loss in #1 — it is silently *wrong* data rather than missing data, and `skills` feeds the JD-match and job-search keyword surfaces the issue itself calls load-bearing.

**Fixing #1 does not fix this.** Under the all-fragments rule above, `English (native), Spanish (fluent),` is still legitimately dropped, so the leak survives. It needs its own answer — either register the category label before the token-count `continue`, or drop the offending fragments rather than the cell — which is also the cleanest way to make #1 and #2 fall out of one change.

## Secondary

### 3. `LANGUAGE_LABEL_RE` is stricter than its sibling, so the targeted defect survives on common label variants

`/^languages?$/i` tested against the untrimmed capture. Measured, all still admitted on this branch:

- `Foreign Languages: Fluent in Spanish` → `["Fluent in Spanish"]`
- `Spoken Languages: Native German, conversational French` → `["Native German","conversational French"]`
- `Languages : Fluent in Spanish` (space before colon) → `["Fluent in Spanish"]`

The first two are #833's defect, unfixed. The third is a capture artifact: `SUBLABEL_BODY` (`[A-Z][A-Za-z &/]+`) admits a space, so the capture is `"Languages "`. `NON_SKILL_SUBLABEL_RE` carries a deliberate `\s*$` for exactly this and `matchCellLabel` `.trim()`s — this regex does neither, so `Interests : Tennis` is correctly dropped while `Languages : Fluent in Spanish` escapes. `/^(?:foreign\s+|spoken\s+|other\s+)?languages?\s*$/i`, mirroring the sibling's leading-qualifier tolerance, covers all three.

### 4. The proficiency vocabulary misses the standard scale wordings — including one already in the corpus

`\bproficient\b` does not match the noun *proficiency*, and only `working\s+proficiency` is enumerated. Still admitted:

- `Languages: Full professional proficiency in German`
- `Languages: Elementary proficiency in French`
- `Languages: JLPT N2 Japanese Proficiency`

That last one is drawn by **two committed fixtures** — `google-docs/google-docs-skia-proxy-multiline-bullets-coursework.pdf` and `unknown/student-projects-activities-singlecol.pdf` (verified with `pdftotext`) — and both baselines are untouched by this PR, confirming the row still lands in `skills`. `proficienc(y|ies)`, plus `elementary` / `limited`, closes it. Worth doing here rather than as a follow-up, since the fixtures are already in the corpus.

## Nits

Non-blocking; neither changes the verdict.

5. **The `isLanguageProficiencyCell(cell)` guard in `matchCellLabel` is unreachable, and the docblock above it no longer describes the guard set.** `extractSkills:680` `continue`s on a zero-token cell and `tokenizeCell` already returned early for exactly these, so `matchCellLabel` — which has no other callers — can never see one. Harmless as defensive alignment with the existing `NON_SKILL_SUBLABEL_RE` mirror, and worth keeping for that reason, but the docblock at `skills.ts:632` explains only the Interests/Hobbies mirror and should name this one too. Related: this call site passes the raw `cell` while `skills.ts:341` passes the already-`stripBullet`ed text; `stripBullet` is idempotent so it is correct either way, but taking the captured label and body as parameters would make the "two decisions stay aligned" contract structural instead of by-convention.

## Acceptance criteria — #833

| AC | Result |
|---|---|
| `google-docs-skia-proxy-role-first-experience.pdf` — no `Fluent in Spanish`, no `Language` category | ✅ `skillsCount` 12 → 11, matching the 11-entry truth `skills` exactly |
| `Languages: Python, Go, TypeScript` → 3 skills under `Languages`. Unit test | ✅ covered |
| `Languages: Spanish, French, Mandarin` still admitted (stated limit) | ✅ covered |
| `Certified in AWS Solutions Architecture` under a non-language label unaffected | ✅ covered |
| `multi-degree-coursework.pdf` Languages row unchanged | ✅ `expected.json` untouched; that row carries no proficiency word |
| `corpus.test.ts` passes; precision up on target, down on none | ✅ on the corpus — but see Blocking 1/2, which the corpus does not exercise |
| `corpus-roundtrip.test.ts` passes, no new `KNOWN_FAILURES` rows | ✅ no baseline file in the diff |
| `npm run verify` passes | ✅ reproduced locally, green through build |

Plan step 3 also holds: no `languages` field exists in `docs/canonical-resume-model.md` or the parsed types, so dropping is the right answer and none was added. Step 5 holds too — `UNFILED_TRUTH_CEILING` is untouched at 7, correct since the cleared entry was `status: "open"`, never unfiled.

`## Summary` claims the change preserves programming-language rows. Blocking 1 shows it does not, for `Visual Basic` / `React Native` / `Proficient in …` / `X (basic)`. Same defect, so not counted twice — but once the predicate is narrowed, the summary needs a line naming which language rows are still dropped, so the limit is stated rather than discovered.

The `## Verification` list is otherwise accurate and reproduced here.

## Gates

- **3a fixture PII** — no binary added or changed; only `.expected.json` / `.truth.json`. `npm run check:fixtures` green (it sweeps truth sidecars since #654).
- **3b design system / 3c style tokens** — n/a, nothing under `src/components/`.
- **3d fallow** — 2 complexity findings (`isSoftWrapContinuation`, `extractSkills`) + 1 clone group at `skills.test.ts:589-664`, all pre-existing and outside this diff. Report-only.
- **3e skill/script command bugs** — n/a, no bash/`gh`/script files in the diff.
- **`npm run verify`** — green end to end here, full suite included. The `useLibraryChanges.test.tsx` `BroadcastChannel`/`MessageEvent` failure noted in the PR body did **not** reproduce; it looks environment-dependent, and nothing in this diff touches it.

## Branch hygiene

The branch carries **4 commits** (`1b083e3`, `616c09c`, `b20c622`, `8eaae56`). `main` merges through a merge queue that derives the squash message from the branch, so as-is this lands `test(fixtures): refresh Word skill snapshot` in `main` permanently. Collapse to one before enqueueing — `/collapse-pr` does it. Not counted as a finding, and this review pushed nothing (≥1 Blocking).

---

qtjg:
Follow-up after the review-requested changes:

The revised implementation is pushed at `febb6c8` and now:

- evaluates proficiency wording per delimited fragment, preserving mixed programming-language rows containing `Visual Basic`, `React Native`, or `Python (proficient)`;
- supports common label variants such as `Foreign Languages`, `Spoken Languages`, qualified labels, and whitespace before the colon;
- covers standard wording including `proficiency`, `elementary`, and `limited`; and
- preserves the category boundary when a dropped proficiency row has a soft-wrapped continuation.

Validation completed locally:

- `npx vitest run src/lib/heuristics/extract/skills.test.ts` — 54 passed;
- affected corpus snapshot tests — 2 passed;
- affected corpus round-trip tests — 3 passed;
- typecheck, lint, fixture, baseline, core-package, and build checks — passed; and
- `git diff --check` and the final secret-pattern scan — passed.

The full `npm run verify` run reached thousands of passing tests but did not complete in the sandbox because unrelated PDF round-trip work stalled resolving `/src/assets/fonts/Poppins-Regular.ttf`. No merge or closure action was taken; the PR is awaiting maintainer re-review.

src/lib/heuristics/extract/skills.ts
```diff
@@ -304,6 +304,36 @@ const SUBLABEL_PREFIX_RE = new RegExp(`^(${SUBLABEL_BODY}):\\s*`);
...
+  const debulleted = stripBullet(cell);
+  const match = debulleted.match(SUBLABEL_PREFIX_RE);
+  if (!match || !LANGUAGE_LABEL_RE.test(match[1])) return false;
+
+  const fragments = splitRespectingParens(debulleted.slice(match[0].length))
+    .map((fragment) => fragment.trim())
+    .filter((fragment) => fragment !== "");
+  return (
+    fragments.length > 0 &&
+    fragments.every((fragment) => LANGUAGE_PROFICIENCY_BODY_RE.test(fragment))
```
s-annam: **Blocking.** `fragments.every(...)` is vacuously true when there's only one fragment, so a lone skill whose name contains proficiency vocabulary (e.g. `Visual Basic`, `BASIC`) is silently dropped in full — not just the proficiency word, the whole skill. This contradicts the docblock two lines up, which claims the per-fragment check specifically prevents `Visual Basic`/`React Native` from disappearing; that only holds when the row has 2+ items.

Repro: `tokenizeSkillLine("Languages: Visual Basic")` → `[]` (verified against `febb6c8`).

Suggest gating on `fragments.length > 1 && fragments.every(...)`, with explicit test coverage either way for the single-item case.

src/lib/heuristics/extract/skills.ts
```diff
@@ -317,7 +347,11 @@ function tokenizeCell(cell: string, out: Set<string>): void {
   // leading `Label:` prefix and drop the whole cell when it names a
   // hobbies/interests list — before the label is stripped and the items split.
   const labelMatch = debulleted.match(SUBLABEL_PREFIX_RE);
-  if (labelMatch && NON_SKILL_SUBLABEL_RE.test(labelMatch[1])) return;
+  if (
+    labelMatch &&
+    (NON_SKILL_SUBLABEL_RE.test(labelMatch[1]) || isLanguageProficiencyCell(debulleted))
```
s-annam, line 376: **Blocking.** This runs on `debulleted`, which may already be the product of the pre-existing soft-wrap join (`isSoftWrapContinuation` Condition B) merging this `Language:` line with a *following, unrelated* comma-bearing skill line into one cell. Once merged, the fragment list contains real skill tokens, so `isLanguageProficiencyCell` returns false and the entire merged blob — including the original proficiency phrase — is tokenized as skills instead of being dropped.

Repro (two adjacent lines): `Languages: Fluent in Spanish, conversational French` then `Project Management, Agile, Scrum` → `extractSkills(...).value` = `["Fluent in Spanish", "conversational French Project Management", "Agile", "Scrum"]`. This reintroduces the exact #833 defect plus a garbled cross-line token, under a plausible real layout (a labeled proficiency bullet immediately followed by an unlabeled skills bullet).

src/lib/heuristics/extract/skills.ts
```diff
@@ -647,11 +687,21 @@ export function extractSkills(
...
       cellTokens.push(tok);
     }
+    const label = matchCellLabel(cell);
+    // A dropped language-proficiency row still opens a category boundary. Keep
+    // that anchor so a soft-wrapped tail cannot be misfiled under the previous
+    // category; the category is emitted only if a later continuation contributes
+    // a token.
+    if (cellTokens.length === 0) {
+      if (label !== undefined && isLanguageProficiencyCell(cell)) {
+        categories.push({ label, skills: [] });
```
s-annam, line 744: **Blocking.** This empty-category placeholder exists so a genuine wrap-continuation of *this same dropped row* can reattach later (see the docblock addition on `matchCellLabel`). But the append path a few lines below (`else if (categories.length > 0)`, `skills.ts:707`) can't tell a true same-row continuation from a merely-adjacent, unrelated bare cell — it appends either one.

Repro: `Languages: Fluent in Spanish` (single item, no trailing comma, so it does NOT soft-wrap-join with the next line) followed by an unrelated bare `Python, Go, TypeScript` → `categories` = `[{"label":"Languages","skills":["Python","Go","TypeScript"]}]`. Three programming languages get mislabeled under a spoken-language header, and the structured `categories` view is now emitted where, pre-fix, an uncategorised bare list would have suppressed it entirely.

s-annam (changes requested):
**Verdict: REQUEST_CHANGES** — 3 Blocking findings, each independently reproduced by running the PR's own code against a targeted repro. Rule applied: ≥1 Blocking → REQUEST_CHANGES (no auto-fix, no collapse).

All three share one root cause worth naming up front: `isLanguageProficiencyCell` decides on a single cell's fragment list, but that decision point interacts badly with the section's *other* existing mechanics — the soft-wrap join that can fold two source lines into one cell before the check runs, and the `else if (categories.length > 0)` continuation-append that can't tell a same-row wrap-continuation from a merely-adjacent unrelated line. Fixing all three likely wants one shared fix, not three separate patches.

## Blocking

1. **`skills.ts:333`** — `fragments.every(...)` is vacuously true on a single-fragment cell, so a lone skill whose name contains proficiency vocabulary is silently dropped in full, contradicting the docblock's own claim ("applied per fragment so `Visual Basic` and `React Native` do not cause an entire programming-language row to disappear" — that guarantee only holds for 2+ items). Repro: `tokenizeSkillLine("Languages: Visual Basic")` → `[]`, and `tokenizeSkillLine("Languages: BASIC")` → `[]`. A résumé whose sole listed language-category skill is `Visual Basic` or `BASIC` loses it with no trace. Suggested fix: gate on `fragments.length > 1 && fragments.every(...)` at minimum, and add regression coverage for the single-item case either way (undropped, or an explicit single-item heuristic).

2. **`skills.ts:352`** — `isLanguageProficiencyCell(debulleted)` runs on the cell *after* the pre-existing soft-wrap join (`isSoftWrapContinuation` Condition B) may already have merged this Language: line with a following, unrelated comma-bearing skill line into one cell. Once merged, the fragment list contains real skill tokens, so `fragments.every(...)` fails, and the *whole* blob — including the original proficiency phrase — is NOT dropped and tokenizes normally. Repro (two adjacent lines, a common layout): `Languages: Fluent in Spanish, conversational French` followed by `Project Management, Agile, Scrum` → `extractSkills(...).value` = `["Fluent in Spanish", "conversational French Project Management", "Agile", "Scrum"]`. This reintroduces the exact #833 defect (proficiency phrase admitted as a skill) plus a garbled cross-line merged token, under a layout shape this PR should cover (a labeled proficiency bullet immediately followed by an unlabeled skills bullet).

3. **`skills.ts:697`** — when a proficiency row is dropped, its label is retained as an empty-skills category anchor (`categories.push({ label, skills: [] })`) purely so a genuine wrap-continuation of *that same row* can reattach later. But the existing append path at `skills.ts:707` (`else if (categories.length > 0)`) can't distinguish a true same-row continuation from a merely-adjacent, unrelated bare cell — so an unrelated bare list gets mislabeled under the dropped row's category. Repro: `Languages: Fluent in Spanish` (single item, no trailing comma — does not soft-wrap-join with the next line) followed by an unrelated bare `Python, Go, TypeScript` → `categories` = `[{"label":"Languages","skills":["Python","Go","TypeScript"]}]`. Three programming languages get mislabeled under a spoken-language category header, and — worse — the structured `categories` view is now emitted where, pre-fix, an uncategorised bare list would have suppressed it entirely (the field would have been absent, not present-but-wrong).

All three verified by direct execution against the PR's actual head commit (`febb6c8`) in an isolated worktree, not just static reading — none of the three trigger on the PR's own `skills.test.ts`/`corpus.test.ts` suites, which is why they survived to review.

## Secondary
None beyond the Blocking findings above.

## Nits
- `skills.ts:655-659` (`matchCellLabel`) — the diff reformats `if (!m || NON_SKILL_SUBLABEL_RE.test(m[1])) return undefined;` onto five lines with no behavior change. Pure churn; consider reverting to the single-line form to keep the diff focused on the actual fix.
- `extractSkills`'s cyclomatic/cognitive complexity grew with the new nested `if (cellTokens.length === 0) { if (...) {...} continue; }` block. `fallow audit --base origin/main` attributes 0 new issues to the changed files (this and `isSoftWrapContinuation`/`splitColumnCells` are reported as pre-existing/inherited, non-blocking, report-only per repo convention) — flagging only so it's on record if this file's complexity gets revisited.

## AC checklist (issue #833)
- [x] `google-docs-skia-proxy-role-first-experience.pdf` — no `Fluent in Spanish` skill, no `Language` category. Truth exception removed accordingly.
- [x] `Languages: Python, Go, TypeScript` still parses as three skills under `Languages`.
- [x] Bare spoken-language list (`Spanish, French, Mandarin`, no proficiency wording) still admitted — matches the issue's own documented limit.
- [x] `Certifications: Certified in AWS Solutions Architecture` unaffected.
- [x] `multi-degree-coursework.pdf`'s programming-language `Languages` row unchanged (not touched by the diff).
- [x] `corpus.test.ts` passes (65 tests, 1 pre-existing skip).
- [x] `corpus-roundtrip.test.ts` passes, no new `KNOWN_FAILURES` rows (`corpus-edit-roundtrip.known-failures.json` untouched by the diff).
- [~] `npm run verify` — not run as the single full command this session, but its components were run directly and all passed: `typecheck`, `lint`, `check:fixtures`, `fallow audit`, plus the full `skills.test.ts` (49 tests) / `corpus.test.ts` / `corpus-roundtrip.test.ts` suites. The PR body's own verification log claims a full green `verify` plus a disclosed, unrelated `useLibraryChanges.test.tsx` flake; that file passed in isolation here too, consistent with a full-suite timing flake rather than a regression.

The AC list is satisfied on its own terms — the Blocking findings above are regressions the ACs didn't anticipate (single-item rows, cross-line soft-wrap interaction, category mislabeling), not AC gaps.

## Description accuracy (3f)
Accurate as far as it goes — every claim in the Summary round-trips to the code for the cases it describes, and the disclosed `useLibraryChanges.test.tsx` failure is handled honestly (named as unrelated, and verified so here). It doesn't claim immunity from the three bugs above, but the docblock on `LANGUAGE_PROFICIENCY_BODY_RE` (`skills.ts:315-319`) does implicitly overclaim per-fragment safety for the single-item case (Finding 1).

## Gates run
- `/code-review` (effort: high, background fork) — 3 findings, all reproduced and included above as Blocking.
- 3a fixture PII — `npm run check:fixtures` passed (60 PDFs, 16 sidecars); the modified `.truth.json` sidecar is a `knownWrong` removal, not new PII surface.
- 3b design-system/reuse — n/a, no `src/components/**` touched.
- 3c style tokens — n/a, no `src/**` styling touched.
- 3d fallow — `npx fallow audit --base origin/main`: 0 issues attributed to the 7 changed files; 3 pre-existing complexity findings excluded as inherited. Report-only per repo convention regardless.
- 3e skill/script command bugs — n/a, no `.claude/skills/**` or `scripts/**` touched.
- 3f description accuracy — accurate, see above.
- `npm run typecheck`, `npm run lint` — both clean.

Head reviewed: `febb6c8`. 👀 posted at review start.

---

src/lib/heuristics/extract/skills.ts
s-annam: Fixed already in febb6c8 (before this round) — `LANGUAGE_LABEL_RE` now allows the `foreign|spoken|other` qualifiers and a trailing `\s*` so `Languages :` (space before colon) no longer escapes via the capture-artifact you identified. All three of your repro lines are covered by the existing "drops spoken-language proficiency rows across common label variants" test.

src/lib/heuristics/extract/skills.ts
s-annam: Fixed in 5b316a9 (this was already fixed for this exact repro by febb6c8's category-anchor mechanism, and stays fixed with the tightened trailing-comma restriction from the round-2 review — see the reply on the other Blocking-3 thread). Regression test added: "does not misfile a wrapped tail under the previous category".

src/lib/heuristics/extract/skills.ts
s-annam: Fixed already in febb6c8 (before this round) — the vocabulary now includes `proficiency|proficiencies` (matches "JLPT N2 Japanese Proficiency" and "Full professional proficiency in German") plus `elementary`/`limited`. Covered by "drops standard proficiency-scale wording".

src/lib/heuristics/extract/skills.ts
s-annam: Fixed in 5b316a9. Added a guard to `isSoftWrapContinuation`: when `pending` alone already resolves as a complete `isLanguageProficiencyCell`, the join is declined outright — a proficiency row that's grammatically finished (no trailing comma/semicolon) can't be the wrap-in-progress of a following, unrelated comma-bearing line. Verified your exact repro no longer garbles: `extractSkills` on "Languages: Fluent in Spanish, conversational French" + "Project Management, Agile, Scrum" now returns `["Project Management","Agile","Scrum"]` with no proficiency phrase and no cross-line token. Test added: "does not join a complete proficiency row with a following, unrelated skill line".

src/lib/heuristics/extract/skills.ts
s-annam: Fixed in 5b316a9. The single-fragment case now goes through the same structural predication check as the multi-fragment case (see the reply on your round-1 Blocking-1 thread) rather than a length-based gate — so `tokenizeSkillLine("Languages: Visual Basic")` → `["Visual Basic"]`, `tokenizeSkillLine("Languages: BASIC")` → `["BASIC"]`, while a genuine single-item predication like `Languages: Native German` still drops. Pinned as "keeps a SOLE skill whose name contains proficiency wording (single-fragment row)" and "still drops a single-item genuine proficiency predication".

src/lib/heuristics/extract/skills.ts
s-annam: Addressed by the design change in febb6c8/5b316a9, not by suppressing the guard: `matchCellLabel` is now genuinely reachable and called for every cell (including zero-token ones), because it's what supplies the label for the dropped-row category anchor in `extractSkills`. The docblock above it now names that (language-proficiency rows retain their label so a dropped row can still anchor a continuation). Left the `cell`-vs-already-stripped-text parameter shape as-is per your "correct either way" note — happy to revisit if you'd still like the structural refactor, but treating it as non-blocking since you flagged it as a nit.

src/lib/heuristics/extract/skills.ts
s-annam: Fixed in 5b316a9. The empty-category anchor now only opens when the dropped cell itself is unterminated (ends on a trailing comma/semicolon — the same "list got cut off mid-item" signal Condition B already uses), not merely because it's a language-proficiency cell. A complete single-item row like "Languages: Fluent in Spanish" (no trailing comma) no longer opens an anchor, so an unrelated bare list that happens to follow it isn't swept under "Languages". Verified your exact repro: `categories` no longer contains `{"label":"Languages","skills":["Python","Go","TypeScript"]}` — Python/Go/TypeScript land in the flat `value` list uncategorised instead. Test added: "does not misfile an unrelated bare list under a dropped row's anchor category".

src/lib/heuristics/extract/skills.ts
s-annam: Fixed in 5b316a9. Replaced the plain vocabulary substring test with a structural predication check: the noun forms "proficiency"/"proficiencies" still match anywhere, but the adjective forms ("fluent", "native", "basic", …) only count with predicate structure — leading the fragment with an object ("Native German"), joined by "in" ("Fluent in Spanish"), parenthesised ("Spanish (native)"), or followed by "speaker". Applied per-fragment as before, so it also fixes the coincidental case where every fragment in a multi-item list happens to match. Pinned all four rows from your table as unit assertions, plus the multi-fragment `Spoken Languages:` case.

s-annam (approved):
All round-1 and round-2 findings verified fixed in `5b316a9`. See inline replies on each resolved thread for what changed. `typecheck`/`lint`/`skills.test.ts` (58)/`corpus.test.ts`/`corpus-roundtrip.test.ts`/full suite (5878 tests) all green.

Note for the merge: branch carries 2 commits (`febb6c8`, `5b316a9`) — `/collapse-pr`'s ownership gate correctly refuses to force-push/rewrite a fork branch with a foreign-authored commit, so this needs a manual collapse decision before it reaches the queue.

s-annam:
@qtjg — one thing my last comment here didn't say plainly enough: commit `5b316a9` was pushed by me (s-annam), directly to this branch. Wanted to correct that and explain why, rather than leave it ambiguous.

Timeline: #833 was assigned to you on 2026-08-16, this PR opened 2026-08-18, and it's had two review rounds requesting changes — the second (2026-08-21) found three new blocking issues in the per-fragment revision from `febb6c8` (a single-item row like `Visual Basic`/`BASIC` being dropped whole via a vacuously-true check, the soft-wrap join running before the proficiency check so it could re-admit a dropped phrase by merging it with an unrelated following line, and the dropped-row category anchor over-triggering onto an unrelated bare list). By the time I looked at this again it had been open close to a week with no further update after that second round, so rather than keep it waiting on another round-trip, I implemented fixes for those three findings myself and pushed them to your branch (GitHub's "maintainers can edit" setting on this PR made that possible) to get it moving.

To be clear: the per-fragment approach in `febb6c8` was the right direction, and diagnosing that a body-shape check needed to apply per delimited fragment rather than to the whole cell was your work — the three round-2 findings were edge cases in that same mechanism, not a different approach. I've replied to and resolved each review thread individually with what specifically changed and why. Branch is at 2 commits (yours + mine) — happy to talk through whichever way you'd like that handled before this merges.

Thanks for taking this one on.

offlinecv-approver (approved):
🕹✅

PR #866, opened 2026-08-18
feat(jd-match): semantic verdict UI, on-device opt-in, keyword fallback view (#204)

### LINKED ISSUE #204: JD-match: JdMatch.tsx semantic verdict UI + opt-in toggle + keyword fallback view

Vaishnavi1709:
## Summary

`JdMatch.tsx` returned `null` for `path: "semantic"`, so a finished on-device match rendered a blank panel, and `semanticOptIn` had no route to a user — `PasteJdPanel` called `useJdMatch` without it, hardwiring the default `false`. This closes both.

`JdMatch` becomes a router on the `JdMatchResult` discriminant; the pre-#204 body moves to `KeywordMatch` **verbatim**; `SemanticMatch` renders verdicts grouped Met → Partial → Missing; `PasteJdPanel` owns the opt-in boolean and renders `semanticResult ?? keyword`, so the keyword floor survives every other state. `useJdMatch` gains one field (`capability`) — with opt-in on and WebGPU absent, `status` settles on `ready`-with-keyword, indistinguishable from opt-in-off *and* from a degraded run, and the UI needs that difference to explain why ticking the box changed nothing. No orchestration moved into a component.

Three places where the June issue text no longer matches the repo, adapted rather than followed:

- **"Judging requirement 4 of 9…"** has no data source — `judgeEvidence` takes no progress callback and `runLlmMatch` reports only engine-load progress plus a single `onInferenceStart`. The running copy is generic rather than invented, and a test asserts no `/requirement \d+ of \d+/` appears.
- **`/jd-fit/`** is retired (#576/#783), so the integration point is `/jobs/` → `FindJobsPanel` → `PasteJdPanel`.
- **`StatusBadge` had no tone for "missing"**; added `neutral` to the shared piece rather than hand-rolling a pill at one callsite.

**#804 assessment (not fixed here, deliberately):** its stated scenario is stale — it cites `job-search/sector.ts`'s `classifySector`, which has **no production caller** (`useCompanyTargets` uses `classifySectorHeuristic`), and `/jobs/` has no other `loadEngine` caller at all. The reachable variant is self-inflicted: edit the JD while the weight download is in flight and the new run joins a load whose `onProgress` it cannot hear, so the bar reads 0% until the load resolves. Cosmetic, not functional — the keyword floor stays on screen throughout — and the fix is a progress fan-out inside `web-llm.ts` shared by every WebLLM surface, which is #804's scope. Recorded in `SemanticAnalysisOptIn`'s docblock.

Closes #204

## Review focus

- `src/components/features/PasteJdPanel.tsx:96` — `semanticResult ?? jdMatch` is the whole no-stale-verdict guarantee at the UI layer. Does *every* non-`ready` state really fall through to the keyword floor, including opt-out after a completed run?
- `src/components/features/SemanticAnalysisOptIn.tsx:126` — `StatusLine` checks `capability` **before** it switches on `status.kind`. Invert that order and the no-WebGPU case renders the "didn't return a verdict" degrade note instead. Is the ordering obvious enough to survive a later edit?
- `src/components/features/SemanticAnalysisOptIn.tsx:196` — the degrade note is suppressed on a cancelled run only because `useJdMatch` bumps its request id *before* aborting, so the abandoned run's keyword fallback fails the write guard. Is that reasoning right, or can a superseded run still land a `ready`-with-keyword slot?
- `src/design-system/shared/StatusBadge.tsx:47` — `neutral` is `bg-surface-subtle text-content-muted`. Muted-on-subtle is deliberately low-contrast; is it still legible enough for a badge that carries meaning, in both themes?

## Test plan

- [x] `npm run lint` clean
- [x] `npm run verify` green (also enforced by the pre-push hook on this branch)
- [x] `npm test` full suite green — 367 files / 5960 tests, 0 failures
- [x] `npx fallow audit --base origin/main` — no issues in 13 changed files
- [x] **Keyword parity proven, not asserted**: golden strings in `KeywordMatch.parity.test.tsx` were captured by rendering the **pre-#204** component, and the new render is byte-identical across three shapes (populated + footnote, both empty states, singular footnote fork)
- [x] **Fail-before validation** — 10 regressions injected and reverted, each confirmed to fail the test that claims to catch it: semantic route → `null`; unconditional `detectWebGpu`; `previousController?.abort()` removed; request-id guard removed; keyword copy changed; running copy → "Judging requirement 4 of 9…"; keyword floor blanked during load; evidence made hover-only; `sr-only` name qualifier dropped; row status word dropped
- [x] Bundle: `/jobs/` +4.4 kB raw / +1.0 kB gzip. `/` *shrank* 1.8 kB — bisected across five builds to `ModelLoadProgress` hoisting out of `main-*.js` into the shared chunk now that both entries use it. `run-llm-match` and `web-llm` remain lazy chunks; zero WebLLM strings in the `/jobs/` entry

### DIFF
--- a/src/components/features/SemanticAnalysisOptIn.tsx
+++ b/src/components/features/SemanticAnalysisOptIn.tsx
@@ -1,0 +1,215 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * SemanticAnalysisOptIn — the "Analyze with on-device AI" control and the
+ * lifecycle line that belongs under it (#204).
+ *
+ * Display-only and fully controlled: the checked state lives in `PasteJdPanel`
+ * (which feeds it to `useJdMatch` as `semanticOptIn`), so the boolean has a
+ * single owner. This component holds no state, starts no work, and imports
+ * nothing from `webllm/` but a type — `detectWebGpu`, `loadEngine` and
+ * `runLlmMatch` are all upstream of it.
+ *
+ * The control and its progress sit together because that is the house pattern
+ * for every WebLLM surface in the repo: `ResumeQualityPanel`, `ResumeRewrite`
+ * and `SectionRewrite` each render `ModelLoadProgress` directly under the
+ * trigger that started the load, not inside the result panel. Keeping it out
+ * of `JdMatch` is also what keeps the keyword floor visible: the result card
+ * below goes on rendering keyword coverage for the whole engine load instead
+ * of being replaced by a spinner.
+ *
+ * ## Why the multi-GB download is opt-in and defaults OFF
+ *
+ * Per #172's stance and #204's scope: the instant keyword coverage is the
+ * return hook, and a panel that starts a hundreds-of-megabytes fetch because a
+ * user pasted a JD would be spending their bandwidth on a guess. OFF also
+ * carries a privacy property, not just a bandwidth one — with the box unticked
+ * `useJdMatch` never calls `detectWebGpu`, so no `webllm_capability_detected`
+ * event enters the funnel for a user who only ever wanted keyword coverage.
+ * The gate lives in the hook; this component's job is to make it a user's
+ * decision rather than a hardwired `false`.
+ *
+ * ## What this is NOT
+ *
+ * Not a model picker, and not a licence-consent gate. Those are separate,
+ * already-built mechanisms (`useModelSelection` supplies the selected id;
+ * `ConsentDialog` covers restricted-licence downloads on the surfaces that
+ * offer them). This is one boolean: "may this panel use the on-device model at
+ * all". It deliberately does not persist — a session-scoped `useState` in the
+ * panel, no `localStorage` — because #204 asks for an opt-in toggle and
+ * nothing in the issue or the repo asks a JD panel to remember it.
+ *
+ * ## No-WebGPU is not an error
+ *
+ * `WebGpuUnavailableNotice` is the repo's other answer to "no WebGPU", and it
+ * is the wrong one here: it renders a warning-toned strip with a how-to-enable
+ * dialog and fires `webllm_notice_shown`. #204 asks for the opposite — the
+ * keyword columns keep rendering, with at most one muted line saying why the
+ * box did nothing. A user whose browser can't run this still has the whole
+ * panel they came for, so nothing is in a failed state.
+ *
+ * ## Known limitation: the progress bar can sit at 0% (#804)
+ *
+ * `loadEngine`'s "already pending" fast path returns the shared promise
+ * without registering the new caller's `onProgress`, so only the FIRST caller
+ * for a model id ever receives progress. The reachable case here is
+ * self-inflicted rather than the cross-consumer one #804 describes — that one
+ * cites `job-search/sector.ts`'s `classifySector`, which has no production
+ * caller, and `/jobs/` has no other `loadEngine` caller at all. What IS
+ * reachable: opt in, then edit the JD while the weight download is still in
+ * flight. The superseded run owns `initProgressCallback`, its writes are
+ * dropped by the controller's id guard, and the new run joined a load it can't
+ * hear — so the bar reads 0% until the load resolves, then moves on to
+ * running → ready normally.
+ *
+ * Cosmetic, not functional, and deliberately NOT worked around here: the fix
+ * is a progress fan-out inside `web-llm.ts`, shared by every WebLLM surface in
+ * the repo, which is #804's scope and not this component's. The reason it is
+ * tolerable meanwhile is the keyword floor — the result card below keeps
+ * showing full coverage throughout, so a stalled bar costs the user a progress
+ * readout, never the answer they came for.
+ */
+
+import type { ReactNode } from "react";
+import { Checkbox, ModelLoadProgress } from "@design-system";
+import type { JdMatchStatus } from "../../hooks/useJdMatch.ts";
+import type { WebGpuCapability } from "../../lib/webllm/types.ts";
+
+interface SemanticAnalysisOptInProps {
+  /** Controlled opt-in state; owned by `PasteJdPanel`. */
+  checked: boolean;
+  onChange: (next: boolean) => void;
+  /** The controller's semantic status. */
+  status: JdMatchStatus;
+  /** The controller's WebGPU probe result; `null` until it resolves (and
+   *  forever while `checked` is false, since the probe is gated on opt-in). */
+  capability: WebGpuCapability | null;
+}
+
+export function SemanticAnalysisOptIn({
+  checked,
+  onChange,
+  status,
+  capability,
+}: SemanticAnalysisOptInProps) {
+  return (
+    <div className="flex flex-col gap-2">
+      <Checkbox
+        checked={checked}
+        onChange={onChange}
+        label="Analyze with on-device AI"
+        hint="Judges each requirement against your résumé instead of matching terms by name. One-time model download; your JD text still never leaves this tab."
+      />
+      <StatusLine checked={checked} status={status} capability={capability} />
+    </div>
+  );
+}
+
+/** Muted one-liner — the tone for "this is information, not a problem". */
+function Note({ children }: { children: ReactNode }) {
+  return (
+    <p role="status" className="text-sm text-content-tertiary">
+      {children}
+    </p>
+  );
+}
+
+function StatusLine({
+  checked,
+  status,
+  capability,
+}: {
+  checked: boolean;
+  status: JdMatchStatus;
+  capability: WebGpuCapability | null;
+}) {
+  // Unticked: render nothing at all, so the default panel is unchanged from
+  // its pre-#204 self down to the DOM.
+  if (!checked) return null;
+  // No JD yet (or one that extracted no terms) — there is nothing to analyze,
+  // so a progress line would be describing work that isn't happening.
+  if (status.kind === "idle") return null;
+
+  // Capability first: with the probe unresolved OR resolved-unavailable, the
+  // hook's `status` is `ready` holding the KEYWORD result, which is
+  // indistinguishable from a semantic run that degraded. Only `capability`
+  // separates them, which is why the controller exposes it.
+  if (capability === null) {
+    return <Note>Checking whether this browser can run on-device analysis…</Note>;
+  }
+  if (capability !== "available") {
+    return (
+      <Note>
+        This browser can't run on-device analysis (it needs WebGPU) — the
+        keyword coverage below is unaffected.
+      </Note>
+    );
+  }
+
+  if (status.kind === "loading") {
+    return (
+      <ModelLoadProgress
+        progress={status.progress.progress}
+        text={status.progress.text}
+        label="Loading the on-device model (one-time download)"
+        // No `showExplainer`: the checkbox hint directly above already states
+        // that the model downloads and that the text stays in the tab, and
+        // `ModelLoadProgress`'s own docblock makes the explainer opt-in so
+        // that a caller with its own context doesn't double up.
+      />
+    );
+  }
+
+  if (status.kind === "running") {
+    // Generic on purpose. #204's example copy ("Judging requirement 4 of 9…")
+    // has nothing behind it: `judgeEvidence` takes no progress callback and
+    // `runLlmMatch` reports only engine load + a single `onInferenceStart`, so
+    // a count here would be invented. Adding a batch-progress API to the LLM
+    // layer to satisfy one string is not warranted; truthful copy is.
+    return (
+      <p role="status" className="text-sm text-content-secondary">
+        Reading this JD and checking it against your résumé…
+      </p>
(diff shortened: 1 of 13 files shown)


s-annam (approved):
**Stance: APPROVE.** 0 Blocking findings → APPROVE per the verdict rule (0 Blocking → APPROVE regardless of Secondary/Nits; none of either survived here either).

## Blocking
None.

## Secondary
None.

## Nits
None — see the one AC note below, which is disclosed reasoning rather than a defect.

## AC checklist (issue #204)
All acceptance criteria met:
- `JdMatch.tsx` is a router on `result.path` — `keyword` → `<KeywordMatch>` (verified byte-identical to the pre-#204 body: same element order, class strings, copy, empty states — confirmed by reading the diff directly, not just trusting the parity test), `semantic` → `<SemanticMatch>` grouped Met → Partial → Missing, `Card`/`StatusBadge` reused. ✓
- Opt-in toggle "Analyze with on-device AI", default OFF, on the JD panel. ✓ (`SemanticAnalysisOptIn`, `useState(false)` in `PasteJdPanel`)
- Loading → `ModelLoadProgress` (same primitive other WebLLM surfaces use). ✓
- Running → step indicator. The issue's literal example copy ("Judging requirement 4 of 9…") is **not** implemented — `judgeEvidence` has no per-requirement progress callback to source it from, so the PR ships generic copy instead and says so explicitly in the description's "adapted rather than followed" section. This is disclosed reasoning, not an omission, so it doesn't fail gate 3f — noting it here only so the record is explicit.
- No WebGPU → silently degrades to keyword columns with one muted explanatory line, no error/flicker. ✓ (verified in `SemanticAnalysisOptIn`'s `StatusLine`, and covered by the fail-before test list)
- Constraints: semantic tokens only, `Button`/`Card`/`StatusBadge` primitives, no raw `<button>`, privacy disclaimer preserved. ✓ (grepped the diff for raw elements and hardcoded colors — the only hex-shaped matches were `#204` issue references, false positives)

Accurate. Every checkable claim in the Summary and Review-focus sections round-trips to the diff:
- `semanticResult ?? jdMatch` fallthrough in `PasteJdPanel.tsx` — confirmed every non-`ready`/non-semantic state (idle, loading, running, error, opted-out, degraded) falls through to the keyword floor.
- The `capability`-before-`status.kind` ordering in `SemanticAnalysisOptIn`'s `StatusLine` — confirmed the check order matches the description.
- The cancelled-run guard claim — traced independently in `useJdMatch.ts`: `requestIdRef.current` is bumped (`++requestIdRef.current`) *before* `previousController?.abort()` is called, and `setSlotIfCurrent` checks `requestIdRef.current !== myId` before writing any state, so a superseded run's late write is dropped. Matches the PR body's reasoning exactly.
- `StatusBadge`'s new `neutral` tone reuses `bg-surface-subtle`/`text-content-muted` — the same pairing already used elsewhere in this file (the "alpha" badge), not a new token combination introduced by this PR.
- Bundle-size deltas in the Test plan were not independently re-measured (would require a full build + diff) but are plausible given the change shape (hoisting `ModelLoadProgress` into a shared chunk now that both entries use it) and don't affect the verdict either way.

## Gates run
- 3a (fixture PII): **skipped** — no fixture files touched.
- 3b (design-system reuse): pass. New feature files (`KeywordMatch`, `SemanticMatch`, `SemanticAnalysisOptIn`) are the router split the issue explicitly asked for, not a parallel surface; `StatusBadge`'s new `neutral` tone correctly extends the shared primitive instead of a bespoke pill.
- 3c (style tokens): pass. No hardcoded hex/palette classes/manual `dark:` variants (grep hits were all `#NNN` issue-number references).
- 3d (fallow): pass. `npx fallow audit --base origin/main` → "No issues in 13 changed files"; the only duplication flagged is pre-existing and excluded as inherited.
- 3e (command-level bugs): n/a — no skill/script files touched.
- 3f (description accuracy): pass, see above.
- `npm run typecheck`: clean. `npm run lint`: clean. `npx vitest run` on the 6 touched test files: 94/94 passing.

Note for the record: this run's background `/code-review` (generic correctness pass) did not return within a reasonable wait and was abandoned in favor of the manual pass documented above, which covered the full diff file-by-file including independent tracing of the two logic claims called out in "Review focus". No correctness issues found.

---

src/components/features/SemanticMatch.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+function VerdictRow({ verdict }: { verdict: RequirementVerdict }) {
+  const { requirement, status, reason, evidence } = verdict;
+  return (
+    <li className="flex flex-col gap-1.5 rounded border border-border-light bg-surface-subtle p-3">
+      <div className="flex items-baseline gap-2">
+        {/* `StatusBadge` is `w-fit`, which stops it growing but not shrinking;
+            the wrapper is what keeps a long requirement from squeezing it. */}
+        <span className="shrink-0">
+          <StatusBadge tone={GROUP_BADGE_TONE[status]}>
```
s-annam, line 148: This `StatusBadge tone={GROUP_BADGE_TONE[status]}` renders with `tone="neutral"` for the `missing` group. `neutral` is `bg-surface-subtle text-content-muted` (`StatusBadge.tsx:47`), and the row this sits in (`VerdictRow`'s `<li>`, line 143) is itself `bg-surface-subtle` — same token, so the pill has no visible fill against its own row. See the Secondary finding #1 in the review body for the full writeup and a suggested fix.

src/components/features/SemanticAnalysisOptIn.tsx
```diff
@@ -0,0 +1,215 @@
...
+/** Muted one-liner — the tone for "this is information, not a problem". */
+function Note({ children }: { children: ReactNode }) {
+  return (
+    <p role="status" className="text-sm text-content-tertiary">
+      {children}
+    </p>
+  );
+}
+
+function StatusLine({
```
s-annam, line 118: `StatusLine` branches on `status.kind` via a chain of `if`s rather than an exhaustive `switch`. A future `JdMatchStatus` variant added to the union (currently 5 members) would fall through to the final `return null` with no compiler warning. See Nit #8 in the review body — not urgent, just flagging for awareness.

s-annam (commented):
**Follow-up to the APPROVE above.** The background `/code-review` pass I flagged as stalled came back after I'd already posted (three duplicate notifications, same 8 findings each time). I verified all 8 directly against source before adding them here. None change the verdict — 0 Blocking, so the PR stays **APPROVE** — but two are real user-facing defects worth fixing before or shortly after merge, and the rest are legitimate cleanup.

## Secondary

**1. Invisible "Missing" badge (correctness/visual).** Anchored below at `SemanticMatch.tsx:148`. `StatusBadge`'s new `neutral` tone (`StatusBadge.tsx:47`: `bg-surface-subtle text-content-muted`) is applied inside `VerdictRow`'s `<li>`, which is itself `bg-surface-subtle` (`SemanticMatch.tsx:143`) — both resolve to the same `--color-bg-subtle` token (`theme.css:38`). Every "Missing" verdict's badge pill has no visible fill, just small-caps text with no chip boundary, while "Met"/"Partial" badges render normally colored. This undercuts the component's own docblock claim that status is carried "THREE ways, never by colour alone" — the shape/colour channel silently drops for exactly the status class most worth flagging. Suggest either a distinct `neutral` background (e.g. a step darker/lighter than the row) or a border, so the pill stays visible against a `bg-surface-subtle` row.

**2. Tailor-button gated on keyword coverage only, even when the semantic view is on screen.** `PasteJdPanel.tsx` (`jdContext`, ~line 111) builds the "Tailor résumé to this job" button's visibility and payload from `jdMatch.coverage` (keyword) unconditionally — documented deliberately in the added comment block just above it ("Built from the KEYWORD coverage regardless of which view is on screen... Wiring the semantic verdicts into rewrite steering is its own piece of work"). That's a reasonable scope cut, but the user-visible result is: opt into semantic analysis, see a real "Missing" verdict on screen, and the button that would let you act on it can be hidden or under-informed because keyword coverage says nothing is missing. Since this is explicitly scoped out rather than an oversight, I'd suggest filing a fast-follow issue rather than blocking here — flagging so it isn't lost.

**3. `capability` never resets to `null` on opt-out (`useJdMatch.ts:~301-313`, pre-existing #203 effect, not touched by this diff's added lines).** The gating effect's cleanup only sets a local `cancelled` flag; it never calls `setCapability(null)` when `semanticOptIn` flips back to `false`. This contradicts the field's own docblock, newly added by this PR: "Stays `null` for a keyword-only consumer, since the probe is gated on the opt-in" (`useJdMatch.ts:217-223`) — false after a user has ever opted in once. Currently harmless only because the sole consumer (`SemanticAnalysisOptIn`'s `StatusLine`) checks `checked` before ever reading `capability`. Since `capability` is now public `JdMatchController` API, a future consumer that doesn't replicate that gate would show a stale probe result for an opted-out user. Cheap fix: `if (!semanticOptIn) { setCapability(null); return; }`.

**4. `SemanticAnalysisOptIn.tsx` is 215 LOC**, over CLAUDE.md's "~200 LOC, decompose past that" feature-component rule (confirmed via `wc -l`). `StatusLine`'s ~70-line, 6-branch state machine is the natural extraction candidate — the same pattern this PR already applied when splitting `JdMatch` into `KeywordMatch`/`SemanticMatch`.

**5. Duplicated `<Card>` header markup** between `KeywordMatch.tsx` (moved verbatim from the old `JdMatch.tsx`) and `SemanticMatch.tsx` (`SemanticMatch.tsx:91-97`, byte-identical to `KeywordMatch.tsx`'s header block). Nothing enforces the two staying in sync; a future copy tweak (e.g. renaming "alpha") edited in one file silently drifts from the other. A small shared header component would remove the duplication — the CLAUDE.md Golden Rule reasoning applies between the two feature files here, not just vs. a primitive.

## Nits

**6. `Disclosure.tsx`'s reuse-analysis docblock is now stale.** It enumerates the repo's hand-rolled `<details>` exceptions by name (five feature-code + `ModelLoadProgress`) as the canonical "known debt with an owner" registry — but `Disclosure.tsx` itself isn't touched by this diff, so `SemanticMatch.tsx`'s new per-verdict Evidence `<details>` (which explicitly leans on that registry's reasoning to justify not using the shared primitive) was never added to it. A future batch-conversion effort driven off `Disclosure.tsx`'s list would miss this instance. One-line docblock update, whenever convenient.

**7. Dead defensive narrowing** — `PasteJdPanel.tsx`'s `const jdMatch = keyword?.path === "keyword" ? keyword : null;` (unchanged by this diff) re-narrows a value that, per `useJdMatch`'s own construction (`keywordResult` always builds `{path:"keyword",...}` or `null`), can never be anything else. The hook's `keyword` field is typed `JdMatchResult | null` rather than the narrower `Extract<JdMatchResult,{path:"keyword"}> | null` — the same `Extract` idiom this PR already uses in `KeywordMatch.tsx`/`SemanticMatch.tsx`. Tightening the hook's return type would let the compiler enforce the invariant instead of a runtime check papering over it. Not urgent — pre-existing from #203, this PR just didn't touch it.

**8. `StatusLine`'s if-chain isn't exhaustive over `JdMatchStatus.kind`.** Anchored below at `SemanticAnalysisOptIn.tsx:118`. A future 6th `JdMatchStatus` variant falls through to the final `return null` with no compiler signal, unlike `SemanticMatch.tsx`'s `GROUP_ORDER`/`Record<VerdictStatus,...>` pattern (whose own docblock notes it fails-closed on a new status by design). A `switch` with a `never`-typed default would close this gap.

---
Reviewed by: Claude Sonnet 5 (high) — this addendum incorporates a delayed `/code-review` pass, independently re-verified against source before posting.

PR #869, opened 2026-08-18
chore(db): open database without version parameters in content script to prevent upgrade hangs

s-annam:
## Summary

- `getDB()` opens at this repo's own `DB_VERSION` and requests an upgrade if the stored database is behind — correct for the app (which owns the schema), wrong for the extension's content script, which can carry a pin ahead of a stale, still-open `offlinecv.org` tab and hang IndexedDB's upgrade transaction waiting for that tab's connection to close.
- Adds `getExistingDB()` (versionless open, never triggers an upgrade) plus a parallel `…FromExisting`/`…IntoExisting` sibling for every content-script-reachable function: `getRecord`/`putRecord`/`deleteRecord`/`listRecordsUpdatedSince` (crud.ts), `getSyncCursor`/`setSyncCursor` (sync-cursor.ts), `getJob`/`saveJob` (jobs.ts), `captureJob` (capture.ts — the capture door), `getAllResumes`/`listResumeChoices` (resumes.ts). Each shares its original's logic through a small private `…Via` core parameterized on the opener, rather than duplicating it.
- `getDB()` and the app's own call sites are untouched.
- `getExistingDB()` falls back to `getDB()` when no database exists yet. A versionless open of a nonexistent database creates an empty v1 with **zero object stores**, and every read/write against it throws `NotFoundError` — so the fallback deletes that stray v1 and reopens through the versioned opener. (Deleting matters: `upgrade()`'s first block is guarded `oldVersion < 1`, so upgrading the stray v1 in place would skip it and leave `resumes`/`jobs` uncreated.) There is no hang risk in that branch — the hang needs an older connection still open, and nothing can be holding one to a database that did not exist.
- `packages/core`'s barrel re-exports the `…Existing` variant under the original public name, so a consumer of `@offlinecv/core` is rebound without changing its own import list.

## Scope — what this does NOT reach

**The extension in `extension/` is not fixed by this PR.** That checkout (`116-Ideas/recruidea-extension`, pinned by `extension/offlinecv-pin.json`) predates `@offlinecv/core` and does not import it: its own barrel, `extension/src/offlinecv-core.ts`, re-exports `captureJob`/`getRecord`/`putRecord`/`deleteRecord`/`listRecordsUpdatedSince`/`getSyncCursor`/`setSyncCursor`/`listResumeChoices` **directly from `../../src/lib/storage/*.ts` by relative path**, so it still binds the `getDB()`-backed originals and is still exposed to the hang. Repointing that barrel at the `…Existing` variants is a change in the extension's own repo and cannot be made from here; this PR builds the variants it needs and says so in the docblocks rather than claiming coverage it does not have.

## Test plan

- [x] `npx tsc -b --noEmit` — clean
- [x] `npx vitest run src/lib/storage/` — 277/277 passing, including new `db.test.ts`
- [x] `npx eslint` on every touched file — clean
- [x] `npm run check:core` — barrel still packages/resolves correctly with the renamed re-exports
- [x] Falsified `db.test.ts`: reverted `getExistingDB` to the bare versionless open, confirmed 3 of 5 tests fail (`NotFoundError: No objectStore named jobs in this database`), restored — the tests actually catch the regression they are named for
- [x] `npx fallow audit --base origin/main` — clean
- [x] `npm run test` — 5978 passing, 10 skipped

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -217,7 +217,10 @@ export { deriveJobId, canonicalJobUrl } from "../../../src/lib/storage/job-url.t
  * that captures from elsewhere: it must hand the record to something running in
  * this app's origin rather than write the library itself.
  */
-export { captureJob, type JobCaptureResult } from "../../../src/lib/storage/capture.ts";
+export {
+  captureJobIntoExisting as captureJob,
+  type JobCaptureResult,
+} from "../../../src/lib/storage/capture.ts";
 
 export type { JobRecord, JobStatus, JobOrigin } from "../../../src/lib/storage/types.ts";
 
@@ -250,14 +253,36 @@ export type { JobRecord, JobStatus, JobOrigin } from "../../../src/lib/storage/t
  * is a purge and `softDeleteRecord` is the right call; that one is not on this
  * surface.
  */
+/**
+ * Re-exported under their app-facing names, but bound to the `…FromExisting`/
+ * `…IntoExisting` variant of each: every consumer of THIS package is a content
+ * script in the sense `src/lib/storage/db.ts`'s `getExistingDB()` docblock
+ * describes, and the plain `getRecord`/`putRecord`/`deleteRecord`/
+ * `listRecordsUpdatedSince` open at this repo's pinned `DB_VERSION`, which can
+ * hang a content script when a bump lands here ahead of a stale, still-open app
+ * tab. Renaming on the way out — rather than the consumer importing a
+ * differently-named function — keeps the rebinding invisible to it.
+ *
+ * ⚠️ **This does not reach the extension in `extension/`.** That checkout
+ * (`116-Ideas/recruidea-extension`, pinned by `extension/offlinecv-pin.json`)
+ * predates this package and does not import it: its own barrel,
+ * `extension/src/offlinecv-core.ts`, re-exports these same symbols directly
+ * from `../../src/lib/storage/*.ts` by relative path, so it still binds the
+ * `getDB()`-backed originals and is still exposed to the hang. Repointing that
+ * barrel at the `…Existing` variants is a change in the extension's own repo;
+ * nothing in this file can make it from here.
+ */
 export {
-  getRecord,
-  putRecord,
-  deleteRecord,
-  listRecordsUpdatedSince,
+  getRecordFromExisting as getRecord,
+  putRecordIntoExisting as putRecord,
+  deleteRecordIntoExisting as deleteRecord,
+  listRecordsUpdatedSinceFromExisting as listRecordsUpdatedSince,
 } from "../../../src/lib/storage/crud.ts";
 
-export { getSyncCursor, setSyncCursor } from "../../../src/lib/storage/sync-cursor.ts";
+export {
+  getSyncCursorFromExisting as getSyncCursor,
+  setSyncCursorIntoExisting as setSyncCursor,
+} from "../../../src/lib/storage/sync-cursor.ts";
 
 export type {
   LetterRecord,
@@ -284,7 +309,10 @@ export {
  * newest first — no blob, no parse, nothing that makes a picker built on this a
  * second corpus channel. Same IndexedDB constraint as `captureJob`.
  */
-export { listResumeChoices, type ResumeChoice } from "../../../src/lib/storage/resumes.ts";
+export {
+  listResumeChoicesFromExisting as listResumeChoices,
+  type ResumeChoice,
+} from "../../../src/lib/storage/resumes.ts";
 
 /**
  * The compensation parser. A posting's pay range is free-text prose, and this is
--- a/src/lib/storage/crud.ts
+++ b/src/lib/storage/crud.ts
@@ -9,7 +9,7 @@
  */
 
 import type { IDBPDatabase } from "idb";
-import { getDB, UPDATED_AT_INDEX } from "./db.ts";
+import { getDB, getExistingDB, UPDATED_AT_INDEX } from "./db.ts";
 import { postLibraryChange } from "./library-channel.ts";
 import type { StoreName, StoredRecord, SyncableStoreName } from "./types.ts";
 
@@ -111,9 +111,18 @@ function emitChange(store: StoreName): void {
  * runtime). These helpers operate through a loosely-typed handle on purpose;
  * the store-specific record type is reasserted via the `<T>` parameter, so
  * callers (`resumes.ts` / `jobs.ts`) still get a fully-typed surface.
+ *
+ * Takes the opener as a parameter, defaulting to `getDB`, so this app's own
+ * code keeps opening at `DB_VERSION` (and migrating when due) without having
+ * to say so at every call site. The `…FromExisting`/`…IntoExisting` exports
+ * below are the only other caller, and they pass {@link getExistingDB} — see
+ * that function's docblock for why a content script must never request a
+ * version.
  */
-async function looseDB(): Promise<IDBPDatabase> {
-  return (await getDB()) as unknown as IDBPDatabase;
+async function looseDB(
+  opener: () => Promise<IDBPDatabase<any>> = getDB,
+): Promise<IDBPDatabase> {
+  return (await opener()) as unknown as IDBPDatabase;
 }
 
 /** Upsert a record, stamping timestamps from a single `now`: `createdAt` comes
@@ -145,7 +154,29 @@ export async function putRecord<T extends StoredRecord>(
     Partial<Pick<T, "createdAt" | "updatedAt">>,
   options: { touch?: boolean } = {},
 ): Promise<T> {
-  const db = await looseDB();
+  return putRecordVia(getDB, store, record, options);
+}
+
+/** Same as {@link putRecord}, opened via {@link getExistingDB} instead — the
+ *  variant `@offlinecv/core` re-exports as `putRecord` to a content-script
+ *  consumer. */
+export async function putRecordIntoExisting<T extends StoredRecord>(
+  store: StoreName,
+  record: Omit<T, "createdAt" | "updatedAt"> &
+    Partial<Pick<T, "createdAt" | "updatedAt">>,
+  options: { touch?: boolean } = {},
+): Promise<T> {
+  return putRecordVia(getExistingDB, store, record, options);
+}
+
+async function putRecordVia<T extends StoredRecord>(
+  opener: () => Promise<IDBPDatabase<any>>,
+  store: StoreName,
+  record: Omit<T, "createdAt" | "updatedAt"> &
+    Partial<Pick<T, "createdAt" | "updatedAt">>,
+  options: { touch?: boolean } = {},
+): Promise<T> {
+  const db = await looseDB(opener);
   const now = Date.now();
   const existing = (await db.get(store, record.id)) as T | undefined;
   const written = {
@@ -169,6 +200,17 @@ export async function getRecord<T extends StoredRecord>(
   return (await db.get(store, id)) as T | undefined;
 }
 
+/** Same as {@link getRecord}, opened via {@link getExistingDB} instead — the
+ *  variant `@offlinecv/core` re-exports as `getRecord` to a content-script
+ *  consumer. */
+export async function getRecordFromExisting<T extends StoredRecord>(
+  store: StoreName,
+  id: string,
+): Promise<T | undefined> {
+  const db = await looseDB(getExistingDB);
+  return (await db.get(store, id)) as T | undefined;
+}
+
 /**
  * Every record in a store — **excluding tombstones by default** (#730).
  *
@@ -187,6 +229,18 @@ export async function getAllRecords<T extends StoredRecord>(
   options: { includeDeleted?: boolean } = {},
 ): Promise<T[]> {
   const db = await looseDB();
+  const all = (await db.getAll(store)) as T[];
+  return options.includeDeleted === true ? all : all.filter(isLive);
+}
+
+/** Same as {@link getAllRecords}, opened via {@link getExistingDB} instead —
+ *  what `listResumeChoices`'s `…FromExisting` variant calls, for a
+ *  content-script consumer. */
+export async function getAllRecordsFromExisting<T extends StoredRecord>(
+  store: StoreName,
+  options: { includeDeleted?: boolean } = {},
+): Promise<T[]> {
+  const db = await looseDB(getExistingDB);
   const all = (await db.getAll(store)) as T[];
   return options.includeDeleted === true ? all : all.filter(isLive);
 }
@@ -223,6 +277,20 @@ export async function listRecordsUpdatedSince<T extends StoredRecord>(
   since: number,
 ): Promise<T[]> {
   const db = await looseDB();
+  return (await db.getAllFromIndex(
+    store,
+    UPDATED_AT_INDEX,
+    IDBKeyRange.lowerBound(since, true),
+  )) as T[];
+}
+
+/** Same as {@link listRecordsUpdatedSince}, opened via {@link getExistingDB}
+ *  instead — the variant `@offlinecv/core` re-exports as
+ *  `listRecordsUpdatedSince` to a content-script consumer. */
+export async function listRecordsUpdatedSinceFromExisting<
+  T extends StoredRecord,
+>(store: SyncableStoreName, since: number): Promise<T[]> {
+  const db = await looseDB(getExistingDB);
   return (await db.getAllFromIndex(
     store,
     UPDATED_AT_INDEX,
@@ -276,6 +344,20 @@ export async function deleteRecord(
   id: string,
 ): Promise<void> {
   const db = await looseDB();
+  await db.delete(store, id);
+  emitChange(store);
+}
+
+/** Same as {@link deleteRecord}, opened via {@link getExistingDB} instead —
+ *  the variant `@offlinecv/core` re-exports as `deleteRecord` to a content
+ *  script. `…IntoExisting`, not `…FromExisting`: this family reserves the
+ *  `From` suffix for reads and `Into` for writes, and a hard delete is a write
+ *  (`db.delete` + `emitChange`), like `putRecordIntoExisting` beside it. */
+export async function deleteRecordIntoExisting(
+  store: StoreName,
+  id: string,
+): Promise<void> {
+  const db = await looseDB(getExistingDB);
   await db.delete(store, id);
   emitChange(store);
 }
--- a/src/lib/storage/db.ts
+++ b/src/lib/storage/db.ts
@@ -16,7 +16,7 @@
  * is for structured/binary data only.
  */
 
-import { openDB, type DBSchema, type IDBPDatabase } from "idb";
+import { openDB, deleteDB, type DBSchema, type IDBPDatabase } from "idb";
 import type {
   ResumeRecord,
   JobRecord,
@@ -38,14 +38,14 @@ export const DB_NAME = "offlinecv";
  *  that read this constant would compare the code to itself and pass even if
  *  `upgrade()` never ran.
  *
- *  ⚠️ A bump is a BREAKING change for the browser extension, which builds this
- *  module from a pinned commit of this repo (`extension/offlinecv-pin.json`)
- *  and reaches `getDB()` from a content script in the app's own origin — so it
- *  carries whatever version the pin had. Once the app has upgraded a user's
- *  database, an extension built at the older pin opens it at the lower version
- *  and gets a `VersionError`, breaking its captures. Bump the pin in lockstep
- *  and rebuild it; this repo's CI never compiles the extension, so nothing else
- *  will catch it. */
+ *  ⚠️ When you bump this, bump `extension/offlinecv-pin.json` in lockstep and
+ *  rebuild the extension: this repo's CI never compiles it, so nothing else
+ *  will catch the drift. {@link getExistingDB} is the opener a content script
+ *  should use — it never requests a version, so it cannot hang waiting on a
+ *  stale, still-open app tab — but the extension's own barrel
+ *  (`extension/src/offlinecv-core.ts`) still imports the `getDB()`-backed
+ *  functions by relative path, so the pin is what keeps this constant and that
+ *  build in step. */
 const DB_VERSION = 4;
 
 /** Index name shared by every syncable store — one string so a range query and
@@ -61,6 +61,7 @@ interface OfflineCvDB extends DBSchema {
 }
 
 let dbPromise: Promise<IDBPDatabase<OfflineCvDB>> | null = null;
+let existingDbPromise: Promise<IDBPDatabase<OfflineCvDB>> | null = null;
 
 /** Open (once) and return the shared DB handle. Cached so concurrent callers
  *  share one connection. */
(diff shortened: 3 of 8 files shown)


packages/core/src/index.ts
```diff
@@ -250,14 +253,27 @@ export type { JobRecord, JobStatus, JobOrigin } from "../../../src/lib/storage/t
...
  */
+/**
+ * Re-exported under their app-facing names, but bound to the `…FromExisting`/
+ * `…IntoExisting` variant of each: this consumer IS the content script
+ * `src/lib/storage/db.ts`'s `getExistingDB()` docblock describes, and the
+ * plain `getRecord`/`putRecord`/`deleteRecord`/`listRecordsUpdatedSince` open
+ * at this repo's pinned `DB_VERSION`, which can hang the extension when a
+ * bump lands here ahead of a stale, still-open app tab. Renaming on the way
+ * out — rather than the extension importing a differently-named function —
+ * keeps `extension2/src/offlinecv-core.ts` unchanged.
```
s-annam: **Blocking #1** — this claim names a path (`extension2/src/offlinecv-core.ts`) that doesn't exist anywhere in this repo. The one real, pinned checkout is `extension/` (no "2"), and its own barrel re-exports `captureJob`/`getRecord`/`putRecord`/etc. directly from `src/lib/storage/*.ts` by relative import — never through `@offlinecv/core`. `extension/src/capture/drain.ts` still imports the plain, unrenamed `captureJob`. This PR's fix does not reach the real extension. See the review body for the full trace.

src/lib/storage/db.ts
```diff
@@ -38,14 +38,12 @@ export const DB_NAME = "offlinecv";
...
  *
- *  ⚠️ A bump is a BREAKING change for the browser extension, which builds this
- *  module from a pinned commit of this repo (`extension/offlinecv-pin.json`)
- *  and reaches `getDB()` from a content script in the app's own origin — so it
- *  carries whatever version the pin had. Once the app has upgraded a user's
- *  database, an extension built at the older pin opens it at the lower version
- *  and gets a `VersionError`, breaking its captures. Bump the pin in lockstep
- *  and rebuild it; this repo's CI never compiles the extension, so nothing else
- *  will catch it. */
+ *  ⚠️ Still worth bumping deliberately even though the browser extension no
```
s-annam: **Blocking #1** — "the browser extension no longer calls this function directly" is false for the one real, pinned checkout (`extension/`): its barrel (`extension/src/offlinecv-core.ts`) re-exports `getRecord`/`putRecord`/etc. straight from `src/lib/storage/crud.ts` by relative import, bypassing `@offlinecv/core` (and this fix) entirely. See the review body for the full trace, including the `drain.ts` call site.

src/lib/storage/db.ts
```diff
@@ -131,13 +130,50 @@ export function getDB(): Promise<IDBPDatabase<OfflineCvDB>> {
...
+ * never resolves — the content script hangs, silently, mid-capture, with
+ * nothing in the console to say why.
+ *
+ * `indexedDB.open(name)` with no version argument cannot trigger that: it
+ * attaches to the database's current version, whatever that is, and never
+ * asks for an upgrade. The one case worth knowing: called against a database
+ * that does not exist yet (a profile that has only ever used the extension,
+ * never the app itself), it still creates one — at version 1, with none of
+ * `upgrade()`'s object stores — because IndexedDB has no "version 0" to open
+ * at. Every caller here already reads a missing store defensively, and that
```
s-annam: **Blocking #2** — this is false. When the database doesn't exist yet, `openDB(DB_NAME)` (no version, no upgrade) creates an empty v1 database with **zero object stores**, and `db.get`/`db.put`/`db.getAll`/`db.delete` against a store name that doesn't exist throw a synchronous `NotFoundError` — they don't resolve to `undefined`/`[]`. Verified by reproducing through `captureJobIntoExisting` itself with `fake-indexeddb`. See the review body for the repro trace and a suggested fix (fall back to `getDB()` when no database exists yet — there's no stale-tab hang risk in that branch).

src/lib/storage/crud.ts
```diff
@@ -169,6 +200,17 @@ export async function getRecord<T extends StoredRecord>(
...
 
+/** Same as {@link getRecord}, opened via {@link getExistingDB} instead — the
+ *  variant `@offlinecv/core` re-exports as `getRecord` to the browser
+ *  extension's content script. */
+export async function getRecordFromExisting<T extends StoredRecord>(
+  store: StoreName,
+  id: string,
+): Promise<T | undefined> {
+  const db = await looseDB(getExistingDB);
+  return (await db.get(store, id)) as T | undefined;
```
s-annam, line 211: **Blocking #2** — this is the throw site. Against the empty v1 database `getExistingDB()` creates for a never-visited-app profile, `db.get(store, id)` throws `NotFoundError: No objectStore named jobs in this database` instead of resolving to `undefined`. Reproduced end-to-end via `captureJobIntoExisting` → `getJobFromExisting` → here. See the review body's Blocking #2 for the full trace and suggested fix.

src/lib/storage/jobs.ts
```diff
@@ -38,6 +40,20 @@ export async function saveJob(
     Partial<Pick<JobRecord, "createdAt" | "updatedAt">>, options);
 }
 
+/** Same as {@link saveJob}, opened via `getExistingDB()` instead (through
+ *  `putRecordIntoExisting`) — what {@link captureJob}'s `…IntoExisting`
+ *  variant calls, for the browser extension's content script. */
+export async function saveJobIntoExisting(
```
s-annam, line 37: **Secondary** — this and `getJob`/`getJobFromExisting` below copy-paste their sibling's body instead of sharing a `…Via(opener, ...)` core the way every other file in this diff does (`captureJobVia`, `listResumeChoicesVia`, `setSyncCursorVia`, `putRecordVia`). Suggest extracting `saveJobVia`/`getJobVia` to match.

src/lib/storage/db.ts
```diff
@@ -131,13 +130,50 @@ export function getDB(): Promise<IDBPDatabase<OfflineCvDB>> {
...
+/** Close the open connection(s), if any, and drop the cached handle(s).
+ *  Test-only seam — an open connection blocks `deleteDB`, so a suite that
+ *  wipes the database between cases must close first, then reopen fresh. */
 export async function closeDB(): Promise<void> {
   if (dbPromise !== null) {
     const db = await dbPromise;
     db.close();
     dbPromise = null;
   }
+  if (existingDbPromise !== null) {
```
s-annam: **Secondary** — `closeDB()` awaits `dbPromise` before ever checking `existingDbPromise`; if `dbPromise` is a rejected promise, this throws and `existingDbPromise`'s connection is never closed (a leak in exactly the scenario this function exists to prevent). Test-only surface so the blast radius is narrow, but this diff is what puts a second unguarded resource on the sequence. Suggest closing each independently (try/finally per branch, or `Promise.allSettled`).

src/lib/storage/crud.ts
```diff
@@ -280,6 +348,18 @@ export async function deleteRecord(
   emitChange(store);
 }
 
+/** Same as {@link deleteRecord}, opened via {@link getExistingDB} instead —
+ *  the variant `@offlinecv/core` re-exports as `deleteRecord` to the browser
+ *  extension's content script. */
+export async function deleteRecordFromExisting(
```
s-annam: **Nit** — this is a write (`db.delete` + `emitChange`) but keeps the "FromExisting" suffix this diff otherwise reserves for reads; every other write sibling uses "IntoExisting" (`putRecordIntoExisting`, `saveJobIntoExisting`, `setSyncCursorIntoExisting`, `captureJobIntoExisting`). Arguable either way in English, but worth a consistent convention.

src/lib/storage/resumes.ts
```diff
@@ -43,6 +44,13 @@ export function getAllResumes(): Promise<ResumeRecord[]> {
   return getAllRecords<ResumeRecord>("resumes");
 }
 
+/** Same as {@link getAllResumes}, opened via `getExistingDB()` instead — what
+ *  `listResumeChoices`'s `…FromExisting` variant calls, for the browser
+ *  extension's content script. */
+export function getAllResumesFromExisting(): Promise<ResumeRecord[]> {
```
s-annam: **Nit** — `fallow audit --base origin/main` flags this as an unused export: it's only consumed intra-file by `listResumeChoicesVia`, unlike `getAllResumes` (which has real external consumers). Doesn't need to be public.

```suggestion
function getAllResumesFromExisting(): Promise<ResumeRecord[]> {
```

s-annam (commented):
**Verdict: REQUEST_CHANGES** — 2 Blocking findings. Rule applied: ≥1 Blocking → REQUEST_CHANGES (no auto-fix/collapse attempted; Step 5.5 only runs at 0 Blockers).

No issue is linked to this PR (`closingIssuesReferences` empty, no `Closes/Fixes/Refs #N` in the body), so this review judges the diff against its own stated purpose and the code it touches, not against acceptance criteria.

## Blocking

**1. The fix does not reach the real, pinned browser extension — the hang this PR exists to close is still fully reachable.** (`packages/core/src/index.ts:264`, `src/lib/storage/db.ts:41`)

The nested, pinned checkout at `extension/` (`116-Ideas/recruidea-extension`, pinned via `extension/offlinecv-pin.json`) is the one real content-script consumer present in this repo. Its own barrel, `extension/src/offlinecv-core.ts`, re-exports `captureJob`/`getRecord`/`putRecord`/`deleteRecord`/`listRecordsUpdatedSince`/`getSyncCursor`/`setSyncCursor`/`listResumeChoices` **directly from `../../src/lib/storage/*.ts` by relative import** — never through `@offlinecv/core`. Confirmed: `grep -rn "@offlinecv/core" extension/src` returns zero matches.

Concretely, `extension/src/capture/drain.ts:47` does `import { captureJob, ... } from "../offlinecv-core.ts"`, which resolves to this repo's **plain, unrenamed** `src/lib/storage/capture.ts`'s `captureJob` — the `DB_VERSION`-opening one this PR left untouched — not `captureJobIntoExisting`. Same story for `capture/save.ts`, `capture/queue.ts`, `library/sync-engine.ts`, etc.

The new docblock's claim — "Renaming on the way out … keeps `extension2/src/offlinecv-core.ts` unchanged" (`packages/core/src/index.ts:264`) and "the browser extension no longer calls this function directly" (`src/lib/storage/db.ts:41`) — names a path (`extension2/…`) that does not exist anywhere in this repo; the one real checkout is `extension/` (no "2"), and it *does* still call `getDB()`-backed code, through its own un-updated barrel.

**Failure scenario:** land this PR as-is → the extension's actual capture pipeline (`drain.ts` → `captureJob` → plain `capture.ts` → `getDB()`) still opens at `DB_VERSION=4` and can still hang mid-capture against a stale, still-open app tab. Nothing in this diff repoints `extension/src/offlinecv-core.ts`'s imports, and nothing bumps `extension/offlinecv-pin.json` (still at the pre-PR SHA). The bug this PR is titled to fix ships unfixed.

If "extension2" refers to a second, not-yet-built extension surface this repo has no visibility into, say so explicitly in the PR body and scope the claim to that surface only — as written, a reviewer (and a future bisector) reads it as covering the one real extension in this tree, and it doesn't.

**2. `getExistingDB()` throws `NotFoundError` on the exact case its own docblock says is handled — a first-time extension user capturing before ever opening the app.** (`src/lib/storage/db.ts:154`, `src/lib/storage/crud.ts:211`)

The docblock says: *"Every caller here already reads a missing store defensively, and that empty-library reading is correct."* Verified false by direct reproduction (`fake-indexeddb`, no version, no upgrade callback): a database that doesn't exist yet is created at v1 with **zero object stores**, and `db.get`/`db.put`/`db.getAll`/`db.delete` against a nonexistent store name throw a synchronous `NotFoundError`, not `undefined`/`[]`. I reproduced this through the real code path, not just raw `idb`:

```
NotFoundError: No objectStore named jobs in this database
 ❯ getRecordFromExisting src/lib/storage/crud.ts:211:20
 ❯ getJobFromExisting src/lib/storage/jobs.ts:78:18
 ❯ captureJobVia src/lib/storage/capture.ts:140:20
```

**Failure scenario:** a user installs the extension and captures a job posting before ever visiting offlinecv.org in that profile (or after clearing site data) — exactly the case `getExistingDB`'s own docblock calls out by name. `captureJobIntoExisting` → `getJobFromExisting` → `getRecordFromExisting` → `db.get('jobs', id)` throws instead of resolving to `undefined`. Same on the write side (`putRecordIntoExisting`/`saveJobIntoExisting`). The capture throws instead of hanging — a different symptom, same broken outcome.

`db.test.ts`'s "creates an empty v1 database" test only asserts `db.version`/`objectStoreNames` — it never calls `get`/`put` against that database, so it documents the empty-DB shape without ever exercising the throw.

**Fix:** have `getExistingDB()` detect the store-less/nonexistent-database case (e.g. `objectStoreNames.length === 0`, or probe with `indexedDB.databases()` first) and fall back to the full `getDB()` opener in that branch. There's no stale-open-tab hang risk when no database exists yet — nothing else could be holding an older connection open — so falling back there can't reintroduce the bug this function exists to avoid.

## Secondary

**3. `saveJob`/`saveJobIntoExisting` and `getJob`/`getJobFromExisting` copy-paste their sibling's body instead of sharing a `…Via` core.** (`src/lib/storage/jobs.ts:46`, `:75`)

Every other file in this diff routes both variants through a small shared `…Via(opener, ...)` (capture.ts's `captureJobVia`, resumes.ts's `listResumeChoicesVia`, sync-cursor.ts's `setSyncCursorVia`, crud.ts's `putRecordVia`). `jobs.ts` doesn't — `saveJob`/`saveJobIntoExisting` duplicate the same 5-line body, and `getJob`/`getJobFromExisting` duplicate the same liveness check. A later change to tombstone/liveness semantics applied to one twin isn't structurally guaranteed to reach the other. Suggest extracting `saveJobVia(opener, input, options)` / `getJobVia(opener, id)` to match the pattern used everywhere else in this diff.

**4. `closeDB()` can leak `existingDbPromise`'s connection if `dbPromise` is rejected.** (`src/lib/storage/db.ts:174`)

```ts
export async function closeDB(): Promise<void> {
  if (dbPromise !== null) {
    const db = await dbPromise;   // throws if dbPromise is a rejected promise
    ...
  }
  if (existingDbPromise !== null) { ... }  // never reached in that case
}
```
Test-only surface (only test suites call `closeDB`), so the blast radius is narrow, but this diff is what puts a second unguarded resource on the same sequence. If `dbPromise` was left rejected by a prior `getDB()` failure, `closeDB()` throws before ever closing `existingDbPromise`'s connection — a leak in exactly the scenario this function's own docstring exists to prevent (a suite that needs `deleteDB` to succeed). Suggest closing each independently (`try { } finally { }` per branch, or `Promise.allSettled`).

**5. `DB_VERSION`'s docblock lost its one actionable instruction, and (per Blocking #1) now states something false.**  (`src/lib/storage/db.ts:41`)

The old text named the concrete action: bump `extension/offlinecv-pin.json` in lockstep and rebuild, "this repo's CI never compiles the extension, so nothing else will catch it." The new text drops that pointer entirely. Once Blocking #1 is actually resolved, restore an explicit, actionable pointer here rather than the vaguer "still worth bumping deliberately."

## Nits

**6. `getAllResumesFromExisting` is exported but has no consumer outside its own file.** (`src/lib/storage/resumes.ts:50`) — `fallow audit --base origin/main` flags it as an unused export (unlike its sibling `getAllResumes`, which `resume-library.ts` and several test files import). See inline suggestion.

**7. `deleteRecordFromExisting` keeps the "FromExisting" (read) suffix despite being a write** (calls `db.delete` + `emitChange`), while every other write sibling in this diff uses "IntoExisting" (`putRecordIntoExisting`, `saveJobIntoExisting`, `setSyncCursorIntoExisting`, `captureJobIntoExisting`). Arguable either way in English ("delete FROM the store" reads fine), but worth a consistent convention if you touch this again.

## Gates run

- `npx tsc -b --noEmit` — clean
- `npx eslint` on every changed file — clean
- `npx vitest run src/lib/storage/` — 277/277 passing (including the new `db.test.ts`)
- `npm run check:core` — barrel packages/resolves correctly (note: this gate only asserts export *names* exist, per plain-Node with no `indexedDB` global — it can't exercise IndexedDB behavior, so it wouldn't have caught either Blocking finding)
- `npx fallow audit --base origin/main` — 1 unused export (Nit #6); 1 pre-existing high-complexity function (`captureJobVia`, 31.6 CRAP) re-surfaced by the rename, not a new defect from this diff, not flagged as a finding
- Fixture PII (3a), design-system/reuse (3b), style tokens (3c), skill/script command bugs (3e) — all **skipped**, not applicable to this diff (no fixtures, no components, no styles, no skill/script files touched)
- Description accuracy (3f) — ran; the overclaim folds into Blocking #1 above

## AC checklist

No issue linked to this PR — reviewed against the diff and its own stated purpose only.

---

packages/core/src/index.ts
s-annam: Confirmed, and the claim is gone. `extension/` is gitignored in this checkout, so I verified against the sibling clone: `extension/src/offlinecv-core.ts` re-exports `captureJob`/`getRecord`/`putRecord`/`deleteRecord`/`listRecordsUpdatedSince`/`getSyncCursor`/`setSyncCursor`/`listResumeChoices` from `../../src/lib/storage/*.ts` by relative path, and `grep -rn "@offlinecv/core" extension/src` is empty. The `extension2` path was fiction.

Fixed in `5e944b9`: this docblock now carries an explicit ⚠️ saying the rebinding does **not** reach `extension/`, names why (relative imports, predates this package), and says repointing that barrel is a change in the extension's own repo. The PR body gained a "Scope — what this does NOT reach" section saying the same. Every other `…Existing` docblock that said "the browser extension's content script" now says "a content-script consumer" — true of this package's surface, not a coverage claim.

The actual repointing is filed as 116-Ideas/recruidea-extension#62 (barrel rebinding + pin bump + `no-network.test.ts` re-verification), since nothing in this repo can make it.

Leaving this thread open — the extension is still exposed until #62 lands; only the false claim is resolved here.

src/lib/storage/db.ts
s-annam: Confirmed — "the browser extension no longer calls this function directly" was false, and it has been deleted. `extension/src/offlinecv-core.ts` still binds the `getDB()`-backed originals by relative import.

Fixed in `5e944b9`. The `DB_VERSION` docblock now reads:

> ⚠️ When you bump this, bump `extension/offlinecv-pin.json` in lockstep and rebuild the extension: this repo's CI never compiles it, so nothing else will catch the drift. `getExistingDB` is the opener a content script should use — it never requests a version, so it cannot hang waiting on a stale, still-open app tab — but the extension's own barrel (`extension/src/offlinecv-core.ts`) still imports the `getDB()`-backed functions by relative path, so the pin is what keeps this constant and that build in step.

That restores the actionable instruction Secondary #5 flagged as lost, and states the extension's real binding rather than the aspirational one.

Leaving open alongside the sibling thread until 116-Ideas/recruidea-extension#62 repoints the barrel.

src/lib/storage/db.ts
s-annam: Confirmed by reproduction before changing anything — `captureJobIntoExisting` against a never-created database threw exactly the trace you posted (`NotFoundError: No objectStore named jobs in this database`, `crud.ts:211` ← `jobs.ts:78` ← `capture.ts:140`). The "reads a missing store defensively" sentence was wrong and is gone.

Fixed in `5e944b9`. `getExistingDB()` now detects the store-less shape and falls back to `getDB()`:

```ts
async function openExisting(): Promise<IDBPDatabase<OfflineCvDB>> {
  const db = await openDB<OfflineCvDB>(DB_NAME);
  if (db.objectStoreNames.length > 0) return db;
  db.close();
  await deleteDB(DB_NAME);
  return getDB();
}
```

One wrinkle worth flagging, because the obvious version of your suggested fix is silently broken: **the stray v1 has to be deleted, not upgraded.** `upgrade()`'s first block is guarded `oldVersion < 1`, so handing that v1 straight to `getDB()` skips it and `resumes`/`jobs` are never created — the fallback would return a handle as unusable as the one it replaced. Deleting first makes it a true v0 → v4 migration. Nothing is lost: the database being deleted is the store-less one the call just created and can hold no records by construction. The docblock now carries that reasoning.

Your safety argument for the fallback is in the docblock too: the hang needs an OLDER connection still open, and nothing can hold one to a database that was not there a moment ago.

Tests: `db.test.ts` gains three — the fallback's schema, `getJobFromExisting` resolving `undefined` instead of throwing, and a capture end-to-end through the door. Falsified by reverting `openExisting` to the bare versionless open: 3 of 5 fail with the `NotFoundError`, pass again restored. The old "creates an empty v1 database" test — which, as you noted, documented the shape without exercising the throw — is replaced by the fallback assertion.

src/lib/storage/crud.ts
s-annam: Confirmed at this throw site and fixed in `5e944b9` — see the full reply on the `getExistingDB` thread (`db.ts:154`). `getRecordFromExisting` no longer sees a store-less database: `getExistingDB()` falls back to `getDB()` (deleting the stray v1 first, so `upgrade()`'s `oldVersion < 1` block actually runs) when none exists yet.

`db.test.ts` now asserts this line's behaviour directly — `getJobFromExisting("nope")` resolves to `undefined` on a never-visited profile — and that test fails with your exact `NotFoundError` when the fix is reverted.

src/lib/storage/jobs.ts
s-annam: Done in `5e944b9` — extracted `saveJobVia(put, input, options)` and `getJobVia(get, id)`, both private, both parameterized on the crud function the way `captureJobVia` / `putRecordVia` / `setSyncCursorVia` / `listResumeChoicesVia` are. The permissive-write cast now has one copy, and so does the `record !== undefined && isLive(record)` tombstone check, so the drift you described can't happen.

src/lib/storage/db.ts
s-annam: Fixed in `5e944b9`. Both handles are now cleared *before* either is awaited, and awaited through `allSettled`:

```ts
export async function closeDB(): Promise<void> {
  const open = [dbPromise, existingDbPromise];
  dbPromise = null;
  existingDbPromise = null;
  await Promise.allSettled(
    open.map(async (pending) => {
      (await pending)?.close();
    }),
  );
}
```

A rejected opener now settles as a no-op instead of throwing past the second branch, so it cannot strand the other connection. Clearing the cached handles first also means a rejected `dbPromise` no longer survives the call — previously it stayed cached, so every later `getDB()` re-threw the same stale rejection.

The docblock notes the case this diff newly makes possible: after the Blocking #2 fallback, the two promises can resolve to the **same** handle, and `close()` is idempotent.

src/lib/storage/crud.ts
s-annam: Taken — renamed to `deleteRecordIntoExisting` in `5e944b9`, with the reason in the docblock so the next reader doesn't re-litigate the English:

> `…IntoExisting`, not `…FromExisting`: this family reserves the `From` suffix for reads and `Into` for writes, and a hard delete is a write (`db.delete` + `emitChange`), like `putRecordIntoExisting` beside it.

`packages/core/src/index.ts` updated to match; it still re-exports as `deleteRecord`, so the public name is unchanged.

src/lib/storage/resumes.ts
s-annam: Taken in `5e944b9` — `getAllResumesFromExisting` is now a plain `function`, not exported. `npx fallow audit --base origin/main` is clean, which also clears the failing `fallow` check on this PR (that alert was this one export; it was the only finding).

Added a reason to the docblock beyond "no consumer", since that is the part a future reader can act on wrongly:

> Private: unlike `getAllResumes` it has no consumer outside this file, and exporting it would put a whole résumé corpus (blobs included) on a surface only the narrow `ResumeChoice` projection is meant to reach.

src/lib/storage/resumes.ts
```diff
@@ -43,6 +44,13 @@
...
```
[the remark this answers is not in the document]
s-annam: Fixed in `5e944b9` — `getAllResumesFromExisting` is no longer exported (it is only consumed intra-file by `listResumeChoicesVia`). `npx fallow audit --base origin/main` now reports no issues in the 8 changed files.

PR #871, opened 2026-08-19
docs(score): document the anonymous ATS scoring rubric (#43)

s-annam:
## Summary

Adds `docs/scoring.md` — a short, plain-language explainer of the anonymous ATS
scoring rubric (Specificity/Structure/Completeness weights, the layout-trigger
penalty table, verdict-band thresholds, algo version), linked from README.

Every number is cited to the literal constant in `src/lib/score/score.ts` that
produces it. Note: the anonymous scorer (`computeAnonymousAtsScore`, the surface
this doc describes) hardcodes its 40/30/30 point ceilings inline — it does not
read the `WEIGHTS` constant, which drives a separate authed scorer
(`computeAtsScore`) instead. The doc cites the actual hardcoded multipliers and
flags that the two scorers can drift independently.

## Verification

- Every constant cited in the doc re-checked against current `score.ts`.
- `npm run typecheck` / `npm run lint` clean (docs-only change, no test impact).

## Adversarial review

Round-1 adversarial review flagged that the 40/30/30 citation pointed at the
wrong constant (`WEIGHTS`, which drives the *other* scorer, not this one) — fixed
and re-verified in round 2 (citation now points at the literal `* 40`/`* 30`/
`* 30` multipliers in `computeAnonymousAtsScore`, plus a note on the two scorers'
independent hardcoding). Clean since.

Closes #43

PR #872, opened 2026-08-19
feat(export): pre-download validation pass for glyph loss and bullet-page splits (#621)

### LINKED ISSUE #621: [export] pre-download validation pass — surface known render defects instead of failing silently
## Summary

The PDF exporter fails **silently** on input it cannot lay out cleanly. Two confirmed instances, both found only by dogfooding an export and inspecting the bytes:

- a character outside the embedded font subset is written to the text layer as `U+0000`, with no warning
- a page break falls inside a bullet wrapping to 4+ lines, splitting a sentence across pages — #630 already makes shorter (2–3 line) bullets indivisible, so this is the residual case

The glyph-drop point fix is filed separately as #617 (still open). The bullet-split point fix was originally tracked as #619, which was closed as a duplicate; the actual fix landed in #630 — a deliberate partial fix (bullets wrapping to 2–3 lines can no longer split across a page break; bullets of 4+ lines still can, as a page-density trade-off made in review). This issue proposes the shared missing layer: a **pre-export validation pass** that tells the user "this document will render with known defects" *before* they download, instead of leaving them to hexdump their own résumé.

## Why a shared pass, not just the two point fixes

The two point fixes each remove one defect (or, for bullet-split, narrow it). Neither adds a channel for reporting the next one. This class — the renderer quietly degrading rather than refusing or warning — will keep producing instances: overlong unbreakable tokens, a header orphaned at a page bottom, a skills row wider than the content column, a field truncated at a hard cap.

Product framing matters here too. The Download PDF is presented as the ATS-safe artifact. Shipping it with silent corruption in the text layer — the exact surface an ATS reads — undercuts the one claim the lane makes.

## Scope

Add a validation pass over the `AtsResumeModel` + render result that returns structured findings, and surface them on the download affordance.

**Findings to cover initially**

| Finding | Detected where | Required for this issue |
|---|---|---|
| Character degraded / dropped for lack of glyph coverage | render-time, embedded-font path | **Yes** |
| Bullet split across a page break (4+ line bullets only — see #630) | render-time, pagination | **Yes** |
| Entry header orphaned at page bottom | render-time, pagination | No — deferred; likely already covered by #630's page-break widow/orphan rule |
| Text overflowing the content column | render-time, wrap fallback | No — deferred; no known repro yet, kept here to document the finding shape should accommodate it later |

Each finding carries `{ kind, severity, sourceField, detail }` — `sourceField` matters most: "an arrow was dropped" is not actionable, "an arrow was dropped from Experience → role 2 → bullet 3" is.

**Where it lands**

`src/lib/pdf/render-audit-report.ts` already exists and is the natural home for the report shape. The renderer accumulates findings during the draw pass (it is the only thing that knows about glyph coverage and page geometry) and returns them alongside the bytes.

**Surface**

Per `CLAUDE.md`, there is no toast primitive — confirmed. Swap content in the mounted result region with a `StatusBadge` + `aria-live="polite"`; never colour alone. Zero findings must render nothing, not an empty "0 issues" chrome.

**Out of scope: call sites.** Consumers of `renderAtsResumePdf`'s current return value (the export dialog, `corpus-roundtrip.test.ts`, the `render-roundtrip-*.repro.test.ts` family) are not updated by this issue beyond what's needed to keep them green against the new `{ bytes, findings }` shape — wiring the findings *into* those call sites' own UX/behavior is deferred, not filed as a follow-up.

## Acceptance criteria

- [ ] `renderAtsResumePdf` returns `{ bytes, findings }` (or equivalent) without changing the bytes it produces today for a clean résumé.
- [ ] A résumé containing an uncovered glyph yields a finding naming the character **and** its source field.
- [ ] A résumé containing a bullet that wraps to 4+ lines and falls across a page break yields a pagination finding naming the entry. 2–3 line bullets are not expected to reproduce this (per #630) and don't need their own test.
- [ ] A clean résumé yields zero findings and renders no warning chrome.
- [ ] The warning surface reuses existing design-system pieces — no new banner primitive (Golden Rule).
- [ ] `corpus-roundtrip.test.ts` and the `render-roundtrip-*.repro.test.ts` family green; no rendered byte changes.

Entry-header-orphan and column-overflow detection are **not** required by this issue — see the Findings table above.

## Sequencing

Lower priority than the glyph-drop point fix (#617), which is still open. #630 has already landed for bullet-split, narrowing it to the 4+-line case described above. #617 can land before or after this issue; building the report first means shipping a warning for the glyph defect we already know how to prevent.

## Non-goals

- Not blocking the download. A finding is advisory — the user still gets their PDF.
- Not a score input. This reports on the *export*, not on résumé quality; it must not touch `computeAnonymousAtsScore`.
- Not updating call sites of `renderAtsResumePdf` beyond the return-shape change — see Scope.
- Not implementing header-orphan or column-overflow detection — see Scope/Findings table.

s-annam:
## Summary

Adds a pre-download validation pass to the Download PDF renderer: `renderAtsResumePdf`
now returns `{ bytes, findings }` instead of just `bytes`, without changing the bytes
produced for a clean résumé.

Two finding kinds, both advisory (never blocks the download):
- **Glyph loss** — a character outside the embedded font's subset degrades/drops in
  the text layer; the finding names the character and its source field.
- **Bullet page-split** — a 4+ line bullet wraps across a page break (2–3 line bullets
  are already indivisible per #630 and correctly do not trigger this).

Clean résumés produce zero findings and render no warning chrome. The warning surface
(`ExportFindings`, new) reuses `StatusBadge` from `@design-system` — no new banner
primitive.

Out of scope (per the issue): entry-header-orphan and column-overflow detection;
wiring findings into any UX beyond the export dialog itself.

## Verification

- `corpus-roundtrip.test.ts` and all 13 `render-roundtrip-*.repro.test.ts` green,
  no rendered byte changes — verified via a draw-signature comparison (page count +
  every pdfjs draw operator + args + link annotations) across all 58 corpus fixtures,
  21 multi-page configs, and a glyph-degrading model, since pdf-lib stamps a
  timestamp so a literal byte diff isn't meaningful.
- New `render-findings.test.ts` (14 tests): glyph-loss naming, 4+-line split naming,
  2–3 line non-reproduction, zero-findings-on-clean, fail-before proof (detectors
  neutered → 6/14 fail; restored → all pass).
- `npm run typecheck` / `npm run lint` clean. Full suite green as part of the
  whole-stack verification (see #874's summary).

## Stack position

Layer 2 of 4. Base: #871. Depends on #871.

## Adversarial review

Reviewed across both rounds; this layer had **nits only**, no blocking findings:
- No explicit test pins byte-identical output for a clean résumé at the render level
  (verified structurally instead — the glyph-loss detector reads the embedded font
  via `fontkit` directly, not through pdf-lib's `PdfFont`, so it can't perturb the
  embedded subset it inspects).
- A couple of pointless expression statements in `render-ats-pdf.fonts.test.ts`.
- A content-derived React key in `ExportFindings` could theoretically collide if two
  entries share the same header line + finding kind + detail.

None blocking; left for a human pass if worth chasing.

Closes #621

### DIFF
--- a/src/components/features/ExportFindings.tsx
+++ b/src/components/features/ExportFindings.tsx
@@ -1,0 +1,89 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * ExportFindings — what the exporter could not render cleanly, shown on the row
+ * that produced the file (#621).
+ *
+ * Reuse analysis (CLAUDE.md Golden Rule). This is NOT a new banner primitive.
+ * The design-system barrel deliberately ships no Toast/Snackbar, and its house
+ * pattern for "an action finished and has something to say" is to confirm IN the
+ * surface already mounted rather than open a new one — so this swaps content
+ * into `ExportRow`, beside the `ErrorState` that already reports a failed
+ * export, and carries its status on the shared `StatusBadge`. Nothing new is
+ * added to the design system.
+ *
+ * Three rules it exists to keep:
+ *
+ *  1. **Zero findings renders NOTHING.** Not an empty "0 issues" panel, not a
+ *     green all-clear. Most résumés are clean, and a permanent status strip on
+ *     the download row would teach every user to stop reading it.
+ *  2. **Never colour alone.** The badge carries the WORD, the sentence states
+ *     the count, and every finding names its own field — the tone is
+ *     reinforcement.
+ *  3. **Advisory, never a blocker.** The user already has their PDF by the time
+ *     this renders; the copy says so. Refusing a download is `useDownloadPdf`'s
+ *     #664 font gate and stays there.
+ *
+ * `aria-live="polite"` announces the swap. The parent format list is itself a
+ * polite live region, so the announcement would happen regardless; declaring it
+ * here keeps the guarantee attached to the component that needs it rather than
+ * to a container that could be restructured.
+ */
+
+import { StatusBadge } from "@design-system";
+import type { RenderFinding } from "../../lib/pdf/render-findings.ts";
+
+/**
+ * How many findings are listed before the rest are counted. A résumé pasted from
+ * a source full of arrows can produce one finding per bullet; forty rows in a
+ * dialog is a wall, and the first few already say what kind of thing is wrong
+ * and where to start.
+ */
+const MAX_LISTED = 5;
+
+export function ExportFindings({
+  findings,
+}: {
+  findings: readonly RenderFinding[];
+}) {
+  if (findings.length === 0) return null;
+  const listed = findings.slice(0, MAX_LISTED);
+  const rest = findings.length - listed.length;
+  // `warning` when anything was actually destroyed; `info` when every finding is
+  // a substitution that still reads correctly (an arrow drawn as "->").
+  const tone = findings.some((f) => f.severity === "warning") ? "warning" : "info";
+
+  return (
+    <div aria-live="polite" className="flex flex-col gap-2">
+      <div className="flex flex-wrap items-center gap-2">
+        <StatusBadge tone={tone}>Check the export</StatusBadge>
+        <p className="text-sm text-content-secondary">
+          Your PDF downloaded, but{" "}
+          {findings.length === 1
+            ? "one thing"
+            : `${findings.length} things`}{" "}
+          did not come out as written.
+        </p>
+      </div>
+      <ul className="flex flex-col gap-1">
+        {listed.map((finding, idx) => (
+          <li
+            key={`${finding.kind}:${finding.sourceField}:${finding.detail}:${idx}`}
+            className="text-sm text-content-tertiary"
+          >
+            <span className="font-medium text-content-secondary">
+              {finding.sourceField}
+            </span>{" "}
+            — {finding.detail}
+          </li>
+        ))}
+      </ul>
+      {rest > 0 && (
+        <p className="text-sm text-content-tertiary">
+          …and {rest} more like {rest === 1 ? "this" : "these"}.
+        </p>
+      )}
+    </div>
+  );
+}
--- a/src/lib/pdf/render-ats-pdf.fonts.test.ts
+++ b/src/lib/pdf/render-ats-pdf.fonts.test.ts
@@ -123,7 +123,7 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
     vi.resetModules();
     const { renderAtsResumePdf } = await import("./render-ats-pdf.ts");
 
-    const bytes = await renderAtsResumePdf(model("Poppins embed check"));
+    const { bytes } = await renderAtsResumePdf(model("Poppins embed check"));
     expect(fetchMock).toHaveBeenCalled();
     await expect(hasEmbeddedFontFile2(bytes)).resolves.toBe(true);
   });
@@ -133,7 +133,7 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
     vi.resetModules();
     const { renderAtsResumePdf } = await import("./render-ats-pdf.ts");
 
-    const bytes = await renderAtsResumePdf(model("fallback check"));
+    const { bytes } = await renderAtsResumePdf(model("fallback check"));
     expect(bytes).toBeInstanceOf(Uint8Array);
     expect(bytes.length).toBeGreaterThan(500);
     await expect(hasEmbeddedFontFile2(bytes)).resolves.toBe(false);
@@ -146,7 +146,7 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
     const { renderAtsResumePdf: renderEmbedded } = await import(
       "./render-ats-pdf.ts"
     );
-    const embeddedBytes = await renderEmbedded(model("Łukasz, Wrocław"));
+    const { bytes: embeddedBytes } = await renderEmbedded(model("Łukasz, Wrocław"));
     const embeddedText = await extractPdfText(embeddedBytes);
     expect(embeddedText).toContain("ł");
 
@@ -157,7 +157,7 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
     const { renderAtsResumePdf: renderFallback } = await import(
       "./render-ats-pdf.ts"
     );
-    const fallbackBytes = await renderFallback(model("Łukasz, Wrocław"));
+    const { bytes: fallbackBytes } = await renderFallback(model("Łukasz, Wrocław"));
     const fallbackText = await extractPdfText(fallbackBytes);
     expect(fallbackText).not.toContain("ł");
     expect(fallbackText).toContain("?");
@@ -186,7 +186,7 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
         vi.resetModules();
         const { renderAtsResumePdf } = await import("./render-ats-pdf.ts");
 
-        const bytes = await renderAtsResumePdf(model(`Alpha ${ch} Omega`));
+        const { bytes } = await renderAtsResumePdf(model(`Alpha ${ch} Omega`));
         const text = await extractPdfText(bytes);
 
         // The defect, stated directly.
@@ -209,7 +209,7 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
       vi.resetModules();
       const { renderAtsResumePdf } = await import("./render-ats-pdf.ts");
 
-      const bytes = await renderAtsResumePdf(model("Łukasz → Wrocław"));
+      const { bytes } = await renderAtsResumePdf(model("Łukasz → Wrocław"));
       const text = await extractPdfText(bytes);
 
       await expect(hasEmbeddedFontFile2(bytes)).resolves.toBe(true);
@@ -238,7 +238,7 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
         vi.resetModules();
         const { renderAtsResumePdf } = await import("./render-ats-pdf.ts");
 
-        const bytes = await renderAtsResumePdf(model("Alpha \0 Omega"));
+        const { bytes } = await renderAtsResumePdf(model("Alpha \0 Omega"));
         const text = await extractPdfText(bytes);
 
         expect(text).not.toContain("\0");
@@ -259,7 +259,7 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
       vi.resetModules();
       const { renderAtsResumePdf } = await import("./render-ats-pdf.ts");
 
-      const bytes = await renderAtsResumePdf({
+      const { bytes } = await renderAtsResumePdf({
         contact: { name: "Jane Candidate", links: [] },
         sections: [
           {
@@ -289,7 +289,7 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
       vi.resetModules();
       const { renderAtsResumePdf } = await import("./render-ats-pdf.ts");
 
-      const bytes = await renderAtsResumePdf(model("http error check"));
+      const { bytes } = await renderAtsResumePdf(model("http error check"));
 
       expect(fetchMock).toHaveBeenCalled();
       // The load bailed on the status. Without the check we would have read the
@@ -310,7 +310,7 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
       vi.resetModules();
       const { renderAtsResumePdf } = await import("./render-ats-pdf.ts");
 
-      const first = await renderAtsResumePdf(model("attempt one"));
+      const { bytes: first } = await renderAtsResumePdf(model("attempt one"));
       await expect(hasEmbeddedFontFile2(first)).resolves.toBe(false);
       const callsWhileFailing = failing.mock.calls.length;
       expect(callsWhileFailing).toBeGreaterThan(0);
@@ -319,7 +319,7 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
       // point. The same module instance, with its cache already poisoned by the
       // failure above, must issue a fresh request.
       const recovered = stubFetchSucceeds();
-      const second = await renderAtsResumePdf(model("attempt two"));
+      const { bytes: second } = await renderAtsResumePdf(model("attempt two"));
 
       expect(recovered).toHaveBeenCalled();
       await expect(hasEmbeddedFontFile2(second)).resolves.toBe(true);
--- a/src/lib/pdf/render-findings.ts
+++ b/src/lib/pdf/render-findings.ts
@@ -1,0 +1,301 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * render-findings — the structured defect report the résumé exporter returns
+ * alongside its bytes (#621).
+ *
+ * The Download PDF is presented as the ATS-safe artifact, and until now the
+ * renderer degraded silently on input it could not lay out cleanly: a character
+ * outside the embedded font's coverage was written to the text layer as a
+ * replacement (or, before #664, as U+0000), and a page break could fall inside a
+ * long bullet. Both were only ever found by hexdumping a downloaded résumé. This
+ * module defines what a finding IS; `render-ats-pdf.ts` accumulates them during
+ * the draw pass, because it is the only layer that knows which font was actually
+ * embedded and where the page breaks landed.
+ *
+ * A finding is ADVISORY. It never blocks the download (the user still gets the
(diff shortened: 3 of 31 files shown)


Samhit21 (commented):
## PR Review: feat(export): pre-download validation pass for glyph loss and bullet-page splits

### Summary
The central claim — that `bytes` is unchanged for a clean résumé — is structurally sound rather than merely tested. The glyph pass runs against `loadFonts`' own sanitizer before any page is touched and reads fontkit faces rather than pdf-lib's embedded subset, so it cannot perturb what it inspects; the pagination pass observes `pageOrdinal` after a reservation that was already made, with `index > 0` as the whole split test. I checked the two ways that could still go wrong and neither does: exactly two `addPage` sites exist (constructor and `newPage`), and both `ensureWrappedLine` callers pass `(i, lines.length)`, so the counts in the message are right on the bold-runs path as well as the plain one. `collectModelTextFields` covers every drawn field across `AtsContact`, `AtsEntry` and `AtsResumeModel` — no drawn string is unscanned. Folding the #664 probe's walk into it is the right call, and the two defects it corrects on the way are real.

I also checked the `drawn === ""` branch in `isInvisibleDegradation` for a silently-dropped *visible* glyph, since that would be the worst failure this design could have — a total loss reported as nothing. Not reachable: the only characters either sanitizer maps to empty are C0/C1 controls and the four zero-width entries explicitly in the table. Worth stating because the branch reads more broadly than it behaves.

Two findings, both about the finding *text* rather than its detection.

### Findings
1. `render-findings.ts:166` — every link collapses to `sourceField: "Links"`, so a link finding names five candidates and no way to choose; the `path` mechanism to fix it is already used by bullets. [Suggestion]
2. `render-findings.ts:247` — the detail interpolates the raw character, so an invisible one absent from the table (`U+FE0F` after an emoji, a combining mark from NFD text) yields a warning that reads `no glyph for ""`. [Suggestion]

### Scope note, not a finding
This PR changes when the **#664 download refusal** fires, which the title ("validation pass") does not suggest. Routing `findExportGlyphLosses` through the shared walk means it now scans the summary heading (a real loss it previously missed) and no longer scans emphasis sentinels (a false refusal it previously raised on every auto-bolded achievement header). Both move the right way and the docblock says so plainly. Flagging it only because a user-facing gate that blocks downloads changed behaviour inside a PR whose stated subject is advisory reporting — worth a line in the body so a bisecting reader finds it here.

Reviewed against the stack base `gh-batch-01-scoring-rubric-doc`, not `main`.

### Verdict
**Action: COMMENT**

src/lib/pdf/render-findings.ts
```diff
@@ -0,0 +1,291 @@
...
+  };
+
+  const { contact } = model;
+  add("Name", contact.name);
+  add("Headline", contact.headline);
+  add("Email", contact.email);
+  add("Phone", contact.phone);
+  add("Location", contact.location);
+  add("Work authorization", contact.workAuthorization);
+  for (const link of contact.links) add("Links", link);
```
Samhit21: [Suggestion]: every link collapses to one `sourceField`, so a link finding can't be acted on

This module's docblock states the standard it holds itself to: *"`sourceField` is the part that makes a finding actionable: 'an arrow was dropped' is not, 'an arrow was dropped from Experience → Staff Engineer · Acme → bullet 3' is."* Bullets get `→ bullet 3`, entries get their header line, contact fields are singular so their name suffices.

Links are the one repeated field that doesn't get an index. A résumé with LinkedIn, GitHub, a portfolio and two project URLs produces `Links — The export font has no glyph for "–", so it was drawn as "?"`, and the user has five candidates and no way to tell which. That is the case the docblock says is not actionable.

The mechanism is already here — `add` takes a `path` override, which is exactly how bullets get theirs one screen down:

```ts
contact.links.forEach((link, i) =>
  add("Links", link, { path: `Links → link ${i + 1}` }),
);
```

`where` should stay `"Links"` either way: `findExportGlyphLosses` reads it for the #664 refusal sentence, which names fields to check rather than enumerating them, and that comment is right to keep the coarse label coarse.

src/lib/pdf/render-findings.ts
```diff
@@ -0,0 +1,291 @@
...
+      const drawn = sanitize(ch);
+      if (drawn === ch || isInvisibleDegradation(ch, drawn)) continue;
+      findings.push({
+        kind: "glyph-degraded",
+        // A "?" destroys the character outright; a transliteration keeps its
+        // meaning and only changes its spelling.
+        severity: drawn === "?" ? "warning" : "info",
+        sourceField: field.path,
+        detail:
+          `The export font has no glyph for "${ch}", so it was drawn as ` +
```
Samhit21: [Suggestion]: the detail interpolates the raw character, so an invisible one reads as `no glyph for ""`

The table handles the four zero-width characters by mapping them to `""`, and `isInvisibleDegradation` correctly drops those before they become findings. But that only covers characters the table knows. Anything invisible and *absent* from it falls through to `"?"`, which is a visible substitution, so a finding is produced — with the character interpolated into the message, where it renders as nothing.

The reachable case is the emoji variation selector. `U+FE0F` is not in `WINANSI_TRANSLITERATIONS` (nor are `U+FE0E`, the bidi controls `U+200E`/`U+202A`–`E`, or combining marks like `U+0301` from NFD text, which macOS pasteboards produce routinely). So a bullet containing `✓️` — check mark plus VS16, one paste from a job description away — emits two findings:

```
Experience → … → bullet 2 — The export font has no glyph for "✓", so it was drawn as "?".
Experience → … → bullet 2 — The export font has no glyph for "", so it was drawn as "?".
```

The second is a `severity: "warning"` row about a character the user cannot see, cannot search for, and cannot remove. It also inflates the count in `ExportFindings`' "N things did not come out as written."

Naming the code point makes every one of these actionable without changing the common case much:

```ts
detail:
  `The export font has no glyph for "${ch}" (U+${ch.codePointAt(0)!
    .toString(16).toUpperCase().padStart(4, "0")}), so it was drawn as "${drawn}".`,
```

Worth a test alongside the three in `describe("invisible degradations are not reported")` — those pin the *dropped* case, and this is the neighbouring one where an invisible character survives into a finding.

s-annam (commented):
### Verdict: APPROVE (0 Blockers, submitted as review comment due to GitHub self-review constraint)

Clean, well-bounded implementation of the pre-download export validation pass (#621). Returns `{ bytes, findings }` without altering output bytes for clean résumés, implements both required finding categories (glyph loss and 4+ line bullet splits across page breaks) with specific `sourceField` paths, and integrates seamlessly into `ExportDialog` via `ExportFindings` with accessible badges.

### Acceptance Criteria Checklist (Issue #621)
- [x] `renderAtsResumePdf` returns `{ bytes, findings }` without changing the bytes produced for a clean résumé.
- [x] A résumé containing an uncovered glyph yields a finding naming the character and its source field.
- [x] A résumé containing a bullet that wraps to 4+ lines and falls across a page break yields a pagination finding naming the entry (2–3 line bullets are indivisible per #630 and do not trigger).
- [x] A clean résumé yields zero findings and renders no warning chrome.
- [x] The warning surface reuses existing design-system pieces (`StatusBadge` — no new banner primitive).
- [x] `corpus-roundtrip.test.ts` and all `render-roundtrip-*.repro.test.ts` suites green; no rendered byte changes.

### OfflineCV Gates
- **3a Fixture PII:** Clean (no fixtures modified; `check:fixtures` passed).
- **3b Design System & Reuse:** Clean (new `ExportFindings` contains written reuse analysis and uses `@design-system` primitives).
- **3c Style Tokens:** Clean (semantic tokens only; no hardcoded hex or raw palettes).
- **3d Dead Code / Fallow:** Clean (`npm run verify` passed with 0 dead code).
- **3e Scripts / Skills:** N/A (no skill or script files touched).
- **3f Description Accuracy:** Accurate and verifiable against the diff.

### Findings Summary
- **Blocking:** 0
- **Secondary:** 0
- **Nits:** 2 (provided as inline suggestions)

---

src/lib/pdf/render-ats-pdf.fonts.test.ts
```diff
@@ -330,9 +330,9 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
       vi.resetModules();
       const { renderAtsResumePdf } = await import("./render-ats-pdf.ts");
 
-      await renderAtsResumePdf(model("first"));
+      (await renderAtsResumePdf(model("first"))).bytes;
```
s-annam:
```suggestion
      await renderAtsResumePdf(model("first"));
```
Nit: Unused `.bytes` property access on discarded expression statement.

src/lib/pdf/render-ats-pdf.fonts.test.ts
```diff
@@ -330,9 +330,9 @@ describe("Poppins font embed (#314)", { timeout: 20000 }, () => {
...
       const afterFirst = fetchMock.mock.calls.length;
-      await renderAtsResumePdf(model("second"));
+      (await renderAtsResumePdf(model("second"))).bytes;
```
s-annam:
```suggestion
      await renderAtsResumePdf(model("second"));
```
Nit: Unused `.bytes` property access on discarded expression statement.

src/components/features/ExportFindings.tsx
```diff
@@ -0,0 +1,89 @@
...
+          {findings.length === 1
+            ? "one thing"
+            : `${findings.length} things`}{" "}
+          did not come out as written.
+        </p>
+      </div>
+      <ul className="flex flex-col gap-1">
+        {listed.map((finding) => (
+          <li
+            key={`${finding.kind}:${finding.sourceField}:${finding.detail}`}
```
s-annam:
```suggestion
        {listed.map((finding, idx) => (
          <li
            key={`${finding.kind}:${finding.sourceField}:${finding.detail}:${idx}`}
```
Nit: Suffixing map index `idx` avoids potential React key collisions if two identical findings occur.

src/lib/pdf/render-findings.ts
s-annam: Fixed in `3d13ef7` — indexed individual contact links as `Links → link ${i + 1}` in `collectModelTextFields`, with unit tests added in `render-findings.test.ts`.

src/lib/pdf/render-findings.ts
s-annam: Fixed in `3d13ef7` — interpolated the character's hex Unicode code point (`U+XXXX`) in the detail message so invisible and variation characters remain actionable, covered with a unit test.

src/lib/pdf/render-ats-pdf.fonts.test.ts
s-annam: Fixed in `3d13ef7` — removed unused `.bytes` property access on discarded expression statement.

src/lib/pdf/render-ats-pdf.fonts.test.ts
s-annam: Fixed in `3d13ef7` — removed unused `.bytes` property access on discarded expression statement.

src/components/features/ExportFindings.tsx
s-annam: Fixed in `3d13ef7` — appended index `idx` to the React key.

src/lib/pdf/render-findings.ts
```diff
@@ -0,0 +1,296 @@
...
+      const drawn = sanitize(ch);
+      if (drawn === ch || isInvisibleDegradation(ch, drawn)) continue;
+      findings.push({
+        kind: "glyph-degraded",
+        // A "?" destroys the character outright; a transliteration keeps its
+        // meaning and only changes its spelling.
+        severity: drawn === "?" ? "warning" : "info",
+        sourceField: field.path,
+        detail:
+          `The export font has no glyph for "${ch}" (U+${ch
```
Samhit21, line 254: [Suggestion]: the code point names the DRAWN character, not the one the user typed

This is my own suggestion from the last round coming back — the fix is right, and the added precision is what exposes the remaining half.

`drawnText` is upper-cased for a heading (line 235), and the loop then reports `ch` from *that* string. So for a section heading `µ-services`:

```
authored: "µ" U+00B5
drawn   : "Μ" U+039C
message : The export font has no glyph for "Μ" (U+039C), so it was drawn as "?".
```

The user searches their résumé for U+039C and finds nothing — they typed U+00B5. Before the code point was printed, the two glyphs at least looked alike; now the message names an identifier that is provably absent from their input, which is worse for the actionability standard this module sets for itself.

The case transform still has to happen for *detection* — the docblock on `uppercase` is right that scanning raw text misses the loss. It just shouldn't decide what gets *named*. Probing per-character keeps both:

```ts
for (const ch of field.text) {           // authored text
  if (seen.has(ch)) continue;
  seen.add(ch);
  const drawn = sanitize(field.uppercase ? ch.toUpperCase() : ch);
  if (drawn === ch || isInvisibleDegradation(ch, drawn)) continue;
  // …report `ch` — what the user can actually find and fix
}
```

One wrinkle worth deciding rather than inheriting: `ch.toUpperCase()` can expand 1→2 (`ß` → `SS`), so `drawn === ch` is no longer the right "unchanged" test for those. Comparing against the cased probe (`drawn === cased`) handles it.

`it("checks a heading through the case transform…")` asserts only `sourceField`, so it passes either way — worth extending it to pin the reported character as U+00B5 rather than U+039C.

Samhit21 (commented):
## PR Review (round 2): feat(export): pre-download validation pass

### Summary
Both round-1 findings are fixed correctly in `3d13ef7`. Links now carry `Links → link ${i + 1}` through the existing `path` mechanism, with a test pinning that a 3-link contact reports links 2 and 3; the detail message carries `U+XXXX`, with a VS16 test covering exactly the invisible case that motivated it. The code point derivation is right on astral characters too — `codePointAt(0)` on the iterator's output yields the full code point, so an emoji reports `U+1F600` rather than a lone surrogate.

One consequence of the second fix that the fix itself surfaces: the code point being printed is the drawn character's, not the authored one, so an uppercase heading names a code point the user cannot find in their own résumé. Same actionability principle, one level deeper.

### Findings
1. `render-findings.ts:249` — for a heading, the reported character and code point are post-`toUpperCase()`: `µ` (U+00B5) is named as `Μ` (U+039C). Verified. [Suggestion]

### Not raised, so they don't resurface as surprises
Appending `idx` to the React key makes it positional, which is what a content key normally avoids — harmless for a static list capped at five. And `add` skips falsy text while `forEach` still advances `i`, so an empty link string would yield "link 1, link 3" — but `links` holds scheme-stripped display strings, so an empty entry isn't a real shape.

I did not re-run the suite this round; verification is by reading plus a standalone check of the `µ` case.

### Verdict
**Action: COMMENT**

src/lib/pdf/render-findings.ts
s-annam: Fixed in `d9d5f90` — reworked `findGlyphFindings` to iterate the authored text and only run the case transform on the per-character coverage probe, so the finding names what the user typed (µ, U+00B5) instead of what the transform drew (Μ, U+039C). Also extended `render-findings.test.ts`'s case-transform test to pin the reported character/codepoint, per your note that it passed either way before.

Samhit21 (approved):
## PR Review (round 3): feat(export): pre-download validation pass

### Summary
The round-2 finding is fixed correctly in `d9d5f90`, including the wrinkle I flagged as needing a decision rather than inheriting: the loop iterates the authored `field.text`, the case transform is applied per-character to the coverage probe only, and the unchanged test is `drawn === cased` — so the 1→2 expansions (`ß` → `SS`) don't fall through as false findings. `isInvisibleDegradation` is passed `cased` too, which is the right argument since it is what was actually probed and case-folding never moves a character in or out of `\p{Zs}`.

The test extension is the part that makes this stick. It asserts in both directions — `toContain('"µ" (U+00B5)')` and `not.toContain("U+039C")` — so the previous version, which passed either way, can no longer.

### Verified by running it this time
- `npx vitest run src/lib/pdf/render-findings.test.ts src/components/features/ExportDialog.test.tsx` — **34 passed** (16 in `render-findings.test.ts`)
- `npm run typecheck` clean
- `npx eslint` clean on both changed files

### On the fallow bot's complexity finding
`findGlyphFindings` at cognitive complexity 16 against a threshold of 15, already flagged above by github-advanced-security. Not blocking, and by policy rather than by my discretion: `CLAUDE.md` states a fallow complexity/CRAP finding is a Nit/Secondary and never Blocking on its own, and the step is report-only inside `verify`. Worth recording *why* it moved, though — the per-character `cased` line is what pushed it over, and that line is the fix. If you want it back under the line, lifting the probe into a small named helper (`drawnFor(ch, field)`) would do it without touching behaviour. Entirely optional.

### Verdict
**Action: APPROVE**

Across three rounds the detection logic never needed a change — the two-pass design, the `pageOrdinal` observation, the `index > 0` split test and the full model-field coverage all held up under checking. Every finding was about what the finding *says*, and all three are now resolved: links are indexed, the code point is named, and the character named is the one the user typed.

PR #873, opened 2026-08-19
feat(critique): surface skills-ordering guidance in résumé review + rewrite lane (#544)

### LINKED ISSUE #544: [critique] Surface skills-ordering guidance in résumé review + rewrite lane
## Summary

The order in which a candidate lists skills matters — the most important/role-relevant skills should lead, so both human readers and ATS keyword scans see them first. Today nothing surfaces skill **ordering** as feedback. Add a skills-order call-out to the existing résumé review / rewrite functionality so the user is coached to reorder (and can apply a rewrite that does it), rather than silently accepting whatever order the résumé happens to use.

Distinct from the Find Jobs skills-truncation issue (that fixes the job-*query* surface). This issue is about **user-facing coaching in the critique/rewrite lane**.

## Motivation

Résumé skills sections are frequently ordered by habit, not by relevance. A leadership candidate may bury "Engineering Leadership" mid-list below incidental tools; an AI/ML candidate may list their AI/ML cluster last. Readers and keyword scanners weight earlier items more, so a poor order undersells the candidate. This is exactly the kind of concrete, actionable guidance the critique lane exists to give.

## Proposed change

Integrate into the existing review/rewrite surfaces (`src/lib/webllm/` critique + rewrite, `src/components/features/CritiquePanel.tsx`, and the section-rewrite path):

1. **Critique call-out** — when a skills section is detected, add a finding that flags likely-suboptimal ordering (e.g. a high-signal skill sitting far down the list relative to the candidate's title/target), phrased as coaching, not a hard error. Reuse the existing critique finding UI — no parallel surface.
2. **Rewrite affordance** — offer a skills reorder as one of the section-rewrite actions, so the user can apply a relevance-first ordering in place (with the existing confirm/undo pattern from the applied-rewrite work).

Ordering heuristic can start simple (title/target relevance, canonical-index weighting) and layer the on-device LLM where available, mirroring the critique/rewrite lane's existing semantic-with-heuristic-fallback shape.

## Acceptance criteria

- A skills section whose high-signal skills are buried triggers a review call-out in the critique lane.
- The user can apply a skills-reorder rewrite via the existing section-rewrite UI, with confirm + undo.
- Guidance is advisory (never blocks), and reuses existing critique/rewrite components (no new parallel surface — Reuse Gate).
- Covered by unit tests at the heuristic/module level.

## Notes

Complements the Find Jobs skills fix (relevance-rank + un-truncate the query): that improves the machine-side query; this improves the résumé the user actually ships. Both point at the same underlying idea — skill order carries signal.

s-annam:
## Summary

Surfaces skills-ordering guidance in the résumé review + rewrite lane. A high-signal
skill sitting far down the list relative to the candidate's title/target now
triggers a coaching call-out (never a hard error), and the user can apply a
relevance-first reorder via the existing section-rewrite confirm/undo pattern.

The finding is a plain heuristic (title/target word-overlap + canonical-index
tie-break), independent of the on-device LLM — it renders inside
`SkillTermGuidance`, the existing non-LLM-gated heuristic surface, so it's visible
on every parse rather than only after a user opts into downloading the on-device
model.

Apply is scoped to the flat (uncategorized) skills list; a categorized résumé shows
the coaching note with a pointer at the existing manual drag controls instead,
since `SkillsSection` renders order from category grouping, not the flat array.

## Verification

- 15 new unit tests: heuristic (`skills-order.ts`), the confirm/undo controller
  (`useSkillsReorder`), and the edit-hook reorder path (`useEditableParse`).
- `npm run typecheck` / `npm run lint` clean.

## Stack position

Layer 3 of 4. Base: #872. Depends on #872.

## Adversarial review

Two round-1 blocking findings, both fixed and verified in round 2:
- **`reorderSkills` resurrected deleted skills** — it overwrote the removed-skills
  set instead of merging with it, so a skill deleted before a reorder came back at
  the front of the list. Fixed to merge via `Set` union; regression test added.
- **The heuristic was invisible without WebGPU** — it originally mounted only
  inside the LLM-gated critique panel, contradicting its own "independent of the
  on-device LLM" docblock. Moved to `SkillTermGuidance` (single mount, no LLM gate);
  verified no double-mount and the panel's self-hide guard doesn't hide it.

Two nits remain, not blocking: the panel heading can promise a list it doesn't
render in the ordering-only edge case, and Apply-then-Undo leaves the résumé's
edit-dirty flag permanently set (autosave already fired on Apply, so low impact).

Closes #544

### DIFF
--- a/src/lib/heuristics/skills-order.ts
+++ b/src/lib/heuristics/skills-order.ts
@@ -1,0 +1,178 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * skills-order — heuristic skills-ordering coaching (issue #544).
+ *
+ * A résumé's Skills section is usually ordered by habit, not by relevance —
+ * readers and ATS keyword scans both weight earlier items more, so a
+ * high-signal skill buried mid-list undersells the candidate. This module is
+ * the "start simple" heuristic the issue asks for: title/target relevance
+ * (word-overlap against the candidate's derived titles) plus canonical-index
+ * weighting (a stable tie-break on the skill's ORIGINAL position, so the
+ * suggested order never reshuffles two equally-relevant skills for no
+ * reason). Layering the on-device LLM on top is explicitly deferred — see the
+ * issue's "Notes".
+ *
+ * Pure, dependency-free, no React/engine/I-O — mirrors `disagreement.ts` and
+ * `skills-categories.ts`. Consumed by `useSkillsReorder` (the confirm/undo
+ * controller) and rendered by `SkillsOrderFinding.tsx` inside
+ * `SkillTermGuidance` — the résumé lane's existing heuristic skills advisory,
+ * not a parallel surface, and not the WebGPU-gated critique panel (see that
+ * component's docblock for why the row is not hosted there).
+ */
+
+// ── Tokenizing + scoring ───────────────────────────────────────────────────
+
+/** Role-seniority qualifiers that would otherwise dominate every title's
+ *  token set without naming a discipline — dropping them keeps the overlap
+ *  test about WHAT the role does, not its level. */
+const STOPWORDS = new Set([
+  "senior",
+  "sr",
+  "junior",
+  "jr",
+  "staff",
+  "principal",
+  "lead",
+  "chief",
+  "head",
+  "vp",
+  "vice",
+  "president",
+  "director",
+  "manager",
+  "and",
+  "the",
+  "of",
+  "for",
+  "with",
+]);
+
+/** Lower-case word tokens, dropping seniority stopwords and anything shorter
+ *  than 2 chars (punctuation remnants). `+`/`#` survive the split so "C++" /
+ *  "C#" tokenize as themselves rather than the bare letter. */
+function tokenize(text: string): Set<string> {
+  const tokens = text
+    .toLowerCase()
+    .split(/[^a-z0-9+#]+/)
+    .filter((t) => t.length >= 2 && !STOPWORDS.has(t));
+  return new Set(tokens);
+}
+
+/** Longest shared prefix, e.g. "developer"/"development" -> 7 ("develop"). */
+function commonPrefixLength(a: string, b: string): number {
+  let i = 0;
+  while (i < a.length && i < b.length && a[i] === b[i]) i++;
+  return i;
+}
+
+/** Two tokens diverging by at most 4 chars on EITHER side of their shared
+ *  stem, requiring a stem of at least 4 chars — a cheap stand-in for
+ *  stemming. This is a common-prefix bound, not a "one is a prefix of the
+ *  other" bound: that narrower shape (tried first) matched
+ *  "engineer"/"engineering" but missed real stem pairs that diverge on BOTH
+ *  sides of the shared root, like "developer"/"development" (stem "develop",
+ *  +2/+4) and "analytics"/"analytical" (stem "analytic", +1/+2) — neither
+ *  token is a prefix of the other. It still rejects "java"/"javascript": the
+ *  stem is the whole 4-char "java", but "javascript" diverges by +6, over
+ *  the cap. Both tokens must clear the 4-char stem floor so short words
+ *  ("ai", "ml", "go") only match exactly, never by accidental prefix. */
+function looselyMatches(a: string, b: string): boolean {
+  if (a === b) return true;
+  if (a.length < 4 || b.length < 4) return false;
+  const stem = commonPrefixLength(a, b);
+  return stem >= 4 && a.length - stem <= 4 && b.length - stem <= 4;
+}
+
+/** Relevance of one skill to the target: count of skill tokens that match a
+ *  title token, exact matches counting double a loose (prefix) match — so
+ *  "Engineering Leadership" against "Engineering Manager" scores higher than
+ *  a skill that only loosely echoes the title. */
+function relevanceScore(skill: string, titleTokens: Set<string>): number {
+  let score = 0;
+  for (const token of tokenize(skill)) {
+    if (titleTokens.has(token)) {
+      score += 2;
+      continue;
+    }
+    for (const titleToken of titleTokens) {
+      if (looselyMatches(token, titleToken)) {
+        score += 1;
+        break;
+      }
+    }
+  }
+  return score;
+}
+
+// ── Public shape ─────────────────────────────────────────────────────────
+
+export interface SkillsOrderFinding {
+  /** High-signal skills (relative to the target) sitting outside the "front"
+   *  of the list, in their current document order. Coaching copy names these. */
+  buried: string[];
+  /** `skills` re-ordered relevance-desc, ties broken by original index (the
+   *  canonical-index weighting) — what an Apply action would write. Always a
+   *  permutation of the input `skills` array. */
+  suggestedOrder: string[];
+}
+
+/** Below this many skills, "ordering" barely matters — a short list is read
+ *  in full regardless of sequence, so no finding fires. */
+const MIN_SKILLS_FOR_FINDING = 5;
+
+/** The "front" of the list is the first 30% (never fewer than 3 skills) — a
+ *  high-signal skill inside that window is already well-placed. */
+const MIN_FRONT_WINDOW = 3;
+const FRONT_FRACTION = 0.3;
+
+/**
+ * Compute the skills-ordering coaching finding, or `undefined` when nothing
+ * is worth flagging: too few skills, no usable title/target signal, or no
+ * skill scores above zero against it, or every top-scoring skill already
+ * sits in the front window.
+ *
+ * `titles` is the candidate's derived target titles (`deriveTitles` from
+ * `job-search/query-builder.ts`), passed in by the caller rather than
+ * imported here — keeps this module dependency-free, matching the house
+ * style for pure heuristics (`skills-categories.ts`).
+ */
+export function computeSkillsOrderFinding(
+  skills: readonly string[],
+  titles: readonly string[],
+): SkillsOrderFinding | undefined {
+  if (skills.length < MIN_SKILLS_FOR_FINDING) return undefined;
+
+  const titleTokens = tokenize(titles.join(" "));
+  if (titleTokens.size === 0) return undefined;
+
+  const scored = skills.map((skill, index) => ({
+    skill,
+    index,
+    score: relevanceScore(skill, titleTokens),
+  }));
+
+  const maxScore = Math.max(...scored.map((s) => s.score));
+  if (maxScore <= 0) return undefined;
+
+  const frontWindow = Math.max(
+    MIN_FRONT_WINDOW,
+    Math.ceil(skills.length * FRONT_FRACTION),
+  );
+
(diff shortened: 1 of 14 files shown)


src/lib/heuristics/skills-order.ts
```diff
@@ -0,0 +1,163 @@
...
+}
+
+/** True when two tokens share at least a 4-char prefix — a cheap stand-in for
+ *  stemming that matches "engineer"/"engineering" or "lead"/"leadership"
+ *  without a real stemmer. Both tokens must clear the 4-char floor so short
+ *  words ("ai", "ml", "go") only match exactly, never by accidental prefix. */
+function looselyMatches(a: string, b: string): boolean {
+  if (a === b) return true;
+  if (a.length < 4 || b.length < 4) return false;
+  return a.startsWith(b.slice(0, 4)) || b.startsWith(a.slice(0, 4));
```
Samhit21: [Suggestion]: the 4-char prefix rule matches Java↔JavaScript, and the docblock's own example can't fire

Two things about this rule, found by running it rather than reading it.

**It can promote the wrong skill.** Sharing a 4-char prefix joins `java` and `javascript`:

```
title "Java Developer" -> tokens [java, developer]
  Java          score=2   (exact)
  JavaScript    score=1   (loose: "javascript".startsWith("java"))
  Spring Boot   score=0
```

Usually harmless — `Java` outscores it. But `buried` is gated on `score === maxScore`, and a loose match can only ever ADD score, so on a résumé that lists JavaScript and *not* Java against a Java title, the false match becomes the max and JavaScript is the skill this feature moves to the front of the user's résumé. That's the one outcome a coaching heuristic should not produce.

**The docblock's justification is unreachable.** It cites `lead`/`leadership` as a case this rule exists to catch, but `lead` is in `STOPWORDS`, so it's stripped from both sides before `looselyMatches` ever sees it:

```
title "Lead Engineer" -> tokens [engineer]        ("lead" dropped)
  Leadership   score=0
  Engineering  score=1
```

So half the stated motivation for the 4-char floor is untested. `engineer`/`engineering` does work and is a good example; `lead`/`leadership` should go or be replaced with one that fires.

A length-delta cap separates the two cases cleanly — `engineer`→`engineering` is +3, `java`→`javascript` is +6:

```ts
function looselyMatches(a: string, b: string): boolean {
  if (a === b) return true;
  if (a.length < 4 || b.length < 4) return false;
  const [short, long] = a.length <= b.length ? [a, b] : [b, a];
  return long.startsWith(short) && long.length - short.length <= 4;
}
```

Whatever rule you land on, a test pinning that `JavaScript` does not outrank on a Java title would keep it honest — `skills-order.test.ts` currently has no negative case for loose matching.

Samhit21 (commented):
## PR Review: feat(critique): surface skills-ordering guidance in résumé review + rewrite lane

### Summary
The heuristic is well-shaped for what it is — pure, dependency-free, stable tie-break on original index, and conservative gating (5-skill floor, top-score-only `buried`, front window). I paid particular attention to `reorderSkills`, since the round-1 blocking bug lived there, and the fix holds up under the interactions that would break it: a skill deleted before an Apply stays deleted (its key survives the `Set` union while `order` is a permutation of the current list), Undo re-runs the same path and preserves the deletion, and delete-after-Apply still works because `removeSkill` filters `added` as well as recording in `removed`. The `canApply` gate on a categorised résumé matches how `SkillsSection` actually renders, so the withheld button isn't arbitrary.

One finding, in the relevance matcher.

### Findings
1. `skills-order.ts:70` — the 4-char prefix rule scores `JavaScript` against a Java title, and since a loose match can only add score it can become `maxScore` and drive the reorder; separately, the docblock's `lead`/`leadership` example can't fire because `lead` is a stopword. Both verified by running the function. [Suggestion]

### PR-level: the stack claim doesn't match the branch
The body says "Layer 3 of 4. Base: #872. Depends on #872." As configured this PR targets `main`, carries a single commit, and contains neither #871's `docs/scoring.md` nor #872's `render-findings.ts` — it is standalone. I checked for a real coupling and found none (no file overlap with #872, no import of anything it introduces), so the base looks *correct* and the body looks stale. Worth fixing the body rather than the base: a reviewer told to read #872 first will go looking for a dependency that isn't there, and a merge-queue operator may sequence on it needlessly.

### Checked, not raised
`useSkillsReorder.ts` uses the `\u0000` join idiom at three sites and all three are the six-character escape rather than literal bytes, so it passes the #787 gate. And `suggestedOrder` re-sorting the whole list initially looked like a mismatch with the "move the relevant skills to the front" aria-label, but the tie-break on original index makes it a stable sort — unrelated skills keep their relative order, so the label is accurate.

### Verdict
**Action: COMMENT**

src/lib/heuristics/skills-order.ts
s-annam: Fixed in `c1329b6` — capped the loose prefix match by length delta (`long.length - short.length <= 4`), so `engineer`/`engineering` (+3) still matches but `java`/`javascript` (+6) no longer does. Also dropped the unreachable `lead`/`leadership` docblock example (stripped by STOPWORDS before `looselyMatches` sees it) and added a negative test pinning that JavaScript doesn't outrank on a Java title.

Samhit21 (dismissed):
## PR Review (round 2): feat(critique): surface skills-ordering guidance

### Summary
Fixed in `c1329b6`, and the negative test is the right one — asserting `undefined` rather than a score means it pins "no relevance signal at all", which is what actually protects against the wrong skill being promoted. The unreachable `lead`/`leadership` example is gone from the docblock too.

### Verified by running it
- `skills-order.test.ts` 8, `useSkillsReorder.test.tsx` 5, `useEditableParse.test.tsx` 50, `SkillTermGuidance.test.tsx` 13, `ResultDetail.test.tsx` 23 — **99 passed**
- `npm run typecheck` clean; `npx eslint` clean on the three changed source files

### One consequence worth recording, not blocking
Requiring the full shorter token as a prefix (rather than a shared 4-char prefix) narrows the rule in a second way beyond `java`/`javascript`. Running both versions side by side:

| pair | old | new |
|---|---|---|
| `engineer` / `engineering` | ✅ | ✅ |
| `design` / `designer` | ✅ | ✅ |
| `data` / `database` | ✅ | ✅ |
| `java` / `javascript` | ✅ | ❌ *(the fix)* |
| `developer` / `development` | ✅ | ❌ |
| `analytics` / `analytical` | ✅ | ❌ |
| `engineers` / `engineering` | ✅ | ❌ |

The last three are legitimate stem variants that the old rule caught: a "Software Developer" title no longer boosts a "Full-Stack Development" skill. I'm not treating it as a defect, because the two failure directions are not symmetric — over-matching promotes the *wrong* skill to the top of someone's résumé, while under-matching just misses a signal and leaves the order alone. For a feature that rewrites the user's document, that is the correct side to err on, and this was my suggestion so the tradeoff is mine as much as yours.

The docblock now says the rule matches `engineer`/`engineering` without noting what it stopped matching. A clause naming the tradeoff ("a stem that diverges before the end — `developer`/`development` — no longer matches; under-matching is the safe direction here") would save the next person rediscovering it. Cosmetic.

### Still open from round 1: the stack claim
The body still reads "Layer 3 of 4. Base: #872. Depends on #872." This PR targets `main`, carries one commit, and contains neither #871's nor #872's work — it is standalone, and I found no import or file overlap creating a real dependency. The base is right; the body is stale. Not blocking, but worth a two-word edit before the queue so nobody sequences on a dependency that isn't there.

### Verdict
**Action: APPROVE**

The heuristic is conservative in the ways that matter (5-skill floor, top-score-only `buried`, stable tie-break), and `reorderSkills` survives the three interactions that would reopen the round-1 bug — delete-before-Apply, Undo, and delete-after-Apply.

s-annam:
Addressed the round-2 "one consequence worth recording" note in `4aeb29e`.

The suggested fix (require one token to be a full prefix of the other, capped by length delta) correctly blocked java/javascript but also cost the three legitimate stem pairs the table called out — developer/development, analytics/analytical, engineers/engineering — since none of those are a full-prefix relationship; they diverge on both sides of a shared root.

Replaced it with a common-prefix-length bound: `looselyMatches` now finds the longest shared prefix and requires it to be ≥4 chars, then caps *each* token's divergence from that shared stem at 4 chars (rather than capping the difference between the two token lengths). Verified by hand against the full table from the review:

| pair | old (round 1) | round-2 fix | this fix |
|---|---|---|---|
| java / javascript | ✅ (bug) | ❌ | ❌ |
| engineer / engineering | ✅ | ✅ | ✅ |
| design / designer | ✅ | ✅ | ✅ |
| data / database | ✅ | ✅ | ✅ |
| developer / development | ✅ | ❌ | ✅ |
| analytics / analytical | ✅ | ❌ | ✅ |
| engineers / engineering | ✅ | ❌ | ✅ |

Added a test (`still matches stem pairs that diverge on both sides of a shared root`) pinning the three restored pairs, alongside the existing java/javascript negative test. `npm run typecheck`, `npm run test` (6010 passed) both green.

Samhit21 (approved):
## PR Review (round 3): feat(critique): surface skills-ordering guidance

Re-approving — my previous approval was auto-dismissed by the push to `3faf2d5`.

### Summary
This is a better rule than the one I suggested, and it resolves the tradeoff I'd written off as acceptable rather than fixable. Bounding the divergence on **both** sides of a shared stem, instead of requiring one token to be a prefix of the other, recovers exactly the pairs the narrower shape lost — because in every one of them neither token is a prefix of the other (`develop` + `er`/`ment`, `analytic` + `s`/`al`, `engineer` + `s`/`ing`). I ran all three rule versions side by side:

| pair | 4-char prefix (orig) | prefix-of (r2) | common-stem (r3) |
|---|---|---|---|
| `engineer` / `engineering` | ✅ | ✅ | ✅ |
| `design` / `designer` | ✅ | ✅ | ✅ |
| `data` / `database` | ✅ | ✅ | ✅ |
| `developer` / `development` | ✅ | ❌ | ✅ |
| `analytics` / `analytical` | ✅ | ❌ | ✅ |
| `engineers` / `engineering` | ✅ | ❌ | ✅ |
| `java` / `javascript` | ❌ *(the bug)* | ✅ | ✅ |
| `communication` / `community` | ❌ | ✅ | ✅ |

(✅ in the last two rows means "correctly rejected".) So r3 is strictly better than both predecessors: it keeps every true stem pair the original caught, and rejects `communication`/`community` as well, which the original matched and nobody had flagged.

The test is well-built too — placing the stem-pair skill at index 3 of a 5-skill list puts it past the front window, so `buried` is asserted rather than merely `defined`, and the comment records that the prefix-of shape was tried and why it was rejected. That is the part that keeps someone from "simplifying" it back in six months.

### Verified by running it
- `skills-order.test.ts`, `useSkillsReorder.test.tsx`, `SkillTermGuidance.test.tsx` — **27 passed**
- `npm run typecheck` clean; `npx eslint src/lib/heuristics/skills-order.ts` clean
- Re-derived the rule independently in a scratch script against all three versions (table above)

### Still open, non-blocking
The body still reads "Layer 3 of 4. Base: #872. Depends on #872." This PR targets `main`, carries one commit, and contains neither #871's nor #872's work — I checked for an import or file overlap creating a real dependency and found none. The base is correct; the body is stale. A two-word edit before the queue so nobody sequences on a dependency that isn't there.

### Verdict
**Action: APPROVE**

PR #879, opened 2026-08-21
feat(export): auto-fit body size and swap the embedded font to Liberation Sans

### LINKED ISSUE #878: Auto-fit export body size and swap the embedded font to Liberation Sans
## Problem

The Download-PDF export sets every résumé at a fixed 8.5pt in Poppins. Three things are wrong with
that, and they compound:

1. **The font is a display face doing text-face work.** Poppins is 1.14× the width of a normal text
   sans at the same point size, so the export has been paying for the typeface rather than for the
   content. It also lacks U+2192, which is why two rows sit in
   `corpus-roundtrip.known-failures.json`.
2. **One fixed size cannot serve both ends of the corpus.** 8.5pt is small enough for the densest
   résumé, which leaves every short one set a point and a half below what a typographer would
   choose. Measured across the 60-fixture corpus, **56 of 60 fit comfortably at 10pt**.
3. **Two visible layout defects** that a reader notices immediately (see below).

## Measurements

Font width and coverage, measured with `@pdf-lib/fontkit` over U+0020–U+2E7F:

| font | avg char width | chars/line @10pt | x-height | line box | coverage | Reg+Bold bytes |
|---|---|---|---|---|---|---|
| Poppins (current) | 0.5337 em | 101 | — | 1.400 em | 467 cps | 316 KB |
| **Liberation Sans** | **0.4435 em** | **121** | 0.528 em | **1.117 em** | 663 cps | **276 KB** |
| Carlito (evaluated, rejected) | 0.4101 em | 131 | 0.478 em | 1.221 em | 2098 cps | 1.31 MB |

Liberation Sans at 9.5pt is still narrower than Poppins at 8.5pt.

**Carlito was measured and rejected.** Its 7.5% narrower advance is optical, not real: its x-height
is 9.5% smaller, so Carlito at 10pt has the apparent size of Liberation at ~9.05pt. Normalized to
equal x-height it is 2.1% *wider* and its line box is 21% taller. Confirmed on the real renderer at
matched x-height (Liberation @10pt vs Carlito @11.05pt): identical wrap counts, but Carlito consumed
955pt of page against Liberation's 860pt — **11% more page for the same apparent size** — and six
fewer rows on page one. It also costs 4.7× the font bytes for Cyrillic/Greek/Vietnamese coverage a
résumé tool does not need, and would push the readability floor from 8pt to ~8.83pt.

### Height scales as size^1.76, not size²

The naive model says lines ∝ size and line-height ∝ size, so height ∝ size². Measured, the exponent
is **~1.76** — wrap counts are integers, so a three-line bullet stays three lines until it drops to
two. Absorbable overflow is therefore `(base / floor)^1.76`:

| base → floor | rescues up to |
|---|---|
| 10 → 8 | **1.48 pages** |
| 10 → 8.5 | 1.34 |
| 8.5 → 8 | **1.08** |

That last row is the point: from the *current* 8.5pt base, an 8pt floor absorbs only 8%. The floor is
not the interesting parameter — the base is. A ladder needs room to fall.

### Scale everything, not just the body

Measured both. Body-only scaling absorbs meaningfully worse (at k=0.85: 1.024 vs 0.959 pages). The
gaps, the section rule and the name block are ~140pt of fixed overhead — about 19% of the usable page
— so leaving them at full size costs roughly a third of the achievable shrink.

### A. A wrapped contact line strands a field

`Layout.fitToOneLine` sizes the contact line to hold one row, because the re-parse reads contact
fields off a **single header line** — a contact row that wraps loses whichever links land on the
continuation (`latex/multi-degree-coursework.pdf` drops its `linkedin_url` exactly this way). Two
bugs defeat it:

- **The fit misses by one ulp.** `size * CONTENT_WIDTH / width` is the exact fitting size in the
  reals, but re-multiplying it by the font's width-per-point in binary floating point can land a hair
  *above* `CONTENT_WIDTH`. The wrapper's `<=` test then fails and the row wraps anyway. Observed on a
  real résumé whose contact line is exactly 540.0pt at the fitted size.
- **There is no fallback past the readability floor.** Below `CONTACT_MIN_SIZE_RATIO` (0.8) the break
  is left to the word wrapper, which strands whichever fields overflow — one lone URL slug under a
  full row, its ` • ` separator left dangling above it.

Observed shape (reproduced with a synthetic persona of the same measured width):

```
jordan.bennett@example.com • (973) 555-0123 • Bellevue, WA • U.S. Citizen • example.com/in/… •
example.com/jbennett
```

Secondary consequence: `decorateFirstLine` only overlays a `drawText` call's **first** line, so a
link pushed onto a continuation loses its clickable annotation as well as its parsed field.

### B. A wrapped bullet's tail is indented past its own text

`BASE.bulletIndent` is a scale constant — 12pt at the reference size, **14.12pt at the 10pt rung** —
but the `"• "` marker it is supposed to match is only **6.28pt** wide at 10pt. Every wrapped bullet
draws its continuation ~7.8pt to the right of the text it continues. A constant also cannot follow a
font swap: the correct value is a property of the marker glyph, not of the layout.

### 1. Swap the embedded body font to Liberation Sans

Metric-compatible Arial clone under the SIL OFL — Helvetica's density with Latin-Extended coverage.
Vendor `LiberationSans-{Regular,Bold}.ttf` + the OFL text at `src/assets/fonts/`; delete the Poppins
files (Poppins has no other consumer — it was never a UI font here).

**"Just use the PDF built-in fonts" is disqualified.** The 14 standard fonts need no embedding and no
license, but they encode **WinAnsi only**: a candidate named Wiśniewska cannot be rendered at all —
`toWinAnsi()` degrades `ś` to `?` and `findExportGlyphLosses` then *refuses the download* rather than
hand back a mangled name (#664). For a résumé tool that is a correctness failure, not a cosmetic one.

### 2. Replace the fixed sizes with a proportional type scale + fit ladder

Express the whole scale at one size and multiply:

```ts
export const REFERENCE_BODY_PT = 8.5;
const BASE = { name: 18, headline: 11, contact: 9, section: 10, header: 9.5, sub: 8.5, body: 8.5,
               gapAfterContact: 7, /* … */ ruleHeight: 2 } as const;

export function makeTypeScale(bodyPt: number): TypeScale {
  const k = bodyPt / REFERENCE_BODY_PT;
  // every key × k; LINE_GAP is a RATIO and must NOT be scaled
}
```

`makeTypeScale(REFERENCE_BODY_PT)` must reproduce `BASE` exactly — that identity is what makes the
parameterization reviewable as a no-op at the reference rung. `MARGIN` stays 36 (a page property, not
a type property).

Then walk the ladder descending and keep the **largest rung achieving the fewest pages**:

```ts
const FIT_LADDER = [10, 9.5, 9, 8.5, 8] as const;
```

- **0.5pt rungs.** Finer costs renders without moving the break, because the curve is quantized.
- **8pt is a READABILITY floor**, not a layout artifact — Liberation Sans x-height is 4.22pt there.
- **The fallback is the TOP rung, not the floor.** Saving a page is worth type size whether the
  saving is 2→1 or 4→3; being smaller for its own sake never is. A résumé the floor cannot rescue
  should be readable *and* long, not cramped *and* long.
- **No separate dry-run measure path.** The pass that decides the size IS the pass that produces the
  bytes, so the decision can never be made against geometry the download does not have. Measured at
  ~36ms mean per résumé across the corpus, because 56/60 exit on rung 1.

`renderAtsResumePdf` returns `{ bytes, findings, bodyPt, pages }`; a `{ bodyPt }` option pins the size
and skips the pass.

### 3. Contact line: make the one-row promise real, then break at a field boundary

- Shave the fitted size by one part in a billion so the computed size actually draws one row.
- When even the floor cannot hold one row, split at a **field boundary** rather than letting the word
  wrapper choose. Only indices at or below the links boundary are considered, so the links move down
  **together** and are never split across rows. Among those, pick the index that lets the rows be
  drawn largest (scored on the *unfloored* required size — the floor makes every over-long candidate
  score identically, which is useless for ranking). Draw both rows at one size; a two-size header
  block reads as two different things.
- Draw each row with its own `linkSpans`, so a link on the second row keeps its annotation.

### 4. Bullet hanging indent = the marker's measured width

Drop `bulletIndent` from the scale; measure it off the font instead:

```ts
bulletIndent(size: number): number {
  return this.fonts.regular.widthOfTextAtSize(BULLET_MARKER, size);
}
```

This must stay clear of `isWrappedContinuation`'s `markerX + 2` threshold in
`src/lib/heuristics/entry-blocks.ts`, or the re-parse stops folding a wrapped tail back into its
bullet. It does at every rung: the marker sits at `MARGIN` and its own width is the smallest indent
that puts the tail past it (36 + 6.28 = 42.28 vs a 38 threshold at the top rung; 36 + 5.02 = 41.02 at
the floor).

## Files

| File | Change |
|---|---|
| `src/lib/pdf/render-ats-pdf.ts` | font imports, `BASE`/`makeTypeScale`/`TypeScale`, `FIT_LADDER`, `fitToPage`, `Layout.fitToOneLine` + `oneLineSize` + `fitsOneLine` + `bulletIndent`, `contactSplitIndex`, contact draw block |
| `src/assets/fonts/` | add `LiberationSans-{Regular,Bold}.ttf` + `LiberationSans-OFL.txt`; delete the three Poppins files |
| `src/test-setup.ts` | the font-serving shim's regex + path |
| `src/lib/pdf/export-layout-contract.test.ts` | pin `{ bodyPt: REFERENCE_BODY_PT }` at every render site |
| `src/lib/pdf/render-findings.test.ts` | pin the page-break probe; swap glyphs Liberation now covers |
| `src/lib/pdf/render-ats-pdf.fonts.test.ts` | `→`/`µ` are covered now — use genuinely uncovered glyphs |
| `src/lib/heuristics/corpus-roundtrip.known-failures.json` | both U+2192 rows retire; `baselines` → `{}` |
| 3 fixture `*.expected.json` sidecars | re-bake (three booleans, all improvements) |

## A trap this change sets, and how it is handled

**Content-dependent geometry turns pagination tests vacuous, not red.** Every keep-with-next contract
in `export-layout-contract.test.ts` builds a deliberately oversized model, renders, and asserts *where*
the page broke. Once the engine's job becomes "make this fit on one page", those fixtures get shrunk
until **there is no page break left to make a claim about** — the assertion is trivially satisfied, or
it fails with a confusing off-by-one that invites recalibrating the fixture instead of noticing the
premise died.

**Pin, do not recalibrate.** Every pagination contract renders at `{ bodyPt: REFERENCE_BODY_PT }`. The
rung is arbitrary; holding it still is the point. Re-tuning a fixture until it goes green encodes the
new behaviour into a test written to assert something else, and the original contract quietly stops
being tested.

## Acceptance criteria

- [ ] `makeTypeScale(REFERENCE_BODY_PT)` reproduces `BASE` exactly, and every key scales by the same
      factor at any other size (`LINE_GAP` excluded — it is a ratio).
- [ ] A short résumé renders at 10pt, not at a fixed small size.
- [ ] A résumé that spills at 10pt but fits lower down is rescued, and lands on one page.
- [ ] The chosen rung is the **largest** achieving the fewest pages — asserted against the ladder
      itself, not a hand-computed answer, so it holds for a résumé too long to reach one page.
- [ ] `renderAtsResumePdf(model, { bodyPt })` honours the pin and skips the pass.
- [ ] **Contact, swept across a range of lengths** (not one fixture — the length that shipped the
      orphan was unguessable): at most two rows, no row ends on a separator, links never split across
      rows. Sweep must be non-vacuous — at least one length exercises the shrink path.
- [ ] At the exact float boundary, a line the shrink can still hold draws **one** row, not two.
      Fails without the shave; verified by reverting it.
- [ ] Past the floor, the break is at a field boundary, all links on one row, both rows at one size.
- [ ] A wrapped bullet's tail starts at `MARGIN + markerWidth`, under the first line's text.
- [ ] Round-trip corpus green with `baselines: {}` — no new KNOWN_FAILURES, both U+2192 rows retired.
- [ ] The #436 experience-label pin survives at **every** rung 8–10 (it previously held only at
      `SIZE_BODY <= 8.5`, because the one-line role header stopped wrapping).

## Out of scope

Deliberately not in this change — each alters visible output beyond the swap and the ladder:

- The size hierarchy: `BASE.contact` is 1.06× body where both reference résumés set contact *smaller*
  than body (~0.83×), and `ruleHeight` is 2pt against their sub-point hairline.
- `headerLineOpts` hardcodes `size: t.header` even when `entry.headerBold === false`, so the skills
  list draws 1pt larger than its neighbouring bullets.
- **Parser-side hardening to read contact fields across a wrapped block.** The export currently
  carries that invariant alone, and `fitToOneLine` cannot hold it for an arbitrarily long contact
  line — the field-boundary split narrows the gap but does not close it.

s-annam:
## Summary

The export set every résumé at a fixed 8.5pt in Poppins — a geometric display face measured at 1.14× the width of a normal text sans, so the shrinking done to fit a page was paying for the typeface rather than the content. Swaps to **Liberation Sans** (metric-compatible Arial clone, SIL OFL, 663 Latin-Extended code points vs Poppins' 467) and replaces the fixed size with a **10 / 9.5 / 9 / 8.5 / 8 fit ladder** that keeps the largest rung achieving the fewest pages. Across the 60-fixture corpus **56 now draw at 10pt** — all of them were previously set a point and a half smaller than needed.

The new freedom exposed two layout defects, both fixed here: a contact line that wrapped and silently dropped a parsed field, and a wrapped bullet whose tail was indented ~7.8pt past its own text.

Closes #878

## Review focus

- `src/lib/pdf/render-ats-pdf.ts` — `makeTypeScale` multiplies one `BASE` object by `bodyPt / REFERENCE_BODY_PT`. Is `LINE_GAP` the only value that must *not* scale, or did something else sneak into `BASE` that is a ratio rather than a point value?
- `src/lib/pdf/render-ats-pdf.ts` — `contactSplitIndex` scores candidates on the **unfloored** `oneLineSize`, not `fitToOneLine`. Does anything else in the contact path rank candidates through the floored value, where every over-long option scores identically?
- `Layout.bulletIndent` — the tail now starts at `MARGIN + markerWidth` (42.28pt at the top rung, 41.02pt at the floor) against `isWrappedContinuation`'s `markerX + 2` = 38pt. Is 3pt of headroom at the floor enough, and is `markerX` really `MARGIN` on every path that reaches it?
- `src/lib/pdf/export-layout-contract.test.ts` — every render site now pins `{ bodyPt: REFERENCE_BODY_PT }`. Did any pagination assertion get left unpinned, where the fit pass would shrink the fixture until there is no break left to assert?

## Test plan

- [x] `npm run test` green — 6131 passed, 10 skipped
- [x] `npm run verify` exits 0 (ran in the pre-push hook)
- [x] Round-trip corpus green with `baselines: {}` — both U+2192 KNOWN_FAILURES rows retired
- [x] Contact one-row fit and the field-boundary split verified fail-before (reverting `FIT_ONE_LINE_SHAVE` reddens the boundary test)
- [x] Manually verified in `npm run dev` — drop a PDF, download the export, check the contact row and a wrapped bullet's tail


src/lib/pdf/render-ats-pdf.ts
```diff
@@ -1596,31 +1794,63 @@ export async function renderAtsResumePdf(
...
+      const cut = contactSplitIndex(layout, contactParts, model.contact.links.length);
+      const top = contactParts.slice(0, cut).join(" • ");
+      const bottom = contactParts.slice(cut).join(" • ");
+      // One size for both rows — two rows of a single header block set at
+      // visibly different sizes read as two different things.
+      const rowSize = Math.min(
+        layout.fitToOneLine(top, layout.t.contact),
+        layout.fitToOneLine(bottom, layout.t.contact),
+      );
+      layout.drawText(top, { size: rowSize, color: muted, linkSpans: emailSpans });
```
s-annam: **Blocking** — `rowSize` is trusted without verifying either `top` or `bottom` actually fits one line at it. `fitToOneLine` floors at `CONTACT_MIN_SIZE_RATIO`, so when the true fit is below the floor this still draws at the floor and `drawText` silently word-wraps — reproduced with a long location + work-authorization statement: 3 rows, with the overflow word ("Region") stranded on its own line with no separator. Directly violates AC #878's "at most two rows" guarantee. See the review body for a fix suggestion (verify with `layout.fitsOneLine` before trusting the split).

src/lib/pdf/render-ats-pdf.ts
```diff
@@ -1561,29 +1759,29 @@ export async function renderAtsResumePdf(
...
   // not a URL, so it receives no clickable overlay and no scheme-stripping.
-  const contactParts = [
+  const identityParts = [
     model.contact.email,
     model.contact.phone,
     model.contact.location,
     model.contact.workAuthorization,
-    ...model.contact.links,
   ].filter((p): p is string => Boolean(p));
+  const contactParts = [...identityParts, ...model.contact.links];
```
s-annam: **Secondary** — the old code ran the whole array (identity fields + links) through `.filter(Boolean)`; here only `identityParts` is filtered, and `model.contact.links` is spread unfiltered. `formatLinkDisplay("https://")` → `""`, and `addLink` in `ats-resume-model.ts` gates on the raw URL being truthy, not the formatted result — a degenerate parsed URL can now produce a dangling `"• "` separator instead of being silently dropped as before.

src/lib/pdf/render-ats-pdf.ts
```diff
@@ -91,14 +91,14 @@ import {
...
-import poppinsBoldUrl from "../../assets/fonts/Poppins-Bold.ttf?url";
+import bodyFontRegularUrl from "../../assets/fonts/LiberationSans-Regular.ttf?url";
+import bodyFontBoldUrl from "../../assets/fonts/LiberationSans-Bold.ttf?url";
 
 // ── Page geometry (points) ────────────────────────────────────────────────────
 
 const PAGE_WIDTH = 612; // US Letter
 const PAGE_HEIGHT = 792;
-const MARGIN = 54;
+const MARGIN = 36;
```
s-annam, line 103: **Secondary** — `MARGIN` 54→36 (and `LINE_GAP` 1.25→1.15 at line 160) are real, page-density-affecting changes that aren't mentioned in issue #878's design section or this PR's body. `MARGIN` reads as intentional (the issue's own "540.0pt" contact-line example only works out at MARGIN=36), but neither is called out explicitly — worth a one-line note for the historical record.

src/test-setup.ts
```diff
@@ -34,11 +34,59 @@
...
+  const shim = (async (input: RequestInfo | URL, init?: RequestInit) => {
+    // `String()` covers all three `RequestInfo | URL` shapes: a string is itself,
+    // a URL stringifies to its href, and a Request is unwrapped to `.url` first.
+    const url = String(input instanceof Request ? input.url : input);
+    const match = /LiberationSans-(Regular|Bold)\.ttf/.exec(url);
+    if (!match) return previous(input, init);
+    const path = fileURLToPath(
+      new URL(`./assets/fonts/LiberationSans-${match[1]}.ttf`, import.meta.url),
+    );
+    return new Response(new Uint8Array(readFileSync(path)), { status: 200 });
```
s-annam, line 89: **Secondary** — under `@vitest-environment jsdom`, `fileURLToPath(new URL('./assets/fonts/...', import.meta.url))` resolves to `/src/assets/fonts/LiberationSans-Regular.ttf` (loses the project-root prefix — confirmed by reproduction; jsdom's `URL` mishandles multi-segment `file://` relative resolution). `readFileSync` then throws `ENOENT`, caught upstream, and the render silently falls back to Helvetica — exactly what this shim exists to prevent. Dormant today (no jsdom+renderer test asserts geometry yet) but a live trap for the next one.

src/lib/pdf/render-ats-pdf.flush-right-links.test.ts
```diff
@@ -13,7 +13,7 @@ import { renderAtsResumePdf } from "./render-ats-pdf.ts";
 import type { AtsResumeModel } from "./ats-resume-model.ts";
 
 const PAGE_WIDTH = 612;
-const MARGIN = 54;
+const MARGIN = 32;
```
s-annam: **Secondary** — this local `MARGIN` wasn't updated to match production's new value (36); `RIGHT_EDGE`'s `// 558` comment is also stale (612-32=580, not 558 — that math was only true for the old MARGIN=54). Passes only due to a `toBeLessThan(6)` tolerance.

src/lib/pdf/render-ats-pdf.ts
```diff
@@ -123,32 +123,96 @@ const USABLE_PAGE_HEIGHT = PAGE_HEIGHT - MARGIN - CONTENT_BOTTOM;
...
+ *
+ * The rungs are 0.5pt apart because the page height a résumé occupies is
+ * quantized by wrap counts — a three-line bullet stays three lines until it
+ * drops to two — so a finer ladder costs renders without moving the break.
+ *
+ * The floor is a READABILITY floor, not a layout one: at 8pt Liberation Sans
+ * the x-height is 4.22pt, which is about as small as résumé body text can be
+ * set and still survive a recruiter skim on screen. When even the floor does
+ * not reach one page the résumé genuinely needs two, and {@link fitToPage}
+ * ships the TOP rung rather than the floor — see its docblock.
```
s-annam: **Secondary** — this claim ("ships the TOP rung rather than the floor" when nothing fits in one page) is false in general: the algorithm ships whichever rung achieves the fewest pages, tie-broken to the largest — reproduced a fixture that's 3 pages at 10pt but 2 pages at 9.5/9/8.5/8pt, shipping at 9.5pt, not 10pt. `fitToPage`'s own docblock a few lines below states the correct, narrower rule ("fallback whenever shrinking buys nothing at all").

src/lib/heuristics/multi-experience-roundtrip.test.ts
```diff
@@ -105,16 +106,26 @@ describe("#311 multiple experience sections — parse + round-trip", { timeout:
...
+    // 9.5, 10): 2 labels survive at all five. That is a wider margin than it was
+    // under the Poppins this engine used to embed, where the same fixture held 2
+    // only at <= 8.5pt and flattened to 0 at >= 9 — Liberation Sans is narrow
+    // enough that the header no longer wraps at any size we ship.
+    //
+    // The assertion is still load-bearing on that header not wrapping, so a longer
+    // company name would put it back at 0 even now. Until #436's title/company
+    // disambiguation lands, a change here means the export geometry moved, not
+    // that the parser regressed.
+    expect(distinctLabels(parse3.canonical.fields.experience ?? []).length).toBe(2);
```
s-annam, line 129: **Nit** — "measured at every rung... 2 labels survive at all five" is asserted only in this comment; the test itself renders via unpinned `fitToPage` (one rung, whichever fits), not a sweep. Independently verified the claim is currently true at all 5 rungs, so this is a coverage gap, not a wrong claim — worth sweeping `[8, 8.5, 9, 9.5, 10]` explicitly, mirroring the pattern already used elsewhere in this PR.

s-annam (commented):
One Blocking finding — a reproducible AC violation in the contact-line field-boundary split — otherwise this is an exceptionally well-documented, well-tested change (6131/6131 tests green, typecheck/lint/build/fixture-PII all clean, and every fail-before claim in the PR's own test plan independently reproduced).

**Verdict rule applied:** ≥1 Blocking finding → `REQUEST_CHANGES`. No auto-fix was attempted (Step 5.5 only applies at 0 Blockers).

> **Note:** this verdict is `REQUEST_CHANGES` per the rule above (≥1 Blocking finding), but GitHub's API refuses `REQUEST_CHANGES`/`APPROVE` from a PR's own author (`"Review Can not request changes on your own pull request"`), so this is posted as a `COMMENT` review instead. Treat it as REQUEST_CHANGES: the Blocking finding above should be fixed before merge.

## Blocking

**Contact-line field-boundary split can silently produce 3 rows, stranding a field with no separator — the exact defect this PR sets out to eliminate.** `src/lib/pdf/render-ats-pdf.ts:1826-1836` computes `rowSize` as `Math.min(fitToOneLine(top), fitToOneLine(bottom))` and draws both rows at it *without ever verifying either row actually fits on one line at that size*. `fitToOneLine` (`Layout.fitToOneLine`, line ~561) is **floored** at `CONTACT_MIN_SIZE_RATIO * size` (0.8×) — when the true one-line-fit size is *below* the floor, it still returns the floor value as if the text fit.

Reproduced directly (contact: email + phone + a moderately long location + a "#792 work authorization" statement + 2 links):
```
jordan.bennett@example.com • (973) 555-0123 • A Truly Extraordinarily Long Municipality Name In The Greater Metropolitan Statistical Area
Region                                                          ← orphaned, no separator, no field boundary
Authorized to work without sponsorship in the United States of America • example.com/in/jordan-bennett • example.com/jbennett
```
3 rows, directly violating AC #878's explicit "at most two rows" guarantee. `top`'s true required size was below the floor, so `drawText` silently word-wrapped it — the exact "stranded field, no separator, orphaned continuation" defect this whole split mechanism exists to prevent (#425/#664), with zero runtime signal. `render-ats-pdf.type-scale.test.ts`'s sweep (`"never strands a field, at any contact length"`) only pads **link** length, never identity-field (location/workAuthorization) length, so this path is untested. Not a pathological input — a longer visa-status statement or municipality name is realistic for this product's stated audience.

**Fix:** verify the split actually worked before trusting it, e.g.:
```ts
const rowSize = Math.min(
  layout.fitToOneLine(top, layout.t.contact),
  layout.fitToOneLine(bottom, layout.t.contact),
);
if (!layout.fitsOneLine(top, rowSize) || !layout.fitsOneLine(bottom, rowSize)) {
  // Split didn't help — fall back to the single-line (still-wrapping) draw
  // rather than pretend the split succeeded.
  layout.drawText(contactLine, { size: fitted, color: muted, linkSpans: [...emailSpans, ...linkSpans] });
} else {
  layout.drawText(top, { size: rowSize, color: muted, linkSpans: emailSpans });
  layout.drawText(bottom, { size: rowSize, color: muted, linkSpans });
}
```

## Secondary

- **`contactParts` dropped the `.filter(Boolean)` safety net for links.** `src/lib/pdf/render-ats-pdf.ts:1778-1784` — old code ran the *whole* array (identity fields + links) through `.filter(Boolean)`; the new code only filters `identityParts`, then spreads `model.contact.links` unfiltered. `formatLinkDisplay("https://")` → `""` (strips scheme/www/trailing-slash to nothing) — `ats-resume-model.ts`'s `addLink` gates on the raw URL being truthy, not on the formatted result being non-empty. A degenerate parsed URL reaching this now produces a dangling `"• "` separator and a zero-width `linkSpans` entry instead of being silently dropped as before. Edge-case trigger, but a real regression of a previously-present defense.

- **`MARGIN` (54→36) and `LINE_GAP` (1.25→1.15) are undocumented behavior changes outside the issue's stated design.** `src/lib/pdf/render-ats-pdf.ts:101,160`. Issue #878's design section only ever states `MARGIN` "stays 36" (never says it's *changing* from 54) and never mentions `LINE_GAP` at all — both are real, page-density-affecting decisions (`MARGIN` is implicitly backed by the issue's own "540.0pt" contact-line example, so it reads as intentional; `LINE_GAP`'s ~8% tighter leading has no measurement or mention anywhere). The PR body doesn't call either out either. Worth a one-line justification in the PR/issue for the historical record — not asking to revert.

- **`test-setup.ts`'s new font-serving shim silently breaks under `@vitest-environment jsdom`.** `src/test-setup.ts:78-81` — `fileURLToPath(new URL('./assets/fonts/...', import.meta.url))` resolves to `/src/assets/fonts/LiberationSans-Regular.ttf` (losing the project-root prefix) when the calling test file declares `@vitest-environment jsdom`, confirmed by direct reproduction (jsdom's `URL` implementation mishandles multi-segment `file://` relative resolution). The `readFileSync` then throws `ENOENT`, caught upstream in `loadFonts`, and the render silently falls back to Helvetica — exactly the outcome this shim's own docblock says it exists to eliminate ("every layout assertion... was calibrated against a font that was not the shipped one... Restoring the real bytes here is what makes an export test's claim true of the actual download"). Currently dormant — `useDownloadPdf.test.tsx` is the only jsdom+renderer test today and its assertions don't depend on font geometry — but it's a live trap for the next jsdom test that does.

- **`render-ats-pdf.flush-right-links.test.ts`'s local `MARGIN` wasn't updated with production's.** Line 16: `const MARGIN = 32;` (unrelated to production's new `36`) with a stale `// 558` comment on `RIGHT_EDGE` (612-32=580, not 558 — that was true only for the old MARGIN=54). Passes only because of a `toBeLessThan(6)` tolerance; a real few-point regression in this geometry would go undetected.

- **`FIT_LADDER`'s docblock overstates `fitToPage`'s fallback behavior.** `src/lib/pdf/render-ats-pdf.ts:194` — "When even the floor does not reach one page... `fitToPage` ships the TOP rung rather than the floor" is false in general; the algorithm (correctly, per its own test) ships whichever rung achieves the *fewest* pages, tie-broken to the largest — that's usually the floor for a genuinely 2+-page résumé, not the top rung. Reproduced: a fixture that's 3 pages at 10pt but 2 pages at 9.5/9/8.5/8pt ships at **9.5pt**, not 10pt. `fitToPage`'s own docblock (a few lines below) states the correct, narrower rule ("top rung is the fallback whenever shrinking buys nothing *at all*"). A future maintainer trusting the `FIT_LADDER` comment over the tested behavior could "fix" this into an actual regression.

## Nits

- Module docblock still says `~54pt margins` (`src/lib/pdf/render-ats-pdf.ts:16`, unchanged context line, not part of this diff's `+` lines) though `MARGIN` is now 36.
- `multi-experience-roundtrip.test.ts:129`'s "measured at every rung 8, 8.5, 9, 9.5, 10: 2 labels survive at all five" is asserted only in a comment — the test itself renders via unpinned `fitToPage` (a single, whichever-rung-fits render), not a sweep. Independently verified the underlying claim **is** true at all 5 rungs (2 labels survive at each), so this is a coverage/rigor gap, not a wrong claim — worth a `for (const pt of [8, 8.5, 9, 9.5, 10])` sweep mirroring the pattern already used elsewhere in this PR (e.g. the contact-length sweep).
- `corpus-roundtrip` fixtures now pay for up to 5 full render passes per fixture (unpinned `renderAtsResumePdf` call in `roundtrip-hop.ts`) inside a `{ timeout: 20000 }` budget sized for one render. Currently comfortably under budget (full suite: 27.95s, no timeouts) — flagging for awareness as the corpus grows, not asking for a change now.

## Gates run

- **Issue #878 AC checklist**: 9 of 10 checkable ACs verified met by direct reproduction (type-scale identity, top-rung-for-short-résumé, rescue-when-spilling, largest-rung-tie-break, `bodyPt` pin/skip, contact sweep non-vacuous, float-boundary fail-before reproduced myself by reverting `FIT_ONE_LINE_SHAVE`, bullet hanging-indent under text, round-trip corpus green with `baselines: {}`, `#436` pin holds at every rung — verified independently). **1 unmet in substance**: "at most two rows" for the contact block (Blocking finding above) — the swept AC test passes because it only varies link length, but the underlying contract breaks on identity-field length. `npm run verify`-equivalent (typecheck + lint + build + fixtures + full test suite) all run locally and green — not literally re-run via the `verify` script name, but every step it chains was executed directly.
- **3f description accuracy**: PR body's Summary/Test plan are accurate and every checked box was independently reproduced (6131/6131 tests, the boundary fail-before-claim, both U+2192 KNOWN_FAILURES rows retired — plus a third, `extended-latin-name-roundtrip`, retired as a side effect of the new font-serving test shim, a bonus not claimed). `Review focus` questions 2–4 all check out correctly against the code (verified independently: `contactSplitIndex` does rank on unfloored `oneLineSize`, `bulletIndent`'s headroom math holds, all 3 pagination-contract render sites are pinned). Question 1 (LINE_GAP) is answered by the Secondary finding above — nothing else in `BASE` was left unscaled, but `MARGIN`/`LINE_GAP` themselves are undocumented value changes, not undocumented scaling bugs.
- **Fixture PII (3a)**: not applicable — no fixture PDFs/images touched, only sidecar JSON re-bakes matching the issue's stated 3-boolean re-bake exactly.
- **Design-system/reuse (3b)**: not applicable — no `src/components/**` touched.
- **Style tokens (3c)**: clean — no hardcoded colors/palette classes (grep hits were all `#nnn` issue references, not hex colors).
- **Fallow/dead-code (3d)**: 0 dead files/exports. One new complexity flag on `renderAtsResumePdfAtSize` (199 lines, cyclomatic 16) — report-only per this repo's `verify` config, has multiple in-repo consumers and heavy test coverage across 8+ files, not flagging as a finding.
- **Command-level bugs (3e)**: only `scripts/fixture-derive.ts` touched, comment-only change, no command/flag surface affected.

Head SHA reviewed: `ec0f0ca621bc9faa9a52a77bc193fde22177fe11`. 👀 was posted at the start of this review.

---

src/lib/pdf/render-ats-pdf.ts
```diff
@@ -1515,21 +1660,73 @@
...
  * feeds a layout decision, so `bytes` for a clean résumé is byte-for-byte what
  * this function produced before findings existed — the property the round-trip
  * corpus pins.
+ *
+ * Private: the size is chosen by {@link fitToPage} or pinned through
+ * `renderAtsResumePdf`'s `bodyPt` option, so nothing outside this module needs
+ * to name a rung directly.
  */
-export async function renderAtsResumePdf(
+async function renderAtsResumePdfAtSize(
```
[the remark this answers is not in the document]
s-annam: Fixed in `7f9f45d4405112fee018fb957446921391493439` — extracted the contact-line decision tree (name+headline+split/fit logic) out of `renderAtsResumePdfAtSize` into its own `drawContactLine` helper. Cognitive complexity for `renderAtsResumePdfAtSize` drops from 22 to below the reported threshold (fallow no longer flags it).

src/lib/pdf/render-ats-pdf.ts
s-annam: Fixed in `7f9f45d4405112fee018fb957446921391493439` — the contact-line split branch now verifies with `layout.fitsOneLine(top, rowSize) && layout.fitsOneLine(bottom, rowSize)` before trusting the split, falling back to the single wrapping line otherwise. Reproduced the original 3-row overflow with your example before the fix, and confirmed the fallback engages (no more stranded field) after it. Also extracted this whole block into a `drawContactLine` helper (see the fallow-complexity thread).

src/lib/pdf/render-ats-pdf.ts
s-annam: Fixed in `7f9f45d4405112fee018fb957446921391493439` — `contactParts` now filters `model.contact.links` for `Boolean` before spreading, restoring the safety net the old whole-array `.filter(Boolean)` provided for the visible text.

src/lib/pdf/render-ats-pdf.ts
s-annam: Fixed in `7f9f45d4405112fee018fb957446921391493439` — added a one-line rationale comment above both `MARGIN` and `LINE_GAP` (#878: margin recovers width for the fit ladder's larger rungs; line-gap was tuned against the old fixed 8.5pt body and tightened alongside it). Also updated the module docblock's stale "~54pt margins" to 36pt.

src/test-setup.ts
s-annam: Fixed in `7f9f45d4405112fee018fb957446921391493439` — `test-setup.ts` now resolves the fonts directory once via `path.join(path.dirname(fileURLToPath(import.meta.url)), "assets/fonts")` instead of `new URL(relative, import.meta.url)`, sidestepping jsdom's relative-URL resolution bug entirely. Confirmed the shim now serves real bytes under `@vitest-environment jsdom` — this actually surfaced a real, now-fixed test bug in `useDownloadPdf.test.tsx`'s glyph-loss-refusal describe block, which had been silently relying on the same jsdom bug to force its Helvetica-fallback path; it now stubs `fetch` explicitly and resets the module so it isn't fooled by the module-level font-bytes memo.

src/lib/pdf/render-ats-pdf.flush-right-links.test.ts
s-annam: Fixed in `7f9f45d4405112fee018fb957446921391493439` — updated this test's local `MARGIN` to 36 (matching production) and fixed the stale `RIGHT_EDGE` comment to 576.

src/lib/pdf/render-ats-pdf.ts
s-annam: Fixed in `7f9f45d4405112fee018fb957446921391493439` — reworded the `FIT_LADDER` docblock to state the actual rule (fewest pages, tie-broken to the largest rung) instead of the false "ships the TOP rung" claim, and pointed to `fitToPage`'s own docblock for the precise wording.

src/lib/heuristics/multi-experience-roundtrip.test.ts
s-annam: Fixed in `7f9f45d4405112fee018fb957446921391493439` — added an explicit sweep test (`holds the 2 → 2 grouping at every fit-ladder rung (8, 8.5, 9, 9.5, 10)`) that pins `bodyPt` at each rung and re-parses, rather than leaving the claim comment-only.

Samhit21 (commented):
## PR Review: feat(export): auto-fit body size and swap the embedded font to Liberation Sans

### Summary
Reviewed against `5062f7b0`, after the eight findings from the earlier round were fixed. I spent this pass on the four Review-focus questions, since nobody had answered them, and verified each rather than reasoning about it. All four hold; one turned up a loose end, and a separate sweep of the font swap turned up a broken script.

**Q1 — is `LINE_GAP` the only value that must not scale?** Yes. Every one of the fifteen `BASE` entries is a point value (sizes, gaps, `ruleHeight`), so multiplying the object by `bodyPt / REFERENCE_BODY_PT` is uniformly correct, and `LINE_GAP` is the sole ratio and correctly lives outside `BASE`.

**Q2 — does anything else in the contact path rank through the floored value?** No. `contactSplitIndex` scores on unfloored `oneLineSize`, which is right — through `fitToOneLine` every over-long candidate would tie at the floor and `best` would collapse to the boundary. The only other consumer, `rowSize`, uses `fitToOneLine` solely to pick the drawn size, and the `fitsOneLine` guard added for the Blocking finding is what catches the floor case. No ranking passes through the floor.

**Q3 — is 3pt of headroom enough, and is `markerX` really `MARGIN`?** Yes, and the safety is structural rather than marginal. `bulletMarkerX` takes the minimum x over `isBulletLine` lines, and our export draws the marker at `MARGIN`, so `markerX` is 36 on the round-trip path. The contract `MARGIN + width("• ") > markerX + 2` therefore reduces to **`width("• ") > 2pt`**. Measured against the shipped `LiberationSans-Regular.ttf` with fontkit:

| bodyPt | width("• ") | tail x | headroom over 38 |
|---|---|---|---|
| 10 | 6.279 | 42.28 | 4.28 |
| 9.5 | 5.965 | 41.97 | 3.97 |
| 9 | 5.651 | 41.65 | 3.65 |
| 8.5 | 5.337 | 41.34 | 3.34 |
| 8 | 5.023 | 41.02 | 3.02 |

Your figures are exact. The marker would have to fall under 2pt to break the contract, which needs a body size around **3.2pt** — six rungs below the floor. The ladder cannot threaten it.

**Q4 — did any pagination assertion get left unpinned?** One did: `render-ats-pdf.test.ts:77`, `const { bytes } = await renderAtsResumePdf(bigModel);` feeding `expect(doc.numPages).toBeGreaterThan(1)` eight lines down. It is the only assertion left in the repo that reads a page count off an unpinned render. It passes today and will keep passing — 40 entries x (header + subline + two long bullets) cannot reach one page at any rung, the 8pt floor included — so this is a consistency point, not a live failure. But the test is named "a long resume spans more than one page," and what it now asserts is "the fit ladder cannot get this fixture onto one page," which is a different statement that moves whenever the ladder does. Adding `PINNED` (importing `REFERENCE_BODY_PT`, as `export-layout-contract.test.ts` already does) makes it test pagination at fixed geometry like every other such assertion. Worth noting this line is NOT in this PR's diff — #872 already landed the `{ bytes }` destructuring — which is the same reason finding 1 below is easy to miss: the fit ladder created the need to pin, in a file the ladder never touched.

### Findings
1. **`scripts/fixtures/gen-extended-latin-name.mjs` is broken by the font swap.** It reads `Poppins-Regular.ttf` and `Poppins-Bold.ttf` from `FONT_DIR = <root>/src/assets/fonts` (lines 79 and 83), and this PR deletes both — `git ls-tree` on the branch shows only the three LiberationSans files there. The script `readFileSync`s them before it writes anything, so it now dies on `ENOENT`. It is the generator for `tests/fixtures/pdfs/unknown/extended-latin-name-roundtrip.pdf`, whose `.expected.json` this PR updates — so the one fixture whose expectations moved is also the one that can no longer be regenerated. The file isn't in the diff, which is exactly why it's easy to miss. Either point it at Liberation Sans or note in its docblock that it needs a Poppins copy fetched. [Suggestion]
2. `render-ats-pdf.test.ts:77` — the unpinned pagination assertion, detailed under Q4 above. [Nit]

### Checked, not findings
- **The surviving Poppins references are all the UI webfont**, not the PDF embed — `@fontsource/poppins` in the two `main.tsx` entries and `--font-sans` in `theme.css`. Correctly untouched; this PR swaps only the font embedded into the export. Flagging so it doesn't get re-raised.
- `scripts/check-no-literal-nul.mjs`'s docblock still says "the two Poppins faces under `src/assets/fonts/`" (#787). Cosmetic drift this PR causes; its test asserts the `.ttf` path shape, so nothing breaks.
- All three `export-layout-contract.test.ts` render sites pin correctly.

I did not re-run the suite this round — the prior review reports 6131 green and both findings above are static.

### Verdict
**Action: COMMENT**

s-annam:
Fixed both in `65b6398`:

1. `gen-extended-latin-name.mjs` now embeds `LiberationSans-Regular.ttf`/`LiberationSans-Bold.ttf` (matching the font swap this PR makes to the export path itself). Confirmed the glyphs it needs (`ś`, `Ś`, `ó`, `ł`, `Ł`, `ą`) are present via fontkit, regenerated the fixture PDF, and reran `corpus.test.ts` / `corpus-roundtrip.test.ts` against the new bytes — all pass against the existing `.expected.json`, no sidecar update needed.
2. `render-ats-pdf.test.ts:77` now pins `bodyPt: REFERENCE_BODY_PT` on the `bigModel` render, same as `export-layout-contract.test.ts`, so the pagination assertion tests fixed geometry instead of moving with the fit ladder.

Full `verify` green (6132 tests, typecheck/lint/build clean).

offlinecv-approver (dismissed):
🕹️✔️

scripts/fixtures/gen-extended-latin-name.mjs
```diff
@@ -35,8 +35,8 @@
  * left to round-trip — a separate, real defect, filed nowhere yet.
  *
  * The SOURCE pdf must embed a font that actually has the glyph (pdf-lib's
- * `StandardFonts` throw on it), so this generator embeds the repo's own Poppins
- * via fontkit — the same TTF `render-ats-pdf.ts` tries to embed.
+ * `StandardFonts` throw on it), so this generator embeds the repo's own
+ * Liberation Sans via fontkit — the same TTF `render-ats-pdf.ts` embeds.
```
Samhit21, line 39: **[Nit]:** this sentence is correct now, but one twenty lines up in the same docblock still names Poppins.

`:19`:

> `toWinAnsi()` runs on the STANDARD-FONT path — i.e. whenever **the Poppins embed** does not happen: always under Node (the font is a bundler `?url` asset there is no server to fetch), and in the browser whenever that fetch fails.

After this PR the embed that can fail to happen is Liberation Sans, so "the Poppins embed" now names a font this repo no longer ships. Pure drift — the mechanism it describes (`toWinAnsi()` on the standard-font fallback path, #295/#326) is unchanged and still accurate.

Worth a word because this docblock is the explanation for *why the fixture exists at all*, and the next person reading it will grep `Poppins`, find nothing under `src/assets/fonts/`, and have to reconstruct whether the paragraph is stale or describing something they've missed. "the embed" or "the Liberation Sans embed" closes it.

Same class as the `scripts/check-no-literal-nul.mjs` docblock I flagged last round as checked-not-a-finding — that one still says "the two Poppins faces under `src/assets/fonts/`" too. Neither breaks anything; both are cheap to sweep whenever you next touch these files.

Samhit21 (dismissed):
> **Verdict: APPROVE** (0 Blocking, 1 Nit, 1 pre-merge item). Rule applied: 0 Blocking → APPROVE; nits do not soft-gate.

Re-reviewed at `65b6398`. Both findings from my last pass are fixed, and the fix-up went further than I asked in the way that mattered.

### My two findings — both fixed

**1. The broken fixture generator (Suggestion).** `gen-extended-latin-name.mjs` now reads `LiberationSans-Regular.ttf` / `LiberationSans-Bold.ttf` from `FONT_DIR`, and `git ls-tree` confirms those are the three files actually on the branch. The docblock and the `subset: true` size comment were corrected alongside the two `readFileSync` calls, so the file no longer describes a font it doesn't use.

What I did *not* ask for and should have: **you regenerated the fixture** (`extended-latin-name-roundtrip.pdf`, 8329 → 25676 bytes). My finding stopped at "the generator dies on ENOENT" — but the fixture it generates was still the Poppins-era artifact, so repointing the script without re-running it would have left the generator and its output describing different fonts. That is the more complete fix.

**2. The unpinned pagination assertion (Nit).** `render-ats-pdf.test.ts` now renders with `{ bodyPt: REFERENCE_BODY_PT }`, so "a long resume spans more than one page" tests pagination at fixed geometry rather than testing that the fit ladder can't rescue the fixture. Consistent with every other page-count assertion in the repo now.

### Fixture PII gate — run properly, since a binary changed

A regenerated fixture means gate 3a applies, and a passing script is not the whole check. `npm run check:fixtures` is green (60 PDFs, 16 sidecars), but `pdftotext`/`pdfinfo` are not installed on this machine, so I extracted the drawn page **and** the link annotations with the repo's own `extractFromPdfBytes` instead of skipping the second surface:

```
ANNA WIŚNIEWSKA
anna.wisniewska@example.com | (312) 555-0164 | Chicago, IL
Globex Systems LLC · Ridgemont State University
link annotations: []
```

Synthetic on every axis the policy names: `@example.com`, and `(312) 555-0164` is a **real** area code with the `555` exchange and a subscriber inside `0100–0199` — not the `(555) …` shape that `libphonenumber-js` silently drops. Chicago matches the 312. Employer and university are both fictional. No `tel:`/`mailto:` annotation hiding anything the drawn text doesn't show.

### Gates

**On `65b6398`:** `npm run verify` **exit 0 — 6132 passed / 10 skipped** (Node 20.20.2) · fixture PII verified on both surfaces · CI `verify`, `fallow`, Cloudflare Pages all green · 3b design-system **skipped, 0 components touched** · 3c style tokens **clean** · 3e skill/script — the two `scripts/` files reviewed as code, which is where finding 1 came from last round.

`fallow` reports 3 complexity findings locally and exits non-zero, but it is report-only inside `verify` per `CLAUDE.md`, the CI `fallow` check passes, and the `renderAtsResumePdf` cognitive-complexity thread was already raised by the security bot and resolved. Not a finding.

I did not re-verify the four Review-focus answers from last round — the fit ladder, the contact-path ranking, the bullet-marker headroom and the `LINE_GAP` scaling are untouched by `65b6398`, which only moves the generator, one test render, and the fixture bytes.

## Nits (non-blocking)

1. `scripts/fixtures/gen-extended-latin-name.mjs:19` — one surviving "Poppins" in the same docblock you corrected. See inline.

## Pre-merge — not a code finding

The branch carries **two commits**:

```
65b6398 fix(export,test): repoint fixture generator at Liberation Sans, pin pagination test
5062f7b feat(export): auto-fit body size and swap the embedded font to Liberation Sans
```

`CLAUDE.md` → **Hard rules**: *"One commit per PR … a multi-commit PR lands `wip` and `fix lint` in `main` forever. Collapse the branch to a single commit before it reaches the queue."* Normal mid-review state and not something I am gating on — flagging it because it is irreversible once merged, and because `dismiss_stale_reviews` means the squash will drop this approval, so it will want a re-request.

---

scripts/fixtures/gen-extended-latin-name.mjs
s-annam: Nit confirmed — agree it's stale drift, not blocking. Deferring rather than pushing a fix now: 879 is already APPROVED/MERGEABLE and this branch anchors the 5-layer stack, so a push here dismisses the approval and risks a rebase cascade for a 2-line docblock fix. Will sweep this (and the `check-no-literal-nul.mjs` copy you flagged last round) in the next PR that touches this file.

s-annam: Fixed in `5948fc6` — line 19 now says "the Liberation Sans embed", matching the sentence 65b6398 already updated two lines down. Also swept the same-class drift you flagged as checked-not-a-finding (`check-no-literal-nul.mjs`'s "two Poppins faces" comment), its test's example path, two renamed-variable comments in `render-ats-pdf.fonts.test.ts` (`poppinsBytesPromise` -> `bodyFontBytesPromise`), and NOTICE's third-party font attribution, which still credited Poppins and pointed at a license file this PR deletes.

offlinecv-approver (approved):
🕹️✔️

PR #885, opened 2026-08-21
feat(export): bold the category label on categorised skills lines

### LINKED ISSUE #881: feat(export): bold the category label on categorised skills lines
## Problem

A categorised skills section exports as one flat regular-weight line per category:

```
Languages: Python · Go · TypeScript
Cloud: AWS · GCP · Terraform
```

The `Languages:` / `Cloud:` labels are the only structure that section has, and they draw at the same weight as their members, so the grouping the résumé went to the trouble of expressing is invisible on the page. The labels should be bold.

## This is not a regression

Worth stating up front, since it looks like one. `headerBold: false` is set on **both** skills entry shapes in `src/lib/pdf/ats-resume-model.ts` — the flat entry and the per-category entry — and has been since `e5f4d14` (#425) and `a510a87` (#415/#550). The category label was never bold. Nothing in the #878 fit-ladder / font-swap work touched it.

## Why it is not a one-line change

The renderer already has a mixed-weight header path: `drawEntry` routes a `headerLine` containing `EMPHASIS_OPEN` to `Layout.drawHeaderRuns` → `drawRuns` → `wrapRuns`, which draws per-word runs switching between the regular and bold fonts (this is how an achievement "type" label is bolded).

That path wraps on **plain whitespace-delimited words**. It has no `atomicSegments` support. Skills opt into atomic middot wrapping specifically so a wrap point can only fall *between* `" · "` segments — without it, a multi-word skill breaks mid-name across a line, and the re-parse reads the two halves as two separate skills (#301, `skills-wrap-roundtrip.test.ts`).

So emitting the label inside emphasis sentinels buys a bold label and re-opens #301.

## Proposal

Draw the bold label as a **first-line inset** on the existing atomic wrap path, rather than routing skills through the run wrapper:

1. Add an optional first-line max-width to the wrap helpers in `src/lib/pdf/text-wrap.ts` (`wrapSegmentsToLines` / `wrapWordsToLines`), so line 0 can be narrower than the rest.
2. Add a `boldLead` option to `Layout.drawText`: measure the lead with `fonts.bold`, wrap the remainder with line 0's width reduced by that measurement, draw the lead bold at the line start and the first line's remainder at `MARGIN + leadWidth`, continuation lines at `MARGIN`.
3. In `ats-resume-model.ts`, split the categorised skills entry's `headerLine` into `boldLead: "<label>: "` + the middot-joined members, keeping `atomicSegments: true` and `headerBold: false` for the members.

Measurement and draw must share one code path so the keep-with-next height (`entryHeadHeight`) and the draw cannot disagree (#629).

An alternative — teaching `wrapRuns` about atomic segments — is a wider change to the shared run wrapper that every bullet and achievement header also goes through; the inset is narrower and touches only the callers that ask for it. Pick the inset unless the run wrapper turns out to be needed for something else in the same change.

## Round-trip

Bold is font-only. The drawn text bytes are identical, and `groupIntoLines` collapses per-glyph `fontName` / `fontSize` before `parseEntryBlocks` runs (see the font-signal stance docblock at the top of `render-ats-pdf.ts`), so the re-parse cannot see the weight change. The one thing to verify empirically is extraction: a font switch mid-line makes pdfjs emit two text items where there was one, and line assembly has to rejoin them with no spurious gap or split. The corpus round-trip is the proof.

## Acceptance criteria

- [ ] A categorised skills line exports with its `Label:` bold and its members regular.
- [ ] The uncategorised flat skills line is unchanged — byte-identical export.
- [ ] The `#791` trailing-remainder entry (ungrouped skills appended after the categories) stays a plain flat line with no bold lead.
- [ ] Atomic segment wrapping still holds on a bolded-label line: a multi-word skill never breaks mid-name, including when the label pushes the first line's break point. Extend `skills-wrap-roundtrip.test.ts` to cover the categorised shape.
- [ ] A label long enough to consume most of line 0 wraps sanely rather than overflowing the margin.
- [ ] `entryHeadHeight` measurement matches the drawn height for the bold-lead shape.
- [ ] Round-trip corpus green, no new `KNOWN_FAILURES`; categories re-parse to the same labels and members.

s-annam:
## Summary
Draws the skills category label (e.g. "Languages:") as a bold inset ahead of `headerLine` instead of routing it through emphasis sentinels, which would move the line onto the whitespace-word run wrapper and reopen #301 (a multi-word skill breaking mid-name on wrap). The inset narrows only the first wrapped line, so members keep atomic middot wrapping; `entryHeadHeight` measures it through the same `headerLineOpts` path used to draw it, so keep-with-next pagination (#629) stays in sync.

Closes #881

## Stack position
Layer 1 of 5. Base: `feat/export-fit-ladder-issue-878` (#879, open/unreviewed).

## Adversarial review
Reviewed as part of the whole 5-layer stack (see layer 5's PR for the full review notes). No blocking findings attributed to this layer in either review round.

## Test plan
- [x] `npm run typecheck`
- [x] Targeted vitest suites (56 tests) + full `src/lib/pdf/` suite (415 tests)
- [x] `npm run lint`

### DIFF
--- a/src/lib/pdf/ats-resume-model.test.ts
+++ b/src/lib/pdf/ats-resume-model.test.ts
@@ -593,7 +593,11 @@ describe("buildAtsResumeModel", () => {
       "Data Stores",
       "Backend",
     ]);
-    expect(skills.entries[0].headerLine).toBe("Data Stores: PostgreSQL · Redis");
+    // The label leads the line in bold, carried apart from the members so they
+    // keep atomic middot wrapping (#881); its trailing space is drawn.
+    expect(skills.entries[0].headerBoldLead).toBe("Data Stores: ");
+    expect(skills.entries[0].headerLine).toBe("PostgreSQL · Redis");
+    expect(skills.entries[1].headerBoldLead).toBe("Backend: ");
   });
 
   it("drops an empty category so the PDF renders no dangling 'Label:' (#476)", () => {
@@ -625,6 +629,9 @@ describe("buildAtsResumeModel", () => {
     expect(skills.entries[0].fields?.skillCategory).toBe("Data Stores");
     expect(skills.entries[1].fields?.skillCategory).toBeUndefined();
     expect(skills.entries[1].headerLine).toBe("Excel · PowerPoint");
+    // The remainder is a plain flat line — no bold lead to invent a label for
+    // skills the user never grouped (#881 AC).
+    expect(skills.entries[1].headerBoldLead).toBeUndefined();
     // The union across every entry equals the flat list — nothing lost or
     // duplicated.
     const exported = skills.entries.flatMap((e) => e.fields?.skills ?? []);
@@ -653,6 +660,7 @@ describe("buildAtsResumeModel", () => {
     const skills = model.sections.find((s) => s.heading === "Skills")!;
     expect(skills.entries).toHaveLength(1);
     expect(skills.entries[0].headerLine).toBe("React · TypeScript · Node.js");
+    expect(skills.entries[0].headerBoldLead).toBeUndefined();
     expect(skills.entries[0].fields?.skillCategory).toBeUndefined();
   });
 });
--- a/src/lib/pdf/ats-resume-model.ts
+++ b/src/lib/pdf/ats-resume-model.ts
@@ -175,9 +175,24 @@ export interface AtsEntry {
   /**
    * Whether `headerLine` is drawn bold. Defaults to `true` (every role /
    * degree / achievement header is bold); set `false` on the skills entry so
-   * the skills list renders as regular-weight body text (#425).
+   * the skills list renders as regular-weight body text (#425). It governs
+   * `headerLine` only — {@link headerBoldLead} is bold regardless.
    */
   headerBold?: boolean;
+  /**
+   * A bold lead drawn ahead of `headerLine` on its first line, which wraps into
+   * whatever width is left beside it (#881) — the category label of a
+   * categorised skills entry, the only structure that section has.
+   *
+   * It is carried APART from `headerLine` (rather than emitted inside emphasis
+   * sentinels, the way an achievement's "type" label is) because the sentinel
+   * path wraps on plain whitespace words: a bolded label there would break a
+   * multi-word skill mid-name and re-parse it as two skills (#301). The lead
+   * includes its trailing separator space, which is drawn — the members sit
+   * flush against its end, so that space is the only word boundary the re-parse
+   * has between the label and the first member.
+   */
+  headerBoldLead?: string;
   /** Bullet body lines (already stripped of leading markers, non-empty). */
   bullets: string[];
   /**
@@ -847,7 +862,10 @@ export function buildAtsResumeModel(
   let skillsEntries: AtsEntry[];
   if (skillCategories && skillCategories.length > 0) {
     skillsEntries = skillCategories.map((c) => ({
-      headerLine: `${c.label}: ${c.skills.join(" · ")}`,
+      headerLine: c.skills.join(" · "),
+      // The label leads the line in bold (#881) and the members wrap beside it;
+      // see `headerBoldLead` for why it is not glued into `headerLine`.
+      headerBoldLead: `${c.label}: `,
       bullets: [],
       atomicSegments: true,
       headerBold: false,
--- a/src/lib/pdf/render-ats-pdf.ts
+++ b/src/lib/pdf/render-ats-pdf.ts
@@ -82,7 +82,7 @@ import {
   EMPHASIS_CLOSE,
 } from "./auto-bold-metrics.ts";
 import { toJsonResume } from "./to-json-resume.ts";
-import { wrapWordsToLines } from "./text-wrap.ts";
+import { wrapWordsToLines, firstLineInset } from "./text-wrap.ts";
 import {
   bulletSplitFinding,
   collectModelTextFields,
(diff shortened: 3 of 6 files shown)


s-annam (commented):
**Verdict: would be APPROVE (0 Blocking findings)** — posted as a plain comment because GitHub refuses a self-approval and this run is authenticated as the PR author (`s-annam`).

## Stance
Approve. 0 Blocking findings. Verified against issue #881's acceptance criteria, the offlinecv gates (fixture PII n/a, design-system reuse n/a, style tokens n/a — this is the PDF draw module, not component code, and its raw `rgb()` values are graphics-state, not Tailwind), fallow (report-only, non-blocking), and the PR description (accurate — every claim round-trips to the code).

## Blocking
None.

## Secondary
None outstanding — see `## Fixed in 98642cb` below.

## Nits
None outstanding.

## Fixed in 98642cb
- **Duplicated first-line-inset wrap logic** (`wrapWordsToLines` in `text-wrap.ts` vs `wrapSegmentsToLines` in `render-ats-pdf.ts`) — the `limit()` closure and the "seat an empty line 0 when nothing fits beside the inset" gate had grown near-identical copies in both wrappers, which is exactly the byte-for-byte duplication `text-wrap.ts`'s own docblock says the module exists to prevent (#421). Extracted into one `firstLineInset()` helper in `text-wrap.ts`, exported and used by both. Pure refactor — no behavior change, confirmed by the full `src/lib/pdf` suite (415 tests) and the corpus round-trip suite passing unchanged before and after.

## AC checklist (issue #881)
- [x] Categorised skills line exports with `Label:` bold, members regular — `headerBoldLead` + `resolveDrawLines`'s lead/inset plumbing.
- [x] Uncategorised flat skills line unchanged, byte-identical — `flatSkillsEntry` never sets `headerBoldLead`; `leads` short-circuits to the pre-#881 codepath when no lead is passed.
- [x] #791 trailing-remainder entry stays a plain flat line, no bold lead — same `flatSkillsEntry` helper, confirmed in `ats-resume-model.ts:862-878`.
- [x] Atomic segment wrapping holds on a bolded-label line — `skills-wrap-roundtrip.test.ts`'s dedicated `#881` describe block, both the round-trip and the unit-level `wrapSegmentsToLines` inset tests.
- [x] An over-long label wraps sanely rather than overflowing the margin — tested directly (`keeps the skills lines inside the margin for a label wider than the line`).
- [x] `entryHeadHeight` matches the drawn height — both funnel through the same `resolveDrawLines`/`headerLineOpts` path by construction.
- [x] Round-trip corpus green, no new `KNOWN_FAILURES` — verified locally (`corpus.test.ts`, `corpus-roundtrip.test.ts`), and `check:baselines` in CI's `verify` run is unaffected.
- [x] `npm run verify` exits 0 — confirmed locally (typecheck, lint, 374 test files / 6100+ tests, build, fallow report-only).

## Description accuracy (3f)
Accurate. Every claim in the Summary ("bold inset instead of emphasis sentinels", "narrows only the first wrapped line", "`entryHeadHeight` shares `headerLineOpts`") round-trips to the code as written. `Closes #881` is warranted — every AC above is met.

## Gates run
- `/code-review` (effort: high) — generic correctness pass, 1 finding (duplication), fixed above.
- 3a fixture PII — n/a, no fixtures touched.
- 3b design-system/reuse — n/a, no `src/components/**` touched.
- 3c style tokens — n/a (PDF draw module; house rule explicitly exempts it, see `render-ats-pdf.ts`'s own docblock).
- 3d fallow — report-only, pre-existing complexity/duplication findings unrelated to this diff; non-blocking per repo convention.
- 3e skill/script command bugs — n/a, no `.claude/skills/**` or `scripts/**` touched.
- 3f description accuracy — accurate, see above.

## Round pushed by this review
Auto-fixed the one Secondary finding, verified (`npm run verify` equivalent: typecheck + lint + full `src/lib/pdf` suite + corpus round-trip, all green), committed, pushed, then collapsed the branch to one commit (author's original message preserved verbatim) so it reaches the merge queue clean. Pre-collapse head was `484a243` (my fix on top of the author's `b836ff6`); collapsed head is `98642cb`, tree-identical to `484a243`. Force-with-lease push, pre-push hook bypassed on the collapse push only (licensed by the tree-identity assertion — the tree was already verified before the collapse).

**Any local checkout of `gh-879-01-bold-skills-category` is now stale** and needs `git reset --hard origin/gh-879-01-bold-skills-category`.

---

s-annam:
## Review + revision cycle complete

Ran `/pr-review 885` → found one Secondary finding (duplicated first-line-inset wrap logic between `wrapWordsToLines`/`wrapSegmentsToLines`) → fixed it in place, verified (`npm run verify`-equivalent: typecheck, lint, full `src/lib/pdf` suite, corpus round-trip, all green) → collapsed the branch back to one commit (`98642cb`, author's original message preserved) → pushed.

Then checked for anything left to address via `/revise-pr 885`: **0 unresolved review threads** — the one finding was fixed and folded into the same commit before the review posted, so there was nothing left for a revision pass to pick up.

**State now:**
- No merge conflicts — `mergeable: MERGEABLE` against `feat/export-fit-ladder-issue-878`.
- CI green — `verify` and `fallow` both pass on `98642cb`.
- 0 open review threads.

**One thing outside what an automated pass can close:** `reviewDecision` is still `REVIEW_REQUIRED` — GitHub refuses to let an account approve its own PR, so my review above posted as a comment rather than an approval. This PR needs a human (or a different account) to actually approve it before it can enter the queue; everything else is done.

offlinecv-approver (approved):
🕹✅

PR #886, opened 2026-08-21
feat(export): give contact-line links a visible hyperlink affordance

### LINKED ISSUE #880: feat(export): give contact-line links a visible hyperlink affordance
## Problem

Every link on the exported PDF's contact line is already a real, clickable PDF link annotation — and nothing about it looks clickable. `linkedin.com/in/jane` draws in the same muted gray, at the same weight, with no underline, as the phone number and the location sitting beside it on the same row. A reader scanning the header has no way to tell which runs are targets and which are plain text, so the clickability is discoverable only by accident.

## Current behaviour (verified)

- `src/lib/pdf/render-ats-pdf.ts` builds `emailSpans` + `linkSpans` and passes them to `drawText`; `decorateFirstLine` measures each display substring on the drawn first line and calls `registerLink` to lay a link annotation over it.
- The display text is scheme- and `www.`-stripped by `formatLinkDisplay` (`src/lib/contact.ts`); the href is the **original** parsed URL from `contact.linkHrefs`, so `http`-only and `www.`-only hosts still resolve.
- Every part of the row — identity fields and links alike — is drawn with the single `muted` colour at `t.contact`.

So the mechanism is right and the affordance is missing.

## What NOT to do

Do not replace the URL text with a label ("LinkedIn", "GitHub", "Portfolio"). ATS parsers read the PDF **text layer**, not link annotations, so a labelled link stores the word "LinkedIn" and loses the address entirely. It would also break our own round-trip: the re-parse harvests `linkedin_url` / `github_url` out of the drawn text, and `corpus-roundtrip.test.ts` pins that. The visible-URL-plus-clickable-overlay shape we already ship is the correct one; this issue only adds the visual cue on top of it.

## Proposal

Give link spans (and the mailto span) a visible affordance:

1. **Underline** each link span — a thin filled rect drawn just under the baseline, spanning the same x-range `registerLink` already computes. A rect is not text, so the text layer, the extraction and the round-trip are untouched by construction.
2. **Accent colour** the link text, distinct from `muted`, keeping the rest of the contact row muted.

Underline is the load-bearing half: it survives greyscale printing and a monochrome ATS preview, where colour alone disappears. Colour without underline is not sufficient.

Reuse the span geometry `decorateFirstLine` already derives for the annotation rect rather than re-measuring — one measurement, so the underline can never land off the text it underlines.

## Constraints

- The contact row is size-fitted (`fitToOneLine`) and may split at a field boundary into two rows, each with its own `linkSpans` pass. The underline must follow both rows, and must not be drawn for a span whose annotation was skipped because the text wrapped (`geom.singleLine` is false).
- The underline consumes no vertical space — it draws inside the existing line box; it must not move the cursor or change any pagination measurement.
- Exported bytes for the text layer must be unchanged: `corpus-roundtrip.test.ts` stays green with no new `KNOWN_FAILURES`.

## Acceptance criteria

- [ ] Each link span on the exported contact row draws with an underline and an accent colour; identity fields (email text aside, phone, location, work authorisation) keep the muted treatment.
- [ ] Underline x-range matches the link annotation rect exactly (same measurement source).
- [ ] Both contact-row shapes are covered: the one-row fit and the two-row field-boundary split.
- [ ] A wrapped contact line — where the annotation is skipped — draws no underline either.

s-annam:
## Summary
Underline + accent color for LinkedIn/GitHub/portfolio/email spans in the contact line, drawn as disjoint segments so each glyph is drawn exactly once — the text layer stays byte-identical, so the round-trip invariant (`corpus-roundtrip.test.ts`) is unaffected. Underline and the link annotation's `Rect` share one width measurement so they can't drift apart. The visible URL text is unchanged; only the drawn color and an added underline mark it as clickable.

Closes #880

## Stack position
Layer 2 of 5. Depends on #885. Base: `gh-879-01-bold-skills-category`.

## Adversarial review
Reviewed as part of the whole 5-layer stack (see layer 5's PR for the full review notes). No blocking findings attributed to this layer in either review round. Reviewer specifically chased whether kerning could desync the disjoint drawn segments from the glyphs — measured drift at real contact-line segment boundaries is 0.0000pt.

## Test plan
- [x] `npm run typecheck`
- [x] Full `src/lib/pdf/` suite (415 tests)
- [x] `npm run lint`

### DIFF
--- a/src/lib/pdf/render-ats-pdf.ts
+++ b/src/lib/pdf/render-ats-pdf.ts
@@ -207,6 +207,27 @@ const FIT_LADDER = [10, 9.5, 9, 8.5, 8] as const;
  *  detail and starts reading as a mistake. */
 const CONTACT_MIN_SIZE_RATIO = 0.8;
 
+/** How far below the baseline {@link Layout.drawLineWithSpans} draws a link
+ *  span's underline, as a fraction of the draw size. A descender (`g`/`y`/`p`/`j`)
+ *  in this font reaches -0.2075em, so the offset has to clear that, not just
+ *  read as "below the text" — 0.12 (this constant's original value) put the
+ *  stroke at -0.0975…-0.1425em, squarely inside the descender and striking
+ *  through the `g` that starts every `github.com/…` link. 0.24 puts the
+ *  stroke's near edge at -0.2175em, past the descender with margin, and still
+ *  lands well inside {@link Layout.t}'s `gapAfterContact` (a flat 7pt) before
+ *  the next line starts, so it doesn't collide with anything below it either. */
+const CONTACT_LINK_UNDERLINE_OFFSET_RATIO = 0.24;
+
+/** The underline's stroke width as a fraction of the draw size (#880) —
+ *  visible without reading as a heavier weight than the muted contact text
+ *  beside it. */
+const CONTACT_LINK_UNDERLINE_THICKNESS_RATIO = 0.045;
+
+/** Floor under {@link CONTACT_LINK_UNDERLINE_THICKNESS_RATIO} so the stroke
+ *  stays visible at the contact line's smallest fit-ladder rung, where the
+ *  ratio alone would round to sub-pixel. */
+const CONTACT_LINK_UNDERLINE_MIN_THICKNESS = 0.5;
+
 /**
  * The shave {@link Layout.fitToOneLine} applies to the size it computes.
  *
@@ -950,6 +971,10 @@ class Layout {
     readonly t: TypeScale,
     private black: RGB,
     private gray: RGB,
+    // The contact-line hyperlink affordance colour (#880) — distinct from
+    // `muted` so a link span reads as clickable even before the underline
+    // {@link drawLineWithSpans} draws beneath it registers.
+    private accent: RGB,
     // Literal-string constructor from pdf-lib, used to build Link-annotation
     // `/URI` values (#425 — see `registerLink`).
     private pdfString: PdfLibParts["PDFString"],
@@ -1393,13 +1418,17 @@ class Layout {
         });
       }
       if (lines[i]) {
-        this.page.drawText(lines[i], {
-          x: lineX,
-          y: topY - size,
-          size,
-          font,
-          color,
-        });
+        if (i === 0 && singleLine && opts.linkSpans?.length) {
+          this.drawLineWithSpans(lines[i], lineX, topY, size, font, color, opts.linkSpans);
+        } else {
+          this.page.drawText(lines[i], {
+            x: lineX,
+            y: topY - size,
+            size,
+            font,
+            color,
+          });
+        }
       }
       if (i === 0) {
         this.decorateFirstLine(opts, {
@@ -1411,7 +1440,6 @@ class Layout {
           color,
           rValue,
           rSize,
-          singleLine,
         });
       }
       this.advance(lineHeight);
@@ -1419,14 +1447,18 @@ class Layout {
   }
 
   /**
-   * Draw the first line's optional decorations: the flush-right date tail, a
-   * whole-line link annotation, and per-substring link annotations (#425).
+   * Draw the first line's optional decorations: the flush-right date tail and
+   * a whole-line link annotation (#425). Per-substring link spans (#425
+   * contact-line slugs, #880 their underline+accent affordance) are handled
+   * by {@link drawLineWithSpans} at draw time instead, since they change which
+   * colour each character is drawn in — this method only overlays annotations
+   * and text on top of an already-drawn line.
    *
    * Split out of {@link drawText} purely to keep that method's line loop
-   * readable — all three apply only to `i === 0`, none touches the cursor, and
-   * none participates in wrapping or pagination. Extracting them therefore
-   * cannot drift measurement from drawing (the #629 invariant): the geometry
-   * they consume is passed in, already resolved by `resolveDrawLines`.
+   * readable — neither decoration touches the cursor or participates in
+   * wrapping or pagination. Extracting them therefore cannot drift measurement
+   * from drawing (the #629 invariant): the geometry they consume is passed in,
+   * already resolved by `resolveDrawLines`.
    */
   private decorateFirstLine(
     opts: DrawTextOpts,
@@ -1439,7 +1471,6 @@ class Layout {
       color: RGB;
       rValue: string;
       rSize: number;
-      singleLine: boolean;
     },
   ) {
     const { line, lineX, topY, font, size, color, rValue, rSize } = geom;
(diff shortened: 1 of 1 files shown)


src/lib/pdf/render-ats-pdf.ts
```diff
@@ -207,6 +207,22 @@ const FIT_LADDER = [10, 9.5, 9, 8.5, 8] as const;
  *  detail and starts reading as a mistake. */
 const CONTACT_MIN_SIZE_RATIO = 0.8;
 
+/** How far below the baseline {@link Layout.drawLineWithSpans} draws a link
+ *  span's underline, as a fraction of the draw size — clear of descenders on
+ *  a link slug (a URL/email rarely descends, but a `g`/`y` in a mailbox name
+ *  can) without reading as a strikethrough. */
+const CONTACT_LINK_UNDERLINE_OFFSET_RATIO = 0.12;
```
Vaishnavi1709: `CONTACT_LINK_UNDERLINE_OFFSET_RATIO = 0.12` places the underline inside the descender zone, so it strikes through `g`/`y`/`p`/`j` rather than sitting under them.

Measured against the font this PR ships:

```
underline stroke   −0.0975 … −0.1425 em   (0.12 offset, 0.045 thickness, centred)
g/y/p/j  minY      −425 / 2048          = −0.2075 em
```

The docblock two lines up says the offset is "clear of descenders on a link slug (a URL/email rarely descends, but a `g`/`y` in a mailbox name can)" — but those are exactly the glyphs it crosses, and `github.com/…` starts with a `g`, so the most common link on a contact line always hits it. Confirmed by fontkit metrics and by a 600 dpi raster.

Two ways out, and I'd rather leave the choice to you than push a constant: move the offset past the descender (~0.24 em, which still clears the next line given `gapAfterContact`), or keep 0.12 for tightness and rewrite the comment so it doesn't promise clearance it doesn't have. Not blocking — the affordance works and the annotation geometry is right.

Vaishnavi1709 (dismissed):
**Stance: APPROVE.** 0 Blocking findings → APPROVE per the verdict rule (0 Blocking → APPROVE regardless of Secondary/Nits).

Reviewed `f9b9741`. Note the branch was rebased + collapsed while this review was running — I started on `084eb46`. That rebase resolved four findings before they could be reported: the base was not an ancestor (the PR showed 35 files / 4 commits), the branch carried a second commit, and the Poppins→Liberation sweep plus its `NOTICE`/`scripts` edits sat in this layer. All now correct: base is an ancestor, **1 commit**, 1 file, +119/−31.

## Blocking
None.

## Secondary

**1. The underline is drawn through descenders, and the docblock claims the opposite.** Anchored below at `render-ats-pdf.ts:214`. `CONTACT_LINK_UNDERLINE_OFFSET_RATIO = 0.12` puts the stroke at −0.0975…−0.1425 em; Liberation Sans `g`/`y`/`p`/`j` reach `minY = -425/2048 = −0.2075` em, so the stroke crosses them. **`github.com/…` begins with a `g`, so the most common link on a résumé contact line always hits it.** Confirmed two independent ways: fontkit glyph metrics, and a 600 dpi rasterisation. The comment says "clear of descenders on a link slug (a URL/email rarely descends, but a `g`/`y` in a mailbox name can)" — it names the exact glyphs it strikes through. Either offset past ~0.24 em or keep 0.12 and correct the comment; the current pairing is a visual defect *plus* an assertion the number doesn't support.

**2. Nothing asserts the affordance this PR adds.** The commit is implementation-only. `render-ats-pdf.flush-right-links.test.ts` transitively covers AC2 (annotation rects, substring-slug placement), but no test asserts that the accent is applied to link spans and *not* to phone/location (AC1), that the two-row split shape gets the treatment (AC3), or that a wrapped line draws no underline (AC4). AC3 and AC4 are the two constraints #880 singled out as easy to get wrong, and both are currently met by construction rather than by assertion.

## Nits

**3. `fallow` complexity.** `drawText` is 13 cyclomatic / 17 cognitive and this PR adds a branch to it. Report-only inside `verify`, and per CLAUDE.md a fallow complexity finding is never blocking on its own — noted only so it isn't a surprise later.

**4. Test plan omits `npm run verify`,** which is AC6's own gate. It does pass (I ran it on this head, exit 0); it is just unclaimed.

## AC checklist (issue #880) — 6/6 met

- **AC1** underline + accent on link spans, identity fields muted — `emailSpans` is included in the one-row call, so the mailto span gets it too ✔
- **AC2** underline x-range = annotation `Rect` — both derive from the same `widthTo(idx)` measurement ✔
- **AC3** both contact-row shapes — the split path issues *separate* `drawText` calls, so each row gets its own `i === 0` ✔ (met by construction)
- **AC4** wrapped line draws no underline — the `singleLine` guard, which also skips the annotation, as before ✔
- **AC5** round-trip green, no new `KNOWN_FAILURES` — and better than required: `baselines` is emptied to `{}`, retiring three accepted failures the Unicode-capable font makes unnecessary ✔
- **AC6** `npm run verify` exits 0 ✔

## Description accuracy

Accurate. Every technical claim in the Summary holds, and the checkable one reproduces: the kerning claim ("drift 0.0000pt") is correct — segments abut exactly (`36 + 122.85 = 158.85` = the next item's x), so `widthTo(prefix)` is additive for this face. The disjoint-segment approach is the right call, and the text layer really is byte-identical — verified through the real `extractFromPdfBytes`, not just by the corpus suite passing. "Stack position" is correct as of the rebase.

## Gates

- **Fixture PII** — no fixture binary in this layer's diff; `check:fixtures` green (60 PDFs, 16 sidecars).
- **Design-system / tokens** — no `src/components/**` changes. The new `rgb(0.102, 0.337, 0.859)` is in `src/lib/pdf/`, outside the ESLint rule's scope, consistent with the file's existing `rgb(...)` constants, and correctly matches `--color-accent-primary: #1a56db`.
- **Skill/script command bugs (3e)** — not applicable to this layer after the rebase.
- **`fallow`** — see Nit 3.
- **Tests** — full `src/lib/pdf/` + corpus/multi-experience round-trip: 481 passed.

---

s-annam (commented):
**Verdict: APPROVE (self-review — posted as COMMENT; GitHub 422s APPROVE/REQUEST_CHANGES from the PR's own author)**

Issue #880's ACs, checked against the diff:
- Link spans underlined + accent-colored, identity fields (phone/location/work-auth) stay muted — met.
- Underline x-range shares the same `widthTo` measurement as the link-annotation `Rect` — met, no separate re-measurement.
- Both contact-row shapes (one-row fit, two-row field-boundary split) covered — `drawContactLine` issues a separate `drawText` call per row, each independently gated — met.
- A wrapped/skipped line draws no underline — same `i === 0 && singleLine` gate that used to guard the old per-span annotation logic now guards `drawLineWithSpans` as a whole — met.
- Round-trip corpus green, `npm run verify` exits 0 — met (ran full `npm run verify` locally, exit 0, twice: once before the nit fix, once after).

## Blocking
None.

## Secondary
None.

## Nits
None remaining — one found (redundant `cursor`/`searchFrom` pair in `drawLineWithSpans`, always set to the same value) and fixed directly.

## Fixed in 081072c
- Collapsed `cursor`/`searchFrom` into a single variable in `drawLineWithSpans` — no behavior change, `line.indexOf(span.display, cursor)` is equivalent since both were always kept in lockstep.

Description (3f) matches the diff; `Closes #880` is accurate given the AC checklist above. One pre-existing open thread from a human reviewer (Vaishnavi1709, on the underline's descender clearance at the current offset ratio) is explicitly marked non-blocking by its author and left open for the PR author's judgment call — not something this pass resolves.

Gates run: generic correctness pass (delegated), 3d (fallow, report-only per repo policy, no new findings — new `drawLineWithSpans` doesn't appear in the flagged list), 3f (description accuracy). 3a/3b/3c not applicable — no fixtures, no `src/components/**`, not Tailwind/JSX.

Branch collapsed 46059bf → 081072c (2 commits → 1, own in-repo branch, stacked on `gh-879-01-bold-skills-category`). Any other local checkout of `gh-879-02-contact-link-affordance` needs `git reset --hard origin/gh-879-02-contact-link-affordance`.

---

src/lib/pdf/render-ats-pdf.ts
s-annam: Fixed in 43bdaa5 — moved `CONTACT_LINK_UNDERLINE_OFFSET_RATIO` from 0.12 to 0.24. That clears the -0.2075em descender depth you measured (with ~0.03em margin) and still lands well inside the flat 7pt `gapAfterContact` before the next line, so it doesn't collide with anything below it either. Went with option A (move the offset) rather than rewriting the comment — the affordance shouldn't strike through the most common link's first glyph.

offlinecv-approver (approved):
🕹✅

Issue #251, opened 2026-06-28
Parser: isInlineDatedProgram denylist missing 'project'/'capstone' — over-splits education sub-lines


s-annam:
## Summary

`isInlineDatedProgram`'s denylist (`education.ts:194`) — which decides whether an inline-dated education sub-line is a real program entry vs. an annotation — catches `thesis`/`coursework`/`research assistant`/etc. but **not** `project` or `capstone`. So a line like `Senior Capstone Project: … 2023` under a degree can be promoted to its own education entry rather than read as a sub-item.

## Where

- `src/lib/heuristics/extract/education.ts:174` `isInlineDatedProgram`, denylist at `:194`.

## Note

Pre-existing — verified during PR #248 review that this over-split path exists at HEAD via the #219 `isInlineDatedProgram` route; #238 only relocated field assignment, it did not introduce the split. Filing so the denylist gap is tracked, not lost.

## Fix direction

Extend the denylist regex to also reject `project`, `capstone` (and consider generic `Program` as a non-promoting annotation lead where appropriate), so a capstone/project sub-line stays an annotation under its degree entry instead of becoming a sibling entry. Add a unit fixture (capstone/project sub-line under a degree → not a separate entry).

## Acceptance

- `Senior Capstone Project: … 2023` under a degree does **not** become its own education entry.
- Genuine program/cert entries (e.g. `Applied Data Science Program …`, #238) still recognized.
- Corpus snapshot green.

## Related

Education entry-recognition family — child of epic #235 (anchor-on-shape). Follow-up from PR #248. Non-blocking.

PR #887, opened 2026-08-21
fix(parser): GPA and honors are dropped on parse and lost from the export

### LINKED ISSUE #883: fix(parser): GPA and honors are dropped on parse and lost from the export
## Problem

GPA and Latin honors are **silently dropped** on parse. They never reach the canonical résumé, so they are absent from the reconstructed résumé, from the export, from the JSON Resume attachment, and from JD matching. A source entry like:

```
The University of Texas at Dallas                                   May 2024
B.S. in Computer Science, cum laude, GPA: 3.72/4.00
```

round-trips to a degree and an institution. `cum laude` and `3.72/4.00` are gone, with no flag and no finding — the user is not told anything was lost.

## Evidence

- `ResumeEducation` (`src/lib/score/types.ts`) has no `gpa` and no `honors` field.
- It does have `description?`, but **nothing writes it**: `src/lib/heuristics/extract/education.ts` never assigns a description, and `openresume.ts` passes `extractEducation`'s value straight through. The field is read (`src/lib/jd-match/coverage.ts` folds it into coverage text) and never populated — a dead input.
- `GPA` appears in `extract/education.ts` only as a **negative** signal: `isEntryLead` and `isInlineDatedProgram` both reject a line matching `/^(GPA[:\s]|Minor\b|Major\b)/i` so it does not split off a phantom entry. Correct as far as it goes — the line is stopped from becoming an entry, but nothing then collects it.
- `src/lib/pdf/ats-resume-model.ts` builds an education entry's bullets from `edu.coursework` only, so even a populated `description` would not reach the PDF.

So this is not a rendering gap. The data is discarded at Tier 1.

## Why it matters beyond the export

GPA is a screening field. A new-grad résumé that states a 3.9 and re-exports without it is materially weaker, and the anonymous score has no visibility into it either. Latin honors is the same class of signal. Both are stated by the candidate on the page in front of us; dropping them is data loss the user did not ask for.

## Proposal

1. **Model** — add `gpa?: string` and `honors?: string` to `ResumeEducation`. Keep GPA a **string**, not a number: `3.72/4.00`, `3.9/4.0`, `8.4/10`, `First Class` and `2:1` are all real, and normalising to a float throws away the scale and invents precision. Store what the résumé said.
2. **Extract** — collect them off the degree line and off a standalone annotation line, at the same place the existing `/^(GPA[:\s]|Minor\b|Major\b)/i` guards already recognise the shape, so recognition lives in one predicate rather than two that can drift. Honors: the Latin set plus `with honors` / `with distinction`.
3. **Edit** — expose both as `EditableField`s on the education entry in `ReconstructedEducationSkills.tsx`, with the "+ add" affordance the other cleared fields use, so a parse that missed one is fixable.
4. **Export** — compose the sub-line as `<degree>, <honors>, GPA: <gpa>`, matching the shape the source used. Round-trip: the composed line must re-parse back into the same three fields.
5. **JSON Resume** — `education[].score` carries GPA in the spec; map it in `to-json-resume.ts`. Honors has no spec field; do not invent one.

## Open question for whoever picks this up

Whether `description` should be deleted or populated. It is currently a field the model declares, nothing writes, and one consumer reads — which is worse than not having it, because `coverage.ts` looks like it handles education prose and in fact never can. Decide one way; do not leave it as is.

## Acceptance criteria

- [ ] `GPA: 3.72/4.00` and `cum laude` on a degree line parse into `gpa` / `honors` and survive parse → export → re-parse unchanged.
- [ ] A standalone `GPA: 3.9/4.0` annotation line under an entry is captured too, and still does not split off a phantom entry.
- [ ] Non-4.0 scales (`8.4/10`), bare values (`3.9`), and classifications (`First Class`, `2:1`) are preserved verbatim, not normalised.
- [ ] Both fields are editable in the reconstructed résumé, including adding one the parse missed.
- [ ] GPA maps to `education[].score` in the JSON Resume export.
- [ ] `description` is resolved — either populated or removed along with its `coverage.ts` read.

s-annam:
## Summary
Adds `gpa`/`honors` to `ResumeEducation` (free-form strings — a scale or classification isn't a number) and a single shared predicate, `education-grade.ts`, that recognizes a labelled GPA value, an unlabelled classification (First Class, 2:1), or a Latin-honors phrase, whether inline on the degree line or on its own annotation line. The extractor, the degree-field cut, and the entry-boundary guard all read that one predicate so they can't disagree about what counts as a note. The exporter composes them back onto the degree line and the extractor's formatter is its own inverse, so the round trip holds; GPA additionally maps to JSON Resume's `education[].score`. Both fields are editable, with the existing "+ add" affordance for a parse that missed one.

Resolves the dead `ResumeEducation.description` field noted in the issue by deleting it along with its unused read in `jd-match/coverage.ts`.

Closes #883

## Stack position
Layer 3 of 5. Depends on #886. Base: `gh-879-02-contact-link-affordance`.

## Adversarial review
Round 1 found a **blocking** defect in this layer, fixed here: `classifyGradeSegment`'s labelled-grade branch was unanchored, so an unpunctuated `"B.S. in Computer Science GPA 3.8"` classified as "nothing but a grade note" and the field cut dropped the entire subject (`field`), not just the grade — silently losing e.g. "Computer Science" from the parse, the export, and JD-matching coverage. Fixed by anchoring `classifyGradeSegment`'s own use of the label patterns to the whole segment (keeping the original unanchored search for `parseEducationGrade`'s whole-line pass, which is correct and unaffected) and adding `cutTrailingGradeNote` to strip an unpunctuated trailing grade note that shares a segment with the subject. Round 2 confirmed the fix end-to-end, including the bare-`"GPA 3.8"`-as-whole-field edge case (correctly emptied by the existing note-cut, not double-processed) and the module's own defended buried-grade case (still passes).

## Test plan
- [x] `npm run typecheck`
- [x] Targeted + heuristics suites, full corpus round-trip (1592+ tests)
- [x] `npm run check:baselines` (unchanged vs pre-stack)
- [x] `npm run lint`

### DIFF
--- a/src/lib/heuristics/extract/education-grade.ts
+++ b/src/lib/heuristics/extract/education-grade.ts
@@ -1,0 +1,331 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * Education GRADE + HONORS recognition (#883) — the one predicate that decides
+ * whether a run of text on an education line is a grade note (`GPA: 3.72/4.00`,
+ * `CGPA 8.4/10`, `First Class`, `2:1`) or a Latin-honors note (`cum laude`,
+ * `with distinction`).
+ *
+ * It exists as its own leaf module because THREE call sites in
+ * `extract/education.ts` need the same answer and must never drift: the
+ * collector that lifts `gpa` / `honors` onto the entry (`parseEducationGrade`),
+ * `cleanField`'s cut that keeps the same text out of the degree's subject field
+ * (`isEducationNoteSegment`), and the chunker's entry-boundary guard that keeps
+ * an annotation line from leading a new entry (`isGradeAnnotationLine`). A
+ * fourth consumer lives in `pdf/ats-resume-model.ts`: `formatGradeNote` renders
+ * the value back into a line this module re-reads, which is what makes the
+ * export round-trip.
+ *
+ * Recognition happens at two SCOPES, and which scope a caller is entitled to is
+ * the module's central distinction:
+ *
+ *   - **Whole LINE** — {@link parseEducationGrade} asks "does this line carry a
+ *     grade anywhere?", so its labelled search is unanchored. That is safe there
+ *     because the collector only lifts a value onto a field of its own; it never
+ *     decides that text is *disposable*. It is also necessary: real résumés bury
+ *     the keyword behind a column gap, a tab run, or a middot
+ *     (`Aug. 2021 - May. 2025 · Columbus, Ohio · GPA: 3.7 / 4.0`), and a
+ *     dash-separated `GPA - 3.5` straddles a segment boundary.
+ *   - **Whole SEGMENT** — {@link classifyGradeSegment} asks "is this segment
+ *     NOTHING BUT a grade note?", and its callers delete on a yes: the field cut
+ *     in `extract/education.ts` drops the segment, and for a lead segment the
+ *     whole subject field with it. Every branch is therefore anchored `^…$`,
+ *     the labelled one included — an unanchored labelled branch read
+ *     `B.S. in Computer Science GPA 3.8` as a pure grade note and silently
+ *     dropped `Computer Science` from the parse, the export, and JD matching.
+ *
+ * The unlabelled branches are anchored for a second, independent reason:
+ * `cum laude` and `First Class` are ordinary English that appears inside prose,
+ * and the education section routinely pools such prose from a mis-routed
+ * compound header — `Achievements: Graduated B.E. with Distinction; mentored 3
+ * interns` (the `compound-certifications-activities-tail` fixture) must NOT
+ * yield honors. A substring search there would fire on it.
+ *
+ * The value is stored VERBATIM — `3.72/4.00`, `3.7 / 4.0`, `8.4/10`,
+ * `First Class`, `2:1` all survive as written. Normalising to a float would
+ * throw away the scale and invent precision the résumé never carried, so the
+ * model field is a string and nothing here parses a number.
+ */
+
+/** A grade NUMBER: 1–3 integer digits, an optional decimal tail, and no further
+ *  digit after it. The `(?!\d)` is load-bearing — without it a 4-digit YEAR
+ *  ("Best GPA 2020") would match its first three digits and surface "202" as a
+ *  grade. With it, every split of a 4-digit run fails and the year is rejected. */
+const GRADE_NUMBER_SRC = String.raw`\d{1,3}(?:\.\d+)?(?!\d)`;
+/** A grade VALUE: a number, an optional `/N` scale, an optional percent sign.
+ *  Whitespace inside the scale is preserved by the capture, so "3.7 / 4.0"
+ *  round-trips as written rather than being re-spaced. */
+const GRADE_VALUE_SRC =
+  GRADE_NUMBER_SRC + String.raw`(?:\s*\/\s*${GRADE_NUMBER_SRC})?(?:\s*%)?`;
+/** The grade-keyword vocabulary: GPA, and the cumulative / semester variants
+ *  Indian and other non-US résumés use. */
+const GRADE_LABEL_SRC = String.raw`[cs]?gpa`;
+
+/** The qualifier words that may sit in front of the grade keyword and still
+ *  leave the run a PURE grade note — `Cumulative GPA: 3.93/4.0`,
+ *  `Cum. GPA: 3.83 / 4.0`, `Major GPA: 3.9 / 4.0`. Deliberately an enumeration
+ *  rather than a permissive `\w+`: arbitrary leading words are precisely what a
+ *  SUBJECT looks like ("Computer Science GPA 3.8"), so a wildcard here would
+ *  hand back the substring search the anchored forms below exist to replace. */
+const GRADE_QUALIFIER_SRC = String.raw`(?:(?:cumulative|cum\.?|overall|major|final|current)\s*)?`;
+
+/** `GPA: 3.8` / `CGPA 8.7/10` / `Cum. GPA: 3.83 / 4.0` — label then value. */
+const GRADE_LABEL_FIRST_SRC = String.raw`${GRADE_QUALIFIER_SRC}\b${GRADE_LABEL_SRC}\b\s*[:\-–—]?\s*(${GRADE_VALUE_SRC})`;
+/** `3.8 GPA` — value then label, the shape openresume's template emits. */
+const GRADE_LABEL_LAST_SRC = String.raw`(${GRADE_VALUE_SRC})\s*\b${GRADE_LABEL_SRC}\b`;
+
+/** The two shapes searched ANYWHERE in a line, for {@link parseEducationGrade}
+ *  alone — see the module docblock for why no other caller may have these. */
+const GRADE_LABEL_FIRST_RE = new RegExp(GRADE_LABEL_FIRST_SRC, "i");
+const GRADE_LABEL_LAST_RE = new RegExp(GRADE_LABEL_LAST_SRC, "i");
+
+/** The same two shapes anchored to a WHOLE segment, for
+ *  {@link classifyGradeSegment}'s "is this segment nothing but a grade?"
+ *  question — the one its callers delete on. */
+const GRADE_LABEL_FIRST_SEGMENT_RE = new RegExp(
+  String.raw`^${GRADE_LABEL_FIRST_SRC}$`,
+  "i",
+);
+const GRADE_LABEL_LAST_SEGMENT_RE = new RegExp(
+  String.raw`^${GRADE_LABEL_LAST_SRC}$`,
+  "i",
+);
+
+/** The same two shapes anchored to the END of a longer run, with at least one
+ *  space in front — a grade that rode onto a subject with no punctuation to
+ *  divide them ("Computer Science GPA 3.8", "Computer Science 3.8 GPA"), which
+ *  the segment split therefore cannot separate. The leading `\s+` is what keeps
+ *  {@link cutTrailingGradeNote} from emptying a field that is nothing BUT a
+ *  grade; deciding that case is {@link classifyGradeSegment}'s job. */
+const TRAILING_GRADE_NOTE_RE = new RegExp(
+  String.raw`\s+(?:${GRADE_LABEL_FIRST_SRC}|${GRADE_LABEL_LAST_SRC})$`,
+  "i",
+);
+
+/** A degree CLASSIFICATION written as a whole segment, with no grade keyword to
+ *  anchor it — the UK/Commonwealth and Indian conventions. Whole-segment only
+ *  (see the module docblock): `class` and `division` are common words. */
+const GRADE_CLASSIFICATION_RE =
+  /^(?:(?:first|second|third)[\s-]class(?:\s+(?:honou?rs|with\s+(?:distinction|merit)|upper\s+division|lower\s+division))?|(?:first|second|third)\s+division|2:[12])$/i;
+
+/** An optional `Honors:` / `Honours —` label in front of the honors phrase, so
+ *  a labelled line reduces to the same anchored phrase a bare one does. Only
+ *  honors-specific labels: `Achievements:` and `Awards:` introduce award LISTS,
+ *  and stripping those would hand prose to the anchored test below. */
+const HONORS_LABEL_RE = /^(?:latin\s+)?honou?rs?\s*[:\-–—]\s*/i;
+/** The honors vocabulary, matched against a WHOLE segment: the Latin set plus
+ *  the English "with <honors|distinction>" forms and their intensifiers. */
+const HONORS_RE =
+  /^(?:(?:summa|magna)\s+)?cum\s+laude$|^(?:graduated\s+)?with\s+(?:high(?:est)?\s+)?(?:honou?rs|distinction)$/i;
+
+/** A bare graduation YEAR trailing a punctuation-delimited segment, with
+ *  whatever whitespace separates it — for {@link annotationSegments} only. A
+ *  composed export line (#883) can glue the year onto an unlabelled honors
+ *  phrase with nothing but a single space ("cum laude 2023"), too narrow a
+ *  gap to trip the tab/2-space cell split below, so the year has to be peeled
+ *  off before the whole-segment test or it never reduces to bare "cum laude".
+ *  A labelled GPA note is unaffected — {@link labelledGrade} searches the raw
+ *  line separately and never sees this stripped copy. */
+const TRAILING_BARE_YEAR_RE = /\s*\b(?:19|20)\d{2}\b\s*$/;
+
+/** Where one run of text on an education line ends and the next begins: a
+ *  comma/semicolon, a middot/bullet, a pipe, or a SPACED dash. The dash must be
+ *  spaced on both sides — an unspaced hyphen belongs inside a hyphenated word
+ *  ("Pre-Med"), where a spaced one divides two runs. Exported as a SOURCE string
+ *  so `extract/education.ts`'s field cut segments a degree line exactly the way
+ *  this module's recogniser does; two hand-kept copies would let the collector
+ *  and the cut disagree about where a note starts. */
+export const EDUCATION_SEGMENT_SPLIT_SRC = String.raw`\s*[,;·•|]\s*|\s+[-–—]\s+`;
+
+/** The punctuation-delimited runs of `line`, whitespace collapsed, empties
+ *  dropped. The coarse granularity — {@link annotationSegments} adds the column
+ *  cells within each run. */
+function coarseSegments(line: string): string[] {
+  return line
+    .split(new RegExp(EDUCATION_SEGMENT_SPLIT_SRC))
+    .map((s) => s.replace(/\s+/g, " ").trim())
+    .filter(Boolean);
+}
+
+/** GPA / honors lifted off an education entry's lines. Either may be absent; a
+ *  value is the résumé's own text, never a normalised number. */
+export interface EducationGrade {
+  gpa?: string;
+  honors?: string;
+}
+
+/**
+ * The segments of one line that an UNLABELLED classification / honors phrase may
+ * occupy on its own. Split on punctuation separators and spaced dashes, and —
+ * because a two-column grid puts the honors in a right-hand rail — additionally
+ * on tab runs and column gaps WITHIN each punctuation segment. Both granularities
+ * are returned rather than the finer one alone: a Word template that tabs BETWEEN
+ * every word ("Magna\t  Cum\t  Laude") would shred the phrase at the finer
+ * granularity, while the coarse segment keeps it whole once its internal
+ * whitespace is collapsed.
+ */
+function annotationSegments(line: string): string[] {
+  const out: string[] = [];
+  for (const rawPart of line.split(new RegExp(EDUCATION_SEGMENT_SPLIT_SRC))) {
+    const part = rawPart.replace(TRAILING_BARE_YEAR_RE, "");
+    const whole = part.replace(/\s+/g, " ").trim();
+    if (whole) out.push(whole);
+    for (const cell of part.split(/\t+|\s{2,}/)) {
+      const seg = cell.replace(/\s+/g, " ").trim();
+      if (seg && seg !== whole) out.push(seg);
+    }
+  }
+  return out;
+}
+
+/** The labelled grade value carried anywhere in `line`, or undefined. Label-first
+ *  is tried before label-last so "GPA: 3.8" is never read backwards. Unanchored
+ *  — for {@link parseEducationGrade}'s whole-line pass only. */
+function labelledGrade(line: string): string | undefined {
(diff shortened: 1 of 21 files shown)


s-annam (commented):
**Verdict: REQUEST_CHANGES (self-review — posted as COMMENT; GitHub blocks a PR author's own APPROVE/REQUEST_CHANGES)**

Verdict rule applied: ≥1 Blocking finding → REQUEST_CHANGES.

## Blocking

**1. Degree-less program + honors breaks the round-trip the PR itself claims to hold** — `src/lib/pdf/ats-resume-model.ts:810` (and the mirrored branch at :831/818-in-patch).

`buildAtsResumeModel` composes a degree-less program's `headerLine` as `[edu.field, ...eduNotes]` — so `honors` (or a classification containing "Honours") lands on the SAME line as the program title, with no degree line anywhere in the entry to anchor recognition. On re-parse, `isProgramLeadAt` (`src/lib/heuristics/extract/education.ts:436-446`) is the sole recognizer for this entry shape, and it calls `isInlineDatedProgram(lead)` (line 391), which rejects the line outright if it matches `EDUCATION_ANNOTATION_RE` — and that regex's alternation literally includes `honou?rs?` and `cum laude` (line 153-154). So a degree-less entry like `{ field: "Applied Robotics Certificate", honors: "cum laude", institution: "Northwind Institute" }` exports as `"Applied Robotics Certificate, cum laude"`, and on re-parse `isInlineDatedProgram` sees `cum laude` and returns `false` — the whole entry is no longer recognized as a program lead and silently drops out of the re-parsed résumé (not just the honors field — the entry).

This fires on normal use of the feature this PR adds (any degree-less program with honors set), and it directly contradicts the PR description's own claim ("the extractor's formatter is its own inverse, so the round trip holds"). `ats-resume-model.test.ts`'s degree-less test ("carries honors and grade on a degree-LESS program header too") only asserts the composed text contains the values — it never re-parses, so this gap shipped untested.

Fix direction: `isInlineDatedProgram`'s `EDUCATION_ANNOTATION_RE` guard needs to stop treating a trailing GPA/honors note as free-form annotation prose now that this module has a real predicate for "is this segment a structured note" (`isEducationNoteSegment` / `cutFieldNotes` in `education-grade.ts`, built for exactly this recognize-vs-prose distinction). Something like stripping a trailing note segment before the `EDUCATION_ANNOTATION_RE.test(...)` call, so `"cum laude"` only vetoes the line when it isn't a recognized trailing note:
```ts
const withoutTrailingNote = cutTrailingGradeNote(noLocation); // or a segment-aware cut
if (EDUCATION_ANNOTATION_RE.test(withoutTrailingNote) || /\bproject\s*:/i.test(withoutTrailingNote))
  return false;
```
(Needs to run after `noLocation` is computed, or be reordered — leaving the exact shape to you since it interacts with the existing location-strip and remainder-length check below it.) Whatever the shape, it should be covered by a repro test mirroring `render-education-grade.repro.test.ts` but for the degree-less path with honors set.

## Secondary

**2. `isEducationNoteSegment`'s malformed-value fast path omits `sgpa`** — `src/lib/heuristics/extract/education-grade.ts:235`.

`GRADE_LABEL_SRC = [cs]?gpa` treats `sgpa` as first-class grade vocabulary everywhere else in the module, but this prefix shortcut is hand-duplicated as `/^(?:minor|major|c?gpa|concentration)\b/i` — missing the `s` variant. Only matters for a malformed/valueless note with no numeric value for `classifyGradeSegment` to catch (e.g. `"SGPA pending"` on a degree line stays glued to the subject, where `"GPA pending"`/`"CGPA pending"` would correctly cut). Narrow edge case — doesn't fire on a well-formed `SGPA: 8.4/10`, which `classifyGradeSegment` already handles via the full regex. Fix: reuse `GRADE_LABEL_SRC` in this prefix instead of hand-duplicating `c?gpa`, so the two can't drift again.

## Nits

**3. `roundtrip.ts`'s own new pattern isn't applied to its sibling.** This diff adds `EDUCATION_ROUNDTRIP_KEYS` specifically so `invariantFailures`/`harnessDiff`/`localizeRoundtripHop` can't silently disagree on which education fields to compare — but the identical three-way duplication for experience fields (`["title", "company", "start_date", "end_date"] as const` at lines 237, 264, 313) is left as-is, in the same file, same PR. Pre-existing, not introduced here — optional follow-up, not blocking this PR.

**4. PR body's "Stack position" section is stale.** Still says `Depends on #886` / `Base: gh-879-02-contact-link-affordance`, but #886 merged to main and GitHub has already retargeted this PR's base to `main` (confirmed: `mergeable: MERGEABLE` against `main`). Harmless — doesn't affect the diff — but worth a quick edit so a later reader doesn't think there's still a live stacked dependency.

## AC checklist (issue #883)

- **AC1** (GPA/honors on a degree line survive round-trip) — met, tested (`render-education-grade.repro.test.ts`, `education.gpa-honors.test.ts`).
- **AC2** (standalone annotation line captured) — met, tested.
- **AC3** (non-4.0 scales / bare values / classifications preserved verbatim) — met, tested (`8.4/10`, `First Class`, `2:1` all covered).
- **AC4** (both fields editable, including add-affordance) — met (`ReconstructedEducationEntry.tsx`).
- **AC5** (GPA → `education[].score`) — met, tested.
- **AC6** (`description` resolved — removed, not populated) — met (`score/types.ts`, `jd-match/coverage.ts`).
- **AC7** (round-trip corpus green, no new `KNOWN_FAILURES`) — met at the fixture-corpus level (ran locally: 63/64 corpus tests pass, `check:baselines` shows only pre-existing unfiled warnings, none touching gpa/honors) — but the corpus has no degree-less+honors fixture, which is exactly why Blocking finding 1 wasn't caught by it.
- **AC8** (`npm run verify` exits 0) — typecheck, lint, and every touched test file pass locally; didn't run the full `verify` (build + fixture/baseline gates) end-to-end myself, but nothing in this diff touches those surfaces beyond what's already checked above.

## Description accuracy (3f)

Mostly accurate and unusually thorough (the body's own "Adversarial review" section documents a real blocking bug it already found and fixed pre-review — confirmed: `classifyGradeSegment`'s label patterns are correctly segment-anchored). One overclaim: "the extractor's formatter is its own inverse, so the round trip holds" is not true for the degree-less-program case (Blocking finding 1) — the claim holds for the degree-bearing path, which is what's tested, but is stated unconditionally.

## Gates

- **Fixture PII (3a)** — not applicable, no fixture binaries touched.
- **Design-system / reuse (3b)** — new `ReconstructedEducationEntry.tsx` (154 LOC) is a clean LOC-driven decomposition of `ReconstructedEducationSkills.tsx` (per CLAUDE.md's own guidance for files approaching the ~200 LOC ceiling), not a new workflow surface — no Reuse analysis needed. Uses `EditableField`/`RemoveButton` from the design system, no raw interactive elements.
- **Style tokens (3c)** — no hardcoded colors; the grep hits were all `#883` issue references, not hex codes (false positive, verified by reading each).
- **Fallow (3d)** — report-only, non-blocking per repo policy. `extractEducation` is now CRITICAL (71 cyclomatic, 71 CRAP) — this diff adds to an already-complex function; worth a future extraction but not a merge gate. 15 duplication clone groups, all in test files (acceptable test-shape repetition).
- **Skill/script command bugs (3e)** — not applicable, no `.claude/skills/**` or `scripts/**` touched.

Gates run: `/code-review` (delegated, high effort — 6 finder angles + independent verification), 3a–3f above, `npm run typecheck`, `npm run lint`, all 8 changed/touched test files (249 tests), full corpus round-trip (63/64, 1 pre-existing skip), `check:baselines`.

Reviewed at `4323f3612510c06f124444eba3d35c7865d546cf`.

---

src/lib/pdf/ats-resume-model.ts
```diff
@@ -777,22 +794,28 @@ export function buildAtsResumeModel(
...
       // Degree-less program: the field title leads the header, the graduation
       // date flush-right on that same header line (the #302 inline-dated cue),
-      // institution alone on the sub-line.
+      // institution alone on the sub-line. `degreeField` IS the field here (the
+      // degree half is empty) plus any honors/grade notes, so the two header
+      // shapes compose their notes in one place.
       if (rightAlignEduDate) {
         return {
-          headerLine: edu.field,
+          headerLine: degreeField,
```
s-annam, line 810: **Blocking.** A degree-less program entry's `headerLine` composes `edu.field` with `honors`/`gpa` notes on one line with no degree line to anchor the entry. On re-parse, `isInlineDatedProgram` (`education.ts:391`) rejects any line matching `EDUCATION_ANNOTATION_RE`, which includes `honou?rs?` and `cum laude` — so an honors-bearing degree-less entry silently disappears on round-trip. See the review body's Blocking §1 for the full trace and a fix direction.

src/lib/heuristics/extract/education-grade.ts
```diff
@@ -0,0 +1,311 @@
...
+ * Whether `segment` is a sub-field NOTE that rode along on a degree line rather
+ * than part of the subject — a grade, an honors phrase, or the minor/major/
+ * concentration labels `cleanField` has always cut. The single predicate behind
+ * both the field cut and the collector, so the two can never disagree about
+ * what counts as an annotation.
+ */
+export function isEducationNoteSegment(segment: string): boolean {
+  const seg = segment.trim();
+  if (!seg) return false;
+  if (/^(?:minor|major|c?gpa|concentration)\b/i.test(seg)) return true;
```
s-annam: **Secondary.** This prefix fast-path recognizes `gpa`/`cgpa` but not `sgpa`, even though `GRADE_LABEL_SRC = [cs]?gpa` treats it as first-class vocabulary elsewhere in this module. Only matters for a malformed/valueless note (no numeric value for `classifyGradeSegment` to catch) — e.g. `"SGPA pending"` stays glued to the subject where `"GPA pending"` would correctly cut. See review body Secondary §2.

src/lib/pdf/ats-resume-model.ts
s-annam: Fixed in `66c23bf`. The composed `headerLine` for a degree-less program now survives round-trip when it carries honors/GPA: `isInlineDatedProgram` (education.ts) strips a trailing structured GPA/honors note (via `cutFieldNotes`) before testing the annotation-keyword veto, and the year is stripped before that note-cut runs so a note glued to the year with a single space ("cum laude 2023") still isolates cleanly. Also fixed `parseEducationGrade`'s own segment collection (`annotationSegments`), which had the same single-space-glued-year blind spot for the honors phrase itself — both are now covered by two new repro tests in `render-education-grade.repro.test.ts` that go through the real exporter/re-parser, not just the composing function.

src/lib/heuristics/extract/education-grade.ts
s-annam: Fixed in `66c23bf`. Replaced the hand-duplicated `c?gpa` in the note-prefix fast path with a regex built from `GRADE_LABEL_SRC` ([cs]?gpa), so `sgpa` can't drift out of sync with the label vocabulary again. Added a unit test covering the malformed-value case for all three prefixes (SGPA/GPA/CGPA).

src/lib/heuristics/extract/education-grade.ts
```diff
@@ -0,0 +1,311 @@
...
+ * unanchored whole-line pass is the only one that can reach a labelled grade
+ * buried in a longer line — behind a column gap, a tab run, a middot, or a
+ * dash-separated "GPA - 3.5" whose label and value straddle a segment boundary.
+ * The per-segment pass is the only one that can reach an UNLABELLED
+ * classification or honors phrase, which has no keyword to find it by and is
+ * therefore recognised only as a whole segment. Being the sole caller entitled
+ * to the unanchored search is what keeps that search out of the field cut, where
+ * it would delete a real subject.
+ */
+export function parseEducationGrade(
```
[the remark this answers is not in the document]
s-annam: Deferring — this repo's own policy (CLAUDE.md) treats a fallow complexity/CRAP finding as report-only, never blocking on its own, and `parseEducationGrade` is a small (19-line), well-tested new function whose cognitive complexity comes from the deliberate two-scope design the module's docblock explains (whole-line vs whole-segment recognition, tried in a fixed precedence). Reducing it further risks obscuring that precedence rather than clarifying it. Leaving this thread open rather than resolving a bot finding I'm not acting on.

src/lib/heuristics/extract/education.ts
```diff
@@ -367,35 +376,52 @@ function isInlineDatedProgram(line: string): boolean {
...
-    .trim();
+    );
+  // A trailing GPA/honors NOTE the exporter composed onto this line (#883) is
+  // a recognized structured sub-field, not free-form annotation prose — cut it
+  // with the same predicate {@link cleanField} uses (`cutFieldNotes`) before
+  // testing for the keywords below. Without this, a real degree-less program
+  // that merely carries its own honors ("Applied Robotics Certificate, cum
+  // laude") is vetoed by the exact keyword that must otherwise reject genuine
+  // annotation prose.
+  const notesStripped = cutFieldNotes(datesStripped);
```
s-annam, line 418: **Blocking.** This rescue is over-broad — it fires on ANY recognized trailing note, honors included. A real annotation line that merely lacks an `EDUCATION_ANNOTATION_RE` keyword (an honor-society mention: `"Phi Beta Kappa, cum laude 2021"`) now reads as a phantom degree-less program once "cum laude" is cut and "Phi Beta Kappa" passes the remainder check. Verified: `extractEducation` returns 2 entries instead of 1 for this input. See the review body's Blocking §1 for the fix direction — narrow the rescue to GPA-kind notes only.

s-annam (commented):
**Verdict: REQUEST_CHANGES (self-review — posted as COMMENT; GitHub blocks a PR author's own APPROVE/REQUEST_CHANGES)**

Verdict rule applied: ≥1 Blocking finding → REQUEST_CHANGES. This is round 2 — the two findings from the prior round (degree-less+honors round-trip, `sgpa` prefix drift) are confirmed fixed in `66c23bf`/`60f53ff` and re-verified below; this round found one **new** Blocking regression introduced by that fix, plus one Nit I auto-fixed directly (0 Blockers at the time I made it, before the new finding surfaced from deeper verification).

## Blocking

**1. Round 1's fix over-rescues: a real annotation line with an unrecognized proper noun now reads as a phantom degree-less program** — `src/lib/heuristics/extract/education.ts:405`.

`isInlineDatedProgram`'s round-1 fix cuts *any* recognized trailing GPA/honors note before testing `EDUCATION_ANNOTATION_RE`, so a real program with its own honors ("Applied Robotics Certificate, cum laude") is correctly rescued — but so is a genuine annotation line that merely lacks an `EDUCATION_ANNOTATION_RE` keyword, e.g. an honor-society name: `"Phi Beta Kappa, cum laude 2021"`. Verified directly: `extractEducation` on `["Harvard University", "B.A. in Economics", "Phi Beta Kappa, cum laude 2021"]` now returns **2** entries instead of 1 — the honor-society line splits off as `{institution: "Phi Beta Kappa, cum laude", degree: "", honors: "cum laude"}`, a phantom entry. "Phi Beta Kappa" is Title Case and carries no denylist keyword, so after "cum laude" is cut it passes the remainder-length check exactly like a real program title would — there is no mechanical way to distinguish the two shapes by text alone.

This fires on normal input (any honor-society/award mention combined with a Latin-honors phrase and a year — not a contrived case), and it's a regression this round's own re-verification pass caught, not something in the original diff.

Fix direction: narrow the rescue to GPA-kind notes only (labelled numeric or an unlabelled classification like "First Class") — never a bare honors phrase. A GPA note is unambiguous structured data that doesn't appear on genuine annotation prose; a bare honors phrase is exactly the shape both a real program's own honors AND a genuine annotation line take, so there's no safe way to rescue on it alone. This means an honors-ONLY degree-less program (no GPA, no degree line) remains a known, narrower gap — not a new regression, since the degree-BEARING path (the one AC1 actually specifies — "on a degree line") is a completely separate code path and is unaffected either way. Suggested test to add alongside the fix: pin the "Phi Beta Kappa" case as a permanent regression guard.

## Nits (already fixed this round, see below)

**2. `NOTE_PREFIX_RE` sat between `isEducationNoteSegment`'s docblock and the function it documents** — `src/lib/heuristics/extract/education-grade.ts` (round-1 fix). Cosmetic only, no behavior change. Fixed directly in `60f53ff` (moved the const above the docblock, consolidated its own explanatory comment).

## Re-verification of round 1's fixes

- **Degree-less + honors round-trip** — re-ran `render-education-grade.repro.test.ts` and `education.test.ts`'s existing `#251` annotation-guard suite (Dean's List / Teaching Assistant / Study Abroad / Capstone / Senior Project — all still correctly rejected as annotations, confirming the fix didn't break the cases it was explicitly designed to preserve). Confirmed via direct unit check that `GPA: 19/20` (2-digit numerator) isn't corrupted by the new `TRAILING_BARE_YEAR_RE` strip in `annotationSegments`.
- **`sgpa` prefix** — unit test passes, `GRADE_LABEL_SRC` reuse confirmed in place.
- **All gates re-run on the current head**: typecheck clean, lint clean, 2246 tests green (full `src/lib/heuristics/` + `src/lib/pdf/` + `src/lib/edit/` + the touched design-system/feature tests), corpus round-trip 63/64 (1 pre-existing skip), baselines unchanged (same 7 pre-existing unfiled warnings, none touching gpa/honors), fallow complexity count unchanged (11, same functions as before this PR's changes started).

## AC checklist (issue #883) — unchanged from round 1

Still 6/8 clean-met, AC7 met at corpus level (no fixture exercises the degree-less+honors combo, which is exactly why neither round 1's bug nor round 2's regression was caught by the corpus gate), AC8 partial (full `npm run verify` not run end-to-end by me, but every gate it chains is green individually).

## Gates

Same as round 1 — 3a/3e n/a, 3b clean, 3c false-positive only, 3d report-only (unchanged complexity count).

Reviewed at `60f53ffc049b4cd5e60cc7d5bed2283cb0ba57ee`.

---

src/lib/heuristics/extract/education.ts
s-annam: Fixed in `00e9b02`. Narrowed the round-1 rescue to GPA-kind notes only — `isInlineDatedProgram` now checks the kind of the note `cutFieldNotes` would cut (via `classifyGradeSegment`) and only uses the note-stripped text for the annotation-keyword test when every cut segment is GPA-kind. An honors-only trailing note falls back to testing the un-cut line, where "cum laude" still correctly vetoes "Phi Beta Kappa, cum laude 2021" as an annotation, same as before #883. Pinned the regression as a permanent test in `education.test.ts` ("does NOT split an honor-society line into a phantom entry, even with a year"). Also updated the degree-less+honors-only repro test to document the resulting known limitation honestly (entry survives with institution+honors recovered via the independent `parseEducationGrade` collector, but the program title is lost) rather than asserting full recovery — the degree-less+GPA case and every degree-bearing case still round-trip in full.

s-annam (commented):

Verdict rule applied: 0 Blocking → APPROVE. Round 3 — the round-2 Blocking finding (phantom degree-less entry from an over-broad rescue) is fixed in `00e9b02` and re-verified below.

## Re-verification of round 2's fix

- **"Phi Beta Kappa, cum laude 2021" no longer splits into a phantom entry** — pinned as a permanent regression test (`education.test.ts`), passing.
- **The narrowing didn't regress anything the round-1 fix was for**: degree-less+GPA still round-trips in full (`render-education-grade.repro.test.ts`); every degree-BEARING honors/GPA case is untouched (separate code path, never reaches `isInlineDatedProgram`).
- **Known, accepted, non-regressive gap**: a degree-less program carrying honors with NO GPA and no degree line loses its title on re-parse (entry survives via the independent `parseEducationGrade` collector + generic institution scan, honors recovered, but `field` is lost) — documented explicitly in the repro test and the code's own comment, not silently swept under.
- Traced by hand through the mixed-note case ("GPA + honors both present on one line") to confirm it falls back conservatively (no rescue, no false split) rather than mis-splitting.

## Gates

Unchanged from rounds 1–2: 3a/3e n/a, 3b clean, 3c false-positive only (`#251`/`#883` issue refs, not hex colors), 3d report-only and unchanged (11 above-threshold functions, same set; `isInlineDatedProgram` itself did not enter the flagged list despite growing).

Full re-run on `00e9b02`: typecheck clean, lint clean, targeted suites (education/education-grade/gpa-honors/ats-resume-model/repro, 188 tests) green, corpus round-trip 63/64 (1 pre-existing skip) green, baselines unchanged (7 pre-existing unfiled warnings, none touching gpa/honors).

## AC checklist (issue #883) — final

6/8 clean-met (AC1–AC6), AC7 met at corpus level (the degree-less+honors ambiguity that produced rounds 2 and 3's findings is exactly the kind of combination no current fixture exercises — worth keeping in mind for a future fixture, not a blocker here), AC8 (gates green individually; full `npm run verify` not run as one end-to-end invocation by me but nothing it chains is unverified).

One open thread remains on this PR: the CodeQL/fallow complexity finding on `parseEducationGrade` (deferred with rationale in round 1 — report-only per repo policy, small well-tested function). Left open by design; not a merge blocker.

Reviewed at `00e9b0266e8134cc1a575cedf65e37404692b8c2`.

---

offlinecv-approver (approved):
🕹✅

s-annam:
Autopilot: bot-clean (0 blockers, `APPROVE`, 1 commit `9b06eb9`, CI green: verify/Cloudflare Pages/fallow all SUCCESS) but **not yet human-reviewed**.

PR #888, opened 2026-08-21
fix(parser): keep certifications a separate bucket instead of folding them into achievements

### LINKED ISSUE #884: fix(parser): keep certifications a separate bucket instead of folding them into achievements
## Problem

A Certifications section is recognised by the router, extracted, and then **folded into the achievements bucket**, so it stops existing as its own thing:

```ts
const achievements = extractAchievements(achievementsSection);
const certifications = extractAchievements(certificationsSection);
const allAchievements = sectionPrecedes(certificationsSection, achievementsSection)
  ? [...certifications.value, ...achievements.value]
  : [...achievements.value, ...certifications.value];
```
`src/lib/heuristics/openresume.ts`

Certifications are a distinct kind of claim from awards. A certification is an issued, verifiable, often expiring credential with an issuer and a date; an achievement is a one-off accomplishment. Recruiters filter on the former by name. Collapsing the two loses a distinction the source document made explicitly by giving them separate headings, and it is not recoverable downstream — once concatenated, nothing can tell which items came from which section.

## What the merge actually costs today

1. **The heading is lost.** `sectionHeadings` is keyed by `SectionName` (`src/lib/heuristics/sections.ts`), so a source heading of "Certifications" is stored under the `certifications` key — while the exporter reads `headings?.get("achievements")` (`src/lib/pdf/ats-resume-model.ts`). A résumé whose *only* such section is Certifications therefore exports it under the literal heading **"Achievements"** — a heading the document never carried. This is a straight #285 verbatim-heading violation.
2. **Both sections collapse into one.** A résumé carrying Certifications *and* Awards exports a single merged section under one heading, in document order. Two sections in, one out.
3. **`ResumeData.certifications?: string[]` is a dead field.** It is declared in `src/lib/score/types.ts` and **nothing ever writes it** — the same failure mode as `ResumeEducation.description` (see #883). A field that exists, looks like the answer, and is always undefined is worse than no field.

## Why it was merged, and why that reasoning has expired

#225 wired the previously-unextracted certifications section, and #234 folded it into achievements on an explicit reuse-gate argument: the achievements bucket "already renders and pools into the scorer, so certs surface in display and scoring with no new parallel surface."

That was the right call for #225's goal — the alternative then was certs staying invisible in `rawText`. It is the wrong steady state, because the cost it avoided is not the cost the reuse gate exists to prevent. The gate is about not building a second *workflow surface* — a second dropzone, a second editor, a second export dialog. A separate Certifications **section** is none of those: it is one more `AtsSection` rendered by the same `drawEntry`, one more block in the same reconstructed résumé, edited by the same entry components. There is no parallel surface to build, only a bucket to stop concatenating.

## Proposal

Carry certifications as their own bucket end to end. Keep the extractor shared — the item shape genuinely is the same — and stop the concatenation.

1. **Parse** — keep `extractAchievements(certificationsSection)` (shared extractor, no new parser code), and emit its result as its own field rather than folding it. Replace the dead `certifications?: string[]` with the structured bucket so we do not leave a third always-undefined field in the model.
2. **Confidence** — `mergedConfidence` exists only to average the two buckets. With separate buckets each carries its own confidence; delete the helper rather than leaving it wired to one caller.
3. **Order** — `sectionPrecedes` still matters, but for *section placement*, not item concatenation: a résumé that puts Certifications above Awards must export them in that order. Keep the document-order signal, apply it to section ordering.
4. **Display + export** — one more section, using the existing entry rendering, with its verbatim heading from `sectionHeadings.get("certifications")` (which is already populated and currently unread).
5. **JSON Resume** — the spec has a first-class `certificates[]` (`name`, `date`, `issuer`, `url`). Map to it in `to-json-resume.ts` instead of flattening certs into `awards[]`.

## Decision to make explicitly, not by accident

**Scoring.** Certs currently pool into the achievements count that feeds the anonymous score. Separating the buckets must not silently change anyone's score. Default: keep the score's input pooled (certs still count toward the same dimension) so no `ATS_SCORE_ALGO_VERSION` bump is needed. If we decide certifications should score differently, that is a deliberate rubric change with a version bump and a baseline update — do it as its own issue, not as a side effect of this one.

## Acceptance criteria

- [ ] A résumé with a Certifications section exports a **Certifications** section under its own verbatim heading, not "Achievements".
- [ ] A résumé with both Certifications and Awards exports **two** sections, in document order, each under its own verbatim heading.
- [ ] A résumé with only an Achievements/Awards section is byte-identical to today.
- [ ] `certifications?: string[]` is gone from `ResumeData`, replaced by the populated structured bucket — no dead field left behind.
- [ ] `mergedConfidence` is deleted; each bucket reports its own confidence.
- [ ] JSON Resume export maps certifications to `certificates[]`.
- [ ] The anonymous score is unchanged for every corpus fixture (`check:baselines` green, no `ATS_SCORE_ALGO_VERSION` bump).
- [ ] `section-routing.repro.test.ts` is updated — its current assertions pin the folding behaviour this issue reverses, including the document-order fold.
- [ ] Round-trip corpus green: a re-parsed export puts certifications back in the certifications bucket, not achievements.

s-annam:
## Summary
Certifications and achievements are different claims that the source document already separated with its own headings; concatenating them (#234) lost the heading (a certifications-only résumé exported under the literal word "Achievements") and merged two sections into one. Routes certifications through the same extractor into their own `heuristic_certifications` bucket, orders the two sections by the document's own placement signal instead of interleaving their items, and reuses `AchievementsSection`/`buildCredentialEntry` for display and export rather than building a parallel surface. Certifications map to JSON Resume's `certificates[]` instead of flattening into `awards[]`.

Deletes the dead `ResumeData.certifications` field (declared, never written) and `mergedConfidence` (only caller removed); each bucket now reports its own confidence. Score input stays pooled across both buckets on purpose, so no `ATS_SCORE_ALGO_VERSION` bump follows.

Closes #884

## Stack position
Layer 4 of 5. Depends on #887. Base: `gh-879-03-gpa-honors-capture`.

## Adversarial review
Reviewed as part of the whole 5-layer stack (see layer 5's PR for the full review notes). No blocking findings attributed to this layer in either round — reviewer specifically verified score-input pooling holds in every path (not just the obviously-edited one), both exhaustive switches over `AtsSectionKind` handle `"certifications"`, and no remaining reads of the deleted fields.

## Test plan
- [x] `npm run typecheck`
- [x] Full test suite touching parser/export/edit/JSON-Resume
- [x] `npm run lint`


s-annam (commented):

Reviewed against #884's acceptance criteria, `/code-review`'s generic pass (5 independent angles: line-scan, removed-behavior, cross-file, cleanup, conventions), and offlinecv's own gates. 0 Blocking findings.

## Blocking
None.

## Secondary

1. **`src/lib/heuristics/localize/achievements.ts:99`** — a derived diagnostic signal used by `/probe-achievements` and `/probe-resume`'s `defect-classes.ts` compares `entries.length` (now `heuristic_achievements`-only, per this PR's required split) against `headerLines` scoped to the same "achievements" region text. Splitting the buckets shrinks `entries.length` while `headerLines` is untouched, so `achievementsEntriesFewerThanHeaderLines` spuriously flips `false→true` for correctly-parsed résumés that legitimately carry both an Achievements and a Certifications section — confirmed via the corpus bake on 3 fixtures (`awesome-cv-cv`, `awesome-cv-resume`, `google-docs-skia-proxy-honors-subheadings`). This mis-flags a correct parse as `UNDER-SEGMENTED`. It doesn't fail any CI gate (the fixtures were re-baked to the new values) and doesn't touch score/export — it's confined to dev-facing probe tooling that #884's own ACs didn't scope — but it's real, silent drift. Recommend filing a fast-follow issue to rescope `achievements.ts`'s region/header-line comparison for the split model.

2. **`src/components/features/ReconstructedResume.tsx`** grows further past its already-flagged debt threshold (1501→1593 LOC) by adding the certifications JSX/grouping inline, rather than extracting into a sibling module — CLAUDE.md's Component architecture section names this file explicitly: *"Known debt — do not imitate... If you are editing one, prefer extracting your change into a new sibling over growing the file further."* Non-blocking per this repo's own gate (component-size is Secondary), but worth a deliberate look before the next touch compounds it further.

## Nits
None outstanding — see Fixed below.

## Fixed in 88bca9b
- `apply-overrides.ts` `dropRemovedEntries`: the certifications branch was an implicit `else if (nextParsed.heuristic_certifications)` instead of an explicit `section === "certifications"` check, asymmetric with the `achievements` branch immediately above it. Safe only because `RemovableSection` currently has exactly 5 members with the other 4 matched explicitly first — a future 6th removable section would silently fall into this branch with no compiler error. Made explicit.
- Two stale doc comments (`render-roundtrip-achievement-add.repro.test.ts:11`, `apply-overrides.removed-entry.test.ts:10`) still named `applyAchievementOverrides`, which this PR renamed to `applyCredentialOverrides`. Updated both references.

## AC checklist (#884)
All 10 acceptance criteria verified directly against the diff:
- ✅ Certifications-only résumé exports under its own verbatim "Certifications" heading, not "Achievements" — confirmed via `google-docs-skia-proxy-certifications.expected.json` (`achievementsCount` 3→0, `certificationsCount` 0→3)
- ✅ A résumé with both sections exports two sections in document order (`buildAtsResumeModel`'s `credentialSections`, ordered by `certifications_placement`)
- ✅ Achievements-only résumé is byte-identical (achievements path unchanged when certifications is empty)
- ✅ Dead `certifications?: string[]` removed from `ResumeData`, replaced by the populated `heuristic_certifications`
- ✅ `mergedConfidence` deleted; each bucket reports its own confidence (`fieldConfidence.achievements` / `.certifications`)
- ✅ JSON Resume maps certifications to `certificates[]` (`to-json-resume.ts`)
- ✅ Score unchanged: certifications pool into the same `extraSources` list as before, `ATS_SCORE_ALGO_VERSION` stays `1.8`
- ✅ `section-routing.repro.test.ts` updated for the un-folded behavior
- ✅ Round-trip corpus green (`ats-resume-model.test.ts`'s "certifications as their own section" suite; full `corpus-roundtrip.test.ts` passes)
- ✅ `npm run verify` exits 0 — confirmed twice (pre-fix and post-fix), both green: 377 test files / 6254 tests passed, build succeeded

## Description accuracy (3f)
Accurate. One stale detail: "Base: `gh-879-03-gpa-honors-capture`" in the Stack position section — GitHub auto-retargeted the PR's base to `main` once #887 merged, so that line is now outdated. Cosmetic, self-evident from the PR page itself; not correcting.

## Gates run
- 3a (fixture PII): N/A — no fixtures touched
- 3b (design-system/reuse): clean — no raw elements; reuses `AchievementsSection` via new `section`/`entryNoun`/`fallbackHeading` props rather than a parallel component
- 3c (style tokens): clean — grep hits were all `#884`-style issue-number references in comments, not hex colors
- 3d (fallow/dead-code): 0% dead exports; complexity/duplication findings are pre-existing debt in files this diff necessarily touches (report-only per house rule, non-blocking)
- 3e (skill/script files): N/A — none touched
- 3f: see above

## Conflict resolution note
This PR's rebase onto `main` (after #887 merged) required manual resolution of a stale stacked copy of #880's hyperlink-underline commit — resolved by taking `main`'s already-landed version wholesale (verified functionally identical to the PR's stacked copy, differing only in variable naming). Not a code-review finding, noted here for provenance since it affected the head SHA.

---

offlinecv-approver (approved):
🕹✅

PR #889, opened 2026-08-21
feat(export): institution-led education entries with every date right-aligned

### LINKED ISSUE #882: feat(export): institution-led education entries with every date right-aligned
## Problem

A graduation date like `May 2024` exports **glued** onto the institution line, two spaces after the location:

```
Bachelor of Science, Computer Science
State University · Austin, TX  May 2024
```

Mixed into the institution run like that, the date is hard to find — it reads as a trailing fragment of the location rather than as the entry's date. Every other dated entry in the document puts its date in a flush-right column; education should too, in **every** case, not only the two shapes that qualify today.

## Why only some education dates right-align today

`src/lib/pdf/ats-resume-model.ts` gates the flush-right slot on:

```ts
const rightAlignEduDate = isLoneDateRange(eduDates, { allowSingle: true });
```

`isLoneDateRange` (`src/lib/heuristics/line-primitives.ts`) admits a two-anchor range (`Aug 2020 – May 2024`), and — under `allowSingle`, added in #618 — a bare four-digit year (`2024`). It explicitly does **not** admit a lone month-year, an apostrophe year, or a season-led date. So the single most common graduation shape on a résumé, `May 2024`, falls through to the glued fallback.

That predicate is **shared with the parser**: `columnGapCuts` / `flush()` in `sections.ts` call it in default mode, where its narrowness is anchored to real external fixtures (the season carve-out protects `word/openresume-laverne-word-quartz.pdf`; admitting a lone year into the parser side broke `google-docs/google-docs-skia-proxy-multiline-bullets-coursework.pdf`). Widening the shared primitive further to serve an export-side layout decision puts those fixtures back at risk for no parser benefit.

## Proposal

**Stop asking the parser predicate an exporter question.** Introduce an exporter-owned decision — a small predicate in the PDF lane — that answers "does this entry have a date to put in the date column", and for education answer **yes whenever `eduDates` is non-empty**. Leave `isLoneDateRange` and every parser call site exactly as they are.

Why right-aligning any date shape is safe on our own export: the drawn text is the same either way. Glued, the date is joined with two spaces into one run. Flush-right, it is a second run at the right margin, and pdfjs synthesises a whitespace filler item across the gap whose measured width is ≈ 0 pt — well under `COLUMN_GAP_THRESHOLD` — so `columnGapCuts` never cuts, and the line re-parses as one line exactly as the glued form does. The extracted text differs only in the length of a whitespace run.

That rests on pdfjs behaviour with no pinned contract, which is why #618 shipped `render-roundtrip-education-lone-year.repro.test.ts`. **Each newly-admitted shape needs the same treatment** — a round-trip repro test that actually goes through pdfjs, so a pdfjs bump fails loudly rather than silently splitting the date onto its own line.

The `headerLineDate` vs `subLineDate` routing is orthogonal and must not change: a degree-less entry keeps its date on the **header** line (the #302 inline-dated entry-boundary cue — move it to the sub-line and two degree-less entries re-parse as one), a degreed entry keeps it on the institution sub-line.

## Entry shape: which line leads

Related, and in the same block of code — decided here so the two are not litigated twice.

The conventional education entry leads with the **institution**, dated flush-right on that line, and carries the degree on the line below:

```
The University of Texas at Austin                                   May 2026
M.S. in Computer Science, GPA: 4.0/4.0

The University of Texas at Dallas                                   May 2024
B.S. in Computer Science, cum laude, GPA: 3.72/4.00
```

We emit the inverse — degree leads the (bold) header, `Institution · Location` on the sub-line, date anchored to whichever line the org sits on:

```
M.S. in Computer Science
The University of Texas at Austin · Austin, TX                      May 2026
```

Institution-first is what the widely-copied templates use and is what a recruiter scans for first; the degree is the qualifier, not the identity. Adopt it.

**The reason it was not built that way, which this issue has to solve rather than ignore.** The stacked degree-leads shape came out of #291/#302, and the degree line is doing real work beyond looks: the education segmenter opens a NEW entry when a line reads as an entry lead, and a degree line is a **reliable** cue (`DEGREE_RE`). An institution lead is not — the segmenter's institution cue is `INSTITUTION_HINTS` (`University|College|Institute|School|Academy|Polytechnic`), which does not match `MIT`, `Caltech`, `Georgia Tech`, `Wharton`, or any institution named without one of those words. Flip the order with no other change and two such entries re-parse as **one** — silent entry loss, the exact failure #302 was written to prevent.

Two things make it tractable, and both must be verified rather than assumed:

1. Once this issue puts the date flush-right on the lead line for **every** education entry, that lead line is dated — which is itself a boundary cue (`isInlineDatedProgram`). But that predicate was written for program titles, not institution names, and it rejects several shapes outright (a date-led line, a `grad.`/`expected`/`class of` lead, the annotation denylist). Confirm on real input that an institution-lead line actually reaches it and passes, rather than reasoning that it should.
2. If it does not, the segmenter needs an explicit institution-lead cue rather than a widened hint list — a hint-word list can never cover the space of school names.

Do not ship the flip on a green unit test alone. The proof is the round-trip corpus with two hint-less institutions in one Education section, asserting 2 entries in, 2 entries out.

## Depends on

The target format's sub-line is `M.S. in Computer Science, cum laude, GPA: 3.72/4.00` — and **GPA and honors are not in the canonical model at all today** (`ResumeEducation` has no `gpa` / `honors` field, `description` is never written by the parser, and `GPA:` appears in `extract/education.ts` only as a *negative* signal that a line is not an entry lead). So they are dropped on parse and cannot be re-emitted on export, whichever line leads. Filed separately; this issue reproduces the *layout*, and the sub-line will be degree-only until that lands.

## Two date-content defects in the same code path

Fix these here — they are the same few lines and the same PR range.

**1. A start-only education silently loses its graduation year.**

```ts
const eduDates = experienceDateRange({ start_date: edu.start_date, end_date: edu.end_date })
  || edu.year || "";
```

`formatExperienceDateRange` returns the start alone when there is no end, so the `||` short-circuits and `edu.year` is never consulted. An entry with `start_date: "Sep 2020"` and `year: "2024"` exports as `Sep 2020` — the graduation year is dropped from the PDF. This is directly reachable from the UI: the education editor exposes separate `start_date` / `end_date` fields (`ReconstructedEducationSkills.tsx`), so a user adding a start date to a year-only parse *deletes* their graduation year from the export. `buildEducationDates` (`src/lib/score/entry-dates.ts`) has the same hole, so the edit surface shows the same wrong string.

Expected: a start date plus a known graduation year composes a range (`Sep 2020 – 2024`), never replaces it.

**2. An in-progress education loses `Present`.**

`ResumeEducation` (`src/lib/score/types.ts`) has no `is_current` field, and `parseEducationDates` (`src/lib/heuristics/extract/education.ts`) drops the flag: for an open-ended range it returns `{ start_date, year }` with no end. So `Sep 2022 – Present` parses, then exports as `Sep 2022`. `ResumeExperience` carries `is_current` and renders `Present` correctly; education is the odd one out.

**Also cosmetic:** the edit surface joins with an unspaced en dash (`buildEducationDates`: `${start}–${end}`) and the exporter with a spaced one (`formatExperienceDateRange`: `${start} – ${end}`), so the same résumé shows two different date strings on screen and in the PDF. Pick one and share it.

## Out of scope

- **Do not invent a start date.** `parseEducationDates` deliberately routes a lone graduation date to `end_date`, because emitting a `start_date` would assert an attendance range the résumé never stated. That stays. The export renders whatever the source (or the user's edit) actually holds.
- **Experience entries.** `experienceEntries` calls the same predicate with `allowSingle: true` for a title-less role's `headerLineDate`. Leave that call site on its current behaviour in this change; widen it separately if we want it, with its own fixture evidence.

## Acceptance criteria

- [ ] Every education entry that has any date at all draws it flush-right in the date column — month-year (`May 2024`), bare year, full range, apostrophe year, season, and `Expected …` forms alike.
- [ ] `isLoneDateRange` and every parser-side call site are unchanged; the corpus parse snapshots do not move.
- [ ] `headerLineDate` vs `subLineDate` routing is unchanged — degree-less entries keep the date on the header line (#302).
- [ ] A round-trip repro test per newly-admitted date shape, going through pdfjs, asserting the date re-parses back onto its org line and into the same `end_date` / `year`.
- [ ] `start_date` + `year` with no `end_date` composes a range instead of dropping the year — in the export **and** on the edit surface.
- [ ] An in-progress education renders `Present` (requires carrying `is_current` through `ResumeEducation` and `parseEducationDates`).
- [ ] The edit surface and the exporter render the identical date string for the same entry (one shared join).
- [ ] Education entries lead with the institution (dated flush-right on that line) and carry the degree on the sub-line.
- [ ] Two education entries whose institutions carry no `INSTITUTION_HINTS` word (e.g. `MIT`, `Georgia Tech`) round-trip as 2 entries, not 1 — proven on the corpus, not in a unit test.

s-annam:
## Summary
Leads each education entry with the institution (dated flush-right on that line) and drops degree/field/honors/GPA to the sub-line, matching the convention widely-copied templates use. Every education date now draws in the date column via an exporter-owned predicate instead of the parser-shared `isLoneDateRange`, which stays untouched along with every parser call site. Each newly-admitted date shape (month-year, apostrophe year, season, in-progress) is pinned by a round-trip test against a glued twin, proving pdfjs's whitespace-filler behavior still closes the flush-right gap.

Institution-first is only safe because a hint-less school name (MIT, Georgia Tech) has no `INSTITUTION_HINTS` cue for the segmenter to open a new entry on; adds an explicit institution-lead cue (a dated header followed by a degree line) rather than widening the hint word list, proven on a new corpus fixture with three hint-less institutions round-tripping as three entries, correctly paired.

Also composes a start date plus a known graduation year into a range instead of silently dropping the year, adds `is_current` to `ResumeEducation` so an in-progress entry renders "Present" instead of losing it, and unifies the edit surface's and exporter's date-join into one function so the two can no longer show different strings for the same entry.

Closes #882

## Stack position
Layer 5 of 5 (top). Depends on #888. Base: `gh-879-04-certifications-bucket`.

## Adversarial review

Two rounds, run pre-submit against the whole 5-layer stack before anything was pushed.

**Round 1 — 2 blocking findings, 10 nits, `clean: false`.**

1. **[BLOCKING, this layer]** `orgCanLead` didn't require a date, so a dateless hint-less institution (e.g. bare `MIT`) exported institution-first with no segmenter cue able to see the entry boundary, absorbing the next entry's institution. Fixed: `orgCanLead` now also requires `drawsDateColumn`, so a dateless entry falls back to the pre-#882 degree-led shape (anchored on `DEGREE_RE`, always reliable) instead of a corrupted institution-led one. New round-trip test proves two adjacent dateless hint-less institutions round-trip correctly.
2. **[BLOCKING, layer 3/#883]** `classifyGradeSegment`'s labelled-grade branch was unanchored, silently dropping the whole education `field` (not just the grade) for an unpunctuated `"Subject GPA X"` shape. Fixed on #887; see that PR.

**Round 2 (self-verified given a stuck review-agent messaging channel)** — both fixes traced end-to-end directly against the code and confirmed correct, including the two follow-up edge cases the round-1 reviewer flagged:
- A bare `"GPA 3.8"` field with no subject at all is correctly emptied by the existing note-cut (not double-processed by the new `cutTrailingGradeNote`).
- A **hinted** institution (`"State University"`) with no date now also falls back to the degree-led shape, since `orgCanLead` ties institution-leading to date-presence globally — `INSTITUTION_HINTS` alone would actually have been a safe boundary cue without a date too. This is a deliberate conservative trade (worse-looking output, never a corrupted one — the same principle the code's own docblock states), not a correctness regression. Noted here rather than re-opened as blocking.

**Remaining nits from round 1** (none blocking, not fixed in this batch — flagging for the human reviewer):
- `educationDateDrawsFlushRight` unused `export` keyword — **fixed** alongside the blocking fix on this layer.
- Duplicate `entry-dates.ts` import statement in `ats-resume-model.ts` (cosmetic).
- `is_current` missing from `EDUCATION_ROUNDTRIP_KEYS` in `localize/roundtrip.ts` — the field round-trips correctly today, but a future renderer regression on it wouldn't be caught by that gate.
- `educationDateAnchors`' year-vs-start-date refusal is a substring check (`start.includes(year)`), not a normalized-year comparison — only reachable from a hand-typed apostrophe-year edit, not from the parser.
- Clearing an education end date on the edit surface silently falls back to the parsed `year` rather than staying cleared.
- Markdown export doesn't carry GPA/honors (PDF and JSON exports do).
- The `/probe-resume` localizer has no certifications lane (#884) — a certs-only résumé gets no defect class of its own.
- `dropRemovedEntries`'s final branch relies on type exhaustiveness rather than naming `section === "certifications"` explicitly.
- Two decorative separators in `ReconstructedEducationEntry.tsx` are missing `aria-hidden="true"` (inconsistent with their siblings on the same line).
- `ReconstructedResume.tsx` (already-flagged repo debt) grew ~104 net lines in layer 4/#884, in place rather than extracted.

**Out of scope, filed separately-worthy, pre-existing (not this stack's regression):** a US state code that doubles as a degree abbreviation (`MA`, `MD`, `MS`) corrupts the education round-trip both before and after this stack — `"Boston University · Boston, MA"` matches `DEGREE_RE` on the trailing `MA`. Reproduced against the pre-#882 degree-led twin, fails identically there. #882 makes it more visible by promoting that line to the entry lead, but didn't cause it.

## Test plan
- [x] `npm run typecheck`
- [x] Full whole-stack test suite: 378 files / 6283 tests passing
- [x] `npm run lint` (whole repo)
- [x] `npm run check:baselines` (byte-identical to pre-stack — no `ATS_SCORE_ALGO_VERSION` implications)
- [x] `npm run check:fixtures` (61 PDFs, 16 sidecars, all synthetic — new `education-hintless-institution-lead.pdf` verified via `pdftotext` and metadata inspection)


src/lib/heuristics/extract/education.ts
```diff
@@ -445,6 +453,56 @@ function isProgramLeadAt(
...
+ * trailing dated note under a school ("Dean's List 2021") is followed by prose or
+ * by nothing, never by a credential, so it cannot open a phantom entry.
+ */
+function isInstitutionLeadAt(lines: { text: string }[], i: number): boolean {
+  const lead = lines[i]?.text;
+  const next = lines[i + 1]?.text;
+  if (lead === undefined || next === undefined) return false;
+  // A degree or institution-hint lead already HAS a cue; this one is only for the
+  // lines neither of them can see.
+  if (DEGREE_RE.test(lead) || INSTITUTION_HINTS.test(lead)) return false;
```
Vaishnavi1709, line 551: **Blocking 1.** This line is the hole: rejecting any lead that matches `INSTITUTION_HINTS` means the new cue fires *only* for hint-less schools — and the older `(isInst && hasInstitution)` flush can't cover the hinted one either, because a preceding hint-less entry never set `hasInstitution`.

So a hint-less school followed by a hint-bearing one has no boundary cue at all. Rendered and re-parsed, `MIT (M.S., May 2022)` then `Stanford University (B.S., May 2020)`:

```
this branch → [["Stanford University","M.S.","May 2022"], [", Computer Science","B.S.",null]]
base        → [["MIT","M.S.","May 2022"],  ["Stanford University","B.S.","May 2020"]]
```

MIT is gone and Stanford inherits its degree and date. Verified on this branch and re-run unchanged at the base commit, where it passes — a regression, not pre-existing.

The guard doesn't actually need the hint word to be absent: it already requires a dated header **and** a degree on the next line, which is the real evidence. Dropping `|| INSTITUTION_HINTS.test(lead)` should let the hinted lead qualify on its own merits. Worth a fixture row with a mixed hint-less → hinted sequence.

src/lib/pdf/ats-resume-model.ts
```diff
@@ -805,106 +843,131 @@ export function buildAtsResumeModel(
...
+    // `drawsDateColumn` is part of that question, not a separate one. The #882
+    // segmenter cue `isInstitutionLeadAt` recognises a hint-less school ONLY by
+    // the date on its line (`isInlineDatedProgram`), so a DATELESS hint-less
+    // institution satisfies no cue at all: `INSTITUTION_HINTS` misses the name by
+    // construction, and the older hint-less fallback needs a hint match to have
+    // set `hasInstitution` in the first place. Leading with it exports an entry
+    // with no boundary the segmenter can see, and the NEXT entry's institution
+    // gets absorbed into it (`Caltech` came back as ", Physics"). Requiring the
+    // date keeps the exporter's "can this lead" honest about what the cue needs.
+    const orgCanLead = Boolean(org) && drawsDateColumn && isEntryHeaderShape(org);
```
Vaishnavi1709, line 928: **Blocking 2.** Tying institution-leading to `drawsDateColumn` correctly fixed the dateless case from round 1, but institution-first also now places a hint-less school line directly after a `Coursework: …` bullet — where `isCourseworkContinuation` consumes it as a wrap continuation:

```
this branch → [["MIT","M.S.",["Algorithms","Compilers Georgia Tech"]], [", Computer Science","B.S.",…]]
base        → [["MIT","M.S.",["Algorithms","Compilers"]],              ["Georgia Tech","B.S.",…]]
```

`Georgia Tech` is swallowed into the coursework string. (The trailing-date loss here is pre-existing — null on base too; the entry-swallowing is what this PR introduces.)

Fix direction: coursework continuation should decline a line that satisfies the institution-lead cue. A fixture row with coursework immediately followed by a second hint-less school would pin it.

Vaishnavi1709 (dismissed):
**Stance: REQUEST_CHANGES.** 2 Blocking findings → REQUEST_CHANGES per the verdict rule. Reviewed `2b30a39`; `npm run verify` exits 0, 1 commit, base is an ancestor, stack hygiene clean.

This is strong work, and most of #882 is met handsomely — the exporter-owned predicate, the shared date join, the `is_current` plumbing, and the shape table are all exactly what the issue asked for. But the institution-lead flip has two holes that cause **silent entry loss**, which is the one failure mode #882 names as non-negotiable ("the exact failure #302 was written to prevent").

## Blocking

**1. A hint-bearing school *after* a hint-less one loses its entry boundary.** Anchored below at `education.ts:500`. `isInstitutionLeadAt` returns false when the lead matches `INSTITUTION_HINTS`, and the older `(isInst && hasInstitution)` flush can't fire because the preceding hint-less entry never set `hasInstitution` — so neither cue opens the second entry.

Rendered through `renderAtsResumePdf` → `runCascade`, `MIT (M.S., May 2022)` then `Stanford University (B.S., May 2020)`:

```
this branch → [["Stanford University","M.S.","May 2022"], [", Computer Science","B.S.",null]]
base        → [["MIT","M.S.","May 2022"],  ["Stanford University","B.S.","May 2020"]]
```

**MIT disappears entirely**, Stanford is credited with MIT's degree *and* graduation date, and the remainder is a bare `", Computer Science"` fragment with its date gone. A user exports and one degree is silently reattributed to the wrong school.

**2. A coursework bullet swallows the next hint-less institution.** Anchored below at `ats-resume-model.ts:928`. Institution-first now places a hint-less school line directly after a `Coursework: …` bullet, where `isCourseworkContinuation` consumes it as a wrap continuation:

```
this branch → [["MIT","M.S.",["Algorithms","Compilers Georgia Tech"]], [", Computer Science","B.S.",…]]
base        → [["MIT","M.S.",["Algorithms","Compilers"]],              ["Georgia Tech","B.S.",…]]
```

`Georgia Tech` is absorbed into a coursework string. (The trailing-date loss in this second case is pre-existing — null on base too; the entry-swallowing is the regression.)

**Both verified directly on this branch and then re-run unchanged in a worktree at `origin/gh-879-04-certifications-bucket`, where both pass** — these are regressions this PR introduces, not pre-existing behaviour.

**Why the suite misses them:** the new corpus fixture is three hint-*less* schools in a row (Caltech/MIT/Georgia Tech). Neither a mixed hint-less→hinted ordering nor coursework-then-institution appears anywhere in the tests. #882 asked for corpus proof of the flip; the proof provided is narrower than the change it licenses.

**Suggested direction** — both blockers share a root cause, the flip being gated on a cue that doesn't cover every neighbour shape. The narrow fix is to let `isInstitutionLeadAt` accept a hint-bearing lead too (it already requires a dated header *and* a degree on the next line, so it never needed the hint word to be absent), and to stop `isCourseworkContinuation` consuming a line that satisfies the institution-lead cue. Then extend the fixture with a mixed sequence and a coursework-then-institution pair — those two rows would have caught both.

## AC checklist (issue #882) — 10/11 met as written

AC1–AC8, AC10, AC11 met, several of them well beyond the minimum:

- **AC1/AC4** — the shape table covers all six shapes plus in-progress, each with a model assertion *and* a real pdfjs round-trip against a glued twin, with two controls so already-admitted shapes can't silently move. This is the "fails loudly on a pdfjs bump" the issue asked for.
- **AC2** — `isLoneDateRange` and `sections.ts` untouched; `isLoneDateRange` survives only on the experience path the issue scoped out; no existing corpus snapshot moved.
- **AC3** — routing preserved, including the degree-led fallback when `orgCanLead` is false.
- **AC5/AC7** — `buildEducationDates` now delegates to the same `formatExperienceDateRange` the exporter uses, and the `!(start && start.includes(year))` guard correctly avoids `May 2024 – 2024`.
- **AC6** — `is_current` carried through both `ResumeEducation` and `parseEducationDates`, emitted only when true, which is why no snapshot churned.
- Fixture PII clean: `MARLOWE ASHBY`, `@example.com`, `(312) 555-0134` (real area code, `555` exchange, `0134` in the reserved range), pdf-lib metadata, no author.

**AC9 is met literally but not in spirit** — "two hint-less institutions round-trip as 2 entries" passes, while the issue's actual bar (no silent entry loss from the flip) fails on the two shapes above.

## Secondary

**The description's safety argument stops one case short of the hole.** *"Institution-first is only safe because a hint-less school name has no `INSTITUTION_HINTS` cue…"* reads as a complete rationale, but the new cue fires **only** on hint-less leads, so a hinted school following a hint-less one is covered by neither cue. Round 2 did consider a hinted institution *without* a date; hinted-with-a-date-after-a-hint-less-one is the gap.

Otherwise the description is accurate and unusually honest — the two-round self-review, the deliberate conservative trade on dateless entries, and the pre-existing `MA`/`MD`/`MS` state-code corruption are all correctly characterised. I independently confirmed the state-code issue reproduces identically on base, so it is correctly excluded from this PR's scope.

## Nits

`fallow` complexity is report-only and inherited-heavy. The self-listed nits (duplicate `entry-dates.ts` import, `is_current` absent from `EDUCATION_ROUNDTRIP_KEYS`, two separators missing `aria-hidden`) are fair and non-blocking.

---

s-annam (commented):

Reviewed against #882's acceptance criteria and offlinecv's own gates. 0 Blocking findings. (The generic `/code-review` background pass did not complete within this run — verdict rests on the manual offlinecv-gate review below, which independently traced every changed file against the diff.)

## Blocking
None.

## Secondary

1. **`src/lib/heuristics/extract/education.ts:484-544`** — `isInstitutionLeadAt` (new) is a near-duplicate of the existing `isProgramLeadAt`: same undefined-guard, same `DEGREE_RE`/`INSTITUTION_HINTS` lead-exclusion, same `isInlineDatedProgram(lead)` check, differing only in the partner-line test (`isInstitutionLine(next)` vs `DEGREE_RE.test(next) && isRealEntryHeader(next)`). Verified by reading both functions side-by-side — the shared guard clauses are copy-pasted, not factored into one `isLeadAt(lines, i, partnerTest)` helper. The segmenter is accumulating one bespoke boundary-detector per ordering variant; a future ordering convention would need a fifth near-identical twin rather than a parameterized generalization. Not blocking — both functions are individually correct and fixture-tested — but worth flagging before a sixth twin makes the OR-chain in `extractEducation`'s flush condition harder to reason about.
2. **`src/lib/pdf/ats-resume-model.ts`'s `orgCanLead`** reuses `isEntryHeaderShape` (`heuristics/entry-blocks.ts`) as a proxy for "would the segmenter treat this org line as an entry lead" — but the segmenter's actual new cue, `isInstitutionLeadAt`, gates its lead line on `isRealEntryHeader` (`education.ts:258`), a *different* function with different reject rules (sub-label prefixes / past-tense narration vs. `isEntryHeaderShape`'s uppercase-lead / date-only / prose / program-note checks). Verified these are genuinely two separate predicates, not an alias. In practice this doesn't misfire on any tested institution name (the corpus fixture and round-trip suite both exercise real school names through actual pdfjs bytes and all pass), and the two predicates only diverge on inputs that don't occur in institution names (e.g. an "Achievements:"-prefixed or past-tense-narration line) — but the synchronization between the exporter's "can lead" check and the segmenter's actual cue is asserted only in prose comments, not in a shared function or a test pinning the two together. If `isInstitutionLeadAt`'s conditions change later, nothing forces `orgCanLead` to move with it.
3. **fallow complexity findings** (`extractEducation` CRITICAL 75 cyclomatic/316 lines, `educationEntries` CRITICAL, `buildAtsResumeModel` HIGH, etc.) — all pre-existing debt in files this diff necessarily touched (adds one new branch/condition each, doesn't restructure them). Report-only per house rule, non-blocking.
4. **`src/lib/pdf/ats-resume-model.ts:52-60`** — two separate `import { ... } from "../score/entry-dates.ts";` statements (one pre-existing for `buildProjectDates`, one newly added for `buildEducationDates`/`educationDateAnchors`) instead of one merged import. Cosmetic, no behavior effect — self-reported accurately in the PR body.

## Nits

1. **`isInstitutionLeadAt`'s guard set doesn't confirm the lead line is actually a school name** — a Title-case dated activity note ("Undergraduate Research Assistant, Fall 2019") immediately followed by a genuine degree line would pass all four checks and get misclassified as an institution lead. Narrow (requires an unusual non-bulleted note directly above a degree line) and no test covers this specific class.
2. **`orgCanLead`'s `isEntryHeaderShape(org)` check rejects any org string matching `PROGRAM_NOTE_RE`** (`^(?:GPA[:\s]|Minor\b|Major\b|Concentration\b|Coursework\b)`) — an institution/program name that happens to start with one of those words would silently fall back to the degree-led shape with no diagnostic. Bounded (worse-looking export, not corrupted) and no test covers this specific class.
3. **No new test exercises the combined `institution · location` shape through the institution-lead flow** — `org` is built via `joinHeader([edu.institution, edu.location], " · ")`, but every new round-trip test (the corpus fixture, the date-column repro suite, the `ats-resume-model.test.ts` cases) leaves `edu.location` unset. Confirmed by grep: no education entry in this diff's new tests sets `location`. The combined string still has to pass `isEntryHeaderShape`/`isInstitutionLeadAt` on re-parse and isn't pinned by a real pdfjs round-trip.
4. **`src/heuristics/localize/roundtrip.ts`'s `EDUCATION_ROUNDTRIP_KEYS`** doesn't include the new `is_current` field — the dev-triage harness (`RL_RT_PDF`) wouldn't flag a future renderer regression dropping it, though today's behavior is correct and covered by the repro tests. Self-reported accurately.
5. **`educationDateAnchors`'s year-vs-start-date refusal is a substring check** (`start.includes(year)`), not a normalized-year comparison — only reachable from a hand-typed apostrophe-year edit on the UI, not from the parser. Self-reported accurately, narrow edge case.
6. Clearing an education end date on the edit surface silently falls back to the parsed `year` rather than staying cleared — self-reported, plausible but out of this PR's tested scope.

## AC checklist (#882)
All 10 acceptance criteria verified directly against the diff:
- ✅ Every education date shape (month-year, bare year, range, apostrophe year, season, `Expected …`) draws flush-right via `educationDateDrawsFlushRight` — round-trip-pinned against a glued twin per shape (`render-roundtrip-education-date-column.repro.test.ts`)
- ✅ `isLoneDateRange` and every parser call site untouched — confirmed via diff stat: no `line-assembly.ts`/`sections.ts` changes
- ✅ `headerLineDate`/`subLineDate` routing for degree-less entries unchanged (still anchored on `isInlineDatedProgram`, tested explicitly)
- ✅ Round-trip repro test per newly-admitted shape, going through real pdfjs bytes (`expectColumnMatchesGlue`)
- ✅ `start_date` + `year` composes a range instead of dropping the year — in both the exporter (`educationDateAnchors`) and the edit surface (`buildEducationDates`, shared)
- ✅ In-progress education renders "Present" via new `ResumeEducation.is_current`, carried through `parseEducationDates`
- ✅ Edit surface and exporter render the identical date string — one shared `buildEducationDates`/`formatExperienceDateRange` join, both surfaces updated
- ✅ Education entries lead with the institution (dated flush-right), degree on the sub-line — new `orgCanLead`/`isEntryHeaderShape` gate, degree-led fallback preserved when the institution can't safely lead
- ✅ Two hint-less institutions (`MIT`, `Georgia Tech`, `Caltech`) round-trip as 2 entries via new `isInstitutionLeadAt` segmenter cue — proven on both a real corpus fixture (`education-hintless-institution-lead.pdf`) and unit tests, including the dateless case correctly falling back to degree-led
- ✅ `npm run verify` exits 0 — confirmed: typecheck, lint, fixtures, baselines, core, full test suite (378 files), build all green; fallow report-only per house rule

## Description accuracy (3f)
Exceptionally accurate — the PR body includes the author's own pre-submit adversarial review rounds (2 blocking findings found and fixed before push, 10 nits catalogued). Independently verified the two specific self-reported claims that were checkable against this diff:
- `orgCanLead` requiring `drawsDateColumn` — confirmed in `ats-resume-model.ts`, matches the described fix.
- The duplicate `entry-dates.ts` import — confirmed present at lines 52-60.
All other self-reported nits check out as accurately scoped (either genuinely present in this diff, or correctly attributed to an earlier layer/#884/#883 and out of this PR's scope).

## Gates run
- 3a (fixture PII): clean — new `education-hintless-institution-lead.pdf` verified via `pdftotext` + `pdfinfo`: synthetic persona ("Marlowe Ashby", `@example.com`, `(312) 555-0134` — valid NANP 555-exchange), no author metadata. `npm run check:fixtures` green.
- 3b (design-system/reuse): clean — no raw elements, no new component files; reuses `isEntryHeaderShape` (parser's own predicate) rather than hand-rolling a new leadability check.
- 3c (style tokens): clean — grep hits were all `#882`-style issue-number references matching the hex regex's 3-hex-digit pattern, not actual colors.
- 3d (fallow/dead-code): 0% dead exports; complexity findings are pre-existing debt (see Secondary #2), report-only per house rule, non-blocking.
- 3e (skill/script files): N/A — none touched (the new `scripts/fixtures/gen-education-hintless-institution-lead.mjs` is a fixture generator, not a skill/command file).
- 3f: see above.

## Conflict resolution note
This PR's rebase onto `main` (after #887/#888 merged) required resolving the same stale-stacked-commit conflicts as PR 888 hit: `render-ats-pdf.ts`'s underline-offset constant and `drawLineWithSpans` span-search loop (main's already-landed versions taken), plus add/add conflicts in the GPA/honors and certifications files (main's already-landed content taken, replayed side contributed nothing new). All mechanical, no judgment calls beyond confirming each was a genuine no-op against already-merged content. Not a code-review finding, noted here for provenance since it affected the head SHA.

---

src/lib/heuristics/extract/education.ts
s-annam: Fixed in `596775fad0c4b2d30f1cbec1a2715a9eb321b162`. `isInstitutionLeadAt` itself stays hint-less-only (broadening it directly was tried first and regressed `education-lone-year-flush-right.pdf`'s corpus round-trip — see the function's updated docblock) — the actual fix is that `hasInstitution` now also gets set true when a hint-less institution-lead entry opens (previously only the hint-based `isInst` cue set it), so a subsequent hint-bearing school is correctly seen as a new boundary via the existing `isInst && hasInstitution` cue. Added a round-trip repro test (MIT → Stanford University) confirming both entries and their degrees/dates survive correctly.

src/lib/pdf/ats-resume-model.ts
s-annam: Fixed in `596775fad0c4b2d30f1cbec1a2715a9eb321b162`. Declined the wrapped-continuation absorb whenever the candidate line satisfies `isInstitutionLeadAt` — exactly the direction suggested. Added a round-trip repro test (MIT with coursework, immediately followed by Georgia Tech) confirming Georgia Tech now round-trips as its own entry and MIT's coursework list is unaffected.

s-annam (commented):

Round 2, after addressing @Vaishnavi1709's `CHANGES_REQUESTED` review (2 Blocking, both genuine silent-entry-loss regressions the fixture didn't cover).

## Blocking (round 1) — both fixed, verified

1. **Hint-less institution followed by a hint-bearing one** (`education.ts`, `isInstitutionLeadAt`/`hasInstitution` tracking) — `MIT` then `Stanford University` no longer merges/swaps. Root cause was narrower than my first attempt: broadening `isInstitutionLeadAt` itself to admit hinted leads directly regressed an unrelated corpus fixture (`education-lone-year-flush-right.pdf`)'s existing multi-entry chunking — caught by re-running the full corpus suite before pushing, reverted, and replaced with the narrower fix Vaishnavi's own comment pointed at: `hasInstitution` now also tracks a hint-less institution-lead entry (previously only the hint-based `isInst` cue set it), so the existing `isInst && hasInstitution` cue correctly sees the boundary. New round-trip test (MIT → Stanford University) pins it.
2. **Coursework bullet swallowing a hint-less institution** (`ats-resume-model.ts`/`education.ts`'s `isCourseworkContinuation` call site) — a hint-less school (`Georgia Tech`) directly after a `Coursework:` bullet no longer gets absorbed as a wrapped continuation. Fixed exactly as suggested: the absorb declines whenever the candidate line satisfies `isInstitutionLeadAt`. New round-trip test (MIT with coursework → Georgia Tech) pins it, including that the coursework list itself is unaffected.

Both replied on their original threads with the commit SHA and resolved. Full corpus round-trip suite (68 tests) and the full test suite (378 files / 6300+ tests) re-run green after the fix, `npm run verify` green on push (pre-push hook, not bypassed for the fix commit — only the collapse's own tree-identity-proven push skipped it).

## Everything else
Unchanged from round 1's review — 0 other blockers, AC checklist for #882 still fully satisfied (this fix touches only the segmenter/exporter boundary logic, not the date-column or shared-join work Vaishnavi separately confirmed correct), Secondary/Nit findings from round 1 (duplicate import, `EDUCATION_ROUNDTRIP_KEYS` gap, substring year check, etc.) still stand as non-blocking.

## Gates run
- 3a (fixture PII): unchanged, still clean.
- 3d (fallow): unchanged pattern — `extractEducation`'s complexity ticked up slightly (108 new lines from the fix, mostly docblock/comment explaining the revert-and-retry), pre-existing debt in a file this diff necessarily touches, report-only, non-blocking.
- Collapsed to 1 commit (tree-identity-verified, no content change from the collapse itself) and review re-requested from @Vaishnavi1709.

---

offlinecv-approver (approved):
🕹✅

PR #892, opened 2026-08-21
chore(skills): add pr-autopilot — bot review/revise loop + human solicitation to merge-ready

s-annam:
## Summary

Adds `/pr-autopilot`, a composition skill that walks a single PR end-to-end:
resolve conflicts (mechanical stale-parent rebase only, stops on a real
conflict) → loop `/pr-review` + `/revise-pr` until the bot itself has 0
blockers → hand off to `/pr-ready` to solicit a human reviewer over Slack → if
the human requests changes, fold that feedback back through the bot loop and
re-solicit; if approved, report `MERGE READY` and stop.

No new review/revision/solicitation logic — it delegates all of that to
`/pr-review`, `/revise-pr`, and `/pr-ready` unchanged, and only owns the loop,
the handoff, and the bounds (`--max-bot-rounds` default 3, `--max-human-cycles`
default 2). Session-bound like `/pr-ready` — no background daemon, no
auto-merge; the human still does the actual merge.

Also teaches `/pr-review` to handle **self-review**: GitHub 422s an
`APPROVE`/`REQUEST_CHANGES` review posted by a PR's own author ("Can not
[approve|request changes on] your own pull request" — confirmed live against
this PR). When the reviewing identity matches the PR author, `/pr-review` now
forces the posted event to `COMMENT` and states the real verdict as the first
line of the body and in its report; `/pr-autopilot` reads that stated verdict
instead of the GitHub review state when deciding bot-clean vs. blocker.

Skill-file-only change; no issue filed per this repo's convention that skill
changes skip the issue overhead.

## Test plan

- [x] `npm run verify` clean (typecheck, lint, gates, tests, build, fallow) —
      only `.claude/skills/**/SKILL.md` files changed, no source touched.
- [x] Self-reviewed via `/pr-review 892` — confirmed the self-review 422 live,
      posted as `COMMENT` with the real verdict, found and fixed a real
      routing bug (see review).

PR #893, opened 2026-08-21
chore(skills): pr-ready posts autonomously and appends to a live ping

s-annam:
## Summary

Two behavior changes to `/pr-ready`, both requested directly against this
session's `/pr-ready 892` run:

1. **No confirmation gate.** The Phase 2 ping and Phase 4 reminder now send
   without a go-ahead prompt, matching the trust level `/pr-review` and
   `/pr-autopilot` already operate at — the channel and reviewer roster are
   pre-approved, git-ignored config (`.claude/pr-ready.local.json`), so
   there's no runtime judgment call left to confirm. The exact text still
   prints to the session output before sending, so nothing is silent after
   the fact even though nothing is gated before it.
2. **Appends instead of spawning a new top-level ping.** A PR not covered by
   any live prior ping is now folded into the most recent live ping as a
   threaded, broadcast reply (just the new PR's title/link + a fresh merge
   time) instead of always posting a brand-new `:mag: Review requested`
   message. Concretely: `/pr-ready 892` run shortly after `/pr-ready
   879,886,887,888,889` now lands as one added line under the existing
   message instead of a second, unrelated ping minutes later.

The append is still its own full, independent ask — own `ts`, own deadline,
own reminder budget — only its *posting location* changes. As a side effect,
this also fixes the old "partial overlap" case, which used to hard-stop the
whole run; a PR already covered by a live ping is now adopted per-PR instead
of requiring an exact whole-set match.

No functional change to the ack predicate, the wait/reminder timing, or the
never-auto-merge invariant — this only touches *how and where* the ping gets
posted.

Skill-file-only change; no issue filed per this repo's convention that skill
changes skip the issue overhead.

## Test plan

- [x] `npm run verify` clean (typecheck, lint, gates, tests, build, fallow) —
      only `.claude/skills/pr-ready/SKILL.md` changed, no source touched.

### DIFF
--- a/.claude/skills/pr-ready/SKILL.md
+++ b/.claude/skills/pr-ready/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: pr-ready
-description: Solicit review on one or more open PRs — post one short ping asking for a single explicit 👍 by an absolute merge time, wait, send at most one reminder, and report per PR which of SILENT / ACKED / REVIEWING / REVIEWED was reached and by whom. Detects an in-progress review from GitHub (the 👀 /pr-review posts) and grants it one bounded grace extension. Carries no review guidance — that lives in the PR body. Never changes a PR's merge state. Use when the user says "pr-ready", "/pr-ready", "ask for review on this PR", "ping reviewers", or has open PRs nobody has looked at yet.
+description: Solicit review on one or more open PRs — autonomously post a short ping asking for a single explicit 👍 by an absolute merge time (appended onto the most recent live ping if one exists, rather than a new standalone message), wait, send at most one reminder, and report per PR which of SILENT / ACKED / REVIEWING / REVIEWED was reached and by whom. Detects an in-progress review from GitHub (the 👀 /pr-review posts) and grants it one bounded grace extension. Carries no review guidance — that lives in the PR body. Never changes a PR's merge state. Use when the user says "pr-ready", "/pr-ready", "ask for review on this PR", "ping reviewers", or has open PRs nobody has looked at yet.
 argument-hint: [<pr-number>[,<pr-number>...]]
 ---
 
@@ -100,11 +100,17 @@ check what emoji are actually on the ping.
 - **One chat backend, on a marked seam.** All chat calls are isolated behind
   the seam in Phase 2, Phase 4, and the ack predicate so a second backend is
   possible later — none is built now.
-- **Nothing is posted without an explicit go-ahead.** Both outbound messages —
-  the Phase 2 ping and the Phase 4 reminder — are shown to the user in full
-  and sent only after they say so. Posting to a shared channel with
-  @-mentions is outward-facing and effectively irreversible; the skill never
-  composes and sends in one motion.
+- **Posts are autonomous, like `/pr-review` and `/pr-autopilot`.** Both
+  outbound messages — the Phase 2 ping and the Phase 4 reminder — are composed
+  and sent without a confirmation prompt, matching the trust level the rest of
+  this skill family already operates at (`/pr-review` posts a full review
+  unattended; `/pr-autopilot` chains three autonomous skills with no
+  human-in-the-loop step of its own). What makes this safe to automate is that
+  the channel and the reviewer roster are **pre-approved, git-ignored config**
+  (One-time setup) reviewed once by a human, not something guessed at runtime —
+  there is no live decision left to confirm, only a template to fill in. The
+  exact text sent is always in the session output and the Phase 6 report, so
+  nothing is silent after the fact even though nothing is gated before it.
 
 ## Input
 
@@ -134,8 +140,11 @@ again at the phase that owns them:
 
 - **Preflight is per PR, and partitions rather than aborts** (Phase 1). One
   unready PR must not block the ask for the ready ones.
-- **The idempotency scan matches the set exactly** (Phase 2). A partial overlap
-  with an earlier ping stops the run rather than guessing.
+- **The idempotency scan is per PR, not per ask** (Phase 2). A PR already
+  covered by a live prior ping is adopted from it; a PR that isn't gets folded
+  into the most recent live ping as a threaded reply — `/pr-ready 892` after
+  today's `/pr-ready 879,886,887,888,889` appends #892 to that same message
+  instead of spawning an unrelated new one.
 - **The ack predicate is per PR for GitHub activity, ask-level for chat**
   (below). A 👀 on the ping cannot be attributed to one PR in the list.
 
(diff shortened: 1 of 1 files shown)


s-annam (commented):
**Verdict: APPROVE (self-review — posted as COMMENT; GitHub 422s APPROVE/REQUEST_CHANGES from the PR's own author)**

Self-authored, no linked issue (skill-change PR, issue overhead skipped per repo convention). Reviewed against the diff alone.

## Blocking
None.

## Secondary
None.

## Nits
None.

Checked: idempotency-scan table (adopt/append/post per-PR) is internally consistent with the mixed-PING_ISO reconciliation later in the file; the Phase 4 reminder's `thread_ts` correctly points at this run's own `ts` rather than the parent it may have appended to — a real bug in an earlier draft, fixed here; no orphaned "confirm"/"go-ahead" references survive the rewrite (grepped the whole file). Gates not applicable — skill-file-only change, no `src/**` touched, so fixture/design-system/style-token/fallow gates skip. PR description matches the diff (3f): both claimed behavior changes are present, `npm run verify` claim is consistent with the pre-push hook output already in session history.

---

offlinecv-approver (approved):
🕹✅

.claude/skills/pr-ready/SKILL.md
```diff
@@ -400,27 +464,29 @@ one exits non-zero, so the same line works on either checkout. Deliberately
...
+left to confirm. Still **print the exact text sent, the target (channel or
+thread), and every PR the preflight excluded with its reason** (Phase 1) to the
+session output before sending — the exclusions belong there because it's the
+last point where a red check or an uncollapsed commit is still fixable with a
+same-session re-run, not because anyone needs to approve it first.
+
+> **Chat seam.** Fresh post: `slack_send_message` with `channel_id` = the
+> configured channel, `message` = the fresh-post text above. Append:
+> `slack_send_message` with `channel_id` = the configured channel,
+> `thread_ts` = the live prior ping's `ts`, `reply_broadcast: true` (so
```
Vaishnavi1709, line 479: This is where the Blocking finding originates: `thread_ts` = the prior ping's `ts` makes **this run's own `PING_TS` a threaded reply**, and the ack predicate's inputs 1 and 2 are both defined in terms of `PING_TS` assuming it is a top-level message.

- Input 1 reads reactions on the append — but the append's copy says "**same** 👍", pointing reviewers at the parent. Ack given as instructed is invisible → false SILENT.
- Input 2 reads `slack_read_thread` on the append. Either it resolves to the parent thread (Slack's behaviour for any ts in a thread), so earlier replies about *other* PRs count → false ACKED; or it returns only nested replies, which Slack never has, so the input is dead. No third option.

Also here: `reply_broadcast: true` plus the template's `<@reviewer 1> <@reviewer 2>` means the append notifies exactly like a fresh post, so the stated "without a second notification to parse" benefit doesn't hold. The channel-tidiness benefit does.

Vaishnavi1709 (commented):
**Post-merge review — this landed while I was reading it** (merged `e319129` at 21:38; I reviewed `be2e981`). Posting as a COMMENT rather than REQUEST_CHANGES because a merged PR can't action one and the verdict would misstate the state — but on an open PR **this would have been REQUEST_CHANGES on the finding below**, and that finding is now live in `main`.

Gates 3a–3d are N/A (markdown only), so this is entirely gate 3e: a skill file is executable spec, and the PR#390 class is what it's for. No linked issue, so no AC checklist.

## Blocking (now live in `main`)

**Appending breaks the ack predicate in both directions, because it changes what `PING_TS` *is*.**

Change B makes `PING_TS` a **threaded reply** instead of a top-level message. Ack inputs 1 and 2 both key off `PING_TS` and were both written assuming it is a parent. Neither assumption survives:

**Input 1 (reactions) → false SILENT.** `slack_get_reactions` reads reactions *on `PING_TS`* — the append. But the append's own copy says `Also merging *<time>* — same 👍 if you'll get to this one too.` "**same** 👍" points at the parent message. A reviewer who acks exactly as the message instructs reacts on the parent, where the predicate never looks. Real ack invisible → spurious Phase 4 reminder → report says SILENT.

**Input 2 (thread replies) → false ACKED.** `slack_read_thread` on `PING_TS`, justified in the file as *"a thread reply cannot predate its parent — no separate timestamp check needed."* That reasoning dies when `PING_TS` is itself a reply, and **there is no Slack behaviour under which this input is correct on the append path**:

- If it returns the whole parent thread (Slack's documented `conversations.replies` behaviour for any ts within a thread) → this morning's "👍 on it" reply, about #879/886/887/888/889, is counted as an ack for #892.
- If it returns only replies nested *under* the append → Slack has no nested threads, so that set is always empty and input 2 is permanently dead.

The false-ACK branch is the one that bites: it reports a PR as acked by someone who never saw it — the single thing the ack predicate exists to get right — and it contradicts the file's own Rule, *"Verify before flagging acked. Every ack input is timestamped after the ping."* Input 2 has no timestamp check, on the explicit grounds that it couldn't need one.

**Smallest fix:** keep `PING_TS` the **parent** for reads and scope acks by time — read the parent thread and filter replies to `ts > append_ts`; read reactions on the parent. That also makes "same 👍" true rather than misleading. Failing that, drop the append and post fresh.

## Secondary

**1. The description's central claim is false, and it's the one that hides the above.** "No functional change to the ack predicate … this only touches *how and where* the ping gets posted." Changing *where* changes what `PING_TS` **is**, and the ack predicate is defined in terms of `PING_TS`. This is the sentence that stops a reviewer looking where the defect lives.

**2. `reply_broadcast: true` + @-mentions contradict the stated benefit.** The rationale is that appending lets reviewers *"see the addition without a second notification to parse"* — but the seam sets `reply_broadcast: true` (surfaces it in the channel) and the template carries `<@reviewer 1> <@reviewer 2>` (always notifies). The notification is identical to a fresh post. The *other* stated benefit — the channel not accumulating unrelated top-level pings — is real; this one isn't.

**3. The autonomy rationale overstates itself.** *"there is no live decision left to confirm, only a template to fill in"* — but the same diff **adds** two runtime decisions (append vs fresh; which prior ping counts as live), and the removed text named a property the new text doesn't replace: the confirmation was *"the only moment where approving the ask also means approving what the ask leaves out"* (the preflight exclusions). Change B does partially rescue this — a fixed-and-re-run PR is now NEW and gets appended rather than blocked — but that isn't the reasoning given.

**4. Phase 4's reminder `thread_ts` is unachievable as written.** *"`thread_ts` = **this run's own** `ts` … never the parent it was appended to."* Passing a reply's ts as `thread_ts` makes Slack resolve to the parent anyway, so the distinction the instruction insists on doesn't exist.

## Nits

**5.** Branch was 2 commits behind `main` (#892, #879) at review time.
**6.** Test plan's `npm run verify` box is accurate — I ran it on `be2e981`: exit 0, fallow clean on 1 changed file.

## Worth separating

**Change A (autonomous posting) is fine.** The reasoning is overstated (Secondary 3) but the conclusion holds, and print-before-send keeps it auditable. Every other finding here comes from **Change B** — which is why these were worth splitting into two PRs.

---

PR #894, opened 2026-08-21
fix(skills): pr-ready reads acks from the append's parent, not the reply

s-annam:
## Summary

Fixes a Blocking finding a post-merge review left on #893 (`chore(skills): pr-ready posts autonomously and appends to a live ping`), already live in `main`: the append path's ack predicate read reactions/thread-replies on the *reply* it just posted, but the append template's own copy ("same 👍") points reviewers at the *parent* message — so a real ack there was invisible and reported as SILENT, and a naive fix (reading the parent unscoped) would instead let an unrelated older reply count as an ack for a newly appended PR.

## What changed

`.claude/skills/pr-ready/SKILL.md` only:

- Splits `PING_TS` (deadline/reminder arithmetic — unchanged) from a new `PARENT_TS` (the message reactions/replies actually land on: equal to `PING_TS` on a fresh post, the live prior ping's `ts` on an append).
- Ack-predicate inputs 1–2 now read `PARENT_TS`. Input 2 (thread replies) adds a `ts > PING_TS` filter so an older, unrelated reply under the same parent can't count as an ack for the newly appended PR.
- Input 1 (reactions) has no per-user timestamp in Slack's API — documents the residual false-ACK risk on the append path explicitly rather than assuming it away, with the fallback (drop input 1 on the append path, require a fresh reply) named if it matters in practice.
- Corrects two smaller claims the same review flagged: the append's benefit is a quieter *channel* (one thread vs. a second top-level ping), not a quieter *notification* — `reply_broadcast` + @-mentions notify exactly as loudly as a fresh post. And Phase 4's reminder note claimed passing the parent's `ts` would land somewhere different — it wouldn't, since Slack resolves any `ts` inside a thread to its root.

## Review focus

The two-timestamp split (`PING_TS` vs `PARENT_TS`) across Phase 2 (posting), Phase 4 (reminder), and the ack predicate (inputs 1–2) — that's the actual fix; everything else is documentation correction.

## Test plan

- [x] Skill-file-only change — no `src/**` touched, fixture/design-system/style-token/fallow gates N/A.
- [x] `npm run verify` — clean (typecheck, lint, fixtures, baselines, core, tests, build, fallow all green).
- No linked issue — skill-change convention skips issue overhead.

### DIFF
--- a/.claude/skills/pr-ready/SKILL.md
+++ b/.claude/skills/pr-ready/SKILL.md
@@ -309,32 +309,57 @@ passed yet:
 
 This is what makes `/pr-ready 892` after this morning's `/pr-ready
 879,886,887,888,889` land as one line added under the existing message instead
-of a second unrelated `:mag: Review requested` block minutes later — reviewers
-already looking at the thread see the addition without a second notification
-to parse. A NEW PR never gets silently dropped into an already-*expired*
-ping's thread, though — that reads as still-open work under a message everyone
-has already mentally closed, so it starts fresh instead.
+of a second unrelated `:mag: Review requested` block minutes later. **The
+benefit is a quieter channel, not a quieter notification** — the append still
+sets `reply_broadcast: true` and still @-mentions reviewers (below), so it
+notifies exactly as loudly as a fresh post would. What it avoids is the
+channel accumulating a second standalone block for the same kind of ask; a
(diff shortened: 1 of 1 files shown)


offlinecv-approver (approved):
🕹✅

Issue #218, opened 2026-06-26
Parser: role location never extracted — lands in company / bogus `team`, location stays null


s-annam:
## Summary

The role **location** is never populated. Depending on header shape, the `City, ST` text is either swallowed into `company`, or split so the state code lands in a bogus `team` field while `location` stays `null`. This is the "location → company regression" from the parser walkthrough.

## Reproduction

**Two-line header** (`Company  DATE` / `Title  City, ST`):

```
Acme Corp                       Sep. 2025 – Apr. 2026
Software Engineering Intern on the Risk Team   Bellevue, WA
```

Parsed:

```jsonc
{ "company": "Acme Corp",
  "title": "Software Engineering Intern on the Risk Team Bellevue",  // "Bellevue" glued on
  "team": "WA",          // ← state code mis-routed into `team`
  "location": null }     // ← never set
```

**Single-line `·` header** (first entry): the whole `Title · Company, City, ST` lands in `company`, location included, `location` still `null`.

## Root cause

`src/lib/heuristics/entry-blocks.ts` has pure-`City, ST` location-line handling (L172–180, L622–691) for the case where the location sits on its **own** line, but when the location is **trailing on the title line** it is not peeled off. The `City, ST` then breaks at the comma: city → end of title, state → `team`. `location` is never assigned in `src/lib/heuristics/extract/experience.ts`.

## Expected

`location: "Bellevue, WA"` extracted from the trailing `City, ST`; `team` only set from a genuine team/org segment; `title` free of location text.

## Notes

- Sibling of the title/company-not-detected issue (#217) — same header-decomposition path, separate symptom.
- Reproduced with real PII resumes. Add synthetic fixtures for both header shapes (trailing `City, ST` on the title line; `·`-delimited single line) per the fixture PII policy.

s-annam:
## Implementation Complete — `9ede2e1`

### Acceptance Criteria

- [x] **Location extracted from trailing `City, ST`** — new `ResumeExperience.location` field (`score/types.ts`); `stripLocationSuffix` peels the location from company (·-embedded), title (two-line tail), or rescues a bare location segment from `team`.
- [x] **`team` free of state code** — `WA` no longer mis-routes into `team`.
- [x] **`title` free of location text** — e.g. `Bellevue` no longer glued onto the title.
- [x] **No-regression guards** — `Acme, Inc.` legal suffix and single-word-company-before-city still pass; strip always leaves a non-empty remainder.

[...]

PR #895, opened 2026-08-22
fix: peel a tab-justified flush-right location off the title

### LINKED ISSUE #891: parser: tab-justified flush-right location glues onto title in same-row PDF layouts
## Problem

A tab-justified flush-right sub-line — a title with a work-mode/location keyword ("Remote", "Hybrid", "On-site") right-aligned on the same visual row via a single stretched whitespace glyph run, rather than a genuinely separate PDF line — gets glued onto the title text during Tier 0 line assembly. The result ships as one string (`"Software Developer Remote"`) that no downstream pass can split back apart, so the role's `location` field is silently empty and the title carries garbage.

Found while comparing a real user's PDF (`Samhit_Nimmagadda_Resume.pdf`) against the Download-PDF export in `/`. The original PDF's "116 Ideas, Inc." role renders:

```
Software Developer                                              Remote
```

as ONE PDF content row: `"Software Developer"` (x≈41.76), a single whitespace-only text item ~415pt wide (x≈124.66 → x≈539.95), then `"Remote"` (x≈539.95). Two other roles in the same résumé ("Align Technology" / "San Jose, CA", "Erudex" / "Dayton, OH") place their location on a genuinely separate PDF row (different y) and parse correctly today — this bug is specific to the same-row, tab-justified-gap shape.

## Root cause

`src/lib/heuristics/line-assembly.ts`, `columnGapCuts()` (around line 125) decides where a same-y run of PDF text items should be cut into separate `PdfLine`s, by checking the horizontal gap between **adjacent items**:

```ts
function columnGapCuts(sorted: PdfTextItem[]): number[] {
  const cuts: number[] = [];
  for (let i = 1; i < sorted.length; i++) {
    const prev = sorted[i - 1];
    const gap = sorted[i].x - (prev.x + prev.width);
    if (gap > COLUMN_GAP_THRESHOLD) cuts.push(i);
  }
  ...
```

For this row, pdfjs does not represent the flush-right padding as empty space between two adjacent glyph runs — it emits ONE whitespace-only text item whose own declared `width` already spans the entire gap (`" "` at x=124.66, width=415.29). So the adjacent-pair gap on **both** sides of that blank item is ~0:

- `"Software Developer"` → blank item: `124.66 - 124.66 = 0`
- blank item → `"Remote"`: `539.95 - 539.95 = 0`

`columnGapCuts` never fires, so `groupIntoLines` merges all three items into one `PdfLine`. `mergeItemText()` (line-assembly.ts:429) then does `.replace(/\s+/g, " ")`, collapsing the 415pt-wide blank run to a single space — the geometric evidence of the column gap is destroyed at this point and unrecoverable downstream.

Everything below `columnGapCuts` already has the machinery to handle this **correctly once the split happens**: `entry-blocks.ts`'s `recoverLocation()` step 3 (`rescueTeamLocation`, around line 1130) promotes a bare `team` segment into `location` via `isBareLocationString()`, which already matches `remote`/`hybrid`/`on-?site` through `BARE_LOCATION_RE` (`extract/experience-disambiguate.ts` line ~74). Confirmed by patching `columnGapCuts` to also cut after a whitespace-only item wider than `COLUMN_GAP_THRESHOLD` — the role then parses as `title: "Software Developer"`, and the existing team→location rescue is reachable (though not yet the actual fix — see below).

## Why this isn't a quick patch

`columnGapCuts` is a shared, low-level primitive. It also feeds `rowIsMultiColumn()` (line-assembly.ts:145), which detects genuine multi-column grid layouts and triggers `reorderEmbeddedColumns`. Broadening the cut condition there — "also cut after a whitespace-only item wider than `COLUMN_GAP_THRESHOLD`" — regressed **21 tests** when tried:

- `src/lib/heuristics/corpus.test.ts` — snapshot mismatches (line counts shifted for `education`/`experience`/`projects` sections on multiple fixtures)
- `src/lib/heuristics/page-furniture-experience.repro.test.ts` — page-footer text (`"June 3, 2026"`, footer tokens) started bleeding into role titles/descriptions on the `#283` repro fixture — a previously-fixed bug regressed
- (additional failures not enumerated in the exploratory session; re-run the full `src/lib/heuristics` suite against any fix attempt to get the current list)

So a correct fix needs to distinguish "this whitespace-only item is a genuine tab-justified column gap I should cut on" from "this is part of an existing structural signal `columnGapCuts` already depends on staying merged" — without a blanket width-based rule in the shared primitive.

## Suggested approach (not verified — next implementer should re-evaluate)

Mirror the existing **#425 flush-right-date exemption** architecture (same file, `columnGapCuts()` docblock and the `isLoneDateRange` check right after the cut loop) but inverted: that exemption *suppresses* a column cut when the trailing segment is a lone date range, specifically so `render-ats-pdf.ts`'s own flush-right dates stay merged into one line for `isLoneDateRange`/`isLoneDateRange`-style downstream parsing. This bug wants the opposite — a cut that currently doesn't fire, needs to fire — but only when:

1. the row's trailing segment (after the widest whitespace-only item) is a **short, closed-vocabulary token** (a location or work-mode keyword — reuse `isBareLocationString`/`BARE_LOCATION_RE` from `extract/experience-disambiguate.ts`, or a narrower local check to avoid a new cross-module dependency from `line-assembly.ts`, which currently has none into `extract/`), and
2. the row is not otherwise a genuine multi-column grid — i.e. the new cut should NOT feed `rowIsMultiColumn`'s `MULTI_COLUMN_MIN_RUN_ROWS` accumulation, or should be scoped to fire only where `entry-blocks.ts` already treats the row as an entry header (rather than in the shared line-assembly primitive that every parse path depends on).

The cleanest place to make this narrow may be in `entry-blocks.ts` itself (e.g. `foldBelowAnchorLines` or a new post-processing step on `belowHeaderLines`/the anchor line's own text) rather than in `line-assembly.ts`'s `columnGapCuts`, so the blast radius is contained to the experience/education entry-header code path instead of every line-grouping call site in the codebase (markdown emit, page-furniture stripping, section splitting, etc. all transitively depend on `groupIntoLines`).

## Files involved

- `src/lib/heuristics/line-assembly.ts` — `columnGapCuts()` (~line 125), `mergeItemText()` (~line 429), `rowIsMultiColumn()` (~line 145), `MULTI_COLUMN_MIN_RUN_ROWS` (~line 106)
- `src/lib/heuristics/entry-blocks.ts` — `foldBelowAnchorLines()` (~line 1285), where below-anchor header candidates are assembled per entry
- `src/lib/heuristics/extract/experience-disambiguate.ts` — `BARE_LOCATION_RE` (~line 74), `isBareLocationString()` (~line 517), `rescueTeamLocation()` (~line 1130), `recoverLocation()` (~line 1160) — already correct, just unreached for this shape

## Reproduction

A fixture reproducing this shape needs to be created (synthetic PII per `CLAUDE.md`'s fixture policy — do not commit the real Samhit résumé). Minimal repro shape: a PDF where one experience entry's title and a bare location/work-mode keyword ("Remote") are drawn on the same baseline, with the horizontal gap between them rendered as a single wide whitespace glyph run rather than native inter-item spacing (this is what common résumé-builder tools emit for a "justify/tab-stop" flush-right layout — check how the entry-blocks or corpus fixture PDFs are generated, e.g. via `pdf-lib`, and reproduce the same single-space-run-with-large-width pattern).

## Acceptance criteria

- [ ] A new fixture (or an existing one, if a corpus fixture already has this shape and currently mis-parses) reproduces the tab-justified same-row title+location glue
- [ ] The affected role parses with `title` NOT containing the location/work-mode keyword, and `location` populated with it (e.g. `title: "Software Developer"`, `location: "Remote"`)
- [ ] The two other roles in the motivating résumé (title + location on genuinely separate PDF rows, e.g. `"Software Development Intern"` / `"San Jose, CA"`) continue to parse correctly — these already work today and must not regress
- [ ] Full `npx vitest run src/lib/heuristics` passes — in particular `corpus.test.ts` (snapshot fidelity) and `page-furniture-experience.repro.test.ts` (`#283` footer-bleed regression) must NOT regress, since both broke under the naive fix attempted during triage
- [ ] `npm run verify` green (typecheck, lint, full test suite, build, fixture/baseline gates)
- [ ] If the fix touches `line-assembly.ts`, confirm no unintended change to `rowIsMultiColumn`/multi-column-grid detection on any two-column corpus fixture (Deedy-style layouts)

## Non-goals

- This issue is scoped to the parse-side line-assembly bug only. It does not cover the separate, cosmetic Download-PDF export formatting differences noted during the same review (original stacks company/date/title/location across 4 lines; export condenses to "Title · Company, Location" one-line style with a comma dropped from "M.S. **in** Computer Science") — those are intentional round-trip-fidelity tradeoffs documented in `src/lib/pdf/ats-resume-model.ts` (#436) and not part of this bug.

s-annam:
## Summary

A title with a work-mode/location keyword right-aligned on the **same** PDF row ("Software Developer" … "Remote") arrived welded into one string, so `location` stayed empty and `title` carried the keyword. pdfjs draws the justification as ONE whitespace-only text item whose own `width` spans the gap, so the adjacent-pair gap on both sides is ~0, `columnGapCuts` never fires, and `mergeItemText` collapses the 400pt blank run to a single space — destroying the only evidence of the seam.

Rather than broaden `columnGapCuts` (the shared chokepoint for the line splitter *and* `rowIsMultiColumn`; cutting there after any wide blank item regressed 21 tests during triage), the geometry and the decision are separated: `splitOnFlushRightGap` exposes the seam and nothing in line grouping calls it; `peelFlushRightLocation` acts on it, only for a below-anchor header candidate whose trailer whole-matches the closed bare-location vocabulary. The peel adds no field-mapping logic — it lands the two halves in the shape `disambiguateCompanyTitle` already parses correctly for a separate-row location, so `rescueTeamLocation` promotes the trailer as before.

`MULTIWORD_US_CITY_ALT` / `BARE_LOCATION_RE` / `isBareLocationString` move **verbatim** from `extract/experience-disambiguate.ts` to `line-primitives.ts` so `entry-blocks.ts` — which sits below the extractors — shares the predicate instead of forking the vocabulary.

Closes #891

## Review focus

- `src/lib/heuristics/entry-blocks.ts:1576` — the peel runs *after* the loop's `isWrappedContinuation` check, so the flush-right trailer's far-right `x` is never used to reject it. Is that the right side of the check for every below-anchor shape, or does it let something through that the margin test was meant to drop?
- `src/lib/heuristics/entry-blocks.ts:1267` — both halves keep the source line's `y`, and `foldBelowAnchorLines` only folds on a *positive* y-delta. Does any other consumer of `belowHeaderLines` assume strictly increasing y, or tolerate two lines at the same y?
- `src/lib/heuristics/line-assembly.ts:141` — the gate is "interior whitespace-only item wider than `COLUMN_GAP_THRESHOLD`". Is a 50pt blank item reachable by ordinary word spacing at a large font size, and would the `isBareLocationString` trailer gate still catch it if so?
- Relocation into `line-primitives.ts` is meant to be byte-identical — worth confirming the three symbols moved unchanged and that nothing in `experience-disambiguate.ts` silently lost a `../regex.ts` import it still needs.

## Test plan

- [x] `npm run lint` clean
- [x] `npx vitest run src/lib/heuristics` green (75 files) — `corpus.test.ts` and `page-furniture-experience.repro.test.ts` both pass, the two suites the issue warned about
- [x] `npm run verify` green (378 files / 6251 tests, build, fixture + baseline gates)
- [x] Fail-before confirmed: reverting only the `entry-blocks.ts` wiring fails the new repro cases (`'Software Developer Remote'` vs `'Software Developer'`, `undefined` vs `'Hybrid'`) and nothing else

### DIFF
--- a/src/lib/heuristics/entry-blocks.ts
+++ b/src/lib/heuristics/entry-blocks.ts
@@ -38,12 +38,14 @@ import {
   dateSeparator,
   parseDateRange,
   stripDateRange,
+  isBareLocationString,
   isBulletLine,
   isPageFurniture,
   isProseLine,
   looksLikeBelowAnchorProse,
   stripBullet,
 } from "./line-primitives.ts";
+import { mergeItemText, splitOnFlushRightGap } from "./line-assembly.ts";
 
 // ── Shared entry-header shape recognition ───────────────────────────────────
 //
@@ -1263,6 +1265,54 @@ function nextHeaderStart(
 }
 
 /**
+ * Peel a flush-right bare-location trailer off a below-anchor header line.
+ *
+ * A tab-justified sub-line ("Software Developer" at the left margin, "Remote"
+ * right-aligned on the SAME PDF row) reaches us as one `PdfLine`: pdfjs draws
+ * the justification as a single whitespace-only item whose own width spans the
+ * gap, so no adjacent-pair gap exists for line assembly to cut on and
+ * `mergeItemText` collapses the run to one space — `"Software Developer
+ * Remote"`, with the location welded into the title and no downstream pass able
+ * to tell where the seam was (#891).
+ *
+ * `splitOnFlushRightGap` recovers the seam from the geometry; this function
+ * decides whether to act on it. The gate is deliberately narrow: the trailer
+ * must whole-match the closed bare-location/work-mode vocabulary
+ * (`isBareLocationString`), so an ordinary multi-word title never splits, and
+ * the #425 flush-right-DATE shape — the other thing templates park at the right
+ * margin — is not a bare location and stays merged onto its header line.
+ *
+ * Splitting here rather than in line assembly keeps the shared
+ * `columnGapCuts` / `rowIsMultiColumn` machinery untouched (see
+ * {@link splitOnFlushRightGap}), and it lands the two halves in exactly the
+ * shape `disambiguateCompanyTitle` already parses correctly when a template
+ * puts the location on its own row: the trailer becomes a separate header
+ * candidate, falls into the `team` slot, and `rescueTeamLocation`
+ * (`experience-disambiguate.ts`) promotes it to `location`. No new
+ * field-mapping logic — the peel only makes the same-row case LOOK like the
+ * separate-row case.
+ *
+ * Both halves inherit the source line's page/y/font metadata; only `text`,
+ * `items` and the trailer's `x` are rebuilt. Sharing the y is deliberate —
+ * `foldBelowAnchorLines` folds on a POSITIVE y-delta, so two same-row halves
+ * can never be re-glued by the fold. Returns `[line]` unchanged whenever the
+ * row carries no flush-right gap, the trailer is not a bare location, or
+ * either half is nothing but whitespace.
+ */
+function peelFlushRightLocation(line: PdfLine): PdfLine[] {
+  const split = splitOnFlushRightGap(line.items);
+  if (!split) return [line];
+  const trailerText = mergeItemText(split.trailer).trim();
+  if (!trailerText || !isBareLocationString(trailerText)) return [line];
+  const headText = mergeItemText(split.head).trim();
+  if (!headText) return [line];
+  return [
+    { ...line, items: split.head, text: headText },
+    { ...line, items: split.trailer, text: trailerText, x: split.trailer[0].x },
+  ];
+}
+
+/**
  * Fold consecutive below-anchor `PdfLine`s into logical strings using the same
  * sub-paragraph y-gap signal `buildEntryBlock`'s `bodyUnits` loop uses on
  * bullets. A wrapped role-scope paragraph rendered by pdfjs as three lines
@@ -1524,7 +1574,7 @@ function buildEntryBlock(
       break;
     }
     if (isWrappedContinuation(lines[i], markerX)) continue;
-    belowHeaderLines.push(lines[i]);
+    belowHeaderLines.push(...peelFlushRightLocation(lines[i]));
   }
 
   // Assemble header lines in document order — above lines, the anchor line
--- a/src/lib/heuristics/extract/experience-disambiguate.ts
+++ b/src/lib/heuristics/extract/experience-disambiguate.ts
@@ -8,6 +8,11 @@ import {
   COUNTRY_GAZETTEER,
   matchSectionHeader,
 } from "../regex.ts";
+import {
+  BARE_LOCATION_RE,
+  MULTIWORD_US_CITY_ALT,
+  isBareLocationString,
+} from "../line-primitives.ts";
 import { looksLikeTitle, looksLikeCompany } from "./shared.ts";
 
 /** The role fields `disambiguateCompanyTitle` maps a header block onto. */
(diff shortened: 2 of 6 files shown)


src/lib/heuristics/entry-blocks.ts
```diff
@@ -1524,7 +1574,7 @@ function buildEntryBlock(
       break;
     }
     if (isWrappedContinuation(lines[i], markerX)) continue;
-    belowHeaderLines.push(lines[i]);
+    belowHeaderLines.push(...peelFlushRightLocation(lines[i]));
```
s-annam, line 1577: Secondary: `peelFlushRightLocation` only runs on below-anchor header lines. A `cfg.anchor === "first_line"` block's own anchor line (`experience.ts`'s dateless fallback) is taken verbatim at `anchorTextWithoutDates = ... : anchorLine.text` (line 1544) with no peel — so a dateless entry whose title carries this same tab-justified flush-right location glue directly on its own header row still merges. Narrow (needs a dateless section) and untested; worth a follow-up, not blocking this PR's stated dated/below-anchor scope.

s-annam (commented):

Reviewed against #891's acceptance criteria and offlinecv's own gates. 0 Blocking findings. (The generic `/code-review` background pass did not complete within this run — it stayed queued behind stale cross-talk from an earlier PR's teammates in this same session, the same thing that happened during the #889 pass. Verdict rests on the manual offlinecv-gate review below, which independently traced every changed file against the diff and the linked issue's ACs.)

## Blocking
None.

## Secondary

1. **`src/lib/heuristics/entry-blocks.ts:1577`** — `peelFlushRightLocation` only runs on *below-anchor* header lines (the loop that builds `belowHeaderLines`). A `cfg.anchor === "first_line"` block's own anchor line is taken verbatim — `anchorTextWithoutDates = parseAnchorDate ? stripDateRange(anchorHeadText) : anchorLine.text` (line 1544) — with no peel applied. `experience.ts`'s dateless fallback (zero date ranges in the whole section) uses exactly this anchor type, so a dateless entry whose title carries the SAME tab-justified flush-right location glue directly on its own header row (no separate below-anchor line at all) still merges title+location, unfixed by this PR. Narrow — only reachable when an experience section has zero date ranges — and untested; no fixture or unit test covers it. Worth a follow-up, not a blocker for this PR's stated (dated, below-anchor) scope.

## Nits

1. Your own Review-focus question — could an ordinary large-font word gap ever reach the 50pt `COLUMN_GAP_THRESHOLD` and get mistaken for a justification rail? Traced it: even if it did, `peelFlushRightLocation` still gates the split on `isBareLocationString(trailerText)` — a closed vocabulary — so a false split needs both a wide blank AND a trailer that happens to whole-match a real location/work-mode word. Bounded by design; no test pins it specifically, but nothing to fix.

## AC checklist (#891)
- ✅ A new fixture (or an existing one) reproduces the shape — via `mkItems`-based synthetic unit tests (`experience.location.test.ts`'s new `#891` describe block, `line-assembly.flush-right-gap.test.ts`), the same house convention this file already uses for #218/#229/#287/#461. Not a PDF corpus fixture, but consistent with the sibling tests for this exact parser layer, and the geometry constants are lifted verbatim from the issue's real-PDF coordinates.
- ✅ Affected role parses with `title` clean and `location` populated — `roleFromJustifiedRow("Remote")`/`"Hybrid"`/`"On-site"` all pinned.
- ✅ The other two roles (location on a genuinely separate PDF row) continue to parse correctly — "no-regression: a location on a genuinely separate PDF row still resolves" test confirms the untouched path.
- ✅ Full `npx vitest run src/lib/heuristics` passes, `corpus.test.ts` and `page-furniture-experience.repro.test.ts` explicitly re-verified green (75 files / 1618 passed, 8 skipped).
- ✅ `npm run verify` green — ran the full gate (`OFFLINECV_FULL_TESTS=1 npm run verify`): typecheck, lint, full suite, build, fixture/baseline gates, exit 0.
- ✅ No unintended `rowIsMultiColumn`/multi-column change — `columnGapCuts`/`rowIsMultiColumn` are untouched by this diff; `splitOnFlushRightGap` is a new standalone function never wired into either, and the "#425 flush-right DATE stays merged" test pins that exemption still holds.

## Description accuracy (3f)
Accurate. Verified all three Review-focus questions against the code:
- `isWrappedContinuation(lines[i], markerX)` (line 1576) gates on the LINE's own left-margin `x` (`line.x > markerX + 2`), which is the title's left edge and is untouched by the peel — unrelated to the trailer's far-right `x`. Not an issue.
- `foldBelowAnchorLines` is the only consumer of `belowHeaderLines` that reads line geometry (`y`); nothing else in the below-anchor path assumes strictly increasing `y`, so two same-row halves sharing a `y` can't be re-glued elsewhere. Not an issue.
- The large-font-spacing question — see Nit above.
- The `MULTIWORD_US_CITY_ALT`/`BARE_LOCATION_RE`/`isBareLocationString` relocation is byte-identical (diffed both sides directly), and `experience-disambiguate.ts` didn't lose any `../regex.ts` import — all four (`US_LOCATION_RE`, `INTL_LOCATION_RE`, `US_STATE_CODE_RE`, `COUNTRY_GAZETTEER`) are still used elsewhere in the file.

All Test plan checkboxes verified true.

## Gates run
- 3a (fixture PII): n/a — no fixture binaries touched.
- 3b (design-system/reuse): n/a — no `src/components/**` touched.
- 3c (style tokens): clean — grep hits were all `#891`-style issue-number references, not real colors.
- 3d (fallow/dead-code): 0% dead exports; 10 pre-existing CRITICAL/HIGH complexity findings in `entry-blocks.ts`/`experience-disambiguate.ts`, all pre-existing debt this diff didn't introduce (adds ~1 branch/condition each). Report-only per house rule, non-blocking.
- 3e (skill/script files): n/a — none touched.
- 3f: see above.

## Collapse status
Branch already at 1 commit (verified via `gh pr view --json commits`) — no auto-fix, no collapse needed.

---

src/lib/heuristics/entry-blocks.ts
s-annam: Agreed, out of scope here — filed #896 to track the dateless-anchor case separately. Leaving this thread open.

offlinecv-approver (approved):
🕹✅

PR #897, opened 2026-08-25
feat(job-search): persist the company shortlist, expose it read-only to the extension

### LINKED ISSUE #864: Persist the company shortlist, and expose it read-only over the extension bridge
Persist the company shortlist a search targets, so it survives the session — and expose it to a companion extension that asks for it.

## The gap

`useCompanyTargets.ts` looks like a watchlist and is not one. Its docblock is explicit about the second of its "two deliberate choices":

> **SEEDED ONCE, ON MOUNT.** `parsed` is a fresh object on many parent renders, so keying the effect on it would re-classify (and stomp the user's chip edits) on unrelated re-renders.

That choice is correct for what the hook does today — it is scratch state for *one* search, seeded from the sector guess, edited with chips, handed to `searchJobs`. But it means the shortlist is **in-memory only**. Reload the tab and the companies a user hand-picked are gone, replaced by whatever the heuristic classifier suggests for their résumé's sector.

So a user who has decided "these are the eight companies I care about" has to re-decide it every single visit. `src/lib/storage/db.ts` already persists résumés, jobs, boards, letters and a sync cursor. The one thing that is purely a statement of user intent is the one thing that is not saved.

## What already exists

| Piece | Where | State |
|---|---|---|
| Curated company registry (`CompanyEntry`, `Ats`) | `src/lib/job-search/company-registry.ts` | present |
| Sector-seeded selection + add/remove chips | `src/hooks/useCompanyTargets.ts` | present, **ephemeral** |
| Board fan-out over selected companies | `src/lib/job-search/company-boards.ts` | present |
| IndexedDB with a versioned `upgrade()` ladder | `src/lib/storage/db.ts` — `DB_NAME = "offlinecv"`, `DB_VERSION = 4`, stores `resumes` / `jobs` / `boards` / `letters` / `sync` | present |
| Same-tab bridge to a companion extension | `src/lib/extension-profile.ts` | present, scoped to a résumé digest |

Nothing here needs inventing. This is a new store on an existing ladder, a writer beside an existing reader, and one more message on an existing bridge.

### 1. A persisted `watched` store

A new object store on the existing `db.ts` ladder — `DB_VERSION` 4 → 5, with a matching branch in `upgrade()`. The module's own docblock states the rule: the `upgrade()` callback fires once for the range `(oldVersion, DB_VERSION]`, so the new branch must be guarded rather than assumed to run alone.

The record is the shortlist entry a search already understands:

```ts
interface WatchedCompany {
  id: string;            // `${ats}:${slug}`, the natural key
  ats: Ats;              // "greenhouse" | "lever" | "ashby"
  slug: string;
  displayName: string;
  addedAt: number;
}
```

`id` is `${ats}:${slug}` rather than a random uuid because the same company on two vendors is two different boards, and the same company added twice is one row. Deriving the key from the pair makes that a storage property instead of a caller's discipline.

### 2. `useCompanyTargets` reads it, and the seed becomes a fallback

The hook keeps its seeded-once behaviour **only when the store is empty**. Once a user has saved a shortlist, that shortlist is what a fresh mount shows — the sector heuristic stops overwriting a decision the user already made.

This is the load-bearing behavioural change, and it must not regress the reason the current design exists: a re-render must still not stomp edits. Reading persisted state once on mount and treating it as the seed preserves that property rather than weakening it.

### 3. Explicit save, not implicit

Toggling a chip for one search should **not** silently rewrite the saved shortlist — that would make an exploratory search permanently change the user's targets. A distinct affordance ("save these as my companies" / a per-chip pin) writes to the store. Everything else stays scratch state.

**Decide and document which affordance it is** in the hook's docblock, in the same style as the two existing deliberate choices. The distinction between "targets for this search" and "companies I am watching" is the whole feature; leaving it implicit is how the two collapse back into one.

### 4. Expose the shortlist over the extension bridge

`src/lib/extension-profile.ts` is the existing same-tab `postMessage` door between this app and an installed companion extension. It is currently scoped to a résumé digest, and its docblock says outright that nothing in it fetches and it reaches no network — a property that must survive.

Add a **read-only** message for the watched shortlist: the extension asks, the app answers with the stored list. No write path from the extension into this store in this issue — the app owns its own data, and a one-way read is the smallest thing that makes the shortlist useful to a companion without handing an extension a writer.

**The bridge's no-network claim is unaffected and must stay asserted.** Reading an IndexedDB store and posting it to a same-tab listener touches no network primitive; whatever test pins that property today must still pass unchanged.

## What this does not reach, and why it is not a dead end

`company-registry.ts` already records the boundary, and it is worth restating so nobody scopes this issue as "watch any company":

> **STRUCTURAL LIMITATION** — large self-hosted-careers employers: Apple, Google, Meta, and most other FAANG-scale companies run their own careers site rather than a Greenhouse/Lever/Ashby board... The actual boundary is that their responses carry no `Access-Control-Allow-Origin` header, so a browser blocks OUR origin from reading them — a CORS rule, not a missing endpoint.

So a saved shortlist here covers the three vendors this app can actually read from a page origin. Self-hosted employers stay on the path `company-search-link.ts` already provides — linking the user's own browser, with the user's own session, into the employer's own careers search. This issue does not change that split and should not try to.

## Acceptance criteria

- [ ] `DB_VERSION` is 5 and `upgrade()` carries a guarded branch creating the `watched` store
- [ ] A saved shortlist survives a full page reload
- [ ] With an empty store, `useCompanyTargets` behaves exactly as it does today — sector-seeded, edits not stomped by re-render
- [ ] With a non-empty store, the saved shortlist is what a fresh mount shows; the sector heuristic does not overwrite it
- [ ] Toggling a chip for one search does not mutate the saved shortlist unless the user takes the explicit save action
- [ ] Adding the same `ats` + `slug` twice yields one row
- [ ] The chosen save affordance is documented in `useCompanyTargets.ts`'s docblock as a third deliberate choice, with its reasoning
- [ ] The extension bridge answers a read of the shortlist and offers no write path
- [ ] The bridge's existing no-network assertion still passes unchanged
- [ ] Lint, types and tests green

## Out of scope

- Any background or timed polling of the saved boards. This issue persists a list; it does not add a scheduler.
- A write path from the extension into this store.
- Extending `Ats` beyond the three vendors, or any attempt to read a self-hosted careers site from this origin — see the CORS boundary above.
- A shared or account-backed shortlist. This app is local-first and has no account; the store is per-browser, like everything else in `db.ts`.
- Re-verifying that a saved slug still resolves. Board churn is documented as expected drift; a periodic re-verify pass remains future work.

s-annam:
## Summary

`useCompanyTargets` selected companies from a sector guess but kept the shortlist in memory only, so a reload replaced a user's hand-picked companies with the heuristic's suggestion again. Adds a persisted `watched` IndexedDB store (DB_VERSION 4 → 5) and a per-chip pin affordance (`isWatched`/`toggleWatched`) that saves independently of the per-search selection toggle, so a saved shortlist survives a reload and wins over the sector seed on mount without stomping in-session edits. Also adds a read-only `get-watched-companies` message on a new, separately-audited extension bridge (`watched-companies-bridge.ts`), per the issue's clarification comment.

Closes #864

## Review focus

- `src/hooks/useCompanyTargets.ts` — `loadWatchlist()` isolates the new IndexedDB read from the pre-existing sector/registry dynamic imports, so a blocked/unavailable IndexedDB degrades only the shortlist, never the storage-independent sector-heuristic suggestions. Does the failure isolation actually hold in every combination (chunk-load failure vs read failure vs both)?
- `src/hooks/useCompanyTargets.ts:toggleWatched` — the optimistic `setWatchedKeys` runs before the `watchedApiRef` null-check, so a failed module-load lets the pin visually toggle without persisting (self-resolves on remount). Worth a look, not blocking.

## Test plan

- [x] `npx tsc -b --noEmit` clean
- [x] `npm run lint` clean
- [x] `npm run test` green (full suite, 381 files / 6336 passed)
- [x] Adversarial review: 2 rounds, 1 blocking finding fixed (storage-failure regression isolation), 2 nits surviving (see Review focus above and the v4→v5 migration-coverage nit)

## Adversarial review

Two rounds, standalone `/implement-issue` review loop (independent `code-reviewer` subagent per round, adversarial framing).

**Round 1** — 1 blocking, 1 nit.
- **Blocking**: `useCompanyTargets.ts`'s mount effect shared one try/catch between the new `getWatchedCompanies()` IndexedDB read and the pre-existing sector-classification/company-registry dynamic imports. A blocked/unavailable IndexedDB (private browsing, a content blocker, corporate policy) would blank the sector-heuristic suggestions too — a regression, since that path had no storage dependency before this PR. **Fixed**: extracted `loadWatchlist()`, which never rejects — the chunk-load and the read each get their own try/catch, falling back to an empty list on either failure while keeping `watchedApiRef` populated so `toggleWatched` still works going forward. Covered by two new tests simulating a blocked `indexedDB.open`.
- **Nit**: no v4→v5 migration test using the real-legacy-version-database-first pattern `letters.test.ts`/`sync-schema.test.ts` use (existing tests only exercise the fresh-v0-install path). Left open — coverage gap, not a suspected defect.

**Round 2** — re-verified the fix fresh (not just re-read the report), 0 blocking, 1 new nit.
- **Nit**: `toggleWatched`'s optimistic `setWatchedKeys` runs before the `watchedApiRef` null-check, so a failed *module-load* (rarer than the blocked-storage case round 1 fixed) lets the pin visually toggle without persisting — self-resolves on remount, no crash or data loss. Left open.

`clean` after round 2 (zero blocking findings survive). Both surviving nits are also called out under Review focus above.

### DIFF
--- a/src/hooks/useCompanyTargets.ts
+++ b/src/hooks/useCompanyTargets.ts
@@ -10,7 +10,7 @@
  * both need it: the panel passes `selected` to `searchJobs`, the component
  * toggles it.
  *
- * TWO DELIBERATE CHOICES:
+ * THREE DELIBERATE CHOICES:
  *
  * 1. HEURISTIC CLASSIFIER, NOT `classifySector`. `sector.ts` also exports the
  *    semantic `classifySector`, which loads a WebLLM model when WebGPU is
@@ -27,7 +27,22 @@
  *    its `JobQuery` — "local scratch state seeded once from the parse".
  *    NOTE: `exhaustive-deps` is NOT enabled in this repo, so nothing would
  *    have flagged the alternative; the empty dep array is a decision, not an
- *    oversight.
+ *    oversight. Since #864, the seed source itself branches on mount: the
+ *    persisted watchlist (`watched-companies.ts`) wins over the sector guess
+ *    when it is non-empty — see choice 3 below — but which source wins is
+ *    still decided exactly once, at mount, for the same reason.
+ *
+ * 3. A PER-CHIP PIN, NOT THE SEARCH-SELECTION TOGGLE, IS THE SAVE AFFORDANCE
+ *    (#864). Toggling a chip for THIS search (`toggle`/`isSelected`) and
+ *    saving a company to the PERSISTED shortlist (`toggleWatched`/`isWatched`)
+ *    are two different user intents and stay two different controls. A
+ *    "save current selection" button was the other option considered and was
+ *    rejected: it would silently DISCARD any previously pinned company that
+ *    isn't in this session's sector-suggested pool (a user who pinned
+ *    companies across two different sector guesses over two visits would have
+ *    one visit's save button erase the other's picks). A per-chip pin has no
+ *    such collision — pinning and unpinning compose one company at a time,
+ *    regardless of which sector guess is currently on screen.
  *
  * The registry and the sector taxonomy are dynamic-imported (the cascade-tier
  * pattern) so neither reaches the entry chunk — the panel renders before they
@@ -38,6 +53,54 @@ import { useCallback, useEffect, useRef, useState } from "react";
 import type { HeuristicParsedResume } from "../lib/heuristics/types.ts";
 import type { CompanyEntry } from "../lib/job-search/company-registry.ts";
 import type { Sector } from "../lib/job-search/sector.ts";
+import type { WatchedCompany } from "../lib/job-search/watched-companies.ts";
+
+/** Signatures of the dynamically-imported save/remove functions, without a
+ *  static import — the module itself is lazy-loaded (see the mount effect
+ *  below), and `typeof import(...)` names its exports' types with no runtime
+ *  cost. */
+type WatchedCompaniesApi = {
+  save: typeof import("../lib/job-search/watched-companies.ts").saveWatchedCompany;
+  remove: typeof import("../lib/job-search/watched-companies.ts").removeWatchedCompany;
+};
+
+/**
+ * The watchlist half of the mount seed, isolated from the sector half.
+ *
+ * A storage failure — IndexedDB blocked by private browsing, a content
+ * blocker, or corporate policy, so `openDB()` throws — must cost ONLY the
+ * watchlist. The sector heuristic and the company registry have no storage
+ * dependency and worked regardless before #864; sharing one try block with
+ * this read would let a blocked `openDB()` blank `sector`/`suggested` too and
+ * degrade a working surface to the keyless-only search.
+ *
+ * The two failure modes are kept apart on purpose. A failed MODULE LOAD leaves
+ * no api to call, so `toggleWatched` becomes a no-op. A failed READ still
+ * hands back the api, so pinning saves and unpinning removes for the rest of
+ * the session — the shortlist just didn't seed. Both answer with the same
+ * empty list an untouched store gives, so the seeding path downstream cannot
+ * tell "no storage" from "nothing saved yet", which is exactly right.
+ */
+async function loadWatchlist(): Promise<{
+  api: WatchedCompaniesApi | null;
+  watched: WatchedCompany[];
+}> {
+  let module: typeof import("../lib/job-search/watched-companies.ts");
+  try {
+    module = await import("../lib/job-search/watched-companies.ts");
+  } catch {
+    return { api: null, watched: [] };
+  }
+  const api = {
+    save: module.saveWatchedCompany,
+    remove: module.removeWatchedCompany,
+  };
+  try {
+    return { api, watched: await module.getWatchedCompanies() };
+  } catch {
+    return { api, watched: [] };
+  }
+}
 
 /**
  * How many companies a sector suggests (#542). Together with
@@ -61,6 +124,22 @@ export const COMPANY_LIMIT = 14;
 /** Stable identity for a registry entry. `slug` alone collides across vendors. */
 export function companyKey(entry: CompanyEntry): string {
   return `${entry.ats}:${entry.slug}`;
+}
+
+/** Same identity as {@link companyKey}, read off a `WatchedCompany` — kept
+ *  separate rather than routing a `WatchedCompany` through `companyKey`
+ *  itself, since the two types diverge (a watched company has no `sectors`). */
+function watchedCompanyToKey(entry: WatchedCompany): string {
+  return `${entry.ats}:${entry.slug}`;
+}
+
+/** A watched company, reshaped as the `CompanyEntry` `suggested`/`selected`
+ *  expect. `sectors: []` is a placeholder: nothing downstream of `suggested`/
+ *  `selected` reads a `CompanyEntry`'s `sectors` field (verified via grep —
+ *  the only reader is `companiesForSector` itself, which never sees a
+ *  watched-derived entry), so an empty array costs nothing. */
+function toCompanyEntry(entry: WatchedCompany): CompanyEntry {
+  return { name: entry.displayName, ats: entry.ats, slug: entry.slug, sectors: [] };
 }
 
 /** Add/remove one key from a selection set, returning a new set. */
@@ -88,6 +167,12 @@ export interface CompanyTargets {
   toggle(entry: CompanyEntry): void;
   /** Re-suggest against `runnerUp`; no-op when there isn't one. */
   switchToRunnerUp(): void;
+  /** True once this entry is on the persisted shortlist (independent of
+   *  `isSelected` — see the hook's docblock, third deliberate choice). */
+  isWatched(entry: CompanyEntry): boolean;
+  /** Pin/unpin `entry` on the persisted shortlist. Does NOT change
+   *  `isSelected`/`selected` for the current search. */
+  toggleWatched(entry: CompanyEntry): void;
 }
 
 export function useCompanyTargets(parsed: HeuristicParsedResume): CompanyTargets {
@@ -96,6 +181,9 @@ export function useCompanyTargets(parsed: HeuristicParsedResume): CompanyTargets
   const [runnerUp, setRunnerUp] = useState<Sector | null>(null);
   const [suggested, setSuggested] = useState<CompanyEntry[]>([]);
   const [keys, setKeys] = useState<ReadonlySet<string>>(() => new Set());
+  const [watchedKeys, setWatchedKeys] = useState<ReadonlySet<string>>(
+    () => new Set(),
+  );
 
   // The lazily-imported registry lookup, kept so `switchToRunnerUp` can
   // re-query without a second dynamic import round-trip.
@@ -104,29 +192,66 @@ export function useCompanyTargets(parsed: HeuristicParsedResume): CompanyTargets
   >(null);
   const parsedRef = useRef(parsed);
   parsedRef.current = parsed;
+  // The lazily-imported save/remove functions, kept so `toggleWatched` can
+  // call them without a second dynamic import round-trip.
+  const watchedApiRef = useRef<WatchedCompaniesApi | null>(null);
+  // Mirrors `watchedKeys` so `toggleWatched` reads current state without a
+  // stale closure — the same reason `parsedRef` mirrors `parsed`.
+  const watchedKeysRef = useRef(watchedKeys);
+  watchedKeysRef.current = watchedKeys;
 
   // Mount-only on purpose — see the docblock.
   useEffect(() => {
     let cancelled = false;
     void (async () => {
       try {
-        const [{ classifySectorHeuristic }, { companiesForSector }] =
-          await Promise.all([
-            import("../lib/job-search/sector.ts"),
-            import("../lib/job-search/company-registry.ts"),
-          ]);
+        const [
+          { classifySectorHeuristic },
+          { companiesForSector },
+          watchlist,
+        ] = await Promise.all([
+          import("../lib/job-search/sector.ts"),
+          import("../lib/job-search/company-registry.ts"),
+          // Resolves rather than rejects on a storage failure, so a blocked
+          // IndexedDB cannot take the two imports beside it down with it.
+          loadWatchlist(),
+        ]);
         if (cancelled) return;
         const guess = classifySectorHeuristic(parsedRef.current);
-        const pool = companiesForSector(guess.sector, COMPANY_LIMIT);
+        const watched = watchlist.watched;
+
+        // Every setter below fires together, after every await has settled —
+        // a throw partway through (a chunk that never loads) must not leave
+        // `sector` set while `suggested`/`keys` stay empty.
+        watchedApiRef.current = watchlist.api;
         lookupRef.current = companiesForSector;
         setSector(guess.sector);
         setRunnerUp(guess.runnerUp ?? null);
+        setWatchedKeys(new Set(watched.map(watchedCompanyToKey)));
+
+        // A saved shortlist wins over the sector guess — see choice 2/3 in
+        // the docblock. `sector`/`runnerUp` above are set regardless, for
+        // the header text and the "try other sector" affordance; only the
+        // POOL a fresh mount shows changes source.
+        // Capped and ordered most-recently-pinned-first, same as the sector
+        // branch caps at COMPANY_LIMIT — the watchlist otherwise grows past it
+        // across visits/sectors with nothing to bound the board fan-out (#897
+        // review). A bare slice would drop pins in storage-insertion order,
+        // which has no relationship to what the user pinned most recently.
+        const pool =
+          watched.length > 0
+            ? [...watched]
+                .sort((a, b) => b.addedAt - a.addedAt)
+                .slice(0, COMPANY_LIMIT)
+                .map(toCompanyEntry)
+            : companiesForSector(guess.sector, COMPANY_LIMIT);
         setSuggested(pool);
         setKeys(new Set(pool.map(companyKey)));
       } catch {
-        // Chunk failed to load (offline first-load). Company targeting is
-        // additive: leaving `suggested` empty degrades to the keyless-only
-        // search rather than breaking the panel.
+        // The sector/registry chunk failed to load (offline first-load) —
+        // `loadWatchlist` never reaches here, by construction. Company
+        // targeting is additive: leaving `suggested` empty degrades to the
+        // keyless-only search rather than breaking the panel.
       } finally {
         if (!cancelled) setReady(true);
       }
@@ -138,6 +263,23 @@ export function useCompanyTargets(parsed: HeuristicParsedResume): CompanyTargets
 
   const toggle = useCallback((entry: CompanyEntry) => {
     setKeys((current) => toggleKey(current, companyKey(entry)));
+  }, []);
+
+  const toggleWatched = useCallback((entry: CompanyEntry) => {
+    const key = companyKey(entry);
+    const wasWatched = watchedKeysRef.current.has(key);
+    setWatchedKeys((current) => toggleKey(current, key));
+    const api = watchedApiRef.current;
+    if (!api) return;
+    // Swallowed for the same reason `loadWatchlist` swallows the read: on a
+    // profile where storage is blocked the write rejects on every pin, and an
+    // unhandled rejection per click is the only thing that would come of
+    // propagating it. The optimistic state above still shows the pin for this
+    // session; nothing is silently lost that storage could have kept.
+    void (wasWatched
+      ? api.remove(entry.ats, entry.slug)
+      : api.save(entry)
+    ).catch(() => {});
   }, []);
 
   const switchToRunnerUp = useCallback(() => {
@@ -163,5 +305,7 @@ export function useCompanyTargets(parsed: HeuristicParsedResume): CompanyTargets
     isSelected: (entry) => keys.has(companyKey(entry)),
     toggle,
     switchToRunnerUp,
+    isWatched: (entry) => watchedKeys.has(companyKey(entry)),
+    toggleWatched,
   };
 }
--- a/src/lib/watched-companies-bridge.ts
+++ b/src/lib/watched-companies-bridge.ts
@@ -1,0 +1,116 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * The read-only app → extension door for the persisted company watchlist
+ * (#864).
+ *
+ * A SEPARATE module from `extension-profile.ts`, per that module's own scope
+ * (one fixed résumé digest, click-gated) — see the reasoning there and the
+ * `company-search-link.ts` precedent for why a second concern gets a second,
+ * separately-audited file rather than widening an already-audited one.
+ *
+ * This channel runs the OPPOSITE direction of every flow in
+ * `extension-profile.ts`: there the app always initiates a request and the
+ * extension replies; here the EXTENSION initiates a request
+ * (`get-watched-companies`) and this module answers it — so this file owns
+ * an always-listening responder rather than a one-shot request/reply.
+ *
+ * **READ-ONLY.** There is no message here that lets the extension write into
+ * the `watched` store — the app owns its own data, and adding a write path is
+ * explicitly out of scope for #864.
+ *
+ * **NO NETWORK.** Reading IndexedDB and posting to a same-tab listener
+ * touches no network primitive — the same property `extension-profile.ts`
+ * asserts of itself, and it holds here for the same reason.
+ *
+ * **Why the "explicit user action, or nothing" rule from `extension-profile.ts`
+ * doesn't carry over.** That module gates on a click because sharing creates a
+ * second on-device copy of a résumé digest outside this origin — a location
+ * the user didn't pick. This responder listens ambiently for the page's
+ * lifetime instead, and that's deliberate: the watchlist is a handful of
+ * company names the user already chose to save *in this app*, not sensitive
+ * corpus text, and answering "what's on it" on request is the whole point of
+ * a read-only door — gating each read behind a click would make the shortlist
+ * useless to the extension. The property that DOES still hold is READ-ONLY:
+ * this file never lets the extension write back, so there's no new copy this
+ * app doesn't already have.
+ */
+
+import { EXTENSION_CHANNEL, isFromThisPage } from "./extension-profile.ts";
+import type { WatchedCompany } from "./job-search/watched-companies.ts";
+
+/** Wire shape of one watched company, as posted to the extension. Kept as its
+ *  own type rather than a re-export of `WatchedCompany` — this is the WIRE
+ *  contract, and the two may drift on purpose later even though they are
+ *  identical today. */
+interface WatchedCompanyWire {
+  id: string;
+  ats: string;
+  slug: string;
+  displayName: string;
+  addedAt: number;
+}
+
+type IncomingRequest = {
+  channel: typeof EXTENSION_CHANNEL;
+  type: "get-watched-companies";
+};
+type OutgoingReply = {
+  channel: typeof EXTENSION_CHANNEL;
+  type: "watched-companies";
+  /** False only on a read failure (storage blocked, IndexedDB threw) — an
+   *  untouched store still replies `ok: true` with an empty list, so the
+   *  extension can tell "no information" apart from "nothing saved". */
+  ok: boolean;
+  companies: WatchedCompanyWire[];
+};
+
+function isGetWatchedCompaniesRequest(value: unknown): value is IncomingRequest {
+  if (value === null || typeof value !== "object") return false;
+  const message = value as { channel?: unknown; type?: unknown };
+  return message.channel === EXTENSION_CHANNEL && message.type === "get-watched-companies";
+}
+
+function toWire(company: WatchedCompany): WatchedCompanyWire {
+  return { ...company };
+}
+
+/**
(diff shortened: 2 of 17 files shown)


s-annam (commented):

## Stance
Reviewed against Issue #864 acceptance criteria, the issue clarification comment, fallow audit, and offlinecv quality gates.

All acceptance criteria are met:
- `DB_VERSION` bumped to 5 with guarded `upgrade()` branch creating `watched` store.
- Shortlist persists across page reloads in IndexedDB and seeds `useCompanyTargets` on mount (without stomping in-session chip edits).
- Distinct per-chip pin affordance (`isWatched`/`toggleWatched`) documents the third deliberate choice in `useCompanyTargets.ts`.
- Read-only extension bridge (`watched-companies-bridge.ts`) answers `get-watched-companies` without network access or write paths, using lazy dynamic import to prevent bundle bloat.
- Comprehensive test coverage across storage migrations (v4 → v5), hook isolation during storage errors, component interactions, and extension bridge message flows.
- 0 blocking findings, all gates green (`verify`, `fallow`, `lint`, `typecheck`, `test`).

## Fixed in 8712582
- `src/lib/watched-companies-bridge.ts`: Made `WatchedCompanyWire` an internal interface to clear fallow dead-code/unused-type warning, converted `watched-companies.ts` import to dynamic import on request, and added `.catch()` fallback to prevent unhandled rejections during storage errors.
- `src/lib/job-search/watched-companies.test.ts`: Added explicit `storage: schema migration v4 → v5 (#864)` test verifying legacy v4 profile upgrade and data preservation.
- Collapsed branch to exactly 1 commit (`8712582`) for merge-queue compliance.

---

s-annam:
Autopilot: MERGE READY after 1 bot round(s) and 0 human cycle(s) — approved, no conflicts, checks green, no unresolved threads. Merge is yours whenever you're ready.

src/lib/watched-companies-bridge.ts
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * Same-origin checks mirror `extension-profile.ts`'s `isFromThisPage`: a
+ * `postMessage` to `window` is visible to every same-page listener, so
+ * `event.source === window` rejects a post from an embedded frame and the
+ * origin check rejects a cross-origin frame holding a handle to this window.
+ */
+export function listenForWatchedCompaniesRequests(): () => void {
+  const appOrigin = window.location.origin;
+  const onMessage = (event: MessageEvent<unknown>) => {
+    if (event.source !== window || event.origin !== appOrigin) return;
```
s-annam: This same-origin/same-window check duplicates `extension-profile.ts`'s `isFromThisPage` (`event.source === window && event.origin === appOrigin`, negated here). It's correct, but `isFromThisPage` isn't exported, so this file re-derives a security-relevant check by hand instead of importing it — the two files' origin checks can now drift independently. Secondary, not blocking: consider exporting `isFromThisPage` from `extension-profile.ts` and importing it here in a follow-up.

src/lib/watched-companies-bridge.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (!isGetWatchedCompaniesRequest(event.data)) return;
+    void import("./job-search/watched-companies.ts")
```
s-annam, line 93: Dynamic-imports `./job-search/watched-companies.ts` fresh on every incoming request instead of caching the resolved module reference once, unlike `useCompanyTargets.ts`'s `watchedApiRef` in this same PR. Nit — the module loader already memoizes per specifier, so the real cost is negligible; flagging only for pattern consistency.

s-annam (commented):

Well-scoped, well-tested implementation of #864. The persisted `watched` IndexedDB store, the guarded v4→v5 migration, the per-chip pin as a *third* deliberate choice distinct from the search-selection toggle, and the separately-audited read-only extension bridge all follow the issue's clarification comment and the repo's own precedent (`board-cache.ts`, `company-search-link.ts`). Every acceptance criterion in #864 is met, including the storage-failure isolation the PR's own adversarial review round 1 caught and fixed (`loadWatchlist()` isolating a blocked IndexedDB from the pre-existing sector/registry suggestion path). `npm run verify` is green (typecheck, lint, fixtures, full test suite, build, fallow report-only).

## Blocking

None.

## Secondary

- **PR description is stale on one of its own disclosed nits.** Round 1's adversarial-review note says: *"no v4→v5 migration test using the real-legacy-version-database-first pattern `letters.test.ts`/`sync-schema.test.ts` use ... Left open — coverage gap, not a suspected defect."* That gap is actually closed in this same diff: `src/lib/job-search/watched-companies.test.ts`'s `describe("storage: schema migration v4 → v5 (#864)")` opens a real v4 legacy database via `openDB(DB_NAME, 4, {...})`, seeds a job record, then asserts `getDB()` migrates it to v5 with the `watched` store present and the seeded record intact — exactly the pattern named as missing. Worth correcting the PR body so a later reader doesn't go looking for a gap that isn't there.
- `src/lib/watched-companies-bridge.ts` reimplements `extension-profile.ts`'s `isFromThisPage` same-origin check inline (see inline comment) instead of importing it. The logic is correct today, but a security-relevant same-tab-origin check now exists in two files that can drift independently.

## Nits

- `src/lib/watched-companies-bridge.ts:77` — re-runs `import("./job-search/watched-companies.ts")` on every incoming `get-watched-companies` message instead of caching the resolved module once, unlike `useCompanyTargets.ts`'s own `watchedApiRef` in this same PR. Low real cost (the module loader memoizes per specifier), just an inconsistency with the sibling pattern.
- `src/hooks/useCompanyTargets.ts` — `loadWatchlist()` now sits inside the mount effect's `Promise.all`, so a slow/contended IndexedDB gates the whole suggestion pool (sector *and* registry), not just the watchlist half — a new latency dependency on a path that used to be storage-independent. `loadWatchlist`'s own failure isolation keeps this non-fatal, just worth knowing.
- Already disclosed and independently confirmed accurate: `toggleWatched` (useCompanyTargets.ts:263) flips the optimistic `watchedKeys` UI state before checking `watchedApiRef`, so a failed *module load* (not the more likely blocked-storage case round 1 fixed) shows a pin that didn't persist — self-resolves on remount, matches the PR body's own Round 2 nit.
- `db.ts`'s own docblock flags that a `DB_VERSION` bump needs a lockstep bump of `extension/offlinecv-pin.json` in the nested companion extension repo, which this repo's CI can't verify. Not something this diff can address (separate repo) — just confirm it's tracked before the extension next rebuilds against this schema.

## AC checklist (#864)

All nine checkable acceptance criteria are met — DB_VERSION 5 with a guarded `oldVersion < 5` branch, reload persistence (proven via the migration + read-back tests), empty-store behavior unchanged, non-empty store wins over the sector seed, chip toggle never mutates the saved shortlist (and is tested doing exactly that), same `ats`+`slug` upserts to one row, the hook's docblock documents the third deliberate choice with its rejected alternative, the bridge is read-only with no write message type, and the existing no-network assertion on `extension-profile.ts` is untouched. Lint/types/tests all green.

Summary and scope match the diff. The one inaccuracy is the stale Round 1 nit noted above — everything else in Summary, Review focus, and Test plan checks out against the code.

---

src/hooks/useCompanyTargets.ts
```diff
@@ -104,29 +192,58 @@ export function useCompanyTargets(parsed: HeuristicParsedResume): CompanyTargets
...
         setRunnerUp(guess.runnerUp ?? null);
+        setWatchedKeys(new Set(watched.map(watchedCompanyToKey)));
+
+        // A saved shortlist wins over the sector guess — see choice 2/3 in
+        // the docblock. `sector`/`runnerUp` above are set regardless, for
+        // the header text and the "try other sector" affordance; only the
+        // POOL a fresh mount shows changes source.
+        const pool =
+          watched.length > 0
+            ? watched.map(toCompanyEntry)
```
Samhit21: [Suggestion]: the watchlist seed bypasses `COMPANY_LIMIT`, so the board fan-out becomes unbounded

`COMPANY_LIMIT`'s own docblock 110 lines up in this file makes it load-bearing, not cosmetic:

> Together with `DEFAULT_PER_COMPANY_CAP` (8 by default, `role-keywords.ts`) this bounds the company-board half of a search at 14 * 8 = 112 postings before ranking — a real sample without a firehose... **It is also the width of the board fan-out**, which the concurrency limiter then meters.

The sector branch respects it (`companiesForSector(guess.sector, COMPANY_LIMIT)`); the watchlist branch is `watched.map(toCompanyEntry)` with no cap. And the next two lines auto-select the whole pool (`setKeys(new Set(pool.map(companyKey)))`), so every watched company is armed by default — `selected` flows uncapped through `FindJobsPanel.tsx:148` → `useJobSearch` → `searchJobs` → `makeBoardProviders(companies)`, which builds one provider per company. I traced the whole path: `COMPANY_LIMIT` at the seed is the only cap that exists.

The watchlist also grows past 14 easily, since it accumulates across visits and sectors: pin 5 fintech, "try other sector", pin 10 more, and a returning user is now at 15 pre-selected boards. Nothing bounds it.

```ts
const pool =
  watched.length > 0
    ? watched.slice(0, COMPANY_LIMIT).map(toCompanyEntry)
    : companiesForSector(guess.sector, COMPANY_LIMIT);
```

A bare `slice` silently drops pins, so it probably wants a deterministic order (most-recently-added first — `addedAt` is already on the record) and ideally a note in the UI when the shortlist exceeds the cap. Whichever shape you pick, the invariant this file documents should hold on both branches.

src/hooks/useCompanyTargets.ts
```diff
@@ -104,29 +192,58 @@ export function useCompanyTargets(parsed: HeuristicParsedResume): CompanyTargets
...
         setSector(guess.sector);
         setRunnerUp(guess.runnerUp ?? null);
+        setWatchedKeys(new Set(watched.map(watchedCompanyToKey)));
+
+        // A saved shortlist wins over the sector guess — see choice 2/3 in
+        // the docblock. `sector`/`runnerUp` above are set regardless, for
+        // the header text and the "try other sector" affordance; only the
+        // POOL a fresh mount shows changes source.
+        const pool =
+          watched.length > 0
```
Samhit21, line 242: [Question]: is pinning intended to be a one-way door for company discovery?

Once `watched` is non-empty, this branch means a fresh mount never shows the sector pool again — the chip row is exactly the watchlist, on every subsequent visit. So a user who pins one company loses the surface they'd use to pin a second one.

`switchToRunnerUp` is the only way back to registry suggestions, and it's an odd fit: it's labelled as trying the *other* sector, it needs a non-null `runnerUp`, and reaching your own sector's pool takes two clicks (the second `lookup(sector)` call, since the pair swaps). The watchlist pool itself is then unrecoverable for the rest of the session.

I don't think this is a bug against #864 — "non-empty store wins over the sector seed" is an explicit AC, so replacing the pool is what was asked for. Flagging it because the AC's consequence for *discovery* may not have been the intent, and it's cheap to address now (append the capped sector pool below the watched entries, watched ones pinned and selected) versus after users have shortlists.

src/lib/watched-companies-bridge.ts
```diff
@@ -0,0 +1,98 @@
...
+      .then((m) => m.getWatchedCompanies())
+      .then((companies) => {
+        const reply: OutgoingReply = {
+          channel: EXTENSION_CHANNEL,
+          type: "watched-companies",
+          companies: companies.map(toWire),
+        };
+        window.postMessage(reply, appOrigin);
+      })
+      .catch(() => {
```
Samhit21, line 104: [Nit]: the failure reply is indistinguishable from a genuinely empty watchlist

`.catch()` posts `companies: []`, the same message a user who has pinned nothing gets. The extension cannot tell "read failed / storage blocked" from "nothing saved", so a transient IndexedDB failure reads as an affirmative empty shortlist on the other side.

`loadWatchlist` in `useCompanyTargets.ts` makes the same collapse deliberately and says why — but its consumer is the seeding path in this repo, where the two really are equivalent. Here the consumer is another codebase that may cache or act on the answer.

Either carry the distinction on the wire (an `ok: false`, or an omitted `companies`) or add a line to the docblock saying the collapse is intended and that the extension must treat an empty list as "no information". Right now a reader of the bridge has no way to know which it is.

src/lib/watched-companies-bridge.ts
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Start answering `get-watched-companies` requests from the extension. Call
+ * once per page (see `useWatchedCompaniesBridge`); returns the unsubscribe.
+ *
+ * Same-origin checks mirror `extension-profile.ts`'s `isFromThisPage`: a
+ * `postMessage` to `window` is visible to every same-page listener, so
+ * `event.source === window` rejects a post from an embedded frame and the
+ * origin check rejects a cross-origin frame holding a handle to this window.
+ */
+export function listenForWatchedCompaniesRequests(): () => void {
```
Samhit21, line 88: [Question]: why doesn't `extension-profile.ts`'s "explicit user action, or nothing" rule carry over here?

The module docblock points at `extension-profile.ts` for the reasoning and mirrors two of its three stated properties — **NO NETWORK** and same-origin posting — but not the third, which that file gives its own heading:

> ## An explicit user action, or nothing
>
> Storing it in the extension creates a **second on-device copy outside this origin** — not egress, and not a break in the custody claim, but a new location the user did not pick. So `shareResumeProfile` runs on a click and on nothing else: no ambient push on load, none on save, none on edit.

This responder answers ambiently for the lifetime of the page, so the extension can pull the shortlist whenever it likes and persist it, with no gesture and no way back out (there's no `clearSharedWatchedCompanies` counterpart).

The list is far less sensitive than a résumé corpus, and an always-listening responder is inherent to the extension-initiates direction this file calls out — so this may well be the right call. But the docblock currently invites the comparison while leaving the one principle that cuts against it unmentioned. Worth either a sentence saying why the copy-outside-this-origin argument doesn't apply to a company list, or a gate.

Samhit21 (commented):
## PR Review: feat(job-search): persist the company shortlist, expose it read-only to the extension

### Summary
Well-constructed and unusually well-documented — the v4→v5 migration is properly additive and tested against a real legacy database, the pin-vs-select separation is argued in the docblock with its rejected alternative, and `loadWatchlist`'s two-try/catch isolation (caught by the PR's own round 1) is the right fix for the right reason. Prior review already covered the `isFromThisPage` duplication, the uncached dynamic import, the `toggleWatched` optimistic-toggle nit, and the stale Round-1 note in the PR body; none of that is repeated below. I verified two of the docblocks' load-bearing claims and both hold: `CompanyEntry.sectors` really does have exactly one reader (`company-registry.ts:318`), so the `sectors: []` placeholder is safe, and `watched` is correctly absent from `backup.ts`'s replace-mode wipe.

### Findings
1. **`useCompanyTargets.ts:238` — the watchlist seed bypasses `COMPANY_LIMIT`.** The sector branch caps the pool at 14; the watchlist branch doesn't cap at all, and the pool is auto-selected. `selected` reaches `makeBoardProviders` uncapped, so the "14 × 8 = 112 postings, not a firehose" bound this file documents stops holding for exactly the users the feature is for. The shortlist accumulates across visits and sectors, so exceeding 14 is ordinary, not pathological. **The main finding.**
2. **`useCompanyTargets.ts:237` — pinning is a one-way door for discovery.** A non-empty watchlist permanently replaces the sector pool on mount, so there's no surface left to pin a second company from except "try other sector". Consistent with the AC as written; flagging the consequence in case it wasn't the intent.
3. **`watched-companies-bridge.ts:87` — the error reply is `companies: []`,** identical to a genuine empty shortlist. Fine inside this repo; less fine for a cross-repo consumer that may cache it. Wire distinction or a docblock note.
4. **`watched-companies-bridge.ts:72` — no consent gate,** where the sibling module it cites treats "an explicit user action, or nothing" as a named principle for the same second-copy-outside-this-origin reason. Probably defensible for a company list; currently unaddressed in a docblock that invites the comparison.

Findings 1 and 2 share a root (the pool ternary replacing rather than merging) but want different fixes, so they're separate comments.

### Verdict
**Action: COMMENT** — nothing here is merge-blocking. Finding 1 is the one I'd want resolved before this reaches users with shortlists, since the fan-out width is a property the repo has deliberately bounded elsewhere.

s-annam:
Autopilot: **reviewed, no blockers** — bot round 1/3 converged (0 blockers, self-review posted as COMMENT with semantic verdict APPROVE), then `/pr-ready`'s solicitation drew a human review from @Samhit21 (`COMMENTED`, not `CHANGES_REQUESTED`) at 22:13 UTC.

No conflicts, checks green, mergeable. Since the human review is `COMMENTED` rather than `CHANGES_REQUESTED`, this run stops here per its own rules rather than manufacturing a revision cycle out of a non-blocking review — but the review's finding 1 (`useCompanyTargets.ts:238` — the watchlist seed bypasses `COMPANY_LIMIT`, so a saved shortlist can grow the board fan-out past the 14-company bound the file documents) is worth your own judgment call before merge, per the reviewer's own "I'd want this resolved" note.

`/pr-ready`'s wait already ran once this session. Re-running `/pr-autopilot 897` will re-check current state (e.g. if you push a fix and want the bot pass to re-verify it, or want to re-solicit).

src/lib/watched-companies-bridge.ts
s-annam: Fixed in 1618828 — exported `isFromThisPage` from `extension-profile.ts` and imported it here instead of re-deriving the same origin check.

src/lib/watched-companies-bridge.ts
s-annam: Not fixing — you already called this negligible ("the module loader already memoizes per specifier"), and hoisting a cached ref here would add a mutable module-level (or WeakRef-guarded) reference to a file whose whole design is a stateless, always-listening responder, for a saving that doesn't exist in practice. Leaving it as-is.

src/hooks/useCompanyTargets.ts
s-annam: Fixed in 1618828 — the watchlist branch now sorts most-recently-pinned first and slices to COMPANY_LIMIT before mapping to CompanyEntry, matching the sector branch's existing bound. Added a test (`caps a shortlist larger than COMPANY_LIMIT`) that pins COMPANY_LIMIT+3 and asserts the seeded pool stays at COMPANY_LIMIT.

src/hooks/useCompanyTargets.ts
s-annam: Not changing this round — you already flagged it as not-a-bug-against-#864 ("non-empty store wins over the sector seed" is the explicit AC), and appending the capped sector pool below the watched entries is a real product call: does the appended pool auto-select the same as watched entries, or come in unselected? That's not mine to decide unilaterally in a review-fix commit. Leaving this thread open — worth its own follow-up issue if we want to change the discovery UX, separate from this PR.

src/lib/watched-companies-bridge.ts
s-annam: Fixed in 1618828 — added an `ok: boolean` field to the wire reply, true only when the read actually succeeded. The catch branch now posts `ok: false, companies: []`, distinguishable from an untouched store's `ok: true, companies: []`. Added a test for the failure path using the same `withStorageBlocked` helper CompanyTargets.test.tsx already has.

src/lib/watched-companies-bridge.ts
s-annam: Answered in 1618828 — added a docblock paragraph explaining why: the gate exists in extension-profile.ts because sharing creates a second on-device copy of sensitive résumé text outside this origin. A watchlist is a handful of company names the user already chose to save in this app, and answering "what's on it" on request is the whole point of a read-only door — gating each read behind a click would make the shortlist useless to the extension. READ-ONLY is the property that still holds: no new copy gets created here that this app doesn't already have.

s-annam:
## Revision round: addressed review feedback

Commit `1618828` responds to Samhit21's review and a self-review comment.

**Fixed:**
- **`COMPANY_LIMIT` bypass** (`useCompanyTargets.ts:238`) — the watchlist-seed pool was unbounded, so a shortlist grown past 14 across visits fed an uncapped board fan-out. Now sorted most-recently-pinned-first and sliced to `COMPANY_LIMIT`, matching the sector branch's existing bound. Covered by a new test that pins `COMPANY_LIMIT + 3` companies and asserts the seeded pool stays capped.
- **Bridge read-failure ambiguity** (`watched-companies-bridge.ts:87`) — a failed IndexedDB read and a genuinely empty watchlist both posted `companies: []`, indistinguishable to the extension. Added an `ok: boolean` field to the wire reply; `ok: false` now marks a real read failure. Covered by a new failure-path test using the same `withStorageBlocked` helper `CompanyTargets.test.tsx` already has.
- **Duplicated origin check** (`watched-companies-bridge.ts:75`) — exported `isFromThisPage` from `extension-profile.ts` and imported it here instead of re-deriving the same same-origin check by hand.
- **Consent-gate question** (`watched-companies-bridge.ts:72`) — added a docblock paragraph explaining why the "explicit user action, or nothing" rule from `extension-profile.ts` doesn't carry over to this bridge: the gate there exists because sharing creates a second on-device copy of sensitive résumé text; a watchlist is a handful of company names the user already chose to save in this app, and answering "what's on it" on request is the point of a read-only door. READ-ONLY is the property that still holds.

**Not changed, replied on the thread:**
- **Dynamic-import caching nit** (`watched-companies-bridge.ts:77`) — reviewer's own comment called the cost negligible; not worth adding a cached module ref for.
- **One-way discovery door** (`useCompanyTargets.ts:237`) — a real product question (should the sector pool reappear once a watchlist exists?), but not a bug against #864's AC, and the shape of a fix (does the appended pool auto-select?) isn't mine to decide unilaterally in a fix commit. Left open for a follow-up issue if we want to revisit the discovery UX.

Gates: `npm run typecheck` clean, full suite green (381 files, 6339 tests). Re-requested review from Samhit21 since the push dismissed their prior review.

Separately, still outstanding before merge: `extension/offlinecv-pin.json` needs bumping to match this branch's `DB_VERSION` 4→5 change — not part of this revision round, flagged earlier.

Samhit21 (approved):
## PR Review: round 2 — all four findings addressed

Re-reviewed `1618828` against the four findings from my previous round. I read the fix diff rather than the summary, and all four hold up.

### Verified

1. **`COMPANY_LIMIT` bypass — fixed, correctly.** `[...watched].sort((a, b) => b.addedAt - a.addedAt).slice(0, COMPANY_LIMIT).map(toCompanyEntry)`. Copies before sorting (no mutation of the array handed back by `loadWatchlist`), orders most-recently-pinned first, then caps — so `selected` can no longer exceed the bound `makeBoardProviders` fans out on. The comment explaining why a bare slice was wrong is the right thing to have written down.
2. **Bridge read-failure ambiguity — fixed.** `ok: boolean` on `OutgoingReply`, `true` on the success path and `false` in the catch, with a docblock on the field saying an untouched store still answers `ok: true`. Additive on the wire, so it costs a consumer that ignores it nothing.
3. **Duplicated origin check — fixed.** `isFromThisPage` is now exported from `extension-profile.ts` and imported here, and the call site reads `if (!isFromThisPage(event, appOrigin))`. The security-relevant check exists once. (That one was your own finding, not mine — noting it since it landed in the same commit.)
4. **Consent-gate question — answered, and I accept the answer.** The new docblock paragraph draws the distinction I was asking about: the gate in `extension-profile.ts` exists because sharing creates a second on-device copy of résumé text, and a watchlist of company names the user already saved in this app is not that. Naming READ-ONLY as the property that still holds is the right framing. My comment asked for "a sentence saying why the argument doesn't apply, or a gate" — this is the former, and it's a better sentence than I'd have written.

Both non-fixes are reasonable and I'm not pushing back on either. The dynamic-import nit was negligible by my own admission. The one-way discovery door is genuinely a product call — whether an appended sector pool arrives selected or unselected isn't something to settle unilaterally in a review-fix commit, and a follow-up issue is the right home for it.

The two new tests are non-vacuous: the cap test pins `COMPANY_LIMIT + 3` and asserts the pool stays at `COMPANY_LIMIT`, and the bridge failure test drives the real `withStorageBlocked` path and asserts `ok: false`.

### ⚠️ Before this reaches the merge queue: collapse to one commit

The branch is now two commits (`8712582` + `1618828`). CLAUDE.md's third hard rule:

> **One commit per PR.** `main` merges through a merge queue that derives the squash message from the branch, so a multi-commit PR lands `wip` and `fix lint` in `main` forever. Collapse the branch to a single commit before it reaches the queue.

Your own round-1 self-review noted collapsing to exactly one commit for this reason, so this is just the fix-up push undoing that. Not a code problem and not why I'd hold the PR — but it needs doing before merge, and an approval shouldn't be read as "merge as-is".

### Nit (non-blocking)

- **`CompanyTargets.test.tsx` — the new cap test doesn't actually pin the sort.** It asserts the pool is `COMPANY_LIMIT` long and that every entry is in the saved set, both of which stay true if the `.sort((a, b) => b.addedAt - a.addedAt)` is deleted: slicing 14 of 17 in storage-key order still yields 14 saved entries. So the "most-recently-pinned-first" half of the fix — the half the code comment argues for at length — is untested, and a later refactor could drop it silently. Pinning distinct `addedAt` values and asserting the surviving names would close it. Worth mentioning given #862 in `main` was specifically about making stable-sort assertions non-vacuous.

### Verdict

**Action: APPROVE.** Every finding I raised is resolved or answered, CI is green on `1618828` (`verify` 4m27s, `fallow`, Cloudflare Pages), and the residual items are one nit and one merge-hygiene step. Squash before queueing.

Issue #765, opened 2026-08-05
Epic: three-tier artifacts — standard → company → job, for résumé and cover letter


s-annam:
Today a résumé and a cover letter are each a **flat, standalone artifact**. This epic replaces that with a three-tier hierarchy — **standard → company → job** — for both.

## The framing (this is the load-bearing part)

The tempting reading is "three scopes, pick one." That reading is wrong and would produce three independent copies that drift.

**The standard artifact is the SOURCE.** A company artifact is a derivation of it; a job artifact is a derivation of that. The standard résumé is what you have; the standard cover letter holds the candidate-specific material that has no place on a résumé — a career-change narrative, a relocation or visa fact, a reason for a gap. The system then customizes downward.

That reframing matters because it decides the data model: **base + delta**, not three rows. Improving the base has to reach the derivations that have not overridden the thing you improved.

It also answers the obvious objection. A *generic* cover letter is weak in a bad market, and nothing here proposes submitting one. Standard is the base you tailor **from**, and the resolution chain is what makes tailoring cheap: write the "why this company" paragraph once for the four roles you are applying to there, and tailor only the "why this role" part per posting.

## Letters and résumés are NOT symmetric — this is the sequencing

| | Cover letter | Résumé |
|---|---|---|
| Artifact | Prose (`LetterRecord.body`, one markdown string) | Structured (`CanonicalResume`: sections, roles, bullets) |
| What a derivation IS | A different body | A field- and bullet-level **delta** |
| Merge semantics needed | None | Yes — and this is the hard part |
| Blocked on | Nothing | Persisted deltas + a base-drift rule |

So **letters can ship the full three tiers before résumés can.** A letter needs only a scope key. A résumé needs a delta that survives its base changing underneath it.

## What already exists (verified at HEAD, not assumed)

The résumé half is much closer than it looks:

- **`src/lib/edit/apply-overrides.ts` is already a base + delta engine.** It takes a frozen base parse plus ordered override maps (`BulletOverrides`, `DescriptionOverrides`, `ContactOverrides`, `AddedBullets`, … — declared in `src/hooks/useEditableParse.ts`) and folds them into a fresh `{ parsed, rawText, sections }` the scorer re-grades from. It is pure and total.
[...]

PR #898, opened 2026-08-25
feat(storage): optional jobId + companyKey letter scope keys (cover-letter contract v2)

### LINKED ISSUE #766: Letter scope keys: optional jobId + companyKey (cover-letter contract v2)
Part of #765. **Read #765 first** — it carries the standard-is-the-source framing, the never-inferred rule for company identity, and why letters can ship all three tiers before résumés can.

This issue is the store and contract half only. No UI: that is the sibling issue.

## Problem

`LetterRecord.jobId` is **required** (`src/lib/storage/types.ts:377`; `docs/cover-letter-contract.md` §1), and its rule in `LETTER_RECORD_RULES` is `required: true` (`src/lib/storage/letter-contract.ts:106`). Deleting a job **cascades** to every letter that names it (`src/lib/storage/letters.ts:102`).

So a letter that is not for one specific job cannot exist. Neither a company letter nor a standard letter has a home.

The contract's own justification for requiring `jobId` is in §5:

> A letter reaches every surface through its job — a letters list is a list *within* a job. So a letter whose job is gone is unreachable: nothing can list it, open it, edit it, or delete it.

That reasoning was correct when written and is **exactly what this issue invalidates**: once a standard/company letter surface exists, a jobless letter is reachable. The doc change is therefore a rewrite of the argument, not a tweak to a table row.

## Design

Two optional keys form a three-case lattice. Exactly one reading applies to any record:

| `jobId` | `companyKey` | Reading |
|---|---|---|
| set | — | Letter for one job (every letter that exists today) |
| — | set | Letter for one company |
| — | — | Standard letter |

Both are `isNonEmptyString` **when present**, so `""` stays a refusal rather than becoming a silent fourth state.

A record carrying both is **refused**. It has no single reading, and accepting one would mean picking a precedence rule that no producer could guess.

### Cascade rules per scope

- Deleting a job cascades to **job letters only**. This is free, not a new guard: `lettersForJob` filters `letter.jobId === jobId` (`src/lib/storage/letters.ts:81`), and `undefined` never equals a real id. Write the reasoning into the docblock so nobody later "fixes" it with a filter.
- Deleting the last job at a company **does not** touch the company letter. It was never that job's.
- There is no delete-company action, so a company letter and a standard letter are only ever deleted explicitly.

### Company key derivation

There is no company entity — `JobRecord.company` is free text and may be empty (`src/lib/storage/types.ts:198`). `companyKey` is derived by normalisation and is **advisory**: per #765, it only ever drives a suggestion the user confirms. Nothing auto-attaches a letter to a job.

Reuse: `src/lib/job-search/raw-postings.ts:32` holds a private `normalizeField` (lowercase + whitespace collapse) for cross-provider dedup. Not exported, no legal-suffix handling, and in `job-search/` — `storage/` importing it would invert the layering. Write a new one in `storage/` whose docblock names the sibling and says why they are not shared.

## Steps

1. **`src/lib/storage/types.ts`** — `jobId?: string`, add `companyKey?: string`. Rewrite the `jobId` docblock: the three-case lattice, why both-set is refused, and why the cascade cannot reach a jobless letter.

2. **`src/lib/storage/letter-contract.ts`** — set `jobId`'s rule to `required: false` (keep `isNonEmptyString`), add the matching `companyKey` rule. Bump `LETTER_CONTRACT_VERSION` to `2` (currently `1`, at line 65). Add the both-set refusal to `validateLetterRecord` — it is a cross-field rule, so it does not live in `LETTER_RECORD_RULES` and needs its own check after `checkDeclaredFields`.

   Note the mapped type over `keyof Required<LetterRecord>` will not compile until `companyKey` has a rule. That is the intended drift guard, not an obstacle.

3. **`src/lib/storage/company-key.ts`** (new) — derive a key from a free-text company name. Lowercase, collapse whitespace, strip trailing legal suffixes (`Inc`, `Inc.`, `LLC`, `Ltd`, `Limited`, `Corp`, `Corporation`, `Co`, `GmbH`, `S.A.`, `B.V.`, `Pty`), drop surrounding punctuation. Returns `undefined` for an empty or all-punctuation input, so a job with no company simply has no key. Pure, zero-dep.

4. **`src/lib/storage/letters.ts`** — `saveLetter`'s required `Pick` drops `jobId` (leaving `body`). Add `lettersForCompany(companyKey)` and `standardLetters()`. Extend `deleteLettersForJob`'s docblock with the by-construction safety note from above.

5. **`src/hooks/useJobLetters.ts`** — the grouping loop does `grouped.get(letter.jobId)`, which would create a literal `undefined` map key for a jobless letter. Skip those, and expose company/standard letters separately from `byJobId`.

6. **`docs/cover-letter-contract.md` → v2** — §1 (both keys optional, `""` still refused, both-set refused), §5 (**rewrite** the reachability argument and add the per-scope cascade table), §6 (bump, plus: a v1 producer that always sends `jobId` keeps working untouched — the bump says there is more it MAY send), §8, and the known-gap paragraph (a jobless letter has no `jobId` to dangle, so merge-import strands strictly fewer letters). Note the sync dependency and link https://github.com/s-annam/recruidea-extension/issues/55.

## Acceptance criteria

- [ ] A letter saves with `companyKey` and no `jobId`, and reads back from `lettersForCompany`
- [ ] A letter saves with neither key and reads back from `standardLetters`
- [ ] A record with **both** keys is refused by `validateLetterRecord`, with a reason naming both fields
- [ ] `jobId: ""` and `companyKey: ""` are still refused (absent ≠ empty)
- [ ] Deleting a job tombstones its own letters and leaves company and standard letters live — asserted directly, since this is the invariant the whole lattice rests on
- [ ] `deleteLettersForJob` is idempotent, as it is today
- [ ] A company letter and a standard letter both survive export → import → export byte-identically
- [ ] `useJobLetters` produces **no** map entry keyed `undefined` when a jobless letter is in the store
- [ ] The company normaliser maps `Northwind`, `Northwind Inc.`, `northwind, inc.` to one key, and returns `undefined` for `""` and for `"  ,  "`
- [ ] `LETTER_CONTRACT_VERSION` is `2` and `docs/cover-letter-contract.md` says v2 in its header
- [ ] A v1-shaped record (always sends `jobId`, never `companyKey`) still validates and stores unchanged
- [ ] `npm run verify` green

## Out of scope

- Any UI. The picker, the resolution chain, and the standard/company editors are the sibling issue.
- Auto-attaching a letter to a job by company match. Advisory only — see #765.
- The `company_key` column on the remote `cover_letters` table, and the sync mapper that refuses jobless letters — https://github.com/s-annam/recruidea-extension/issues/55.

## Blocking dependency

`letter-mapping.ts` in `s-annam/recruidea-extension` **refuses a pulled letter with no `local_job_id`** while happily pushing one, so a standard or company letter uploads and can never come back — it silently fails to appear on a second device. Land https://github.com/s-annam/recruidea-extension/issues/55 before or with this issue, or multi-device users lose every letter this issue makes possible.

Samhit21:
Closes #766. Part of #765 — the store-and-contract half only; no UI.

## What changed

`LetterRecord.jobId` was required and deleting a job cascaded to every letter naming it, so a letter that is not for one specific job could not exist. Both scope keys are now optional and form a three-case lattice, with exactly one reading per record:

| `jobId` | `companyKey` | Reading |
|---|---|---|
| set | — | Letter for one job (every letter that exists today) |
| — | set | Letter for one company |
| — | — | Standard letter |

A record carrying **both is refused**, with a reason naming both fields — it has no single reading, and resolving it by precedence would mean picking a rule no producer could guess. Each key stays `isNonEmptyString` *when present*, so `""` remains a refusal rather than a silent fourth state.

- **`types.ts`** — `jobId?`, new `companyKey?`, docblocks carrying the lattice and why both-set is refused.
- **`letter-contract.ts`** — `jobId` → `required: false`, new `companyKey` rule, `LETTER_CONTRACT_VERSION` → `2`, and the both-set refusal in `validateLetterRecord` (cross-field, so it cannot live in `LETTER_RECORD_RULES`).
- **`company-key.ts`** (new) — pure, zero-dep normaliser: lowercase, collapse whitespace, drop punctuation, strip one trailing legal suffix. Returns `undefined`, never `""`, so a job with no company yields "no key" rather than a key the contract refuses. Its docblock names `normalizeField` in `job-search/raw-postings.ts` and says why they are deliberately not shared (layering, and the two want different roughness).
- **`letters.ts`** — `saveLetter`'s required `Pick` drops to `body`; new `lettersForCompany` and `standardLetters`; `deleteLettersForJob` gains the by-construction safety note.
- **`useJobLetters.ts`** — the grouping loop no longer creates a literal `undefined` map key; `byCompanyKey` and `standard` are exposed beside `byJobId`. The pure `groupByScope` moved to module scope so the hazard is testable without a render.
- **`docs/cover-letter-contract.md` → v2** — §1 (+ a new §1.1 on derivation), §5 **rewritten**, §6, §8, the known-gap paragraph, and the appendix.

## On §5's rewrite

The doc change is a rewrite of the argument, not a tweak to a table row. v1 justified the cascade with:

> A letter reaches every surface through its job … So a letter whose job is gone is unreachable.

That was correct when written and is exactly what this issue invalidates — once a standard/company surface exists, a jobless letter is reachable on its own terms. What survives is the narrower claim, and it is enough: a letter naming a *specific, now-deleted* job is unreachable. That is a statement about job letters only, which is why `jobId` could become optional without weakening it.

The cascade is scope-limited **by construction, not by a new guard**: `lettersForJob` matches `letter.jobId === jobId`, and `undefined` never equals a real id. Both the docblock and §5 say so explicitly, so nobody later "fixes" it with a filter — widening it would delete exactly what the lattice protects, and adding a redundant `companyKey === undefined` check would be a weaker copy of a rule the scoping already makes airtight.

## Acceptance criteria

All met, each with a direct assertion:

- [x] Company letter saves with no `jobId`, reads back from `lettersForCompany`
- [x] Standard letter saves with neither key, reads back from `standardLetters`
- [x] Both keys → refused, reason names both fields
- [x] `jobId: ""` and `companyKey: ""` still refused (absent ≠ empty)
- [x] Deleting a job tombstones its own letters and leaves company + standard letters live — asserted directly, including the delete-the-last-job-at-a-company case
- [x] `deleteLettersForJob` still idempotent, now with jobless letters in the store
- [x] Company + standard letters survive export → import → export byte-identically (`exportedAt` and the `updatedAt` `putRecord` restamps excluded, the pre-existing rule for every record type)
- [x] `useJobLetters` produces no `undefined`-keyed entry
- [x] `Northwind` / `Northwind Inc.` / `northwind, inc.` → one key; `""` and `"  ,  "` → `undefined`
- [x] `LETTER_CONTRACT_VERSION === 2`, asserted against the doc header value
- [x] A v1-shaped record still validates and stores unchanged

## Out of scope, as specified

No UI (the picker, resolution chain and editors are the sibling issue), no auto-attaching a letter to a job by company match, and no `company_key` column or sync mapper.

`lettersForCompany`, `standardLetters` and `deriveCompanyKey` are exported from their own modules but deliberately **not** re-exported through the `storage` barrel or `@offlinecv/core`: nothing in this build reads a company or standard letter yet, and a barrel entry with no importer is dead surface (fallow flags it as exactly that — see below). The change that adds the first caller adds the barrel entry with it. A comment at the barrel site says so, so the omission reads as a decision rather than an oversight.

## ⚠️ Blocking dependency, unverified from here

https://github.com/s-annam/recruidea-extension/issues/55 — `letter-mapping.ts` refuses a pulled letter with no `local_job_id` while happily pushing one, so a standard or company letter uploads and can never come back, failing silently on a second device. **That must land before or with this**, or multi-device users lose every letter this PR makes possible. I could not check its status: the extension repo is not readable from this account.

## Verification

`typecheck` and `lint` clean. Full suite: 6335 passed, 1 failure — `render-ats-pdf.type-scale.test.ts > picks the LARGEST rung…` timed out at 20s under a load average of ~51 from the run itself, and passes in 3.8s in isolation on an idle machine. Unrelated to this change, which touches no PDF code. The final `verify` was not re-run to green locally; CI is the check that matters here.

---

## Update — fallow fix (`148c40c`)

The first push failed the `fallow` check with three `unused-export` errors: `lettersForCompany`, `standardLetters` and `deriveCompanyKey` were re-exported from `src/lib/storage/index.ts` with no importer anywhere (every consumer, tests included, imports from the source module directly). Removed the three barrel entries and left a comment saying why they're absent and when to add them — the same resolution #897 used for its own `unused-export` finding.

`verify` passed on both pushes. Local re-check after the fix: `typecheck`, `lint`, the three affected suites (61 tests), and `fallow audit --base origin/main` all clean.

### DIFF
--- a/src/lib/storage/company-key.ts
+++ b/src/lib/storage/company-key.ts
@@ -1,0 +1,117 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 The offlinecv Authors
+
+/**
+ * Derive `LetterRecord.companyKey` from a free-text company name (#766).
+ *
+ * There is no company entity in this build. `JobRecord.company` is whatever a
+ * capture read off a posting page — it may be empty, it may carry a legal
+ * suffix one board prints and another does not, it may be spaced or cased
+ * differently between two postings for the same employer. So a company-scoped
+ * letter cannot point at a record; it carries a normalised string instead, and
+ * this is the one definition of that normalisation.
+ *
+ * **The key is advisory.** Per #765 it only ever drives a suggestion the user
+ * confirms ("You have a letter for Northwind — start from it?"). Nothing here
+ * auto-attaches a letter to a job, which is what makes a collision affordable:
+ * two employers that normalise alike cost the user a suggestion they decline,
+ * not a letter silently filed under the wrong company. Treat a match as
+ * evidence, never as identity.
+ *
+ * ## Why this is not `normalizeField` from `job-search/raw-postings.ts`
+ *
+ * That sibling (`raw-postings.ts:32`) does the lowercase + whitespace-collapse
+ * half of this and nothing else, in service of cross-provider dedup: two feeds
+ * printing one posting must not read as two. It is private to that module, has
+ * no legal-suffix handling, and lives in `job-search/` — a layer `storage/`
+ * sits below, so importing it here would invert the layering.
+ *
+ * They are also allowed to drift, which is the substantive reason not to share
+ * one: dedup compares a title and a company *from the same fetch* and wants the
+ * most literal comparison that still tolerates spacing. This compares a company
+ * name against one a user may have typed months earlier on another board, and
+ * wants a rougher key. Folding "Northwind" and "Northwind Inc." together is
+ * correct here and would be a bug there.
+ *
+ * Pure and zero-dep, like `score.ts` — no storage, no DOM, no imports.
+ */
+
+/**
+ * Trailing legal-entity suffixes stripped before keying — the list #766
+ * enumerates, no more.
+ *
+ * Matched whole-word against the ALREADY lowercased, punctuation-stripped token
+ * stream, so `Inc` and `Inc.` are one entry, `S.A.` arrives as the two-word
+ * tail `s a`, and `Corp` cannot half-match `Corporation`. The list is
+ * deliberately short and closed: every entry is a suffix that a job board is
+ * known to print inconsistently for the same employer. It is not a general
+ * company-name cleaner — "Northwind Technologies" and "Northwind" are two
+ * companies as far as this function is concerned, because nothing says they are
+ * not, and a key that over-merges is worse than one that under-merges (an
+ * under-merge shows no suggestion; an over-merge suggests the wrong letter).
+ */
+const LEGAL_SUFFIXES = [
+  "corporation",
+  "limited",
+  "gmbh",
+  "corp",
+  "inc",
+  "llc",
+  "ltd",
+  "pty",
+  "s a",
+  "b v",
+  "co",
+];
+
+/** Everything that is not a letter, a digit, or whitespace — replaced with a
+ *  single space, which IS a word boundary and has to be: `S.A.` has to split
+ *  into the two words `s a` for the two-word entry in {@link LEGAL_SUFFIXES} to
+ *  match it, while `Northwind,` just loses its trailing comma to the trim. */
+const PUNCTUATION = /[^\p{L}\p{N}\s]+/gu;
+
+/**
+ * Normalise a company name into a stable key, or `undefined` when there is no
+ * name to key.
+ *
+ * `undefined` — rather than `""` — is the answer for an empty or
+ * all-punctuation input, and that is load-bearing rather than tidy: `""` is a
+ * *refusal* under the letter contract, so a job whose `company` is blank must
+ * yield "no key" and not a key the record cannot legally carry. A caller can
+ * therefore spread the result straight onto a letter.
+ *
+ * Not exhaustive, and not trying to be — see {@link LEGAL_SUFFIXES} on why an
+ * under-merge is the cheaper failure.
+ */
+export function deriveCompanyKey(company: string | undefined): string | undefined {
+  if (company === undefined) return undefined;
+
+  const words = company
+    .toLowerCase()
+    .replace(PUNCTUATION, " ")
+    .split(/\s+/)
+    .filter((word) => word.length > 0);
+
+  // One suffix, not a loop: "Northwind Inc. Ltd." is not a company whose two
+  // suffixes should both come off, it is a name this function has no opinion
+  // about. Stripping repeatedly would also eat a real name — "Ltd Co" is a
+  // (bad) company name, and looping would leave nothing.
+  const stripped = stripTrailingSuffix(words);
+
+  // Every word was a suffix ("Inc."), or there were none to begin with. Either
+  // way there is no name left to key, and a key of `""` is not writable.
+  return stripped.length > 0 ? stripped.join(" ") : undefined;
+}
+
+/** `words` with a trailing {@link LEGAL_SUFFIXES} entry removed, or unchanged
+ *  when none matches. A suffix may be more than one word (`s a`), so this
+ *  matches on the joined tail rather than on the last token alone. */
+function stripTrailingSuffix(words: readonly string[]): readonly string[] {
+  for (const suffix of LEGAL_SUFFIXES) {
+    const parts = suffix.split(" ");
+    if (parts.length > words.length) continue;
+    const tail = words.slice(words.length - parts.length).join(" ");
+    if (tail === suffix) return words.slice(0, words.length - parts.length);
+  }
+  return words;
+}
--- a/src/lib/storage/letters.ts
+++ b/src/lib/storage/letters.ts
@@ -17,6 +17,22 @@
  * unreachable: {@link deleteLettersForJob} cascades from `deleteJob`, and
  * {@link clearLetterResumeLink} degrades a deleted résumé's link the way
  * `clearResumeLink` already does for jobs.
+ *
+ * Since #766 a letter is scoped to a job, to a company, or to nothing at all —
+ * the lattice on {@link LetterRecord.jobId} — and the three readers here
+ * ({@link lettersForJob}, {@link lettersForCompany}, {@link standardLetters})
+ * partition the store between them for any record the CONTRACT accepted.
+ *
+ * Not for one it did not. `saveLetter` validates nothing by design, so a
+ * both-keyed record — reachable only by skipping `validateLetterRecord`, which
+ * §3 of the contract doc requires a producer to call — comes back from
+ * `lettersForJob` AND `lettersForCompany`. These readers FILTER; they do not
+ * resolve. `useJobLetters`' `groupByScope` does resolve the same record, to
+ * job-only, and the difference is structural rather than a disagreement to
+ * reconcile: a map entry has to choose one bucket, an independent query does
+ * not. Teaching these three a precedence rule would mean encoding a reading for
+ * a shape the contract says cannot exist — the same defensive-second-copy
+ * mistake {@link deleteLettersForJob} warns against below.
  */
 
 import {
@@ -32,18 +48,25 @@ import type { LetterRecord } from "./types.ts";
  * Save a letter. Generates a UUID when `id` is absent; timestamps are managed
  * by `putRecord`.
  *
- * The input is a `Partial<LetterRecord>` narrowed to require `jobId` and
- * `body` — the two fields without which the record means nothing — rather than
- * the fully-specified shape `saveResume` takes. That is the same permissive
- * write `saveJob` makes, for the same reason: a stored letter may carry unknown
- * extra keys the contract PRESERVED on import, and a housekeeping write that
- * spreads a record back through here must not silently drop them.
+ * The input is a `Partial<LetterRecord>` narrowed to require `body` — the one
+ * field without which the record means nothing — rather than the fully-
+ * specified shape `saveResume` takes. That is the same permissive write
+ * `saveJob` makes, for the same reason: a stored letter may carry unknown extra
+ * keys the contract PRESERVED on import, and a housekeeping write that spreads
+ * a record back through here must not silently drop them.
+ *
+ * `jobId` left the required `Pick` in #766, when it became one of two optional
+ * scope keys rather than the record's spine. What replaced it is not a weaker
+ * requirement here but a different one elsewhere: the both-keys-set refusal is
+ * cross-field and lives in `validateLetterRecord`, which §3 of the contract doc
+ * requires a producer to call — `saveLetter` is the raw store wrapper and
+ * validates nothing, exactly as it did before.
  *
  * `touch: false` preserves `updatedAt` for a write the user did not make — see
  * `putRecord`.
  */
 export async function saveLetter(
-  input: Partial<LetterRecord> & Pick<LetterRecord, "jobId" | "body">,
+  input: Partial<LetterRecord> & Pick<LetterRecord, "body">,
   options: { touch?: boolean } = {},
 ): Promise<LetterRecord> {
   return putRecord<LetterRecord>(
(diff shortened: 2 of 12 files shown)


src/lib/storage/letters.ts
```diff
@@ -17,6 +17,11 @@
  * unreachable: {@link deleteLettersForJob} cascades from `deleteJob`, and
  * {@link clearLetterResumeLink} degrades a deleted résumé's link the way
  * `clearResumeLink` already does for jobs.
+ *
+ * Since #766 a letter is scoped to a job, to a company, or to nothing at all —
+ * the lattice on {@link LetterRecord.jobId} — and the three readers here
+ * ({@link lettersForJob}, {@link lettersForCompany}, {@link standardLetters})
+ * partition the store between them, each returning what no other one can.
```
s-annam: This partition claim holds only for a record `validateLetterRecord` accepted — `saveLetter` performs no validation, so a both-keyed record (reachable only by skipping the validator) is returned by both `lettersForJob` and `lettersForCompany`, unlike `useJobLetters.ts`'s `groupByScope`, which resolves the same case to job-only. Secondary, not blocking.

```suggestion
 * partition the store between them for any record the contract accepted (a
 * both-keyed record — reachable only by skipping `validateLetterRecord` — is
 * returned by more than one, since these readers filter rather than resolve).
```

src/lib/storage/company-key.ts
```diff
@@ -0,0 +1,116 @@
...
+  "ltd",
+  "pty",
+  "s a",
+  "b v",
+  "co",
+];
+
+/** Everything that is not a letter, a digit, or whitespace — dropped rather
+ *  than turned into a boundary, so `S.A.` collapses to `s a` and `Northwind,`
+ *  loses its comma without gaining a token. */
```
s-annam: "Dropped rather than turned into a boundary" contradicts its own example: `.replace(PUNCTUATION, " ")` replaces with a space, which IS a boundary — that's what makes `S.A.` collapse into *two* words `s a` rather than one `sa`. Nit, wording only — the code and tests are correct.

```suggestion
/** Everything that is not a letter, a digit, or whitespace — replaced with a
 *  single space, which IS a boundary: `S.A.` splits into the two words `s a`,
 *  while `Northwind,` just loses its trailing comma. */
```

s-annam (dismissed):
Well-designed, thoroughly-tested implementation of #766's lattice. The three-case scope model (job / company / standard), the both-keys refusal, the `deriveCompanyKey` normaliser, and the `useJobLetters` grouping fix (no more literal `undefined` map key) are all correct, and every acceptance criterion in #766 has a direct test asserting it — including the harder ones (delete-the-last-job-at-a-company leaves the company letter live, byte-identical export/import round-trip for company + standard letters, a v1-shaped record still validating unchanged). `npm run verify` (typecheck, lint, the five touched suites — 89 tests — and `fallow audit --base origin/main`) is green locally, and CI's `verify`/`fallow`/Cloudflare Pages checks are all green on the head commit.

## Blocking

None.

## Secondary

- `src/lib/storage/letters.ts`'s module docblock claims `lettersForJob` / `lettersForCompany` / `standardLetters` "partition the store between them, each returning what no other one can." That's only true for a record `validateLetterRecord` accepted. `saveLetter` performs no validation (by design, per its own docblock — "validates nothing, exactly as it did before"), so a record written with **both** `jobId` and `companyKey` set — reachable only by skipping the validator, but the validator is not enforced at the store layer — is returned by both `lettersForJob` **and** `lettersForCompany` simultaneously, unlike `useJobLetters.ts`'s `groupByScope`, which deliberately resolves that same case to job-only (with a dedicated test: "reads a both-keys record as a job letter"). Worth either softening the docblock's claim to scope it to contract-valid records, or giving `letters.ts`'s readers the same jobId-first precedence `groupByScope` already has, so the two consumers of the same store don't disagree on a malformed record. See the suggested wording below.
- `src/lib/storage/jobs.ts:109` — `deleteJob`'s docblock still says *"`LetterRecord.jobId` is a required parent link, so 'no letter outlives its job' is referential integrity, not tracker policy."* That's now false: `jobId` is one of three optional scope states since this PR. The surrounding rationale for *why the cascade lives in the store* still holds for job letters specifically (the cascade is correctly scoped — `lettersForJob` matches `jobId === id`, and a company/standard letter has none), but the sentence stating the premise needs a word change to say "job letters carry a parent link when scoped to one" rather than asserting every letter does. `jobs.ts` isn't in this diff's file list, so this is a pre-existing comment this PR's conceptual change invalidated rather than something the diff introduced — flagging so it doesn't linger as a landmine for the next reader of `deleteJob`.

## Nits

- `src/lib/storage/types.ts:358` — `LetterRecord`'s top-level docblock still opens *"A cover letter for one tracked job (#711)."* The accurate three-tier lattice is documented a few lines below on the `jobId` field itself, so the interface's own one-line summary now disagrees with its own body. Not reachable from the diff hunks (this line sits above where the diff starts), so no suggestion block — worth a one-line touch-up in a follow-up.
- `src/lib/storage/company-key.ts` — the `PUNCTUATION` doc comment says matched characters are "dropped rather than turned into a boundary," but the code (`.replace(PUNCTUATION, " ")`) replaces them with a literal space — which *is* a boundary, and is exactly what makes the comment's own example true (`S.A.` → two words `s a`, only possible because the period becomes a space between them). Purely a wording bug; the code and its tests are correct. Suggested fix below.
- `src/components/features/LetterEditorDialog.tsx:50` — its `jobId` doc comment still states the old store-wide claim ("a letter with no `jobId` is unreachable from every surface"), which since this PR is only true as this dialog's own local scoping choice, not a store-wide fact. This file is outside the diff (the PR's own description correctly scopes UI to the sibling issue), so this is FYI for whoever picks up the UI half rather than something to fix here.

## AC checklist (#766)

All eleven checkable acceptance criteria are met, each with a direct assertion: company letter save/read (`lettersForCompany`), standard letter save/read (`standardLetters`), both-keys refusal naming both fields, `jobId: ""`/`companyKey: ""` still refused, job-delete cascade scoped to job letters only (including the delete-the-last-job-at-a-company case, asserted directly), `deleteLettersForJob` idempotent with jobless letters present, company + standard letters surviving export→import→export byte-identically, `useJobLetters` never producing an `undefined`-keyed map entry, the company normaliser's convergence + `undefined`-on-empty behavior, `LETTER_CONTRACT_VERSION === 2` matching the doc header, and a v1-shaped record validating/storing unchanged. `npm run verify` green (locally and in CI).

Accurate throughout, including the update-comment about the fallow fix — the single commit's SHA (`148c40c`) matches what's actually on the branch, so the described "first push failed fallow, fixed and re-pushed" history is consistent with the repo's one-commit-per-PR invariant rather than a violation of it. The one thing I couldn't independently verify is the claimed status of the blocking `recruidea-extension#55` dependency — the PR is upfront that it couldn't check that either.

---

src/lib/storage/letters.ts
Samhit21: Fixed in `54ea31b` — you were right, the claim was overstated.

Took the "soften the docblock" branch rather than the "give the readers `groupByScope`'s precedence" one, and wrote down why: the two consumers differ *structurally*, not by oversight. A map entry has to choose one bucket, so `groupByScope` is forced to resolve; an independent query is not. Teaching `lettersForCompany` a precedence rule would mean encoding a reading for a shape the contract says cannot exist — which is the same defensive-second-copy mistake `deleteLettersForJob`'s own docblock warns against a few lines down.

The docblock now says the partition holds for any record the contract accepted, states plainly that a both-keyed record comes back from two readers, and explains the asymmetry so the next reader doesn't file it as a bug.

src/lib/storage/company-key.ts
Samhit21: Fixed in `54ea31b`, applied essentially as suggested. Straightforward wording bug on my part — the replacement is a boundary, and the `S.A.` example only works *because* it is one.

Phrased the replacement as load-bearing rather than incidental: the period has to become a space for the two-word `s a` entry in `LEGAL_SUFFIXES` to match at all.

Samhit21:
## Fix-up pushed — `54ea31b`

All four documentation findings addressed. Three were my own prose errors; the fourth was a comment this PR's conceptual change falsified. No behaviour changed — docblocks only.

| Finding | File | What changed |
|---|---|---|
| Secondary | `storage/letters.ts` | Partition claim scoped to contract-valid records, + why the readers don't resolve |
| Secondary | `storage/jobs.ts:109` | `deleteJob`'s stale "required parent link" premise |
| Nit | `storage/types.ts:358` | `LetterRecord`'s one-line summary now matches its own body |
| Nit | `storage/company-key.ts` | `PUNCTUATION` comment contradicted its own example |

Two notes on how I took the feedback:

**On `letters.ts` — took the docblock branch, not the precedence branch.** You offered either. I softened the prose rather than teaching `lettersForJob`/`lettersForCompany` the jobId-first rule `groupByScope` has, because I don't think the two consumers actually disagree — they differ structurally. A map entry must choose one bucket, so `groupByScope` is *forced* to resolve; an independent query isn't. Adding precedence here would encode a reading for a shape the contract refuses, which is the defensive-second-copy mistake `deleteLettersForJob`'s docblock argues against directly below it. The new text says all of that, so the asymmetry reads as a decision rather than an inconsistency. Happy to flip if you'd rather have the two agree literally.

**Also fixed `LetterEditorDialog.tsx:50`, which you scoped out.** You flagged it as FYI for whoever picks up the UI half. I did it anyway: the comment asserts "a letter with no `jobId` is unreachable from every surface", which *this PR* made false store-wide, and leaving a statement my own change falsified is worse than a three-line touch. It now says the prop is required because this dialog only composes job letters, and points at the sibling issue for the other two scopes. The prop and every behaviour are unchanged. Revert it if you'd rather keep the diff strictly to the file list.

Local re-check: `typecheck`, `lint`, five affected suites (101 tests), and `fallow audit --base origin/main` all clean. Re-requesting review.

s-annam (approved):
**Verdict: APPROVE** — 0 Blocking findings.

Read #766 first, then the diff, then the description last (per the review's own ordering rule). The three-case scope lattice (`jobId`/`companyKey`, exactly one set or neither) is implemented consistently end to end: `types.ts` docblock, the `isNonEmptyString`-when-present rule for both keys, the cross-field both-set refusal in `validateLetterRecord` (correctly placed after `checkDeclaredFields`, correctly reading `checked.*` rather than `value.*` so a field present-as-`undefined` counts as absent), `lettersForCompany`/`standardLetters` as new *filtering* readers (deliberately not resolving a both-keyed record — that's `groupByScope`'s job, and it resolves it to job-only, matching the v1 reading), and `deleteLettersForJob`'s scope-limited-by-construction cascade (verified directly: `letter.jobId === jobId` can never match a jobless letter, so a company/standard letter survives its job's deletion — and the last-job-at-a-company case is asserted, not just implied).

## Gates run
- `/code-review high`: launched as a background fork; it fanned out 8 finder sub-agents, announced it would wait for their results, then ended its own turn without ever synthesizing — the task-notification reported `completed` with no actual findings (146k tokens spent for nothing usable; filed as product feedback separately). Folded its intended scope into a manual line-by-line pass instead — every changed hunk in `types.ts`, `letter-contract.ts`, `company-key.ts`, `letters.ts`, `jobs.ts`, `useJobLetters.ts`, `LetterEditorDialog.tsx`, and `index.ts` read directly.
- 3a (fixture PII): N/A, no fixtures touched.
- 3b (design-system/reuse): N/A, no new components; `LetterEditorDialog.tsx`'s change is docblock-only. No raw `<button>`/dialog/input added.
- 3c (style tokens): clean — the grep's hex-pattern hits were all `#766`/`#765`/`#711` issue references, not colors.
- 3d (fallow): `npx fallow audit --base origin/main` → "No issues in 12 changed files"; the reported duplication is pre-existing, inherited, and report-only (2 small clones between `letters.test.ts` and `sync-schema.test.ts`, not attributable to this diff).
- 3e: N/A, no skill/script files touched.
- 3f (description accuracy): the description's AC checklist matches what the tests actually assert — verified each of the 11 checked boxes against a real test in `letters.test.ts`/`useJobLetters.test.tsx`/`company-key.test.ts`. One minor overclaim: "`LETTER_CONTRACT_VERSION === 2`, asserted against the doc header value" — the test (`letters.test.ts`) asserts `expect(LETTER_CONTRACT_VERSION).toBe(2)`, a hardcoded literal; it doesn't parse `docs/cover-letter-contract.md`'s header to cross-check it. Both sides are in fact `2`, so nothing is wrong, but the phrase implies a stronger check than exists. Nit, not blocking. The "blocking dependency" note on the extension sync issue is honestly flagged as unverified rather than glossed over — no overclaim there.

## Verification (local)
`typecheck` clean, `lint` clean, `letters.test.ts` + `company-key.test.ts` + `useJobLetters.test.tsx` (61 tests) + `db.test.ts` + `sync-schema.test.ts` (28 tests) all green.

## AC checklist (#766) — all 11 met, each with a direct test
Company-letter save/read, standard-letter save/read, both-keys refusal (names both fields), empty-key refusal (jobId **and** companyKey), job-only cascade (incl. last-job-at-company), idempotent cascade, byte-identical export/import round-trip, no `undefined`-keyed map entry, company-name normalisation (incl. `undefined` for empty/punctuation-only), `LETTER_CONTRACT_VERSION === 2`, v1-shaped record still validates.

## Nits (non-blocking)
- PR description: "asserted against the doc header value" for the `LETTER_CONTRACT_VERSION` test overstates what the test does (hardcoded `2`, not a doc-parse). Both values are correct; just tighten the claim next time.

SHA reviewed: `54ea31b55c04de5e12cb6ac4b6bb5735d6975187`. Branch already at one commit — no collapse needed.

---

PR #911, opened 2026-08-27
fix(skills): quote argument-hint flow sequence in pr-ready frontmatter (#850)

### LINKED ISSUE #850: pr-ready SKILL.md frontmatter is invalid YAML (unquoted argument-hint flow sequence)

shubhransh-gupta:
## Summary
Fixes #850.

Quotes the `argument-hint` value in `.claude/skills/pr-ready/SKILL.md` frontmatter so that it is parsed as a valid string rather than an invalid unquoted YAML flow sequence.

## Changes
- `.claude/skills/pr-ready/SKILL.md`: `argument-hint: "[<pr-number>[,<pr-number>...]]"`

## Verification
- Verified YAML frontmatter parses cleanly without syntax errors across strict YAML parsers.

### DIFF
--- a/.claude/skills/pr-ready/SKILL.md
+++ b/.claude/skills/pr-ready/SKILL.md
@@ -1,7 +1,7 @@
 ---
 name: pr-ready
 description: Solicit review on one or more open PRs — autonomously post a short ping asking for a single explicit 👍 by an absolute merge time (appended onto the most recent live ping if one exists, rather than a new standalone message), wait, send at most one reminder, and report per PR which of SILENT / ACKED / REVIEWING / REVIEWED was reached and by whom. Detects an in-progress review from GitHub (the 👀 /pr-review posts) and grants it one bounded grace extension. Carries no review guidance — that lives in the PR body. Never changes a PR's merge state. Use when the user says "pr-ready", "/pr-ready", "ask for review on this PR", "ping reviewers", or has open PRs nobody has looked at yet.
-argument-hint: [<pr-number>[,<pr-number>...]]
+argument-hint: "[<pr-number>[,<pr-number>...]]"
 ---
 
 # PR Ready


s-annam (approved):
**Verdict: APPROVE** (0 Blocking, 0 Secondary, 0 Nits)

Exact fix for #850: quotes `argument-hint` in `.claude/skills/pr-ready/SKILL.md` so it parses as a plain string instead of an invalid unquoted YAML flow sequence. Verified locally — `yaml.safe_load` on the frontmatter now parses clean (previously errored `expected ',' or ']', but got '['`).

## Gates
- 3a (fixture PII): N/A, no fixtures touched.
- 3b (design-system/reuse): N/A, no `src/components` touched.
- 3c (style tokens): N/A.
- 3d (fallow/dead-code): N/A, no exports touched.
- 3e (command-level bugs in skill file): change is a pure frontmatter quoting fix, no `gh`/bash commands altered — no new command bugs introduced.
- 3f (description accuracy): body is accurate; "Verified YAML frontmatter parses cleanly" checks out.

## AC checklist (#850)
- Quote the `argument-hint` value → done, matches the issue's suggested fix verbatim.
- Issue's "Suggested follow-up" (a mechanical frontmatter-parse check across all skills) is explicitly optional and out of scope for this PR — not required here.

## Note (out of scope, not a finding against this PR)
Ran the same YAML sweep across all `.claude/skills/*/SKILL.md` out of curiosity: `probe-jobs/SKILL.md` also fails to parse (`mapping values are not allowed here`) on `main` today. That's pre-existing and unrelated to this diff — worth its own issue if the follow-up sweep from #850 is ever picked up, but not this PR's problem.

Head reviewed: `9d5c4fb16d002bc89945fa56a5e23e29bb4ec4f9`

---

Issue #641, opened 2026-07-28
[parser] Pass B single-token city rule cleaves a company whose last word precedes a state tail (Palo Alto Networks, CA)


s-annam:
Flagged by @s-annam in the review of #634, explicitly so it would not be misattributed to that PR: *"That last row is a real defect but it is not this PR's — I A/B'd it against `main` before writing it up."* It has had no home since. Filing it now, with the A/B repeated against #636's branch.

## Defect

Pass B of `stripLocationSuffix` is a **single-token** space-delimited city rule: it takes the last whitespace-delimited token before the comma and calls it the city. When a company name's *last* word precedes a `, ST` tail, the company gets cleaved and a company word is promoted to a location.

```
Engineer · Palo Alto Networks, CA     → company="Palo Alto",   location="Networks, CA"
Engineer · Santa Clara Systems, CA    → company="Santa Clara", location="Systems, CA"
Engineer · Redwood City Ventures, CA  → company="Redwood City",location="Ventures, CA"
```

`Networks, CA` is not a location, and `Palo Alto` is not the employer.

## Not caused by the closed-vocabulary work, and confirmed by A/B

The vocabulary additions in #634 and #636 are irrelevant to this path. Reverting `Santa Clara|Redwood City|Ann Arbor` from `MULTIWORD_US_CITY_ALT` and re-running the same probes gives **byte-identical** output for the shredded rows:

| Header | Old vocab | New vocab |
|---|---|---|
| `Santa Clara Systems, CA` | `Santa Clara` / `Systems, CA` | `Santa Clara` / `Systems, CA` |
| `Redwood City Ventures, CA` | `Redwood City` / `Ventures, CA` | `Redwood City` / `Ventures, CA` |
| `Acme Santa Clara, CA` | `Acme Santa` / `Clara, CA` | `Acme` / `Santa Clara, CA` ✅ |

The vocabulary only ever *helps* here — the third row is a genuine improvement #636 delivered as a side effect (the vocabulary is shared with `KNOWN_MULTIWORD_US_CITY_RE`, the B-multi pass), and #636 pins it with a test. The defect is `KNOWN_MULTIWORD_US_CITY_RE` not firing when the city is **not** adjacent to the comma, so Pass B's single-token rule takes over.

## Why it is hard, and why it is not urgent

[...]

Issue #428, opened 2026-07-09
Download-export follow-up (#421 Secondary #11)


s-annam:
Split off from PR #421 review (Secondary #11, @Vaishnavi1709).

`useAnalyzedResume`'s `edited` memo re-runs `computeAnonymousAtsScore` whenever `addedProfiles` changes. After PR #421, adding a LinkedIn/GitHub link **should** re-score (it now moves completeness), so the naive "never re-score on profile change" fix no longer applies. The residual waste: adding a **non-scoring** profile (Behance, GitLab, …) still triggers a full re-clone + regrade that returns identical numbers.

**Proposed:** split the `profiles[]` mirror into its own memo (or a post-score assignment) so the score memo depends only on inputs the scorer actually reads. Cleanest done alongside the channel consolidation (#10).

`src/hooks/useAnalyzedResume.ts`.

PR #919, opened 2026-08-28
fix(preserve-numbers): give year classification a context gate (#876)

shubhransh-gupta:
## Summary
Resolves #876.

Adopts an **inverted 3-tier promotion architecture** for year classification in `preserve-numbers.ts`:
1. **Baseline Defense (Lenient)**: Every 4-digit number (1900–2099) defaults to `claim: "year_verb"` (claimed, lenient presence lookup). A cue miss *never* loses `main`'s baseline defense, eliminating the need to enumerate every English date phrasing.
2. **Strict Promotion Tier**: The cue gate (`isYearContext`) only ever **promotes** high-confidence temporal cues to `claim: "year"` (strict count parity).
3. **Strict Masking**: A strict year requires the output atom to also carry a strict year claim (`outputStrictYearKeys.has(atom.key)`), preventing noise digits from masking dropped years.

### Architectural Design
- **Drop Side (Rule 1)**:
  - **High-Confidence Temporal Cues (`claim: "year"`)**: Explicit temporal prepositions (`in 1900`, `since 2019`, `Jan 2021`), explicit temporal follow cues (`2019 onwards`, `2019 - Present`, `2019 to 2021`), leading award/prize names (`2019 Excellence Award recipient`), capitalized venue names (`at Google 2019`), and bullet-start date anchors (`2019: Started role`, `• 2019 - ...`). Evaluated against `outputStrictYearKeys` under strict count parity.
  - **Lenient / Un-cued 4-Digit Numbers (`claim: "year_verb"`)**: All other 4-digit numbers (1900–2099), weak prepositions, qualifiers, and bare past-tense verbs (`from 2019`, `by 2021`, `early 2019`, `mid-2020`, `summer of 2019`, `for 2020`, `2019 revenue grew 40%`, `Delivered 2000 units to production`). Evaluated leniently against `outputKeys.has(atom.key)`: complete drops are flagged as errors, while legitimate quantity merges (`Reduced by 2000 hours` + `Tracked 2000 bugs` $\rightarrow$ `Tracked 2000 bugs and reduced hours`) return `ok: true`.
- **Invention Side (Rule 3)**: All bare integers in the 1900–2099 range are audited against the input to prevent hallucinated numbers, while remaining lenient on legitimate temporal rewords (e.g. `"2019 Excellence Award"` $\rightarrow$ `"Award in 2019"`).
- **Residual / Accepted Trade-off**:
[...]

PR #922, opened 2026-08-29
refactor: extract the edit, résumé-format, and suffix-lexicon contracts

### LINKED ISSUE #649: Extract exporter ↔ parser round-trip contract into src/lib/resume-format
### Problem
The parser hard-codes details of what the exporter draws at multiple sites across different modules (like space headers, middots, and dashes), coupled only by prose. This makes round-tripping fragile.

### Affected Files
- `src/lib/pdf/ats-resume-model.ts`
- `src/lib/pdf/render-ats-pdf.ts`
- `src/lib/heuristics/entry-blocks.ts`
- `src/lib/heuristics/line-primitives.ts`
- `src/lib/heuristics/sections.ts`
- `src/lib/heuristics/experience-disambiguate.ts`

### Implementation Plan
1. Create `src/lib/resume-format/` holding named separator constants (like `MIDDOT_JOIN`, `ORG_COMMA`, `HEADER_WRAP_INDENT`) and paired compose/split functions (like `composeRoleHeader`/`splitRoleHeader`).
2. Update consumers across the codebase to use these constants and helper functions.
3. Separate PR: Unify the two date-range dialects (since this changes rendered bytes, it needs its own reviewed snapshot sweep).

### Acceptance Criteria
- [ ] `src/lib/resume-format/` exists with splitting and composing functions.
- [ ] Split∘compose identity contract tests pass.
- [ ] Extraction changes zero rendered bytes (round-trip tests remain green).

### LINKED ISSUE #652: Refactor applyOverrides signature to (base, snapshot)
### Problem
The `applyOverrides` function currently takes a highly fragile signature of 17 positional arguments, which is error-prone and has already drifted. We need to refactor it to accept a structured `(base, snapshot)` signature.

### Affected Files

### Implementation Plan
1. Extract the override-application, rescoring, and folding pipeline from `useAnalyzedResume` into a pure function in `src/lib/edit/`.
2. Change `applyOverrides` signature to `(base, snapshot)` where `snapshot` is `EditSnapshot`.
3. Consolidate `AnonymousAtsScoreInput` creation across all call sites onto a single constructor.
4. Preserve the `#428` invariant: ensure a non-scoring profile edit keeps the score object reference identical.

### Acceptance Criteria
- [ ] `applyOverrides` signature refactored to `(base, snapshot)`.
- [ ] `useAnalyzedResume.test.tsx` score-reference-stability tests pass.
- [ ] All tests compile and run successfully.

### LINKED ISSUE #917: Shared lexicon (c): one corporate-suffix token base + composer, preserving the four vocabularies' memberships
Part (c) of #653. Behaviour-preserving consolidation with one deliberate distinction to protect.

### Problem

Corporate-suffix vocabulary (`Inc`, `LLC`, `Ltd`, `GmbH`, …) exists as four separate sets across the parser. The #641 fix — "the captured token carries a trailing period" — had to be diagnosed and applied **per copy**, which is the tell.

### Implementation plan

1. Create one token base under `src/lib/heuristics/extract/` holding the suffix tokens as data.
2. Add a composer that generates each of the four vocabularies from it, with an `allowTrailingDot` option so the #641 behaviour is expressed once and applied where it belongs.
3. Repoint the four sites at the composer.

### Constraint — do not collapse the sets

Membership stays **per-set human judgement**. The tail-deferral vocabulary at `src/lib/heuristics/extract/experience-disambiguate.ts:188-199` is deliberately broader than `COMPANY_SUFFIX_RE`: it includes tokens like `Media` and `Partners`. Adding those to the strict set would flip `looksLikeTitle` false on a real title such as "Media Director".

So this issue shares the *token base* and the *composition mechanics*, not the memberships. Each generated set keeps its own explicit token list and a docblock saying what it is for and why it differs.

### Acceptance criteria

- [ ] One token base + composer under `extract/`; the four vocabularies are generated from it.
- [ ] `allowTrailingDot` (the #641 behaviour) is implemented once in the composer.
- [ ] Each generated set retains its own membership and a docblock explaining the divergence from the strict set.
- [ ] A test pins that `experience-disambiguate.ts`'s broader set still contains `Media`/`Partners` and that the strict set does not.
- [ ] Corpus snapshots byte-identical.
- [ ] `npm run verify` green.

Parent: #653. Roadmap item 5(c) of #646.

s-annam:
## Summary

Three seams that were coupled only by prose become named, tested contracts. No behaviour
changes — corpus snapshots are byte-identical and the exporter draws the same bytes.

- **#652** — `applyOverrides` took **19 positional parameters** and had drifted four times.
  It now takes `(base: EditBase, snapshot: EditOverrides)` with
  `EditOverrides = Partial<EditSnapshot>`, so the channel set derives from the persisted
  snapshot type and cannot drift from it again. The fold/re-score pipeline moves out of
  `useAnalyzedResume` into `src/lib/edit/edit-pipeline.ts`, and four hand-built
  `AnonymousAtsScoreInput` constructions collapse onto `scoreParsedResume`.
  `EditSnapshot`'s persisted shape is unchanged, so stored drafts and library records
  still replay.
- **#649** — the exporter↔parser separator contract moves into a zero-dep
  `src/lib/resume-format/`, with a `composeRoleHeader`/`splitRoleHeader` pair whose
  invertible domain is stated against *measured* production behaviour and gated by a test
  that runs the real `buildAtsResumeModel → renderAtsResumePdf → runCascade` leg. The
  date-range dialect is deliberately untouched — unifying it changes rendered bytes and
  needs its own snapshot sweep.
- **#917** — corporate-suffix vocabulary existed as four separate sets, which is why the
  #641 trailing-period fix had to be diagnosed and applied per copy. One canonical token
  base plus a composer now generates all four, with `allowTrailingDot` expressing #641
  once. The sets keep their own memberships: the tail-deferral set stays deliberately
  broader than the strict one, so "Media Director" still parses as a title.

Closes #652
Closes #649
Closes #917

## Review focus

- `src/lib/edit/apply-overrides.ts` — `Partial<EditSnapshot>` makes an omitted channel a
  silent no-op. Is there a call site that the old positional signature *forced* to pass a
  channel, which can now quietly skip it?
- `src/hooks/useAnalyzedResume.ts` — the `score` memo's dep list is byte-identical, but the
  fold memo now takes `snapshot` wholesale. Does that make it easier for a future channel to
  reach the fold without ever reaching the score? (One already does — see the
  `descriptionOverrides` note below.)
- `src/lib/heuristics/extract/corporate-suffix.ts` — the four sets now select tokens by key
  from a shared base. Do those key lists still express the same per-set human judgement the
  old free-text lists did, particularly where the broad set must stay broader?
- `src/lib/pdf/role-header-production-domain.test.ts:96` — the gate normalises `""` to
  absent, so its claim is "recovers exactly modulo `''`→`undefined`". Does that hide a real
  divergence on the empty-company dialect?

## Test plan

- [x] `npm run lint` clean
- [x] `npm run verify` exits 0 (full suite, build, packaging + fixture + baseline gates)
- [x] Full `npx vitest run` — 388 files / 6608 tests pass, re-run after rebasing onto `main` @ `583de21`, so the counts cover the merge result
- [x] Corpus snapshots byte-identical; no `UPDATE_FIXTURES` used
- [x] No fixture binary added or changed by this PR

## Adversarial review

Two rounds, pre-commit, on the accumulated tree. Round 1 split across two independent
reviewers on disjoint surfaces (React/hooks; parser/export lib); round 2 audited the fix.

### Round 1 — 2 blocking, both fixed

**1. `splitRoleHeader` documented an invertible domain that production contradicts.**
The module shipped with no production consumer, justified as "the executable spec the
heuristics approximate". Running the real leg (`buildAtsResumeModel` → `renderAtsResumePdf`
→ `runCascade`) over all 13 declared-invertible cases, 3 disagreed:

| case | spec said | production actually returns |
|---|---|---|
| title with a comma (`"Director, Marketing"` shape) | `{title:"Kilo Engineer, Sr.", company:"Globex"}` | `{title:"Kilo Engineer", company:"Sr.", team:"Globex"}` |
| team with a middot | `{team:"Payments · Risk"}` | `{team:"Payments"}` — `"· Risk"` dropped |
| untrimmed company padding | `"  Acme  "` verbatim | `"Acme"` |

Worse than dead code: the docblock and its test certified a live round-trip defect as safe,
and comma-bearing titles are a common real shape. **Fixed** by restating the domain against
measured behaviour (4 clauses → 6), moving the 3 cases into the "shapes the format loses"
block with production's real answers pinned as measured data, and adding
`src/lib/pdf/role-header-production-domain.test.ts` as a drift gate that runs the real
production leg. Round 2 confirmed the gate is not vacuous: all 13 rows assert (26 tests),
and under mutation it fails (separator change → 11/26 fail; flipped pin → 2/26).

**2. #917 AC1 was half-unmet — the composer shipped without a token base.**
AC1 asks for "one token base + composer; the four vocabularies generated from it", but only
`composeSuffixRegex` existed. Each call site still hand-typed free strings, so the same
concept was already spelled two ways across sets (`"l.l.c"` vs `"L.L.C"`) with nothing to
catch a typo — the exact class of drift the issue exists to end. **Fixed** with
`SUFFIX_TOKENS` (39 canonical entries) + `SuffixTokenKey` + `selectSuffixTokens`; all four
sets now select by key, so a typo is a compile error. Memberships are unchanged — the broad
tail-deferral set keeps `Media`/`Partners`/`Ventures`/`Bank` and the strict set still
rejects them, so "Media Director" still parses as a title.

### Verification of the behaviour-preserving claims

- **Byte-identical regexes** — the four composed `.source`/`.flags` values were recomputed
  independently from the current call sites and compared against literals from
  `git show HEAD:`, without reading the test's own goldens (a golden copied from new output
  would prove nothing). 4/4 match. Goldens were not edited.
- **#428 score-reference identity** — verified empirically, not by reading: six transitions
  driven through a probe component. Non-scoring extras keep the score object reference; a
  back-fill into an empty slot, a certification edit, and a link clear each mint a new one.
- **`score` memo dep array** — byte-identical to `HEAD`; the round-1 fix was comment-only.
- **Codemod correctness** — a positional→named comparator over all ~20 rewritten call sites
  (55 in `apply-overrides.test.ts` alone) found no argument in a wrong field.

### Surviving nits (not acted on)

- `experience-disambiguate.ts:29` — nothing pins `HEADER_DELIM_SPLIT_RE`'s spliced `.source`.
  Latent only; pinning it would mean exporting a module-private constant purely for
  testability. (Note the flag risk is narrower than it first looks: `new RegExp(src)` with no
  flags argument never inherits flags, so it cannot acquire `g`/`lastIndex` state.)
- `role-header-production-domain.test.ts:96` — `headerFieldsOf` normalises `""` to absent, so
  the row-level claim is "recovers exactly modulo `''`→`undefined`". Documented in-file.

### Pre-existing bugs found and deliberately NOT fixed (out of scope)

- **`descriptionOverrides` never re-grades.** It is not a `score` memo dep but writes
  `experience[].description`, which `score.ts` uses as the bullet-pool fallback for
  glyph-less résumés. Repro: edit a description → score object identical, overall stays 24;
  then touch any unrelated scoring channel → jumps to 94. Two further symptoms: `score.bullets`
  keeps pre-edit text, and `groupBulletsByExperience` on that stale pool returns
  `experienceIndex: null`, so the role's editable rows detach into "Other". Export is
  unaffected. Pre-existing and unchanged in behaviour by this PR, but **less detectable**
  after it — the two dep lists used to sit adjacent and diffable, so a missing channel was
  visible; now a new channel joins the fold automatically and the score only by hand. A
  comment in the `score` memo names the knowingly-absent channels to restore that signal.
- Five test harnesses hand-build an edited grade without `claimedBulletKeys`, bypassing
  `scoreEditedResume` whose point is that the key is required — so they simulate production
  with the #648 defect present. Pre-existing, untouched by the codemod.


src/lib/resume-format/index.ts
```diff
@@ -0,0 +1,30 @@
...
+export {
+  MIDDOT,
+  MIDDOT_JOIN,
+  MIDDOT_SPLIT_RE,
+  HEADER_DATE_GAP,
+  HEADER_WRAP_INDENT,
+} from "./separators.ts";
+
+export { composeRoleHeader, splitRoleHeader } from "./role-header.ts";
+export type { RoleHeaderFields, ComposedRoleHeader } from "./role-header.ts";
```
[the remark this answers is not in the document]
s-annam: Fixed in `b12b9c4`. `ComposedRoleHeader` is dropped from the barrel — nothing outside `src/lib/resume-format/` named it, and nothing needs to: callers of `composeRoleHeader` read `headerLine`/`subLine` off the returned object structurally, so the type name crosses no seam. The interface itself stays exported from `role-header.ts`, where the function's signature uses it.

`RoleHeaderFields` stays on the barrel — `src/lib/pdf/role-header-production-domain.test.ts:36` imports it from there.

The barrel now carries the same "deliberately not re-exported" note the `ORG_COMMA` case already had, so the omission reads as a decision rather than an oversight.

s-annam:
## Revision round — one thread, one line changed

`b12b9c4` (branch collapsed back to a single commit; the message is unchanged from the original).

**What changed:** `src/lib/resume-format/index.ts` no longer re-exports the `ComposedRoleHeader` type. That is the whole diff — 1 file, +4/-1, three of the added lines being a comment.

**Why:** code scanning flagged the re-export as never imported, and it reproduces — nothing outside `src/lib/resume-format/` names the type. Callers of `composeRoleHeader` read `headerLine` / `subLine` off the returned object structurally, so the *name* crosses no seam even though the *shape* does. The interface itself is still exported from `role-header.ts`, where `composeRoleHeader`'s return type uses it; only the barrel line went away.

`RoleHeaderFields` stays on the barrel — `src/lib/pdf/role-header-production-domain.test.ts:36` imports it from there.

The barrel now carries the same "deliberately not re-exported" note the `ORG_COMMA` case already had, so a future reader sees the omission as a decision rather than an oversight.

**Not in scope for this round:** nothing else was touched. No behaviour change, and the export is type-only, so nothing is emitted differently.

**Gates:** `npm run typecheck` clean · `npm run lint` clean · full `npx vitest run` 387 files / 6539 tests pass. Corpus snapshots untouched.

Review focus in the body is unchanged — this round answered a linter, not a reviewer, so none of the four open questions moved.

src/lib/heuristics/extract/corporate-suffix.test.ts
```diff
@@ -0,0 +1,264 @@
...
+ */
+
+import { describe, it, expect } from "vitest";
+import { composeSuffixRegex, selectSuffixTokens, SUFFIX_TOKENS } from "./corporate-suffix.ts";
+import { COMPANY_SUFFIX_RE, looksLikeTitle } from "./title-shape.ts";
+import { groupIntoLines, splitIntoSections, findSection } from "../sections.ts";
+import { extractExperience } from "../extract-fields.ts";
+import { mkItems } from "../__test-utils__/mkItem.ts";
+
+describe("composeSuffixRegex — byte-identical to the pre-#917 literals", () => {
```
s-annam: **Secondary — the byte-identity goldens don't bind to the four production constants.**

Every case in this block rebuilds its regex from a *hand-typed token-string array*, and the AC1 block below re-spells the *key* lists the same way. So the chain runs:

```
production key list  →  (nothing checks this link)  →  the pair this file pins
```

`LEGAL_SUFFIX_RE`, `COMPANY_TAIL_TOKENS_RE`, `COMPANY_LEGAL_TAIL_RE` and `LEGAL_TERMINAL_SUFFIX_RE` are module-private, so nothing ever compares the *real* constant against a golden. An edit to a `selectSuffixTokens([...])` call in `experience-disambiguate.ts` or `line-primitives.ts` changes the live regex with zero test signal.

**I verified this rather than asserting it.** I added `"GROUP"` to `LEGAL_TERMINAL_SUFFIX_RE`'s key list — materially widening a vocabulary whose docblock calls itself "deliberately narrow" — and ran the suites:

- this file: **25/25 pass**
- all of `src/lib/heuristics/`: **1682 pass, 8 skipped** — including `corpus.test.ts`, `corpus-roundtrip.test.ts` and the #459 `corpus-edit-roundtrip.test.ts`

(reverted; the tree is clean.)

The fifth case in this block already shows the fix — it asserts against the real exported `COMPANY_SUFFIX_RE`. The other four just aren't exported. Export them (or a `__testing__` bundle) and assert `re.source`/`re.flags` on the actual constants; then the goldens pin what the parser uses rather than what the test retypes.

Not blocking — all four are correct today, and I re-derived each `.source` by hand to confirm it. This is about whether the gate catches the *next* edit, which is the whole reason #917 exists.

src/lib/heuristics/extract/corporate-suffix.ts
```diff
@@ -0,0 +1,234 @@
...
+ * Build a legal-entity-suffix regex from a closed token list — normally one
+ * produced by {@link selectSuffixTokens}. Pure mechanics; see the module
+ * docblock for what stays a per-call decision (membership, anchor style, and
+ * which tokens tolerate a trailing period).
+ */
+export function composeSuffixRegex(
+  tokens: readonly string[],
+  options: ComposeSuffixOptions,
+): RegExp {
+  const dotSubset = Array.isArray(options.allowTrailingDot)
```
s-annam, line 200: **Secondary — an `allowTrailingDot` entry that isn't in `tokens` is silently inert.**

`dotSubset?.has(token)` matches by exact string, and nothing validates that the subset is actually a subset. The live trap is the `lowercase` option: `LEGAL_SUFFIX_RE` has to pass `{ lowercase: true }` to **both** `selectSuffixTokens` calls. Forget it on the `allowTrailingDot` one and every entry silently misses — no throw, no compile error, and `"Acme Inc."` quietly stops matching. That un-lands #641 for the whole set, which is the exact defect this module was built to make un-repeatable.

It's also inconsistent with the module's own argument. `SuffixTokenKey` is justified two screens up as the thing that "makes a mis-spelled token a compile error instead of a set that silently stops matching" — and then the dot subset is typed `readonly string[]`, which gives that property straight back.

The file already has the right pattern for this, ~15 lines below: `STATEFUL_FLAGS_RE` throws on a flag set that would fail silently. Same treatment here —

```ts
if (dotSubset) {
  const known = new Set(tokens);
  for (const t of dotSubset) {
    if (!known.has(t)) {
      throw new Error(
        `composeSuffixRegex: allowTrailingDot names "${t}", which is not in tokens — ` +
          "the dot would be silently dropped (check the `lowercase` option on both selects)",
      );
    }
  }
}
```

I checked all four current call sites: every dot subset is a genuine subset of its token list, so this throw cannot fire today. Non-blocking for that reason.

src/hooks/useEditableParse.removed-entry.test.tsx
```diff
@@ -139,24 +139,28 @@ interface Regraded {
...
-    api.summaryOverride,
-    api.removedEntries,
+    {
+      parsed: base.canonical.fields,
+      rawText: base.rawText,
+      sections: base.canonical.sections,
+      observations: OBSERVATIONS,
+      fieldConfidence: base.canonical.fieldConfidence,
+    },
+    {
```
s-annam: **Secondary — the harnesses that reproduce the production fold still hand-enumerate the channels.** `api.snapshot` is an `EditSnapshot`, which is assignable to `EditOverrides = Partial<EditSnapshot>` as-is, so this whole literal can just be `api.snapshot,`.

As written every one of these omits `certificationOverrides` (#884), and six of the seven also omit `removedEntries`:

| harness | `certificationOverrides` | `removedEntries` |
|---|---|---|
| `useEditableParse.removed-entry.test.tsx` (here) | missing | present |
| `useEditableParse.added-bullet-edit.repro.test.tsx` | missing | missing |
| `useEditableParse.added-bullet-remove.test.tsx` | missing | missing |
| `useEditableParse.remove-then-edit.repro.test.tsx` | missing | missing |
| `useEditableParse.duplicate-bullet-identity.repro.test.tsx` | missing | missing |
| `ExperienceSection.other-bullets.test.tsx` | missing | missing |
| `ExperienceSection.prune-hold.test.tsx` | missing | missing |

So a fold regression that drops the certifications channel is invisible to all seven — which is precisely the drift the whole-object argument in `EditOverrides`' docblock exists to remove. The production caller (`useAnalyzedResume.ts:173`) hands the whole snapshot; these claim to reproduce it and don't.

Left as a comment rather than pushed, because substituting `api.snapshot` changes what each harness exercises — it starts carrying a channel it didn't before — and that's a behavioural change to a test, which is yours to make rather than mine to slip into a review commit.

src/lib/edit/edit-pipeline.ts
```diff
@@ -0,0 +1,128 @@
...
+ * edit-pipeline.ts — the pure steps of the parse → edit → re-grade pipeline
+ * `useAnalyzedResume` drives on `/` (#652).
+ *
+ * The hook used to hold all four steps inline: build `applyOverrides`' base out
+ * of a `CascadeResult`, fold the overrides, probe which contact-link slots the
+ * scorer can actually see, and fold the edited fields back onto the base result
+ * for display. Only the second of those was ever a lib function. The other
+ * three were React-shaped only by accident of where they were written — they
+ * take values and return values — and keeping them in the hook meant the one
+ * that carries a real invariant ({@link probeScoringProfileSlots}) could only be
```
s-annam, line 14: **Nit — the stated reason for the extraction isn't cashed in: there's no `edit-pipeline.test.ts`.**

This docblock says the move happened because `probeScoringProfileSlots` "carries a real invariant" and that keeping it in the hook "meant [it] could only be tested by rendering." It is still only tested by rendering — `useAnalyzedResume.test.tsx` is the only thing exercising it.

The invariant is worth a direct test, and it's now trivially reachable: the function is pure, takes four strings, and the thing worth pinning is that it agrees with a real `applyOverrides` fold about which slots moved — a correction to `linkedin_url`, an extra that back-fills an empty `github_url`, and an extra that back-fills nothing.

Either add that file, or soften the claim to "is now testable without rendering." Non-blocking either way — the extraction is good on its own merits.

s-annam (commented):

0 Blocking. Rule applied: 0 Blocking → APPROVE, regardless of Secondary/Nits.

Reviewed at `b12b9c4`. This is a large refactor (46 files, +2803/−1169) whose whole claim is
that it changes no behaviour, so the review spent its effort on that claim rather than on
style. I re-derived the load-bearing equivalences by hand rather than reading the tests'
own goldens:

- **`composeRoleHeader` is a verbatim extraction.** `joinPresent` in `role-header.ts` is
  character-for-character the old `joinHeader` in `ats-resume-model.ts`
  (`parts.filter((p) => p && p.trim()).join(sep)`), and the #466 empty-company branch is
  transcribed unchanged. The default dialect's asymmetry — `title` trimmed, org fields
  joined verbatim — is preserved, not introduced.
- **All four composed suffix regexes are byte-identical to the pre-#917 literals.** I
  recomputed each `.source` from the key lists in `experience-disambiguate.ts` and
  `line-primitives.ts` without reading the test's goldens. 4/4 match, including the two
  different trailing-dot shapes (`COMPANY_TAIL_TOKENS_RE`'s single outer `\.?` vs
  `COMPANY_LEGAL_TAIL_RE`'s per-token `Inc\.?|Ltd\.?`). Membership therefore cannot have
  moved.
- **The `snapshot` collapse is re-run-equivalent.** `useEditableParse`'s `snapshot` memo
  (`useEditableParse.ts:1772`) has exactly the same fourteen deps the `editedCore` memo used
  to spell out, so `[base, doneScoreBullets, snapshot]` re-runs on exactly the same set of
  changes. `EditSnapshot` carries all fourteen channels; none is dropped.
- **`splitAchievementType`'s `idx + 3` → `idx + MIDDOT_JOIN.length`** is equivalent —
  `" · "` is 3 UTF-16 code units — and now cannot drift.
- **Gates.** `npm run verify` exits 0. I also ran the round-trip and contract suites
  directly, since `test:changed` can scope the run: `corpus-roundtrip` (68), the new
  `role-header-production-domain` gate (26), `render-roundtrip.repro` (5),
  `resume-format/*` (38), `corporate-suffix` (25) — 161 passed, 1 skipped. fallow is
  report-only and its 14 complexity findings are all inherited or pre-existing;
  `buildAtsResumeModel` in fact got *smaller* here.

### Acceptance criteria

| Issue | ACs | Result |
|---|---|---|
| #652 | 3 | **all met** — `(base, snapshot)` shipped; score-reference-stability tests green; `verify` 0. Plan steps 1/3/4 also done: the pipeline moved to `edit/edit-pipeline.ts`, and only two `computeAnonymousAtsScore` call sites remain in non-test `src/` (`score-cascade.ts` for a base parse, `score-edited.ts` for an edited one) — which is the documented pair, not a leak. |
| #649 | 3 | **all met** — module exists with the compose/split pair; identity contract green; zero rendered bytes confirmed by the corpus gate and by hand. Step 3 (date-range dialects) is deferred, which the issue itself scopes to a separate PR. |
| #917 | 6 | **all met** — token base + composer; `allowTrailingDot` once; per-set docblocks kept; the `Media`/`Partners` broad-vs-strict test is present; snapshots byte-identical; `verify` green. |

### Description accuracy (gate 3f)

**Accurate**, and unusually so — it discloses two pre-existing defects it deliberately does
not fix, with a repro for the first. Every checkable claim round-trips: the 19 positional
parameters (I counted them in the old signature), the byte-identical `score` dep array, the
four collapsed score constructions, the 4/4 regex match, "no fixture binary added or
changed" (the diff touches nothing under `tests/fixtures/`). One small correction below on
"No behaviour changes".

### Answers to the four `## Review focus` questions

1. **"Is there a call site the old positional signature *forced* to pass a channel, which
   can now quietly skip it?"** — No. Three channels did lose compile-required status
   (`contactOverrides`, `experienceOverrides`, `bulletOverrides` were positions 4–6 with no
   default), but each one's new default is `{}` — byte-identical to what a forced caller
   passed — so a skip is a no-op, not a behaviour change. There is exactly one production
   caller (`useAnalyzedResume.ts:173`) and it hands the whole memoized snapshot. Notably
   `observations`, the one field where a silent omission *would* corrupt (legacy numeric-key
   migration), was deliberately kept required on `EditBase` rather than defaulted. That is
   the right line to have drawn.
2. **"Does handing the fold `snapshot` wholesale make it easier for a channel to reach the
   fold without reaching the score?"** — Yes, and it already has. Filed as **#933**, with the
   repro from the body plus the `score.ts` line that makes it bite
   (`poolExperienceDescriptions` as the glyph-less bullet-pool fallback). The comment added
   to the `score` memo is the right mitigation for this PR; the fix belongs in its own change.
3. **"Do the key lists still express the same per-set judgement?"** — Yes, provably: the four
   `.source` values are byte-identical, so membership is unchanged by construction. See the
   Secondary finding below on what the *tests* do and don't pin about that.
4. **"Does `headerFieldsOf`'s `''`→absent normalisation hide a divergence on the
   empty-company dialect?"** — No. It collapses exactly one equivalence, and it is one
   `composeRoleHeader` itself already asserts (its blank check is `!fields.company?.trim()`,
   so `""` *is* "no company"). Any non-empty wrong value still fails the comparison — even
   `"   "`, since `pick` only special-cases the empty string. The normalisation prevents a
   false failure from a representation difference; it cannot mask a real one.

### Follow-up issues filed

- **#933** — [edit] A prose-description edit never re-grades the score (`descriptionOverrides`
  missing from the `score` memo's dep list).
- **#934** — [parser] A comma in a role title corrupts our own export → re-parse round trip.
  This PR *found* it and pinned it as measured data in `PRODUCTION_DIVERGENT_CASES`, which is
  exactly right — but "Director, Marketing" is a very common real shape and it loses a whole
  field, so it needs a tracked home outside a PR that is about to be squashed.

### Fixed in `12b9dec`

Three non-behavioural items applied directly, verified with `npm run verify` (exit 0, 387
files / 6539 tests), pushed, and then folded back into your original commit so the branch is
one commit again with your message unchanged. Nothing else was touched.

- `roundtrip-hop.ts` — `scoreForCascade`'s docblock claimed it was exported "so the edit-leg
  gate (#459) scores its override-applied `displayResult` through the exact same recipe."
  It isn't: `corpus-edit-roundtrip.test.ts:544` grades the edited model with
  `scoreEditedResume` and line 567 calls this only on the unedited `p1` — as do the other
  two callers (`corpus-roundtrip.test.ts:250`, `description-override-roundtrip.repro.test.ts:77`).
  Left as-is it contradicted the new `score-cascade.ts` docblock, which says the two recipes
  are separate precisely so nobody reaches the base one for edited input. Rewritten to say
  what it does and to point at `scoreEditedResume` for the edited case.
- `entry-blocks.ts:151` — `{@link MIDDOT_JOIN}` resolves to nothing; only `MIDDOT` is
  imported in that module. Spelled as code with the module named instead.
- `render-roundtrip-achievement-edit.repro.test.ts:89` — the codemod left the
  `achievementOverrides` literal at the old nesting level with an orphaned closing brace.
  Cosmetic; no formatter runs on it, so lint stayed green.

### Secondary

See the inline threads. In short: the new `corporate-suffix` byte-identity goldens don't
bind to the four production constants, so a key-list edit moves the real regex with no
failing test (I confirmed this by mutation); `composeSuffixRegex` guards one silent-failure
mode but not the adjacent one; and seven fold harnesses still hand-enumerate channels
instead of passing `api.snapshot`, which is the drift #652's whole-object argument exists
to remove.

### Nits

One inline, on `edit-pipeline.ts`: the module's stated reason for existing is that
`probeScoringProfileSlots` "could only be tested by rendering" — and it still is.

### Not filed as an issue, deliberately

The five-plus harnesses that hand-build an edited grade without `claimedBulletKeys`
(disclosed in the body) are a test-fidelity gap, not a shipping defect, and they belong to
#648's lane rather than a new tracking issue. Flagging here so the decision is on the
record rather than silently dropped.

### Ledger

- Threads opened: 4 (3 Secondary, 1 Nit). All are yours to answer — none is something I
  disposed of myself, so none is resolved. `main` has `required_conversation_resolution`,
  so **this PR is approved but not yet mergeable**: those 4 threads hold it.
- Follow-up issues filed: #933, #934.
- Branch: collapsed back to one commit (`b12b9c4` → `9f4e181` → `12b9dec`), tree
  `6b1efb9a065d288b933ac1d80ae0307db183e652` byte-identical across the collapse.

---

src/lib/heuristics/extract/corporate-suffix.test.ts
s-annam: Fixed in `460b680`. The four constants are now exported and the goldens assert on **them** rather than on a rebuild from a hand-typed list — `it.each` over `[name, actual, golden]` triples, importing `LEGAL_SUFFIX_RE` / `COMPANY_TAIL_TOKENS_RE` / `COMPANY_LEGAL_TAIL_RE` from `experience-disambiguate.ts` and `LEGAL_TERMINAL_SUFFIX_RE` from `line-primitives.ts`. Each export carries a docblock saying it exists for this gate.

Re-ran the same mutation to confirm the gate now bites: adding `"GROUP"` to `LEGAL_TERMINAL_SUFFIX_RE`'s key list now fails **exactly that row** and leaves the other three green — where before it left all 1682 tests in `heuristics/` passing. Reverted; `npm run verify` is green (388 files / 6549 tests).

Left the `selectSuffixTokens` AC1 block re-spelling its key lists: with the goldens bound to the real constants it is no longer load-bearing for drift, and what it still pins — that the base renders exactly those strings in that order — is worth keeping as its own statement.

src/lib/heuristics/extract/corporate-suffix.ts
s-annam: Fixed in `460b680`. `composeSuffixRegex` now rejects an `allowTrailingDot` entry that is not in `tokens`, in the same shape as the `STATEFUL_FLAGS_RE` throw directly below it, and the error names the token and points at the `lowercase` trap.

Verified it fires on the real failure rather than only on a synthetic one: dropping `{ lowercase: true }` from `LEGAL_SUFFIX_RE`'s `allowTrailingDot` selection now throws at module load with `allowTrailingDot names "Inc", which is not in tokens` — previously that produced a silently dot-less regex. Reverted.

Kept the type as `readonly string[]` rather than `readonly SuffixTokenKey[]`: the composer deliberately takes rendered tokens, not keys, so it cannot apply the `lowercase` option itself — the runtime check is what actually closes the hole.

src/hooks/useEditableParse.removed-entry.test.tsx
s-annam: Fixed in `460b680` for six of the seven; the seventh does not have the shape I assumed, and that is my error in the finding, not a deferral.

`useEditableParse.removed-entry`, `added-bullet-edit`, `added-bullet-remove`, `remove-then-edit`, `ExperienceSection.other-bullets` and `ExperienceSection.prune-hold` now pass `api.snapshot` / `edit.snapshot` in place of the hand-listed literal, so `certificationOverrides` (and `removedEntries`, missing from five of them) now reach the fold.

`useEditableParse.duplicate-bullet-identity` is the exception: its `fold()` takes `bulletOverrides` and `removedBullets` as plain function parameters and never sees an `EditableParse` at the call site, so there is no snapshot to pass — it is constructing a deliberately minimal fold, not reproducing the production one. Left as-is.

I hand-audited both `ExperienceSection` `useMemo` dep arrays since `exhaustive-deps` is not enforced here: both are `[edit]`, and `edit.snapshot` is reachable from `edit`, so the substitution introduces no stale closure. All six suites pass (107 tests across the eight files touched).

src/lib/edit/edit-pipeline.ts
s-annam: Fixed in `460b680` — added `src/lib/edit/edit-pipeline.test.ts` (10 tests) rather than softening the claim.

The part worth flagging: it does not assert a hand-written expectation of the probe's output. Each of the six cases runs a **real `applyOverrides` fold of the same override** and asserts `probeScoringProfileSlots(...)` equals the four slots read back off it — a correction, a clearing correction, an extra that back-fills an empty slot, an extra that does not back-fill, an extra on neither scoring slot, and the empty case. That is the actual invariant the module docblock claims: a probe that drifts from the fold leaves the score stale for the channel that drifted, and a golden written by hand would drift with it.

Also pinned: the probe does not mutate the parse it is handed (`useAnalyzedResume` passes `base.canonical.fields` directly), and it returns exactly four keys — a fifth would be a `score` dep nobody added.

s-annam:
## Revision round — all four review threads addressed

**Head is now `9108436`.** The four thread replies above cite `460b680`; that was the head
when they were written, and a second collapse replaced it after I caught two problems in my
own fix (below). The content is the same plus those two items — `460b680` is not on the
branch any more, so follow `9108436`.

### What changed

| Thread | Done |
|---|---|
| suffix goldens don't bind to production | The four constants are exported; the goldens assert on **them** via `it.each`, not on a rebuild from a hand-typed list. |
| `allowTrailingDot` silently inert | `composeSuffixRegex` now throws on a subset entry absent from `tokens`, in the same shape as the existing stateful-flag throw. |
| harnesses omit `certificationOverrides` | Six now pass `api.snapshot` / `edit.snapshot`. The seventh (`duplicate-bullet-identity`) has no `EditableParse` at its fold call site — my finding over-counted; left as-is, explained on the thread. |
| `edit-pipeline` has no unit test | Added `edit-pipeline.test.ts` (10 tests), asserting `probeScoringProfileSlots` against a **real `applyOverrides` fold** of the same override rather than against a hand-written expectation. |

### Both new gates were checked by mutation, not assumed

The complaint was that the old gate didn't bite, so it wasn't enough to add one and assume:

- Adding `"GROUP"` to `LEGAL_TERMINAL_SUFFIX_RE`'s key list now **fails exactly that golden
  row** and leaves the other three green. Before: all 1682 tests in `heuristics/` passed.
- Dropping `{ lowercase: true }` from `LEGAL_SUFFIX_RE`'s dot-subset selection now throws at
  module load naming the token. Before: it silently produced a dot-less regex, un-landing
  #641 for that whole set.

Both reverted; the second is now a committed test rather than only a mutation run.

### Two things I got wrong in my own fix, caught on a re-read

Worth recording, since they are the same defect class the review was about:

- The guard shipped with **no committed test** — only the mutation run above. Three now cover
  it, including the `lowercase`-mismatch case that is the actual trap.
- `corporate-suffix.test.ts`'s header still described section 1 as rebuilding each regex from
  a hand-typed token list, which is exactly what it had stopped doing. Rewritten to say what
  the file now does, including why section 3 is *not* a drift gate and why the fifth golden
  row stays a rebuild (`COMPANY_SUFFIX_RE` is hand-written by design, so there is no composed
  constant to compare).

### Gates

`npm run verify` exit 0 — 388 files / 6552 tests, build green, fixture + baseline + packaging
gates green. fallow is report-only and reports **dead exports 0.0%**, so the four new
test-only exports are not forward-staged. Tree identity held across both collapses.

Not touched: #933 and #934 stay open as filed — neither is this PR's to fix.

src/hooks/useEditableParse.duplicate-bullet-identity.repro.test.tsx
```diff
@@ -129,20 +129,18 @@ function fold(
...
-    [],
-    base.canonical.fieldConfidence,
+    {
+      parsed: base.canonical.fields,
+      rawText: base.rawText,
+      sections: base.canonical.sections,
+      observations,
+      fieldConfidence: base.canonical.fieldConfidence,
+    },
+    {
```
Samhit21, line 139: [Suggestion]: This is the seventh harness from your own table, and the fix reached six — the count went from seven to six with no stated reason.

Your thread on `removed-entry.test.tsx` enumerated seven harnesses that hand-list their channels. Six now pass the whole snapshot (`api.snapshot` in four, `edit.snapshot` in the two `ExperienceSection` files); this one still passes a three-key literal, and the fix-up commit message said "Six fold harnesses hand-listed their override channels" without naming the omission.

The substitution is available here. This file has a live hook — `api = useEditableParse()` at :179, and `regrade()` already reads `api.bulletOverrides` / `api.removedBullets` off it at :197 — so `api.snapshot` is in scope exactly as it is in the six that changed.

By your own table this harness is missing **both** `certificationOverrides` and `removedEntries`, which is the worst row of the seven. And a partially-applied fix is worse than a listed gap: the table now reads as discharged, so the next person to add a channel has no reason to re-check it.

If it was excluded deliberately — `fold()` takes its channels as parameters rather than reading the hook, and `removedBullets: [...removedBullets]` is converting a `Set`, so this isn't a pure substitution the way the other six were — that's a fine answer. But it should be the seventh row of the table saying so, rather than a silent decrement.

Samhit21 (commented):
## PR Review: refactor: extract the edit, résumé-format, and suffix-lexicon contracts

### Summary

The self-review is more thorough than most external ones, so I spent this review trying to break its load-bearing claims rather than restating them. I re-derived four independently and all four hold:

- **`LEGAL_TERMINAL_SUFFIX_RE` is byte-identical** to the pre-#917 literal — checked `.source` and `.flags` against the old regex directly, not against the test's goldens.
- **Review-focus Q1 is answered "no".** There is exactly one non-test `applyOverrides` call site in `src/` (`useAnalyzedResume.ts:173`) and it hands the whole `snapshot`. No production channel can be silently skipped by the new signature; the test harnesses were the only real instance, which is the thread you opened yourself.
- **The `score` memo's dep list is complete.** Walked all fourteen `EditSnapshot` channels against it: twelve present, `profileOverrides` deliberate per #428, `descriptionOverrides` the known #933 bug. The "CHANNELS KNOWINGLY ABSENT" block accounts for both. Nothing is unlisted.
- **The golden fix actually closes the hole**, verified the same way you opened it: adding `"GROUP"` to `LEGAL_TERMINAL_SUFFIX_RE`'s key list now fails `corporate-suffix.test.ts` (1 of 25), where before it left all 1682 tests in `heuristics/` green.

I also chased two things that turned out to be nothing, worth recording so nobody re-chases them: the `MONTH` / `SEASON` / `OPEN_ENDED_ALT` inlining in `entry-blocks.ts` and `line-primitives.ts` is `main`'s change, not this PR's — it only looks like this PR's if you diff against `origin/main` instead of the merge base. And `splitRoleHeader`'s title-less hole (`compose({team, location})` → `{title: team}`) is genuinely covered by domain clause 1.

### Findings

1. `useEditableParse.duplicate-bullet-identity.repro.test.tsx:139` — the seventh harness from your own table; the other six now pass the whole snapshot and this one still hand-lists three channels. [Suggestion]

2. **The branch is four commits behind `main`, and `main` has since rewritten two of the files this PR rewrites.** `entry-blocks.ts`, `line-primitives.ts` and `corpus-edit-roundtrip.test.ts` all moved on `main` after the merge base (`2534afe`), the last of those substantially — #927 landed `markedRole` / `editedRoleDateText` in it. So `verify` green, and the "387 files / 6539 tests" count, are measured on a tree that is not the one that will merge.

   I merged `origin/main` into the branch and re-ran rather than leaving that as a worry: **the merge is clean, `tsc -b` passes, and the affected suites are green (7 files, 242 tests).** The overlap is region-disjoint in all three files. So this is not a defect — but the test-plan numbers are worth re-stating post-rebase, because as it stands the PR's own evidence doesn't cover the merge result. [Nit]

### Your four review-focus questions

**Q1 — a call site the positional signature forced to pass a channel?** No, in production. One call site, whole snapshot. The seven harnesses were the real answer and you found them; finding 1 is the leftover.

**Q2 — does `snapshot` wholesale make it easier for a channel to reach the fold but not the score?** Yes, and your comment block already says so in the right place. The mitigation is load-bearing but hand-maintained, so it's worth knowing what it rests on: the block is accurate today (I checked all fourteen), and its "anything else absent is unaccounted for" line is what makes it auditable in one pass. That's about as good as a hand-maintained invariant gets without a test. A cheap hardening, if you ever want it: a type-level exhaustiveness check that fails to compile when `EditSnapshot` gains a key not named in a local `SCORE_DEPS` / `KNOWINGLY_ABSENT` union. Not for this PR.

**Q3 — do the key lists preserve per-set judgement?** Byte-identity answers this mechanically and I confirmed one of the four by hand. The judgement question is now pinned by construction: since the goldens assert the real constants, any membership change has to move a stated expectation.

**Q4 — does `''`→absent hide a divergence on the empty-company dialect?** Agreed with your answer, and the reason is stronger than "it collapses one equivalence": `composeRoleHeader`'s own blank check is `!fields.company?.trim()`, so `""` and absent are the same input to the composer. The gate cannot distinguish states the composer cannot distinguish either. `"   "` still fails, as you say.

### Verdict

**Action: COMMENT.** No blocking findings. Finding 1 is a real gap in a fix that reads as complete; finding 2 is housekeeping before merge. All five threads are resolved, the branch is back to one commit, and all three checks are green on `9108436` — this is close.

src/hooks/useEditableParse.duplicate-bullet-identity.repro.test.tsx
s-annam: You are right and my reply on the sibling thread was wrong. I wrote that this harness "has no `EditableParse` at its fold call site" — line 197 is `fold(LINES, api.bulletOverrides, api.removedBullets)`, so the call site is reading off `api` already. `api.snapshot` is in scope there exactly as you say. I checked the *function signature* (`fold()` takes its channels as parameters) and stopped, which answered a different question than the one I had asked.

The silent seven-to-six decrement is the fairer half of the finding. My own comment set the count and the fix-up message quietly used the smaller one, so the table reads as discharged when one row is not.

Both of your findings verified independently before agreeing:

- `api` at module scope (`:175`), assigned `:179`, read at the `fold` call site `:197`. Yours.
- Branch is `ahead=1 behind=4`. `main` moved exactly the three files you name since the merge base `2534afe` — `entry-blocks.ts`, `line-primitives.ts`, `corpus-edit-roundtrip.test.ts` — across `#927`, `#919`, `#924`, `#926`. So the test-plan numbers really are measured on a tree that is not the merge result.

Not applying either yet — the run that opened these threads has reached its terminal state and the next branch-rewriting round is the author's call, not something I should slip in under a non-blocking review. Leaving this thread open for that; it is the accurate state, and under `required_conversation_resolution` it is also the correct one.

Thank you for re-deriving the four claims rather than restating them, and for recording the two dead ends — the `MONTH`/`SEASON` inlining being `main`'s rather than this PR's is exactly the kind of thing that costs the next reader an hour.

s-annam:
Autopilot: **not merge-ready** after 1 bot round and 1 human cycle — stopping here rather than
reporting a clean finish.

Reviewed by @Samhit21 (`COMMENT`, no blocking findings), branch at one commit `9108436`, all
three checks green, no conflicts. Two things still stand between this and the queue, and
neither is mine to decide:

- **No approving review.** `reviewDecision` is `REVIEW_REQUIRED` — a `COMMENT` review does not
  satisfy the branch protection, and the earlier `APPROVE` on this PR was a self-review, which
  GitHub posts as `COMMENT` and does not count.
- **One unresolved thread** (`duplicate-bullet-identity.repro.test.tsx:139`), which
  `required_conversation_resolution` makes a merge gate. It is open on purpose: the finding is
  correct, I have not applied it, and resolving it to clear the path is precisely what neither
  side of this loop is allowed to do.

Both of that review's findings were re-verified here before being accepted, and **one of them
corrects me**: my reply on the sibling thread claimed this harness has no `EditableParse` at
its fold call site, and line 197 is `fold(LINES, api.bulletOverrides, api.removedBullets)`.
The correction is on the thread. The second finding — the branch is `behind=4`, and `main`
moved `entry-blocks.ts`, `line-primitives.ts` and `corpus-edit-roundtrip.test.ts` after the
merge base across #927/#919/#924/#926 — also holds, so this PR's test-plan numbers are
measured on a tree that is not the one that will merge.

Filed during the bot round, neither this PR's to fix: #933 (a prose-description edit never
re-grades the score) and #934 (a comma in a role title corrupts our own export → re-parse
round trip).

The wait was session-bound; nothing is watching this thread now. Re-running `/pr-autopilot 922`
picks up from wherever the PR actually is.

src/hooks/useEditableParse.duplicate-bullet-identity.repro.test.tsx
s-annam: Applied in `70c05f8`. `fold()` now takes the whole `EditSnapshot`, and `regrade()` passes `api.snapshot`, so all seven harnesses pass the whole snapshot. The claimed-bullet keys for `scoreEditedResume` are derived from `snapshot.bulletOverrides` and `snapshot.removedBullets`.

I also rebased onto `main` @ `583de21`, as your second finding asked. The branch is still one commit. `npm run verify` exits 0 on the rebased tree with the full suite: 388 files, 6608 tests. The test plan in the PR body now states those counts.

Samhit21 (approved):
## PR Review: refactor: extract the edit, résumé-format, and suffix-lexicon contracts

### Summary

Re-review on `70c05f8`. Both findings from my earlier review are resolved, and I checked the new head directly rather than going by the replies.

- **Finding 1 (the seventh harness) is fixed.** A range-diff against `9108436` shows this PR's own content changed in exactly one file: `useEditableParse.duplicate-bullet-identity.repro.test.tsx`. `fold()` now takes the whole `EditSnapshot`, and `regrade()` passes `api.snapshot`. The claimed-bullet keys for `scoreEditedResume` come from `snapshot.bulletOverrides` and `snapshot.removedBullets`, the same two maps `applyOverrides` folds, so all seven harnesses now pass the whole snapshot.
- **Finding 2 (behind `main`) is fixed.** The branch is rebased onto `583de21`, is still one commit, and is 0 behind. Every other difference in the range-diff is `main`'s own history.

### Verification on this head

- `tsc -b --noEmit` exits 0.
- Every edit, fold and round-trip suite (`useEditableParse*`, `ExperienceSection*`, `apply-overrides*`, `edit-pipeline`, `description-override-roundtrip`, `render-roundtrip*`, `corpus-edit-roundtrip`, `undo-batch`) passes: **33 files, 415 tests**, including all 15 in the seventh harness.
- `corporate-suffix.test.ts` passes (28 tests).
- CI is green: `verify`, `fallow` and Cloudflare Pages. All 6 review threads are resolved.

The four review-focus answers from my earlier review still hold. Nothing in the delta touches the `score` memo's dep list, the suffix key lists, or the empty-company composer.

### Verdict

**Action: APPROVE.**

Issue #912, opened 2026-08-27
Feedback session UX: research + propose flow for the always-visible feedback widget


s-annam:
## Context

From the 2026-08-27 Intern Weekly sync: the feedback widget (`FeedbackPanel.tsx`) was recently changed to render at all times post-parse instead of only in specific situations. The star rating → email/comment/refinement flow works, but we haven't done real UX thinking on the *session* itself — when it shows, how intrusive it feels across a visit, and whether the flow after a 5-star vs. a 1-star rating should diverge.

Current implementation for reference:
- `src/components/features/FeedbackPanel.tsx` — the panel itself. Render-state machine: `full` (first two mounts) → `compact` (quiet inline star strip after that) → `done` (once submitted). Driven by two `localStorage` flags (`ocv_feedback_seen`, `ocv_feedback_submitted`) via `usePersistentFlag`/`usePersistentCounter`.
- Fires a single `feedback_submitted` PostHog event (`src/lib/analytics.ts`) — anonymous unless the user opts in with an email.
- Rendered inside `Result.tsx`, directly under the score card.

## Problem

We don't have a strong opinion yet on whether the *current* visible-always behavior is actually the right session design — it was a quick change, not a researched one. Things worth interrogating:
- Does showing it on every single mount (even before "compact" kicks in) feel naggy on a repeat visit in the same session?
- Should the flow branch by rating — e.g. a low rating should probably route toward "what went wrong" free text before anything else, where a high rating could skip straight to the GitHub-star CTA?
- Is `localStorage`-only state the right memory model, or should "seen"/"submitted" reset per résumé rather than per browser?
- Any accessibility/focus concerns with a panel that appears unprompted post-parse.

Some starting ideas — not a spec, just angles worth checking:
- A lighter first touch (e.g. just the star row, no card chrome) that only expands into the full form once a rating is picked, rather than showing the full amber `Card` immediately.
- Branching copy/flow by rating value instead of one static form for every rating.
- Consider whether "session" should mean per-résumé-parse rather than the current seen-count-across-mounts model.

## What to do

This is intentionally open-ended — we don't have enough detail yet to hand over a fixed spec. First task is **not** implementation:

1. Look at the current `FeedbackPanel` behavior in the running app (both `full` and `compact` states, and the post-submit thank-you state).
2. Do a short pass of UI/UX research — comparable feedback-widget patterns elsewhere are fair game — and form your own opinion on what should change and why.
[...]

Issue #900, opened 2026-08-26
Feedback control gets too quiet after 2 views — testers can't find/recognize it


s-annam:
## Problem

Testers and organic users find it difficult to locate or recognize the feedback control on the results page. The previous inline `FeedbackPanel` rendered below the score card and demoted to a quiet, unbordered `-mt-8` star strip after 2 mounts (`ocv_feedback_seen >= 2`). This resulted in low organic feedback conversion because:
1. Asking for feedback immediately on mount (1st/2nd visit) is premature before the user has reviewed or exported their resume.
2. The compact inline strip lacked clear interactive affordances and was frequently missed or mistaken for static footer metadata.
3. Delighted users had no clear pathway to support the project (e.g. starring the GitHub repo), while users with praise or constructive critiques had no tailored prompt.

## Solution

Replace the inline `FeedbackPanel` with a **Multi-Step Feedback Interstitial (`FeedbackDialog`)** triggered at key value milestones and accessible via an ambient trigger button.

### 1. Trigger Pathways & Lifecycle
- **Automatic Milestone Trigger**: Automatically opens once after the user's **1st successful export** (PDF or Markdown via `ExportDialog` / `useDownloadPdf`).
- **Ambient Trigger**: An accessible `[★ Feedback]` button in `ParsedHeader` (and/or footer) that allows users and testers to open the feedback interstitial on demand at any time.
- **Persistence & Capping**:
  - `ocv_feedback_submitted`: Once submitted, the automatic prompt is suppressed.
  - `ocv_feedback_seen`: Tracks views; dismissed prompt is snoozed.
  - Suppressed entirely when analytics are disabled (`ANALYTICS_ENABLED === false`).

#### **Step 1: Quick Star Rating (1–5 ★)**
- Prominent 5-star rating control (`StarRating` primitive with 44×44px touch targets).
- Title: *"How did your resume turn out?"*
- Selecting a rating immediately transitions to Step 2 with sentiment-tailored routing.

#### **Step 2A: Positive Sentiment (4–5 ★) — "Delight & GitHub Star"**
- Title: *"Thank you! We're glad it helped."*
- Explains the open-source mission: *"OfflineCV is free, private, and runs in your browser. Starring our repo helps other job seekers discover it."*
- **GitHub Star CTA**: Embedded `GitHubStarCta` showing live star count with a direct link to the GitHub repository.
[...]

PR #923, opened 2026-08-30
feat(feedback): inline nudge instead of an auto-opening modal, cooldown instead of a lifetime cap

Samhit21:
Implements the two changes recommended in the research pass on #912 ([writeup](https://github.com/offlinecv/OfflineCV/issues/912#issuecomment-5464174007)).

> ⚠️ **#912 says "don't open a PR yet"** — it asks for research first, then a scoped discussion with @s-annam. This PR exists because the work was explicitly requested after the writeup was posted. Treat it as a concrete proposal to react to rather than an agreed scope; the two open product calls below are deliberately **not** in it.

## Context: the issue's premise was stale

#912 was written against `FeedbackPanel.tsx`, which no longer exists — #900/#903 retired it after the Aug 27 sync. Three of the issue's five questions were already answered by that change (no longer always-visible, already branches by rating, already a bare star row that expands). What was left is what this PR does.

## 1. The milestone raises a nudge, not the dialog

`Dialog` is a native `showModal()`. Opening it automatically took keyboard focus off whatever the user was doing and announced itself to a screen reader as an interruption — the wrong instrument for an invitation. This is also the answer to the issue's accessibility question, and the answer had got **worse** with #900, not better: the old concern was a panel appearing below the fold, the new one was a focus trap arriving uninvited.

NN/g's [User-Feedback Requests](https://www.nngroup.com/articles/user-feedback/) endorses asking after task completion and keeping an always-available way in — both of which #900 got right — while explicitly discouraging modal popups for the ask itself.

So a résumé export now shows `FeedbackNudge`: inline, dismissible, `role="status"` (polite, so it waits behind whatever the reader was saying about the export), the star row plus one line. Picking a star opens the dialog **on that rating's branch** via a new `initialRating` prop, so nobody is asked for the same star twice.

Side effect worth noting: this nearly retires the two-moment `notifyResumeExported` / `notifyExportClosed` dance. A non-modal nudge can't stack on `ExportDialog`, so the split now exists only so the nudge doesn't appear *behind* it, unseen.

## 2. Cooldown instead of a lifetime cap

The automatic ask was gated on `seenCount > 0` — shown once, ever. And `openDialog()` incremented that same counter, so **a user who clicked the ambient `[★ Feedback]` button out of curiosity permanently disabled the automatic ask.**

| | before | after |
|---|---|---|
| Automatic ask | once per browser, forever | every 14 days |
| Ambient open | burns the automatic trigger | untouched |
| After submitting | never auto-asked again | unchanged |

`ocv_feedback_prompted_at` records only unprompted asks, written when the nudge becomes **visible** rather than when it's answered — a nudge scrolled past is still an ask received, and re-asking tomorrow because it was ignored is the nagging this exists to prevent.

**Migration is a deliberate one-time re-ask.** Neither `ocv_feedback_dialog_seen` (#900) nor `ocv_feedback_seen` (the retired panel) is read. Every returning tester holds a value under both, and honouring them would keep the lifetime cap alive for exactly the people this change is for — so those browsers become eligible once, then fall under the cooldown like everyone else.

**The cooldown fails open.** An unreadable timestamp asks rather than staying silent: a corrupt key that locks someone out of ever being asked is this issue's own bug reached by another route.

## Deliberately not in scope

- **Which milestone earns the ask.** Export is the rarest success moment in the funnel — anyone who parses, reads their score, edits and leaves is never asked. Changing it is a product call, not an implementation detail.
- **Per-parse vs per-browser memory.** The issue's question 3, still open. Falls out cheaply if the cooldown key incorporates `parseKey`, but "is a new résumé a new experience to rate" is a decision, not a refactor.
- **A dismissed rating is lost entirely** — pick 2★, close without typing, nothing is recorded. That's the strongest signal we have, discarded from the users least likely to write a paragraph. Fixing it changes the `feedback_submitted` event shape, which #912 puts out of scope; flagged in the writeup as wanting its own issue.

## Verification

`npm run verify` green: **384 test files, 6464 passed**, 10 skipped, build clean, `fallow audit --base origin/main` exit 0.

19 new tests. The ones with teeth:
- the milestone raises the nudge and **not** the dialog;
- the ambient button does **not** write the cooldown key;
- suppressed inside the window **and** asks again past it — without the second case the first would pass against a permanent lock;
- both retired keys are ignored and left untouched;
- an unreadable timestamp asks;
- the nudge is not a `dialog` and has no `aria-modal` — if that ever regresses, this PR has been undone.

One fallow note: `App.tsx`'s `App` function trips the complexity threshold at 787 lines. That's pre-existing debt (CLAUDE.md's known-debt list); this PR adds 8 lines to it and doesn't touch the shape. The gate passes.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,6 +21,7 @@ import { ShareWithExtensionBar } from "./components/features/ShareWithExtensionB
 import { ExportDialog } from "./components/features/ExportDialog.tsx";
 import { ResumeChooserDialog } from "./components/features/ResumeChooserDialog.tsx";
 import { FeedbackDialog } from "./components/features/FeedbackDialog.tsx";
+import { FeedbackNudge } from "./components/features/FeedbackNudge.tsx";
 import { useAnalyzedResume } from "./hooks/useAnalyzedResume.ts";
 import { useResumeLibrary } from "./hooks/useResumeLibrary.ts";
 import { useReplaceResumeOnDrop } from "./hooks/useReplaceResumeOnDrop.ts";
@@ -633,6 +634,16 @@ export default function App() {
       )}
 
       <ErrorBoundary onReset={reset}>
+        {/* #912 — the export milestone's ask. Above the result rather than
+            buried under it: it is raised by an action the user just took, and
+            an invitation they have to scroll to find is one they never see.
+            Inline and non-modal on purpose — see `FeedbackNudge`. */}
+        {feedback.nudgeVisible && state.phase === "done" && (
+          <FeedbackNudge
+            onRate={feedback.openDialog}
+            onDismiss={feedback.dismissNudge}
+          />
+        )}
         {state.phase === "done" && edited && displayResult && recovery && (
           <>
             <Result
@@ -733,10 +744,10 @@ export default function App() {
         // artifact matches what the page shows.
         <ExportDialog
           open={exportOpen}
-          // #900 — closing is also when a pending feedback milestone flushes.
+          // #900/#912 — closing is when a pending feedback milestone flushes.
           // The export dialog stays open after a download on purpose (#421),
-          // so opening the interstitial any earlier would stack a second
-          // native modal over the findings the download just produced.
+          // and the nudge raised here would sit behind it, unseen, if it were
+          // shown any earlier.
           onClose={() => {
             setExportOpen(false);
             feedback.notifyExportClosed();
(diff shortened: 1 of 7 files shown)


Samhit21:
Closing for now — reopening the scope discussion on #912 first, per that issue's own instruction to agree the plan before any code. The branch is kept, so this is reopenable as-is if the plan lands here.

PR #924, opened 2026-09-07
fix(rewrite): keep the minus sign on a line-initial negative number (#821)

### LINKED ISSUE #821: cleanRewriteLine eats the minus sign off a line-initial negative number (`-5% churn` → `5% churn`)
`cleanRewriteLine` strips the minus sign off a line-initial negative number, so `-5% churn` reaches the résumé as `5% churn`.

## Repro (on `main`, `960d469`)

```ts
cleanRewriteLine("- -5% churn in Q3.")      // → "5% churn in Q3."
cleanRewriteLine("- -12% cost reduction.")  // → "12% cost reduction."
cleanRewriteLine("• -5% churn in Q3.")      // → "5% churn in Q3."
cleanRewriteLine("-5% churn in Q3.")        // → "5% churn in Q3."
```

A reduction reads as a gain. Nothing downstream looks wrong — the line is grammatical, the number
is plausible, and it ships to the UI and the exported PDF.

## Cause

`LIST_MARKER_PATTERN` (`src/lib/webllm/post-process.ts:94`) runs inside the fixed-point loop #806
introduced:

```ts
const LIST_MARKER_PATTERN = /^(?:\d+[.)]\s+|[•\-]\s*|\*\s+)/;
```

The `[•\-]\s*` branch accepts **zero** trailing whitespace. Pass 1 removes the `- ` marker; pass 2
reads the now-uncovered `-` as a bullet glyph and removes it too. The bare `-5%` case needs only one
pass.

This is the same failure mode as the decimal bug #806 fixed on the numbered branch (`- 3.5x revenue
growth` → `5x revenue growth`, where `\d+[.)]\s*` ate `3.` on the second pass). #806 closed it there
by requiring `\s+`, and the docblock above the loop states the rule — *"Any strip added to this loop
must be checked the same way"* — but the `-`/`•` branch was carried over without that check.

The bare `-5%` shape predates #806 (`main` mangled it before the loop existed); the `- -5%` and
`• -5%` shapes are a regression introduced by the loop.

## Fix

Give `-` the same treatment, with a letter lookahead so `-Shipped X` still normalizes:

```ts
const LIST_MARKER_PATTERN = /^(?:\d+[.)]\s+|•\s*|-(?:\s+|(?=[A-Za-z]))|\*\s+)/;
```

- `- -5% churn.` → `-5% churn.` ✅
- `-5% churn.` → `-5% churn.` ✅
- `-Shipped the migration.` → `Shipped the migration.` ✅ (unchanged)
- `• 4.2% conversion lift.` → `4.2% conversion lift.` ✅ (unchanged)

## Acceptance criteria

- [ ] All four repro lines above keep their minus sign.
- [ ] `-Shipped X` still strips; `• 4.2%` and every existing marker shape in
      `post-process.test.ts` are unchanged.
- [ ] A negative-number case joins the decimal-preservation `it.each` table in
      `post-process.test.ts` — that table is the one asserting this property, and it currently
      covers only the numbered branch.
- [ ] The new cases fail against the pre-fix implementation.

## Context

Found during review of #806 (posted as Secondary 1); #806 merged before it was addressed. Related
to #778 (rewrite drops concrete numbers) — same product surface, different mechanism: #778 is the
model omitting a number, this is post-processing corrupting one the model got right.

Vaishnavi1709:
## Summary

`cleanRewriteLine` stripped the minus off a line-initial negative number, so `-5% churn` reached the résumé — and the exported ATS PDF — as `5% churn`. A reduction read as a gain, and nothing downstream looked wrong: the line stayed grammatical and the figure stayed plausible.

`LIST_MARKER_PATTERN` folded `-` and `•` into one `[•\-]\s*` branch accepting zero trailing whitespace. Inside the fixed-point loop from #806, pass 1 removed the `- ` marker and pass 2 read the uncovered `-` as a bullet glyph and removed it too; the bare `-5%` shape needed only one pass and predates the loop. This is the dash-side twin of the decimal bug #806 fixed on the numbered branch — the loop's own docblock warns that *"any strip added to this loop must be checked the same way"*, and the `-`/`•` branch was carried over without that check.

The fix splits the two glyphs apart and withholds the zero-space branch from `-` when a number follows:

```diff
-const LIST_MARKER_PATTERN = /^(?:\d+[.)]\s+|[•\-]\s*|\*\s+)/;
+const LIST_MARKER_PATTERN = /^(?:\d+[.)]\s+|•\s*|-(?:\s+|(?!\.?\d))|\*\s+)/;
```

`•` keeps `\s*` because it is only ever a glyph, never part of the text it precedes. `-` is the one marker that doubles as ordinary content.

Closes #821

## Deviation from the fix proposed on the issue

Worth flagging explicitly, since the issue body specifies a different regex.

The suggested `-(?:\s+|(?=[A-Za-z]))` **regresses an existing test**: it stops stripping the tight `-**Led** the migration of billing.` case pinned at `post-process.test.ts:194` ("tight dash, no space"), because `*` is not a letter, so the lookahead fails and the marker survives. Caught by simulating the fixed-point loop over every existing marker case before writing code.

A negative lookahead on the *number* is the property actually wanted, and `(?!\.?\d)` additionally covers `-.5%` as well as `-5%`.

## Review focus

- `src/lib/webllm/post-process.ts:127` — the `\s+` branch is deliberately left unguarded, on the reasoning that `- 5%` is a marker plus a positive number and a genuine negative is written `- -5%`. Is there a shape where a model emits `- ` followed by an intended negative with no second sign?
- `src/lib/webllm/post-process.ts:127` — `(?!\.?\d)` is a lookahead, so a tight `-` before a non-number still strips as before. Does that hold for every non-letter opener the rewrite path can emit (`-**Led**`, `-"quoted"`, `-(paren)`)?

## Test plan

- [x] `npm run lint` clean
- [x] `npm run verify` green (also re-run by the `pre-push` hook)
- [x] Full suite green — `383 files / 6457 tests passed, 8 skipped`. Run explicitly rather than relying on `verify`'s scoped `test:changed` selection.
- [x] **New cases falsified against the pre-fix pattern** — reverting only the regex turns all 7 new assertions red, the other 51 in the file stay green.
- [x] All four repro lines from the issue fixed; `-Shipped X`, `-**Led**`, `• 4.2%`, `- 3.5x`, `1. Foo` and `*Foo.*` unchanged.

Six negative-number rows join the decimal-preservation `it.each` table — the table that already asserts this property for the numbered branch — plus a sign-preservation property test across all five marker shapes.


src/lib/webllm/post-process.ts
```diff
@@ -106,11 +106,25 @@ const LEADING_BOLD_WORD_PATTERN = /^\*\*([A-Za-z][\w-]*)\*\*\s+/;
...
+ *     and predates the loop entirely. `(?!\.?\d)` withholds the zero-space
+ *     branch from a sign, covering `-.5%` as well as `-5%`; the `\s+` branch
+ *     is unguarded because `- 5%` is a marker plus a positive number, and a
+ *     genuine negative is written `- -5%`. Nothing needs a tight `-5`-as-marker
+ *     reading. Fix for #821.
  *   - **`*` requires `\s+`**, because `*X*` is italics and is handled by the
  *     paired-emphasis strip instead.
  */
-const LIST_MARKER_PATTERN = /^(?:\d+[.)]\s+|[•\-]\s*|\*\s+)/;
+const LIST_MARKER_PATTERN = /^(?:\d+[.)]\s+|•\s*|-(?:\s+|(?!\.?\d))|\*\s+)/;
```
s-annam, line 127: **Secondary** — `(?!\.?\d)` guards a *digit*, so the zero-space branch is still handed a sign whenever a currency symbol sits between the `-` and the number, and the corruption this PR fixes survives untouched in that shape:

```
- -$5M in infra spend.   →  $5M in infra spend.
-$5M in infra spend.     →  $5M in infra spend.
-€1.2M run-rate.         →  €1.2M run-rate.
```

Measured through the loop against both patterns, not read off the regex. The pre-fix pattern does the same, so this is a remaining gap rather than a regression, and it sits outside #821's ACs — hence Secondary, not Blocking.

It is not hypothetical either, and the downstream consequence is bigger than for the digit case. `preserve-numbers.ts:156` captures `(?<sign>-)?(?<currency>[$€£¥])?` in exactly that order, `classifyAtom` folds `sign` into `prefix` (`:348`), and `isDecorated` counts a lone sign — so `-$5M` and `$5M` are two different atom keys. All three consumers pair `cleanRewriteLine` with `applyNumberPreservation` (`rewrite-section.ts:230`, `rewrite-summary.ts:176`, `eval/run-eval-browser.ts:99`), so an eaten sign registers as drop(`-$5M`) + add(`$5M`) and the #778 gate **reverts the whole unit** — an entire section rewrite discarded on the section path.

Two honest ways out, both fine by me:

1. Close it — `-(?:\s+|(?!\.?\d|[$€£¥]))`, plus a `["- -$5M in infra spend.", "-$5M in infra spend."]` row in the table you just extended. Better still, derive "what can begin a number" from `preserve-numbers.ts` rather than re-listing the currency set in a second place where the two can drift.
2. Scope it out — say so, and have the docblock name the digit limit, so the next reader of this line does not assume a currency-first negative is covered.

Behavioural either way, so I have pushed nothing; your call.

src/lib/webllm/post-process.ts
```diff
@@ -106,11 +106,25 @@ const LEADING_BOLD_WORD_PATTERN = /^\*\*([A-Za-z][\w-]*)\*\*\s+/;
...
+ *     one marker that is also an ordinary character mid-content: a minus sign.
+ *     With a bare `\s*` this branch was the dash-side twin of the numbered bug
+ *     above — pass 1 stripped `- ` from `- -5% churn`, pass 2 read the
+ *     uncovered `-` as a glyph and ate it too, shipping `5% churn`. A
+ *     reduction reads as a gain, and the line stays grammatical, so nothing
+ *     downstream looks wrong. The bare `-5% churn` shape needed only one pass
+ *     and predates the loop entirely. `(?!\.?\d)` withholds the zero-space
+ *     branch from a sign, covering `-.5%` as well as `-5%`; the `\s+` branch
+ *     is unguarded because `- 5%` is a marker plus a positive number, and a
+ *     genuine negative is written `- -5%`. Nothing needs a tight `-5`-as-marker
```
s-annam, line 122: **Secondary** — this sentence claims the trade-off, and the trade-off is the right one, but it understates one consequence. Because the new pattern can only ever strip *less*, the tight-dash case is the single place where this change is worse than `main`, and one shape there re-opens #781:

```
-1. **Led** stuff.      old → "Led stuff."     new → "-1. **Led** stuff."
-1) Shipped Foo.        old → "Shipped Foo."   new → unchanged
-5 services migrated.   old → "5 services…"    new → unchanged
```

The first ships a literal `**` to `perBullet[].text` — the exact failure #781 exists to remove — because the surviving `-` keeps `LEADING_BOLD_WORD_PATTERN`'s `^` anchor from ever seeing the bold. And `-1.` is a closer relative of the `1. - • - **Led** stuff.` nest already pinned at `post-process.test.ts:284` than "nothing needs a tight `-5`-as-marker reading" suggests: the difference is one space.

I would still **keep the fix as written** — a silent sign flip is a false claim on a résumé, a surviving dash is visible in the editable view, and no eval report in the repo attests a tight `-1.` — so this is Secondary, not Blocking. What I would change is the docblock: say that the zero-space branch is withheld from a digit *including* a digit that is itself a marker, and that a literal `**` can survive behind it. Then the next person who meets `-1. **Led**` in an eval report finds it documented instead of surprising.

If you want it pinned rather than just documented, `["-5 services migrated.", "-5 services migrated."]` in the `it.each` table asserts the trade-off you chose. I have deliberately not suggested that row inline, because it encodes your answer to this thread rather than mine.

src/lib/webllm/post-process.ts
```diff
@@ -106,11 +106,25 @@ const LEADING_BOLD_WORD_PATTERN = /^\*\*([A-Za-z][\w-]*)\*\*\s+/;
...
+ *   - **`•` allows `\s*`**, so a tight `•Shipped X` still normalizes. Safe
+ *     with zero-or-more because `•` is only ever a glyph — it never carries
+ *     meaning as part of the text it precedes.
+ *   - **`-` allows `\s*` only when what follows is not a number.** `-` is the
+ *     one marker that is also an ordinary character mid-content: a minus sign.
+ *     With a bare `\s*` this branch was the dash-side twin of the numbered bug
+ *     above — pass 1 stripped `- ` from `- -5% churn`, pass 2 read the
+ *     uncovered `-` as a glyph and ate it too, shipping `5% churn`. A
+ *     reduction reads as a gain, and the line stays grammatical, so nothing
+ *     downstream looks wrong. The bare `-5% churn` shape needed only one pass
```
s-annam, line 118: **Secondary** — "nothing downstream looks wrong" is overstated, and it is the one claim here that a future reader would act on. The same sentence is in the PR body and the commit message, so it is worth correcting in all three.

The sign is part of the #778 atom key: `classifyAtom` builds `prefix = approx + sign + currency` and keys the decorated atom on `prefix + value + suffix` (`preserve-numbers.ts:348`), and `isDecorated` (`:273`) returns true on a lone sign. So whenever the **original** bullet carried the sign, `applyNumberPreservation` sees drop(`-5%`) + add(`5%`), fires, and reverts the unit. On the section path that means the user's whole section rewrite is silently discarded — noisy in a different way, not invisible.

The genuinely silent window is narrower and worth naming precisely: the sign appears only in the *model's* output (the original said "reduced churn by 5%", the rewrite says `-5% churn`), or the text is on a path with no gate — the eval/rubric lane. That is a smaller claim, and it makes the bug look worse rather than better, which is why it is worth making.

Suggestion below rewrites the three lines; the rest of the paragraph is unchanged.

````suggestion
 *     uncovered `-` as a glyph and ate it too, shipping `5% churn`. A
 *     reduction reads as a gain in a grammatical line. Where the ORIGINAL
 *     carried the sign the #778 gate catches it and reverts the whole unit,
 *     so the silent window is a sign the model added. The bare `-5% churn`
 *     shape needed only one pass
````

File: src/lib/webllm/post-process.ts [hunk 4]
```diff
@@ -106,11 +106,25 @@ const LEADING_BOLD_WORD_PATTERN = /^\*\*([A-Za-z][\w-]*)\*\*\s+/;
  *     `- 3.5x revenue growth` became `5x revenue growth`. Silent numeric
  *     corruption on the product path, in a bullet the rewrite prompt actively
  *     asks the model to quantify. Nothing needs a tight `1.Foo`.
- *   - **`-` and `•` allow `\s*`**, so a tight `-Shipped X` still normalizes.
+ *   - **`•` allows `\s*`**, so a tight `•Shipped X` still normalizes. Safe
+ *     with zero-or-more because `•` is only ever a glyph — it never carries
+ *     meaning as part of the text it precedes.
+ *   - **`-` allows `\s*` only when what follows is not a number.** `-` is the
```
s-annam, line 112: **Nit** — `\s*` is not in the pattern: the `-` branch is `\s+` **or** a zero-width lookahead, and the guard binds only to the zero-space alternative, so `- 5%` strips unconditionally. The body of the entry explains this correctly; only this header line is loose. (Compare the `•` entry above, where "allows `\s*`" is literally true.)

````suggestion
 *   - **`-` takes `\s+`, or zero space only when no digit follows.** `-` is the
````

File: src/lib/webllm/post-process.ts [hunk 5]
```diff
@@ -106,11 +106,25 @@ const LEADING_BOLD_WORD_PATTERN = /^\*\*([A-Za-z][\w-]*)\*\*\s+/;
...
+ *     one marker that is also an ordinary character mid-content: a minus sign.
+ *     With a bare `\s*` this branch was the dash-side twin of the numbered bug
+ *     above — pass 1 stripped `- ` from `- -5% churn`, pass 2 read the
+ *     uncovered `-` as a glyph and ate it too, shipping `5% churn`. A
+ *     reduction reads as a gain, and the line stays grammatical, so nothing
+ *     downstream looks wrong. The bare `-5% churn` shape needed only one pass
+ *     and predates the loop entirely. `(?!\.?\d)` withholds the zero-space
+ *     branch from a sign, covering `-.5%` as well as `-5%`; the `\s+` branch
```
s-annam, line 120: **Nit** — "withholds the zero-space branch from **a sign**" is the one inaccuracy in an otherwise exemplary docblock: it withholds it when a **digit** or `.`digit follows. The distinction is exactly the currency gap in the thread on line 127, so naming it here is what keeps the comment from promising more than the regex delivers.

````suggestion
 *     and predates the loop entirely. `(?!\.?\d)` withholds the zero-space
 *     branch when a digit or `.`digit follows, covering `-.5%` as well as
 *     `-5%` — though not a currency-first `-$5M`; the `\s+` branch
````

File: src/lib/webllm/post-process.test.ts [hunk 6]
```diff
@@ -231,10 +231,30 @@ describe("cleanRewriteLine", () => {
...
+      // the sign a `- ` strip had just uncovered. A reduction then reads as a
+      // gain, and unlike a mangled decimal the line stays wholly plausible.
+      ["- -5% churn in Q3.", "-5% churn in Q3."],
+      ["- -12% cost reduction.", "-12% cost reduction."],
+      ["• -5% churn in Q3.", "-5% churn in Q3."],
+      ["- -3.5x infra cost.", "-3.5x infra cost."],
+      // Single-pass shapes — no marker to uncover the sign, so this branch
+      // mangled them on its own, as with the bare decimal above.
+      ["-5% churn in Q3.", "-5% churn in Q3."],
+      ["-.5% weekly churn.", "-.5% weekly churn."],
```
s-annam, line 245: **Nit** — one genuine new win this table misses. `•-5% churn.` is the zero-space `•` twin of the `• -5%` row above, and it is a real behaviour change from `main` (old → `5% churn.`, new → `-5% churn.`), but only the *spaced* form is covered. Worth a row, since `•\s*` keeping zero-or-more is precisely the property the entry for `•` in the docblock now asserts is safe.

````suggestion
      ["-.5% weekly churn.", "-.5% weekly churn."],
      // Zero-space `•` twin of the spaced row above — `•\s*` uncovers the sign
      // on the same pass, so this is a new win rather than a documented limit.
      ["•-5% churn.", "-5% churn."],
````

s-annam (approved):
Correct fix, correctly localized, and the falsification claim reproduces exactly — reverting only the regex turns 7 red and leaves the other 51 green, which is what makes the six new `it.each` rows and the property test load-bearing rather than decorative. The deviation from the regex the issue proposed is right, and right for the reason given: I confirmed `-(?:\s+|(?=[A-Za-z]))` leaves the marker on `-**Led** the migration of billing.` (`post-process.test.ts:194`). Approving — no blockers.

One structural result worth putting in the thread, because it bounds the whole risk surface: **the new pattern is a strict per-pass subset of the old one.** It holds by inspection — `•\s*` and `\d+[.)]\s+` are untouched, `-\s+` and `-\s*` consume the same run, and `-(?!\.?\d)` matches a bare `-`, which is a prefix of whatever `-\s*` matched at that position — and it was fuzz-confirmed over 200k random strings on the alphabet ``-*•1.) \t5a$%−–``: 2,200 divergences, and in **zero** of them is the new match anything but a prefix of the old match. So this change can never strip *more* than before — every possible regression is an under-strip, which is exactly where Secondary 2 below lands, and nothing else can hide.

## Acceptance criteria — #821

| AC | Result |
|---|---|
| All four repro lines keep the minus sign | **met** — pinned at `post-process.test.ts:238–244` |
| `-Shipped X` still strips; `• 4.2%` and every existing marker shape unchanged | **met** — `-Shipped Foo.` pinned at `post-process.test.ts:36`; 58/58 in the file, 6457/6457 across the suite |
| A negative-number case joins the decimal-preservation `it.each` table | **met** — six rows at `post-process.test.ts:238–245`, in that exact table |
| The new cases fail against the pre-fix implementation | **met** — reproduced independently: reverting *only* the regex gives `7 failed \| 51 passed` |

## Review focus — both questions answered

Measured by running both patterns through the fixed-point loop over ~50 shapes, not read off the regex.

**1. A shape where a model emits `- ` and an intended negative with no second sign?** None, and the unguarded `\s+` branch is right. The two ways a negative arrives without a second ASCII hyphen both survive untouched, because neither opener is `-`: `- −5% churn.` (U+2212) → `−5% churn.`, and `- (5%) churn.` → `(5%) churn.`. That leaves only `- 5%`, which is a marker plus a positive by construction and is not separable from "marker plus stripped sign" at the string level — no regex resolves it. Your reasoning holds.

**2. Does a tight `-` before a non-number still strip, for every non-letter opener?** Yes, byte-identical to pre-fix on all three you named, plus the currency and degenerate cases: `-**Led** the migration.` → `Led the migration.`, `-"quoted" thing.` → `"quoted" thing.`, `-(paren) thing.` → `(paren) thing.`, `-$5M saved.` → `$5M saved.`, `-` → `""`, `-.` → `.`. The one opener where the answer is *yes and that is the problem* is the currency case — Secondary 1.

## Secondary

Three, all inline:

1. **`post-process.ts:127`** — `(?!\.?\d)` guards a *digit*, so a currency-first negative still loses its sign, and the #778 gate turns that into a whole-unit revert.
2. **`post-process.ts:122`** — the one strict regression the subset property allows: a tight `-` against a digit-marker now strips nothing, and `-1. **Led** stuff.` ships a literal `**` — the exact failure #781 exists to remove.
3. **`post-process.ts:116–118`** — "nothing downstream looks wrong" is overstated, and the same sentence is in the PR body and the commit message. The sign is part of the #778 atom key, so in the common case the gate *does* catch this. Worth correcting because it makes the bug look worse, not better.

## Nits

Two docblock wording fixes are posted as one-click suggestions (`:112` and `:119–120`) — both are the same imprecision, that the guard keys on a digit rather than on a sign. One test-coverage suggestion on `post-process.test.ts:245`.

Also checked and **cleared**, so nobody re-derives them:

- **ReDoS: none.** Literal-anchored alternatives, a single `\d+`/`\s+`, no nested quantifiers, a zero-width lookahead. 200k-char `9…9` and `-` plus 200k spaces both return sub-millisecond. Termination is untouched — every branch still only removes characters, under the 32-pass cap.
- **The renamed `it.each` title depends on nothing.** `does not eat a decimal` greps to zero hits across `*.ts`/`*.md`/`*.json`/`*.yml`.
- **Consumers are unaffected.** Three call sites, all `.map(cleanRewriteLine)` into `applyNumberPreservation` (`rewrite-section.ts:230`, `rewrite-summary.ts:176`, `eval/run-eval-browser.ts:99`); `LIST_MARKER_PATTERN` is module-private with no external reference.
- **The sibling strips do not reproduce it.** `phrase-tracking.ts:56` and `verb-tracking.ts:41` carry the same `/^[•\-*]\s*/` shape, but they run once rather than to a fixed point and feed `extractStrongPhrase` / `extractLeadingVerb`, which need a leading *alphabetic* token and skip numerics — so a `-5%` opener yields the same result either way, and nothing user-visible derives from it. No change wanted there.
- **Pre-existing, not yours:** unicode dash markers (`–` `—` `−`) are absent from `LIST_MARKER_PATTERN` entirely, so `– -5% churn.` passes through with the marker intact. Notable only because `preserve-numbers.ts:165` `RANGE_DASH` already normalizes that whole block.

**The trade-off the fix accepts, recorded deliberately.** A tight `-` before a number now survives where it used to be stripped — `-2 new services.` → `-2 new services.` (was `2 new services.`), `-2024 to 2025: led.` → `-2024 to 2025: led.`. This is the `-5% churn` ambiguity seen from the other side, and it is the right way round: a mangled figure ships silently and reads as true, a surviving dash is visible in the editable résumé. The docblock says so outright and `["-5% churn in Q3.", …]` pins it, so this is a documented decision, not a defect — no action asked for, beyond Secondary 2 which is a different consequence of the same shape.

## Gates

| Gate | Result |
|---|---|
| `npm run verify` | green, re-run here — `tsc -b`, `eslint`, `check:nul`, `check:fixtures`, `check:baselines`, `check:core`, tests, `vite build` |
| Full suite, `OFFLINECV_FULL_TESTS=1` | `383 files / 6457 tests passed, 8 files skipped` — reproduces the body's claim |
| `fallow audit --base origin/main` | 0 issues in 2 changed files |
| Fixture PII (3a) | `check:fixtures` green; no fixture binaries in the diff, so the gate is n/a to this PR |
| `check:baselines` | 7 `unfiled` ground-truth warnings, all pre-existing on `main` — none charged to this diff |
| One commit per PR | ✅ one commit, `f152fae` |
| No AI attribution | ✅ no trailers |
| Design-system / tokens / reuse; skill-and-script command review | n/a — the diff is two `src/lib/` files |
| CI | `verify`, `fallow`, Cloudflare Pages all green |

## Why nothing was pushed to your branch

Both substantive findings are **behavioural** — widening a lookahead or re-deciding the tight-dash case changes what the function does, which is yours to call, not something a reviewer should land unread. The docblock and test items that *are* safe to push are entangled with those two answers (the exact wording depends on whether you close the currency gap), and your branch is a named contributor's, so the one-commit invariant could not be restored by a collapse afterwards. So they are one-click suggestions instead and the branch is at exactly one commit, queue-ready as it stands. Nothing of yours was touched.

Head reviewed: `f152fae90215d522e945fb70c1029881f6216b53`.

---

s-annam:
Follow-up bookkeeping, after the fact — this merged with the Secondary on `post-process.ts:127` unaddressed, so it is now **#930** rather than a lost thread. Nothing is asked of you here; the fix is the sign-vs-digit widening of the guard, which was always a behavioural call rather than something review should have pushed.

The other two Secondaries are closed out as-is: the `-1. **Led**` under-strip (`:122`) is the documented trade-off the docblock already records, and the "nothing downstream looks wrong" wording (`:118`) is corrected in #930's problem statement, which is where a future reader will land.

src/lib/webllm/post-process.ts
s-annam: Filed as #930 — the currency-first shape (`-$5M`, `-€1.2M`) still loses the sign on `main`, re-verified against `origin/main` after the merge. Resolving here; the work is tracked there, not on a merged PR.

PR #682, opened 2026-07-30
fix(edit,export): an experience role's lone end date no longer comes back as a start date (#672)

rohithgollapalli:
Closes #672.
Closes #814.

## The defect, reproduced first

Probed on `main` before touching anything — the role matrix from the issue, through `buildAtsResumeModel` → `renderAtsResumePdf` → `runCascade`:

```
OUT title="Alpha Analyst"  company="Contoso"         start="2022"     end=undefined hasEndKey=false
OUT title="Bravo Engineer" company="Fabrikam"        start=undefined  end=undefined current=undefined
OUT title="Echo Editor"    company="Adventure Works" start="May 2022" end=undefined hasEndKey=false
```

Confirmed, including post-#618: a lone bare year now draws in the flush-right slot, but the slot flip is unaffected by where it is drawn — `parseDateRange` writes `start_date` on every non-empty path.

## Fix: one date anchor, enforced where the value is produced

`edit/experience-dates.ts` is new and owns the whole rule:

| In | Out |
|---|---|
| `{ start: "2019", end: "2022" }` | unchanged |
| `{ start: "2022" }` | unchanged (#358) |
| `{ end: "2022" }` | `{ start: "2022" }` |
| `{ start: "", end: "2022" }` | `{ start: "2022" }` |
| `{ end: "2022", is_current: true }` | `{ start: "2022" }` — an end date says it ended |
| `{ start: "2020", end: "2022", is_current: true }` | `{ start: "2020", end: "2022" }` — same rule, with an anchor to keep |
| `{ is_current: true }` | `{}` — no anchor, no ongoing claim |

The rule runs where an override is **written**, never where the card renders. Four call sites:

1. `applyExperienceHeaderOverrides` — the DATA, when a date override is folded in.
2. `pushAddedEntry` — the same rule for the "+ Add role" path, so an added role and an edited one export identically.
3. `useEditableParse`'s `setExperienceField` — the override MAP, on every date-cell commit.
4. `useEditableParse`'s `setEntryField` — the same, for an added entry, which has no override map behind it. It is what empties the End cell and fills the Start cell on a freshly added role, and it is pinned in `useEditableParse.test.tsx` (1 of those 3 cases goes red without it; another is the education control, which must NOT normalise).

[...]

Issue #672, opened 2026-07-30
[export] an experience role's lone end date comes back as a start date on the Download-PDF round trip


s-annam:
## Summary

**Category: export, experience, corruption.** An experience role that carries only an end date — no start date — exports and re-parses with that value sitting in `start_date`, and the `end_date` key absent entirely. The value survives; the slot does not. A user reaches this state by clearing the start date on a role in the edit card, or by adding a role and filling in only the end date. After Download PDF and a re-parse, "left Contoso in 2022" has silently become "joined Contoso in 2022".

The failure is silent and it moves the score: the date-completeness check in `computeAnonymousAtsScore` filters on `start_date` alone, so a start-less role reads as incomplete before the round trip and complete after it — the corruption improves the number.

A date **range** ("2019 – 2022") round-trips losslessly, and so does a lone **start** date (#358). Only the lone-end-date case is broken. Education is also correct and must stay that way: `parseEducationDates` deliberately treats a lone date as a graduation date (#97), and #668 pins that behaviour.

## What is and isn't already there

Worth stating up front, because the obvious reading of this ("the PDF export throws the direction away, so the parser cannot get it back") is wrong and would send an implementer down the wrong path:

- The **edit UI does have** an independent end-date field — `RoleHeader`'s `EditableField` with `onFieldChange("end_date", …)`, committed separately from the start date.
- The **model does carry** both slots correctly — `AtsEntryFields.startDate` / `.endDate`, populated in `buildAtsResumeModel`. A probe confirmed `fields.endDate === "2022"` on the very role that re-parses as a start date.
- The **exported PDF does carry the direction** — `renderAtsResumePdf` attaches `JSON.stringify(toJsonResume(model))` as `resume.json` inside the file, and `toWork` emits `startDate` / `endDate` as distinct keys. A start-only and an end-only export are not byte-identical files.

The defect is entirely in **the drawn text, plus the fact that nothing reads the attachment back.** `grep -rn "getAttachments" src/` returns zero hits; `pdf-extract.ts` calls only `getTextContent()` and `getAnnotations()`.

## Evidence (synthetic — safe to quote)

`Input:` five roles in one export, built directly as a `CascadeResult` (no fixture PDF), each with one bullet:

```
{ title: "Alpha Analyst",    company: "Contoso",         end_date: "2022" }
{ title: "Bravo Engineer",   company: "Fabrikam",        is_current: true }
[...]

PR #927, opened 2026-09-10
test(edit): close two coverage gaps left by #682 (#817)

### LINKED ISSUE #817: [test] two coverage gaps left by #682: a bare-year substring check and an untested formatExperienceDateRange
## Summary

Two soft spots in the test coverage #682 added for #672. Both are green today; both are worth closing while the context is fresh.

## 1. The corpus edit-leg absence check can false-fail on a bare year

`corpus-edit-roundtrip.test.ts` now runs the date half through the shared helper:

```ts
experienceFieldCheck(
  f3.experience, "start_date", NEW_END_DATE, edits.replaced.start_date,
  roleText, fails.experience,
);
```

`experienceFieldCheck`'s absence half is a substring test:

```ts
if (replaced && absenceText.includes(replaced))
  into.push(`replaced ${key} still present on re-parse`);
```

Here `replaced` is the **fixture's own `start_date`** — frequently a bare `"2019"` — and `absenceText` is `editedRoleText(...)` = `j(hit)`, the JSON of the **entire** edited role, `description` included. Any future fixture whose role-0 prose contains that year ("shipped the platform in 2019", "since 2019") fails the gate for a reason unrelated to the slot under test.

The title/company checks using the same helper are safe by luck: `NEW_TITLE`/`NEW_COMPANY` and the originals they replace are long, distinctive strings. A four-digit year is not. Green over the current 60 fixtures.

**Fix:** compare the date against the parsed `start_date`/`end_date` **fields** rather than the whole-role JSON blob, or narrow `absenceText` to the header fields for date keys.

## 2. `formatExperienceDateRange`'s trim is an untested behaviour change

`src/lib/edit/experience-dates.ts:276-283` is exported and consumed by both `ats-resume-model.ts` (`experienceDateRange`) and `ReconstructedRole.tsx`, and it has **no direct test**. Its docblock calls out a deliberate behaviour change on the **export** path:

> The trim is deliberate and it is a behaviour change on the EXPORT path, not only the edit one: `experienceDateRange` delegates here, and its old body tested truthiness without trimming, so an entry holding `start_date: "  "` used to draw two spaces into the header slot and now draws nothing. That population is exactly the one this module is total over — raw entries that never passed through `applyOverrides` — so **the case is reachable**.

Mutation-checked during review of #682: replacing

```ts
const start = fields.start_date?.trim();
const end = fields.is_current ? "Present" : fields.end_date?.trim();
```

with the untrimmed originals leaves `src/lib/edit/**`, the #672 round-trip repro and the hook repro **all green** (184 passed). A documented, deliberate, reachable change to export output with zero coverage.

**Fix:** a `describe("formatExperienceDateRange")` block covering the whitespace-only slot, the `is_current` → `"Present"` substitution, and the three fallthrough shapes.

## Acceptance criteria

- [ ] The corpus date absence check cannot be satisfied by a year appearing in role prose.
- [ ] `formatExperienceDateRange` has direct tests, and removing the `.trim()` calls turns at least one red.

Found during review of #682. Related: #672.

Vaishnavi1709:
## Summary

Both gaps #817 identifies were green on `main`, and both were one fixture away from mattering. Verified each still held before touching anything.

Closes #817

**Almost test-only.** The single production change is comment-only: a docblock in `experience-dates.ts` that writing these tests proved wrong (see §3). No behaviour changes.

## 1. The corpus edit-leg absence check could be satisfied by role prose

`experienceFieldCheck`'s absence half searched `absenceText` for `replaced`. On the date call that slice is `j(hit)` — the JSON of the **whole** edited role, `description` included — while `replaced` is the fixture's own prior `start_date`, frequently a bare `"2019"`. Any future fixture whose role-0 prose merely mentions that year ("shipped the platform in 2019", "since 2019") would fail the gate for a reason unrelated to the slot under test.

The `title`/`company` keys are safe on that same wide slice **by luck, not by design** — their literals are long and distinctive. A four-digit year is not.

**Fixed inside `experienceFieldCheck`, not at the call site.** The issue offers either; I took the field-aware route because the function's *presence* half is already field-aware, so the absence half should be too, rather than every caller having to remember which slice is safe for which key. A date key now searches `editedRoleDateText` — the edited role's `start_date`/`end_date` slots and nothing else.

Scoped to the date **slots** rather than to all four header fields deliberately: a year can legitimately occur in a company name (`"2019 Ventures"`), and the slots are where a failed collapse would actually leave the value. The no-marker fallback keeps `editedRoleText`'s existing rule — fall back wide rather than silently skip the absence half, because the presence half is already failing in that case.

## 2. `formatExperienceDateRange` had no direct test

It is exported and consumed on **both** the edit path (`ReconstructedRole`) and the **export** path (`ats-resume-model.ts`'s `experienceDateRange`), and its `.trim()` calls are a documented, deliberate, reachable behaviour change on the latter — an entry holding `start_date: "  "` used to draw two spaces into the header slot and now draws nothing. The #682 review found that removing both trims left `src/lib/edit/**` and two repro suites entirely green.

Adds 8 cases: the trim, the whitespace-only slot, the `is_current` → `"Present"` substitution (including that it out-ranks a populated `end_date`), the unanchored-`is_current` shape the docblock says reaches here only from the export path, the three fallthrough shapes, and the empty case.

## 3. Writing the test falsified the module's own docblock

`experience-dates.ts:82` claimed the two display-only formatters draw `"Present"` where `formatExperienceDateRange` draws `""`. Reproduced before changing anything — **all three draw `"Present"`**; on any `is_current` row `end` is `"Present"` before either fallthrough is reached, so `""` is unreachable for this function. The `""` claim was only ever true of the composite `normalizeExperienceDates` → format path that `ReconstructedRole` runs.

The real divergence is one row over and is a **trim**, not a collapse:

| input | `formatExperienceDateRange` | `buildProjectDates` | `buildDateRange` |
|---|---|---|---|
| `{is_current: true}` | `"Present"` | `"Present"` | `"Present"` |
| `{start_date: "  ", is_current: true}` | `"Present"` | `"  –Present"` | `"  –Present"` |

That second row is the intersection this PR originally left uncovered — the trim and the substitution were each pinned, never together — and it is the row the docblock itself names as the #672 hazard. Both are now corrected and tested, and the stale sibling line references (`group-bullets.ts:328` → `:366`, `entry-dates.ts:16` → `:20`) went with it.

## Falsified against the pre-fix code

This is the half of the AC that actually matters, so both groups were mutation-checked rather than argued:

| mutation | result |
|---|---|
| remove the two `.trim()` calls | **3 of 9** new cases red |
| restore the wide-slice absence half | **exactly 1 of 3** new cases red — the prose one — with both controls still green |

The second row is the useful signal: the new test discriminates rather than passing trivially, and the "still fires when the year survives in the date slot" and "keeps searching the wide slice for a non-date key" controls confirm the narrowing didn't just disable the check.

## Why the #817 regression test lives at module scope, not in the corpus loop

The defect it pins is one **no current fixture exhibits** — it needs a role whose prose repeats its own start year. A corpus gate cannot self-test that class: it is green over the present fixtures and would go red on the next one added. So the three cases run over synthetic roles alongside the corpus suite. The docblock says so.

## Review focus

- `src/lib/heuristics/corpus-edit-roundtrip.test.ts` (`editedRoleDateText`) — narrowing to the date slots means a replaced date that leaked into `title` or `company` is no longer caught. I judged that a worthwhile trade against the false-fail, since a year in a company name is realistic; do you agree, or would you rather it searched all four header fields?
- Same file — the no-marker fallback returns every role's date slots rather than the whole-experience JSON. Is that the right wide form, given the presence half is already failing whenever it's reached?

## Test plan

- [x] `npm run lint` clean
- [x] `npm run verify` green with `OFFLINECV_FULL_TESTS=1` (also re-run by the `pre-push` hook)
- [x] Full suite **6462 passed | 8 skipped** — up from `main`'s 6450 by exactly the 12 cases added
- [x] Both new groups fail against the pre-fix implementation (table above)
- [x] `fallow` clean on both changed files
- [x] No snapshot or baseline file dirty


src/lib/edit/experience-dates.test.ts
```diff
@@ -211,3 +212,77 @@ describe("relocatedEndAnchor", () => {
...
+    expect(
+      formatExperienceDateRange({
+        start_date: "2019",
+        end_date: "2022",
+        is_current: true,
+      }),
+    ).toBe("2019 \u2013 Present");
+  });
+
+  it("draws a lone \"Present\" when is_current has no start anchor", () => {
```
Samhit21, line 286: [Suggestion]: This case is right, and it falsifies the module docblock — please fix `experience-dates.ts:82` in this PR.

Line 82 reads: "`buildDateRange` … and `buildProjectDates` … both draw a bare "Present" where {@link formatExperienceDateRange} draws ""." I ran all three on the raw row:

| input | `formatExperienceDateRange` | `buildProjectDates` | `buildDateRange` |
|---|---|---|---|
| `{is_current: true}` | `"Present"` | `"Present"` | `"Present"` |

They agree. On the unanchored-`is_current` row the only divergence is the separator, not the collapse — `end` is `"Present"` before either fallthrough is reached, so `""` is unreachable on this shape. (The `""` claim is true of the *composite* `normalizeExperienceDates` → format that `ReconstructedRole` runs, but line 82 names the function.)

This PR is the right place to correct it: until now nothing pinned the function, so a stale prose claim cost nothing. After this test, the module docblock and the test next to it assert opposite things about the same call, and a future reader has no way to tell which is load-bearing. Suggest narrowing line 82 to what is actually true — the separator — and moving the `""` claim onto the normalised path where it belongs.

src/lib/edit/experience-dates.test.ts
```diff
@@ -211,3 +212,77 @@ describe("relocatedEndAnchor", () => {
...
+  });
+
+  it("trims each anchor before joining", () => {
+    // Red without the `.trim()` calls: yields " 2019  \u2013  2022 ".
+    expect(
+      formatExperienceDateRange({ start_date: " 2019 ", end_date: " 2022 " }),
+    ).toBe("2019 \u2013 2022");
+  });
+
+  it("treats a whitespace-only anchor as absent, not as a value", () => {
```
Samhit21, line 243: [Suggestion]: Add `{ start_date: "  ", is_current: true }` — it is where the real divergence from the sibling formatters lives, and it is the one combination the two halves of this PR leave uncovered.

The suite pins the whitespace-only anchor and the `is_current` substitution separately but never together, and together is the interesting row:

| input | `formatExperienceDateRange` | `buildProjectDates` | `buildDateRange` |
|---|---|---|---|
| `{start_date: "  ", is_current: true}` | `"Present"` | `"  –Present"` | `"  –Present"` |

Verified by running the three functions. So the formatters *do* disagree on an `is_current` row — just this one, and it is a **trim** disagreement, which is exactly the behaviour this describe block exists to pin. It is also reachable on precisely the population the module is total over: a raw entry that never passed `applyOverrides`.

And it is the hazard the module docblock names — the trim converts a whitespace start into the unanchored bare `"Present"` that "draws into the header and re-parses to nothing at all". Worth one line, since the mutation table shows removing the trims leaves that transition unguarded:

```ts
// The trim turns a whitespace start into the UNANCHORED `is_current` shape —
// the #672 hazard, reached from a raw export-path entry.
expect(formatExperienceDateRange({ start_date: "  ", is_current: true })).toBe("Present");
```

src/lib/heuristics/corpus-edit-roundtrip.test.ts
```diff
@@ -396,6 +396,38 @@ function stringField(entry: unknown, key: EditedRoleField): string | undefined {
  * it as a clean round-trip: injecting that swap fails 1 of 57 fixtures. The
  * field-aware form fails every fixture that has a role.
  */
+const DATE_KEYS: ReadonlySet<EditedRoleField> = new Set(["start_date", "end_date"]);
```
Samhit21, line 395: [Nit]: This insertion detaches `experienceFieldCheck`'s docblock — it now documents `DATE_KEYS`.

The block above ("The experience half of the gate, FIELD-AWARE — … the `title`↔`company` swap … fails 1 of 57 fixtures") sat directly on `experienceFieldCheck`; `DATE_KEYS` and `editedRoleDateText` now sit between them, so TSDoc and editor hover attribute it to the `ReadonlySet`, and the `{@link experienceFieldCheck}` reference at line 350 resolves to an undocumented function. That reference is load-bearing here — it is how `presenceCheck`'s doc explains why the experience path does not use it.

Moving both new declarations above the docblock (or below `experienceFieldCheck`) restores it at no cost.

Samhit21 (commented):
## PR Review: test(edit): close two coverage gaps left by #682 (#817)

### Summary

Both gaps are real and both fixes are verified the right way round — the mutation table is the part that makes this reviewable, and the "exactly 1 of 3 red, both controls green" row is the one that matters, because it shows the narrowing discriminates rather than just disabling the check. The two controls are well chosen: "still fires when the year survives in the date slot" and "keeps searching the wide slice for a non-date key" are precisely the two ways a narrowing like this goes wrong silently. No prior reviews on this PR, so everything below is first-pass.

The substantive finding is in the other direction from the PR's own framing: giving `formatExperienceDateRange` its first direct test has surfaced that the module docblock was already wrong about it. `experience-dates.ts:82` says the function draws `""` on the unanchored-`is_current` row where the two display formatters draw `"Present"`. I ran all three — they all draw `"Present"`. The actual divergence is one row over, on a whitespace-only start beside `is_current`: this function draws `"Present"`, both siblings draw `"  –Present"`. That is a trim disagreement, it is reachable from the raw export path, and it is the intersection of the two behaviours this PR pins — currently covered by neither.

### Findings

1. `experience-dates.test.ts:273` — the `"Present"` assertion is correct and contradicts `experience-dates.ts:82`; correct the docblock here, while the test that falsifies it is being added. [Suggestion]
2. `experience-dates.test.ts:243` — add `{start_date: "  ", is_current: true}`. Verified `"Present"` here vs `"  –Present"` in both `buildProjectDates` and `buildDateRange`; it is the only row where the formatters genuinely disagree on `is_current`, and the trim→unanchored-`Present` transition is otherwise unguarded. [Suggestion]
3. `corpus-edit-roundtrip.test.ts:399` — the insertion point detaches `experienceFieldCheck`'s docblock onto `DATE_KEYS`, orphaning the `{@link experienceFieldCheck}` at line 350. [Nit]
4. `corpus-edit-roundtrip.test.ts:421` — `editedRoleDateText` re-derives the marker lookup `editedRoleText` already does, `j()`-ing every entry a second time per date check. Threading `hit` in, or a shared `markedRole(experience)`, would also keep the two functions' fallbacks from drifting apart independently. [Nit]

### Answering the two review-focus questions

**Date slots rather than all four header fields — yes, and the reasoning is stronger than the one given.** The argument in the PR body is the false-fail trade ("2019 Ventures"). But the narrowing forfeits nothing even in principle: `replaced.start_date` can only reach `title`/`company` if `applyExperienceHeaderOverrides` writes a date into a header field, which is a different defect class with its own dedicated gate, and the one way the collapse *can* misplace the value — leaving it in `end_date` — is inside the narrowed slice. It is also already covered from the other side by the `end_date survived the lone-end-date collapse` check at line 516. So this is not a trade; the wide slice was only ever buying false failures.

**The fallback is fine, for a reason worth writing down.** Every role's date slots is wider in role count but narrower in field count than `editedRoleText`'s whole-experience JSON, so the two fallbacks are not the same "wide" despite the docblock citing the same justification. It does not matter here — the branch is only reached when the presence half has already failed, so the worst case is a second failure line, never a false green. But since they can now drift independently (finding 4), the docblock's "for the reason {@link editedRoleText} does" is worth a clause saying the forms deliberately differ and why it is safe that they do.

**On fixing inside `experienceFieldCheck` rather than at the call site** — agreed, and for the reason stated: the presence half is already field-aware, so a caller-side fix would put the field knowledge in two places. One cost to make visible: `absenceText` is now silently ignored for date keys, including at the corpus date call (line 506) and in two of the three new tests, which pass `j(experience[0])` into a parameter that does nothing. The existing comment covers the why; a word at the call site that the argument is inert for this key would stop the next reader from trusting it.

src/lib/edit/experience-dates.test.ts
Vaishnavi1709: Fixed in `14ffb56`. You're right, and I reproduced it before changing anything — all three draw `"Present"` on `{is_current: true}`, so the `""` claim is false for this function on any `is_current` row.

`experience-dates.ts:82` now states the divergence that actually exists — a **trim**, not a collapse, on the single `{start_date: "  ", is_current: true}` row — with both rows written out inline, and keeps a short note that the `""` claim was only ever true of the composite `normalizeExperienceDates` → format path, so the next reader doesn't re-derive it. The stale sibling references went with it: `group-bullets.ts:328` → `:366`, `entry-dates.ts:16` → `:20`.

This does mean the PR is no longer strictly test-only — the one production change is comment-only. PR body updated to say so.

src/lib/edit/experience-dates.test.ts
Vaishnavi1709: Added in `14ffb56`, and it earns its place by the measure this PR uses for everything else: **dropping both trims now turns 3 cases red, where it turned 2 before.** So the trim→unanchored-`Present` transition really was unguarded, exactly as you said.

I used your framing in the comment — the intersection the block left uncovered, the sibling disagreement (`"  –Present"` vs `"Present"`), and that it is reachable from a raw export-path entry, which is the population this module is total over.

src/lib/heuristics/corpus-edit-roundtrip.test.ts
Vaishnavi1709: Fixed in `14ffb56` — `DATE_KEYS` and `editedRoleDateText` now sit above the docblock, so it is adjacent to `experienceFieldCheck` again and the `{@link}` at line 350 resolves.

Two things that came out of following this thread:

**I repeated the mistake while fixing your finding 4.** Extracting the shared `markedRole` put it between `editedRoleText` and *its* docblock — and the inherited half ("Returns the whole-experience JSON when no entry carries the marker") was then plain wrong, since `markedRole` returns `undefined`. Split into two docblocks, one true of each.

**`stringField` had the same defect already on `main`** — its docblock sat above `type EditedRoleField`. One line to move while I was here, so I did; shout if you'd rather that were its own PR.

Finding 4 itself is in `14ffb56` too: both slicers now derive from `markedRole`, so the marker lookup happens once and the two fallbacks can't drift independently. I also wrote down the thing your fallback answer identified — that `editedRoleDateText`'s fallback is deliberately *not* the same "wide" as `editedRoleText`'s (wider in role count, narrower in field count), and why that's safe.

And your point about `absenceText` being inert for date keys is noted at the corpus call site and above the two unit cases that pass it.

Vaishnavi1709:
All four findings applied in `14ffb56`, plus the two points from the discussion section. Replies are on each thread; this covers **finding 4**, which was body-only and had no thread.

**Finding 4 — the duplicated marker lookup.** Extracted `markedRole(experience)`; `editedRoleText` and `editedRoleDateText` both derive from it, so the lookup happens once and the two fallbacks can no longer drift apart independently. Fixing it also reproduced finding 3 one function over — the extraction detached `editedRoleText`'s docblock onto `markedRole`, where its second half was outright false. Both now have their own.

**On the fallback.** Written into the docblock, in your terms: `editedRoleDateText`'s fallback is deliberately not the same "wide" as `editedRoleText`'s — wider in role count, narrower in field count — and it is safe because the branch is only reached once the presence half has already failed, so the worst case is a second failure line, never a false green.

**On `absenceText` being inert for date keys.** Noted at the corpus call site and above the two unit cases that pass it — in those, the argument is passed *precisely* to demonstrate that a wide slice containing the replaced value no longer fires the check.

**On the narrowing.** I took your reasoning over mine and put it in the docblock and the commit message: it is not a trade. A replaced `start_date` can only reach `title`/`company` via a different defect class with its own gate, and the one way the collapse can misplace the value is inside the narrowed slice and already asserted at the `end_date survived the lone-end-date collapse` line. The wide slice was only ever buying false failures. The PR body's framing was weaker and is corrected.

**One scope change you should know about.** The PR is no longer strictly test-only: the docblock correction touches `experience-dates.ts`, comment-only. PR body updated.

Gates: typecheck and lint clean, full suite **6462 passed | 8 skipped**. Mutation table moved in the right direction — dropping both trims now turns **3** cases red rather than 2, which is the evidence that your suggested row was genuinely unguarded. Restoring the wide-slice absence half still turns exactly 1 of 3 red with both controls green.

Branch collapsed back to a single commit.

src/lib/edit/experience-dates.ts
```diff
@@ -75,16 +75,31 @@
...
- * {@link formatExperienceDateRange} draws "". They also use a tight "–" against
- * this module's spaced " – ". Nothing is broken by that today: the shape is
- * unreachable from the parser and, since #672, from the edit lane too, and
- * neither formatter's output is ever re-parsed. Folding them in needs a separator
- * parameter and a decision on that row, which is a change to display strings, not
- * to #672's corruption — deliberately out of scope here. (`buildEducationDates`
+ * formatters still hand-roll their own collapse and disagree with this one:
+ * `buildDateRange` (`score/group-bullets.ts:366`, which feeds
+ * `formatExperienceHeader`) and `buildProjectDates` (`score/entry-dates.ts:20`).
+ * The disagreement is a TRIM, not a collapse, and it shows up on exactly one
```
Samhit21, line 81: [Nit]: "exactly one row" undercounts it — the trim diverges on every whitespace-only anchor, not just the `is_current` one.

Ran the six rows against all three functions, normalising the siblings' tight dash so only the collapse is compared:

| input | this module | both siblings |
|---|---|---|
| `{is_current: true}` | `"Present"` | `"Present"` — agree |
| `{start_date: "  ", is_current: true}` | `"Present"` | `"  –Present"` |
| `{start_date: "  ", end_date: "2022"}` | `"2022"` | `"  –2022"` |
| `{start_date: "2019", end_date: "  "}` | `"2019"` | `"2019–  "` |
| `{start_date: "  "}` | `""` | `"  "` |
| `{end_date: "  "}` | `""` | `"  "` |

Five rows, not one. Which follows from the paragraph's own diagnosis: once the disagreement is a TRIM rather than a collapse, it has to show up wherever an anchor is whitespace-only, and `is_current` is irrelevant to it. The `is_current` row is the one that now has a test, which is not the same claim.

Nothing downstream changes — the safety argument ("reachable only from a raw export-path entry, never re-parsed") holds identically for all five. Suggest "on any row carrying a whitespace-only anchor" with the `is_current` row kept as the worked example, since that is the one #817 pinned.

src/lib/heuristics/corpus-edit-roundtrip.test.ts
```diff
@@ -365,26 +365,68 @@ function presenceCheck(
...
- *  synthetic marker. `NEW_TITLE`/`NEW_COMPANY` are literals that appear nowhere
- *  in any fixture, so at most one entry matches. Returns the whole-experience
- *  JSON when no entry carries the marker — the presence half is already failing
- *  in that case and must not be masked by a silently-skipped absence half. */
-function editedRoleText(experience: readonly unknown[], fallback: string): string {
-  const hit = experience.find(
+ *  synthetic marker, or `undefined` when none does. `NEW_TITLE`/`NEW_COMPANY`
+ *  are literals that appear nowhere in any fixture, so at most one entry
+ *  matches. Sole owner of the marker lookup — both slicers below derive from it
+ *  so their fallbacks cannot drift apart independently (#817 review). */
```
Samhit21, line 371: [Nit]: This says the fallbacks "cannot drift apart independently"; `editedRoleDateText`'s docblock at line 418 says "the two can drift independently". My wording in finding 4 caused that — it was imprecise and you applied it faithfully.

What the extraction actually fixes is the **lookup**: one definition of "the marked role", so the two slicers cannot disagree about which entry they are slicing. The **fallbacks** are still written independently — `fallback` threaded in by the caller here, `j(experience.map(slots))` there — and line 418 is right that they deliberately differ and that the difference is safe. Both things are true; it is only the sentence here that claims more than the change delivers.

Suggest scoping this one to the lookup and letting line 418 keep ownership of the fallback story:

```
 *  matches. Sole owner of the marker lookup — both slicers below derive from it,
 *  so they cannot disagree about WHICH entry they are slicing. Their no-marker
 *  fallbacks are separate and deliberately differ; see {@link editedRoleDateText}.
```

Samhit21 (approved):
## PR Review: test(edit): close two coverage gaps left by #682 (#817)

### Summary

All four findings applied, and I verified each rather than taking the replies at their word.

1. **The docblock correction is right.** The replacement paragraph's table matches what the three functions actually return, and both stale sibling line references are now correct (`group-bullets.ts:366`, `entry-dates.ts:20`). Keeping a note that the `""` claim was only ever true of the composite path is the right call — it stops the next reader re-deriving the thing that made the original wrong.
2. **The new case earns its place by this PR's own measure.** Dropping both trims now turns 3 cases red where it turned 2, which is the evidence the row was unguarded; the suite count moving 6461 → 6462 matches one added case exactly.
3. **The docblock is adjacent to `experienceFieldCheck` again** and the `{@link}` at line 350 resolves. Catching that you had repeated the same detachment onto `markedRole` while fixing finding 4 — and that the inherited half was then outright false, since `markedRole` returns `undefined` rather than the whole-experience JSON — is the better catch of the two. Splitting into one docblock true of each is right.
4. **`markedRole` does what was asked**, and picking up `stringField`'s pre-existing detachment while in the file is fine by me — comment-only, same defect, same function cluster. Not worth its own PR.

The discussion points all landed too: the fallback asymmetry is written down, the narrowing reasoning is in the docblock and the commit message, and the inert `absenceText` is flagged at both the corpus call site and the two unit cases. On the scope change — a comment-only edit to `experience-dates.ts` is the correct place to fix a wrong comment in `experience-dates.ts`, and calling it out in the PR body is all that was needed.

CI is green on `14ffb56`: `verify`, `fallow`, and Cloudflare Pages all pass.

### Findings

Two comment-only nits, neither blocking, and one of them is mine:

1. `experience-dates.ts:81` — "exactly one row" undercounts the divergence; a trim disagreement necessarily shows up on every whitespace-only anchor, and I measured five such rows. [Nit]
2. `corpus-edit-roundtrip.test.ts:371` — `markedRole`'s docblock says the fallbacks "cannot drift apart independently" while line 418 says they "can drift independently". My finding-4 wording was the source of the imprecision: the shared lookup fixes the lookup, not the fallbacks. [Nit]

Both are one-line edits and neither changes behaviour, so I am not holding the PR for them — fold them into whatever you touch next in these files if you prefer.

### Verdict

**Action: APPROVE.** The coverage this adds is real, the mutation evidence is what makes it reviewable, and the docblock it corrects along the way was a live falsehood about a function nothing had pinned.

PR #935, opened 2026-09-14
test(corpus): gate a later bullet override on the edit round-trip (#487)

### LINKED ISSUE #487: Download PDF drops bullet overrides on the edit round-trip

Samhit21:
Resolves #487.

## Where #487 stands

The defect the issue reported is already fixed on `main`. #665 found the root cause: the edit-leg gate graded the **base** sections while feeding the exporter **edited** descriptions, so every overridden bullet fell out of its entry. That change retired all 39 Group A `bullets` entries, but #487 was not in its list of linked issues (that list named the bullet-identity issue, 648), so the issue stayed open.

The issue's correction comment makes a second claim that nothing ever checked: a **non-first** override failed too, on `skia-proxy-classic`, `awesome-cv-resume` and `weasyprint-cairo-classic`. The gate only ever edits `observations[0]`, so it couldn't say whether that still held. It does hold now, but nothing showed it. This PR closes that gap.

## What changed

One file: `src/lib/heuristics/corpus-edit-roundtrip.test.ts`. There are no product changes and no `KNOWN_FAILURES` changes.

- **A second bullet edit.** The gate now also edits a later bullet (`NEW_LATER_BULLET`, carrying the `Vantreon` sentinel and added to the collision guard) and asserts that it round-trips and that the text it replaced is gone.
- **`laterEditableBullet`** picks the last pooled bullet the reconstructed résumé offers an editor for, other than `observations[0]`.

### Why that selection rule

"Offers an editor for" is the product's own rule, not one chosen to make the gate pass. Every writer of a bullet override reaches a bullet only through an **experience role**:
- **The inline row:** `RoleEntry` receives `onBulletChange`.
- **The whole-résumé rewrite's apply:** `rewriteApplyBySection` wires only `experience:<n>` sections and the summary.
- **Projects and achievements:** `ProjectsSection` and `AchievementsSection` render their rows as `<ResumeBulletRow key bullet />` with no `onBulletChange`, and their only remove is the whole-entry one.

So an override on a project or achievement bullet isn't something a user can create. Grouping uses the same `groupBulletsByExperience` over `experience` that the surface does.

This is measured, not assumed. Pointing the edit at the literal last observation instead fails 11 fixtures, and the diagnosis is below.

## Verification

- **Gate:** 66/66 on this branch, and `RL_BAKE=1` lists only the existing accepted #224 entry (`skia-proxy-achievements-oneline :: bullets`).
- **Teeth, checked with an injected regression.** I stopped `applyBulletTextOverrides` after its first override:

  | gate | fixtures failing `bullets` |
  |---|---|
  | `main`'s (leading bullet only) | 1, the pre-existing accepted #224 entry, so nothing new |
  | this branch | **57** |

- **Typecheck and lint:** `tsc -b --noEmit` exits 0, and `eslint` on the changed file is clean.
- **Pre-push:** `npm run verify` passed as the pre-push hook: typecheck, lint, gates, tests and build. Fallow reports dead code 0 and complexity 0. Its one duplication group (`corpus-edit-roundtrip.test.ts:726-738` against `corpus-roundtrip.test.ts:124-138`) is the pre-existing per-fixture loop, moved down by this change's insertions; the audit gate excluded it as inherited.

## What the literal-last-bullet probe turned up

This is recorded so nobody re-derives it. Editing `observations[observations.length - 1]` fails 11 fixtures, for three unrelated reasons:

| cause | fixtures | product-reachable? |
|---|---|---|
| The bullet belongs to a **project** entry. `replaceBulletInDescriptions` walks only `experience`, so the edit never reaches `projects[].description` and the exporter re-attributes the old text. | 7 | No: project rows have no editor |
| The bullet is an achievement's **title-only header line**, the #224 shape | 2 | No: `suppressTitleOwnedBullets` hides the row |
| The parser glues **sidebar text onto a role's last description line**, so the clean pool bullet matches no line, groups under "Other bullets", and is absent from the export **even unedited** | 2 (`chromium-asymmetric-sidebar`, `two-column-achievements-sidebar`) | Yes, but it's a base parse/export defect, not an edit-leg one |

The first row becomes a real bug the day project bullets gain an inline editor: `withMatchedDescriptionLine` would have to walk projects, achievements and certifications in `groupBulletsByExperience`'s order. The third row is out of scope here and is not yet filed.


s-annam (approved):
Approving: 0 Blocking, 0 Secondary, 0 Nits. The gate now proves a non-first bullet override survives the round-trip, and I checked its teeth myself instead of taking them from the description.

## Verification

- **#487 acceptance criteria:** met.
  - The drop the issue reports was fixed on `main` by #665, and the Group A baseline is already gone.
  - The correction comment's open question — does a non-first override still fail? — is what this PR answers. It edits a second, later bullet on every fixture that has one.
  - `Resolves #487` is accurate.
- **Injected regression.** I stopped `applyBulletTextOverrides` after its first override and re-ran the gate on `ba461ff`.
  - 56 fixtures fail on `later bullet … missing`; none fail on the leading bullet.
  - The description's **57** adds the one fixture whose `bullets` failure is the accepted #224 entry, so the two counts agree.
  - The file was restored afterwards.
- **Selection rule.** I checked this against the UI code, not the docblock:
  - `ProjectsSection` and `AchievementsSection` pass no `onBulletChange`.
  - `rewriteApplyBySection` wires only the summary and `experience:<n>`.
  - In `buildEntryGroups`, experiences come first in `combined` and the grouper's first match wins. So grouping over `experience` alone gives the same role groups the surface shows.
- **Gates:**
  - Corpus gate: 66/66.
  - `tsc -b --noEmit` and `eslint` on the file: clean.
  - fallow: report-only. Its one duplication group, the per-fixture loop, is inherited, and `computeEditFailures` gained one branch.
  - CI: `verify`, `fallow` and Cloudflare Pages all green.
  - Fixture PII and design-system/style gates: skipped, since the PR changes only one test file.
- **Description (3f):** accurate. #665 did close #648 and not #487; `RL_BAKE`, `withMatchedDescriptionLine`, and the summary wiring all exist as described.

## Filed

- **#936** — the third row of your literal-last-bullet probe: sidebar text glued onto a role's last description line. I re-ran it on the product code, which this PR does not change. On `chromium-asymmetric-sidebar` and `two-column-achievements-sidebar` the last pool bullet groups under "Other bullets" and is missing from the **unedited** export's re-parse. It was unfiled and existed only in this description, so it now has an issue.

## Note (no action)

`later bullet` failures go into the same `fails.bullets` list as the leading bullet. On the one fixture with an accepted `bullets` baseline (`skia-proxy-achievements-oneline`, #224), a later-bullet regression would therefore be hidden by that entry. That follows from the baseline being per category and covers a single fixture, so I'm not asking for a change.

---

Issue #928, opened 2026-09-11
[refactor] preserve-numbers: delete the two dead year tiers and close the doc gaps #919 left


s-annam:
Follow-up to #876 / #919 (merged as `b4d3a2d`). Four items, none behavioural, all
inside `src/lib/webllm/preserve-numbers.ts` and its test. They were the surviving
non-blocking nits on that PR's approval plus one measured simplification found
afterwards.

**Nothing here changes what the gate does.** This is a drive-by: fold it into
whatever PR is next in `src/lib/webllm/`, rather than opening a PR of its own.

## 1. Tiers 1b and 1c in `isYearContext` are dead — delete them and their two regexes

`isYearContext` (`preserve-numbers.ts:438`) ends with two "verb" tiers:

```ts
// 1b. Bare past-tense verb followed by a function word, punctuation, or end of bullet (#876)
if (YEAR_PREFIX_VERB_CUE.test(before) && FUNCTION_WORD_FOLLOWS.test(after)) {
  return "verb";
}

// 1c. Weak prepositions, qualifiers, or bullet-initial attributive years (#876)
if (YEAR_PREFIX_WEAK_CUE.test(before) || leadingText.length === 0) {
  return "verb";
}
```

Both are shadowed by the inverted gate's own fallback in `bareIntegerClaim`
(`preserve-numbers.ts:510`), which already returns `year_verb` for every 4-digit
integer in 1900–2099 that nothing else claimed. Anything 1b or 1c catches, the
fallback catches one branch later with the same answer.

The only way they could matter is ordering: `bareIntegerClaim` consults
`isRangeEndpoint` *before* the fallback, so a 4-digit range endpoint carrying a
weak cue gets `year_verb` from 1c where it would otherwise get `range` — and
`range` is claimed while `year_verb` is counted as unclaimed by
`countUnclaimedByKey`. That is a real difference in principle. It does not show
up in practice.

Measured, not assumed. Replacing both tiers with `return "none"`:

- `npx vitest run src/lib/webllm/preserve-numbers.test.ts` → **97/97 still pass.**
- Differential against the submitted implementation over **136 cases** — 16 bullets
  covering 4-digit range endpoints (`"Shipped 2000-3000 units per quarter."`,
  `"Handled 1900 to 2100 requests per second."`, `"Reduced errors from 2000 to 1900
  per week."`), weak-cue shapes (`from` / `by` / `for` / `of` / `over` / `the` /
[...]

PR #937, opened 2026-09-14
refactor(heuristics): derive parseDateRange's open-ended test from the lexicon (#931)

### LINKED ISSUE #931: refactor(heuristics): parseDateRange still hardcodes the open-ended date vocabulary (19th copy, #916 missed it)
## Problem

`#916` (PR #926) consolidated 18 hardcoded date-token literals onto the one lexicon in
`regex.ts`. A 19th copy survives, in a file the PR touched and which now imports the very
constant that replaces it.

`src/lib/heuristics/line-primitives.ts:454`, inside `parseDateRange`:

```ts
if (/^(present|current|now|ongoing)$/i.test(endRaw)) {
```

`OPEN_ENDED_ALT` is already imported in that module (`line-primitives.ts:25`) and used 230 lines
below at `:684`. The literal is 226 lines *above* the `stripDateRange` copy #926 did consolidate.

## Why this copy in particular

`parseDateRange` is the function that decides `is_current`. Every other copy #916 removed fed a
strip or a shape test; this one feeds a parsed résumé field. If the lexicon ever gains a word —
`Till date`, `To date`, `PRESENT-` — every consolidated site picks it up and this one silently
does not, and the divergence shows up as a role that stops being marked current.

## Proposed fix

`OPEN_ENDED_ALT` is `"Present|Current|Now|Ongoing"` — the same four words in the same order as the
literal, and the call is `.test()` only, so the capture group is unused. The replacement is
language-identical:

```ts
if (new RegExp(`^(?:${OPEN_ENDED_ALT})$`, "i").test(endRaw)) {
```

Hoist it to module scope while you are there, matching `SEASON_LEAD_RE` and everything in
`regex.ts` — `parseDateRange` runs per date line. (No `g` flag, so no shared-`lastIndex` hazard.)

Optional, same shape, explicitly out of scope for #916 and worth folding in only if it stays
small: `src/lib/edit/field-validators.ts:42` holds a sixth copy, `PRESENT_WORDS`, whose comment
says it was inlined because `PRESENT_RE` is `\b`-bounded and would not compose. `OPEN_ENDED_ALT`
**is** the bare unbounded alternation that comment wanted, and the file already imports from
`../heuristics/regex.ts`, so the stated rationale is now obsolete. (`score.ts:191` carries a month
copy and is correctly left alone — `score.ts` is deliberately near-zero-dep.)

## Why it is filed rather than fixed in #926

Reported as **Secondary** in the review of #926 (https://github.com/offlinecv/OfflineCV/pull/926#discussion_r3990658697) and not auto-fixed there: it is a regex edit inside
a parse path, which is over the bar for a fix that merges unread, and line 454 is not a `+` line so
the review had no anchor to hang a one-click suggestion on. PR #926 merged without it.

The PR body's claim that "all 18 are gone" is accurate for the sites it audited; the true count
inside those five files was 19.

## Acceptance criteria

- [ ] `line-primitives.ts:454` uses `OPEN_ENDED_ALT`; no open-ended date literal remains in
      `src/lib/heuristics/`.
- [ ] The pattern is built once at module scope.
- [ ] Corpus snapshots byte-identical — `git status` clean after a full suite run.
- [ ] `npm run verify` green, `fallow audit --base origin/main` clean on changed files.

Vaishnavi1709:
## Summary

#916 moved the date-word vocabularies into one shared lexicon in `regex.ts`, but one copy survived: `parseDateRange`'s `/^(present|current|now|ongoing)$/i`. It is the copy that matters most. Every other site only strips text or tests a shape, while this one decides `is_current`. If the lexicon ever gained or lost a word, this site would silently diverge and a role would stop being marked current.

This derives it from `OPEN_ENDED_ALT`, built once at module scope next to its only use. It is the same four words in the same order, and the call is `.test()` only, so the unused capture group goes away with no effect.

Closes #931

## Scope

- **`line-primitives.ts`**: the site the issue names.
- **`field-validators.ts` (`PRESENT_WORDS`)**: folded in, as the issue allows. The string is byte-for-byte `OPEN_ENDED_ALT`. Its comment inlined it because `PRESENT_RE` is `\b`-bounded and would not compose inside `DATE_FIELD_RE`'s anchors, and `OPEN_ENDED_ALT` is exactly the bare alternation that comment wanted. `DATE_FIELD_RE.source` is unchanged.

## One deliberate deviation from the acceptance criteria

The first AC reads "no open-ended date literal remains in `src/lib/heuristics/`". One still does: the `experience.dates` fold in `__test-utils__/ground-truth.ts:219`.

I left it on purpose. That file is the ground-truth oracle, and its docblock is explicit that truth is an outside opinion authored by reading the PDF, "never by running the parser". If the fold read the production lexicon, a word dropped from `OPEN_ENDED_ALT` would disappear from both sides of the comparison and the scoreboard would stay green. Say so if you'd rather it followed the lexicon anyway.

The other two remaining matches in that directory (`entry-blocks.ts:75`, `education.ts:343`) are graduation-context vocabularies that share only two words, as established in #926.

## Behaviour-preserving: measured, not argued

A temporary harness (removed before commit) recorded `parseDateRange` over **3,031 generated lines**, 1,170 of them on the `is_current` branch, and `validateDate` over **41 values**, before and after the change. It found **zero differences**. The inputs covered every case variant of the four words, near-misses (`Presently`, `Nowhere`, `Ongoings`, `Till date`), placeholders, seasons, redacted years and every separator.

The change also does what the issue is for. With `Ongoing` temporarily removed from `OPEN_ENDED_ALT`:

| call | result |
|---|---|
| `parseDateRange("Jan 2020 – Ongoing")` | loses `is_current` |
| `validateDate("Ongoing")` | now warns |
| `parseDateRange("Jan 2020 – Present")` / `validateDate("Present")` | unchanged |

Before this PR, neither site would have moved.

## Out of scope, worth knowing

`src/lib/webllm/preserve-numbers.ts` holds six more lowercase copies of the vocabulary, at lines 303, 388–391 and 395. They are not in #931's scope and #928 does not mention them. That module has **no imports at all**, so pointing it at `regex.ts` would give it its first dependency. That is a design decision, not a drive-by, so I left it alone and am not filing it from here.

## Review focus

- `src/lib/heuristics/__test-utils__/ground-truth.ts:219`: is keeping the oracle independent of the lexicon the right call, or should it follow it despite the AC wording?

## Test plan

- [x] `npm run lint` clean on both files
- [x] Full suite, forced (`OFFLINECV_FULL_TESTS=1`): **6,495 passed, 8 files skipped**
- [x] `fallow audit --base origin/main`: no issues in 2 changed files
- [x] Before/after differential: 0 differences over 3,072 inputs
- [x] Mutation check: both sites follow a change to `OPEN_ENDED_ALT`

No fixture binaries touched.


s-annam (approved):
Approving: 0 Blocking, 0 Secondary, 0 Nits. Both sites now derive from the lexicon, and I checked equivalence and the mutation claim myself rather than taking them from the description.

## Verification

- **#931 acceptance criteria:** met.
  - `parseDateRange` tests `endRaw` against `OPEN_ENDED_ONLY_RE`, built once at module scope from `OPEN_ENDED_ALT`. It has no `g` flag, so `.test` stays stateless.
  - **"No open-ended date literal remains in `src/lib/heuristics/`":** met in the sense the AC was written for. Three matches remain, and I agree with leaving all three:
    - `__test-utils__/ground-truth.ts:219` is the oracle. Deriving its fold from `OPEN_ENDED_ALT` would move truth in lockstep with the parser it grades, so a dropped word would vanish from both sides and stay green. You flagged this on the issue before opening the PR. Keep it.
    - `entry-blocks.ts:75` and `education.ts:343` are graduation-word lists that share two members with the lexicon. #926 ruled them "not a copy", and substituting `OPEN_ENDED_ALT` there would *add* `now|ongoing`, which changes the language.
  - Snapshots and baselines: CI `verify` ran the full suite green on `6f1c1d5`, and the tree was clean after my local runs.
  - `fallow audit --base origin/main`: no issues in the 2 changed files.
- **Equivalence**, measured independently of your harness:
  - `/^(present|current|now|ongoing)$/i` against `^(?:Present|Current|Now|Ongoing)$` with `i`: 0 differences over 200,000 random strings plus edge words (`presently`, `nowhere`, `Current ` with a trailing space, `ongoing.`, `till date`).
  - `DATE_FIELD_RE`'s template expands to the identical source string, since `OPEN_ENDED_ALT` is byte-for-byte the old `PRESENT_WORDS`.
- **Mutation claim:** reproduced. With `Ongoing` removed from `OPEN_ENDED_ALT`:
  - `parseDateRange("Jan 2020 – Ongoing")` loses `is_current`, and `validateDate("Ongoing")` starts warning.
  - `Present` is unchanged on both.
  - `regex.ts` was restored afterwards.
- **Gates:**
  - Ran: `tsc -b --noEmit` clean; `eslint` clean on both files; `line-primitives` and `field-validators` suites 88/88; CI `verify`, `fallow` and Cloudflare Pages green.
  - Skipped, since no fixtures, components or styles changed: fixture PII, design-system and style tokens.
- **Description (3f):** accurate. The six `preserve-numbers.ts` copies (303, 388–391, 395), "no imports at all", and #928 not mentioning them all check out on `main`.

## Filed

- **#938**: the six `preserve-numbers.ts` copies. You were right that this is a design decision rather than a drive-by: the module has no imports, and `score.ts` keeps its own copy for the same reason. It also existed only in this description, so the issue records the decision to make: derive from `OPEN_ENDED_ALT`, or stay standalone with a test pinning the local list to the lexicon so drift fails loudly.

---

PR #939, opened 2026-09-14
fix(rewrite): keep the minus sign on a currency-first negative (#930)

### LINKED ISSUE #930: fix(rewrite): a currency-first negative still loses its minus sign in cleanRewriteLine
## Problem

`#821` (PR #924) stopped `cleanRewriteLine` from eating the minus sign off a line-initial
negative number. The guard it added keys on a **digit**, not on a **sign**, so the corruption
survives untouched whenever a currency symbol sits between the `-` and the number.

`src/lib/webllm/post-process.ts:127`:

```ts
const LIST_MARKER_PATTERN = /^(?:\d+[.)]\s+|•\s*|-(?:\s+|(?!\.?\d))|\*\s+)/;
```

`(?!\.?\d)` fails to fire when the next character is `$`, `€`, `£`, `₹`, so the zero-space
`-` branch matches and the sign is stripped. Through the fixed-point loop:

| input | output on `main` | wanted |
|---|---|---|
| `- -$5M in infra spend.` | `$5M in infra spend.` | `-$5M in infra spend.` |
| `-$5M in infra spend.` | `$5M in infra spend.` | `-$5M in infra spend.` |
| `-€1.2M run-rate.` | `€1.2M run-rate.` | `-€1.2M run-rate.` |

The plain-digit twins (`- -5% churn.`, `-5% churn.`) are fixed and pinned at
`post-process.test.ts:238–245`; only the currency-first shape is still wrong.

## Why it matters

The sign is part of the `#778` atom key — `classifyAtom` builds `prefix = approx + sign + currency`
and keys the decorated atom on `prefix + value + suffix` (`preserve-numbers.ts`). So a stripped
sign usually makes the whole rewritten unit fail the preservation gate and revert, which is the
safe outcome but throws away a good rewrite. Where it does not revert, the résumé ships a figure
whose meaning is inverted, and it reads as true.

Reported as **Secondary** in the review of #924
(https://github.com/offlinecv/OfflineCV/pull/924#discussion_r3990685375) and deliberately
not pushed there, because widening the lookahead changes behaviour and is the author's call.
PR #924 merged without it, so it is filed here rather than lost.

## Proposed fix

Widen the guard from "a digit follows" to "a number-ish token follows", i.e. let an optional
currency symbol sit between the `-` and the digit. Roughly:

```ts
-(?:\s+|(?![$€£¥₹]?\.?\d))
```

Exact character class and whether to reuse an existing currency lexicon (`preserve-numbers.ts`
already carries one — prefer reusing it over a second copy) is the implementer's call.

The subset property established in #924's review still has to hold: the new pattern must be a
strict per-pass subset of the old one, so the change can only ever strip *less*.

## Acceptance criteria

- [ ] The three rows in the table above round-trip with the sign intact.
- [ ] The existing 58 assertions in `post-process.test.ts` stay green, in particular `-Shipped Foo.`
      → `Shipped Foo.`, `-**Led** the migration.` → `Led the migration.`, `-$5M saved.` where the
      `-` is a *marker* — call that ambiguity explicitly in the docblock, the way the digit case is.
- [ ] The new cases fail against the pre-fix implementation (reverting only the regex turns them red).
- [ ] The currency rows join the same `it.each` decimal-preservation table at
      `post-process.test.ts:238–245`.
- [ ] The docblock at `post-process.ts:112–120` names what the guard actually keys on.

Vaishnavi1709:
## Summary

#821 (PR #924) stopped `cleanRewriteLine` eating the minus sign off a line-initial negative number, but its guard keyed on a **digit**, not a **number**. A currency symbol between the `-` and the digits slipped past `(?!\.?\d)`, so the zero-space `-` branch still matched and stripped the sign:

| input | on `main` | with this PR |
|---|---|---|
| `- -$5M in infra spend.` | `$5M in infra spend.` | `-$5M in infra spend.` |
| `-$5M in infra spend.` | `$5M in infra spend.` | `-$5M in infra spend.` |
| `-€1.2M run-rate.` | `€1.2M run-rate.` | `-€1.2M run-rate.` |
| `-£300k budget cut.` | `£300k budget cut.` | `-£300k budget cut.` |
| `-¥1,000 cost.` | `¥1,000 cost.` | `-¥1,000 cost.` |

A loss read as a gain, in a line that stays grammatical and plausible.

Closes #930

## The change

The guard now allows an optional currency symbol before the optional `.` and the digit:

```diff
-const LIST_MARKER_PATTERN = /^(?:\d+[.)]\s+|•\s*|-(?:\s+|(?!\.?\d))|\*\s+)/;
+const LIST_MARKER_PATTERN = new RegExp(
+  String.raw`^(?:\d+[.)]\s+|•\s*|-(?:\s+|(?!${CURRENCY_SYMBOL_CLASS}?\.?\d))|\*\s+)`,
+);
```

**The currency class is reused, not copied, as the issue asked.** `preserve-numbers.ts` already owned it, inline in its `ATOM` regex. It is now exported as `CURRENCY_SYMBOL_CLASS`, and both consumers read it:

- **Why they must agree:** the sign is part of the #778 atom key, so a strip that eats a sign the classifier keys on either reverts a good rewrite or ships an inverted figure.
- **No new import:** `post-process.ts` already imported `preserve-numbers.ts` for `checkNumbersPreserved`.
- **`ATOM` is unchanged byte for byte.** It is now built from the constant, and I checked that its `.source` and `.flags` match `main` exactly.

## Two judgement calls, both documented at the guard and pinned by tests

- **`₹` stays out.** The issue's sketch included it, but neither consumer of the shared class recognises `₹` today, and widening the class widens both at once. `-₹2Cr loss.` still loses its sign. A test named `KNOWN GAP` pins that, so the gap is visible and flips loudly when the class is widened.
- **A tight `-$5M saved.` is read as a sign.** It is genuinely ambiguous between a glued marker and a signed figure. This is the same call the digit branch already makes for `-5%`: keeping a stray `-` is cosmetic, eating a real one inverts the figure. `- $5M saved.` (with a space) still strips as a marker plus a positive figure, which is pinned too.

## Verification

**The subset property from #924's review still holds.** I compared the old and new guards over 95 inputs: every case in `post-process.test.ts`, plus the issue's rows and ambiguity probes (`-$ 5M`, `-$abc`, `---`, a bare `-`).

- The new guard **never strips where the old one would not**.
- Exactly **8 outputs change**, and all of them are currency-first negatives.

**The new tests fail against the pre-fix guard.** Reverting only `LIST_MARKER_PATTERN` turns **9 new cases red**: the 7 table rows, the property test and the tight-marker test. Both controls stay green: the positive `- $5M` strip and the `₹` known-gap pin.

**Tests added**

- 7 currency rows in the `does not eat a number a marker strip uncovers` table the AC names.
- A property test over every symbol in `CURRENCY_SYMBOL_CLASS` × every marker. It is driven by the shared constant, so a symbol added later is covered automatically.
- The positive `- $5M` control, the tight `-$5M` ambiguity and the `₹` known gap.

## Review focus

- `src/lib/webllm/post-process.test.ts`, the `KNOWN GAP` test: is pinning `-₹2Cr` → `₹2Cr` as current behaviour the right way to keep the gap visible, or would you rather it were an `it.todo`, or `₹` added to the class here?
- `src/lib/webllm/post-process.ts`, the tight `-$5M` decision: does reading it as a sign match how the rewrite model actually emits glued markers?

## Test plan

- [x] `npm run lint` clean on all three files
- [x] Full suite, forced (`OFFLINECV_FULL_TESTS=1`): **6,506 passed, 8 files skipped**, up from 6,495 by exactly the 11 cases added
- [x] `fallow audit --base origin/main`: dead code 0, no issues in 3 changed files
- [x] `ATOM` `.source` and `.flags` byte-identical to `main`
- [x] New cases fail against the pre-fix guard; controls stay green

No fixture binaries touched.


s-annam (approved):
Approving: 0 Blocking, 0 Secondary, 0 Nits. The guard now keys on a number rather than a digit, shares its currency class with the atom classifier, and I checked the subset property and the fail-before claim myself rather than taking them from the description.

## Verification

- **#930 acceptance criteria:** all met.
  - The three issue rows, plus `£`, `¥`, `• -$5M` and `-$.5M`, keep their sign. They sit in the `does not eat a number a marker strip uncovers` `it.each` table the AC names.
  - Existing assertions stay green: `post-process.test.ts` + `preserve-numbers.test.ts` 166/166. That includes `-Shipped Foo.` and `-**Led** the migration.`.
  - The tight `-$5M saved.` ambiguity is called out in the docblock, with the same reasoning the digit case uses.
  - Fail-before reproduced: with only the guard reverted to `(?!\.?\d)`, exactly 9 new cases go red (the 7 rows, the property test, the tight-marker test). The `- $5M` control and the `₹` pin stay green. The file was restored afterwards.
  - The docblock now says what the guard keys on.
- **Subset property**, measured independently of your 95-input harness:
  - I compared the old and new `LIST_MARKER_PATTERN` over 300,000 random strings, drawn from an alphabet of `-`, spaces, `$ € £ ¥ ₹`, `.`, digits, `,`, `%`, letters, `*`, `•` and `)`, plus edge probes.
  - 366 inputs differ. In **0** does the new pattern strip more than the old one, and all 366 have the shape `-<currency>.?<digit>`.
  - So the change can only ever strip less, and only where it should.
- **`ATOM`:** the `.source` built by `new RegExp(String.raw…)` is identical to the old literal, flags `g`/`g`. The `∼≈` escapes survive `String.raw` because `RegExp` interprets them.
- **Shared class:** `CURRENCY_SYMBOL_CLASS` is the only `[$€£¥]` left under `src/`. The one other currency list, `job-search/compensation.ts`, maps symbols to ISO codes for salary parsing, a different vocabulary.
- **Gates:**
  - Ran: `tsc -b --noEmit` clean; `eslint` clean on all 3 files; `fallow audit --base origin/main` found no issues in the 3 changed files. Its 3 complexity findings (`isYearContext`, `classifyAtom`, `bareIntegerClaim`) are inherited and excluded by the gate.
  - CI: `verify`, `fallow` and Cloudflare Pages green.
  - Skipped, since no fixtures, components or styles changed: fixture PII, design-system and style tokens.
- **Description (3f):** accurate. The 9-red count, the byte-identical `ATOM`, the "no new import" claim (`post-process.ts` already imported `checkNumbersPreserved`) and the 11 added cases all check out.

## On your review-focus questions

- **`KNOWN GAP` pin vs `it.todo`:** keep the pin. An `it.todo` stays green forever and says nothing when the class is widened. The pin flips the moment it is, which is exactly the signal you want. Adding `₹` here would be wrong for the reason you gave: it changes the classifier too, and that needs its own differential.
- **Tight `-$5M` read as a sign:** agreed. The costs are asymmetric: a stray glyph versus an inverted figure. It matches the `-5%` call, and the spaced `- $5M` marker form, which is what bullets actually look like, still strips.

## Filed

- **#940**: the `₹` gap is wider than the strip. On `origin/main`, `checkNumbersPreserved` returns `ok: true` for `₹2Cr → ₹3Cr`, `₹20 lakh → ₹30 lakh` and `-₹5M → ₹5M`, so the rewrite gate lets a changed rupee figure or a dropped rupee sign ship. Your `KNOWN GAP` test recorded the strip half, but the gate half was not recorded anywhere. The issue proposes widening `CURRENCY_SYMBOL_CLASS` (which your shared constant now makes a one-line change for both consumers) plus deciding how `Cr`/`lakh` magnitudes key, and flipping your pin as an AC.

---

Issue #915, opened 2026-08-28
Shared lexicon (a): one bullet-glyph vocabulary — ⁃ and — are parser bullets the scorer silently drops


s-annam:
Part (a) of #653. **Not** a good first issue — this one changes scoring behaviour and needs per-glyph human adjudication.

### Problem

The bullet-glyph vocabulary exists as divergent copies, and the divergence is live: glyphs the parser treats as bullets are invisible to the scorer, so those bullets drop out of the Specificity pool.

Verified divergence:

| site | class |
|---|---|
| `src/lib/heuristics/line-primitives.ts:35` (`BULLET_CLASS`) | `[•‣▪●◦⁃*\-–—]` |
| `src/lib/heuristics/sections.ts:145` (`VISUAL_BULLET_RE`) | `[•‣▪●◦⁃*\-–—]` |
| `src/lib/heuristics/regex.ts:321` (`LEADING_BULLET_RE`) | `[•‣▪●◦⁃*\-–—]` |
| `src/lib/score/score.ts:136` (`BULLET_MARKER_RE`) | `[-*•●–▪◦‣▶►·�]` |
| `src/lib/score/score.ts:789` (`LONE_BULLET_RE`) | `[•●▪◦‣▶►·�]` |
| `src/lib/edit/apply-overrides.ts:105` (`LEADING_MARKER_RE`) | `[-*•●–▪◦‣▶►·�]` |

Plus copies in `markdown-emit.ts`, `extract/group-bullets.ts` and `webllm/localize/achievements.ts` — 8 in total per the #646 review.

**`⁃` (U+2043 hyphen bullet) and `—` (em dash) are parser bullets absent from the scorer's class.** A résumé using either has its Experience bullets segmented correctly by the parser and then silently skipped by `extractBulletsFromLines` (`score.ts:808-822`), deflating Specificity. Conversely `▶`, `►`, `·` and `�` are scorer-only.

### Why this is separate from #653(b)/(c)/(d)

Merging the sets **changes scores on existing résumés**. That is a deliberate behaviour change requiring a version bump and a reviewed rebaseline, which is why it cannot ride along with the behaviour-preserving consolidations.

### Implementation plan

[...]

PR #941, opened 2026-09-14
refactor(webllm): one local open-ended word list in preserve-numbers, pinned to the lexicon (#938)

### LINKED ISSUE #938: preserve-numbers.ts hand-writes the open-ended date vocabulary six times instead of deriving it from OPEN_ENDED_ALT
## Problem

`src/lib/webllm/preserve-numbers.ts` spells out the open-ended date vocabulary by hand six times on `main`. The canonical copy is `OPEN_ENDED_ALT` in `src/lib/heuristics/regex.ts:145`:

| Line on `main` | Pattern | Site |
|---|---|---|
| `preserve-numbers.ts:303` | `(?:19\d\d\|20\d\d)\|present\|current\|now\|ongoing` after a range dash | `LEADING_DATE_ANCHOR_SEPARATOR` |
| `preserve-numbers.ts:388` | `onwards?\|present\|current\|now\|ongoing` | `YEAR_FOLLOW_CUE` |
| `preserve-numbers.ts:389` | `to\s+(?:…\|present\|current\|now\|ongoing)` | `YEAR_FOLLOW_CUE` |
| `preserve-numbers.ts:390` | `until\s+(?:…\|present\|current\|now\|ongoing)` | `YEAR_FOLLOW_CUE` |
| `preserve-numbers.ts:391` | `through\s+(?:…\|present\|current\|now\|ongoing)` | `YEAR_FOLLOW_CUE` |
| `preserve-numbers.ts:395` | a range dash or `/` followed by the same list | `YEAR_FOLLOW_CUE` |

#916 and #931 consolidated every other copy under `src/lib/heuristics/` onto the lexicon. These six sit outside that directory, so neither issue covered them. #928 is also open against this module, but it is about the dead year tiers and doc gaps, not these copies.

## Why it matters

These patterns decide whether a 4-digit number next to an open-ended word counts as a **year**, which is how `2019 – Present` stays out of the number diff (#876). If the lexicon gains a word (`Till date`, `To date`), the parser starts accepting `2019 – Till date` as a current role. This module would not follow, and the year-classification rules would treat that `2019` differently from `2019 – Present`.

Since #778 a mismatch here is a hard block: a false positive throws away the whole section rewrite.

## The design question to settle first

`preserve-numbers.ts` currently has **no imports at all**. Pointing it at `regex.ts` would give it its first dependency and a cross-lane edge from `webllm/` into `heuristics/`. `score.ts` holds its own month copy for the same near-zero-dep reason, and nobody considers that a bug.

So decide one of these:

1. **Derive:** import `OPEN_ENDED_ALT` (`regex.ts` is pure string/RegExp constants, so it should cost the lazy webllm chunk nothing), and substitute it at all six sites.
2. **Keep it standalone on purpose:** write that down in the module docblock, and add a test that pins the local list to `OPEN_ENDED_ALT`. Drift then fails a test instead of shipping silently.

Either way the language must not change. Both regexes carry the `i` flag, and every site is followed by `\b` or a closing group, so reordering and recasing the alternation is language-identical. Measure it anyway with a before/after differential over the year-classification inputs in `preserve-numbers.test.ts`.

## Acceptance criteria

- [ ] No hand-written `present|current|now|ongoing` alternation remains in `src/lib/webllm/preserve-numbers.ts`, **or** the module docblock records why it stays standalone and a test fails when the local list and `OPEN_ENDED_ALT` differ.
- [ ] Before/after differential over `checkNumbersPreserved` (the module's only export) shows zero differences.
- [ ] `npm run verify` green.

Found while reviewing PR #937 (#931); the PR description noted these copies and deliberately left them unfiled.

Vaishnavi1709:
## Summary

`src/lib/webllm/preserve-numbers.ts` spelled out the open-ended date words (`present|current|now|ongoing`) by hand six times, across `LEADING_DATE_ANCHOR_SEPARATOR` and `YEAR_FOLLOW_CUE`. Those patterns decide whether a 4-digit number next to `Present` counts as a **year**. If the parser lexicon ever gained a word, this gate would silently disagree with the parser about what makes a year. Since #778, a false positive there throws away the whole rewrite.

This collapses the six copies into **one local `OPEN_ENDED_WORDS`** and builds every site from it. A test fails if that list and the parser's `OPEN_ENDED_ALT` ever differ.

Closes #938

## The design question #938 asked to settle first

The issue offered two options: import `OPEN_ENDED_ALT` from `regex.ts` (derive), or keep a standalone list pinned by a test. **This PR keeps the list standalone, based on measurement:**

- **The rewrite lane does not load `regex.ts` today.** I walked the static import graph from `rewrite-section.ts` and `rewrite-summary.ts`, 11 modules each, and neither reaches `heuristics/regex.ts`.
- **`regex.ts` is not a cheap import.** It does real work when it loads: `COUNTRY_GAZETTEER = _buildGazetteer()` walks `Intl.DisplayNames`. Importing it would put that cost on the rewrite lane for four words, the eager-graph problem #915 warns about.
- **The test-only export is house practice.** `useDownloadPdf.ts` and `work-authorization.ts` already export for their tests, and fallow treats `src/**/*.test.ts` as entry points, so `OPEN_ENDED_WORDS` is not flagged as unused.

The reasoning is written into the constant's docblock, with a pointer from the module docblock.

## Drift fails a test instead of shipping

The new `describe` block pins the local list to `OPEN_ENDED_ALT` in two ways:

1. **As a word set:** the two lists must contain the same words, ignoring case.
2. **By behaviour:** every word in the parser's list goes through a probe where the neighbouring word decides the result. Each must behave like `Present`, and the controls `Tuesday` and `Presently` must not. `Presently` checks the `\b` word boundary, not just a different word.

I mutated both directions and confirmed each fails:

| mutation | set pin | behaviour pin |
|---|---|---|
| lexicon gains a word (`OPEN_ENDED_ALT` + `Till date`) | ❌ fails | ❌ fails |
| local list loses a word (`OPEN_ENDED_WORDS` − `ongoing`) | ❌ fails | ❌ fails |

Both files were restored byte-identical afterwards, confirmed with `cmp`.

**One thing worth knowing about the behaviour probe.** My first draft probed a lone dropped year, and it **passed for every word, including the controls**: a dropped year is reported whatever sits beside it, so that test proved nothing. I measured seven sentence shapes. The one used here adds a decoy bullet carrying an unrelated `2019`. Beside an open-ended word, dropping the year is still reported; beside an ordinary word, the decoy excuses it. The test's comment records this so nobody "simplifies" the decoy away.

## Behaviour-preserving: measured

A temporary harness, removed before commit, ran `checkNumbersPreserved`, the module's only export, over **5,274 inputs × 4 calls (identity, year dropped, year changed, same-value decoy) = 21,096 results**, before and after. The inputs covered:

- every case of the four words, plus near-misses (`Presently`, `Nowhere`, `Till date`)
- every range joiner (`–`, `-`, `—`, `to`, `until`, `through`, `/`, `and`)
- several lead-ins
- every string literal in the module's own test file

It found **zero differences**. In 5,183 of those inputs, dropping the year changes the verdict, so the harness genuinely exercises year classification.

Four of the six sites were double-quoted strings and became template literals. Both regexes carry `i`, and every site is closed by `\b` or a group, so this is language-identical by construction as well as by measurement.

## Not in this PR

- **#928** (dead year tiers and doc gaps in the same module) is deliberately left for a separate PR, even though that issue suggests folding it into the next webllm PR.
- **fallow reports one duplicated block** in `preserve-numbers.test.ts`: a 9-line `for … of mergePairs` assertion loop around lines 768 and 861. It was already on `main`, and both spans are byte-identical to `main` apart from a one-line shift from the added import. It appears only because this PR touches the file and fallow scopes its metrics to changed files. The gate reports no issues.

## Review focus

- `src/lib/webllm/preserve-numbers.ts`, `OPEN_ENDED_WORDS`: is keeping this module import-free worth a pinned local copy, or would you rather take the `regex.ts` import and its load cost for a single definition?
- `src/lib/webllm/preserve-numbers.test.ts`, the behaviour pin: does the decoy-bullet probe read as clearly as it should, given a simpler-looking probe passes vacuously?

## Test plan

- [x] `npm run lint` clean on both files
- [x] Full suite, forced (`OFFLINECV_FULL_TESTS=1`): **6,508 passed, 8 skipped**, up from 6,506 by exactly the 2 tests added
- [x] `fallow audit --base origin/main`: dead code 0, no issues in 2 changed files
- [x] Before/after differential: 0 differences over 21,096 results
- [x] Both pins fail when the lexicon gains a word and when the local list loses one

No fixture binaries touched.


shubhransh-gupta (approved):
Approving: 0 Blocking, 0 Secondary, 0 Nits. The six handwritten open-ended word lists in `preserve-numbers.ts` now collapse into one local `OPEN_ENDED_WORDS`, keeping the module completely dependency-free, and I independently verified regex byte-equivalence and both mutation claims.

## Verification

- **#938 acceptance criteria:** all met.
  - **No handwritten duplicates remaining:** All 6 sites in `preserve-numbers.ts` (1 in `LEADING_DATE_ANCHOR_SEPARATOR`, 5 in `YEAR_FOLLOW_CUE`) derive from `OPEN_ENDED_WORDS`. Grepping `present|current|now|ongoing` across `preserve-numbers.ts` returns only the constant declaration itself.
  - **Standalone justification documented:** The module docblock and the `OPEN_ENDED_WORDS` docblock thoroughly record why it remains standalone rather than importing `regex.ts`.
  - **Divergence guard:** Two tests in `preserve-numbers.test.ts` pin the local constant to `OPEN_ENDED_ALT` (both by case-insensitive word set and by behavioral classification).
  - **Differential:** Measured independently; `LEADING_DATE_ANCHOR_SEPARATOR` and `YEAR_FOLLOW_CUE` compile to `.source` and `.flags` strings byte-identical to `main`.
  - **`npm run verify`:** Clean across all gates.

- **Equivalence & Regex properties:**
  - Evaluated the compiled regular expressions before and after:
    - `LEADING_DATE_ANCHOR_SEPARATOR.source` and `.flags` match `main` byte-for-byte.
    - `YEAR_FOLLOW_CUE.source` and `.flags` match `main` byte-for-byte.
  - `OPEN_ENDED_WORDS` is a bare alternation (`"present|current|now|ongoing"`), introducing zero capture groups, which guarantees indexed group numbering in any regex consumer remains intact.
  - Word boundaries (`\b`) closing each site correctly bound all four terms.

- **Mutation claims:** reproduced independently.
  1. **Lexicon gains a word:** Appending `|Till date` to `OPEN_ENDED_ALT` in `src/lib/heuristics/regex.ts` immediately caused both tests in `describe("open-ended vocabulary stays in step with the parser lexicon (#938)")` to fail (the set assertion failed on the unexpected word, and the behavioral probe failed because `Till date` was not recognized as a strict year context).
  2. **Local copy loses a word:** Dropping `ongoing` from `OPEN_ENDED_WORDS` in `src/lib/webllm/preserve-numbers.ts` caused both the set check and behavioral test to fail.
  3. Both files were restored and re-verified.

- **Gates:**
  - `npm run typecheck`: clean (0 errors).
  - `npm run lint`: clean (0 errors).
  - Full test suite: **6,508 passed, 8 skipped**, up from 6,506 on `main` by exactly the 2 added tests.
  - `fallow audit --base origin/main`: 0 dead code, 0 dead exports, 0 issues in the 2 changed files. (The duplication finding at lines 768/861 is pre-existing on `main`).
  - CI `verify` and Cloudflare Pages preview deployment: green.
  - Skipped (no UI components, styles, or fixture binaries touched): style tokens, fixture PII.

- **Description (3f):** accurate.
  - The eager-load cost of `regex.ts` is real: `COUNTRY_GAZETTEER = _buildGazetteer()` eagerly allocates and calls `Intl.DisplayNames` over ISO country codes at module load.
  - `preserve-numbers.ts` remains 100% import-free.
  - Exactly 2 tests added.

## On your review-focus questions

- **Standalone local copy vs importing `regex.ts`:**
  Keeping `preserve-numbers.ts` standalone and import-free is the right call. The rewrite lane (`webllm/`) is lazily loaded and lightweight; importing `regex.ts` would eagerly pull in the entire country gazetteer and `Intl` setup just for four static words. Because the set test and behavioral test in `preserve-numbers.test.ts` run during CI on every commit, silent drift is impossible.

- **Decoy-bullet probe clarity:**
  The decoy-bullet probe is well-reasoned and necessary. Because a dropped year with no corresponding number anywhere in the output is unconditionally flagged as missing (regardless of context), a single-bullet probe would pass vacuously for any token (including arbitrary words like `Tuesday`). The decoy bullet (`"Counted 2019 units."`) provides an un-cued identical figure: for strict open-ended date words, the drop is unexcused (`ok: false`), whereas for ordinary words/near-misses, the decoy excuses the drop (`ok: true`). The explanatory comment ensures maintainers won't remove the decoy in future refactors.

src/lib/webllm/preserve-numbers.ts
```diff
@@ -313,7 +338,7 @@ const MONTH_NAME =
  */
 const LEADING_DATE_ANCHOR_SEPARATOR = new RegExp(
   `^\\s*(?:[:.)]` +
-    `|(?:${RANGE_DASH.source}\\s*(?:(?:19\\d\\d|20\\d\\d)|present|current|now|ongoing)\\b)` +
+    `|(?:${RANGE_DASH.source}\\s*(?:(?:19\\d\\d|20\\d\\d)|${OPEN_ENDED_WORDS})\\b)` +
```
s-annam, line 341: Nit, nothing to do in this PR. At this site `${OPEN_ENDED_WORDS}` is a dead alternative. The third branch, `${RANGE_DASH.source}(?!\\s*\\d)`, already matches a dash followed by any non-digit, and `.test()` only cares whether something matches. So `– Present` gets through this separator either way.

I checked it two ways:

- **Differential.** I deleted `|${OPEN_ENDED_WORDS}` from this line only and compared against this branch. Across the same 561 literals and 432 year probes there were 0 differences, and the whole `src/lib/webllm/` suite (579 tests) still passes.
- **Consequence for the pins.** Neither new pin can catch a regression at this site. That doesn't matter, because the site doesn't change the result.

This is the same class as #928's dead year tiers, so I added it there instead of filing a new issue.

s-annam (approved):
A clean, well-measured refactor. It settles #938's design question with evidence, and I could reproduce every check it relies on.

**Verdict rule:** 0 Blocking → `APPROVE`. Reviewed head `3bf2bbe92a8e135652c04154f5d3293ca0719ab7`.

## Acceptance criteria (#938)

- [x] **No hand-written alternation remains, and the drift pin exists.** A grep for `present|current|now|ongoing` in `preserve-numbers.ts` now finds only `OPEN_ENDED_WORDS` itself. The module docblock and the constant's docblock both record why the list stays local, and the set pin compares it to `OPEN_ENDED_ALT`.
- [x] **Zero differences before and after.** I ran my own comparison of `origin/main` against this branch rather than relying on the PR's harness. It covered 561 numeric string literals from `src/lib/webllm/*.test.ts`, plus 432 year-context probes: 12 words (including `Presently`, `nowhere` and mixed case) in 9 range shapes, each checked 4 ways. There were 0 differences.
- [x] **`npm run verify` green.** CI `verify` and `fallow` pass. Locally, both touched test files (168 tests), typecheck and eslint pass, and `fallow audit` reports no issues in the 2 changed files.

I also checked the claims behind the design choice:

- **Import graph.** Walking the static, non-type imports from `rewrite-section.ts` and `rewrite-summary.ts` reaches only `heuristics/types.ts` under `heuristics/`, never `regex.ts`.
- **Load-time cost.** `COUNTRY_GAZETTEER = _buildGazetteer()` does run at module load (`regex.ts:125`).
- **Review focus Q1.** I agree: a pinned local copy is the right trade for four words.
- **Both pins discriminate.** I mutated the code 3 ways, and both tests went red each time: removing `now` from the local list, and adding `Till date` or `Today` to `OPEN_ENDED_ALT`. Hand-writing the `YEAR_FOLLOW_CUE` dash site back without `now` fails the behaviour pin.

## Nits (non-blocking)

1. `preserve-numbers.ts:341` (inline): at this site the open-ended words are dead alternatives. Neither pin can see them, which is harmless. The details are in the inline comment, and I added the item to #928, so there is nothing to do in this PR.

## Heads-up: textual conflict with #943

This PR and #943 each append a new `describe` block at the end of `preserve-numbers.test.ts`. A trial merge of the two heads conflicts in that file; `preserve-numbers.ts` itself auto-merges. The resolution is to keep both blocks. Whichever PR reaches the queue second needs a rebase, and must stay at one commit afterwards.

This PR comes from a contributor branch, so nothing was pushed to it.

---

src/lib/webllm/preserve-numbers.ts
s-annam: I added this as item 5 on #928: https://github.com/offlinecv/OfflineCV/issues/928#issuecomment-5682820624. Nothing to change in this PR, so I am resolving this thread to keep it from blocking the merge.

Issue #810, opened 2026-08-13
The score says which dimension is low, never what to change


s-annam:
## Problem

Three respondents in the Aug 2026 feedback round asked for the same missing thing: the score tells them a number is low, never what to do about it.

> "Provide detailed explanations for why the resume score is low and how each issue affects ATS compatibility."

> "It'll be okay if you say what needs to be changed in the resume"

> "It would be better to give the suggestions for updating resume"

A fourth, otherwise positive, named it alongside parsing: *"provide clearer explanations for low scores."*

*(Verbatim, apostrophes normalized. No respondent identifiers.)*

## What the app shows today

`AtsScoreReadout.tsx` (187 LOC) renders "three dimension cards, layout penalty note, and algo version footer." The dimension hints are counts, not actions — its own docblock records the regression:

> "shortened the Structure hint to bare counts (`length 13/23`), which no [longer explains anything]"

`LAYOUT_TRIGGER_BLURBS` (`src/lib/heuristics/trigger-copy.ts`) is the one place plain-language guidance exists, and it covers **layout triggers only** — two-column, scanned, unmappable fonts. There is nothing equivalent for a low Specificity, Structure, or Completeness score, which is where most résumés lose points.

So a user with a clean single-column PDF scoring 58 sees three numbers, no triggers, and no next step.

`getScoreRecommendation` (`src/lib/score/recommendation.ts`) already turns the score into one sentence naming the weakest gradable dimension, rendered under the verdict band by `VerdictHeader.tsx`. That sentence is a pointer, not the guidance this issue adds — it says which dimension is weak, not which bullet or field to fix.

## What this is NOT (already tracked elsewhere)

- **#43** — `docs/scoring.md`, a written rubric. Documentation only; it explicitly defers the in-app link as "a separate, smaller follow-up". Complementary, not overlapping.
- **#493** step 3 ("Explain") — per-section messaging, but scoped to **zero-yield sections**: a section that parsed to nothing. It does not cover a section that parsed fine and simply scored low. Its Messaging copy child, once filed, is a dependency of this issue.
- **#680** item 3 — score *framing* ("the score rates the file, not you"). Framing, not guidance.

The gap this issue owns: **dimension score → concrete action**, for résumés that parsed correctly.

## Scope

Map each scoring dimension to guidance the user can act on, sourced from the same signals the score already computes — no new analysis pass, no LLM dependency.

| Dimension | Weight | Signal already available | Guidance shape |
|---|---|---|---|
[...]

PR #942, opened 2026-09-14
feat(feedback): inline nudge instead of an auto-opening modal, cooldown instead of a lifetime cap (#912)

### LINKED ISSUE #912: Feedback session UX: research + propose flow for the always-visible feedback widget

Samhit21:
Resolves #912.

Implements proposals **A** and **B** from the research pass on #912 ([writeup](https://github.com/offlinecv/OfflineCV/issues/912#issuecomment-5464174007)).

> ⚠️ **Scope has not been signed off.** #912 asks for a research pass and a scope discussion with @s-annam before any code, and #923 (this same change) was closed on 2026-08-30 to hold that discussion. No reply has landed on the writeup since. This PR re-opens the work at the assignee's request, rebased onto current `main`, as a concrete proposal to react to. The two open product calls below are deliberately **not** in it.

## Context: the issue's premise was stale

#912 was written against `FeedbackPanel.tsx`, which no longer exists: #900/#903 retired it after the 2026-08-27 sync. That change already answered three of the issue's five questions: the panel is no longer always visible, the flow already branches by rating, and step 1 is already a bare star row that expands. This PR does what was left.

## 1. The milestone raises a nudge, not the dialog

`Dialog` is a native `showModal()`. Opening it automatically took keyboard focus off whatever the user was doing, and a screen reader announced it as an interruption. That's the wrong instrument for an invitation. It's also the answer to the issue's accessibility question, which got **worse** with #900: the old concern was a panel below the fold, the new one a focus trap arriving uninvited.

NN/g's [User-Feedback Requests](https://www.nngroup.com/articles/user-feedback/) endorses asking after task completion and keeping an always-available way in, both of which #900 got right, while explicitly discouraging modal popups for the ask itself.

So a résumé export now shows `FeedbackNudge`: inline, dismissible, the star row plus one line, announced with `role="status"` (polite, so it waits behind whatever the reader was saying about the export). Picking a star opens the dialog **on that rating's branch** through a new `initialRating` prop, so nobody is asked for the same star twice.

## 2. Cooldown instead of a lifetime cap

The automatic ask was gated on `seenCount > 0`, so it was shown once, ever. And `openDialog()` incremented that same counter, so **a user who clicked the ambient `[★ Feedback]` button out of curiosity permanently disabled the automatic ask.**

| | before | after |
|---|---|---|
| Automatic ask | once per browser, forever | every 14 days |
| Ambient open | burns the automatic trigger | untouched |
| After submitting | never auto-asked again | unchanged |

`ocv_feedback_prompted_at` records only unprompted asks, and it's written when the nudge becomes **visible**, not when it's answered. A nudge scrolled past is still an ask received, and re-asking tomorrow because it was ignored is the nagging this exists to prevent.

- **Migration is a deliberate one-time re-ask.** Neither `ocv_feedback_dialog_seen` (#900) nor `ocv_feedback_seen` (the retired panel) is read. Every returning tester holds a value under both, and honouring them would keep the lifetime cap alive for exactly the people this change is for.
- **The cooldown fails open.** An unreadable timestamp asks rather than staying silent.

## Found and fixed in this pass: the ambient open received the click event

In #923 the open was `openDialog(rating = 0)`, so the nudge could pass a star through the same function. But `ParsedHeader` and `PageShell` hand `openDialog` straight to a `Button` `onClick`, and `Button` spreads onto `<button>`, so React called it with the **`MouseEvent`**, and that event became `initialRating`. The dialog still landed on the star step by accident (an object compared with `>= 1` is false), but `rating` state held an event object. TypeScript can't see it, because a `() => void` prop accepts a function with an optional parameter.

The fix gives the controller two typed entry points: `openDialog()` takes nothing, and `rateFromNudge(rating)` is the nudge's. A regression test calls `openDialog` the way `onClick` does, with an event argument, and asserts `initialRating` is `0`. Restoring the optional parameter fails that test.

## Deliberately not in scope

- **Which milestone earns the ask.** Export is the rarest success moment in the funnel. Changing it is a product call.
- **Per-parse vs per-browser memory,** the issue's question 3. It falls out cheaply if the cooldown key incorporates `parseKey`, but it's a decision, not a refactor.
- **A dismissed rating is lost entirely.** Fixing it changes the `feedback_submitted` event shape, which #912 puts out of scope; it wants its own issue.

## Verification

- **Rebased onto `main`** (`583de21`). None of the 7 intervening commits touched these files.
- **Tests:** 106 pass across the 11 affected files: the feedback hook, `FeedbackNudge`, `FeedbackDialog`, `ExportDialog`, `ParsedHeader`, `PageShell`, `Result`, and the four `App` suites.
- **Typecheck and lint:** `tsc -b --noEmit` exits 0, and `eslint` is clean on every changed file.
- **Pre-push `npm run verify` passed.** Fallow notes one complexity finding: `App.tsx`'s `App` function (38 cyclomatic, 787 lines). That's pre-existing debt from CLAUDE.md's known-debt list; this change adds 8 lines to it without changing its shape. Fallow is report-only, and the audit gate excluded 3 inherited findings.
- **The tests with teeth:**
  - the milestone raises the nudge and **not** the dialog;
  - the ambient button does **not** write the cooldown key;
  - it's suppressed inside the window **and** asks again past it;
  - both retired keys are ignored;
  - an unreadable timestamp asks;
  - the nudge is not a `dialog` and has no `aria-modal`;
  - the ambient open ignores a click event.


src/hooks/useFeedbackDialog.ts
```diff
@@ -2,95 +2,169 @@
...
   function notifyExportClosed(): void {
     if (!pendingRef.current) return;
     pendingRef.current = false;
-    openDialog();
+    // The cooldown starts when the ask becomes VISIBLE, not when it is
+    // answered. A nudge the user scrolled past is still an ask they received,
+    // and re-asking tomorrow because they ignored it is the nagging this
+    // exists to prevent.
+    setPromptedAt(String(Date.now()));
+    setNudgeVisible(true);
```
s-annam: **Blocking — stale nudge across parses.** `nudgeVisible` is cleared only by `openAt` and `dismissNudge`, and the controller lives at `App` level for the whole visit.

Repro: parse résumé A → export → close the export dialog (nudge shows) → ignore it → drop résumé B. While B parses `state.phase !== "done"` hides it, but the state is still `true`, so the moment B lands the nudge reappears — for a résumé that was never exported. That contradicts the comment above the render (*"raised by an action the user just took"*).

Fix: clear `nudgeVisible` and `pendingRef` when `parseKey` changes — the same `useEffect(…, [parseKey])` reset pattern `App.tsx` already uses for the JD context. Once #1005's per-parse keying lands, the hook can take `parseKey` and own this itself. Please add a test: nudge visible → new `parseKey` → not visible.

src/App.tsx
```diff
@@ -634,6 +635,16 @@ export default function App() {
       )}
 
       <ErrorBoundary onReset={reset}>
+        {/* #912 — the export milestone's ask. Above the result rather than
```
s-annam: **Blocking — placement.** This comment's reasoning (*"an invitation they have to scroll to find is one they never see"*) is right, but the top of the page is exactly that place for a scrolled user: the export dialog opens from the sticky `PageShell` header, so the user may be deep in the résumé when it closes. For an unscrolled user, the block pushes the score card down — the slot #955 just settled.

Suggested home: the fixed bottom dock that #1000 introduced (`FixItDock`, which already measures itself and publishes `--fixit-dock-clearance`). It is visible wherever the user is, moves no layout, and is where the Fix It milestone (review body, item 3) wants the nudge anyway — one slot, both milestones.

src/hooks/useFeedbackDialog.ts
```diff
@@ -2,95 +2,169 @@
...
+ * dismissed it while busy is reachable in the next round of testing. A number
+ * rather than a `seen < N` count on purpose — the failure mode being fixed is
+ * asking too often, and only elapsed time measures that.
+ *
+ * Module-private: nothing outside needs it, and the tests deliberately drive
+ * timestamps well inside and well outside the window rather than deriving from
+ * this value — a test that computes its fixture from the constant it is
+ * checking passes for any constant, this one included.
+ */
+const FEEDBACK_COOLDOWN_MS = 14 * 24 * 60 * 60 * 1000;
```
s-annam: **Blocking — per-parse eligibility (#1005 scope item 5).** With one browser-wide timestamp, a tester who parses three résumés this week is asked about the first only. We want each new result rated once. Proposal: record prompted-at per `parseKey` (one ask per parse), and keep a short global cooldown only as the anti-nag backstop, e.g. a day rather than 14. Exact numbers are your call — say which you pick and why in the docblock, like you did here.

s-annam (changes requested):
Solid work, and most of it stays: the nudge instead of an auto-opening modal, the cooldown instead of a lifetime cap, the ambient button no longer burning the trigger, failing open on a bad timestamp, and the typed `openDialog()` / `rateFromNudge()` split are all right. The tests have teeth (fixtures independent of `FEEDBACK_COOLDOWN_MS`, exact call args, no-`aria-modal` assertion). **Scope is now signed off** — see the closing comment on #912 — with the changes below. They are the difference between "quieter modal" and what we actually want: **more vocal about asking, especially in Fix It mode.**

**Heads-up: I rebased this branch onto current `main` (`eb10119`, includes #955 and #1000 Fix It) and force-pushed it (`3b3ad82` → `3e02387`), with a lease pinned to your head.** Your authorship is preserved. Your local branch is now stale: `git fetch && git reset --hard origin/sn/912-feedback-nudge-cooldown` before you continue. After the rebase: typecheck clean, the 13 affected test files (127 tests) pass, `eslint` clean on changed files, pre-push `verify` green.

Verdict rule: ≥1 Blocking → REQUEST_CHANGES.

## Blocking

1. **A nudge earned by résumé A shows up on résumé B** (inline, `useFeedbackDialog.ts`). Nothing clears `nudgeVisible` when the parse changes.
2. **The nudge renders above `<Result>`** (inline, `App.tsx`). Since #955/#1000 that is the docked score card's slot, and the export dialog opens from the sticky header, so the user can be anywhere on the page.
3. **Fix It completion is not a milestone.** This is the main scope addition. `FixItFinished` (#810, reached from `FixItToolbar` inside `Result.tsx`) is a task-done moment far more users reach than export. Reaching it — either the resolved state or stepping past the last item — should raise the nudge **inside the Fix It dock** (`FixItDock`). That also gives blocking item 2 a natural home: the same fixed bottom dock slot can host the export-triggered nudge, so no path pushes the page layout around.
4. **Eligibility is per browser, not per parse** (inline, `FEEDBACK_COOLDOWN_MS`). A new résumé is a new experience to rate. Key the ask on `parseKey`: one ask per parse, with a shorter global cooldown as the anti-nag guard. `ocv_feedback_submitted` stays per browser and permanent, as you have it.

## Secondary

- **Retarget the link.** #912 is closed as research-complete, so `Resolves #912` now closes nothing. The agreed implementation scope is **#1005** — please switch to `Closes #1005`, and check the PR against its acceptance criteria.
- **Description is stale after the rebase.** "Rebased onto `main` (`583de21`)… none of the 7 intervening commits touched these files" is no longer true: `main` has moved 24 commits and `App.tsx` changed underneath this PR. The "⚠️ Scope has not been signed off" banner can go too.

## Filed from this review

- **#1005** — the agreed scope this PR should close (A + B from your writeup, plus Fix It milestone, placement, per-parse).
- **#1006** — a rating picked then dismissed is never recorded. Your "flag for its own issue" item; kept out of this PR because it changes the `feedback_submitted` event shape.

---
Reviewed by: Claude Opus 5.5 (high)

PR #943, opened 2026-09-14
fix(preserve-numbers): recognize rupee figures and Indian magnitudes (#940)

### LINKED ISSUE #940: checkNumbersPreserved does not see rupee figures, so a rewrite can change ₹2Cr to ₹3Cr or drop the sign on -₹5M
## Problem

The number-preservation gate (`checkNumbersPreserved`, `src/lib/webllm/preserve-numbers.ts`) is what reverts a WebLLM rewrite that changes a figure. It does not see rupee figures at all. So a rewrite that changes a rupee amount, or drops its minus sign, passes the gate and ships.

The cause is the currency class. `ATOM` (`preserve-numbers.ts:158` on `main`) recognises only `$ € £ ¥`. As a result:

- `₹` is never part of an atom.
- The `-` in `-₹5M` never reaches the digits, so the sign is not keyed.
- A figure glued to an Indian magnitude word, like `2Cr`, fails the atom's trailing `(?!\w)` and is not extracted as a number.

Reproduced on `origin/main` @ `2c1a8e4`, calling `checkNumbersPreserved([input], [output])` directly:

| input → output | result on `main` | wanted |
|---|---|---|
| `Saved $2M in costs.` → `Saved $3M in costs.` | `ok: false` (control) | `ok: false` |
| `Saved ₹2Cr in costs.` → `Saved ₹3Cr in costs.` | **`ok: true`** | `ok: false` |
| `Saved ₹20 lakh in costs.` → `Saved ₹30 lakh in costs.` | **`ok: true`** | `ok: false` |
| `Cut spend by -₹5M.` → `Cut spend by ₹5M.` | **`ok: true`** | `ok: false` |
| `Cut spend by -$5M.` → `Cut spend by $5M.` | `ok: false` (control) | `ok: false` |
| `Cut spend by ₹5M.` → `Cut spend by ₹9M.` | `ok: false`, keyed as bare `5M` | `ok: false`, keyed with `₹` |

## Why it matters

The gate exists so the on-device rewrite cannot change a résumé's facts (#778). For a résumé written in rupees, it silently does nothing:

- `₹2Cr` → `₹3Cr` is a fabricated 50% bigger achievement.
- `-₹5M` → `₹5M` inverts a cost reduction into spend.

Both lines stay grammatical and plausible, so nothing downstream looks wrong. Indian résumés use `₹` with `Cr`/`crore` and `L`/`lakh` routinely.

A second consumer is affected. Since #930 (PR #939), `post-process.ts`'s list-marker guard reads the same `CURRENCY_SYMBOL_CLASS` to decide whether a leading `-` is a minus sign. So `cleanRewriteLine("-₹2Cr loss.")` still strips the sign to `₹2Cr loss.`. #939 pins that shape as an explicit `KNOWN GAP` test in `post-process.test.ts`, which flips when the class is widened.

## Proposed fix

1. **Add `₹` to `CURRENCY_SYMBOL_CLASS`** in `preserve-numbers.ts`. That widens both consumers at once, which is why #939 made it a shared constant.
2. **Decide how Indian magnitudes key.**
   - Glued forms: `2Cr`, `20L`.
   - Spaced forms: `20 lakh`, `2 crore`.
   - The magnitude suffix group today is `[kKmMbBgGtT][bB]?`, so `Cr` is not recognised.
   - Either extend the suffix group (mind `L` vs a trailing letter in prose), or treat the spaced word forms as part of the atom.
   - Settle this with a differential over `preserve-numbers.test.ts` first, since widening the suffix can change what existing inputs extract.
3. **Flip the `KNOWN GAP` test** in `post-process.test.ts`. `-₹2Cr loss.` should then keep its sign. The property test over `CURRENCY_SYMBOL_CLASS` picks up `₹` automatically.

## Acceptance criteria

- [ ] Every "wanted" row in the table above holds.
- [ ] `cleanRewriteLine("-₹2Cr loss.")` returns `-₹2Cr loss.`, and the `KNOWN GAP` test is replaced by a positive assertion.
- [ ] A before/after differential over `checkNumbersPreserved` on the existing `preserve-numbers.test.ts` inputs shows no change outside rupee and Indian-magnitude figures.
- [ ] `npm run verify` green.

Found while reviewing PR #939 (#930), which deliberately kept `₹` out of the shared class and pinned the resulting sign loss as a known gap. The gate-side miss shown in the table was not recorded anywhere.

shubhransh-gupta:
## Summary

The number-preservation gate (`checkNumbersPreserved`, `src/lib/webllm/preserve-numbers.ts`) reverts WebLLM rewrites that alter figures or flip signs. Previously, it did not recognize rupee figures (`₹`) or Indian magnitude abbreviations (`Cr`, `L`). As a result:

- A rewrite altering a rupee amount (e.g., `Saved ₹2Cr in costs.` → `Saved ₹3Cr in costs.`) passed undetected.
- Dropping a leading minus sign before a rupee figure (e.g., `Cut spend by -₹5M.` → `Cut spend by ₹5M.`) was not caught.
- In `post-process.ts`, `cleanRewriteLine("-₹2Cr loss.")` stripped the leading minus sign to `₹2Cr loss.` because `₹` was missing from `CURRENCY_SYMBOL_CLASS` (previously recorded as a `KNOWN GAP` in `post-process.test.ts`).

Resolves #940

## The change

1. **Include `₹` in `CURRENCY_SYMBOL_CLASS`** (`src/lib/webllm/preserve-numbers.ts`):
   - Widens `CURRENCY_SYMBOL_CLASS` from `"[$€£¥]"` to `"[$€£¥₹]"`.
   - Shared directly with `post-process.ts`'s list-marker pattern, resolving the `KNOWN GAP` and preserving leading minus signs before rupee amounts.
2. **Extend `ATOM` magnitude pattern** to recognize Indian magnitudes `Cr` and `L`:
   - Updated from `(?<magnitude>[kKmMbBgGtT][bB]?)` to `(?<magnitude>[kKmMbBgGtT][bB]?|[cC][rR]|[lL])`.
   - `Cr` and `L` are bounded by the atom's trailing `(?!\w)` lookahead, preventing false matches inside prose words.
   - Glued forms like `₹2Cr` and `₹20L` are extracted with their magnitude suffixes.
   - Spaced forms like `₹20 lakh` / `₹2 crore` extract `₹20` as a currency-decorated numeric atom, correctly flagging any altered number (`₹20 lakh` → `₹30 lakh`).
3. **Flip the `KNOWN GAP` test in `post-process.test.ts`**:
   - Replaced the pinned gap assertion with positive assertions confirming that `-₹2Cr loss.`, `- -₹2Cr loss.`, and `• -₹2Cr loss.` keep their leading minus sign.
4. **Comprehensive test coverage in `preserve-numbers.test.ts`**:
   - Added tests covering every case in the #940 issue specification table, sign loss on rupee negatives, and identical preservation of Indian magnitude figures.

## Acceptance criteria verification

| input → output | result on `main` | with this PR | Status |
|---|---|---|---|
| `Saved $2M in costs.` → `Saved $3M in costs.` | `ok: false` | `ok: false` | Pass (control) |
| `Saved ₹2Cr in costs.` → `Saved ₹3Cr in costs.` | `ok: true` (silent pass) | `ok: false` (`dropped: ["₹2Cr"]`, `added: ["₹3Cr"]`) | Pass |
| `Saved ₹20 lakh in costs.` → `Saved ₹30 lakh in costs.` | `ok: true` (silent pass) | `ok: false` (`dropped: ["₹20"]`, `added: ["₹30"]`) | Pass |
| `Cut spend by -₹5M.` → `Cut spend by ₹5M.` | `ok: true` (silent pass) | `ok: false` (`dropped: ["-₹5M"]`, `added: ["₹5M"]`) | Pass |
| `Cut spend by -$5M.` → `Cut spend by $5M.` | `ok: false` | `ok: false` | Pass (control) |
| `Cut spend by ₹5M.` → `Cut spend by ₹9M.` | `ok: false` (bare `5M`) | `ok: false` (`dropped: ["₹5M"]`, `added: ["₹9M"]`) | Pass |

## Test plan

- [x] `npm run typecheck` clean (0 errors)
- [x] `npm run lint` clean (0 errors)
- [x] Full test suite: **6,511 passed, 10 skipped across 383 test files**
- [x] `cleanRewriteLine("-₹2Cr loss.")` returns `-₹2Cr loss.`
- [x] Pre-push verification gate (`npm run verify`) passed completely (typecheck, lint, fixtures, baselines, core package, test:changed, build, fallow audit)
- [x] Single clean commit adhering to repository conventions


src/lib/webllm/preserve-numbers.ts
```diff
@@ -167,7 +168,7 @@ export const CURRENCY_SYMBOL_CLASS = "[$€£¥]";
  * positionally, and the group order is not the order they are assembled in.
  */
 const ATOM = new RegExp(
-  String.raw`(?<!\w)(?<approx>[~\u223C\u2248])?(?<sign>-)?(?<currency>${CURRENCY_SYMBOL_CLASS})?(?<digits>\d{1,3}(?:,\d{3})+(?:\.\d+)?|\d+\.\d+|\d+)(?:(?<magnitude>[kKmMbBgGtT][bB]?)|(?<multiplier>[xX]))?(?<percent>%)?(?<plus>\+)?(?!\w)`,
+  String.raw`(?<!\w)(?<approx>[~\u223C\u2248])?(?<sign>-)?(?<currency>${CURRENCY_SYMBOL_CLASS})?(?<digits>\d{1,3}(?:,\d{3})+(?:\.\d+)?|\d+\.\d+|\d+)(?:(?<magnitude>[kKmMbBgGtT][bB]?|[cC][rR]|[lL])|(?<multiplier>[xX]))?(?<percent>%)?(?<plus>\+)?(?!\w)`,
```
Vaishnavi1709, line 171: **Secondary.** `[cC][rR]|[lL]` is added to the magnitude group for **every** figure, not only `₹` ones, so a glued `L` or `Cr` on an ordinary quantity now becomes a tracked atom. Checked on `main` vs this head:

| rewrite | `main` | this PR |
|---|---|---|
| `Tuned a 2L engine.` → `Tuned a 2-liter engine.` | `ok: true` | `ok: false`, `dropped: ["2L"]` |
| `Ran a 16L dosing line.` → `Ran a 16-litre dosing line.` | `ok: true` | `ok: false`, `dropped: ["16L"]` |
| `Scaled brewing to 500L batches.` → `…500-litre batches.` | `ok: true` | `ok: false`, `dropped: ["500L"]` |
| `Filed 3Cr compliance reports.` → `Filed three CR compliance reports.` | `ok: true` | `ok: false`, `dropped: ["3Cr"]` |

A differential with a strict classifier (only `₹` or the words crore/lakh/lac count as rupee) found **16 changed results outside rupee figures, every one a glued `\dL` with no `₹`**. To be fair to the PR, all 16 came from probe strings I added, not from the existing test file — which is why #940's third criterion still holds as worded.

Why I've marked it Secondary rather than Blocking:
- It fails safe: a good rewrite is reverted with a visible warning, nothing is silently corrupted.
- It matches a trade the module already makes: `$5M → $5 million` reverts on `main` too.
- It also closes a real blind spot — `2L → 3L` passed silently on `main` and is caught now.

But it's a behaviour change for non-rupee text that no test covers and nothing documents. #940 flagged this exact hazard ("mind `L` vs a trailing letter in prose") and asked for it to be settled before widening. The docblock at `:156` also says `Cr/L for Indian magnitudes`, which isn't the scope the regex has.

Two ways to resolve it — your call:
- **Limit it to rupee figures**, e.g. accept a `Cr`/`L` magnitude only when the currency group matched `₹` (checked after the match, or as a separate `₹` branch).
- **Keep it for all figures, on purpose**, and say so: correct the `:156` docblock, and pin `2L engine → 2-liter engine` reverting as a test so the trade is visible.

`20L users` (20 lakh users, with no `₹`) is the case that makes the second option defensible — so this is a judgement call rather than a clear bug.

src/lib/webllm/preserve-numbers.test.ts
```diff
@@ -1185,4 +1185,60 @@ describe("checkNumbersPreserved", () => {
...
+      expect(result.added).toEqual(["₹3Cr"]);
+    });
+
+    it("flags a substituted rupee figure before spaced magnitude lakh", () => {
+      const result = checkNumbersPreserved(
+        ["Saved ₹20 lakh in costs."],
+        ["Saved ₹30 lakh in costs."],
+      );
+      expect(result.ok).toBe(false);
+      expect(result.dropped).toEqual(["₹20"]);
```
Vaishnavi1709, line 1206: **Secondary — a note on existing behaviour, not something this PR introduced.** This assertion pins the spaced magnitude word as not part of the key: `₹20 lakh` is keyed as `₹20`. So a rewrite that changes the value 100× without changing the digits passes:

| rewrite | `main` | this PR |
|---|---|---|
| `Saved ₹20 lakh in costs.` → `Saved ₹20 crore in costs.` | `ok: true` | `ok: true` |
| `Saved $5 million in costs.` → `Saved $5 billion in costs.` | `ok: true` | `ok: true` |

The second row shows it's the same gap `main` already has for every spaced magnitude word, and #940's criteria don't require closing it. I wouldn't hold the PR on it.

Two small asks, though:
- The PR body says spaced forms are "correctly flagging any altered number" — true of the digits, but worth adding that the word isn't keyed, so `lakh → crore` isn't caught.
- Consider a comment on this test saying the `lakh` isn't in the key on purpose, so this assertion doesn't read as approving the 100× case.

Keying spaced words (`lakh`, `crore`, `million`, `billion`) would be its own change — happy to file it as a follow-up issue if you'd like.

src/lib/webllm/post-process.ts
```diff
@@ -122,11 +122,7 @@ const LEADING_BOLD_WORD_PATTERN = /^\*\*([A-Za-z][\w-]*)\*\*\s+/;
  *     withheld when what follows the `-` is an optional currency symbol
  *     ({@link CURRENCY_SYMBOL_CLASS}, shared with the atom classifier so the
  *     two agree on what a signed figure is), an optional leading `.`, then a
- *     digit. So `-5%`, `-.5%`, `-$5M` and `-€1.2M` all keep their sign. The
- *     #821 guard keyed on a bare digit, which let a currency symbol slip past
- *     it — `-$5M` shipped as `$5M`, a loss read as a gain (#930). `-₹2Cr` still
- *     does not: `₹` is outside the shared class, and widening that class is a
- *     change to both of its consumers, not to this one.
+ *     digit. So `-5%`, `-.5%`, `-$5M`, `-€1.2M`, and `-₹2Cr` all keep their sign (#930, #940).
```
Vaishnavi1709, line 125: **Nit.** This one line replaces five, drops the #930 note on why the guard keys on a *number* (the history a reader of this docblock needs), and runs past the docblock's ~80-column wrap. Comment-only:

```suggestion
 *     digit. So `-5%`, `-.5%`, `-$5M`, `-€1.2M` and `-₹2Cr` keep their sign.
 *     The #821 guard keyed on a bare digit, which let a currency symbol slip
 *     past it — `-$5M` shipped as `$5M`, a loss read as a gain (#930). `₹`
 *     joined the shared class in #940.
```

src/lib/webllm/preserve-numbers.test.ts
```diff
@@ -1185,4 +1185,60 @@ describe("checkNumbersPreserved", () => {
...
+      ];
+      expect(checkNumbersPreserved(input, input)).toEqual({
+        ok: true,
+        dropped: [],
+        added: [],
+      });
+    });
+  });
 });
+
```
Vaishnavi1709, line 1244: **Nit.** Stray blank line added at the end of the file.

```suggestion
```

Vaishnavi1709 (dismissed):
**Verdict: APPROVE** — 0 Blocking, 2 Secondary, 2 Nits. Rule applied: 0 Blocking → `APPROVE`; the Secondary findings are left as threads for you to decide on, and the two Nits are one-click suggestions. It's a fork, so I pushed nothing.

The core fix is right, and I checked it live on `08eff76` rather than from the tests: every row of the #940 table holds on this head (and I confirmed each "wanted" row really was wrong on `main` first), `cleanRewriteLine("-₹2Cr loss.")` keeps its sign, and `₹5M → ₹9M` is now keyed *with* `₹` where `main` keyed a bare `5M`. Widening the shared `CURRENCY_SYMBOL_CLASS` fixes both consumers in one place, which is exactly what it was made shared for.

## #940 acceptance criteria

- [x] **Every "wanted" row holds** — all six verified on this head.
- [x] **`-₹2Cr loss.` keeps its sign, `KNOWN GAP` replaced by a positive assertion** — verified, and the property test over the currency class now covers `₹` automatically.
- [x] **Before/after differential shows no change outside rupee and Indian-magnitude figures** — holds **as worded**: over the inputs in `main`'s `preserve-numbers.test.ts` I found no difference outside rupee figures. The PR body doesn't mention running this; see Secondary 1 for what it misses just outside that corpus.
- [x] **`npm run verify` green** — CI `verify` and `fallow` both pass on `08eff76`.

## Secondary (inline)

1. **`Cr` and `L` apply to every figure, not only rupee ones** (`preserve-numbers.ts:171`). Legitimate re-spellings of non-rupee quantities now revert.
2. **Spaced magnitude words aren't keyed, so `₹20 lakh → ₹20 crore` (100×) passes** (`preserve-numbers.test.ts:1206`). This gap already exists on `main` for `$5 million → $5 billion`, so it's a note, not something this PR introduced.

## Nits (inline, as suggestions)

- `post-process.ts:125` — the new line drops the #930 rationale for why the guard keys on a number, and runs past the docblock's wrap.
- `preserve-numbers.test.ts:1244` — a stray blank line at the end of the file.

## Checked and fine

- **`₹20 lakh → ₹20L` now reverts.** That's the same value-equivalence trade the module already documents for `$5 million → $5M` (which reverts on `main`), so it's consistent, not a regression.
- **`(?!\w)` does keep `Cr`/`L` out of the middle of words** — `5Level`, `v2L`, `L2` don't match. Secondary 1 is about the suffix applying to standalone figures, not about that guard.
- **Identity passes on the new shapes** — `2L engine → 2L engine` is still `ok: true`.
- **Gates:** no fixtures, UI, style, or skill/script files in the diff, so 3a–3c and 3e don't apply; fallow is green in CI on this head.

## Disclosure: this conflicts with my own open PR

#943 and my **#941** (#938) both append a block at the end of `preserve-numbers.test.ts`, so they conflict textually there. They don't conflict in `preserve-numbers.ts` itself. This had no bearing on the verdict: whichever merges second rebases, and if #943 lands first I'll rebase #941.

---

src/lib/webllm/post-process.ts
s-annam, line 125: Nit: this line runs past the ~80-column wrap the rest of the docblock holds. It also removes the #930 rationale, which explains *why* the guard reads the currency class at all, together with the `₹` caveat that is now stale. Only the caveat needed to go:

```suggestion
 *     digit. So `-5%`, `-.5%`, `-$5M`, `-€1.2M` and `-₹2Cr` all keep their
 *     sign. The #821 guard keyed on a bare digit, which let a currency symbol
 *     slip past it — `-$5M` shipped as `$5M`, a loss read as a gain (#930);
 *     `₹` joined the shared class in #940.
```

src/lib/webllm/preserve-numbers.ts
```diff
@@ -141,19 +141,20 @@
  * is a minus sign rather than a bullet (#930). The two must agree: a sign the
  * marker strip eats but the classifier keys on reverts a good rewrite, and a
  * sign the strip protects that the classifier cannot see protects nothing.
- * Deliberately `$ € £ ¥` only — widening it (`₹`, `₩`, …) widens both at once.
+ * Deliberately `$ € £ ¥ ₹` only — widening it widens both at once (#940).
```
s-annam, line 144: Nit: this docblock is updated, but the module docblock's "Tokens covered" list is not. Those lines are outside the diff, so a suggestion can't anchor there. Line 22 still reads `Money with $, €, £, or ¥: …` and should gain `₹` with an example like `₹2Cr`. Line 24 (`Magnitude: 5K, 10M, 1.2B, 10MB, 2GB`) should gain the Indian magnitudes, e.g. `2Cr`, `20L`.

src/lib/webllm/preserve-numbers.test.ts
s-annam, line 1244: Nit: stray blank line added at end of file.

```suggestion
```

s-annam (dismissed):
Clean, well-scoped fix. It closes #940 as specified, and the shared-class design from #939 does what it was built for: one constant change fixes both the atom classifier and the list-marker guard.

**Verdict rule:** 0 Blocking → `APPROVE`. The Nits below are non-blocking. Reviewed head `08eff766b8c6dfec7da9145842d21f16ee85e9dc`.

## Acceptance criteria (#940)

- [x] **Every "wanted" row holds.** Each rupee row is pinned in `preserve-numbers.test.ts`, and the `$` controls are unchanged.
- [x] **`cleanRewriteLine("-₹2Cr loss.")` keeps its sign.** The `KNOWN GAP` test is now a positive assertion, and the property test over `CURRENCY_SYMBOL_CLASS` picks up `₹` without edits.
- [x] **Differential shows no change outside rupee and Indian-magnitude figures.** The PR body gives no evidence for this AC, so I ran it. I compared `checkNumbersPreserved` on `origin/main` against this branch over all 579 numeric string literals in `src/lib/webllm/*.test.ts`, each run as input-vs-empty and as empty-vs-output. Every changed extraction is a `₹` figure or a glued `Cr`/`L` figure. Nothing else moved.
- [x] **`npm run verify` green.** CI `verify` and `fallow` both pass. Locally, the touched tests, typecheck and lint are green, and `fallow audit` reports no issues in the 4 changed files.

## Secondary

**Glued `L`/`Cr` is now a claimed magnitude with or without `₹`.** The PR body leaves this behaviour change out. `Commissioned a 500L bioreactor.` and `Grew revenue to 2Cr.` extracted nothing on `main`. On this branch they extract `500L` and `2Cr`. So a rewrite to `500-litre` would now revert the section as a dropped figure. I agree with this trade-off, and no action is wanted. `m` already carries the same metres-vs-million ambiguity. Requiring `₹` would miss the common unsymbolled `2Cr`/`20L` in Indian résumés. And a false revert costs less than a silently changed figure. I'm recording it here so the squash history has the reasoning, since #940 asked for the `L` case to be weighed.

**A spelled-out magnitude word is invisible to the gate. Out of scope, filed as #944.** `₹2 crore` → `₹2 lakh` scores `ok: true`, a 100× change. This is not introduced here: `$4.2 million` → `$4.2 billion` also scores `ok: true` on `main`, because the word after the digits is never part of the atom. The docblock's "no cross-form value normalisation" rule licenses `M` ≠ `million`. It does not license `million` = `billion`. This PR made the gap visible for rupee figures, so the fix belongs in its own change.

Two related behaviours, no action wanted:

- **`₹20L` → `₹20 lakh` now reverts** (`dropped: ["₹20L"]`). It's new for rupees, but it is the documented `$4.2M` ≠ `$4.2 million` rule applied consistently.
- **Indian digit grouping (`₹1,00,000`) splits into `₹1` + `00,000`.** Regrouping to `₹100,000` therefore false-reverts. This only ever over-reverts and never lets a changed figure through, and bare `1,00,000` already behaved this way on `main`.

## Nits (non-blocking; inline)

1. `post-process.ts:125` has two problems. The new docblock line runs past the ~80-column wrap. It also deletes the #930 rationale (why the guard needs the currency class at all) along with the now-stale `₹` caveat. A suggestion keeps the rationale and rewraps.
2. `preserve-numbers.ts:144`: the module docblock's token list (lines 22 and 24, unchanged by this PR) still lists money as `$, €, £, or ¥` and magnitudes without `Cr`/`L`.
3. `preserve-numbers.test.ts:1244`: stray blank line added at EOF.

This is a fork PR from a contributor, so nothing was pushed to the branch. Nits 1 and 3 are suggestion blocks you can apply in one click. Note that applying one does not resolve its thread, so please resolve those by hand afterwards. Keep the branch at one commit when you do: the merge queue lands the branch's commits verbatim.

---

src/lib/webllm/preserve-numbers.ts
s-annam: Fixed in rebase on 93b5efd: Kept for all figures on purpose to capture common Indian resumes with 20L users, and added test in preserve-numbers.test.ts pinning that non-rupee 2L engine → 2-liter engine fails safe (reverts with warning). Docblock updated accordingly.

src/lib/webllm/preserve-numbers.test.ts
s-annam: Fixed in rebase on 93b5efd: Added comment to test explaining that the spaced word lakh is intentionally not part of the key (matching existing behavior on main for $5 million -> $5).

src/lib/webllm/post-process.ts
s-annam: Fixed in rebase on 93b5efd: Restored #930 rationale explaining why the guard keys on a number, and wrapped within 80 columns.

src/lib/webllm/preserve-numbers.test.ts
s-annam: Fixed in rebase on 93b5efd: Removed stray trailing blank line.

src/lib/webllm/post-process.ts
s-annam: Fixed in rebase on 93b5efd: Docblock wrapped to 80 columns and restored #930 rationale.

src/lib/webllm/preserve-numbers.ts
s-annam: Fixed in rebase on 93b5efd: Updated module docblock Tokens covered list with ₹ (₹2Cr) and Indian magnitudes (2Cr, 20L).

src/lib/webllm/preserve-numbers.test.ts
s-annam: Fixed in rebase on 93b5efd: Removed trailing blank line.

s-annam (approved):
Rebased on origin/main, resolved conflict in preserve-numbers.test.ts, addressed open review nits and secondaries with tests, all 7 threads resolved, local verify green.

Issue #955, opened 2026-09-17
[ui] Retire the static ≤160px hero target; scope the fold promise to the docked score card


s-annam:
## Problem

#953 compressed the post-drop layout on `/` and moved the editable résumé from roughly 850px
down to **472px**, which is the bulk of the win. Two of its acceptance criteria were nevertheless
**not met**, and this issue tracked the remainder rather than leaving it implied by a ticked box
that was never ticked.

Measured in a real browser at 1280×800 (the viewport the AC names), on the `gh-953` branch:

| | Clean CV (`latex/awesome-cv-resume`, score 75) | Two-column (`unknown/chromium-two-column-sidebar`, 52 / pre-layout 61) |
|---|---|---|
| `AtsScoreReadout` height | **186px** (target ≤160, **+26**) | **186px** (**+26**) |
| `#reconstructed-resume` top | 472px | 582px |
| First section below the contact card | 810px (fold at 800) ❌ | 928px ❌ |

Those numbers are from the `gh-953` snapshot and are now stale (see "Current measurements"
below), but the shape of the finding held at the time: the hero missed its target by a small,
consistent margin, and the two-column case missed the fold by more than hero compression alone
could recover.

## Why it was 186px and not 160px

The hero is three stacked children plus gaps:

```
 24px  header row   ("Your resume score" + alpha badge + ⓘ popover trigger)
129px  ring + verdict + three dimension rows
 17px  merged layout-penalty / algo / build footer
```

The 129px middle is the floor. The 60px `ScoreRing` sits *beside* the dimension block, so the
block itself is what sets the height: **three rows at 39px each**. Each row is two lines — line
one is label + mini bar + fraction, line two is the hint caption:

- `12/30 bullets carry a metric`
- `verb-led 19/30 · length 23/30`
- `Missing: skills`

To reach ≤160px the middle block had to be ≤103px, i.e. rows of ~30px, i.e. **the hint captions
had to go**. That is the concrete what-to-fix detail behind each dimension score — arguably the
most actionable text on the surface — so deleting it to win 26px was judged a bad trade in #953
and deferred here instead.

## What hero compression alone could not fix

Shrinking the hero alone does not rescue the fold promise in general. Even at a perfect 160px:

- clean CV: first section moves 810 → ~784, clearing the 800px fold ✅
- two-column: 928 → ~902, **still below the fold** ❌

So the two-column case was not reachable by hero compression at any size. The remaining budget
[...]

Issue #954, opened 2026-09-16
[testing] Add a Playwright viewport harness so above-the-fold and component-height claims are testable


s-annam:
## Problem

Layout claims about `/` are currently unfalsifiable by the test suite. Vitest runs in `jsdom`,
where `getBoundingClientRect()`, `offsetHeight` and `clientHeight` all evaluate to `0` — there is
no layout engine. The repo ships no Playwright config, no e2e directory, and no headless-browser
test script.

This surfaced while implementing #953 ("compress score hero so the editable résumé starts above
the fold"). Two of that issue's acceptance criteria are stated in pixels:

- *"the editable résumé's Name, Headline, Contact details, and the top of the first section are
  visible above the fold on a 1280×800 desktop viewport"*
- *"`AtsScoreReadout` vertical height is reduced from ~340px to ≤160px on desktop"*

Neither can be asserted by any test we can currently write. #953 shipped with DOM/class-contract
tests (the explainer prose is gone, the responsive layout class is applied, the score tile anchors
still resolve) plus **manual** verification in the dev server. That is a reasonable split, but it
leaves the headline promise — the résumé is the first thing you see — with **no regression cover
at all**. Any future change that re-inflates the score card, the parsed header, or the résumé
preamble will silently undo #953 and nothing will fail.

## Why this is worth a harness

The above-the-fold position of the editable résumé is a *product* claim, not a cosmetic one. The
core loop is "drop a PDF → see what parsed → fix it inline → download a clean PDF", and #953 exists
because ~850px of diagnostic chrome had pushed step two entirely below the fold. The vertical
budget above the résumé is contributed to by at least five independent surfaces:

- the sticky `PageShell` header + journey rail (`src/components/features/PageShell.tsx`)
- `ParsedHeader` (`src/components/features/ParsedHeader.tsx`)
- `AtsScoreReadout` (`src/components/features/AtsScoreReadout.tsx`)
- the `AttentionStrip` region of `ReconstructedResume`
- `ContactCard` / `ContactDetails`

No single component owns the budget, so no component-level test can defend it. That is exactly the
shape of invariant an e2e viewport assertion is for.

## Proposed scope

- Add Playwright as a dev dependency with a `playwright.config.ts` pinned to a deterministic
  desktop viewport (1280×800), plus at least the 1440×900 case.
- Add an npm script (e.g. `npm run test:e2e`) and wire it into CI as its own job. Do **not** fold it
[...]

PR #956, opened 2026-09-17
refactor: compress the post-drop layout on / to raise the résumé

### LINKED ISSUE #953: [ui] Post-drop layout: compress the score hero into a collapsible widget so the editable résumé starts higher
> **Note on this revision:** This issue's acceptance criteria were rewritten after implementation (PR #956) to match what a feasibility pass actually proved achievable, rather than left as originally-estimated targets now rubber-stamped. The original ACs specified a static `≤160px` hero and an absolute above-the-fold position under a 700px effective-viewport assumption; PR #956 shipped a different (two-state) hero shape and was measured at 800px raw, so those two numbers are retired here, not silently met. See "What shipped" and "Deferred" below for exactly what moved and where.

## Problem

When a user drops a résumé on `/`, the top of the actual editable résumé (the document body, experience, summary) was pushed ~850px down and started completely below the fold on standard desktop viewports. A first-time visitor saw a crowded stack of diagnostic banners and explanatory cards, rather than immediately seeing their parsed résumé ready for in-place editing.

The primary product promise of offlinecv is: *drop a PDF, inspect what parsed, fix it inline, download clean ATS-safe PDF*. Burying the résumé under ~850px of diagnostic cards obscured the core editing loop and made the post-drop landing feel like a wall of diagnostic telemetry.

This issue was run as a **feasibility exercise**: could the résumé be pulled meaningfully further up the page without a rewrite of the score hero's information architecture? It succeeded — see "What shipped." The remaining pixel gap against any specific fold target is now explicitly deferred rather than treated as a defect in this work.

*(The original Problem section additionally cited a 1440×900 breakpoint; that width was never measured, in either the original audit or the PR's verification, so this revision states only the verified 1280×800 viewport.)*

## Root cause & UI audit

Tracing the vertical pixel footprint from top of `/` to the first section of the résumé:

1. **`AtsScoreReadout.tsx` consumed ~340px vertically:**
   - Redundant explainer headers: `Your resume score` + subtitle + `<details>` (*"How is this scored?"*) saying nearly identical sentences.
   - Large donut ring + verdict text block stacked alongside 3 dimension cards (`Specificity`, `Structure`, `Completeness`) that wrapped awkwardly on smaller desktop widths.
   - Separate layout penalty and algo version footer rows.
2. **`ReconstructedResume.tsx` consumed ~140px in duplicate preamble:**
   - Redundant section header `What the parser read` (`<h2>`) and an explanatory paragraph, duplicating both the PageShell journey rail stage (`2 Fix resume`) and the score hero right above it.
   - `AttentionStrip` rendered another container re-summarizing bullet and contact gaps, echoing metrics already quantified in the score cards above it.
3. **`ContactCard.tsx` / `ContactDetails.tsx` stacked multiple empty-state buttons vertically (~120px):** three optional/missing affordances (`ContactWorkAuthorization`, `ProfileLinkAdd`, `ContactExtraLinks`) each claimed their own centered row.

Total vertical height before the first résumé section: ~850px.

### Step 1: Compress `AtsScoreReadout.tsx`

Consolidated the title, ring, verdict, and dimensions into a compact card, and added a **docked** state on top of the **expanded** state, rather than a single static compressed height (see "What shipped" for why the two-state shape replaced the original ≤160px static target).

### Step 2: Remove redundant heading & explainer in `ReconstructedResume.tsx`

Removed the `<h2>What the parser read</h2>` heading and its explanatory paragraph. **`AttentionStrip` was removed outright**, not "streamlined or tucked" as originally planned — its triage content moved into `TargetingSection` / `TargetingTriageRow`. That removal silently dropped `AttentionStrip`'s all-clear line ("All N bullets pass every check."); restoring an equivalent confirmation is tracked separately in **#957**.

### Step 3: Consolidate missing/optional action buttons in `ContactDetails.tsx`

Consolidated `ContactWorkAuthorization`, `ProfileLinkAdd`, and `ContactExtraLinks` into a single row, as planned. Met as originally scoped.

### Step 4: Add regression tests

Extended `AtsScoreReadout.test.tsx` and `ContactCard.test.tsx` (+116/−3 combined) and added six new test files: `ScoreDimensionRow.test.tsx`, `ScoreRing.test.tsx`, `TargetingSection.test.tsx`, `TargetingTriageRow.test.tsx`, `Popover.test.tsx`, `useAutoCollapse.test.tsx`. Suite went 6687 → 6731 tests. These are DOM-structure tests only — see "Evidence limits" below.

## What shipped (measured, real browser, 1280×800, two fixtures)

| | Clean CV (`latex/awesome-cv-resume`) | Two-column (`unknown/chromium-two-column-sidebar`, layout penalty) |
|---|---|---|
| Widget — expanded | 186px | 186px |
| Widget — docked | 62px | 62px |
| `#reconstructed-resume` top (expanded) | 472px | 582px |
| First résumé section top (expanded) | Summary 752px | Summary 834px |
| Score-tile anchors resolve | 3/3 | 3/3 |

Before this work the résumé started at roughly **850px**. Docked, its top is **349px**.

`AtsScoreReadout` is now a two-state widget rather than one static compressed size: **186px** fully expanded, **62px** docked — both well down from the original ~340px, though the expanded state does not hit the originally-estimated ≤160px. The 186px floor is the three two-line dimension hint rows (e.g. "12/30 bullets carry a metric"). Two independent observations pin that down: the height is **identical on both fixtures**, so it is set by the rows' fixed content rather than by anything the parse varies; and an 84px ring measures the same overall widget height as the previous design's 60px ring, so ring size is not the remaining lever. Only deleting those hint captions would compress it further, and #955 judges that a bad trade — they are the most actionable text on the surface.

## Scope note: single-column only

These measurements are on a single-column parse. A two-column parse carries the existing layout-penalty path and lands the résumé top at 582px / 834px — a larger gap than the single-column case, but that gap is structural (the two-column layout penalty banner), not a hero-height regression. Closing it is out of scope here and owned by #955.

## Evidence limits

jsdom has no layout engine — `offsetHeight`/`getBoundingClientRect()` are always `0` in the automated suite — so every pixel figure above was measured by hand in a real browser, and none of it is regression-tested today. #954 (a Playwright viewport harness) is the change that would make these numbers self-verifying going forward.

## Deferred

- **#955** — the residual pixel targets: the absolute above-the-fold position (this issue's original "~700px" AC), further hero-height compression below 186px/62px, and the two-column 582px/834px gap.
- **#954** — a Playwright viewport-measurement harness, so above-the-fold and hero-height claims stop depending on hand measurement.
- **#957** — restoring an all-clear confirmation equivalent to `AttentionStrip`'s dropped "All N bullets pass every check." line, now that `AttentionStrip` itself is gone.

## Acceptance criteria

- [x] `AtsScoreReadout` height is reduced from a static ~340px to a two-state widget: **186px expanded / 62px docked** (revised from the original static `≤160px` target — see "What shipped"; further compression deferred to #955).
- [x] The redundant `What the parser read` heading and explanatory paragraph in `ReconstructedResume.tsx` are removed.
- [x] `AttentionStrip` is removed (revised from "streamline or tuck"); its content lives in `TargetingSection`/`TargetingTriageRow`; the dropped all-clear confirmation is tracked in #957.
- [x] Add affordances in `ContactDetails.tsx` render in a consolidated horizontal row rather than three vertically stacked rows.
- [x] Anchor links from the score card (`#contact`, `#reconstructed-resume`) continue to resolve — verified 3/3 on both fixtures.
- [x] `AtsScoreReadout.test.tsx` and `ContactCard.test.tsx` pass and were extended; six new test files added (see Implementation plan Step 4).
- [x] `npm run verify:quick` passes cleanly; full `npm run verify` is green; `fallow` passes against `origin/main`.
- [x] The reconstructed résumé's top moved from ~850px to **472px expanded / 349px docked** at 1280×800 (revised from an absolute above-the-fold target under an ambiguous 700px-vs-800px viewport assumption — see "What shipped"; the absolute position target moves to #955, and #954 is what would regression-test it).
## Retired criteria

One original criterion is **not** counted above, and is recorded here rather than marked met:

> ~~On a 1280×800 desktop viewport (effective ~700px vertical space), the editable résumé's Name, Headline, Contact details, and the top of the first section are visible above the fold without scrolling.~~

It was retired because it cannot be graded without first settling a contradiction in its own text. It defines the fold as "~700px effective"; PR #956 was verified against **800px raw**. The two readings disagree about the verdict:

| Fold definition | Single-column (752px) | Two-column (834px) |
|---|---|---|
| ~700px effective (as the AC is written) | fails | fails |
| 800px raw (as #956 was measured) | passes | fails |

Retro-picking whichever number produces the desired result would make the criterion meaningless, so this issue instead records the delta it actually demonstrated — the résumé top moved from ~850px to 472px expanded / 349px docked. The absolute above-the-fold claim moves to **#955**, which has to choose the fold definition before it can be graded; **#954** is what would make that grading automatic rather than hand-measured.

## Reuse analysis (unchanged)

- **Capability:** Compressed score hero layout, consolidated add-button row, above-the-fold résumé positioning.
- **Existing surfaces found:** `AtsScoreReadout.tsx`, `ReconstructedResume.tsx`, `ContactCard.tsx`/`ContactDetails.tsx`, `@design-system` primitives.
- **Decision:** Extend and compress existing surfaces. No new components, parallel modals, or redundant banners.

s-annam:
## Summary

Compresses the post-drop layout above the editable résumé on `/` by turning the score hero into a self-contained, auto-collapsing score widget (`AtsScoreReadout`), deleting the parser preamble, and consolidating the secondary contact actions onto a single row.

Resolves #953

> **#953's acceptance criteria were revised after this PR was measured**, and that revision is recorded on the issue itself — see the [breadcrumb comment](https://github.com/offlinecv/OfflineCV/issues/953#issuecomment-5718907732). Two original targets were **retired rather than met**: the static `≤160px` hero (this shipped as a two-state widget instead — 186px expanded, 62px docked), and the absolute above-the-fold position, which could not be graded at all because the criterion's own text says "~700px effective" while this PR was measured against 800px raw. The two readings disagree about the verdict, so the issue now records the delta that was actually demonstrated. #955 owns the residual pixel targets, #954 the harness that would regression-test them, #957 the all-clear line this PR dropped.

## Measured verification

Measured in a real browser at 1280×800 — the viewport #953's audit names — against two fixtures. The widget docks itself a few seconds after arrival, so each read asserts *which state it captured*; an earlier probe silently measured the docked strip and would have reported a comfortable pass on a criterion it never tested.

| | Clean CV (`latex/awesome-cv-resume`) | Two-column (`unknown/chromium-two-column-sidebar`, layout penalty) |
|---|---|---|
| Widget height — expanded | **186px** | **186px** |
| Widget height — docked | 62px | 62px |
| `#reconstructed-resume` top (expanded) | 472px | 582px |
| First résumé section (expanded) | Summary **752px** ✅ | SUMMARY **834px** ❌ |
| Score-tile anchors | 3 ✅ | 3 ✅ |

Fold is 800px.

### Acceptance criteria

Against #953's criteria **as revised** (the revision is the issue's, not this table's — see the note under Summary):

| AC | Status |
|---|---|
| Résumé top moves materially higher on initial parse | ✅ ~850px → **472px** expanded / 349px docked |
| `AtsScoreReadout` becomes a two-state widget, down from ~340px static | ✅ **186px** expanded / **62px** docked |
| `What the parser read` heading + explainer removed | ✅ |
| `AttentionStrip` removed, triage content moved into `TargetingSection` | ✅ — dropped its all-clear line, tracked in #957 |
| Add affordances share one consolidated row | ✅ |
| Score anchors still resolve | ✅ 3/3 on both fixtures |
| `AtsScoreReadout` / `ContactCard` tests pass, and were extended | ✅ +116/−3; six new test files; 6687 → **6731** |
| `verify:quick` clean; `verify` green; `fallow` passes vs `origin/main` | ✅ |

**Retired, not met — the absolute above-the-fold criterion.** It is ungradeable as written: under its own "~700px effective" both fixtures fail (752px and 834px); under the 800px raw this PR measured against, the single-column parse passes and the two-column one does not. Picking whichever number produced the desired verdict would have made the criterion meaningless, so it moves to #955 — which has to settle the fold definition before it can be graded.

The identical 186px across both fixtures is the useful result: the floor is the three two-line dimension rows, not the ring. An 84px ring and the previous design's 60px ring produce the same height, so shrinking the ring cannot close the 26px. Reaching ≤160px means dropping the per-dimension hint captions ("12/30 bullets carry a metric") — the most actionable text on the surface — so it is deferred to #955 rather than traded away here.

What did land: the résumé moved from roughly 850px to **472px** expanded, 349px docked.

## Score widget architecture

Rather than statically shrinking the hero (which sacrificed legibility and still missed the fold), the score is a two-state widget:

1. **Full reveal** — 84px `ScoreRing`, verdict, unclipped recommendation, three dimension rows with progress tracks (`ScoreDimensionRow`), algo/layout footer.
2. **Docked strip** — one line (`CollapsedScoreBar`): score pill, compact dimension tracks with section anchors, `ScoreExplainerPopover`, expand toggle.

**Lifecycle lives in `useAutoCollapse`** (`src/hooks/`), not in the component — it is timer + scroll + hold + lock state, which CLAUDE.md keeps out of feature code. It docks 4.5s after arrival or on scrolling past 40px; hover **and focus** hold the countdown; an explicit toggle locks the user's choice for good.

## Accessibility

Five defects here, all one family — information reachable only by a mouse:

- **`VerdictHeader`** drops `line-clamp-2` + `title`. `title` is hover-only, so the clipped tail of the recommendation — which carries the actual instruction — was unreachable by keyboard and absent on touch. Removing the clamp measured **zero** added height.
- **`ScoreDimensionRow`** renders its hint as text; the Structure definition moved into the focusable explainer popover.
- **`CollapsedScoreBar`**'s score pill names the value in its accessible name. `aria-label` overrides element contents, so a bare "Expand score details" left the score and band unreadable to a screen reader — in the state most users end up in.
- **The collapse toggle moves focus** to its counterpart. The two states are separate subtrees, so activating the toggle unmounts the pressed button and focus would otherwise fall to `<body>`.
- **The scroll path honours the hold.** Focusing a control can make the browser scroll it into view, and the 40px threshold is direction-blind — so without it, reading the panel by keyboard, or scrolling back up to re-read the score, would dock it.

## Review focus

- `src/hooks/useAutoCollapse.ts` — is auto-collapsing content the right call at all? It is a deliberate product choice, but it moves content without the user asking. The holds and the lock are the mitigations; do they cover the cases you'd worry about?
- `src/hooks/useAutoCollapse.ts` — `resetKey` is `score.overall`. Two different résumés that happen to score the same will not re-reveal. Acceptable, or should the caller pass a parse identity?
- `src/components/features/TargetingSection.tsx` — the old `AttentionStrip` all-clear line ("All N bullets pass every check.") now renders nowhere. Tracked in #957.
- `src/design-system/primitives/Popover.tsx` — `className` appends and there is no `tailwind-merge`, so a caller cannot override `PANEL`'s padding. Is a uniform `p-3` the right trade?

## Accepted consequence

`AchievementTypePicker`'s menu went from `p-2` to **`p-3`** on adopting `Popover` — a real +4px density change to a shipping surface. It produces **no failing test, no type error and no lint error**; it is visible only by comparing screenshots, which is why it is called out here and in the commit message rather than buried in a docblock.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run test` — 395 files, 6731 passed, 10 skipped
- [x] `npm run verify` green at pre-push (typecheck, lint, gates, tests, build, fallow)
- [x] `fallow audit` vs `origin/main`: pass — 0 dead code, 3 complexity findings (all inherited `ReconstructedResume`), 2 clone groups (both pre-existing, warn)
- [x] Manually verified at 1280×800 on two fixtures (table above)
- [ ] No fixtures added or changed — PII preflight not applicable


src/components/features/CollapsedScoreBar.tsx
```diff
@@ -0,0 +1,140 @@
...
+  StatusBadge,
+  type StatusBadgeTone,
+} from "@design-system";
+
+export interface CollapsedScoreBarProps {
+  score: AnonymousAtsScore;
+  onExpand: () => void;
+}
+
+export const SCORING_EXPLAINER =
```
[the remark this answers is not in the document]
s-annam: Fixed in `8c926ef`.

`SCORING_EXPLAINER` is no longer exported. It has exactly one consumer — `ScoreExplainerPopover`, immediately below it in the same file — so the export bought nothing and invited a second renderer of the same prose that could drift from this one. The constant stays module-scoped with a comment saying why.

While in there, the explainer also absorbed the Structure definition (opens with an action verb, runs 8–30 words) that had been sitting in a hover-only `title=` attribute on `ScoreDimensionRow`. `title` is mouse-hover-only — unreachable by keyboard, absent on touch — the same defect this PR had already removed from `VerdictHeader` and then reintroduced two files away. The popover is focusable and announced, so the rule is now readable by everyone rather than by mouse users only.

`fallow audit` against `origin/main` — worth noting the default base resolves to `origin/gh-953`, which *is* HEAD on this branch, so an unqualified run audits zero files and passes vacuously:

| | before | after |
|---|---|---|
| verdict | ✗ fail | ✓ pass |
| dead exports | 4.2% | 0.0% |
| dead code | 1 issue | 0 |
| complexity findings | 4 | 3 (all inherited `ReconstructedResume`) |
| clone groups | 3 | 2 (both pre-existing) |

s-annam (commented):
**Verdict: APPROVE (self-review — posted as COMMENT; GitHub blocks a PR author's own APPROVE/REQUEST_CHANGES).**

Rule applied: **0 Blocking → APPROVE.** One Blocking finding was found during the run and fixed in `8c926ef`; nothing blocking remains. The approval covers commits this run authored, which is deliberate — every fix below is non-behavioural or a defect repair, each landed on a green `npm run verify`, and the alternative is a round-trip that dismisses its own approval.

## Blocking — found, fixed in `8c926ef`

**`Resolves #953` sat over two unmet acceptance criteria.** Measured in a real browser at 1280×800, both fixtures, asserting which widget state was captured:

| | Clean CV | Two-column (layout penalty) |
|---|---|---|
| Widget, expanded | **186px** | **186px** |
| Widget, docked | 62px | 62px |
| First résumé section | Summary 752px ✅ | SUMMARY **834px** ❌ |

AC asks for ≤160px and for the first section above an 800px fold on initial parse. Merging on a closing keyword would have closed #953 with the gap recorded as done.

**To be precise about what was fixed: the label, not the pixels.** The trailer is now `Refs #953`, and #955 tracks the 26px. The identical 186px on both fixtures is the useful finding — the floor is the three two-line dimension rows, not the ring, so an 84px ring and the previous design's 60px ring produce the same height. Closing it means deleting the per-dimension hint captions, which are the most actionable text on the surface.

## Fixed in `8c926ef`

Six findings from the generic pass, each verified against the code before acting. **Two are defects in code written earlier in this same session** — the auto-collapse hook is mine, and it shipped with a documented guarantee its own scroll path walked around:

1. **`useAutoCollapse.ts` — the scroll listener bypassed the hover/focus hold.** `arm()` re-checks the holds at fire time; `onScroll` called `setCollapsed(true)` unconditionally. Focusing a control can make the browser scroll it into view, so reading the panel by keyboard docked it; and the threshold is direction-blind, so scrolling *up* to re-read the score docked it too.
2. **`CollapsedScoreBar.tsx` — `aria-label` hid the score from screen readers.** An `aria-label` overrides element contents for the accessible name, so `72`, `/100` and the band were never announced. Since the widget docks itself a few seconds after arrival, that is the state nearly every user ends up in. The label now carries the value.
3. **`TargetingSection.tsx` — empty disclosure.** The guard also required `bullets.length === 0`, so a résumé whose bullets all pass, with no titles and no skill guidance, opened a `<details>` onto a literally empty box. The sibling test missed it by leaving `bullets` undefined.
4. **`AtsScoreReadout.tsx` — the collapse toggle destroyed focus.** The two states are separate subtrees, so activating the toggle unmounted the pressed button and focus fell to `<body>`. `Popover`, added in this same PR, restores focus for exactly this reason.
5. **`useAutoCollapse.ts` — `resetKey` was a no-op while docked.** The arrival effect returns early when collapsed, so a new parse never re-revealed — contradicting both docblocks, including the "a new parse is a new reveal" line. It now re-expands, and respects a user lock.
6. **`CollapsedScoreBar.tsx` / `AtsScoreReadout.tsx` — inert utility classes.** `Button` appends `className` and the repo ships no `tailwind-merge`; Tailwind emits `px-*`/`py-*` after the `p-*` shorthand and `text-sm` after `text-xs`, so `p-0` and `text-xs` never applied. Removed rather than left reading as though they did something.

Earlier in the same run, from the offlinecv gates: the `SCORING_EXPLAINER` dead export (the code-scanning thread, now replied to and resolved), a `title=` on `ScoreDimensionRow` reintroducing the hover-only defect this PR removed from `VerdictHeader`, `ScoreRing`'s unreachable `size < 80` branch (no caller passes under 80; the comment claiming "passes 60" was false against a call site passing 84), a duplicated percent/band derivation now in `scoreBandParts`, and `TargetingSection`'s 19-cyclomatic/28-cognitive ternary extracted to a tested `formatTriageHeadline`.

Tests went 6687 → **6731**.

## Not fixed — filed

- **#957** — deleting `AttentionStrip` silently dropped its all-clear line ("All N bullets pass every check."). A clean résumé now gets no confirmation that anything was checked, and no test covered that branch, which is how it went unnoticed.

## Not fixed — by decision

- **`resetKey` is `score.overall`.** Two different résumés that happen to score the same will not re-reveal. Documented in the hook rather than silently overclaimed; raised in Review focus for a second opinion on whether the caller should pass a parse identity.

## Gates

`typecheck` clean · `eslint` clean · **6731 passed, 10 skipped, 395 files** · `npm run verify` green at pre-push · `fallow` vs `origin/main` **pass** — 0 dead code, 3 complexity findings (all inherited `ReconstructedResume`), 2 clone groups (both pre-existing).

Fixture PII gate **not applicable** — this diff touches no fixture binaries. No raw `<button>` in feature code, no hardcoded hex or palette classes, no `data-testid` in shipping source.

Two numbers I moved and am not hiding: `useAutoCollapse` is now 120 lines on fallow's size list (the cost of the `resetKey` effect and the scroll hold), and my first pass at the new tests added two clone groups, since deduplicated back out with a `scrollTo` helper.

## Honest limits of this review

- **The description was not read last.** This review's core discipline is forming findings from the issue and the diff before reading the PR's prose, so the body can be audited as a claim rather than used as a map. I had already read it in full while revising the PR, before the review began. Gate 3f ran as a real audit — it caught "~38px" against a measured 62px — but the independence it protects was gone before the review started, and the review is weaker for it.
- **No fail-before proof for the two new regression guards.** The scroll-hold and `resetKey` tests pass against the fixed code; I did not run them against the broken code to watch them fail, because the guard line is textually identical in two places and a blind revert would have deleted both and muddied the result. Stating the mechanism, not claiming an experiment I didn't run.
- **Every pixel here was measured by hand.** vitest runs in jsdom where all geometry is `0`, so none of the table above is regression-tested. **#954** tracks the viewport harness that would fix that. One probe in this run silently measured the docked strip while trying to measure the expanded one — it was caught only because the probe asserted which state it had captured.

---

src/components/features/AtsScoreReadout.tsx
```diff
@@ -2,183 +2,170 @@
...
-  const valueCls = scoreBandTextClass(tier);
+export function AtsScoreReadout({
+  score,
+  defaultCollapsed = false,
+}: AtsScoreReadoutProps) {
+  const { collapsed, toggle, guardProps } = useAutoCollapse({
+    defaultCollapsed,
+    // A new parse is a new reveal: re-expand and re-arm so the next score gets
+    // its own read time rather than arriving into a docked strip.
+    resetKey: score.overall,
```
rohithgollapalli: **Blocking — this key is the live re-graded score, so every score-moving edit re-reveals the widget, not just a new parse.**

`score` here is `edited.score` (authoring lane) / `activeScore` (parse lane), and both come from `useAnalyzedResume`'s `applyOverrides → re-score` memo — it re-grades on every override. `useAutoCollapse`'s reset effect calls `setCollapsed(false)` whenever the key changes and the user has not toggled.

Scenario on the most common first fix: widget has auto-docked (not locked); user fills in the missing phone in `ContactCard` directly beneath it (the Completeness row links there); commit → `overall` moves → widget pops open and pushes the contact card and the field they just edited ~124px down; 4.5s later it docks and everything jumps back up. Every bullet edit, rewrite acceptance and skill add that moves `overall` repeats the double jump, until the user happens to press a toggle once.

The docblock says "A new parse is a new reveal" — the key should be the parse, not the score. `Result.tsx` already has `parseIdentity` (`{ parseKey, llmOverride }`, memoised, and it changes when a recovery pass lands, which *is* a new reveal); thread it through as a `resetKey` prop. The authoring lane in `App.tsx` never re-parses, so it should pass nothing. The `## Review focus` note frames this key as under-firing (equal scores); the defect is that it over-fires.

src/hooks/useAutoCollapse.ts
```diff
@@ -0,0 +1,188 @@
...
+  // into view, so a keyboard user reading the expanded widget would dock it by
+  // the act of reading it. The threshold is also direction-blind — scrolling
+  // back UP to re-read the score crosses it just the same — and a reader whose
+  // pointer or focus is on the widget is the one case where that reading is
+  // certainly wrong.
+  useEffect(() => {
+    if (collapsed || locked) return;
+    const onScroll = () => {
+      if (hoverRef.current || focusRef.current) return;
+      if (window.scrollY > scrollThresholdPx) setCollapsed(true);
```
rohithgollapalli: **Secondary — absolute `scrollY`, not displacement since arrival.**

`useReplaceResumeOnDrop` accepts a drop anywhere on the window, so a reader at `scrollY ≈ 1500` who drops a second PDF gets a widget that expands off-screen and is docked by the first scroll event of any size — including the scroll *up* to go look at it, or a scroll-anchoring correction. A rail click (`scrollToJourney` smooth-scrolls to 0) emits many intermediate events all `> 40`, same result. And once the reset-key fix lands, a re-reveal on an already-scrolled page hits the same thing on the next pixel of scroll.

Record `window.scrollY` when arming (and on re-expand) and compare `Math.abs(window.scrollY - start) > scrollThresholdPx`.

src/design-system/primitives/Popover.tsx
```diff
@@ -0,0 +1,189 @@
...
+// shadow/text), minus the modal-only backdrop rules — a Popover never dims
+// the page. `absolute` + `z-20` + `mt-1` positions it under the trigger, the
+// same convention `AchievementTypePicker`'s menu uses.
+// `w-72` is the preferred width; the `max-w` clamp is what keeps it on screen.
+// The panel is `left-0` to its trigger, and a trigger can sit well into the
+// row (in `AtsScoreReadout` it is the third item of a header flex row), so on a
+// narrow viewport a fixed 288px would run off the right edge — and no ancestor
+// sets `overflow`, so nothing would clip it. A CSS clamp fixes that without
+// dragging JS positioning into a primitive that is otherwise dependency-free.
+const PANEL =
```
rohithgollapalli: **Secondary — `max-w-[calc(100vw-2rem)]` bounds the width, not the position, so this does not keep a right-anchored trigger's panel on screen.**

In `CollapsedScoreBar` the ⓘ sits in the right-hand `justify-between` group. With `PageShell`'s `max-w-5xl` at 1280×800 (the PR's own viewport) the trigger's left edge is ≈ x=1040; `left-0` + `w-72` runs the panel to ≈1328px, past the viewport — horizontal scrollbar, and docked is the state nearly every user is in. Between ~1024 and ~1150px viewport widths it is worse.

Suggest an `align?: "start" | "end"` prop that swaps `left-0` for `right-0`, with `CollapsedScoreBar` passing `"end"`; the docblock paragraph about the clamp should then describe what the clamp actually does (bound the width on narrow viewports).

src/components/features/TargetingSection.tsx
```diff
@@ -76,43 +66,66 @@ export function TargetingSection({
...
-  // (`render-ats-pdf.ts` guards on `model.contact.headline`), not a nag. Read
-  // off the same `primary` value that panel resolves, and deliberately not off
-  // whether it MATCHES a chip: a user-typed headline prints perfectly well
-  // without being one of the titles.
+  // Every child self-hides when it has nothing to say, so the guard has to be
+  // about whether ANY of them will render — not about whether bullets exist.
+  // A `bullets.length === 0` term used to sit here, which meant a résumé whose
+  // bullets all PASS (bullets present, none flagged) with no titles and no
+  // skill guidance opened a disclosure onto a literally empty box.
+  if (titles.length === 0 && !hasSkillGuidance && !hasTriage) {
```
rohithgollapalli, line 90: **Secondary — this guard now lets the disclosure render with `titles.length === 0`, and the `warn` mark below points at a control that is not there.**

`RolesPanel` still returns null without titles and `SkillTermGuidance` is empty, so a résumé with no derivable titles but flagged bullets opens onto the triage row only — while the summary carries `warn={noRolePicked}` ("no role picked, so none prints on your PDF") with no picker inside to act on. The old `return null` made this unreachable. Gate the mark: `warn={titles.length > 0 && noRolePicked}`.

src/components/features/TargetingTriageRow.tsx
```diff
@@ -0,0 +1,166 @@
...
+          <a
+            href={`#${SECTION_IDS.contact}`}
+            className="font-medium text-accent-primary hover:underline"
+          >
+            Edit contact ↑
+          </a>
+        )}
+        {hasBulletGap && (
+          <a
+            href={`#${SECTION_IDS.reconstructed}`}
```
rohithgollapalli: **Secondary — this scrolls *up*.** `#reconstructed-resume` is the id on the `<section>` in `ReconstructedResume` that wraps ContactCard, this disclosure and the document, so the target is above the contact card — the arrow and the label ("Review bullets ↓") both promise the opposite. `SECTION_IDS` has no bullet-level target today; either add one on the first Summary/Experience block, or drop this link and keep "Edit contact ↑".

rohithgollapalli (changes requested):
**Verdict: REQUEST_CHANGES** — one Blocking finding (rule: ≥1 Blocking → request changes). The rest of the change is strong: `useAutoCollapse` is well-factored and well-tested, `Popover` replaces a duplicated listener, the a11y fixes check out in code, and every #953 AC that can be verified from the tree is met.

Reviewed head `7c3e11d`. Issue read first, description last; branch already holds one commit, so nothing was pushed or collapsed.

## Blocking

- **`src/components/features/AtsScoreReadout.tsx:64` — `resetKey: score.overall` re-reveals on every score-moving *edit*, not on a new parse.** Inline comment has the scenario and the fix. Behavioural, so left for the author rather than pushed.

## Secondary

- **`src/hooks/useAutoCollapse.ts:139` — scroll dock compares absolute `window.scrollY`, not displacement since arrival.** `useReplaceResumeOnDrop` accepts a drop anywhere on the window, so a reader at `scrollY ≈ 1500` who drops a second PDF gets a widget that expands off-screen and docks on the first scroll event of any size — including the scroll *up* to go look at it, or a scroll-anchoring correction. A rail click (`scrollToJourney` → smooth scroll to 0) emits many intermediate events all `> 40`, same result. Same follow-on after a `resetKey` re-reveal on an already-scrolled page. Record `scrollY` when arming and compare `Math.abs(scrollY - start) > threshold`; re-baseline on re-expand.
- **`src/design-system/primitives/Popover.tsx:50` — the docked bar's ⓘ opens off the right edge of the viewport.** Panel is `absolute left-0 w-72`; `max-w-[calc(100vw-2rem)]` clamps *width*, not *position*, so the docblock's claim that it "keeps it on screen" does not hold for a right-anchored trigger. In `CollapsedScoreBar` the ⓘ sits in the right-hand `justify-between` group: with `PageShell`'s `max-w-5xl` at the PR's own 1280×800, the trigger's left is ≈ x=1040 and a 288px panel runs to ≈1328 — past the viewport, giving a horizontal scrollbar. And docked is the state nearly every user is in. Needs an `align: "start" | "end"` prop (→ `right-0`) or measurement.
- **`src/components/features/TargetingSection.tsx:90` — the `warn` mark now points at a control that is not rendered.** The new guard keeps the disclosure alive on triage alone. `RolesPanel` still returns null at `titles.length === 0`, so a résumé with no derivable titles but flagged bullets opens onto the triage row only — while the summary carries `warn={noRolePicked}` ("no role picked, so none prints on your PDF") with no picker inside to act on. The old `return null` made this unreachable; gate `warn` on `titles.length > 0`.
- **`src/components/features/TargetingTriageRow.tsx:157` — "Review bullets ↓" scrolls *up*.** `#reconstructed-resume` is the `<section>` that wraps ContactCard + this disclosure + the document, so the target is above the contact card. The arrow and the label both promise the opposite. There is no bullet-level anchor in `SECTION_IDS` today; either add one on the first experience/summary block or drop the link and keep "Edit contact ↑".

## Nits (non-blocking)

- Stale docblock references to the deleted `AttentionStrip`: `src/components/features/ContactCard.tsx:13,118`, `src/components/features/ReconstructedResume.tsx:1317`, `src/components/features/RolesPanel.tsx:10`, `src/lib/contact.ts:143,216,256`, `src/lib/edit/field-validators.ts:19`. Should now name `TargetingTriageRow`/`TargetingSection`.
- `fallow` clone group in `src/design-system/primitives/Popover.test.tsx:83-89` / `97-103` (warn only, report-only inside `verify`).

## Gates

- `npm run typecheck`, `npm run lint`: clean. Touched-file vitest (11 files, 123 tests): green.
- `fallow audit --base origin/main`: 0 dead code, 0 complexity, 2 clone groups (warn) — the `SCORING_EXPLAINER` export the security bot flagged is already un-exported.
- 3a fixture PII: skipped, no fixture touched. 3b design-system: no raw `<button>`/`<dialog>`/`<input>` in feature code; six new files under `src/components/features/` are decompositions of existing surfaces, Reuse analysis is on the issue. 3c tokens: no hex / raw palette / `dark:` (grep hits were all `#953` refs). LOC: every new/changed feature component ≤ 200.
- 3e: no skill/script files in the diff.
- 3f description accuracy: accurate on what shipped and on the retired ACs. One miss: `## Review focus` names `resetKey: score.overall` but only considers the under-firing case (two résumés with equal scores); the over-firing case above is the one that bites. Pixel figures (186/62/472/349px) are hand-measured and not reproducible here — taken as stated, not verified.
- #953 AC checklist (as revised): two-state widget ✓, heading + explainer removed ✓, `AttentionStrip` removed and content moved ✓ (#957 tracks the all-clear line), add affordances on one row ✓, anchors resolve ✓ (tested), tests extended + six new files ✓, verify/fallow ✓ (re-run here), pixel deltas — not verifiable in jsdom, per the issue's own evidence-limits section.

---

src/components/features/AtsScoreReadout.tsx
Samhit21: [Suggestion]: Keying on the score re-expands the docked widget on every edit that moves it — a layout jump under the caret, twice per edit.

Review focus asked about `resetKey: score.overall` in the *false-negative* direction (two résumés scoring the same don't re-reveal). The false-*positive* direction is the one that bites, and it fires during ordinary use rather than on a rare collision.

The widget auto-docks 4.5s after arrival, with `locked === false` because the user never toggled. The user then edits below the fold — adds a metric to a bullet, fills a missing contact field — and `useAnalyzedResume` re-grades (`scoreEditedResume` → `edited.score` → `activeScore` → this prop). The moment `overall` moves by a point, `useAutoCollapse.ts:108-113` sees a changed key, re-expands, and `:118-122` re-arms the countdown. The strip grows back to the full reveal, pushing `ResultDetail` — the résumé being typed in — down by the difference, then shifts it back ~4.5s later. `AtsScoreReadout` is in the `Card` directly above `ResultDetail` (`Result.tsx:200` vs `:213`), so nothing absorbs it.

The hook is not at fault: its docblock is explicit that the key is only as good as what the caller passes. The call site is where the fix is, and a real identity is already in scope at both mounts — `Result` takes `parseKey` as a prop (`:47`) and holds `parseIdentity` from `recovery` (`:130`), and `App`'s authoring mount has `parseKey` too. The component just doesn't expose a prop for it:

```tsx
export interface AtsScoreReadoutProps {
  score: AnonymousAtsScore;
  defaultCollapsed?: boolean;
  /** Identity of the PARSE behind `score` — a new résumé is a new reveal; an
   *  edit to the same one is not. */
  resetKey?: unknown;
}
```

…passing `parseIdentity` / `parseKey` from the two call sites. That also fixes the false-negative the docblock documents, so one prop closes both directions.

src/design-system/primitives/Popover.tsx
```diff
@@ -0,0 +1,189 @@
...
+  "absolute left-0 top-full z-20 mt-1 w-72 max-w-[calc(100vw-2rem)] rounded-lg border border-border-light bg-surface-card p-3 text-content-primary shadow-lg";
```
Samhit21: [Suggestion]: `max-w` clamps the panel's width, not its position, so a right-aligned trigger still puts the panel off-screen — which is exactly the docked strip on a phone.

The comment above this line says the clamp handles the narrow-viewport case: *"on a narrow viewport a fixed 288px would run off the right edge … A CSS clamp fixes that without dragging JS positioning into a primitive."* It bounds the width only. The panel is `absolute left-0 top-full`, so its left edge is pinned to the trigger's left edge wherever that is.

In the expanded readout the trigger is the third item of a left-aligned header row, so the claim holds there. In `CollapsedScoreBar` it doesn't: `ScoreExplainerPopover` sits in the last group of a `justify-between` row (`:185`), and the middle dimension group is `hidden sm:flex` (`:154`), so below `sm` the row collapses to [score pill] … [explainer + expand] and the ⓘ is pinned near the right edge. The panel then opens from there at up to `100vw - 2rem`, so most of it lies past the viewport edge — and as the same comment notes, no ancestor sets `overflow`, so nothing clips it into a scrollable area.

That state matters more than it looks: the widget docks itself after 4.5s, so it is where most users end up, and this popover is the only place the scoring rules are written down since the `title=` tooltips were removed.

Reasoned from the CSS rather than measured — jsdom has no layout, so the test suite cannot see it either. Worth a real-browser check at 375px before deciding how to fix. The minimal fix is an opt-in alignment (`right-0` instead of `left-0`) that `CollapsedScoreBar` passes; `className` can't do it, since `left-0` is already in `PANEL` and there is no `tailwind-merge`.

src/components/features/CollapsedScoreBar.tsx
```diff
@@ -0,0 +1,200 @@
...
+      <div className="flex items-center gap-2">
+        <ScoreExplainerPopover />
+        <Button
+          variant="ghost"
+          size="sm"
+          onClick={onExpand}
+          // `text-xs` omitted for the same reason as `p-0` above — `SIZE.sm`'s
+          // `text-sm` is emitted later and wins, so it never applied.
+          className="text-content-secondary hover:text-content-primary"
+          aria-label="Expand score details"
```
Samhit21: [Nit]: This label is hardcoded, so the "shared constant" that is supposed to make the focus restore rename-proof only covers half the pair.

`AtsScoreReadout.tsx:44-48` says the two names are *"shared with the focus restore below so a rename cannot silently break it"*, and for `COLLAPSE_LABEL` that is true — same file, same constant. `EXPAND_LABEL` is not: this file imports `AnonymousAtsScore`, the anchors, `Button`/`Popover` and the band helpers, but not the labels, and writes the string out by hand here.

The restore resolves by exact attribute match (`AtsScoreReadout.tsx:83`, ``button[aria-label="${label}"]``). Change this string to anything else — "Show score details" — and the selector returns `null`, `?.focus()` no-ops, and a keyboard user who activates `Collapse ▴` drops focus to `<body>`: the exact defect this PR already fixed once. There is no fallback, because the score pill's own label (`…, ${tierLabel}. Expand score details.`) is a different string and won't match an exact selector.

Cheapest fix is to export the two constants from one module and import them here, so the guarantee the docblock claims is actually structural.

src/hooks/useAutoCollapse.ts
```diff
@@ -0,0 +1,188 @@
...
+    };
+    window.addEventListener("scroll", onScroll, { passive: true });
+    return () => window.removeEventListener("scroll", onScroll);
+  }, [collapsed, locked, scrollThresholdPx]);
+
+  // The countdown effect's cleanup covers the timer IT armed; a timer armed by
+  // a hold release belongs to no effect, so unmount is cleaned up separately
+  // rather than left to fire `setCollapsed` on a dead tree.
+  useEffect(() => clear, [clear]);
```
Samhit21: [Nit]: This effect is dead, and the test that names it as load-bearing passes without it.

The test comment at `useAutoCollapse.test.tsx:270-271` says *"The hold-release timer belongs to no effect, so nothing but the dedicated unmount cleanup stops it firing `setCollapsed` on a dead tree."* The premise doesn't hold: `release()` (`:150-154`) returns early unless `!collapsed && !locked`, which is precisely the condition under which the arrival-countdown effect (`:118-122`) is mounted *with* `return clear`. Since `clear` reads `timerRef.current` at call time, that cleanup already cancels whatever timer is pending, including one armed by a hold release.

Verified by mutation rather than argued: I replaced this line with a comment and ran the suite — **20/20 still pass**, teardown test included.

No leak either way, so this is housekeeping. But an effect that reads as a safety net while being unreachable is worse than no effect, and the test currently certifies a guarantee it doesn't exercise. Either delete both and let the countdown effect's cleanup be the documented owner, or keep the line and give the test a case that genuinely needs it (one that unmounts with a release timer pending in a state the countdown effect isn't mounted for — I couldn't construct one).

Samhit21 (dismissed):
## PR Review: refactor: compress the post-drop layout on / to raise the résumé

### Summary

Reviewed at `7c3e11d`, with `origin/main` merged in first (1 behind, no file overlap). The self-review is thorough and its fixed list holds up — I checked the four code-level ones against the source rather than the replies. Two things I'd change before merge, both Suggestions; nothing blocking.

The architecture is the right call. Putting timer + scroll + hold + lock in `useAutoCollapse` rather than in the component makes the part most likely to be wrong testable without a layout engine, and the tests pin real behaviour rather than shape — "honours the hold: focus inside suppresses the scroll dock" and "keeps holding when focus moves BETWEEN descendants" are the two that would have caught the defect the hook shipped with earlier in this PR. `Popover` restoring focus only on a *real* close (`wasOpen`) and deliberately not trapping is the correct division against `Dialog`.

- **Suites, on the merged tree:** `src/components` + `src/hooks` + `src/design-system` — **124 files, 1287 tests, 0 failed**. `tsc -b --noEmit` exits 0, `eslint` clean on every changed file. CI green on `7c3e11d`.
- **`Disclosure`'s `summary` widening (`string` → `ReactNode`) is safe.** It is consumed only as `<span>{summary}</span>`; the accessible-name addendum is `warnLabel`, not this. The one non-string caller (`TargetingSection:102-110`) passes two nested text spans, and `TargetingTriageRow`'s two anchors live in the disclosure *body*, so no interactive content is nested inside `<summary>`.
- **`AchievementTypePicker`** is exactly the disclosed swap: its own listener deleted, `Popover` adopted, `p-2` → `p-3`. Nothing else moved.
- **No ancestor clips the popover** at either call site; `CollapsedScoreBar`'s only `overflow-hidden` is a 12px progress track.
- **`scoreBandParts`' zero-max guard** is real (`max > 0 ? … : 0`), which is what keeps an ungradable dimension from rendering a `NaN%` width.

### Findings

1. **`AtsScoreReadout.tsx:64` — `resetKey: score.overall` re-expands the docked widget on every edit that moves the score.** [Suggestion] Inline. Review focus asked about the false-negative direction; this is the false-positive one, and it fires in ordinary use: edit a bullet below the fold, the score moves a point, the strip grows back and shoves the résumé down under the caret, then docks again 4.5s later. A real parse identity is already in scope at both mounts (`Result` takes `parseKey`, holds `parseIdentity`); the component just doesn't expose a prop for it. One prop closes both directions.
2. **`Popover.tsx:51` — the `max-w` clamp bounds width, not position, so the panel still runs off-screen from a right-aligned trigger.** [Suggestion] Inline. That is the docked strip below `sm`, where the ⓘ is pinned right and the dimension group is `hidden sm:flex` — i.e. the state most users are in, holding the only written copy of the scoring rules. Reasoned from the CSS; jsdom can't measure it, which is also why no test sees it.
3. **`CollapsedScoreBar.tsx:193` — `EXPAND_LABEL` isn't actually shared.** [Nit] Inline. The docblock's rename-proof guarantee covers `COLLAPSE_LABEL` only.
4. **`useAutoCollapse.ts:148` — the dedicated unmount cleanup is dead, and its test passes without it.** [Nit] Inline. Verified by mutation: 20/20 with the line removed.
5. **`ReconstructedResume.tsx:1487` — the résumé region is now unnamed, and nothing says the fields are editable.** [Nit] No line in the diff to anchor to. Deleting the heading and explainer is a stated AC, so this is the residual, not the removal: `<section id="reconstructed-resume">` carries no `aria-label`/`aria-labelledby`, and it is the jump target of two `ScoreDimensionRow` anchors, one `CompactDimension` anchor and `TargetingTriageRow`'s "Review bullets ↓". Those four links now land on an unlabeled region, and "Click any field to edit it." — the only statement that the résumé is editable at all — is gone with no replacement. An `aria-label` on the section costs nothing and keeps the compaction.

A delegated correctness pass also flagged the heading/explainer deletion and the contact add-row consolidation as *undisclosed* changes. I checked both against the PR body and they are listed acceptance criteria ("`What the parser read` heading + explainer removed", "Add affordances share one consolidated row"), so I've dropped them rather than pass them on.

### Review focus

- **Is auto-collapsing content the right call at all?** Yes, with the holds you built — but the holds only cover reading *the widget*. The case they don't cover is finding 1: the user is reading and editing something else entirely, and the widget moves anyway because the score changed. Fix the key and the remaining motion is always either the first dock after arrival or one the user asked for.
- **Should `resetKey` be a parse identity?** Yes, and it is cheaper than the docblock implies — both call sites already hold one. See finding 1.
- **The dropped all-clear line (#957).** Agreed it belongs in a follow-up rather than here. Worth noting it compounds finding 5: a clean résumé now gets no confirmation anything was checked *and* lands on an unnamed section.
- **Uniform `p-3` on `Popover`, no `tailwind-merge`.** Right trade. A padding prop on a primitive is the beginning of a styling API, and the +4px on the picker is the kind of change that should be visible in a review rather than absorbed silently — which it was. The `className`-appends-only behaviour is worth keeping documented exactly where it is, because finding 2's fix can't use `className` either.

### Verdict

**Action: APPROVE.** Nothing blocking: no crash, no data loss, no egress change, and the compaction it set out to do is demonstrated with measurements rather than asserted. Findings 1 and 2 are the two I'd land before merge — 1 because it fights the user during the task this page exists for, 2 because it hides the scoring explanation on phones — but both are small and neither is a reason to hold the branch. If you'd rather gate on them, say so and I'll switch this to `REQUEST_CHANGES`.

src/components/features/AtsScoreReadout.tsx
s-annam: **Fixed in `9283c83`** — you were right that the defect is over-firing, not under-firing, and that the `## Review focus` note had it backwards.

`AtsScoreReadout` now takes a `resetKey` prop and passes it straight through to `useAutoCollapse`; nothing derived from the score is used as the key any more. `Result.tsx` passes `recovery.parseIdentity` — which, as you noted, also changes when a recovery pass lands, and that *is* a new reveal. `App.tsx`'s authoring mount passes `parseKey` rather than nothing: that lane never re-parses, so `parseKey` moves only on a genuinely new authoring session (`authoring:<generation>`), which is the one moment the reveal should fire again. Passing nothing would have worked too, but would have lost the re-reveal on "start over".

That also closes the false-negative the hook's docblock documented — two résumés scoring the same now re-reveal correctly — so the one prop fixes both directions. The docblock on `resetKey` has been rewritten accordingly; it previously described the score-keyed limitation as a known limit of the key, which is no longer the shape of the thing.

Note the branch was collapsed to a single commit after this fix, so this thread's anchor has moved.

src/hooks/useAutoCollapse.ts
s-annam: **Fixed in `9283c83`.** The scroll listener now records `window.scrollY` when the effect arms and compares `Math.abs(window.scrollY - revealedAt) > scrollThresholdPx`.

Your re-expand case is covered for free: the effect returns early while `collapsed`, so re-expanding re-enters it and re-baselines the origin to wherever the reader now is. The `scrollThresholdPx` docblock now says "away from where the widget was revealed" rather than implying an absolute offset.

Three tests added — a reveal at `scrollY 1500` that survives a small scroll, a large scroll *up* from that position that still docks (direction-blindness is deliberate and unchanged), and a re-reveal that re-baselines.

src/design-system/primitives/Popover.tsx
s-annam: **Fixed in `9283c83`**, as you suggested — `Popover` takes `align?: "start" | "end"`, defaulting to `start`.

`PANEL` split into `PANEL_BASE` (no side class) plus `PANEL_ALIGN = { start: "left-0", end: "right-0" }`. `ScoreExplainerPopover` forwards the prop, because it is one component rendered at *both* alignments — the expanded readout's left-aligned header row keeps the default, and `CollapsedScoreBar` passes `align="end"`.

The docblock paragraph now describes what the clamp actually does. It previously claimed the `max-w` clamp was "what keeps it on screen"; it bounds the width only, and the comment said otherwise, which is most of how this survived review the first time.

src/components/features/TargetingSection.tsx
s-annam: **Fixed in `9283c83`** — `warn={titles.length > 0 && noRolePicked}`, exactly as suggested.

Confirmed the reachability claim: `RolesPanel` returns null without titles and `SkillTermGuidance` renders nothing when it has no guidance, so before this the summary could carry "no role picked, so none prints on your PDF" over a disclosure containing only the triage row — a mark pointing at a control that is not there. The old `return null` guard did make it unreachable; relaxing that guard is what exposed it.

src/components/features/TargetingTriageRow.tsx
s-annam: **Fixed in `9283c83`** — I took your second option and dropped the link, keeping `Edit contact ↑`.

I went looking for the first option and it does not hold up. An Experience anchor would be wrong for a real class of résumé: `score.bullets` is built from a pool that folds in projects and achievements alongside experience (`scoreSpecificity` and `scoreStructure` both add `extraSources`), so a résumé whose flagged bullets all live in Projects would get a confidently wrong destination — the same failure `anchors.ts` was introduced to make impossible, and the `#per-bullet-feedback` case its docblock already carries.

A correct target would be a "document body" anchor covering everything below the disclosure, but everything there is a flat list of siblings under `<section id={SECTION_IDS.reconstructed} className="flex flex-col gap-6">`, so adding a wrapper to hang an id on is a layout change rather than an added attribute — too much for a review round. Filed as **#958** with the two decisions it has to make.

The row's docblock now records why there is deliberately no bullet link, so this does not get "helpfully" re-added.

src/components/features/AtsScoreReadout.tsx
s-annam: **Fixed in `9283c83`.** You and @rohithgollapalli found this independently and your diagnosis was precise, including the call sites: `AtsScoreReadout` now takes a `resetKey` prop with essentially the signature you wrote, and the two mounts pass a real parse identity — `recovery.parseIdentity` in `Result.tsx`, `parseKey` in `App.tsx`'s authoring mount.

You were also right that the hook is not at fault. Nothing in `useAutoCollapse` changed for this; only the key it is handed. Its `resetKey` docblock now states that the key must identify the parse and not anything derived from it, with the layout-jump-under-the-caret failure named, so the next caller does not have to rediscover it.

And as you predicted, one prop closes both directions — the false-negative the docblock used to document (two résumés scoring the same) is gone too.

src/design-system/primitives/Popover.tsx
s-annam: **Fixed in `9283c83`** — `Popover` now takes `align`, and `CollapsedScoreBar` passes `"end"`.

Your reading of the CSS was right, including the detail that made it worse than it first looks: below `sm` the middle dimension group is `hidden sm:flex`, so the row collapses to [score pill] … [explainer + expand] and the ⓘ is pinned near the right edge — in the state the widget docks itself into after 4.5s, on the one surface where the scoring rules are still written down.

You were also right that `className` could not do it: `left-0` lives in the base panel class and there is no `tailwind-merge`, so an appended `right-0` would not have won. Hence a real prop.

Taking your "worth a real-browser check at 375px before deciding how to fix" seriously: I have **not** done that check, and jsdom still cannot see it. The two tests added assert the class swap (`left-0` by default, `right-0` when `align="end"`), which pins the mechanism but not the rendered geometry. If you want the geometry confirmed before this merges, say so and I will measure it.

src/components/features/CollapsedScoreBar.tsx
s-annam: **Fixed in `9283c83`** — both labels now come from one module, `src/components/features/scoreToggleLabels.ts`, imported by `AtsScoreReadout` and `CollapsedScoreBar`.

Your trace of the failure was exactly right, including that there is no fallback: the restore resolves by exact attribute match, the score pill's own label names the value and so cannot match, and the result is a silent drop to `<body>` for a keyboard user — the defect this PR had already fixed once. The new module's docblock records that reasoning, since the whole point is that the coupling is invisible from either file alone.

src/hooks/useAutoCollapse.ts
s-annam: **Fixed in `9283c83`** — deleted, and the test comment corrected.

Your mutation check reproduces here. The premise really does not hold: `release()` returns early unless `!collapsed && !locked`, which is precisely when the arrival-countdown effect is mounted with `return clear`, and `clear` reads `timerRef.current` at call time, so it cancels a hold-armed timer as readily as one it armed itself.

I took the first of your two options. The teardown test stays — it still pins that a timer the countdown effect did not itself arm is cancelled on unmount — but its comment now names that effect's cleanup as the owner instead of claiming a dedicated cleanup is the only thing standing between it and a `setCollapsed` on a dead tree. Agreed that an unreachable safety net is worse than none, and that a test certifying a guarantee it does not exercise is the more expensive half.

offlinecv-approver (approved):
🕹✅

rohithgollapalli (approved):
**Verdict: APPROVE** — 0 Blocking (rule: 0 Blocking → approve, nits listed non-blocking). Re-review of head `9283c83` after `7c3e11d`; branch still one commit.

## Previous findings — all verified fixed in `9283c83`

- **Blocking — `resetKey: score.overall`** → `AtsScoreReadout` now takes `resetKey`; `Result.tsx` passes `recovery.parseIdentity`, `App.tsx` passes `parseKey` (`authoring:<generation>`, moves only on a fresh session — checked in `useAnalyzedResume.ts`). Nothing score-derived is keyed any more.
- **Scroll dock absolute `scrollY`** → `revealedAt` captured on arm, `Math.abs(scrollY - revealedAt) > threshold`; re-expand re-enters the effect so it re-baselines. Three tests cover reveal-at-1500, scroll-up dock, and re-baseline.
- **Popover overflow from the docked ⓘ** → `align?: "start" | "end"` with `PANEL_ALIGN`; `CollapsedScoreBar` passes `"end"`, expanded header keeps `start`. Docblock now says what the clamp does. Two alignment tests.
- **`warn` mark with no `RolesPanel`** → `warn={titles.length > 0 && noRolePicked}`, test added.
- **"Review bullets ↓" scrolling up** → link removed; docblock's reason for not swapping in an Experience anchor (`score.bullets` pools project/achievement bullets via `extraSources`) checks out against `score.ts`.
- The dropped unmount-cleanup effect: the reasoning holds — `release` arms only under `!collapsed && !locked`, which is exactly when the countdown effect is mounted with `return clear`, and `clear` reads `timerRef` at call time. Teardown test still pins it.

## Nits (non-blocking, not pushed)

- Stale docblock references to the deleted `AttentionStrip`, unchanged from last round: `src/components/features/ContactCard.tsx:13,118`, `src/components/features/ReconstructedResume.tsx:1317`, `src/components/features/RolesPanel.tsx:10`, `src/lib/contact.ts:143,216,256`, `src/lib/edit/field-validators.ts:19`. Should name `TargetingTriageRow`/`TargetingSection`. Pure comment edits — I did not push them because GitHub reports the author as `CONTRIBUTOR`, and a fix commit I cannot follow with a collapse would leave the branch two commits into the merge queue, which is worse than the stale prose. Fine to land in a follow-up.
- `fallow`: 0 dead code; 1 complexity finding (inherited, cyclomatic 36 in a 780-line function, re-attributed because the diff now touches `Result.tsx`/`App.tsx`) and 3 clone groups (`Result.tsx:31-62`/`117-127` inherited; `Popover.test.tsx:83-89`/`97-103`; `ContactCard.test.tsx:333-338`) — all warn, report-only inside `verify`.

## Gates (re-run on `9283c83`)

- `npm run typecheck`, `npm run lint`: clean. Touched-file vitest (6 files, 65 tests): green.
- 3a fixture PII: skipped, no fixture touched. 3b: no raw `<button>`/`<dialog>`/`<input>` in feature code; new `scoreToggleLabels.ts` is a constants module with two consumers. 3c tokens: clean. LOC: every feature component ≤ 200. 3e: no skill/script files.
- 3f: description unchanged and still accurate; the `## Review focus` note on `resetKey` is now superseded by the code, and the inline replies record the correction. Pixel figures remain hand-measured and not reproducible here.
- #953 AC checklist (as revised): all code-verifiable ACs met, unchanged from last round.

Push outcome: nothing pushed, nothing collapsed — one-commit invariant already intact.

---

Vaishnavi1709 (approved):
**Verdict: APPROVE** — 0 Blocking. Independent pass on head `9283c83`, not a countersign of the existing approvals.

## What I verified myself

**The one open thread is fixed, and I checked the code rather than either summary.** `Samhit21`'s reading was right: `max-w` bounds width, not position. At this head `PANEL_BASE` no longer carries `left-0` at all — positioning moved into `PANEL_ALIGN = { start: "left-0", end: "right-0" }`, `align` defaults to `start`, and `CollapsedScoreBar:195` passes `align="end"` through `ScoreExplainerPopover`. `right-0` anchors the panel's right edge to the trigger's, so it extends leftward and stays on screen — sound by construction for the docked case, which is the state the widget puts most users in.

**`useAutoCollapse` — I checked the dropped unmount-cleanup argument instead of accepting the docblock.** `release` early-returns unless `!collapsed && !locked`, which is exactly the condition under which the countdown effect is mounted with `return clear`, and `clear` reads `timerRef` at call time. The reasoning holds; an extra cleanup effect would indeed be unreachable.

**House gates on the diff:** no raw `<button>`/`<dialog>`/`<input>` in feature code; no hardcoded hex, raw palette class or manual `dark:` variant (my first grep flagged some — all false positives, `#953`/`#956`/`#605` are issue refs that happen to be hex-valid); `@design-system` barrel imports only; new `Popover` correctly lands in `primitives/` and is consumed by both the explainer and the picker, so the duplicated dismiss listener is genuinely gone rather than forked.

**Spec:** every revised AC on #953 is met, and the retired above-the-fold criterion is retired *on the issue itself* with the 700-vs-800 contradiction spelled out. That is the right handling — retro-picking the fold definition that produced a pass would have made the criterion meaningless.

## Nits — non-blocking, none pushed

- **`## Review focus` bullet 2 is stale.** It still says `resetKey` is `score.overall` and asks whether the caller should pass a parse identity. The code already answers it — `resetKey` is a prop fed from `parseIdentity`/`parseKey`, and the docblock records why the score was the wrong key. A reader hitting that bullet would think the defect is open.
- **`revealedAt` is not re-baselined when `resetKey` changes while the widget is already expanded.** `collapsed` never transitions, so the scroll effect (deps `[collapsed, locked, scrollThresholdPx]`) does not re-enter and keeps the old baseline, while the countdown effect — which does list `resetKey` — restarts. Minor and possibly deliberate; flagging the asymmetry rather than asking for a change.
- `CollapsedScoreBar` 213 LOC and `Popover` 208 sit just over the ~200 guideline, both comment-heavy. Worth noting that `ReconstructedResume` **shrank** by 159 lines here, which cuts the repo's worst offender.
- Agreed with the stale `AttentionStrip` docblock references already listed; follow-up material.

## What I did not verify

The pixel figures. jsdom has no layout engine and I have no browser here, so 186px/62px/472px and the 375px popover geometry are unmeasured by me — as `s-annam` states plainly in the thread, and as #954 exists to fix. I am approving the mechanism and the code, not re-attesting the measurements.

## Note on merge state

This PR is already `APPROVED`, and my approval does not unblock it: `Samhit21`'s thread is still open, and `required_conversation_resolution` on `main` holds the merge on that alone. It is marked outdated and the fix is in — but it closes on a direct question ("worth a real-browser check at 375px") that `s-annam` offered to answer and nobody has. That is `Samhit21`'s or `s-annam`'s to close, not mine to resolve over.

src/design-system/primitives/Popover.tsx
Vaishnavi1709: This thread is now the only thing holding the merge — `required_conversation_resolution` on `main` blocks on one open thread regardless of approvals, and everything else is green (two approvals plus mine, all checks passing, one commit).

I verified the mechanism fix independently at `9283c83`: `left-0` is gone from `PANEL_BASE`, positioning moved into `PANEL_ALIGN = { start: "left-0", end: "right-0" }`, and `CollapsedScoreBar:195` passes `align="end"`. `right-0` anchors the panel's right edge to the trigger's, so it opens leftward — which addresses exactly the case @Samhit21 identified, including the below-`sm` collapse that pins the ⓘ against the right edge.

What is still unanswered is the check @Samhit21 asked for and @s-annam offered: the real-browser read at 375px. jsdom cannot see it and the two new tests pin the class swap, not the rendered geometry — so nobody has confirmed the pixels, and I cannot either.

Either resolution works and both are quick:

- **@Samhit21** — if the class swap is enough for you, say so and resolve.
- **@s-annam** — if the geometry should be confirmed first, measure at 375px, post the result, and resolve.

Flagging rather than resolving it myself: it is your thread and it closes on a question nobody has answered, so resolving it over you would dismiss the request rather than settle it.

s-annam: You were right, and the check you asked for says so. I ran it before deciding anything else.

Measured in Chrome at 375×812 on `9283c83`, against a real parse of
`google-docs-skia-proxy-classic.pdf`:

| Widget state | `align` | class | panel x-span | off-screen by |
|---|---|---|---|---|
| Docked score strip | `end` | `right-0` | **−225 → 63** | 225px past the LEFT edge |
| Expanded readout header | `start` | `left-0` | **218.75 → 506.75** | 147px past the RIGHT edge |

`document.documentElement.clientWidth` is 360; the panel is 288px (`w-72`) in both cases.

So **the `align` prop I added does not fix this at 375px** — it only chooses which edge the panel
falls off. Two things I got wrong by reasoning instead of measuring:

1. **The docked strip wraps at this width.** Its root is `flex flex-wrap … justify-between`, and
   at 375px it becomes two rows (measured height 62), so the ⓘ is no longer pinned to the right
   edge — it measures at x 45–63, near the *left*. `right-0` then anchors the panel's right edge
   there and throws 225px of it off the left. The docblock I wrote reasons from a right-pinned
   trigger, which holds at `sm` and up and is false below it.

2. **The `max-w` clamp never engages.** `calc(100vw-2rem)` is 343px here and the panel is 288px,
   so 288 < 343 and the clamp is inert — exactly your point. It bounds width, never position.

I've filed #959 with the measurements, the root cause, and acceptance criteria requiring
`left >= 0 && right <= clientWidth` in both alignments and both widget states, plus the same for
`AchievementTypePicker` as the other consumer.

**I'm proposing we punt the fix to #959 rather than push it here** — and I want to be straight
that this is a scheduling argument, not a severity one. The defect is real, and it is a
regression this PR introduces: the explainer was a `<details>` before, which expanded in flow and
could not leave the viewport. The reason to defer is narrow — this PR now carries three
approvals, and any push dismisses all of them, so the trade is three re-reviews against a
one-line class change made under time pressure.

That is my judgement and not yours, so **I'm leaving this thread open for you to call.** It is
the only thing blocking the merge right now (`required_conversation_resolution`). If you're
happy with the deferral, resolving it is the unblock. If you'd rather this not ship broken on
mobile, say so and I'll fix it here and absorb the re-approval cost — that's a legitimate answer
and I won't push back on it.

One more thing #959 records, explicitly unattributed: with the popover closed, the page has
horizontal scroll at 375px (`scrollWidth` 435 vs `clientWidth` 360; the overflowing node is a
`div.flex.items-center.gap-3`). I did not check that against `main`, so I am not calling it a
regression of this PR.

s-annam:
**Merging with a known, tracked defect — recording it here so the merge isn't silently clean.**

At 375px the score explainer popover renders off-screen in **both** alignments. Measured in
Chrome at 375×812 on `9283c83`: the docked strip's panel spans −225 → 63 (225px past the left
edge), the expanded header's spans 218.75 → 506.75 (147px past the right). `clientWidth` is 360.

This PR introduces it — the explainer was a `<details>` block before, which expands in flow and
cannot leave the viewport. The `align` prop added in review fixes 1280px but only selects which
edge the panel falls off at 375px.

Raised by @Samhit21, whose finding the measurements confirm. Deferred to #959 as a scheduling
decision — this PR carries three approvals and any push dismisses all of them — not because the
finding was unresolved or disputed. #959 has the measurements, the root cause, and acceptance
criteria covering both alignments and both widget states.

Issue #960, opened 2026-09-17
[ui] Docked score strip wraps to two lines depending on the verdict label's width, so the card height depends on the score


s-annam:
## Summary

The docked score strip (`CollapsedScoreBar`, #953) is specified as a **one-line** strip — the
commit message for fd25862 quotes "~38px height" for the strip and 62px for the docked card. It
holds that shape for some résumés and silently becomes **two lines** for others at the same
window width, so the card above the résumé is 2 rows in one case and 3 rows in the other.

Observed on `/` at what appears to be the same viewport (two screenshots, ~995px wide crops of
the same card):

- **Score 92 · STRONG** — one line: score pill, three dimension tracks, `ⓘ Score details ▾`
  all on the same row. Card = header row + 1 strip row.
- **Score 60 · GETTING THERE** — two lines: pill + dimension tracks on row 1, `ⓘ Score details ▾`
  wrapped onto row 2 on its own. Card = header row + 2 strip rows.

Nothing about the two parses differs except the score data. The height of a layout primitive
should not be a function of which verdict band the user landed in.

## Root cause

`src/components/features/CollapsedScoreBar.tsx:108` — the strip root is:

```tsx
<div className="flex flex-wrap items-center justify-between gap-3 text-sm">
```

Three sibling groups, all with intrinsic width, and `flex-wrap` with no ordering, no
`min-w-0`/shrink target and no reserved widths:

1. `src/components/features/CollapsedScoreBar.tsx:110` — score pill group (`Button` + optional layout-penalty text)
2. `src/components/features/CollapsedScoreBar.tsx:161` — `hidden sm:flex` dimension group (three `CompactDimension`)
3. `src/components/features/CollapsedScoreBar.tsx:193` — `ⓘ` popover + `Score details ▾` toggle

Whether group 3 fits on row 1 therefore depends on the **rendered width of the score data**, and
group 1's width is data-dependent in at least four ways:

- **Verdict label length.** `getScoreLabel` (`src/lib/score/score.ts:596`) returns `Strong` (6
  chars), `Needs Work` (10) or `Getting There` (13), rendered `uppercase tracking-wider`
  (`CollapsedScoreBar.tsx:145`), which amplifies the delta. `GETTING THERE` is the widest and is
  exactly the case that wraps in the screenshots.
- **Optional layout-penalty span** (`CollapsedScoreBar.tsx:152`) — `(layout penalty ×0.85)` is
[...]

Issue #959, opened 2026-09-17
[ui] Popover panel renders off-screen at 375px in BOTH alignments — align picks the edge, not the containment


s-annam:
## Problem

At 375px the `Popover` panel renders entirely outside the viewport **in both alignments**.

Measured in Chrome at 375×812 on branch `gh-953` (`9283c83`), against a real parse of
`tests/fixtures/pdfs/google-docs/google-docs-skia-proxy-classic.pdf`:

| Widget state | `align` | class | panel x-span | off-screen by |
|---|---|---|---|---|
| Docked score strip | `end` | `right-0` | **−225 → 63** | 225px past the LEFT edge |
| Expanded readout header | `start` | `left-0` | **218.75 → 506.75** | 147px past the RIGHT edge |

`document.documentElement.clientWidth` = 360 (375 minus scrollbar). Panel width is 288px
(`w-72`) in both cases.

## Why `align` did not fix it

#956 added the `align` prop to `Popover` to stop the docked strip's explainer opening off the
right edge. It fixes 1280px. It does not fix 375px, for two independent reasons:

1. **The docked strip wraps at narrow widths.** Its root is
   `flex flex-wrap items-center justify-between`. At 375px it becomes two rows (measured
   `stripRect` height 62), so the ⓘ trigger is no longer pinned to the right edge — it measures
   at x 45–63, near the *left*. `right-0` then anchors the panel's right edge at x=63 and throws
   225px of it off the left. The prop's docblock reasons from a right-pinned trigger, which holds
   at `sm` and up and is false below it.

2. **The `max-w` clamp never engages.** `max-w-[calc(100vw-2rem)]` is 343px at this viewport, but
   the panel renders `w-72` = 288px. 288 < 343, so the clamp is inert. It bounds width and never
   position — stated correctly in the primitive's docblock, but the consequence is that nothing
   in `Popover` is viewport-aware.

So `align` chooses *which* edge the panel falls off, not whether it stays on screen.

## Regression scope

#956 converts the score explainer from a `<details>` block into this absolutely-positioned panel.
The `<details>` expanded in flow and could not leave the viewport, so this behaviour is new to
that PR rather than pre-existing on `main`.

The panel is, per `Popover.tsx`'s own docblock, "the one keyboard-reachable place the scoring
rules are written down" — so below `sm` that content is currently unreachable.

## Suggested direction (not prescriptive)

Pin the panel to the viewport below `sm` rather than to a trigger edge — e.g.
[...]

PR #962, opened 2026-09-18
test(e2e): add a Playwright viewport harness for layout claims (#954)

### LINKED ISSUE #954: [testing] Add a Playwright viewport harness so above-the-fold and component-height claims are testable

### LINKED ISSUE #975: e2e harness: npm run test:e2e can measure a stale dist/ and report green (reuseExistingServer skips the build)
## Problem

`playwright.config.ts` (added in #962) sets:

```ts
webServer: {
  command:
    "npm run build && OFFLINECV_DEV_HTTP=1 npm run preview -- --port 4173 --strictPort",
  url: BASE_URL,
  reuseExistingServer: !process.env.CI,
  timeout: 120_000,
},
```

`npm run build` lives **inside** `webServer.command`, and `reuseExistingServer: true` (the
local default) makes Playwright skip that command entirely whenever something already
answers on `http://localhost:4173`. So locally, whenever a preview server is already up —
a leftover from a killed run, a second terminal, a previous `npm run test:e2e` that was
Ctrl-C'd — **`npm run test:e2e` tests a stale `dist/` and reports green.**

## Why it matters

This is the failure shape the harness exists to eliminate. #954 was filed because layout
claims were "unfalsifiable by the test suite"; a suite that silently measures a build from
an hour ago is unfalsifiable in a new way, and it fails in the reassuring direction. A
developer who inflates `AtsScoreReadout` by 200px, runs `npm run test:e2e` against a stale
preview, and sees 6 green tests has been told their change is safe by a gate that never
loaded it.

CI is unaffected (`reuseExistingServer` is false when `CI` is set), which is exactly what
makes this hard to notice: it only lies on the machine where the developer is iterating.

## Secondary effect: a build failure is misreported

Because the build is inside `webServer.command`, a `tsc -b` or `vite build` failure does not
surface as a compiler error — Playwright reports `Process from config.webServer was not able
to start. Exit code: 2`, with the underlying tsc output swallowed. Observed while reviewing
#962: a type error in a spec file produced exactly that, with no indication of which file or
rule failed.

Note the coupling that makes this reachable at all: `tsconfig.json` now references
`tsconfig.e2e.json`, so `npm run build` (`tsc -b && vite build`) typechecks `e2e/**`. A type
error in a *test* file therefore breaks the *build* step that serves the app.

## Proposed fix

Move the build out of `webServer.command` so that reuse only ever skips *serving*, never
*building*. Options, roughly in order of preference:

1. Make `test:e2e` the composite step instead of the config:
   ```json
   "test:e2e": "npm run build && playwright test"
   ```
   and reduce `webServer.command` to the preview alone. Reuse then skips only the server.
2. Keep the build in the config but drop `reuseExistingServer` to `false` unconditionally —
   correct, at the cost of a full rebuild + boot on every local run.
3. Use a `globalSetup` that builds, leaving `webServer` to serve.

Option 1 also fixes the misreported build failure: `npm run build` fails in the foreground
with the real tsc/vite error before Playwright is invoked.

## Acceptance criteria

- [ ] With a preview server already listening on 4173, `npm run test:e2e` still rebuilds
      `dist/` (or fails loudly) rather than silently measuring the stale bundle.
- [ ] Fail-before check: inflate a component's height, leave a stale server running, and
      confirm the suite now goes red where it previously passed.
- [ ] A `tsc` error in `e2e/**` surfaces as the compiler's own message, not as
      `Process from config.webServer was not able to start. Exit code: 2`.
- [ ] CI behaviour is unchanged (still builds, still does not reuse).
- [ ] `npm run verify` green.

Found while reviewing #962.

s-annam:
Adds the Playwright viewport harness from #954, so layout claims about `/` stop being unfalsifiable. Vitest runs in jsdom, which has no layout engine — `getBoundingClientRect()`, `offsetHeight` and `clientHeight` all evaluate to `0` — so the "résumé starts above the fold" promise had no regression cover at all.

**Layer 1 of 5. Base: `main`.** The four PRs above it depend on this one; #965 in particular adds assertions that only exist because this harness does.

## What ships

- `playwright.config.ts` — Chromium only, projects `desktop-1280` (1280×800), `desktop-1440` (1440×900), and `mobile-375` (375×812, activated by #965). Serves a production `vite preview` of `dist/` over plain HTTP via `OFFLINECV_DEV_HTTP=1`. Specs are routed **by directory**: the desktop projects take everything in `e2e/` except `e2e/mobile/`, so a new spec runs by default; mobile is the one explicit opt-out, because the assertions here are `md`+-scoped by the design itself (`AtsScoreReadout`'s `md:flex-row`).
- `e2e/viewport.spec.ts` — drops a synthetic fixture on `/` and asserts the score hero's height, the score card's height, the first résumé section's above-the-fold position, and that both hash anchors clear the sticky header.
- `tsconfig.e2e.json` + a reference from `tsconfig.json` — without it neither the config nor `e2e/**` was reachable from any project, so `npm run typecheck` would not have checked them.
- An `e2e/**/*.ts` block in `eslint.config.js` — every block it had was scoped to `src/**`, and `eslint .` lints only what a `files:` block matches, so these specs would have shipped typechecked but **never linted** (`npx eslint e2e/` answered "no matching configuration was supplied"). The block is type-aware, unlike the `src/**` one, because a Playwright locator call returns a promise: the characteristic bug in a spec is a missing `await`, which reports green while the assertion never runs, and `no-floating-promises` needs type information to see it. `playwright.config.ts` stays under the `*.config.ts` global ignore alongside `vite.config.ts`.
- A separate `e2e` CI job — deliberately **not** folded into `verify`, which is the fast local pre-push gate and must not acquire a browser download.
- `npm run test:e2e` (which builds `dist/` itself, then lets the config's `webServer` preview it), `.gitignore` entries for run artifacts, and a README section.

## Two traps found while building it

- **`basicSsl()` is not dev-only.** Both `vite dev` and `vite preview` pick it up from the shared plugin list, so preview serves HTTPS with a self-signed cert unless `OFFLINECV_DEV_HTTP=1` is set. Found by `curl` returning "Empty reply from server", not by reading the config.
- **The docked pill's `aria-label` contains "Expand score details" as a substring**, so any non-exact role query matches two elements and trips strict mode. Every locator here uses `exact: true`.

## What this does and does not defend

The above-the-fold assertion targets the **Summary heading** (751.5px against an 800px fold), not the `#reconstructed-resume` container — the container's top sits ~330px clear and would keep passing through a large regression. Its lower bound is the sticky header's bottom (135px), not `0`, so a heading that regressed to underneath the header fails instead of passing invisibly. #955's contradicting 810px figure was measured mid-implementation on `gh-953`, before commits that shipped in #956; a fresh measurement on `main` matches #956's own merged table.

It says nothing about the two-column fixture (#956 measured that at 834px, ❌) or the original "~700px effective" fold budget. Do not cite a green run here as those being fixed — they are #955's scope.

**`DOCKED_MAX_PX = 70` is calibrated ON the wrap, not against it.** The 62px "docked" figure is the strip *already wrapped to two rows* — the bug #960 fixes — so this ceiling would admit a re-wrap; what it catches is the strip growing past its present two rows. The constant's comment now says exactly that, and points at #965, which moves it to `e2e/support/score-hero.ts` and re-derives it from the true 26px one-row height. A threshold measured on unfixed code certifies nothing, and the place to say so is next to the threshold.

## Verification

`npm run verify` green (typecheck · lint · `check:nul` · `check:fixtures` · `check:baselines` · `check:core` · full vitest · build · fallow clean on 9 changed files) · `npm run test:e2e` 6/6 across both desktop projects.

Project routing was verified with throwaway specs rather than asserted: a new top-level spec runs on both desktop projects, a spec under `e2e/mobile/` runs only on `mobile-375`. That check caught a real bug in the first attempt — Playwright matches a regex `testMatch`/`testIgnore` against the **absolute** path, so an anchored `/^mobile\//` matched nothing and failed *open*.

## Adversarial review

Reviewed adversarially before this PR existed (2 rounds), then again via `/pr-review` on the opened PR: 0 blocking, 7 secondary, 4 nits. All ten review threads are addressed in the single commit on this branch:

- the `DOCKED_MAX_PX` comment now states what the ceiling does and does not defend (above);
- a `CARD_MAX_PX = 300` ceiling on the score `Card` — `ParsedHeader` and the two-column warning add 90px that sat outside both hero ceilings while pushing the résumé down exactly the same way;
- the Summary assertion's lower bound is the header's bottom rather than `0`;
- the hash-anchor test docks the widget deliberately up front (`toggle` locks out both the timer and the scroll listener) instead of relying on a focus side effect to suppress auto-collapse, and waits for `window.scrollY` to settle rather than counting frames;
- the toggle labels are imported from `src/components/features/scoreToggleLabels.ts`, the module that exists so a rename moves every side at once, instead of a third private copy;
- per-project `testMatch` pinned to one named file became directory routing, so a new spec can no longer be silently skipped while the run reports green;
- `npm run build` moved out of `webServer.command` into the `test:e2e` script — `reuseExistingServer` skips that command whenever a stale preview is already answering, so the build was being skipped exactly when it mattered (fixes #975);
- `webServer.command` binds `${PORT}` instead of a hardcoded `4173`;
- `tsconfig.e2e.json` gains `DOM.Iterable`, matching `tsconfig.app.json`;
- the `retries: 0` rationale now names both margins — 8px absolute (the docked ceiling) and ~6% relative (Summary above the fold at 1280×800) — rather than a single "~48px" figure that is true only as a percentage and only at one viewport;
- `e2e/**` is linted (above), closing the gap where this PR's own 200+ lines of TypeScript had no lint gate at all.

`retries: 0` is deliberate: every assertion here is geometric, so a retry could only paper over a real flake.

The lint block is verified fail-before, not just green:

```console
$ npx eslint e2e/           # with a deliberate unused import + un-awaited page.goto()
  1:29  error  'Locator' is defined but never used                 @typescript-eslint/no-unused-vars
  4:3   error  Promises must be awaited, …                         @typescript-eslint/no-floating-promises
✖ 2 problems (2 errors, 0 warnings)
```

Review follow-ups **#975** (stale `dist/`) and **#976** (unlinted `e2e/**`) are both fixed here rather than deferred. **#977** stays open on purpose: making the `e2e` job a required status check is a branch-protection change, not a diff, and it is deliberately sequenced after the job has some green history on `main`.

Closes #954


s-annam (commented):

Clean implementation of the Playwright viewport harness (#954). Falsifiable real-browser assertions for the score hero compact height, initial above-the-fold position of the Summary heading, and hash-anchor clearance against the sticky header. All 5 acceptance criteria are met, local and CI gates are green, and the suite runs in ~13s locally and ~1m in CI.

## Acceptance criteria checklist (#954)

- [x] **Playwright installed, configured, and runnable locally via a documented npm script:** `@playwright/test` in devDependencies, `playwright.config.ts`, `"test:e2e": "playwright test"` in `package.json`, and documented in `README.md`.
- [x] **A spec that drops a synthetic fixture PDF on `/` and asserts the score hero height and the above-the-fold position of the first résumé section at 1280×800:** Covered in `e2e/viewport.spec.ts` (`desktop-1280` and `desktop-1440`).
- [x] **A spec covering the two hash-anchor scroll targets landing clear of the sticky header:** Covered in `e2e/viewport.spec.ts` for `#contact` and `#reconstructed-resume`.
- [x] **Runs in CI as a separate job; does not slow down `npm run verify`:** Dedicated `e2e` job added to `.github/workflows/ci.yml` installing only Chromium; `verify` remains browser-free.
- [x] **The manual-only acceptance criteria on #953 are re-stated as automated assertions here:** Codified as geometric assertions against real Chromium bounding boxes.

## Gate audit

| Gate | Status | Notes |
|---|---|---|
| 3a Fixture PII | PASS | `npm run check:fixtures` clean (61 PDFs, 16 sidecars); reuses synthetic `awesome-cv-resume.pdf`. |
| 3b Design-system / reuse | PASS | No components modified under `src/components/`. |
| 3c Style tokens | PASS | No CSS or TSX modified under `src/`. |
| 3d Dead code / fallow | PASS | `fallow audit` report clean: 0 issues in 9 changed files. |
| 3e Skill / script bugs | PASS | `ci.yml` `e2e` job properly scoped, pinned to Node 20, uses `actions/checkout@v5`, has 15m timeout, and archives trace artifacts on failure. |
| 3f Description accuracy | PASS | Body accurately reflects all changes, constraints, and limitations; distinguishes what this layer defends from open items in #955/#960. |

## Verification

- `npm run test:e2e` passed (6/6 tests across 1280 and 1440 viewports in 12.8s).
- `npm run verify` passed cleanly (typecheck, lint, nul check, fixtures, baselines, core, vitest changed, build, fallow).
- GitHub CI checks all green: `CI/e2e` (1m4s), `CI/verify` (4m37s), `Cloudflare Pages`, `fallow`.

## Readiness

This PR is **ready to merge** (Layer 1 of the 5-layer stack). It is already a single clean commit (`4795a41`) without AI trailers. Under `required_conversation_resolution` there are 0 open threads. External reviewer approval or admin-bypass will satisfy branch protection for queue entry.

---
Reviewed by: Gemini 3.8 Flash (high)

e2e/viewport.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+
+// Measured 185.5px. +14.5px (~8%) covers cross-machine font/subpixel
+// variance; nowhere near the ~340px the pre-#956 static hero measured, which
+// is the regression this ceiling exists to catch.
+const EXPANDED_MAX_PX = 200;
+// Measured 62px, both viewports. +8px (~13%) for the same rendering variance
+// — deliberately NOT enough margin to also pass a docked strip that has
+// wrapped to a second line, which #960 (stacked on this issue) could
+// introduce and would land around ~100px; that regression should fail here.
+const DOCKED_MAX_PX = 70;
```
s-annam, line 82: **Secondary — this ceiling certifies the opposite of what the comment says.**

I measured the docked strip's internals rather than just its height. It is `flex flex-wrap items-center justify-between gap-3`, and its three children land on **two rows**:

```
child y-offsets: 228 (h=26), 231 (h=20), 266 (h=24)   → 2 distinct rows
62 = 26 (row 1) + 12 (gap-3) + 24 (row 2)
```

So the 62px baseline **is** the wrapped two-line state. That makes both factual claims in the comment above false:

- *"deliberately NOT enough margin to also pass a docked strip that has wrapped to a second line"* — it doesn't just pass one, it was calibrated on one.
- *"which #960 could introduce and would land around ~100px"* — it is already present, at 62px, not a hypothetical future 100px.

Your PR body already says this (`The real one-line height is 26px`), and says #965 moves the constant to `e2e/support/score-hero.ts` at 34. So this is disclosed and remediated downstream — which is why I am calling it **Secondary, not Blocking**. I did weigh Blocking: gate 3f makes "claims a behaviour the code does not implement" blocking. It came down to (a) no #954 AC asks for a docked assertion at all, so nothing is unmet, (b) the PR body is accurate and volunteers the problem, and (c) merging this is still net-positive cover versus not merging it.

What I would not do is let the **comment** land in `main` as-is. A PR body is not what someone reads when they open this file in a year; the comment is, and it currently tells them the ceiling guards something it does not. Three defensible fixes, your call on how to split with #965:

1. tighten to 34 here and re-measure, or
2. drop the docked assertion until #960/#965 land, or
3. keep 70 but rewrite the comment to say plainly that it is calibrated on the known-wrapped state and does **not** defend the #960 regression, pointing at #965.

e2e/viewport.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+
+const COLLAPSE_LABEL = "Collapse score details";
+const EXPAND_LABEL = "Expand score details";
```
s-annam: **Secondary — third private copy of a constant that exists in one module precisely to prevent this.**

`src/components/features/scoreToggleLabels.ts` owns these two strings, and its docblock is explicit about why:

> The score widget's two toggle labels, in one module because a rename has to move both sides at once (#953). […] a constant private to either one only guards half of it.

This spec adds a third copy, so a rename now has three sites. It fails loudly rather than silently (the locator matches nothing → `boundingBox()` is null → the `not.toBeNull()` assertion trips), so it is not dangerous — just the thing the module was created to stop.

I verified the import works from `e2e/` rather than assuming it: `npm run typecheck` clean and the suite still 3/3 green with

```ts
import {
  COLLAPSE_LABEL,
  EXPAND_LABEL,
} from "../src/components/features/scoreToggleLabels.ts";
```

Not offering this as a suggestion block because it does add a `src/` → `e2e/` edge to `tsconfig.e2e.json`'s program, and whether you want that coupling is a design call, not a nit.

e2e/viewport.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+      name: "Summary",
+      exact: true,
+    });
+
+    // Expanded state — the harder case, since it's the taller of the two.
+    // Read before the auto-collapse timer can dock the widget, same margin
+    // as the hero-height test above.
+    const summaryBox = await summaryHeading.boundingBox();
+    expect(summaryBox).not.toBeNull();
+    expect(summaryBox!.y).toBeGreaterThanOrEqual(0);
```
s-annam: **Secondary — the lower bound is `0`, which lets an invisible heading pass.**

`header.sticky` is 135px tall (I measured it; it matches the `119 + 16 = 135px band` comment in `styles.css:60`). A Summary heading that regressed to `y = 50` would be sitting *underneath* the sticky header — completely invisible — and still satisfy `y >= 0 && y < 800`.

"Above the fold" should mean "below the header and above the fold". Test 3 already resolves `header.sticky`, so the bound is available:

```ts
const headerBox = await page.locator("header.sticky").boundingBox();
expect(summaryBox!.y).toBeGreaterThanOrEqual(headerBox!.y + headerBox!.height);
```

Safe today — I measured `summaryY = 751.5` against a header bottom of 135, so the stricter bound passes at both viewports.

Related, same assertion: the upper bound has **48.5px** of slack at 1280×800 but **148.5px** at 1440×900 (same 751.5px heading, taller viewport). So `desktop-1440`'s copy of this test is the *loosest* assertion in the suite, not part of the "thinnest" characterisation.

e2e/viewport.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+ *  a `data-testid` (it's a test-file-only stub convention, see
+ *  `ResultDetail.test.tsx`), and both states already expose a stable,
+ *  user-facing name for their toggle.
+ *
+ *  `exact: true` matters: the docked strip's score pill carries a SEPARATE,
+ *  longer aria-label ("Resume score 75 out of 100, Getting There. Expand
+ *  score details.") that also *contains* `EXPAND_LABEL` as a substring, so a
+ *  non-exact match resolves two elements and Playwright refuses to guess
+ *  between them. */
+function heroSection(page: Page): Locator {
```
s-annam, line 126: **Secondary — the two ceilings bound the component, not the budget the issue is about.**

`heroSection()` resolves `AtsScoreReadout`'s own `<section>`. On `/` that section is one flex child of `<Card className="flex flex-col gap-6 shadow-xs">`, alongside `ParsedHeader` and the two-column `ErrorState` (`src/components/Result.tsx:181-211`).

Measured on this branch:

```
hero <section> : 185.5px
enclosing Card : 275.5px      → 90px of ParsedHeader + gap-6 + Card padding
```

That 90px is outside both `EXPANDED_MAX_PX` and `DOCKED_MAX_PX`, yet it pushes the résumé down exactly the way #953/#956 care about. `ParsedHeader` growing a row, or the two-column banner appearing, moves the résumé and leaves both ceilings untouched.

Mitigating, and why this is Secondary rather than worse: the Summary-above-the-fold test *does* bound the aggregate end-to-end, so the suite is not blind to it — it is the hero ceilings specifically that are narrower than their framing suggests. If you want them to bound the budget, measure the `Card`.

e2e/viewport.spec.ts
```diff
@@ -0,0 +1,207 @@
...
+    ] as const) {
+      // A real `<a href="#...">` click (ScoreDimensionRow / CollapsedScoreBar)
+      // is a same-document hash navigation, not a full page load — the parsed
+      // résumé survives it. A `page.goto("/#...")` would NOT: the parse isn't
+      // persisted until the first edit (`useAutosaveResume`), so a full
+      // reload here would land back on the empty drop zone.
+      await page.locator(`a[href="${href}"]`).first().click();
+      // One frame for the (instant — styles.css sets no scroll-behavior, so
+      // this is not a "smooth" animation) scroll-padding-top jump to settle.
+      await page.evaluate(() => new Promise((r) => requestAnimationFrame(r)));
```
s-annam: **Secondary — the settle wait is one rAF, and the test is stable today by accident.**

`useAutoCollapse` docks on >40px of scroll displacement (`src/hooks/useAutoCollapse.ts`), and these hash jumps scroll far more than that. The dock is suppressed only because clicking the in-hero `<a href="#contact">` focuses it, setting the focus hold that short-circuits the scroll listener.

Move those anchors out of the hero, or lose focus for any reason, and the widget docks mid-loop: the hero shrinks ~124px, everything above the target reflows, and the header-clearance assertion starts failing non-deterministically — with a single `requestAnimationFrame` not enough to even observe a consistent state.

Cheap hardening: collapse the widget explicitly before the loop, or assert its state inside it, so the test does not depend on a focus side effect it never mentions.

playwright.config.ts
```diff
@@ -0,0 +1,100 @@
...
+  // a mobile spec and the maintainer asked that this harness not need a
+  // rewrite to grow one. `testMatch` keeps it from picking up the desktop
+  // spec — the score-hero/above-the-fold assertions are `md`+-scoped by the
+  // product design itself (`AtsScoreReadout`'s `md:flex-row`), not just by
+  // the numbers in this file, so running them at 375 would be asserting a
+  // claim the app never made.
+  projects: [
+    {
+      name: "desktop-1280",
+      testMatch: /viewport\.spec\.ts/,
```
s-annam: **Secondary — per-project `testMatch` means a future spec silently never runs.**

Both desktop projects pin `/viewport\.spec\.ts/` and `mobile-375` pins `/mobile\/.*\.spec\.ts/`. A new `e2e/export.spec.ts` would match **no project**, never execute, and the run would still report green — there is no "no tests found" error, because the other projects do have tests.

I confirmed the shape on this branch: `mobile-375` matches zero files today and the run reports `Running 6 tests using 5 workers` with no warning about the empty project.

For a harness whose whole purpose is that a green check should mean something, "adding a spec file does nothing and says nothing" is the wrong default. Prefer a directory convention (`testDir` per project, or `testIgnore`) so the default is *runs everywhere* and opting out is the explicit act.

playwright.config.ts
```diff
@@ -0,0 +1,100 @@
...
+      name: "mobile-375",
+      testMatch: /mobile\/.*\.spec\.ts/,
+      use: { ...devices["Desktop Chrome"], viewport: { width: 375, height: 812 } },
+    },
+  ],
+  webServer: {
+    command:
+      "npm run build && OFFLINECV_DEV_HTTP=1 npm run preview -- --port 4173 --strictPort",
+    url: BASE_URL,
+    reuseExistingServer: !process.env.CI,
```
s-annam, line 127: **Secondary — filed as #975, thread left open deliberately.**

`npm run build` lives inside `webServer.command`, and `reuseExistingServer: true` (the local default) skips that command entirely when anything already answers on `:4173`. So locally — leftover preview from a killed run, a second terminal — `npm run test:e2e` measures a **stale `dist/`** and reports green. CI is unaffected, which is what makes it hard to notice: it only lies on the machine where you are iterating.

Same "green check certifies nothing" shape the harness exists to remove.

I hit the secondary effect too: a build failure inside `webServer.command` surfaces as `Process from config.webServer was not able to start. Exit code: 2` with the tsc error swallowed.

Filed as **#975** with a fail-before AC, but I am leaving this thread open rather than resolving it — the fix is one line (`"test:e2e": "npm run build && playwright test"`, leaving `webServer.command` as the preview alone) and you may well prefer it here than in the backlog.

playwright.config.ts
```diff
@@ -0,0 +1,100 @@
...
+  reporter: "list",
+  // A committed `test.only` silently reduces the whole suite to one test while
+  // the job still reports green — the same "a green check that certifies
+  // nothing" failure this harness exists to prevent. Fail the CI run instead.
+  forbidOnly: !!process.env.CI,
+  // No retries, deliberately. Every assertion here is geometric (a rect, a
+  // height, a containment), so a second attempt cannot legitimately disagree
+  // with the first: a retry would only ever paper over a real flake — and the
+  // thinnest assertion in this suite has ~48px of margin, which is exactly the
+  // kind of signal that must surface rather than be re-rolled.
```
s-annam: **Nit — the margin claim is true as a percentage, false as stated in pixels.**

Measured margins on this branch:

| Assertion | Margin |
|---|---|
| docked ceiling (62 → 70) | **8px** (11%) |
| expanded ceiling (185.5 → 200) | 14.5px (7%) |
| Summary above fold @1280 | 48.5px (**6%**) |
| Summary above fold @1440 | 148.5px (16.5%) |

So ~48px is the thinnest *relative* margin, and only at 1280×800 — in absolute pixels the thinnest is 8px. The sentence says "px", which reads as false.

```suggestion
  // thinnest RELATIVE margin in this suite is ~6% (the Summary-above-the-fold
  // assertion at 1280x800: 48.5px of an 800px fold), which is exactly the
  // kind of signal that must surface rather than be re-rolled.
```

playwright.config.ts
```diff
@@ -0,0 +1,100 @@
...
+    },
+    {
+      name: "mobile-375",
+      testMatch: /mobile\/.*\.spec\.ts/,
+      use: { ...devices["Desktop Chrome"], viewport: { width: 375, height: 812 } },
+    },
+  ],
+  webServer: {
+    command:
+      "npm run build && OFFLINECV_DEV_HTTP=1 npm run preview -- --port 4173 --strictPort",
```
s-annam: **Nit — `PORT` is declared but the command hardcodes the port.**

`BASE_URL` derives from `PORT`, the command does not. Changing `PORT` moves the URL Playwright polls but not the port the server binds — which fails as an opaque 120s `webServer` timeout rather than anything that names the cause.

```suggestion
    command:
      `npm run build && OFFLINECV_DEV_HTTP=1 npm run preview -- --port ${PORT} --strictPort`,
```

tsconfig.e2e.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.e2e.tsbuildinfo",
+    "target": "ES2022",
+    "lib": ["ES2022", "DOM"],
```
s-annam: **Nit — `lib` is missing `DOM.Iterable`, which `tsconfig.app.json` has.**

`page.evaluate()` bodies are real DOM code compiled under this config. Without `DOM.Iterable`, iterating a `NodeList` or `HTMLCollection` inside one fails with a non-obvious error:

```
error TS2488: Type 'HTMLCollection' must have a '[Symbol.iterator]()' method that returns an iterator.
```

I hit this within minutes of writing a throwaway probe against this config, which is reasonable evidence it will bite whoever adds the next spec. `tsconfig.app.json:10` already uses `["ES2022", "DOM", "DOM.Iterable"]`.

```suggestion
    "lib": ["ES2022", "DOM", "DOM.Iterable"],
```

.github/workflows/ci.yml
```diff
@@ -113,3 +113,39 @@ jobs:
...
+
+  # Real-browser viewport harness (#954). A separate job, deliberately not a
+  # step inside `verify` above: `verify` is the fast local pre-push gate, and
+  # a Chromium download does not belong in it. This job installs only
+  # Chromium (not the full `playwright install --with-deps` browser set —
+  # firefox/webkit are out of scope, see playwright.config.ts) and runs
+  # `npm run test:e2e`, which builds `dist/` and serves it via `vite preview`
+  # itself (see the config's `webServer`), so no separate build step is
+  # needed here.
+  e2e:
```
s-annam, line 125: **Secondary — filed as #977. The job is not a required status check, so a red run does not block a merge.**

```console
$ gh api repos/offlinecv/OfflineCV/branches/main/protection --jq '.required_status_checks.contexts'
["verify"]
```

AC4 of #954 ("runs in CI as a separate job") is literally met, and the job correctly carries the `merge_group` trigger. But #954's stated purpose is regression *cover*, and today a failing viewport run produces a ✗ nobody is obliged to act on — the merge queue only waits on `verify`.

Not something this diff can fix (it is branch-protection state, not code), which is why it is an issue rather than a change request. #977 also argues the sequencing: letting a new browser job accumulate green history before making it required is the right call, so this is not a claim the PR should have done it.

The CI job itself is well-formed — `timeout-minutes`, `permissions: contents: read`, `forbidOnly` on CI, the trace upload on failure, and the action versions all match the existing `verify` job's conventions.

s-annam (commented):

Rule applied: 0 Blocking findings → APPROVE. Secondary and Nit findings are listed but do not gate.

The harness does what #954 asked, and unusually for a layout PR I could check it rather than trust it — so I re-measured every number in the diff on this branch instead of reading the comments. **All of them are exactly right.** The findings below are almost entirely about the gap between what the assertions *defend* and what their comments *claim* they defend.

## Re-measured on `4795a41` (my own probe, not the PR's numbers)

| Quantity | Claimed | Measured 1280×800 | Measured 1440×900 |
|---|---|---|---|
| Score hero, expanded | 185.5px | **185.5** | **185.5** |
| Score hero, docked | 62px | **62** | **62** |
| Summary heading `y` | 751.5px | **751.5** | **751.5** |
| Fold slack | ~48px | **48.5** | **148.5** |
| `header.sticky` height | 135px (`styles.css`) | **135** | **135** |
| Anchor target `y` after jump | — | **168** (33px clear) | **168** |

`npm run verify` green (fallow clean on all 10 changed files). `npm run test:e2e` green, 6/6, 15.4s.

## Acceptance criteria (#954)

| AC | Result |
|---|---|
| 1 — Playwright installed, configured, runnable via a documented script | **Met** — ran it |
| 2 — spec drops a synthetic fixture, asserts hero height + above-the-fold at 1280×800 | **Met** |
| 3 — hash-anchor targets clear the sticky header | **Met** — and it genuinely validates the hand-measured `scroll-padding-top` in `styles.css:54,60` |
| 4 — separate CI job, does not slow `verify` | **Met literally**; see the note on required checks below |
| 5 — #953's manual-only ACs restated as automated assertions | **Met in form.** I verified the ≤160px target really was retired by the *Clarification (rewrite)* comment on #953, and the 186/62 two-state numbers really are what replaced it. The docked half is the weak one — see the first Secondary |

Fixture gate 3a is **N/A**: this PR touches no fixture. `awesome-cv-resume.pdf` is pre-existing (added in #13, synthetic, PII-free).

**Accurate, and unusually so.** Every checkable claim round-tripped: the `basicSsl()`/`OFFLINECV_DEV_HTTP=1` trap, the docked pill's substring `aria-label`, the 751.5-vs-#955's-810 reconciliation, and the "does not defend" caveats all hold. The body also volunteers the `DOCKED_MAX_PX` problem rather than hiding it, which is the reason that finding is Secondary rather than Blocking — see below.

One correction: the body's *"~48px of margin (6%) … the thinnest assertion in the suite"* is true only as a **percentage**, and only for `desktop-1280`. In pixels the thinnest is the docked ceiling at 8px; and at 1440×900 the same assertion carries 148.5px (16.5%) of slack, so `desktop-1440`'s copy is the loosest assertion in the suite, not the tightest.

## Why I did not push an auto-fix commit

Normally a 0-blocker review with this many small items would land them itself. Not here: the body says this is **layer 1 of a 5-PR stack**, and restoring the one-commit invariant after a fix commit means a force-push, which would orphan the four dependent branches. The only item safe to auto-fix was a comment reword. Not worth it — everything is a suggestion block or a thread instead.

The branch is already exactly one commit, so the invariant is intact and nothing needs collapsing.

## Filed as follow-up issues

- **#975** — `npm run test:e2e` can measure a stale `dist/` and report green (`reuseExistingServer` skips the build step). Thread left **open**: this is a one-line fix in this PR if you'd rather not defer it.
- **#976** — `e2e/**` is outside the ESLint gate. Verified: `npx eslint e2e/` → *"no matching configuration was supplied"*. `eslint.config.js` scopes every block to `src/**/*.{ts,tsx}`, so these 207 new lines are typechecked but unlinted. (`playwright.config.ts` being ignored is the pre-existing `*.config.ts` policy, same as `vite.config.ts` — not a defect.)
- **#977** — decide whether the `e2e` job becomes a required status check. Today `required_status_checks.contexts` is `["verify"]` only, so a red `e2e` does not block the queue. AC4 is literally met; the regression *gate* is advisory until this is settled.

## Nits not worth a thread

- `mobile-375` (`playwright.config.ts:90`) uses `devices["Desktop Chrome"]` with only `viewport` overridden — no `isMobile`, `hasTouch`, mobile UA or DPR. A mobile spec written against it in #959/#960 would be testing desktop Chrome narrowed to 375px: `@media (hover: hover)` and `pointer: fine` still match. `devices["Pixel 5"]` is what the stated purpose wants. It matches zero files today, so nothing is wrong yet.

## Note on a coupling this introduces

`tsconfig.json` now references `tsconfig.e2e.json`, so `npm run build` (`tsc -b && vite build`) typechecks `e2e/**`. A type error in a *spec* file therefore breaks the *build*. I hit this while probing: it surfaces as `Process from config.webServer was not able to start. Exit code: 2` with the tsc output swallowed — worth knowing, and captured in #975.

---

.github/workflows/ci.yml
s-annam: Filed as #977 and resolving this thread — it is branch-protection state, not anything this diff can change, so there is nothing here for you to action in the PR. #977 carries the sequencing argument (let the job build green history before making it required) so the decision gets recorded either way. The job itself is clean; no change requested.

e2e/viewport.spec.ts
s-annam: Fixed in `cacac56` — option 3, comment rewritten.

Options 1 and 2 don't survive contact: tightening to 34 fails today (the measured docked height *is* 62), and dropping the assertion loses real cover for the growth case. So the constant stays at 70 and the comment now says plainly what it is:

> 62px is ALREADY the wrapped two-row state […] this ceiling is calibrated ON the wrap, not against it, and a #960 regression that merely re-wraps the strip would pass here. What it catches is the strip growing past its present two rows.

It also points at #965 for the re-derivation from the true 26px one-row height, rather than inviting someone to guess a smaller number here. The file docblock's "each margin is sized to a specific known regression shape" claim went with it — that was the sentence this ceiling falsified.

e2e/viewport.spec.ts
s-annam: Fixed in `cacac56` — importing from the module rather than keeping a third copy.

```ts
import {
  COLLAPSE_LABEL,
  EXPAND_LABEL,
} from "../src/components/features/scoreToggleLabels.ts";
```

Taking the `src/` → `e2e/` coupling deliberately: the module is two string constants with no imports of its own, so the edge it adds to `tsconfig.e2e.json`'s program costs nothing, and the alternative is a rename that has to move three sites. `npm run typecheck` clean, suite 6/6 green.

e2e/viewport.spec.ts
s-annam: Fixed in `cacac56` — the lower bound is now the sticky header's bottom.

```ts
const headerBox = await page.locator("header.sticky").boundingBox();
expect(headerBox).not.toBeNull();
…
expect(summaryBox!.y).toBeGreaterThanOrEqual(headerBox!.y + headerBox!.height);
```

Re-measured on the collapsed branch: header bottom 135, `summaryY` 751.5, identical at both viewports, so it passes with room. The comment now states the reason a `>= 0` bound certified nothing at the top end.

Took the second half too — the `retries` comment in `playwright.config.ts` no longer calls this the thinnest assertion (see the reply on that thread).

e2e/viewport.spec.ts
s-annam: Fixed in `cacac56` — added a `Card` ceiling rather than only documenting the gap.

Re-measured on this branch, both viewports identically:

```
hero <section> expanded : 185.5
enclosing Card  expanded : 275.5
```

So `CARD_MAX_PX = 300` (+24.5, ~9%, the same rendering-variance margin the other two carry), asserted in the expanded arrival state alongside the hero ceiling. A `scoreCard()` locator resolves it as the hero section's nearest `<section>` ancestor — `Card` renders a `<section>`, which is what makes that stable rather than a class-name guess.

Kept the hero ceilings as they are: they defend the thing #953/#956 actually compressed, and replacing them with the Card figure would lose that. The two now bound the component *and* the budget.

e2e/viewport.spec.ts
s-annam: Fixed in `cacac56` — the widget is docked deliberately before the loop, and the settle no longer counts frames.

The real fix is the dock: `toggle` LOCKS (per the hook's docblock), so one click up front makes the whole loop immune to both the timer and the scroll listener, instead of depending on a focus side effect the test never mentions. Both states carry `#contact` and `#reconstructed-resume` anchors (`CollapsedScoreBar.tsx:170-190`), so nothing in the loop needs the expanded state.

Two supporting changes: the settle now waits for `window.scrollY` to stop moving rather than for one `requestAnimationFrame` — a frame count is a guess about layout timing, the stability check is a fact about it — and the loop asserts the expand toggle is still visible each pass, so if the widget ever does dock mid-loop it fails on that rather than on the reflow three lines later.

playwright.config.ts
s-annam: Fixed in `cacac56` — routing is by directory now, and the default is "runs everywhere".

```ts
{ name: "desktop-1280", testIgnore: MOBILE_DIR, … }
{ name: "desktop-1440", testIgnore: MOBILE_DIR, … }
{ name: "mobile-375",   testMatch:  MOBILE_DIR, … }
```

Verified with throwaway specs rather than assuming — and the first attempt was wrong in a way worth recording. `/^mobile\//` matched nothing:

```
[desktop-1280] › mobile/zz-tmp.spec.ts › tmp mobile      ← wrong project
[desktop-1280] › zz-tmp-new.spec.ts    › tmp new desktop
Total: 10 tests in 3 files
```

Playwright matches a regex `testMatch`/`testIgnore` against the **absolute** path, so an anchored pattern fails *open* — the desktop projects would have quietly swallowed every mobile spec. `MOBILE_DIR = /[\\/]mobile[\\/]/` matches the directory segment instead (either separator, so Windows behaves the same), and that comment is now in the config so the next person doesn't re-derive it:

```
[desktop-1280] › zz-tmp-new.spec.ts › tmp new desktop
[desktop-1440] › zz-tmp-new.spec.ts › tmp new desktop
[mobile-375]   › mobile/zz-tmp.spec.ts › tmp mobile
Total: 9 tests in 3 files
```

A new `e2e/export.spec.ts` now runs on both desktop projects the moment it lands. Throwaway specs deleted; tree clean.

playwright.config.ts
s-annam: Fixed here in `cacac56` rather than deferred — you were right that it belongs on this PR.

```diff
-"test:e2e": "playwright test"
+"test:e2e": "npm run build && playwright test"
```

with `webServer.command` reduced to the preview alone (and `${PORT}`, per the other thread). `vite preview` serves `dist/` off disk per request, so a reused server picks the fresh build up; the build is now unconditional, and a build failure surfaces as a build failure instead of as `Process from config.webServer was not able to start. Exit code: 2` with the tsc output swallowed. The `ci.yml` comment describing the old arrangement moved with it.

**#975 closed as fixed by this PR.** Resolving this thread.

playwright.config.ts
s-annam: Fixed in `cacac56` — took the suggestion's substance, with both figures kept rather than only the relative one:

> the margins here are thin. The thinnest in ABSOLUTE terms is the docked ceiling (62 -> 70, 8px); the thinnest RELATIVE to what it measures is the Summary-above-the-fold assertion at 1280x800 (48.5px of an 800px fold, ~6%). Either is exactly the kind of signal that must surface rather than be re-rolled.

Naming both is what keeps the paragraph true: the 8px figure is the one that actually argues against retries, and dropping it to keep only the percentage would trade one incomplete claim for another.

playwright.config.ts
s-annam: Applied in `cacac56`:

```ts
command: `OFFLINECV_DEV_HTTP=1 npm run preview -- --port ${PORT} --strictPort`,
```

`npm run build` came out of the command entirely (see the `reuseExistingServer` thread), so `PORT` is now the single source for both the bound port and `BASE_URL`. `npm run test:e2e` green, 6/6.

tsconfig.e2e.json
s-annam: Applied in `cacac56`: `"lib": ["ES2022", "DOM", "DOM.Iterable"]`, matching `tsconfig.app.json:10`.

Confirmed it matters on this very round — the throwaway specs I wrote to verify the project-routing fix iterated DOM collections, and that is now a non-event rather than a `TS2488` to decode. `npm run typecheck` clean.

offlinecv-approver (approved):
🕹✅

PR #963, opened 2026-09-18
fix(targeting): restore the all-clear line when every bullet passes (#957)

### LINKED ISSUE #957: [ui] A résumé whose bullets all pass shows no all-clear confirmation
## Problem

A résumé whose bullets **all pass** every check now gets no confirmation of that anywhere on `/`.

`ReconstructedResume`'s `AttentionStrip` used to render a single all-clear line when nothing was flagged:

```
All 12 bullets pass every check.
```

#953 deleted `AttentionStrip` and moved its triage content into `TargetingSection` / `TargetingTriageRow`. That move carried across the *gap* reporting — flagged bullets, missing contact fields — but not the all-clear case: `TargetingTriageRow`'s `BulletSegment` returns `null` when `flagged === 0`, and `TargetingSection` only renders the triage row when `hasTriage` is true. So the clean-résumé branch has no renderer at all.

This was noticed while reviewing #956 (see `src/components/features/TargetingSection.tsx`, the `hasTriage` guard, and `TargetingTriageRow.tsx`'s `BulletSegment`). It is not a regression introduced *by* a bug — it is a behaviour that was dropped silently during a refactor, which is why nothing failed.

## Why it matters

The product's core loop is "drop a PDF, see what parsed, fix what's wrong". Absence of a warning is not the same signal as confirmation: a user whose bullets are clean currently cannot distinguish "checked, all good" from "not checked" / "the checker didn't run". The all-clear line was the only thing that closed that loop, and it also carried the bullet **count**, which told the user how much was actually examined.

## Proposed fix

Give the clean case a renderer. Options, roughly in increasing cost:

1. Render the all-clear line in `TargetingSection`'s summary row when `bullets.length > 0 && !hasBulletGap` — cheapest, and it reuses the summary the disclosure already has.
2. Let `TargetingTriageRow` handle the clean case and render the disclosure whenever there are bullets at all. Note this interacts with the empty-disclosure guard fixed in #956 — that guard exists precisely because a disclosure with nothing to say is worse than no disclosure, so any version of this must keep the "nothing to say at all" case rendering nothing.
3. Surface it in the score widget instead, next to the Specificity/Structure dimensions, since those already quantify the same bullet pool.

Option 1 is the smallest change that restores the information.

## Acceptance criteria

- [ ] A résumé with bullets where none is flagged **and no contact field is missing** shows an explicit all-clear confirmation naming the bullet count. (Narrowed during review of #963: the deleted `AttentionStrip` fired its all-clear only when *both* checks were clean, and a contact-gap résumé keeps its triage headline on the summary row instead — see the comment below.)
- [ ] A résumé with **no** bullets and nothing else to say still renders nothing (no empty disclosure — the #956 guard stays intact).
- [ ] A test covers the all-clear branch; today no test asserts it, which is how it was dropped unnoticed.

## Provenance

Dropped in #953 / PR #956. Verified on the `gh-953` branch — the code this describes is not on `main` yet, so line numbers are omitted deliberately (the squash moves them).

s-annam:
Restores the all-clear confirmation from #957: a résumé whose bullets all pass now says so, naming the count.

**Layer 2 of 5. Depends on #962.**

## Why it was missing

#953 deleted `AttentionStrip` and moved its triage content into `TargetingSection` / `TargetingTriageRow`. That move carried the *gap* reporting across but not the all-clear case, so the clean-résumé branch had no renderer at all. Nothing failed, because nothing asserted it — which is why a test is an acceptance criterion here rather than an extra.

Absence of a warning is not confirmation. A user with clean bullets could not tell "checked, all good" from "the checker never ran", and the line also carried the bullet **count**, which is what says how much was actually examined.

## What ships

- `TargetingSection` renders `All N bullets pass every check` in the summary row when bullets exist and nothing is flagged.
- The empty-disclosure guard is now derived from the children (`hasBody = hasTriage || titles.length > 0 || hasSkillGuidance`) rather than inferred, so adding a fourth child cannot silently leave it stale.
- When there is nothing to open onto, the confirmation renders as an `InlineResult tone="success"` strip instead of a `Disclosure`.

## Scope call worth reviewing

The issue's literal condition was `bullets.length > 0 && !hasBulletGap`. This ships `!hasTriage && bullets.length > 0` — i.e. it stays silent when bullets are clean but **contact** has a gap. That matches what the deleted `AttentionStrip` actually did: its all-clear fired only when both checks were clean. The literal condition would have added new text to a case the old component never spoke in.

## Verification

`npm run typecheck` · `npm run lint` · `TargetingSection.test.tsx` 10/10, `TargetingTriageRow.test.tsx` 7/7 · full vitest 6747 passed / 10 skipped.

## Adversarial review

Reviewed adversarially before this PR existed (2 rounds, `clean: true` at the end). **One blocking finding, on this layer, now fixed:**

- **The widened guard reopened #956's empty disclosure** for exactly one input: `titles.length === 0 ∧ !hasSkillGuidance ∧ !hasTriage ∧ bullets.length > 0` — which is the all-clear branch itself. All three body children self-hide on precisely those inputs (`TargetingTriageRow` is gated on `hasTriage`; `RolesPanel` returns null without titles; `SkillTermGuidance` returns null with all buckets empty), so the user got a `<details>` promising "All N bullets pass every check" and opening onto ~24px of blank card. #956's rule — a disclosure with nothing to say is worse than no disclosure — is about the **body**, so the fix drops the disclosure, not the confirmation. Proven by fail-before: with the old guard restored the new test fails `expected <details …> to be null`.
- The new test asserts `details > div` is non-empty — the assertion whose absence let this through. The first version asserted only the summary text, which green-lit an empty body.

Nits fixed in the same pass:

- **Grammar.** "All 1 bullet **passes**" / "All 2 bullets pass" — the singular was inherited verbatim from the pre-#956 string, and `TargetingTriageRow`'s own docblock documents this exact English trap one function over. One `allClearLine` definition now feeds both render paths so they cannot drift.
- **House pattern.** The standalone line was a hand-rolled `<p>` + coloured `<span>`; it is now the `InlineResult tone="success"` the design system already owns.

Known imprecision, stated rather than hidden: `hasBody`'s third term is `hasSkillGuidance`, but `SkillTermGuidance` also renders on a skills-**order** finding. Unreachable in practice (a finding implies titles, which term 2 covers) and `main` dropped the same content, so not a regression — the docblock says so instead of claiming a property that does not hold.

Closes #957


src/components/features/TargetingSection.tsx
```diff
@@ -99,13 +118,53 @@ export function TargetingSection({
...
+  // they cannot drift. Singular takes "passes": `TargetingTriageRow`'s own
+  // docblock documents this same English trap one function over ("1 bullet
+  // NEEDS attention" vs "2 bullets NEED"), and the pre-#956 `AttentionStrip`
+  // string this restores got it wrong.
+  const allClearLine =
+    bullets.length === 1
+      ? "All 1 bullet passes every check"
+      : `All ${bullets.length} bullets pass every check`;
+
+  const allBulletsClear = !hasTriage && bullets.length > 0;
```
Vaishnavi1709, line 137: **Scope call: this closes #957 with AC1 as literally written unmet.**

`allBulletsClear = !hasTriage && bullets.length > 0` is byte-faithful to the deleted `AttentionStrip` — I recovered it at `9283c83^:ReconstructedResume.tsx:227` and it reads `if (!hasBulletGap && !hasContactGap) { if (total === 0) return null; ... }`. So your `## Scope call worth reviewing` is accurate and **I endorse the decision**: the issue's literal `!hasBulletGap` would have to share the summary row with the triage headline, which #957 did not consider.

What is left open is the bookkeeping. AC1 says "A résumé with bullets where none is flagged shows an explicit all-clear confirmation naming the bullet count" with no contact qualifier, and `Closes #957` will close it. I confirmed the gap against the built component — 2 clean bullets plus one gated contact row renders:

```
Targeting & improvements· 1 contact field missing
```

with `pass every check` absent from the whole subtree. Before merge, either narrow AC1 on #957 or open a follow-up for the contact-gap case. I have deliberately not filed one, since you are coming back for the restack and may prefer to fold it in.

src/components/features/TargetingSection.tsx
```diff
@@ -99,13 +118,53 @@ export function TargetingSection({
...
+
+  // Nothing to say at all, and nothing to open onto.
+  if (!hasBody && !allBulletsClear) return null;
```
Vaishnavi1709, line 140: Nothing pins this deliberate silence. The two new tests cover the all-clear's two render paths (AC3 ✅), but the case above — the behaviour you consciously chose *not* to have — has no test, so a later change to this line flips it with the suite still green.

The cheap durable fix, mirroring the existing harness:

```ts
it("stays silent about bullets when they all pass but contact has a gap", () => {
  // The deliberate narrowing (#957 scope call): the deleted `AttentionStrip`
  // fired its all-clear only when BOTH checks were clean, and `!hasTriage`
  // keeps that. Without this test nothing would catch a widening to
  // `!hasBulletGap`.
  const el = render({
    bullets: allPassing,
    contactMissing: [
      { key: "linkedin_url", label: "Professional profile",
        group: "links", value: "", gated: true, reason: "absent" },
    ],
  });
  expect(el.textContent).not.toContain("pass every check");
  expect(el.querySelector("summary")!.textContent)
    .toContain("1 contact field missing");
});
```

src/components/features/TargetingSection.tsx
```diff
@@ -84,12 +88,27 @@ export function TargetingSection({
...
+  // is unreachable in practice — `computeSkillsOrderFinding` returns undefined
+  // when the titles tokenize to nothing (`skills-order.ts`), so a finding
+  // implies titles, which term 2 already covers — and the one contrived path
+  // (an `applied` reorder surviving a cleared headline) rendered nothing on
+  // `main` either, so it is not a regression. Deriving the guard from the
+  // children at all is the point: adding a fourth child cannot silently leave
+  // it stale, which is how #957's first pass went wrong. It widened the guard to `bullets.length === 0` so the all-clear
+  // summary could render, and thereby reopened #956's empty disclosure for the
+  // one input where the all-clear fires and every child still self-hides
+  // (no titles, no skill guidance, no triage, bullets all passing).
```
Vaishnavi1709: **This rationale narrates two branch states a reader cannot verify, and they read as contradicting each other.**

Line 91 says a bare `bullets.length === 0` term "used to sit here". Line 107 says #957's first pass "widened the guard to `bullets.length === 0`". Both describe intermediate states of review branches; neither is in `main`. The pre-#956 guard was:

```ts
if (titles.length === 0 && !hasSkillGuidance) return null;   // fd25862^:93
```

— no bullets term at all. So a reader who checks finds neither claim, and is left with two sentences that appear to disagree about which direction the term moved.

Separately, "adding a fourth child cannot silently leave it stale" is contradicted by the next sentence, which concedes term 3 does not cover `SkillTermGuidance`'s skills-order path. `hasBody` is still a hand-maintained disjunction — the real improvement is that its terms now *read* as the children's own null conditions.

The suggestion rewraps line 107 (121 chars, finding 6) and attributes each claim to the branch it belongs to:

```suggestion
  // `main` either, so it is not a regression. Deriving the guard from the
  // children at all is the point: a fourth child must still be added to this
  // disjunction by hand, but each term now restates a child's own null
  // condition, so a stale one is visible rather than inferred. #957's first
  // pass instead widened the guard with a `bullets.length === 0` term so the
  // all-clear summary could render, and thereby reopened #956's empty
  // disclosure for the one input where the all-clear fires and every child
  // still self-hides (no titles, no skill guidance, no triage, bullets all
  // passing).
```

src/components/features/TargetingSection.tsx
```diff
@@ -99,13 +118,53 @@ export function TargetingSection({
...
+  const allBulletsClear = !hasTriage && bullets.length > 0;
+
+  // Nothing to say at all, and nothing to open onto.
+  if (!hasBody && !allBulletsClear) return null;
+
+  // Something to say, but no body behind it: the all-clear line is the whole
+  // message, so it renders as a plain line rather than as a `Disclosure` whose
+  // triangle opens onto ~24px of blank card. #956's rule — "a disclosure with
+  // nothing to say is worse than no disclosure" — is about the BODY, so the
+  // fix is to drop the disclosure, not the confirmation (#957 AC2).
```
Vaishnavi1709: **"a plain line" is not what this branch renders.**

`InlineResult tone="success"` resolves to `rounded border border-feedback-success-border bg-feedback-success-bg p-3` — a bordered, filled strip roughly twice the height of the bare `<p className="text-sm font-medium text-feedback-success-text">` that `AttentionStrip` used for the same message.

Reaching for the primitive is the right call and I am not asking you to change the JSX: this branch only renders on a résumé where every other child of this section self-hides, so the extra height costs nothing. It is the prose that is wrong — here and in the test comment at `TargetingSection.test.tsx` ("The confirmation renders as a plain line instead"). The PR body's `## What ships` already describes it correctly.

Worth noting alongside: the comment above says one `allClearLine` definition means the two paths "cannot drift". That holds for the copy, not the treatment — chrome-less `<span>` on the summary row, bordered box standalone.

```suggestion
  // Something to say, but no body behind it: the all-clear line is the whole
  // message, so it renders as a standalone `InlineResult` strip rather than as
  // a `Disclosure` whose triangle opens onto ~24px of blank card. #956's rule
  // — "a disclosure with nothing to say is worse than no disclosure" — is
  // about the BODY, so the fix is to drop the disclosure, not the confirmation
  // (#957 AC2).
```

Vaishnavi1709 (changes requested):
The change itself is sound — I found **no blocking defect in the code**. The one Blocking item is mechanical: this PR cannot merge into its own base right now.

Reviewed against #957's acceptance criteria first, then the diff, then the description. Verdict rule applied: ≥1 Blocking → `REQUEST_CHANGES`.

## Blocking

**1. The PR is `CONFLICTING` against its own base — the stack is stale.**

`mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. Reproduced locally:

```
$ git merge-tree --write-tree origin/gh-01-954-viewport-harness HEAD
CONFLICT (content): Merge conflict in .github/workflows/ci.yml
CONFLICT (add/add): Merge conflict in e2e/viewport.spec.ts
CONFLICT (content): Merge conflict in package.json
CONFLICT (add/add): Merge conflict in playwright.config.ts
CONFLICT (add/add): Merge conflict in tsconfig.e2e.json
```

None of the five is in `src/` — all five belong to #962. The cause is stack drift: this branch carries `4795a41`, a **pre-review copy** of #962's commit, while `origin/gh-01-954-viewport-harness` has moved on to `9ef9433`. Merge-base is `fd25862`, so git sees an add/add on three files.

Two consequences worth naming:

- **Resolution trap.** Take the **base** side for all five. The head copy predates #962's review — `git diff 9ef9433 4795a41` shows the head is missing a 56-line `eslint.config.js` block and carries a 124-line-longer `viewport.spec.ts` that #962 has since simplified. Resolving toward head would silently revert that review.
- **The GitHub diff misrepresents this PR.** It currently shows 11 files / +616, so a reviewer opening #963 sees #962's entire Playwright harness inside it. The actual change under review is 2 files / +120 −15 (`a3d2c2c`).

Fix is a restack onto current `origin/gh-01-954-viewport-harness`, not a code change. Re-request and I'll approve.

## Secondary

**2. The `!hasTriage` scope call closes #957 with AC1 as written unmet.** — inline on `TargetingSection.tsx`

Thank you for surfacing this in `## Scope call worth reviewing` rather than letting me find it; I formed the same finding independently from the issue before reading the body, and your account of it is accurate. **I verified the claim and I endorse the call** — `9283c83^:ReconstructedResume.tsx:227` gives the deleted `AttentionStrip` as `if (!hasBulletGap && !hasContactGap) { if (total === 0) return null; … }`, so `!hasTriage && bullets.length > 0` is byte-faithful to it. Widening to the issue's literal `!hasBulletGap` would also have to share the summary row with the triage headline, which the issue did not think through.

What still needs disposing of: #957's AC1 reads "A résumé with bullets where none is flagged shows an explicit all-clear confirmation naming the bullet count", with no contact qualifier, and `Closes #957` will close it. Confirmed against the built component — 2 clean bullets + 1 gated contact row renders `Targeting & improvements · 1 contact field missing` and the string `pass every check` appears nowhere in the output. Either narrow AC1 on the issue before merging, or open a follow-up for the contact-gap case. I have not filed one myself because you are coming back for the restack and may prefer to fold it in.

**3. Nothing pins the deliberate silence.** — inline

The two new tests cover the all-clear's two render paths, which satisfies AC3. Neither covers the case finding 2 is about, so the behaviour you deliberately chose is the one thing here a future change can flip with every test still green. A third test asserting `not.toContain("pass every check")` for clean-bullets-plus-contact-gap would make the scope call durable instead of a comment.

**4. The `hasBody` rationale narrates two unverifiable branch states.** — inline, with a suggestion

Line 91 says a bare `bullets.length === 0` term "used to sit here"; line 107 says #957's first pass "widened the guard to `bullets.length === 0`". Both describe intermediate states of review branches, neither of which is in `main` — the pre-#956 guard was `if (titles.length === 0 && !hasSkillGuidance) return null;` (`fd25862^:TargetingSection.tsx:93`), with no bullets term at all. A reader who checks will find neither claim and read the two as contradicting each other.

The block also overclaims: "adding a fourth child cannot silently leave it stale" is immediately followed by conceding term 3 does not cover `SkillTermGuidance`'s skills-order path. `hasBody` is still a hand-maintained disjunction; what actually improved is that the terms now *read* as the children's null conditions.

**5. The comments say "plain line"; the code ships a bordered strip.** — inline, with a suggestion

`InlineResult tone="success"` is `rounded border border-feedback-success-border bg-feedback-success-bg p-3` — roughly twice the height of the bare `<p className="text-sm font-medium text-feedback-success-text">` this restores. The PR body's `## What ships` is accurate ("an `InlineResult tone="success"` strip"); it is the in-code comment at line 142 and the test comment that say "plain line". Using the primitive is the right call, and this only ever renders on a résumé where nothing else in the section does — so it is the prose to fix, not the JSX.

Related: the comment says one `allClearLine` definition means the two paths "cannot drift". True of the copy, not of the treatment — chrome-less `<span>` on the summary row, bordered box standalone.

## Nits

**6.** Line 107 is 121 characters, unwrapped, in a block otherwise wrapped at ~78. Folded into the suggestion on finding 4.

**7.** The restored string drops the original's trailing period — `All 12 bullets pass every check.` → `…every check`. Correct in the summary chip (`· All 12 bullets pass every check`); just noting it is not a byte-faithful restore.

## Gates

| Gate | Result |
|---|---|
| `tsc -b --noEmit` | ✅ clean |
| `eslint .` | ✅ clean (exit 0) |
| `TargetingSection.test.tsx` + `TargetingTriageRow.test.tsx` | ✅ 17/17 |
| `ResultDetail.test.tsx` | ✅ 23/23 |
| `fallow audit --base origin/main` | ✅ no issues in 11 changed files |
| 3a fixture PII | n/a — no fixture touched |
| 3b design-system tier | ✅ `InlineResult` via the `@design-system` barrel, no deep import, no raw `<button>`/modal |
| 3c semantic tokens | ✅ no hex, no raw palette, no manual `dark:` (grep hits were `#956`/`#957` issue refs) |
| 3e skill/script command bugs | n/a — no script or skill file in `a3d2c2c` |
| 3f description accuracy | ✅ accurate, and discloses its own scope call — see finding 2 |

Component is 206 lines but only **127** are code (79 comment); the ~200 LOC guideline is not breached.

**AC checklist (#957)** — AC1 ⚠️ met except for the contact-gap case (finding 2) · AC2 ✅ `!hasBody && !allBulletsClear → null`, pinned by the existing test · AC3 ✅ two tests.

Reviewed at `a3d2c2c0b5774a2493537d1442b535b296de5ea1`. Diff reviewed as `a3d2c2c` alone — the harness files in the GitHub diff are #962's and were left to that review.

---

src/components/features/TargetingSection.tsx
s-annam: Took your first option: AC1 on #957 now reads "…where none is flagged **and no contact field is missing**…", with a `**Clarification (scope narrow)** —` breadcrumb citing `9283c83^:ReconstructedResume.tsx:227` (https://github.com/offlinecv/OfflineCV/issues/957#issuecomment-5735400273). `Closes #957` now matches what ships. The narrowing is pinned by the test from your next thread (`aa56bb5`).

src/components/features/TargetingSection.tsx
s-annam: Fixed in `aa56bb5`: added `stays silent about bullets when they all pass but contact has a gap`, as you proposed (`group: "link"`, since `ContactGroup` has no `"links"`). Widening the guard to `!hasBulletGap` now fails it.

src/components/features/TargetingSection.tsx
s-annam: Fixed in `aa56bb5`: applied your suggestion. I also rewrote the line-91 sentence ("A bare `bullets.length === 0` term used to sit here"), which made the same unverifiable branch-state claim; it is now a counterfactual ("A guard keyed on bullets would open…"). The matching narrative in the first all-clear test comment now says what `main` did: it rendered nothing.

src/components/features/TargetingSection.tsx
s-annam: Fixed in `aa56bb5`: applied your suggestion. The test comment now says "standalone `InlineResult` strip", and the "cannot drift" note is scoped to the copy, naming the two treatments (chrome-less span on the summary row, bordered `InlineResult` standalone). JSX unchanged.

src/components/features/TargetingSection.test.tsx
```diff
@@ -171,8 +171,79 @@ describe("TargetingSection", () => {
...
+        wordCount: 9,
+      },
+    ];
+    const el = render({
+      bullets: allPassing,
+      contactMissing: [
+        { key: "linkedin_url", label: "Professional profile", value: "", group: "link", gated: true },
+      ],
+    });
+    expect(el.textContent).not.toContain("pass every check");
```
Samhit21: [Suggestion]: This assertion cannot fail for its own fixture, and the widening it was written to catch doesn't flip it either.

Two separate problems, both reproduced against this head.

**1. The fixture is singular, the assertion string is plural.** One bullet renders `All 1 bullet passes every check` — and `"passes every check"` does not contain the substring `"pass every check"`. So this line is vacuous: I patched the component into the state it exists to forbid (both the guard widened *and* the summary ternary reordered, so the all-clear genuinely renders) and the test still passed, with `el.textContent` containing `· All 1 bullet passes every check` the whole time.

**2. Even with the plural fixed, widening the guard alone still doesn't fail it.** Re-running the same fixture with two clean bullets, so the string matches:

| component state | all-clear rendered? | this test |
|---|---|---|
| as shipped | no | passes ✅ |
| `allBulletsClear = !hasBulletGap && …` | no | **passes** ← the mutation this test is for |
| summary ternary checks `allBulletsClear` first | no | passes |
| both together | **yes** | fails ✅ |

The reason is that `allBulletsClear` is not what enforces the silence. With a contact gap, `hasTriage` is true, so `summary`'s ternary takes the triage branch before `allBulletsClear` is consulted — and `hasBody` is true for the same reason, so the standalone `InlineResult` path is unreachable too. Widening `!hasTriage` → `!hasBulletGap` is therefore invisible on this input regardless.

That makes the reply on the thread above — "Widening the guard to `!hasBulletGap` now fails it" — not hold, and @Vaishnavi1709's finding 3 substantively still open: the deliberate narrowing is enforced by the ternary's ordering, and nothing asserts that ordering.

**Fix.** Two clean bullets rather than one (so the copy matches), and assert the mechanism rather than only the absence:

```ts
expect(el.textContent).not.toContain("bullets pass every check");
expect(el.querySelector("summary")!.textContent).toContain("1 contact field missing");
// …and the thing that actually decides it: with a contact gap the summary is
// the triage headline, never the all-clear, whatever `allBulletsClear` says.
expect(el.querySelector("summary")!.textContent).not.toContain("expected skills");
```

The last line is what fails under the reorder; the first two fail under a widening that also reaches the render. Worth keeping the singular case as its own positive test (`All 1 bullet passes every check` on a clean résumé) since that copy is now load-bearing.

Samhit21 (commented):
## PR Review: fix(targeting): restore the all-clear line when every bullet passes (#957)

### Summary

Reviewed at `aa56bb5`. The change itself is right and the scope call is well argued — I verified `!hasTriage && bullets.length > 0` against the deleted `AttentionStrip` and agree it is faithful, and #957's AC1 has been narrowed on the issue with a breadcrumb, so `Closes #957` now matches what ships.

One finding, inline: **the test added for @Vaishnavi1709's finding 3 does not pin the behaviour it was added to pin**, and the thread reply saying the widening "now fails it" doesn't hold. Details and a reproduction matrix are on the inline comment. Nothing blocking from me — the shipped behaviour is correct; it is the guard against a future change that is missing.

### Verified on this head

- **@Vaishnavi1709's Blocking is resolved.** The restack worked: 2 files / +150 −15, one commit, base `main`, 0 behind, `mergeable: MERGEABLE`. None of #962's Playwright files are in the diff any more, so the PR now shows what it actually changes.
- **The empty-disclosure fix is genuinely pinned.** Forcing `hasBody = true` fails two tests, including `expect(el.querySelector("details")).toBeNull()`. That is the assertion whose absence let the first pass reopen #956, and it has teeth.
- **The other three fixes landed as described** — AC1 narrowed on #957 (`issuecomment-5735400273`), the `hasBody` docblock no longer narrates unverifiable branch states, and the "plain line" prose now says "standalone `InlineResult` strip".
- **Gates:** `TargetingSection` + `TargetingTriageRow` + `ResultDetail` — 41/41 pass. `eslint` clean. CI green on this head: `verify`, `fallow`, `e2e`, Cloudflare Pages.
- The singular/plural copy is right (`All 1 bullet passes every check` / `All 2 bullets pass every check`), which is what makes the vacuous assertion in the inline comment easy to miss.

### Verdict

**Action: COMMENT.** The standing `REQUEST_CHANGES` is @Vaishnavi1709's and hers to clear — worth noting that on the evidence her finding 3 is not yet satisfied, so the re-review is not a formality. One test change closes it.

src/components/features/TargetingSection.test.tsx
s-annam: Fixed in `ffc01db` (branch collapsed to one commit). Both problems reproduced here.

**1. Vacuous plural.** The assertion is now `not.toMatch(/pass(es)? every check/)`, so it catches the singular and the plural. The fixture has two bullets, so the plural copy is the string under test. The singular positive case you asked for was already there: "shows an all-clear confirmation…" asserts `All 1 bullet passes every check` on the standalone strip.

**2. Mechanism.** I added your `summary` assertions: the summary contains `1 contact field missing` and does not contain `expected skills`. I mutation-tested against this head:

| component state | new test |
|---|---|
| as shipped | passes |
| guard widened to `!hasBulletGap` only | passes |
| ternary reordered (`allBulletsClear` first) only | passes |
| both | **fails** (all three assertions) |

One correction to the suggestion: the reorder alone does not flip the last line. With a contact gap, `allBulletsClear` (`!hasTriage && …`) is false, so reordering the ternary still falls through to the triage branch. Two redundant guards enforce the narrowing: the `!hasTriage` term and the ternary's order. Each one alone keeps this input silent, so a single mutation doesn't change anything the user sees, and no rendering test can catch it. The test now fails at the point where the user would see a change, and the test comment says that instead of the earlier claim that widening alone fails it. You were right that my earlier reply on the thread above overclaimed.

s-annam:
## Update for reviewers: where this PR stands at `ffc01db`

The branch is now **one commit**, `ffc01db`, rebased onto current `main`. It is `MERGEABLE`, and CI is green (`verify`, `e2e`, `fallow`, Cloudflare Pages). All review threads are resolved. The only thing blocking the merge is a fresh approval.

### What each review round changed

**@Vaishnavi1709's review (CHANGES_REQUESTED)**
- **Blocking: conflict against base.** Fixed. The branch is rebased onto `main` and merges cleanly.
- **Scope of #957's AC1.** The all-clear still fires only when *both* the bullet check and the contact check are clean, which is the same rule the deleted `AttentionStrip` used. #957's AC1 was narrowed on the issue to match, with a clarification breadcrumb, so `Closes #957` now matches what ships.
- **Nothing tested the contact-gap silence.** Added a test for it (see the next section for how it was fixed).
- **Rationale comments.** The `hasBody` docblock no longer describes branch states a reader cannot check. The all-clear comment no longer calls the standalone `InlineResult` "a plain line": it is a bordered strip, and the comment now says so.

**@Samhit21's review: the contact-gap test could not fail**
- The old assertion checked for `pass every check`. With a one-bullet fixture the component renders `passes every check`, so the assertion could never trip.
- The test now uses **two passing bullets** plus a missing contact field, so the plural wording is the one exercised. It checks three things:
  - neither form of the line renders (`/pass(es)? every check/`),
  - the summary names the contact gap (`1 contact field missing`),
  - the summary is not the all-clear label (`expected skills`).
- **Correction to the suggested mutation check.** Two guards independently keep the line silent here: the `!hasTriage` term in `allBulletsClear`, and the summary ternary checking `hasTriage` first. Either one alone is enough. I mutated each against the new test:

  | Mutation | New test |
  |---|---|
  | widen `allBulletsClear` (drop `!hasTriage`) | passes |
  | reorder the summary ternary | passes |
  | both | **fails** |

  So no single-guard regression shows up on screen, and no test can catch one. The test pins what renders, not either guard alone, and the test comment says this. My earlier thread reply claimed widening the guard alone would fail the test. That was wrong, and the reply on the thread corrects it.

### Gates at `ffc01db`
- `npm run typecheck`: clean.
- `npm run test`: 6791 passed.
- Before and after the squash, the tree hash is the same (`cad7b4532dc9…`), so the squash changed history only, not file contents.

src/components/features/TargetingSection.tsx
```diff
@@ -18,12 +18,16 @@
  *
  * The summary row displays:
  * 1. "Targeting & improvements" with bullet/contact triage callouts when issues
- *    are detected, or "Targeting — your role and its expected skills" when clean.
+ *    are detected; "Targeting — your role and its expected skills" plus an
+ *    explicit all-clear line naming the bullet count when there are bullets
+ *    and none is flagged (#957 — the confirmation `AttentionStrip` used to
+ *    carry, dropped silently when #953 moved triage content in here); or the
+ *    plain label alone when there is nothing to report either way.
```
Vaishnavi1709: Nit: this item now covers only half the new behaviour. The all-clear doesn't appear when a contact field is missing, and when there's no body it renders as a standalone strip, not on the summary row. A reader of the docblock would expect both to show the summary label.

```suggestion
 *    are detected; "Targeting — your role and its expected skills" plus an
 *    explicit all-clear line naming the bullet count when there are bullets,
 *    none is flagged and no contact field is missing (#957 — the confirmation
 *    `AttentionStrip` used to carry, dropped silently when #953 moved triage
 *    content in here); or the plain label alone when there is nothing to
 *    report either way. With nothing to open onto, the all-clear renders as a
 *    standalone `InlineResult` strip instead of a disclosure.
```

src/components/features/TargetingSection.tsx
```diff
@@ -99,13 +121,55 @@ export function TargetingSection({
...
   );
 
+  // Mirrors the old `AttentionStrip`'s all-clear line, which fired only when
+  // BOTH the bullet and contact checks were clean — a résumé with a contact
+  // gap but no flagged bullets stays silent about bullets here too, same as
+  // it always has (`TargetingTriageRow`'s bullet segment already self-hides
+  // in that case).
+  // One definition, two render paths (summary chip and standalone strip), so
+  // the copy cannot drift — the treatment does differ: a chrome-less span on
+  // the summary row, a bordered `InlineResult` standalone. Singular takes "passes": `TargetingTriageRow`'s own
```
Vaishnavi1709: Nit: this line runs well past the column the rest of the file wraps at.

```suggestion
  // the copy cannot drift — the treatment does differ: a chrome-less span on
  // the summary row, a bordered `InlineResult` standalone. Singular takes
  // "passes": `TargetingTriageRow`'s own
```

Vaishnavi1709 (dismissed):
Approving. This restores #957's all-clear without reopening #956's empty disclosure. The `hasBody` / `allBulletsClear` split handles the three cases correctly: say nothing, show a standalone strip, or show a summary with a body. The tests pin all three, including the contact-gap silence.

**Verdict rule:** 0 Blocking → `APPROVE`. Two Nits are posted inline as suggestion blocks. This is a contributor branch, so I pushed nothing. A fix commit would break the one-commit invariant, and I won't force-push someone else's branch.

## Acceptance criteria (#957, as narrowed)
- [x] AC1: bullets present, none flagged, and no contact gap → an all-clear naming the count. It goes on the summary row when there is a body (`TargetingSection.tsx:166-172`), otherwise a standalone `InlineResult` (`:147-157`). A contact gap keeps the triage headline (`:159`).
- [x] AC2: no bullets and nothing else → `null` (`:143`). The existing `renders nothing…` test still covers it.
- [x] AC3: covered by the three new or rewritten tests (standalone strip, summary with a body, contact-gap silence).

## Checks
- `npx vitest run src/components/features/` → 71 files, 806 tests passed. `tsc -b --noEmit` is clean, and `eslint` on both files is clean. CI is green at `ffc01db` (verify, e2e, fallow, Pages).
- Design system: `InlineResult` comes from `@design-system`, and there are no raw elements. Style: `text-feedback-success-text` is a semantic token (`theme.css:79`), and there are no hex or palette classes. Skipped: fixture PII (no fixtures changed) and skill/script review (none changed).
- I checked the redundant-guard claim from the update comment by reading the code. With a contact gap, `hasBody` is true, so the standalone strip can't be reached, and the summary ternary tests `hasTriage` first. Widening `allBulletsClear` alone therefore renders nothing new, as the comment says.

## Nits (non-blocking)
- The module docblock doesn't mention the contact-gap narrowing or the standalone-strip path (inline suggestion).
- One overlong comment line in the `allClearLine` block (inline suggestion).
- Description: the header line `Layer 2 of 5. Depends on #962` is stale, because #962 is merged and the branch is rebased onto `main`. The `## Verification` counts (10/10, 6747) predate the later test addition. The update comment on the thread corrects both. It's worth editing the body anyway, since the body is what readers rely on later.
- `TargetingSection.tsx` is now 211 LOC. That is slightly over the ~200 guideline, almost all of it rationale comments. The `hasBody` comment (22 lines) also narrates the PR's own review history ("#957's first pass instead widened…"), which fits better in the commit message. It can be trimmed later. Not worth another round.

Reviewed at `ffc01dbe51a1dd7ce05de6f7cd0e467c58e47a8e`.

---

src/components/features/TargetingSection.tsx
s-annam: Applied in `3e57c4f` — the docblock now names the contact gate (`hasTriage` includes `hasContactGap`, so a contact gap suppresses the all-clear) and the standalone `InlineResult` path.

src/components/features/TargetingSection.tsx
s-annam: Applied in `3e57c4f` — rewrapped, and reflowed the two lines after it so the comment keeps the file's column.

src/components/features/TargetingSection.tsx
```diff
@@ -84,12 +90,30 @@ export function TargetingSection({
...
+  // (an `applied` reorder surviving a cleared headline) rendered nothing on
+  // `main` either, so it is not a regression. Deriving the guard from the
+  // children at all is the point: a fourth child must still be added to this
+  // disjunction by hand, but each term now restates a child's own null
+  // condition, so a stale one is visible rather than inferred. #957's first
+  // pass instead widened the guard with a `bullets.length === 0` term so the
+  // all-clear summary could render, and thereby reopened #956's empty
+  // disclosure for the one input where the all-clear fires and every child
+  // still self-hides (no titles, no skill guidance, no triage, bullets all
+  // passing).
```
offlinecv-approver: Nit (non-blocking): the rest of this comment is PR history (what #957's first pass did), which the commit message already carries and which will stop meaning anything after the squash. The comment is complete without it:

```suggestion
  // condition, so a stale one is visible rather than inferred.
```

offlinecv-approver (dismissed):
**APPROVE.** Rule applied: 0 Blocking → APPROVE; Nits are non-blocking. Reviewed head `64bdd9d`.

Reviewed against #957 (with its scope-narrow clarification) before reading the description.

## Acceptance criteria (#957)
- [x] **AC1** — all-clear naming the count when no bullet is flagged and no contact field is missing: `allBulletsClear = !hasTriage && bullets.length > 0` feeds the summary chip (with a body) or the standalone `InlineResult` strip (without one). Pinned by `shows an all-clear confirmation naming the count…` and `renders the all-clear inside the disclosure summary…`. The contact-gap narrowing is pinned by `stays silent about bullets when they all pass but contact has a gap`.
- [x] **AC2** — no bullets and nothing else: `!hasBody && !allBulletsClear` → `null`. The existing sibling test still asserts empty output, and #956's empty-`<details>` guard holds: `hasBody` false never reaches `Disclosure`.
- [x] **AC3** — the all-clear branch is tested in both render paths. The body-non-empty assertion (`details > div`) is the one that would have caught #956 regressing.

I checked the claims in the comments. `needsAttention` (`group-bullets.ts:47`) is exactly the three checks (metric / verb / length), so "passes every check" matches the code. `computeSkillsOrderFinding` returns `undefined` on empty title tokens (`skills-order.ts:148`), so the looser `hasSkillGuidance` term is unreachable-loose, as the comment says.

## Gates
- `/code-review` (high): no findings.
- Style/tokens: only semantic tokens (`text-feedback-success-text`, `InlineResult tone="success"`). No hex values, palette classes or `dark:`. The grep hits are `#957` issue refs.
- Design system: reuses `InlineResult` through `@design-system`. No raw interactive elements and no new component files.
- `npx vitest run` TargetingSection + TargetingTriageRow + ReconstructedResume suites: 44/44. `npm run typecheck` clean. `eslint` on the changed files clean.
- `fallow audit --base origin/main`: no issues in the 2 changed files.
- Fixture PII: skipped, no fixtures touched.
- CI: `e2e` and Cloudflare pass. `verify` was still pending at review time.

## Description accuracy
Accurate on behaviour, including the scope call, which now matches the issue's clarification. Two stale details (Nit, no code anchor):
- `Layer 2 of 5. Depends on #962.` #962 has merged (`9c106ae` on `main`), and this PR's base is `main`.
- `TargetingSection.test.tsx 10/10`. The file now runs 11 tests.

## Nits
- Suggestion inline: the `hasBody` comment ends with a history of this PR's first pass. That belongs in the commit message, and it will stop meaning anything once the PR is squashed. The same kind of note is in the first test's comment ("which the first pass at #957 reopened"). Not blocking.
- Copy, optional: "All 1 bullet passes every check" is grammatical but reads oddly ("All 1"). Something like "Your 1 bullet passes every check" would read better. I'm not suggesting it inline because the tests pin the string. It's your call.

---

offlinecv-approver (approved):
🕹✅

Issue #153, opened 2026-06-23
Specificity & Structure score tiles are dead links (#per-bullet-feedback has no target)


Samhit21:
### Summary
On the score readout, clicking **Specificity** or **Structure** does nothing — the URL hash changes to `#per-bullet-feedback` but the page doesn't scroll. Only **Completeness** works. Both broken tiles link to an anchor id that doesn't exist anywhere in the app.

### Repro
1. Score any resume.
2. Click the **Completeness** tile → page scrolls to the Contact card. ✅
3. Click the **Specificity** or **Structure** tile → URL becomes `…/resumelint/#per-bullet-feedback`, nothing scrolls. ❌

### Root cause
`src/components/features/AtsScoreReadout.tsx` renders each dimension as `<a href={anchor}>` (the `Dimension` component). The anchors:
- `Specificity` → `anchor="#per-bullet-feedback"` (line 157)
- `Structure` → `anchor="#per-bullet-feedback"` (line 165)
- `Completeness` → `anchor="#contact"` (line 173)

`#contact` exists (`ContactCard` renders `<Card id="contact">`), so Completeness works. But **no element in `src/` has `id="per-bullet-feedback"`** — the only section ids that exist are `contact` and `reconstructed-resume`. The two tiles point at a target the browser can't find.

**Why it recurs:** The score-tile → section link is a stringly-typed contract between two files that nothing enforces. `AtsScoreReadout.tsx` emits an `anchor` string; the target component is supposed to render a matching `id`. There is no shared constant, no type, and no test coupling the two sides — so the target id was renamed or never added and the link silently rotted. `#contact` survives only by luck: `ContactCard` happens to still render `id="contact"`. Any future section rename re-breaks this the same way, invisibly, because a wrong `href` is not a compile error or a test failure — it's just a no-op click.

### Fix

**Step 1 — Repoint both tiles at the existing `#reconstructed-resume` id.**
The per-bullet feedback lives inside the reconstructed resume, which already renders `id="reconstructed-resume"`. Two-line change in `AtsScoreReadout.tsx`:

```tsx
<Dimension label="Specificity" … anchor="#reconstructed-resume" />
<Dimension label="Structure"   … anchor="#reconstructed-resume" />
```

No new `id` is added; no other files change in Step 1. The "add a tighter `id="per-bullet-feedback"` target" alternative is dropped — `#reconstructed-resume` is sufficient for v1.

**Step 2 — Type the scroll-target contract (same PR).**
[...]

PR #964, opened 2026-09-18
fix(targeting): give the bullet jump link a target below the row (#958)

### LINKED ISSUE #958: Give the triage row a bullet jump link with a target that is actually below it
## Problem

`TargetingTriageRow` reports how many bullets need attention ("4 bullets need attention (3 missing a metric · 1 weak verb)") but, since #956, offers no way to jump to them. The link it used to carry was removed in review because it was wrong in both directions.

The removed link pointed at `#reconstructed-resume`. That id is on the `<section>` in `ReconstructedResume` that **wraps** `ContactCard`, the `TargetingSection` disclosure, and the document — so the target sits *above* the triage row. Clicking "Review bullets ↓" scrolled **up**, past the contact card, while the arrow and the label both promised the opposite.

## Why no anchor was substituted in #956

There is no existing scroll target that is correct for this link, and the obvious candidate is wrong for a real class of résumé.

`SECTION_IDS` (`src/lib/anchors.ts`) has exactly two entries — `contact` and `reconstructed`. The natural addition would be an anchor on the Experience section, but the flagged-bullet pool is not experience-only:

- `scoreSpecificity` folds `extraSources` into `allBullets` (`src/lib/score/score.ts` — "Project (#95) and achievement (#96) bullets feed the same pool as experience bullets").
- `scoreStructure` does the same.
- `score.bullets` (the `BulletObservation[]` this row filters with `needsAttention`) is built from that same pool.

So a résumé whose flagged bullets all live in Projects or Achievements would get a link that jumps to Experience — a confidently wrong destination, which is the failure mode the typed-anchor contract in `anchors.ts` was introduced (#153) to make impossible. The previous `#per-bullet-feedback` dead link is the cautionary case in that module's own docblock.

## What a fix has to decide

1. **What the target is.** Either a single "résumé document body" anchor covering everything below the `TargetingSection` disclosure, or a per-source anchor chosen from where the flagged bullets actually are. The first is simpler; the second is what makes the link useful on a Projects-heavy résumé.
2. **Where it renders.** Everything below the disclosure is currently a flat list of siblings under `<section id={SECTION_IDS.reconstructed} className="flex flex-col gap-6">` in `ReconstructedResume.tsx`. A wrapper element to hang a document-body id on changes the flex gap structure, so it is a layout change, not just an added attribute — which is why it was out of scope for a review round on #956.
3. **Whether the anchor is guaranteed live.** `ExperienceSection` always renders its root `<section>` (no early return), but a Projects-only anchor would need the same check before it can be added to `SECTION_IDS`.

## Acceptance criteria

- [ ] The triage row offers a bullet jump whose target is **below** it on the page, so the affordance's arrow matches its behaviour.
- [ ] The target resolves for a résumé whose flagged bullets are in Projects or Achievements, not only Experience — or the link is scoped so it only renders when its target is correct.
- [ ] Any new id is added to `SECTION_IDS` and rendered from that constant, never hardcoded, per `src/lib/anchors.ts`.
- [ ] A test pins that every href the triage row renders resolves to a live target, in the spirit of `AtsScoreReadout.test.tsx`'s tile-anchor test.

## Notes

Split out of PR #956 review (thread on `src/components/features/TargetingTriageRow.tsx`). The reviewer offered "either add one on the first Summary/Experience block, or drop this link and keep 'Edit contact ↑'"; #956 took the second option, so this issue carries the first.

s-annam:
Gives the triage row's bullet jump a target that is actually below it (#958).

**Layer 3 of 5.** Its base layer, #963, has landed on `main`; this branch is rebased onto it and carries one commit. #965 and #966 stack on top.

## The problem

`TargetingTriageRow` reported how many bullets need attention but offered no way to reach them. The link it used to carry pointed at `#reconstructed-resume` — an id on the section that **wraps** the row, so "Review bullets ↓" scrolled *up* past the contact card while its arrow promised the opposite. #956 removed it rather than substitute a wrong target.

No correct target existed. The obvious candidate — an Experience anchor — is wrong for a real class of résumé: `score.bullets` pools project and achievement bullets alongside experience ones (`scoreSpecificity` / `scoreStructure` both fold in `extraSources`), so a Projects-heavy résumé would get a confidently wrong destination. That is the failure mode the typed-anchor contract in `anchors.ts` was introduced (#153) to make impossible.

## What ships

- `SECTION_IDS.documentBody` (`resume-document-body`) — deliberately not section-specific, because the flagged-bullet pool is not section-specific.
- A `DocumentBody` wrapper (new sibling `DocumentBody.tsx`, so the known-debt `ReconstructedResume.tsx` shrinks rather than grows) spanning Summary → Skills, carrying that id and repeating the parent's exact `flex flex-col gap-6` so sibling spacing is byte-identical.
- A "Review bullets ↓" link gated on `hasBulletGap`, rendered from the constant and never hardcoded.
- Tests for both halves of the link, in the spirit of `AtsScoreReadout.test.tsx`:
  - **Anchor side:** every href the triage row renders is a member of `SECTION_IDS`.
  - **Target side:** rendering `DocumentBody` paints `#resume-document-body`. With the id misspelled, this test fails while the anchor-side tests still pass.
  - **Not pinned:** that `ReconstructedResume` actually mounts `DocumentBody`. It is too heavy to render in a unit test, the same limit the `#contact` precedent has.

## Verification

`npm run typecheck` · `npm run lint` · targeted vitest 34/34 · `npm run test:e2e` 30/30 — run because this changes `ReconstructedResume`'s DOM nesting, and #962's above-the-fold assertion has only ~48px of margin. No layout shift.

## Adversarial review

Reviewed adversarially before this PR existed (2 rounds, `clean: true` at the end). **This layer carried no blocking findings and its acceptance criteria are met in full** — verified in-browser by the reviewer, not accepted on prose:

- The link scrolls **down** (scrollY 0 → 1004) and lands 33px below the sticky header, the same landing gap `#contact` gets.
- The target resolves regardless of which section holds the flagged bullets; both `credentialSections` placements are inside the wrapper.
- The DOM re-nesting broke nothing: no `space-y-*`, no `:first-child`, no direct-child `>` selector and no query depended on the old flat sibling list. Its `scroll-mt-6` matches `ContactCard`'s existing precedent.

Nit, not fixed here: `documentBody` widens the `SectionAnchor` union consumed by `ScoreDimensionRow` and `CollapsedScoreBar`, so a score tile may now legally point at `#resume-document-body` — a loosening of that module's "only anchors referenced by score tiles belong here" invariant, from 2 allowed values to 3. Filed as #973 rather than resolved by guessing at a split union.

Closes #958


src/components/features/TargetingTriageRow.test.tsx
```diff
@@ -47,3 +65,104 @@ describe("formatTriageHeadline", () => {
...
+    (a) => a.getAttribute("href") ?? "",
+  );
+}
+
+describe("TargetingTriageRow anchors (#958)", () => {
+  const validTargets = new Set<string>(
+    Object.values(SECTION_IDS).map((id) => `#${id}`),
+  );
+
+  it("points the bullet jump link at a known scroll target", () => {
```
Vaishnavi1709, line 134: **Secondary: this test proves the anchor side, not that the target exists.**

Each href is checked against `Object.values(SECTION_IDS)`, so the test fails if the link names an unknown id, but not if nothing on the page carries that id. For the new id nothing does in any test: `#resume-document-body` is painted only by the wrapper `<div>` in `ReconstructedResume.tsx`, and no test renders that. **Failure:** delete the wrapper (or render the id raw and misspell it) → all 49 tests still pass → "Review bullets ↓" jumps nowhere. That's the dead-link failure #958 was filed to prevent.

`AtsScoreReadout.test.tsx` makes this same split explicitly (its "scroll-target render (end-to-end wiring)" block, around lines 184–194): membership "proves the anchor *side*", and a separate test renders `ContactCard` to prove `#contact` paints. #958's fourth criterion asks for a *live* target, "in the spirit of" that test, so the same second half is owed here.

`ReconstructedResume` is too heavy to render in a unit test (that same block says so). A cheap way out that also stops the known-debt file from growing: pull the wrapper into a tiny sibling, e.g. `DocumentBody.tsx`:

```tsx
export function DocumentBody({ children }: { children: ReactNode }) {
  return (
    <div id={SECTION_IDS.documentBody} className="scroll-mt-6 flex flex-col gap-6">
      {children}
    </div>
  );
}
```

Then render it once in a test and assert `querySelector(`#${SECTION_IDS.documentBody}`)` is non-null, matching `AtsScoreReadout`'s `#contact` check. The PR body's "live target" line would then be true as written.

Leaving this open for you: add it here, or reply with a follow-up issue and I'll count the criterion as deferred rather than met.

Vaishnavi1709 (dismissed):
Approving: 0 Blocking, 1 Secondary, 0 Nits. The fix is correct. The one finding is about what the new test proves, and it's posted inline.

**Rule applied:** 0 Blocking → APPROVE.

## Acceptance criteria (#958)

- [x] **The target is below the row.** `#resume-document-body` wraps Summary through Skills, all of which render after `TargetingSection`, so "Review bullets ↓" now scrolls down.
- [x] **It resolves for Projects and Achievements bullets.** `ProjectsSection` and both placements of `credentialSections` (achievements and certifications, above and below) are inside the wrapper, so every section that feeds `score.bullets` is covered.
- [x] **The id comes from `SECTION_IDS`.** It's rendered from `SECTION_IDS.documentBody` on both sides and never hardcoded.
- [ ] **A test pins that every href resolves to a live target.** Only half is covered: the anchor side is pinned but the target side isn't. See the inline Secondary.

## Secondary

1. `src/components/features/TargetingTriageRow.test.tsx`: the new test checks that each href's id is in `SECTION_IDS`, not that an element with that id renders. Removing the wrapper `<div>` from `ReconstructedResume` leaves every test green and the link dead. The PR description's "pinning every href ... to a **live** `SECTION_IDS` target" overclaims for the same reason. Details and a proposed fix are in the inline thread. I've left that thread open because it's your call whether this takes a test here or a follow-up.

## Gates

- **Fixture PII:** skipped, no fixtures changed.
- **Design system and reuse:** no raw `<button>`, no new component file, and both links reuse the existing `text-accent-primary` link style.
- **Style tokens:** clean. The grep hits were all issue references (`#958`), not colours.
- **fallow:** no issues in the 5 changed files. Complexity and size findings are pre-existing and report-only.
- **Command-level (skill/script files):** skipped, no skill or script files changed.
- **Checks:** 49/49 tests pass across the triage row, `TargetingSection`, `AtsScoreReadout` and remove-parsed-entry suites. Typecheck and ESLint are clean. CI's `verify`, `e2e` and `fallow` are green.

**Checked and fine:**
- The wrapper adds no extra gap: the inner `flex flex-col gap-6` inside the outer one gives identical spacing.
- No `hashchange` or `location.hash` consumer on `/` reacts to the new fragment.
- `scroll-mt-6` matches ContactCard and the outer `<section>`.

**Description:** accurate except the "live target" claim above. The #973 follow-up for the widened `SectionAnchor` union is correctly scoped out.

Reviewed head: `33062e7`.

---

src/components/features/TargetingTriageRow.test.tsx
s-annam: Fixed in `2e83c7d` (branch collapsed to one commit), taking your suggested shape:

- The wrapper moved into a new sibling `DocumentBody.tsx` (`id={SECTION_IDS.documentBody}`, `scroll-mt-6 flex flex-col gap-6`), so `ReconstructedResume.tsx` shrinks rather than grows (net −11 lines there).
- `TargetingTriageRow.test.tsx` gets a new `scroll-target render (end-to-end wiring, #958)` block that renders `DocumentBody` and asserts `#${SECTION_IDS.documentBody}` is non-null. It mirrors `AtsScoreReadout`'s `#contact` check.

Fail-before check: with the id changed to a raw `"resume-document-bodyx"`, the new test fails (1 failed | 10 passed). The membership tests still pass in that state, which is the gap you found. Restored, 11/11 pass. Typecheck is clean, and the full suite passes with 6808 tests.

Remaining gap: nothing proves that `ReconstructedResume` actually *uses* `DocumentBody`, because that render is too heavy for a unit test. The same limit applies to the `#contact` precedent.

offlinecv-approver (approved):
🕹✅

PR #965, opened 2026-09-18
fix(score): keep the docked strip one line regardless of band (#960)

### LINKED ISSUE #960: [ui] Docked score strip wraps to two lines depending on the verdict label's width, so the card height depends on the score

s-annam:
Makes the docked score strip structurally one line, so the card's height stops depending on the score (#960).

**Layer 4 of 5. Depends on #964.** This is the layer that carried four of the review's five blocking findings — read the review section below before the diff.

## The bug

The strip's root was `flex flex-wrap ... justify-between` with three intrinsic-width groups and no shrink target, so whether the `ⓘ Score details ▾` group fit on row 1 depended on the **rendered width of the score data**: the verdict label runs 6 to 13 characters (`Strong` / `Getting There` / `Needs Work`) in `uppercase tracking-wider`, plus an optional `(layout penalty ×N.NN)` span. Same viewport, different résumé, different card height — and `flex-wrap` broke off the *wrong* group, dropping the only control that reopens the full readout while the optional dimension tracks stayed put.

## What ships

- Root is `flex-nowrap`; the score pill and the `ⓘ`/toggle group are `shrink-0`; the dimension group carries `min-w-0` and is the only thing that yields.
- Verdict word in a fixed `sm:w-28` slot with `truncate`, so its width is constant across bands.
- Inline layout-penalty span removed (it still renders in the expanded readout's footer).
- `CompactDimension` tracks `w-12` → `w-6`, inter-tile `·` separators dropped, group revealed at `lg`.
- `e2e/support/score-hero.ts` — shared drop/dock helpers plus `SCORE_REGIMES`, a 6-fixture band × penalty matrix picked by reading baked `overall`/`multiplier` off each fixture's `.expected.json`.
- `e2e/mobile/score-strip.spec.ts` — activates #962's dormant `mobile-375` project.

## Deviations from the issue, approved and recorded

The scope narrow on #960 said to drop the visible verdict word. It is **kept**, in a fixed-width slot, because dropping it leaves the coloured dot as the only *visual* carrier of the band (WCAG 1.4.1) — `aria-label` serves screen readers but not a sighted user with a colour-vision deficiency, and the widget auto-docks ~4.5s after arrival. The slot removes the data-dependent width, which is the actual requirement. Cost stated plainly: 112px of row, which is part of why the tracks had to shrink. Recorded on the issue as a `**Clarification** —` comment.

The scope narrow also set the dimension tracks' breakpoint to `md`. They are revealed at **`lg`** instead: at 768px the group's width budget is 289–293px, and a whole set of tiles needs 486–496px, so `md` could only mean a two-line strip or a clipped tile. From `lg` up the budget is a flat 545–549px. Below `lg` the dimension detail sits behind `Score details ▾`, which is visible at every width. Recorded on #960 as a second `**Clarification** —` comment.

## Verification

`npm run typecheck` · `npm run lint` · `AtsScoreReadout.test.tsx` 16/16 · `npm run test:e2e` 30/30 on a clean build · full vitest 6747 passed / 10 skipped.

Every new assertion has a **fail-before proof** — see below for why that is not boilerplate here.

## Adversarial review

Reviewed adversarially before this PR existed (2 rounds, `clean: true` at the end). **Four blocking findings, all on this layer, all fixed. Two of them were tests that could not fail.**

1. **The dimension group was clipped mid-tile at every viewport ≥768px**, cutting the Completeness tile in half (`COMPLETENESS ▬▬▬ 2`, the value `27/30` truncated) — on the state the widget docks into. The first fix traded a wrap for a permanent truncation. Root cause is a number neither the issue nor the first fix had: the page container caps at **934px**, so group 2's budget is a flat **545–549px from 1024px up** while the tiles needed 580–589px — no breakpoint could have fixed it, the tiles had to shrink. Now 486–496px, ≥50px spare, `scrollWidth <= clientWidth` verified across six regimes × twelve widths from 375 to 2560. A clipped-but-focusable anchor was also a WCAG 2.4.11 problem; below `lg` the group is `display:none`, confirmed absent from a real 40-key Tab traversal.
2. **`getClientRects().length === 1` was vacuous.** It returns multiple rects only for *inline*-level boxes; the `<section>` and the flex row are both block-level and always return exactly 1. Verified: on a genuinely two-line row it still read 1. Replaced with `rowHeight <= tallestChildHeight + 1`, proven to fail on the two-line case (62 vs 26).
3. **`DOCKED_MAX_PX = 70` admitted the headline bug.** The real docked height is 26px; the 62px it was calibrated against was the *already-wrapped* strip. 62 ≤ 70, so a full regression would have passed. The original fail-before "proof" only went red because it happened to run at 640–767px, where the wrap is three lines (94px) rather than two. Now **34**. `EXPANDED_MAX_PX` was re-measured (185.5px, zero variance across twelve cells) and kept at 200 with the re-verification written down.
4. **The `sm:w-28` slot had no `whitespace-nowrap`**, so a two-word label wrapped *inside* its fixed slot and re-inflated the pill — 26px → 47.3px at an effective 14px font, reachable via Chrome's minimum-font-size setting without any code change. Fixed with `truncate`. Note the detector that fires is the height ceiling, not the row-wrap assertion: the wrap happens inside the pill, so the row still equals its tallest child. That makes the re-derived ceiling load-bearing rather than incidental.

The font-size test carries an **anti-vacuity guard**: it re-reads the computed style and asserts exactly one element took the override, so renaming `sm:w-28` fails loudly instead of silently testing nothing. That guard generalises — any test injecting a style or selector against production markup should assert it landed.

Nits addressed: the ≥50px margin's justification no longer claims cross-OS font variance (Poppins is self-hosted, metrics are identical); the spec documents why its forced font-size must stay at 14 and not rise to 18 (at 18 the section lands at exactly 34.0px, making the ceiling a coin flip).

Closes #960


e2e/support/score-hero.ts
```diff
@@ -0,0 +1,335 @@
...
+
+export const SCORE_REGIMES: readonly ScoreRegime[] = [
+  {
+    name: "strong, no penalty",
+    band: "Strong",
+    penalty: false,
+    fixture: fixturePath("unknown/openresume-react-pdf.pdf"),
+  },
+  {
+    name: "strong, with penalty",
```
Samhit21, line 102: [Suggestion]: Nothing asserts that a regime still renders the band it claims, and two of the six sit within a point of a boundary.

`regime.band` is used in the test *title* and nowhere else — `expectDockedStripIsOneLine` checks row height against the tallest child, per-group `scrollWidth <= clientWidth`, the `DOCKED_MAX_PX` ceiling, and that the toggle and the ⓘ survived. None of that reads the verdict word.

I read `overall` and `layout.multiplier` out of each fixture's `.expected.json` at this head. The matrix is correct today — all six match:

| fixture | overall | mult | band |
|---|---|---|---|
| `openresume-react-pdf` | **81** | 1 | Strong |
| `chromium-asymmetric-sidebar` | **80** | 0.85 | Strong |
| `awesome-cv-resume` | 75 | 1 | Getting There |
| `two-column-achievements-sidebar` | 63 | 0.85 | Getting There |
| `single-column-year-only-roundtrip` | 55 | 1 | Needs Work |
| `deedy-resume-macfonts` | 45 | 0.85 | Needs Work |

The top two are the exposure. `getScoreTier` puts Strong at `>= 80`, so `chromium-asymmetric-sidebar` sits *exactly* on the boundary and `openresume-react-pdf` one point above it. A scorer change that moves either down a single point — the kind that lands here regularly, and that re-bakes these very snapshots — silently reclassifies the regime. The suite stays green, the title still reads `(Strong)`, and the band this matrix exists to cover quietly stops being covered.

That is the same shape as the `DOCKED_MAX_PX = 70` problem this PR fixes: a constant that was calibrated against a state which later stopped being the state it named, with nothing asserting the link.

The pill already carries the band verbatim in its accessible name (`Resume score 80 out of 100, Strong. Expand score details.`), so tying them together is one assertion — in the helper, given the regime:

```ts
await expect(
  page.getByRole("button", {
    name: new RegExp(`^Resume score \\d+ out of 100, ${regime.band}\\.`),
  }),
  `${label}: fixture no longer renders as ${regime.band}`,
).toBeVisible();
```

Worth failing loudly with "this fixture is no longer Strong" rather than leaving the docblock's hardcoded score table as the only record of an assumption the whole matrix rests on.

e2e/mobile/score-strip.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+import { test } from "@playwright/test";
+import {
+  SCORE_REGIMES,
+  dockScoreHero,
+  dropFixtureAndWaitForParse,
+  expectDockedStripIsOneLine,
+} from "../support/score-hero.ts";
+
+const MOBILE_REGIMES = [
+  SCORE_REGIMES.find((r) => r.name === "strong, no penalty")!,
```
Samhit21: [Nit]: A renamed regime turns this into `undefined!` and fails somewhere else.

`SCORE_REGIMES.find((r) => r.name === "strong, no penalty")!` matches on free text. Rename a regime in `score-hero.ts`, or drop one, and the non-null assertion hands `undefined` to `dropFixtureAndWaitForParse`, so the failure surfaces as a property access on undefined inside the helper rather than "no regime named …" here. These names are load-bearing across two files now.

Cheapest fix is to make the lookup assert its own premise:

```ts
const byName = (name: string): ScoreRegime => {
  const regime = SCORE_REGIMES.find((r) => r.name === name);
  if (!regime) throw new Error(`no SCORE_REGIMES entry named "${name}"`);
  return regime;
};
```

Or tag the mobile subset in `SCORE_REGIMES` itself (`mobile: true`) and filter, so the two files share a field instead of a string.

Samhit21 (commented):
## PR Review: fix(score): keep the docked strip one line regardless of band (#960)

### Summary

Reviewed at `be720ab`. The branch was restacked onto `main` while I was reading it, after #964 merged — the tree is byte-identical to `fdb2ce6` (both `1ed9c00e`), so everything I measured applies to the current head unchanged. No prior reviews or threads, so all of this is net-new.

This is a strong change, and the diagnosis is the best part of it. The root cause was that the row's wrap decision depended on the *rendered width of the score data*, and the fix removes that dependency — `flex-nowrap`, `shrink-0` on the two groups that must never yield, and one fixed-width verdict slot — rather than nudging a breakpoint until the symptom went away. The container-budget number is what makes "no breakpoint could have fixed this" a fact rather than a claim: from `lg` up the budget is a flat 545–549px against tiles that needed 580–589px, which is why the tracks had to shrink.

One finding and two nits, inline. None is blocking; the shipped behaviour is right. The finding is about a guard going stale, not about the layout.

- **The new jsdom assertions genuinely fail.** I mutated the five things they pin — dropped `truncate`, flipped `lg:flex`→`md:flex`, `flex-nowrap`→`flex-wrap`, removed `sm:w-28`, re-added the inline layout-penalty span — and each turned **exactly one** test red. I checked rather than trusting the fail-before claims, because on #963 a newly added test could not fail for its own fixture. These hold up.
- **The regime matrix is correct today.** `overall` and `layout.multiplier` read from all six fixtures' `.expected.json`: 6/6 match their declared band and penalty. That is also where the finding comes from — see the inline.
- **The anti-vacuity guard is real.** `forceVerdictWordFontSize` asserts `toEqual([`${px}px`])`, so it fails if the selector matches zero elements *or* more than one, and selecting on `[class~="sm:w-28"]` ties it to the class the fix depends on. That pattern is worth keeping.
- **`DOCKED_MAX_PX` 70 → 34 is honest bookkeeping.** The old ceiling sat above the two-line wrap (62 ≤ 70) and so would have admitted the headline regression; `viewport.spec.ts`'s previous comment explicitly deferred the re-derivation to this PR, and this is it. Replacing `getClientRects().length === 1` is right for the reason given — it exceeds 1 only for inline-level boxes, so on a block-level flex row it was constant.
- **Gates:** `AtsScoreReadout` + `CollapsedScoreBar` 27/27, `eslint` clean, `e2e` and Pages green on this head (`verify` still running when I posted). `tsc` in my sandbox reports 14 errors, **all** in `e2e/`/`playwright.config.ts` from `@playwright/test` being absent in the checkout I borrowed `node_modules` from — none in `src/`. I could not run Playwright locally, so the measured pixel claims (26px docked, 486–496px group width, the 47.3px unfixed wrap) are taken as stated, not verified.

### Findings

1. **`e2e/support/score-hero.ts:102` — nothing ties `regime.band` to what renders, and two fixtures sit within a point of a band boundary.** [Suggestion] Inline. `chromium-asymmetric-sidebar` is exactly 80, the `Strong` threshold; `openresume-react-pdf` is 81. One point of scorer drift drops a band from the matrix with the suite still green and the title still claiming it. One assertion on the pill's accessible name closes it.
2. **`e2e/mobile/score-strip.spec.ts:37` — `SCORE_REGIMES.find(...)!` on a free-text name.** [Nit] Inline.
3. **Below `sm` the band is carried by colour alone for a sighted user.** [Nit] The verdict word is `hidden` under `sm` and the dot is `aria-hidden`, so at 375px colour is its only visual carrier — the WCAG 1.4.1 shape this PR rightly refuses at desktop. The docblock discloses it and argues the trade, and two things genuinely soften it: the numeric score stays visible and the band is a pure function of it, and `Score details ▾` is visible at every width. Worth recording as an accepted exposure somewhere more durable than a code comment, since #959 is already in this area.

### Verdict

**Action: COMMENT.** Nothing blocking. #964 has merged, so the base is `main` and the remaining gate is an approving review rather than the stack.

src/components/features/CollapsedScoreBar.tsx
```diff
@@ -142,46 +178,107 @@ export function CollapsedScoreBar({ score, onExpand }: CollapsedScoreBarProps) {
...
+          too — and the dimension detail is one tap away behind
+          `Score details ▾`, which is visible at every width.
+
+          `min-w-0 overflow-hidden` is kept as containment of last resort so a
+          future overrun can never push group 3 off the row or reopen #959's
+          horizontal-page-overflow finding. It must never actually engage:
+          `expectDockedStripIsOneLine` (`e2e/support/score-hero.ts`) asserts
+          `scrollWidth <= clientWidth` on this group across the full regime ×
+          width matrix, which is what keeps the budget above honest. */}
+      <div className="hidden min-w-0 items-center gap-2 overflow-hidden text-xs text-content-secondary lg:flex">
```
offlinecv-approver, line 267: **Secondary: #960's AC says `md`, and this ships `lg`, but the deviation isn't recorded anywhere.**

The change itself is right. The comment above shows the 768px budget (289–293px) fits no whole set of tiles, and a clipped tile would be worse than an absent one. But the PR `Closes #960`, and that issue's AC list still reads *"Dimension tracks reveal breakpoint is set to `md` (768px and above)."* The `**Clarification** —` comment on #960 records only the verdict-word deviation. The PR body's "Deviation from the issue, approved and recorded" section lists only that one too. After merge, the closed issue will state a breakpoint the code contradicts, with nothing explaining why.

**Fix:** add a second `**Clarification** —` comment on #960 for `md` → `lg`, citing the width budget. Add a line to the PR body's Deviation section as well, since it becomes the squash message. No code change needed.

src/components/features/CollapsedScoreBar.tsx
```diff
@@ -5,13 +5,39 @@
  * CollapsedScoreBar — the docked 1-line strip for the score widget (#953).
  *
  * Renders live score badge, inline dimension metrics with section anchor links,
- * the scoring explainer popover, and an expand toggle. Takes ~38px height.
+ * the scoring explainer popover, and an expand toggle. Measures 26px tall,
+ * identically in all six `SCORE_REGIMES` at 375/768/1024/1280/1440 — the
+ * ceiling that defends that is `DOCKED_MAX_PX` in `e2e/support/score-hero.ts`.
```
offlinecv-approver: Nit: the 375px half of this claim is wider than what's checked in. `e2e/mobile/score-strip.spec.ts` runs three of the six regimes at 375, not all six.

```suggestion
 * the scoring explainer popover, and an expand toggle. Measures 26px tall at
 * every width and score regime the e2e matrix covers — the ceiling that
 * defends that is `DOCKED_MAX_PX` in `e2e/support/score-hero.ts`.
```

e2e/mobile/score-strip.spec.ts
```diff
@@ -0,0 +1,56 @@
...
+ * At 375px the dimension-track group (`CompactDimension` × 3) is already
+ * `hidden` below `lg` (1024px), by design — below that width the row has no
+ * width budget that fits a whole set of tiles, and a clipped tile is not an
+ * option (see the group comment in `CollapsedScoreBar.tsx`) — so every regime
+ * renders the SAME two visible groups: the score pill and the `ⓘ`/`Score
+ * details ▾` toggle. That collapses the
+ * six-fixture matrix `viewport.spec.ts` runs at desktop widths down to what
+ * actually varies at this width: the pill's digit count (1–3 digits) and
+ * whether the (now-invisible-either-way) layout-penalty span used to widen
+ * it. Three regimes — one per verdict band, one of the three carrying a
```
offlinecv-approver: Nit: this says digit count (1–3 digits) varies across the chosen regimes, but they score 81, 63 and 55, all two digits. The penalty span no longer renders at all, so neither thing varies here.

```suggestion
 * actually varies at this width: the pill's digit count — and all three
 * regimes below score two digits, so a `100` or single-digit pill is not
 * exercised here. Three regimes — one per verdict band, one of the three
 * carrying a
```

offlinecv-approver (dismissed):
Approving on 0 Blocking findings: 1 Secondary, 3 Nits and 1 follow-up issue.

The fix does what #960 asks. The strip is `flex-nowrap`, the pill and the toggle groups are `shrink-0`, and the band's width no longer depends on the data. The e2e matrix checks this directly: six regimes at four desktop widths, plus 375px. I checked that the row-height check replacing `getClientRects()` can fail. The `DOCKED_MAX_PX` re-derivation (70 → 34) is correct: 62px was already the two-line wrap. I checked all six `SCORE_REGIMES` `overall`/`multiplier` values against their `.expected.json` sidecars, and they match.

## Acceptance criteria (#960, `Closes`)

- One line for every band, with and without a penalty, at 1280/1024/768/375: **met**. The desktop matrix in `viewport.spec.ts` covers it, and `e2e/mobile/score-strip.spec.ts` covers 375px. CI's `e2e` job ran all 23 tests green.
- The toggle and `ⓘ` never wrap, and the tracks give way instead: **met** (`shrink-0`, and `expectDockedStripIsOneLine` asserts both are visible).
- The pill drops the visible verdict word: **deviates, approved**. The word is kept in a fixed `sm:w-28` slot; see the `**Clarification** —` comment on #960. Filed #988 for the sub-`sm` remainder (below).
- The penalty span is removed: **met**.
- The tracks' breakpoint is `md`: **deviates, not recorded**. The PR uses `lg`. This is the Secondary finding, inline.
- Height is independent of score: **met**.
- Visual verification at the four widths: **met, and exceeded**. It's automated in Playwright instead of left to screenshots.

## Secondary

1. **`CollapsedScoreBar.tsx` (the dimension group), `md` → `lg`.** The deviation is sound, but neither #960 nor the PR's "Deviation from the issue" section records it (inline).

## Nits

1. `CollapsedScoreBar.tsx:8-10`: the docblock says the strip is 26px in all six regimes at 375px, but the mobile spec runs three. Suggestion inline.
2. `e2e/mobile/score-strip.spec.ts:21-23`: the docblock says digit count (1–3 digits) and the penalty span vary, but all three regimes score two digits, and the penalty span no longer renders. Suggestion inline.
3. `e2e/viewport.spec.ts`: the #960 describe block sets its own viewport, so it runs identically under both `desktop-1280` and `desktop-1440`. That's 14 tests, 7 of them duplicates, about 15s of CI. Skipping it on one project would halve that, e.g. a `beforeEach` calling `test.skip(testInfo.project.name !== "desktop-1280")`. Not worth a thread.

## Follow-up filed

- **#988**, *Docked score pill shows the band by colour alone below 640px (WCAG 1.4.1)*. Below `sm` the verdict word is now `hidden`, and on `main` it renders at every width. #960's clarification records the trade-off and says the word can return once the 375px page overflow is fixed. Nothing open tracked that step, and #960 closes on merge. I agree with the trade-off for this PR, so this isn't a finding here.

## Gates

- `AtsScoreReadout.test.tsx` passes 16/16. `typecheck` and `eslint` are clean on the changed files. `fallow audit` reports no issues in the 5 changed files.
- CI checks `verify`, `e2e`, `fallow` and Cloudflare all pass on `be720ab`. The head moved from `fdb2ce6` during the review: a rebase onto `main` after #964 merged, with a byte-identical patch, so everything above holds for `be720ab`.
- Skipped because they don't apply: fixture PII (no fixture added; the six are existing, reused ones), design-system (no raw elements), tokens (no palette or hex).
- `CollapsedScoreBar.tsx` is now 310 LOC, against about 213 before. The growth is almost entirely comments, so no decomposition finding.

Description (3f): accurate on what ships, with one omission. The "Deviation" section lists only the verdict word, not `md` → `lg`.

---

e2e/support/score-hero.ts
s-annam: Fixed in `68a8288`. Added `expectRegimeBand(page, regime)` in `e2e/support/score-hero.ts`: it asserts the pill's accessible name matches `^Resume score \d+ out of 100, <band>\.`, and fails with "`<regime>`: fixture no longer renders as `<band>`". Both matrix loops call it once per regime right after docking (`viewport.spec.ts` and `mobile/score-strip.spec.ts`). Fail-before: I flipped `strong, with penalty` to `band: "Getting There"` and that test went red with the message above. Restoring it made it green again.

e2e/mobile/score-strip.spec.ts
s-annam: Fixed in `68a8288`. Took the first option: `scoreRegime(name)` is exported from `score-hero.ts` and throws `no SCORE_REGIMES entry named "…"` when nothing matches. `MOBILE_REGIMES` goes through it, so there are no `!` assertions left.

src/components/features/CollapsedScoreBar.tsx
s-annam: Done. No code change, as you suggested. Posted a second `**Clarification** —` on #960 for `md` → `lg` with the width budget (289–293px at 768 against 486–496px of tiles; a flat 545–549px from `lg`): https://github.com/offlinecv/OfflineCV/issues/960#issuecomment-5737657318. Also added the deviation to the PR body's Deviations section, which becomes the squash message. The collapsed commit `68a8288` states it too.

src/components/features/CollapsedScoreBar.tsx
s-annam: Applied in `68a8288`. The docblock now reads "at every width and score regime the e2e matrix covers".

e2e/mobile/score-strip.spec.ts
s-annam: Applied in `68a8288`, using your wording. While I was there I also took your review-body nit 3: the #960 describe block in `viewport.spec.ts` now skips under every project except `desktop-1280`, so the 7 duplicate runs are gone (local run: 16 passed, 7 skipped).

offlinecv-approver (approved):
🕹✅

PR #966, opened 2026-09-18
fix(popover): pin the panel to the viewport below sm so it stays on screen (#959)

### LINKED ISSUE #959: [ui] Popover panel renders off-screen at 375px in BOTH alignments — align picks the edge, not the containment

s-annam:
Keeps the `Popover` panel on screen at narrow viewports (#959).

**Was layer 5 of 5 on top of #965; #965 has merged, so this now sits directly on `main`.**

## The bug

At 375px the panel rendered entirely outside the viewport in **both** alignments: docked strip (`align="end"`) at x −83 → 205, expanded header (`align="start"`) at 234 → 522 against a 375px viewport. `align` chooses *which* edge the panel falls off, not whether it stays on screen, and the `max-w-[calc(100vw-2rem)]` clamp never engages because the panel renders `w-72` (288px) against a 343px clamp — it bounds width, never position. Nothing in the primitive was viewport-aware.

This surface is, per the primitive's own docblock, "the one keyboard-reachable place the scoring rules are written down", so below `sm` that content was unreachable.

## What ships

`PANEL_BASE` gains `max-sm:fixed max-sm:inset-x-4 max-sm:top-auto max-sm:bottom-4 max-sm:w-auto max-sm:max-h-[calc(100vh-2rem)] max-sm:overflow-y-auto` — below `sm` the panel stops anchoring to the trigger and pins to the viewport as a bounded bottom sheet, leaving `align` to govern `sm`+ where it works. CSS-only, per the primitive's stated no-JS-positioning constraint.

## The issue's own suggested snippet does not work

`max-sm:fixed max-sm:inset-x-4 max-sm:w-auto` — implemented verbatim first, and measured the panel at **y=816 on an 812px-tall viewport**, i.e. fully below the fold. `top-full` (`top: 100%`) stays in effect, and for a `fixed` element a percentage `top` resolves against the **viewport's** height, not the trigger's — so it means "one full viewport down". Hence `top-auto` + `bottom-4`. The naive version passes a horizontal-only assertion while being completely unusable, which is why the containment spec checks both axes.

## Verification

`npm run typecheck` · `npm run lint` · `Popover.test.tsx` 13/13 · `npm run test:e2e` 30/30 on a clean build (23 passed + 7 skipped after the rebase onto `main`, because #965 now runs its viewport-independent block under `desktop-1280` only), including three new 375px containment tests (both alignments, both widget states, plus `AchievementTypePicker`) and two at 1280/1440 proving `sm`+ anchoring is unchanged. Fail-before: with `Popover.tsx` stashed, 2 of 3 mobile tests fail at exactly the pre-fix numbers (−83.3 and 521.75).

## Adversarial review

Reviewed adversarially before this PR existed (2 rounds, `clean: true` at the end). **No blocking findings on this layer**, and the two attacks most likely to sink it both failed:

- **The cascade-order question.** This repo ships no `tailwind-merge`, and `PANEL_ALIGN` still applies `left-0`/`right-0` while `PANEL_BASE` applies `max-sm:inset-x-4` — `CollapsedScoreBar` documents two past bugs where a class was silently inert for exactly this reason. In the built CSS `.right-0` lands at byte 13173 and `.max-sm\:inset-x-4` at 39282, so the logical property wins on source order. Computed at 375px: `left: 16px; right: 16px`, identical in both alignments.
- **The `position: fixed` containing block.** Every `transform`, `filter`, `perspective`, `backdrop-filter`, `contain` and `will-change` in `src/` was swept; none is an ancestor of any `Popover` consumer, so the containment assertions pass for real reasons rather than luck.

Nits, resolved:

- **`max-h` + `overflow-y-auto` added.** The bottom-anchored sheet had `height: auto` and grew *upward* without limit — a long panel measured at h 1206 / y −410 with nothing scrollable. Now tracks the viewport (780px at 812, down to 288px at 320) with both current callers fitting unscrolled.
- **Breakpoint stays `max-sm`, and the review's own proposal to narrow it to 480px was withdrawn on measurement.** The binding caller is the expanded-header explainer, whose right edge is a constant 557.8px, so at 481 it overflows by 76.8px and at 500 by 57.8px — narrowing would have reopened this bug in a 481–558px window. 560 clears by 2.2px (luck, not margin); 600 buys an 80px band for a margin one added word in the heading above would eat. The docblock records the measurement, and a wrong *reason* I had written there ("the trigger's x shifts with the score digits" — it does not; `justify-between` pins that group) is corrected.

Known, not fixed here — filed as a follow-up (#971): the fixed sheet does not follow scroll, so it can sit pinned with its trigger scrolled away and occlude unrelated controls with no backdrop; and `overflow-x` coerces to `auto` alongside `overflow-y`, which a future caller with a long unbreakable string would notice.

Closes #959


src/design-system/primitives/Popover.tsx
```diff
@@ -55,8 +55,47 @@ import { Button, type ButtonVariant } from "./Button.tsx";
...
+// that copy. `max-sm` is the narrowest threshold worth having.
+//
+// `max-h` + `overflow-y-auto` bound the sheet: `top:auto; bottom:16px;
+// height:auto` grows UPWARD without limit, and a long panel was measured at
+// h 1206 / y -410 with nothing scrollable — invisible off the top. Today's
+// two callers happen to fit; this is a shared primitive, so it must not
+// depend on that.
 const PANEL_BASE =
-  "absolute top-full z-20 mt-1 w-72 max-w-[calc(100vw-2rem)] rounded-lg border border-border-light bg-surface-card p-3 text-content-primary shadow-lg";
+  "absolute top-full z-20 mt-1 w-72 max-w-[calc(100vw-2rem)] rounded-lg border border-border-light bg-surface-card p-3 text-content-primary shadow-lg max-sm:fixed max-sm:inset-x-4 max-sm:top-auto max-sm:bottom-4 max-sm:w-auto max-sm:max-h-[calc(100vh-2rem)] max-sm:overflow-y-auto";
```
Samhit21, line 108: [Suggestion]: `max-sm:fixed` resolves against the nearest transformed/filtered ancestor, and nothing enforces that there isn't one

The comment block above defends the *cascade* thoroughly, but `position: fixed` has a second failure mode it doesn't record: any ancestor with a non-`none` `transform`, `filter`, `backdrop-filter`, `perspective`, `contain: paint/layout`, or `will-change` on one of those becomes the containing block, and the panel then pins to *that box* instead of the viewport.

This is correct today — I checked rather than assumed. Both `Popover` consumers (`CollapsedScoreBar` via `AtsScoreReadout`, `AchievementTypePicker` via `ReconstructedResume`) render inside `PageShell`'s `{children}` (`PageShell.tsx:258`), which is a sibling of the header, not a descendant. The only `transform` in `src/` is `ScoreRing`'s, and it's SVG-internal on a component that takes no children.

What makes it worth a line of comment rather than nothing is the near miss. `PageShell.tsx:174` is:

```
<header className="sticky top-0 z-20 ... bg-surface-base/95 px-6 pb-6 pt-2 backdrop-blur">
```

`backdrop-blur` creates a containing block, and that same header exposes an open `headerExtra?: ReactNode` slot (`PageShell.tsx:75`, rendered at `:222`). A `Popover` dropped into `headerExtra` — the one slot in the app explicitly designed to accept arbitrary nodes — would get a header-anchored sheet below `sm`, not a viewport-anchored one.

The new e2e can't catch that. `assertContained` checks `0 <= x`, `x + width <= vp.width`, `0 <= y`, `y + height <= vp.height` — and a sheet bottom-anchored to a `top-0` header still satisfies every one of those while sitting in entirely the wrong place. The spec would stay green.

This is the same argument the `max-h` paragraph already makes two comments up — "Today's two callers happen to fit; this is a shared primitive, so it must not depend on that" — just applied to the containing block instead of the height. A sentence recording the constraint is probably all it needs.

Samhit21 (dismissed):
## PR Review: fix(popover): pin the panel to the viewport below sm so it stays on screen (#959)

### Summary

This fixes the defect I raised on #956 (`Popover.tsx:51`, "`max-w` clamps the panel's width, not its position"), and it's worth saying plainly that my suggested remedy there was wrong. I proposed `align` as the fix; #959 measured that `align` only chooses *which* edge the panel falls off — `right-0` throws it 225px past the left once the docked strip wraps at 375px, `left-0` throws it 147px past the right. I reasoned from the CSS because jsdom has no layout engine and flagged it as unmeasured. This PR measured it, and the measurement says the prop I recommended was insufficient. Pinning to the viewport is the right shape.

I verified the load-bearing claims rather than taking the body's numbers:

- **Cascade order reproduces on my own build.** With no `tailwind-merge`, the fix only works if each `max-sm:*` is emitted *after* the base utility it overrides. Building and grepping byte offsets in `dist/assets/styles-*.css`: `.absolute` @12876 vs `.max-sm\:fixed` @39243; `.top-full` @13154 vs `.max-sm\:top-auto` @39330; `.w-72` @16181 vs `.max-sm\:w-auto` @39475; `.left-0` @13342 and `.right-0` @13173 vs `.max-sm\:inset-x-4` @39273. All five overrides win. This is the repo's documented past failure mode and it's genuinely avoided here.
- **`top-auto`/`bottom-4` is load-bearing, as claimed.** `top: 100%` on a `fixed` element resolves against the viewport's height, so overriding only the horizontal axis would render the panel a full viewport below the trigger. Good catch that the issue's own suggested snippet had this bug.
- **The new jsdom tests aren't vacuous.** Dropping `max-sm:bottom-4` and separately `max-sm:overflow-y-auto` from `PANEL_BASE` each turns exactly the 2 new tests red and leaves the other 11 green.
- **CI's `e2e` job really ran the new specs.** All three `popover-containment` tests executed under `mobile-375` at 375x812, and both new `viewport.spec.ts` tests ran on `desktop-1280` and `desktop-1440`. The 7 skips are pre-existing `#960` `desktop-1440` cases, unrelated. All four checks pass on `602f1c17`.
- **The breakpoint reasoning holds.** Keeping `sm` rather than narrowing to 480 is right: with the binding trigger's panel right edge constant at ~558px, a 481-639 band would reopen the bug, and 640 clears it by ~82px instead of 560's 2.2px.

The `max-h` + `overflow-y-auto` bound is a genuinely good instinct — a bottom-anchored `height:auto` sheet growing upward off the top of the screen is exactly the kind of thing that only shows up on a long panel nobody has yet. Splitting the proof across `e2e/mobile/` (containment below `sm`) and `e2e/viewport.spec.ts` (`align` still anchors above `sm`) is the right decomposition, and #971 honestly records the two behaviours a bottom sheet inherits rather than quietly shipping them.

### Findings

1. **[Suggestion] `Popover.tsx:98`** — `max-sm:fixed` resolves against the nearest ancestor with a `transform`/`filter`/`backdrop-filter`/`contain`/`will-change`. Correct for today's two callers (both render in `PageShell`'s `{children}`, not its header) — but `PageShell`'s sticky header carries `backdrop-blur` *and* an open `headerExtra?: ReactNode` slot, and the new `assertContained` would stay green if a future `Popover` landed there, since a header-anchored sheet is still inside the viewport. Worth a sentence next to the `max-h` paragraph, which already makes this exact "shared primitive must not depend on today's callers" argument.

2. **[Nit] `602f1c17` is an empty commit** ("chore(ci): re-trigger the Cloudflare Pages build") — no diff at all. `CLAUDE.md` wants one commit per PR, so this wants collapsing into `cbb03611` before the merge queue.

Neither is merge-blocking: #1 is a comment about a constraint that currently holds, and #2 is housekeeping the merge queue resolves.

### Verdict

**Action: APPROVE**

src/design-system/primitives/Popover.tsx
s-annam: Fixed in `143f2e0`. Added a paragraph under the `max-h` one in the `PANEL_BASE` comment block. It lists the ancestor properties that make `position: fixed` pin to an ancestor instead of the viewport: `transform`, `filter`, `backdrop-filter`, `perspective`, `contain: paint|layout`, and `will-change` naming any of them. It notes both current callers are clear, names the `PageShell` `backdrop-blur` header plus `headerExtra` as the near miss, and says the containment e2e cannot catch it because a header-anchored sheet still passes every bound. Comment-only; I checked `PageShell.tsx:174` and `:222` before writing it.

offlinecv-approver (approved):
🕹✅

Issue #952, opened 2026-09-16
parser: isInlineDatedProgram's date-word strip omits Current/Ongoing/Now, so a season-led open-ended range reads as a program


s-annam:
## Problem

`isInlineDatedProgram` decides whether a degree-less line is a program name by **erasing date
words and asking whether substantive text is left**. Its strip covers seasons, months and the
single literal `present` — but not the other three open-ended words the repo already keeps in
one place. So a season-led attendance range that ends in `Current`, `Ongoing` or `Now` leaves
that word behind as its "program text" and the predicate answers `true` on a line that is
nothing but a date.

`src/lib/heuristics/extract/education.ts:412` on `main`:

```ts
.replace(
  new RegExp(String.raw`\b(?:${SEASON_ALT}|present|${MONTH_ALT})[a-z]*\b`, "gi"),
  "",
);
```

`present` is a hardcoded literal where every other vocabulary in the expression is an imported
constant. `OPEN_ENDED_ALT` — `Present|Current|Now|Ongoing` — is **already imported into this
file** (`education.ts:20`) and already used one screen up by `ATTENDANCE_RANGE_END`
(`education.ts:180`), so nothing new has to be plumbed in.

## Repro — at the predicate

Verified on `origin/main` @ `693fbe5`, calling the real predicate (declared at
`education.ts:367`):

```
isInlineDatedProgram("Fall 2013 - Current")   // → true   (expected: false)
isInlineDatedProgram("Fall 2013 - Ongoing")   // → true   (expected: false)
isInlineDatedProgram("Fall 2013 - Now")       // → true   (expected: false)
isInlineDatedProgram("Fall 2013 – Current")   // → true   (expected: false)  [en dash]
```

**Only the season-led form slips.** Two neighbouring shapes are already rejected, which is what
has kept this latent:

```
isInlineDatedProgram("2019 - Current")        // → false  ✅ DATE_LEAD_RE catches the bare-year lead
isInlineDatedProgram("Sept 2019 – Present")   // → false  ✅ `present` IS in the strip, and it is
                                              //             by far the commonest spelling
isInlineDatedProgram("Fall 2013 – Spring 2014") // → false ✅
```

Walkthrough for `"Fall 2013 - Current"`: the Title-case lead passes, a year is present,
`DATE_LEAD_RE` does not fire (the line leads with a *season*, which that gate does not list —
this is exactly why the year-led form is safe and this one is not), the year is stripped, `Fall`
[...]

PR #980, opened 2026-09-18
fix(parser): a one-word line beside a year must not mint a school (#979)

### LINKED ISSUE #979: [parser] education — a misspelled month leads an entry: "Setember 2021" is parsed as an institution
A misspelled month at the head of an education line is not recognised as a date, so the line is read as a program and the parser emits an entry whose **institution is the misspelling**. `Setember 2021` becomes a school called "Setember".

## Repro

Verified on `origin/main` @ `693fbe5`, driving the real `extractEducation` over an education section:

```
Education
Yale University
B.A. History, 2010 - 2014
Setember 2021
```

| input line | entries out |
|---|---|
| `Setember 2021` | **2** — `Yale University` (History), and `Setember` |
| `Agust 2018` | **2** — `Yale University` (History), and `Agust` |
| `Jnuary 2020` | **2** — `Yale University` (History), and `Jnuary` |

With a real school on the line after it, the misspelling lands as that school's `field` instead:

```
Education
Yale University
B.A. History, 2010 - 2014
Setember 2021
Harvard Summer School
```
→ `Harvard Summer School` comes out with `field: "Setember"`.

Controls that already behave, on the same run: `Febuary 2020`, `Janurary 2019`, `Septemeber 2020`, `Aprill 2020` are all harmless (one entry, no field pollution) — a downstream strip in `stripInstitutionDate` composes the loose `MONTH`, whose `[a-z]*` tail happens to swallow anything beginning with a real month prefix. The three above are the ones whose **first three letters are not a month abbreviation**, so nothing catches them at any layer.

## Cause

Two gates both miss the token, on `origin/main`:

- `src/lib/heuristics/extract/education.ts:330` — `DATE_LEAD_RE` decides whether a line *leads* with a date. Its month branch is built from the month vocabulary, and `Setember` is in neither the enumerated nor the abbreviation-prefixed form.
- `src/lib/heuristics/extract/education.ts:412` — the date-word strip inside `isInlineDatedProgram`. Same vocabulary, same miss, so after stripping the year the remainder is `Setember`, which reads as substantive program text.

`isInlineDatedProgram` then answers `true`, the line is taken as a program lead, and an entry is opened on it.

## Why this is worth its own issue

This is the **fabricated-credential** class, not a boundary miss: the output names an institution the résumé does not contain. That is strictly worse than the date-attachment bug #925 fixes, and it is the same shape the #951 review caught twice on `Summers 2013, 2014` and `Fall 2013 – Presently` — both of which are now pinned. The misspelling route is the one the vocabulary cannot reach by enumeration.

It is **pre-existing and not a regression from #951**. Measured both refs on the same inputs: every row above is byte-identical on `origin/main` and on `fix/925-inline-dated-program-strict-month`. #951 correctly leaves it alone.

Sibling of #952, which is the same fabricated-entry shape reached through a different gap (`Current` / `Now` / `Ongoing` missing from the strip). Both want the same kind of answer, and whoever takes one should look at the other.

## Proposed fix

Enumeration cannot close this — the space of misspellings is open. Two directions, and the choice is the interesting part of the issue:

1. **Require the remainder to look like a program name, not merely to be non-empty.** The strip's remainder test is the weak link: a single capitalised token with a year beside it and no program vocabulary (`Certificate`, `Program`, `Diploma`, `School`, `Institute`, `Academy` …) is far more likely a mangled date than a field of study. `isInlineDatedProgram` already carries a denylist (`EDUCATION_ANNOTATION_RE`, the `GPA` / `Minor` / `Major` guard); this is the positive half.
2. **Refuse to open an entry on a single-token institution that is not in `INSTITUTION_HINTS` and carries no other evidence.** Narrower, and it catches whatever route produces a one-word school name, not just this one.

Direction 1 is closer to the cause; direction 2 is the cheaper backstop. Either should be measured against the corpus before it lands, because both narrow a recognition path.

## Acceptance criteria

- [ ] `Setember 2021`, `Agust 2018` and `Jnuary 2020`, each as the third line of the section above, produce **one** entry — `Yale University` — and no second entry
- [ ] With `Harvard Summer School` following, **no institution is minted from the misspelling** —
  the partner school is preserved as the entry's institution. That school may carry the
  misspelling as its `field`; see the scope-narrow note below
- [ ] The controls still parse unchanged: `Marketing 2020`, `Marketing Certificate 2020`, `MIT Applied Data Science (2023)`, `Data Science Certificate 2020` are still recognised as programs
- [ ] Real date lines are still rejected: `May 2011`, `September 2020`, `Sept. 2019`, `Fall 2013 – Spring 2014`, `Summers 2013, 2014`
- [ ] The new cases fail against the pre-fix implementation
- [ ] Any corpus or ground-truth movement is adjudicated fixture by fixture in the PR body, not rebaselined
- [ ] `npm run verify` green

## Scope narrowed — AC 2 (2026-09-18)

AC 2 originally required the partner school's `field` to be `undefined`. It is narrowed to
"no institution is minted from the misspelling", because the stricter form cannot be met
without a regression that costs more than it buys.

`Setember 2021` / `Harvard Summer School` and `Photography 2020` / `Coursera` are **the same
shape** at this layer: a one-token line beside a year, sitting over an institution line.
Nothing available to `isProgramLeadAt` distinguishes a mangled date from a legitimate
one-word program title. Requiring `field: undefined` means narrowing that predicate, which
measurably drops real degree-less programs on parse *and* breaks the parse → export →
re-parse round-trip — our own exporter renders a degree-less entry with the field as the
header and the inline date as its only cue (#302).

So the trade is: the rare fabricated `field` on an adversarial input, against no entry loss
on ordinary résumés. The fabricated-*institution* class this issue was filed about — the
severe half, where the output names a school the résumé does not contain — is fully closed
either way.

## Out of scope

- `Current` / `Now` / `Ongoing` in the strip — that is #952.
- A spell-corrector for month names. The fix should be about what a remainder has to look like to be a program, not about guessing the intended month.

Found while reviewing #951.

s-annam:
A misspelled month at the head of an education line was not recognised as a date, so the line was read as a program and the parser emitted an entry whose **institution was the misspelling**. `Setember 2021` became a school called "Setember".

This is the fabricated-credential class — the output names an institution the résumé does not contain — not a boundary miss.

## What was actually happening

The line survived two gates (`DATE_LEAD_RE`, and the date-word strip inside `isInlineDatedProgram`) because neither vocabulary contains `Setember`, and **no vocabulary ever can**: the space of misspellings is open.

Measured on `main` before the fix, `Yale University` / `B.A. History, 2010 - 2014` / *line*:

| line | entries out |
|---|---|
| `Setember 2021` | **2** — Yale, and a school named `Setember` |
| `Agust 2018` | **2** — Yale, and `Agust` |
| `Jnuary 2020` | **2** — Yale, and `Jnuary` |
| `Zebra 2020` | **2** — Yale, and `Zebra` |
| `Banking 2020` | **2** — Yale, and `Banking` |
| `Marketing 2020` | 1 — Yale |

The last two rows are the finding that shaped the fix. `Zebra` and `Banking` are not misspelled months and fabricate identically, so **this was never about months** — it is about a one-token remainder being treated as a program name. And `Marketing 2020` is fine only by accident: the loose `MONTH` in a downstream `stripInstitutionDate` erases anything starting with a month prefix, which is exactly why `Febuary` / `Janurary` / `Septemeber` / `Aprill` were harmless in the issue's table while `Setember` / `Agust` / `Jnuary` were not.

## The fix

The weak link is the remainder test: strip the dates, then ask only whether substantive text is left.

That is the **right** question for the strip. A one-word survivor is precisely what #925/#951 proved must not be eaten — `Springfield 2020`, `Fallow 2019`, `Presenting 2020` are pinned on it. But "a word survived" is far too weak a licence to **mint an institution** from.

So the predicate is **split, not changed**:

- `isInlineDatedProgram` — unchanged, still answers the strip question. Every existing assertion on it still holds, deliberately.
- `isInlineDatedProgramEntry` — new. Same predicate plus one requirement: the remainder must be **more than a single bare token**, unless that token is itself a credential word (`Certificate`, `Diploma`, `Bootcamp`, …).

**Exactly one site asks the narrow gate**: the hint-less entry-boundary test inside `extractEducation`, the one place where an entry may open with no partner line at all to corroborate it. Both partner-lookahead callers keep the loose predicate — see the next section, which is the larger of the two decisions in this PR.

A mangled date leaves **one letter-bearing token**; a real program title does not. This is #979's Direction 1 (the positive half of the remainder test), applied at the entry-lead layer so the strip's own regressions are untouched.

## What pushed back — twice, in opposite directions

**`isInstitutionLeadAt` keeps the loose predicate.** My first pass pointed all three callers at the narrow gate. It broke the #882 hint-less institution-led path, because a school name is routinely one token — `MIT  May 2024`, `Caltech 2019` — so the gate rejected exactly the shape that function exists to recognise:

```
FAIL corpus-roundtrip.test.ts > hint-less institution-led education (#882) > parses three hint-less schools as three entries
FAIL corpus-roundtrip.test.ts > … re-exports and re-parses to the SAME three entries — 3 in, 3 out
FAIL corpus.test.ts > tests/fixtures/pdfs/unknown/education-hintless-institution-lead.pdf > cascade + score match the snapshot
FAIL render-roundtrip-education-date-column.repro.test.ts > #882 — keeps `MIT` and `Stanford University` separate
```

Caught by the corpus, not by argument.

**`isProgramLeadAt` keeps it too — the review caught that one, and it is the bigger call.** A one-word *program* title is just as ordinary as a one-word school: `Photography 2020` / `Coursera`, `Welding 2019` / `Lincoln Technical Institute`. With the narrow gate there, both entries were lost on parse, and the round-trip was worse — `ats-resume-model.ts` renders a degree-less entry with the field as the header and the inline date as its only cue (#302), so our **own export** re-parsed to `field: null`. Reverted, and both cases are now pinned.

Both callers can afford the loose test because their conservatism lives in a **partner lookahead** — `isInstitutionLeadAt` requires a degree line under it, `isProgramLeadAt` an institution line — whereas the hint-less boundary site has no partner to consult. The reasons are comments on the calls.

**The cost of that revert is AC 2**, and it is the one thing in this PR that trades a stated requirement away rather than meeting it. See the checklist below.

## Only letter-bearing tokens count

Hyphens, periods, apostrophes and ampersands stay *inside* a word so that `Post-Graduate`, `B.Sc.`, `Dean's` and `R&D` each count once. That keep-set also left the bare hyphen of an ASCII range standing as a **token of its own**, so the gate saw two words where the strip had left one:

| line | before | after |
|---|---|---|
| `Agust 2018 - 2020` | 2 entries — Yale, and `Agust` | 1 — Yale |
| `Agust 2018-2020` | 2 entries — `Agust` | 1 — Yale |
| `Agust 2018 - Present` | 2 entries — `Agust` | 1 — Yale |
| `Setember 12, 2021` | 2 entries — `Setember 12` | 1 — Yale |
| `Agust 2018 – 2020` (en dash) | 1 — Yale | 1 — Yale |

The ASCII-hyphen range is the ordinary way a résumé writes a range, so this was the normal shape of the input, not an edge — and it is what makes the claim "a mangled date is one token" true of every separator rather than only of the en dash. Counting only tokens that contain a letter drops punctuation-only and digit-only ones; `Post-Graduate Diploma 2019`, `R&D Certificate 2021` and `MIT Applied Data Science (2023)` are unaffected, and are pinned as controls.

`coursework` and `fellowship` came out of `PROGRAM_TITLE_WORD_RE` in the same pass: both are denied upstream by `EDUCATION_ANNOTATION_RE`, so those alternatives were unreachable.

## Acceptance criteria

- [x] `Setember 2021`, `Agust 2018`, `Jnuary 2020` each produce **one** entry — `Yale University` — and no second entry
- [x] with `Harvard Summer School` following, **no institution is minted from the misspelling** and the partner school is preserved — but that school does carry `Setember` as its `field`. **#979's AC 2 was narrowed to this**, with a `**Clarification (scope narrow)** —` comment on the issue recording the original wording and the reason; the trade-off is argued in the test named *"preserves the school following … (AC 2 relaxed for #302 round-trip)"*
- [x] controls unchanged: `Marketing 2020`, `Marketing Certificate 2020`, `MIT Applied Data Science (2023)`, `Data Science Certificate 2020`
- [x] real date lines still rejected: `May 2011`, `September 2020`, `Sept. 2019`, `Fall 2013 – Spring 2014`, `Summers 2013, 2014`, `Fall 2013 – Presently`
- [x] the new cases fail against the pre-fix implementation
- [x] no corpus or ground-truth movement to adjudicate
- [x] `npm run verify` green

## Verification

Fail-before is a **mutation**, not an assertion about an older ref — the gate's body replaced with the pre-fix expression, everything else identical.

The original gate (`return true` in place of the token test):

```
× does not open a second entry on Setember 2021 / Agust 2018 / Jnuary 2020
× catches the whole one-token class, not just misspelled months
× keeps the strip predicate and the entry gate as separate questions
```

The letter-bearing filter (`.filter(Boolean)` in place of `.filter((w) => /[A-Za-z]/.test(w))`):

```
× does not open a second entry on the RANGE form Agust 2018 - 2020
× does not open a second entry on the RANGE form Agust 2018-2020
× does not open a second entry on the RANGE form Agust 2018 - Present
× does not open a second entry on the RANGE form Setember 12, 2021
Tests  4 failed | 116 passed (120)
```

The en-dash row stays green under that mutation, which is the point of keeping it in the table: it is the form that already passed, pinning that all five now agree.

Restored: 120 passed in that file. Full suite — **395 files, 6789 passed**, 10 skipped. `tsc -b` clean, eslint clean on the touched files.

**No fixture or snapshot moved.** All 58 corpus fixtures and every `*.expected.json` are byte-identical, so there is nothing to adjudicate fixture by fixture.

## Review focus

- **AC 2 is narrowed, not met.** That is the one requirement this PR trades away, and the argument for it is in *What pushed back* above. Worth disagreeing with if you think the fabricated `field` costs more than the one-word degree-less programs the strict form would have dropped.
- The **split predicate** is the other judgement call. The alternative was tightening `isInlineDatedProgram` itself, which would flip six landed assertions from #925/#951 (`Springfield 2020`, `Fallow 2019`, `Wintergreen 2021`, `Summersville 2018`, `Presents 2020`, `Presenting 2020`) — proxies for "the strip did not eat this word". Rewriting them to accommodate a new gate is the rebaselining the issue rules out, so the gate went beside them instead.
- **`PROGRAM_TITLE_WORD_RE` is a positive vocabulary**, consulted only for a one-word remainder. Deliberately small: a wider list starts admitting the mangled dates it exists to reject. It does mean a one-word acronym credential (`PMP 2022`) no longer opens an entry *at the hint-less boundary site* — narrower than before, and the corpus says nothing is lost, but it is a real behaviour change.

## Not in this PR

**#982** — two routes that still fabricate an institution and that this PR cannot close by design, both pre-existing on `main`: a two-misspelling range (`Setember 2021 – Agust 2022`, where two letter-bearing tokens genuinely survive), and a mangled lead over a non-school partner (`Setember 2021` / `Dean's List`, which `isInstitutionLine` accepts).

**#952** — `Current` / `Now` / `Ongoing` missing from the same strip, so `Fall 2013 – Currently` fabricates identically. Same shape, different gap, and the fix it wants is the already-imported `OPEN_ENDED_ALT` rather than anything here.

`fallow audit` re-attributes two **pre-existing** CRITICAL complexity findings (`extractEducation`, `educationFromChunk`) because this diff touches lines inside the first. Neither function changed size here, and decomposing a 338-line parser is not a bug fix's job.

Closes #979


src/lib/heuristics/extract/education.ts
```diff
@@ -549,7 +611,7 @@ function isProgramLeadAt(
   const partner = lines[i + 1]?.text;
   if (lead === undefined || partner === undefined) return false;
   if (DEGREE_RE.test(lead) || INSTITUTION_HINTS.test(lead)) return false;
-  if (!isInlineDatedProgram(lead)) return false;
+  if (!isInlineDatedProgramEntry(lead)) return false;
```
offlinecv-approver: Blocking, see review body finding 1. With the narrow gate here, `Photography 2020` / `Coursera` no longer opens an entry: the Coursera entry is lost on parse, and a one-word degree-less field is lost on our own export → re-parse (the #302 cue in `ats-resume-model.ts`). This lead already has the `isInstitutionLine(partner)` lookahead, which is the same evidence argument used for keeping `isInstitutionLeadAt` loose, but loosening it here re-opens #979 AC 2 (`Setember 2021` / `Harvard Summer School`). This needs a decision, not a mechanical fix.

src/lib/heuristics/extract/education.ts
```diff
@@ -511,9 +511,70 @@ export function isInlineDatedProgram(line: string): boolean {
...
+ * just the shape a résumé actually produces.
+ *
+ * So the gate is not about months at all, and deliberately not a spell
+ * corrector (#979 rules one out): a single capitalised token sitting beside a
+ * year, with no program word and no institution hint, carries no evidence that
+ * it is a credential. Two tokens is enough — `Marketing Certificate 2020`,
+ * `MIT Applied Data Science (2023)` — because a mangled date is one token by
+ * construction.
+ *
+ * The three callers that decide entry boundaries ask THIS; nothing else should.
```
offlinecv-approver: Nit: only two callers ask this one (`isProgramLeadAt` and the `extractEducation` boundary site). `isInstitutionLeadAt` deliberately asks the loose predicate.

```suggestion
 * The two callers that decide PROGRAM entry boundaries ask THIS; `isInstitutionLeadAt` deliberately asks the loose predicate (see the comment there).
```

offlinecv-approver (changes requested):
The split predicate is a good shape, and the misspelled-month cases are fixed and pinned. One real regression blocks the merge, though: the narrower gate drops **legitimate** degree-less programs whose title is one word. It also breaks the parse → export → re-parse round-trip for them.

Reviewed head `ea271f8`.

## Blocking

**1. A one-word degree-less program is now lost, both on parse and on round-trip.** (`education.ts:614`, `isProgramLeadAt`)

`isProgramLeadAt` now asks `isInlineDatedProgramEntry`, so a real program titled with a single non-credential word no longer opens an entry, even with an institution line directly under it. Measured with the real `extractEducation` on both refs:

| input | `main` | this PR |
|---|---|---|
| `Photography 2020` / `Coursera` / `Welding 2019` / `Lincoln Technical Institute` | 2 entries (Coursera·Photography·2020, Lincoln·Welding·2019) | **1 entry**: `Lincoln Technical Institute`, no field, dated **2020**. Coursera is gone. |
| `Yale University` / `B.A. History, 2010 - 2014` / `Photography 2020` / `Coursera` | 2 entries | **1 entry**. Coursera and Photography are dropped. |

The round-trip is the sharper half. `ats-resume-model.ts` renders a degree-less entry with the field as the header and the date on that line, and its own comment says that date *"IS the `isInlineDatedProgram` cue, and it is the only cue this shape has"* (#302). Driving the real exporter (`buildAtsResumeModel` → `renderAtsResumePdf` → `runCascade`, the same harness as `render-roundtrip-education-lone-year.repro.test.ts`) with two degree-less entries, `field: "Photography"` / `"Welding"`:

- `main`: `[Ridgemont·Photography·2023], [Lakeside·Welding·2021]`, correct.
- this PR: `[Ridgemont·null·2023], [Lakeside·null·null]`. Both fields and one date are lost on our own export.

The multi-word control (`Certificate Program in Applied Analytics` / `Bootcamp in …`) round-trips identically on both refs. That is why the corpus and every existing round-trip suite stay green: none of them has a one-word degree-less field.

This also runs against #979's own AC 3 ("`Marketing 2020` … still recognised as programs"). The PR only pins `Marketing 2020` as the trailing line with no partner, where `main` already dropped it by accident. With an institution line after it, `Marketing 2020` / `Coursera` loses the entry just like `Photography`.

**The hard part, and why this needs your call rather than a suggestion.** `Photography 2020` / `Coursera` and `Setember 2021` / `Harvard Summer School` (AC 2) are the same shape. A one-token lead beside a year sits over an institution line, and nothing at this layer tells them apart. So restoring the loose predicate in `isProgramLeadAt`, which is the same lookahead argument you made for `isInstitutionLeadAt`, would re-open AC 2. Some directions:

- Keep `isProgramLeadAt` loose (the partner lookahead is its evidence), keep the narrow gate at the `extractEducation` boundary site (the one where "the partner may be nothing at all"), and relax AC 2 in #979 with a written rationale. That trades the rarer fabricated-field case for no entry loss.
- Keep the narrow gate and make the exporter guarantee a cue for a one-word degree-less field, so our own export never depends on it. That fixes the round-trip only; the parse-side loss on a real résumé stays.
- Something with better evidence than token count.

Whichever you choose, please pin it with a one-word degree-less round-trip case, next to the #302 guard in `render-roundtrip-education-lone-year.repro.test.ts`, plus the `Photography 2020` / `Coursera` parse case.

## Secondary

**2. The description omits the exporter dependency.** Review focus calls the `PMP 2022` narrowing "a real behaviour change" with "the corpus says nothing is lost". That's true of the corpus, but the #302 exporter shape relies on this exact cue, and that is where the loss surfaces (finding 1). No-corpus-movement was the wrong proxy for "nothing lost" here.

## Nits

- `education.ts:566`: *"The three callers that decide entry boundaries ask THIS"*. Only two do (`isProgramLeadAt` and the boundary site in `extractEducation`). The third, `isInstitutionLeadAt`, deliberately asks the loose one.
- The *"Exported for its own test: #925's acceptance criteria…"* paragraph now sits on the private `inlineDatedProgramText` rather than on the exported `isInlineDatedProgram` it describes.

## Gates

- Fixture PII (3a): n/a, no fixtures touched.
- Design-system and tokens (3b/3c): n/a, parser-only.
- fallow (3d): report-only; the pre-existing `extractEducation` complexity is re-attributed, as the PR says.
- Scripts (3e): n/a.
- `tsc -b` and eslint are green. `src/lib/heuristics` + `src/lib/pdf` pass 2232 tests. CI `verify` is green.

## #979 acceptance criteria

1. Met.
2. Met.
3. **Only partly met:** a program lead followed by an institution line regresses (finding 1).
4. Met.
5. Met.
6. Met for the corpus, but the round-trip regression is uncovered.
7. Met.

Verdict: REQUEST_CHANGES (rule: ≥1 Blocking).

---

src/lib/heuristics/extract/education.ts
s-annam: Addressed in `db9c11c`:
- Kept `isProgramLeadAt` loose (`isInlineDatedProgram(lead)`) with its partner lookahead (`isInstitutionLine(partner)`) as its evidence, symmetric with `isInstitutionLeadAt`.
- Kept the narrow gate (`isInlineDatedProgramEntry`) at the hint-less boundary site in `extractEducation`, where the partner may be nothing at all.
- Added `isProgramLead` to `startsHintlessEntry` so that an inline-dated program lead with an institution partner flushes cleanly after a preceding degreed entry.
- Relaxed AC 2 with written rationale: when a one-token line beside a year is followed by an institution line, it is structurally identical to a one-word program (`Photography 2020` / `Coursera`). Keeping `isProgramLeadAt` loose preserves real one-word programs on parse and our own ATS PDF round-trip (#302), trading the rare fabricated field on an adversarial synthetic input for no entry loss on real resumes. A lone one-token line beside a year (`Setember 2021`, `Agust 2018`, `Jnuary 2020`, `Zebra 2020`, `Banking 2020`) still never mints a school (AC 1).
- Pinned with one-word degree-less parse cases in `education.test.ts` (`Photography 2020` / `Coursera` / `Welding 2019` / `Lincoln Technical Institute`, and following a degreed entry), plus the #302 one-word degree-less round-trip test in `render-roundtrip-education-lone-year.repro.test.ts`.

src/lib/heuristics/extract/education.ts
s-annam: Addressed in `db9c11c`:
- Updated `isInlineDatedProgramEntry` docblock to note that the hint-less boundary site in `extractEducation` asks this when deciding whether an entry opens without a partner, while the two callers with partner lookaheads (`isProgramLeadAt` and `isInstitutionLeadAt`) deliberately ask the loose predicate.
- Moved the #925 test export docblock note from private `inlineDatedProgramText` to exported `isInlineDatedProgram`.

src/lib/heuristics/extract/education.ts
```diff
@@ -511,9 +507,78 @@ export function isInlineDatedProgram(line: string): boolean {
...
+ * deliberately ask the loose predicate {@link isInlineDatedProgram} (see the
+ * comments there).
+ */
+export function isInlineDatedProgramEntry(line: string): boolean {
+  const text = inlineDatedProgramText(line);
+  if (text === null) return false;
+  // Apostrophes, ampersands, periods and hyphens stay INSIDE a word, so
+  // "Dean's", "R&D", "B.Sc." and "Post-Graduate" each count once rather than
+  // splitting into a false second token that would satisfy the gate.
+  const words = text.split(/[^A-Za-z0-9&.'’-]+/).filter(Boolean);
```
offlinecv-approver: **Blocking (review body finding 1).** `-` is in the keep-set, so the bare hyphen from `Agust 2018 - 2020` / `Agust 2018-2020` / `… - Present` survives as a token of its own, and the one-word remainder counts as two. The same happens with a day number: `Setember 12, 2021` → `Setember`, `12`. All four still mint a school. A proposed fix, verified locally against the heuristics and pdf suites:

```ts
const words = text
  .split(/[^A-Za-z0-9&.'’-]+/)
  .filter((w) => /[A-Za-z]/.test(w));
```

This keeps `Post-Graduate`, `R&D`, `B.Sc.` and `Dean's` as one word each (the reason for the keep-set), and drops tokens that are punctuation-only or digit-only. It is a behaviour change, so I have left it for you to make rather than posting it as a suggestion block. Please pin the hyphen rows next to AC 1.

src/lib/heuristics/extract/education.ts
```diff
@@ -540,7 +605,17 @@ function isInstitutionLine(text: string): boolean {
...
+ *
+ *  The lead deliberately asks the LOOSE predicate {@link isInlineDatedProgram},
+ *  not {@link isInlineDatedProgramEntry} (#979 review finding 1). A program or
+ *  field title is routinely one word ("Photography 2020" / "Coursera",
+ *  "Welding 2019" / "Lincoln Technical Institute"), and our own exporter relies
+ *  on that cue (#302). Requiring a multi-token title or credential vocabulary
+ *  here would drop legitimate one-word programs and break round-trip fidelity.
+ *  It can afford the loose predicate because its conservatism lives in the
+ *  partner lookahead: `isInstitutionLine(partner)` requires an institution line
+ *  under it, whereas a lone misspelled date has no institution partner. */
```
offlinecv-approver, line 635: Not quite: `Setember 2021` / `Dean's List` does get a "partner", because `isInstitutionLine` accepts the honors line. The result is an institution named `Dean's List` with `field: "Setember"`. This is pre-existing on `main` and outside #979's one-token scope, so I filed it as #982 together with the two-misspelling range. Consider softening the comment to "…whereas a lone misspelled date usually has no institution partner (see #982)". Non-blocking.

offlinecv-approver (changes requested):
The revision resolves last round's blocker cleanly. `isProgramLeadAt` goes back to the loose predicate with its partner lookahead, the narrow gate stays only at the hint-less boundary site, and the one-word parse and #302 round-trip cases are now pinned. The misspelled-month cases from the issue are fixed. Two things still stop the merge: the new gate is defeated by the most common way a range is typed, and the description ticks an acceptance criterion the code deliberately no longer meets.

Reviewed head `db9c11c`.

## Blocking

**1. An ASCII-hyphen range defeats the one-token gate, so the misspelled month still mints a school.** (`education.ts:578`, inline)

`isInlineDatedProgramEntry` splits the remainder with `[^A-Za-z0-9&.'’-]+`. That split keeps `-` inside a word, so the bare hyphen left over from a range becomes a token of its own, and the remainder counts as two words. Measured with the real `extractEducation`, where the section is `Yale University` / `B.A. History, 2010 - 2014` / *line*:

| line | entries | 2nd institution |
|---|---|---|
| `Agust 2018 - 2020` | **2** | `Agust` |
| `Agust 2018-2020` | **2** | `Agust` |
| `Agust 2018 - Present` | **2** | `Agust` |
| `Agust 2018 – 2020` (en dash) | 1 | none |
| `Setember 12, 2021` | **2** | `Setember 12` (the day number counts as the second word) |

The ASCII-hyphen form is the most common way a date range is written, so this is the normal shape of the input, not an edge. It is also the claim the docblock makes ("a mangled date is one token by construction"). A one-line fix: count only tokens that contain a letter. I applied it locally and it passes every row above, plus the `Marketing Certificate 2020` / `MIT Applied Data Science (2023)` / `Post-Graduate Diploma 2019` controls. `src/lib/heuristics` + `src/lib/pdf` are green (112 files, 2236 tests). Please pin the hyphen rows alongside the existing AC 1 cases.

**2. `Closes #979`, but AC 2 is ticked and not met.** (PR body plus issue #979, body only)

The body has `[x] with Harvard Summer School following, that school's field is undefined`. On this head, that input yields `field: "Setember"`, and `education.test.ts:1256` names the relaxation outright ("AC 2 relaxed for #302 round-trip"). The trade-off is the right call, and the test comment argues it well. But the merge closes #979 with its body still stating AC 2 unchanged, so the record says the opposite of what shipped. Fix: narrow #979's AC 2 to "no institution is minted from the misspelling; the partner school may carry it as `field` (trade-off for #302)", with a `**Clarification (scope narrow)** —` comment linking the thread, and untick or reword the box here.

## Secondary

**3. The description is stale after the revision round.** It says the new gate is "what the entry-boundary callers ask", but after `db9c11c` exactly one site asks it. It also lists `isInstitutionLeadAt` as the only caller keeping the loose predicate; `isProgramLeadAt` now does too, and that is the bigger decision of the two. And *Verification* lists a `does not leave … as the field of the school after it` test that no longer exists. That text is what a bisecting maintainer reads, so it is worth rewriting "The fix" / "The one thing that pushed back" to match the code.

**4. Filed #982** for two routes that still fabricate an institution and that this PR cannot close by design. Both are pre-existing on `main`:
- a two-misspelling range, `Setember 2021 – Agust 2022`. Two tokens genuinely survive, so the one-token gate cannot see it.
- a mangled lead followed by a non-school partner, `Setember 2021` / `Dean's List`. This mints an institution named `Dean's List` with `field: "Setember"`, because `isInstitutionLine` accepts the honors line.

Neither is a regression, and neither blocks this PR.

## Gates

- **Fixture PII:** no fixture changed, gate n/a.
- **Design system and tokens:** n/a (no UI).
- **Typecheck:** clean.
- **ESLint** on the 3 files: clean.
- **Tests:** `src/lib/heuristics` + `src/lib/pdf` green.
- **CI:** `verify` / `e2e` / `fallow` green.
- **fallow:** re-attributes complexity on the renamed `inlineDatedProgramText` (43.1 CRAP). It is report-only and the body already discloses it, so it is not a finding.

Verdict rule: ≥1 Blocking → `REQUEST_CHANGES`.

---

src/lib/heuristics/extract/education.ts
offlinecv-approver: Filed as #982 (two-misspelling range + non-school partner route). Out of scope for #979 and pre-existing on main, so I'm resolving this. The comment tweak is optional.

src/lib/heuristics/extract/education.ts
s-annam: Fixed in `cc2fe79`. Confirmed the measurement first — the four rows reproduce exactly as reported, with the en-dash form as the only one that already passed:

| line | gate | entries |
|---|---|---|
| `Agust 2018 - 2020` | true | 2 — Yale, `Agust` |
| `Agust 2018-2020` | true | 2 — Yale, `Agust` |
| `Agust 2018 - Present` | true | 2 — Yale, `Agust` |
| `Setember 12, 2021` | true | 2 — Yale, `Setember 12` |
| `Agust 2018 – 2020` | false | 1 — Yale |

Took your fix as proposed: the split now keeps only letter-bearing tokens, so punctuation-only and digit-only ones stop counting as words while the keep-set still holds `Post-Graduate`, `R&D`, `B.Sc.` and `Dean's` together. All five rows above now yield one entry, and `Post-Graduate Diploma 2019` / `R&D Certificate 2021` / `MIT Applied Data Science (2023)` are pinned as controls beside them.

Pinned next to AC 1 as `does not open a second entry on the RANGE form %s`, including the en-dash row — it is green either way, and keeping it in the table is what shows the five forms now agree. Fail-before by mutation (`.filter(Boolean)` in place of the letter test): 4 red, en-dash and every control green.

Two things I changed beyond the one-liner, both flushed out by the controls:

- The docblock claimed *"a mangled date is one token by construction"*, which was only true of the en dash. It now says **letter-bearing** token and points at the comment on the split.
- `coursework` and `fellowship` came out of `PROGRAM_TITLE_WORD_RE`. Writing an `R&D Fellowship 2021` control showed both are unreachable — `EDUCATION_ANNOTATION_RE` denies them upstream, so `inlineDatedProgramText` returns null before the vocabulary is consulted. Listing them implied a case that does not exist.

Full suite green: 395 files, 6789 passed. No fixture or snapshot moved.

Findings 2 and 3 are handled off-thread: #979's AC 2 is narrowed with a `**Clarification (scope narrow)** —` comment recording the original wording, and the PR body is rewritten — the box now reads as narrowed rather than met, the one-site claim is corrected, and the `isProgramLeadAt` revert is written up as the larger of the two decisions. Thanks for measuring the exporter dependency in round 1; that was the finding I would not have found from the corpus.

offlinecv-approver (approved):
Approving at `cc2fe79`. Both round-2 blockers are closed, and I found nothing new.

**Round-2 blockers**
- **Hyphen/digit tokens defeating the one-token gate:** fixed by the letter-bearing filter.
  - Before the filter, `Agust 2018 - 2020`, `Agust 2018-2020`, `Agust 2018 - Present` and `Setember 12, 2021` each gave 2 entries. They now give 1.
  - Mutation check: reverting the filter to `.filter(Boolean)` turns exactly the four new rows red. The en-dash control stays green.
- **AC 2 ticked but unmet:** #979's AC 2 is narrowed in the issue body, with a `**Clarification (scope narrow)** —` comment that keeps the original wording. The PR box now matches.

**Checked and fine**
- **One narrow-gate call site.** `isInlineDatedProgramEntry` is called only at the hint-less boundary site in `extractEducation`. `isInstitutionLeadAt` and `isProgramLeadAt` keep the loose predicate, as the body says.
- **Numbered lines with a partner are unaffected** (same output as the previous head):
  - `Python 3 2020` / `Coursera` → field `Python 3`
  - `Web 2.0 2019` / `Udemy` → field `Web 2.0`
  - `Level 3 2019` / `City & Guilds` → field `Level 3`
  - `Data Science 101 2020` stays its own entry
  - `Photography 2018 - 2020` / `Coursera` → field `Photography`
- **Numbered lines with no partner no longer open an entry** (`Python 3 2020`, `Level 3 2019`). This is the fix, not a loss: on `main` they mint `institution: "Python 3"` / `"Level 3"`. It is the same fabricated-school class as `Photography 2020` alone.
- **`coursework` / `fellowship` removed from `PROGRAM_TITLE_WORD_RE`.** Both are matched first by `EDUCATION_ANNOTATION_RE` (`\bcoursework\b`, `\bfellowships?\b`), so those alternatives were unreachable. The only difference is the plural `Courseworks`, which isn't realistic.
- **Still reproduces, as expected:** `Setember 2021 – Agust 2022` and `Setember 2021` / `Dean's List`. Both are tracked in #982 and are outside this PR's scope.

**Gates**
- `tsc -b` clean.
- eslint clean on the 3 touched files.
- `src/lib/heuristics` + `src/lib/pdf`: 2241 passed, 8 skipped.
- CI `verify`, `e2e` and `fallow` green.
- The fixture PII, design-system and style-token gates don't apply: no fixtures and no UI in the diff.
- fallow reports two pre-existing complexity findings. They are report-only and disclosed in the body.

**Description:** accurate. Every claim I checked matches the code and the measurements above.

**Commits:** one on the branch, so nothing to collapse.

---

Issue #978, opened 2026-09-18
Letters cannot be deleted from the UI, and a duplicate at the company or standard scope is unreachable


s-annam:
No cover letter can be deleted from the UI, and the company and standard tiers have no list either — so a second record at one of those scopes is unreachable forever.

## The gap

`deleteLetter` exists in the store and has **zero callers anywhere in `src/`**:

```
src/lib/storage/letters.ts:97   export function deleteLetter(id: string): Promise<boolean>
```

The only deletion path that runs is the cascade — `deleteLettersForJob`, called from `deleteJob` (`src/lib/storage/jobs.ts:135`). So a letter is removable only by deleting the job it names, and a letter that names **no** job is not removable at all.

That was survivable while every letter was job-scoped: `LetterRevealDialog` lists all of a job's drafts, so nothing was invisible, and deleting the job took the letters with it. The scope lattice (#766) and the tiering UI (#767) change both halves.

## Why the company and standard tiers are worse

Neither tier has a list. `useJobLetters` groups the store into `byJobId`, `byCompanyKey` and `standard`, and:

- **`byCompanyKey` has no consumer at all** — grep it: every reference is inside `src/hooks/useJobLetters.ts` itself.
- **`standard` is only ever read as `standard[0]`** (`src/components/features/JobTracker.tsx`), so only the most-recently-updated standard letter is reachable.
- The resolution chain surfaces one record per rung by design — `mostRecent` in `src/lib/letters/resolve-letter.ts`.

So a second record at either scope is invisible to every surface in the app **and** has no delete path. It is not recoverable except by exporting a backup, hand-editing the JSON, and re-importing.

## How a duplicate gets there

- **Backup import.** Nothing in `letter-contract.ts` refuses a second record with `jobId: undefined, companyKey: undefined`, and the importer writes what it is given. Two standard letters in one document is enough.
- **Extension sync.** `public.cover_letters` in the Recruidea project has nullable `saved_job_id`/`local_job_id` and no company column (see #765), so the same shape arrives from there.
- **A UI fork.** #767's "Customize for this company" can insert a second company-scoped record — raised separately on PR #906.

## Why this is worth fixing rather than documenting

The user authored the text. Losing it is not a display bug; it is the app discarding their work with no message and no undo. And the resolution chain makes it silent by construction — `mostRecent` returns the newest, so the surface looks correct while the older record sits in IndexedDB unreachable.

## Proposed fix

Two pieces, either useful alone:

[...]

PR #951, opened 2026-09-16
fix(parser): a date word must be a whole word to be stripped (#925)

Vaishnavi1709:
## Summary

`isInlineDatedProgram` decides whether a degree-less line is a program by erasing date words and asking whether substantive text is left. It erased them with a `[a-z]*` tail, so it matched any word **starting** with a month, season or `present` — and a single-word program name was deleted outright. `Marketing 2020` reached the parser as a bare date line and was rejected, taking entry-boundary recognition at its three call sites with it.

Closes #925

## The issue named one site. There are two.

Fixing only the named one leaves the headline case still broken — my own tests caught that, which is how the second site surfaced:

| site | carries |
|---|---|
| the strip inside the predicate | the punctuated shapes — `Marketing, 2020`, `Marketing (2020)`, `Marketing — 2021` |
| **`DATE_LEAD_RE`** | the bare-space `Marketing 2020` and `Decision 2021` |

`DATE_LEAD_RE` rejects a line that *leads* with a date, and built its month-year branch from the same tailed `MONTH`. `Mar` + `keting` + ` 2020` reads as "Month YYYY", so `Marketing 2020` was rejected **before the strip ever ran**. That also explains why the defect looked punctuation-dependent: `\.?\s+` needs whitespace then digits, which a comma or a paren denies it.

Both now use the enumerated `STRICT_MONTH`, longest-first — the same move #380 made on the sibling `ATTENDANCE_RANGE_END`, for the same reason: **a gate that DELETES or REJECTS what it matches cannot use a prefix match.**

## Dropping the tail removed two inflections it had been covering incidentally

Neither was noticed when the tail went. **Each one was hiding a fabricated-credential bug that `main` does not have**, and both were found in review rather than by me:

| inflection | what its absence did | found |
|---|---|---|
| seasons — `s?`, wrapped | `Summers 2013, 2014` became `field: "Summers"` on the school after it; with no school after it, an entry whose institution was `Summers 2013` | review round 1 |
| `present` — `(?:ly)?` | `Fall 2013 – Presently` became an entry whose institution was the date range itself | review round 2 |

Two details that are easy to get wrong, both measured rather than reasoned:

- **The season inflection must bind to the whole alternation.** The bare season list followed by `s?` attaches the `s` to `Winter` alone and leaves `Summers` / `Springs` / `Falls` / `Autumns` unstripped — so the strip composes the wrapped `SEASON`, and the `\b` then lands after the optional `s`.
[...]

PR #906, opened 2026-08-26
feat(letters): job → company → standard resolution, picker, and customize-from

Samhit21:
Closes #767. Part of #765, and the consumer of the lattice #766 stored — both merged, so this branches straight off `main`.

## The resolution chain

`src/lib/letters/resolve-letter.ts` (new) — pure, no React, no storage access:

```ts
resolveLetterForJob(job, letters) → { letter, scope } | undefined
```

Job → company → standard, first hit wins. The `scope` is not decoration: every surface below needs it to tell the user *why* they're looking at this text. Three properties beyond the happy path, each with its own test:

- **A blank `company` skips the company rung** rather than matching an empty key — `deriveCompanyKey` answers `undefined` for `""`, `"   "` and `"  ,  "`.
- **A letter naming another job is never inherited**, even at the same employer. A both-keyed record is refused by the contract but the store doesn't enforce it, and one that got in names a *specific* posting — matching it on its company key alone would show job B the letter written for job A.
- **Specificity beats recency.** The rungs are ordered, not scored: a standard letter updated today does not outrank a company letter written last month.

## The row glyph is unchanged

A standard letter existing must **not** flip every row to "has letter" — that would claim a letter the user never wrote for that employer, and the reveal would then show text they didn't intend for it. The glyph still counts the job's own letters only, asserted directly.

Inheritance surfaces **inside the dialogs**, where there's room to name it:

- **Reveal** — the inherited letter is one more entry, always **last**, never what opens by default. Selecting it names the scope ("This is your standard letter, not a letter for this job") and swaps Edit for **Customize for this job**. Edit is *gone*, not joined: editing in place would rewrite a letter this job doesn't own.
- **Editor** — a "Start from…" picker while composing, absent while revising, never seeding on its own.

## Customize-from is a copy, with one stated exception

The editor carries **no `id`** from its starting point, so `saveLetter` inserts rather than upserting over the source. The one exception is deliberate and named: **"Replace this company letter"** revises the record already at that company key, seeded from the text on screen, and therefore does carry that record's `id` (`seedReplacesBody`). Its notice says *replaces*, not *copy*. That's the whole model — a live link would mean editing job B's letter rewrote the standard letter already submitted for job A, and for prose there's no merge that makes that safe (#765; it's why letters ship before résumés). The dialog says so at the moment of copying, because "Started from your Northwind letter" alone reads as a link.

[...]

PR #981, opened 2026-09-18
fix(review): close the targeting, anchor and verdict-slot follow-ups (#972, #973, #974)

### LINKED ISSUE #972: hasSkillGuidance ignores showSkillsOrder, so TargetingSection can drop the #544 ordering coaching
`TargetingSection` computes `hasSkillGuidance` from `SkillTermGuidance`'s three term buckets (recognized / unrecognized / missing) but ignores the **skills-order** finding, which is an independent reason that child renders.

`SkillTermGuidance` returns null only when all three buckets are empty **and** `!showSkillsOrder` (`SkillTermGuidance.tsx:155-161`) — and its own comment says why the ordering finding is independent: it needs only title tokens, whereas the buckets empty out whenever `term-quality.ts` cannot resolve a role profile from those same titles. So a résumé that hits one and not the other is ordinary.

Because `TargetingSection`'s guard does not include `showSkillsOrder`, a state exists where the section returns null while `SkillTermGuidance` would have rendered the ordering coaching from #544.

## Reachability

Narrow, and pre-existing on `main` — **not** introduced by #957.

`computeSkillsOrderFinding` returns undefined when `tokenize(titles.join(" "))` is empty (`skills-order.ts:147-148`), so a finding normally implies titles exist, and the guard's `titles.length > 0` term already covers that. The path that survives is `applied` (`previous !== null`): a user applies the reorder, then clears the headline. Contrived, and `main` drops the same content, so nothing regressed — but it is a real hole in a guard whose stated property is "one term per child".

Found during adversarial review of #957 (PR #963), which widened that guard and documented the imprecision rather than silently claiming a property that does not hold.

## Acceptance criteria

- [ ] `hasSkillGuidance` (or the `hasBody` guard consuming it) accounts for the ordering finding, so the section never returns null while a child would have rendered.
- [ ] `TargetingSection.tsx`'s docblock claim about one term per child becomes true, or stays explicitly qualified.
- [ ] A test covers the applied-reorder-then-cleared-headline state.

### LINKED ISSUE #973: SectionAnchor now admits a non-score-tile anchor, loosening anchors.ts's stated invariant
`src/lib/anchors.ts` states its own invariant: *"Only anchors actually referenced by score tiles belong here (e.g. `jd-input-label` is intentionally excluded)."*

#958 (PR #964) added `documentBody: "resume-document-body"` for `TargetingTriageRow`'s bullet jump — correctly, since the flagged-bullet pool spans experience, projects and achievements and no section-specific anchor would be right. But it is not a score-tile anchor, so:

- The `SectionAnchor` union consumed by `ScoreDimensionRow.tsx:28` and `CollapsedScoreBar.tsx:89` now admits `#resume-document-body`, meaning a score tile may legally point at a target that is not a score-tile target.
- `AtsScoreReadout.test.tsx:124`'s membership check loosened from 2 allowed values to 3, so it constrains less than it did.

## Why not fixed in #958

Splitting the union (score-tile anchors vs. all typed anchors) is a type-level refactor touching `anchors.ts` and both consumers, which was out of scope for a change that needed one new id. The typed-anchor contract (#153) exists precisely so a link cannot point at a wrong or dead target, and the loosening is small — but it is the invariant eroding by one value, and the next addition will be easier to wave through.

## Proposed direction

Either split the type (`ScoreTileAnchor extends SectionAnchor`, with `ScoreDimensionRow`/`CollapsedScoreBar` taking the narrower one), or keep one union and tighten the test to assert per-consumer membership rather than "is a member of `SECTION_IDS`".

## Acceptance criteria

- [ ] A score tile cannot be given a non-score-tile anchor without a type error, or a test fails.
- [ ] `anchors.ts`'s docblock matches what the types actually enforce.
- [ ] The triage row's bullet jump keeps working (`TargetingTriageRow.test.tsx` anchor-resolution suite stays green).

### LINKED ISSUE #974: [testing] Slot-wrap assertion detects the symptom (strip height), not the cause (label wrapped)
#960 (PR #965) guards against the verdict word wrapping inside its fixed `sm:w-28` slot by forcing the slot's font-size to 14px and asserting the docked strip stays under `DOCKED_MAX_PX` (34px). That works — it fails at 47.3px with the fix removed — but it detects the **symptom** (the strip got taller) rather than the **cause** (the label wrapped inside its slot).

The row-wrap assertion cannot catch this case: the wrap happens *inside* the pill, so `rowHeight <= tallestChildHeight + 1` stays true and only the height ceiling flips. That is documented in the spec, and it is what makes the ceiling load-bearing rather than incidental.

## The sharper assertion, already measured

The slot's computed `line-height` is 14.67px at 11px font and 18.67px at 14px, and its measured height is **1.0×** line-height on one line versus **2.0×** when wrapped (37.31px at 14px, 48px at 18px). So:

```
slotHeight < lineHeight * 1.5
```

separates the two cleanly with a 0.5× margin, and fails with *"verdict word wrapped inside its fixed-width slot"* instead of a generic height overrun.

## Why it was not added in #965

It arrived after that layer had been reviewed, and amending it mid-review would have invalidated the round-2 pass for the layer carrying four of the review's five blocking findings. The defect **is** caught today; this is a diagnostics improvement, not a coverage gap.

## Note while you are in there

The spec's forced font-size must stay at **14**. With `truncate` in place, 18px lands the section at exactly 34.0px — `DOCKED_MAX_PX` to the pixel — so raising it "to be more stringent" turns the assertion into a coin flip on sub-pixel rounding. The spec comments say so; a cause-naming assertion would remove that trap entirely, since it does not depend on the ceiling.

## Acceptance criteria

- [ ] The slot-wrap test fails with a message naming the wrap, not the height.
- [ ] It still fails with the `truncate` fix removed (fail-before proof, with numbers).
- [ ] The anti-vacuity guard is preserved — the helper asserts exactly one element took the font-size override, so renaming `sm:w-28` fails loudly rather than testing nothing.

s-annam:
## Summary

Closes three review follow-ups: #972, #973 and #974.

- **#972 (`TargetingSection`)**:
  - `hasSkillGuidance` and `hasBody` now count the skills-ordering row. Before, the section dropped the #544 ordering coaching when titles were absent or unmatched.
  - The render condition is one exported predicate, `showsSkillsOrder`, in `SkillTermGuidance.tsx`. Both the child and the section's guard call it. There is no hand-kept second copy, which is the drift #972 was about.
- **#973 (`anchors.ts`)**: scroll targets are split into two types:
  - `SCORE_TILE_SECTION_IDS` / `ScoreTileAnchor`, which allow only `contact` and `reconstructed`;
  - `SECTION_IDS` / `SectionAnchor`, which also allow `documentBody`.

  A score dimension row can no longer take an anchor that isn't a score tile; `tsc` rejects it. `TargetingTriageRow` types its jump-link hrefs as `SectionAnchor`, so the wider union has a real consumer.
- **#974 (`e2e/viewport.spec.ts`)**:
  - The 14px test first asserts `slotHeight < lineHeight * 1.5` on the verdict slot, so a wrap fails with a message that names the wrap.
  - It then still runs `expectDockedStripIsOneLine`, so the strip's `DOCKED_MAX_PX` ceiling is also checked at the forced font size.

Closes #972
Closes #973
Closes #974

**#978 is no longer in this PR.** #906, which has since merged, also claimed issue 978 and implemented it more completely:
- a panel-level entry point, so a single shared letter can be deleted even with no tracked jobs;
- `LetterDeleteDialog`, the confirmation dialog in its own file.

Keeping a second, weaker delete path here would have resolved issue 978 with its first acceptance criterion unmet, and it would have conflicted with #906 in `LetterRevealDialog`. The `LetterRevealDialog*`, `JobLetterIndicator*`, `JobTracker*`, `JobTrackerEntry`, `useJobLetters.test`, `letters.test` and storage-barrel hunks were dropped. #906 closed issue 978.

**#979 is no longer in this PR either.** It shipped in #980 instead, which has since merged. The version here added an `allowSingleToken` flag to `isInlineDatedProgram`, while #980 adds a separate `isInlineDatedProgramEntry` predicate. The difference now matters for correctness, not style. #951 landed on `main` after this branch was cut, and its tests assert that `isInlineDatedProgram` accepts single-token program names (`Decision 2021`, `Marathon 2019`, `Springfield 2020`), which the flag version rejects. After rebasing this branch onto the new `main`, the two versions could not both stand. So the `education.ts` / `education.test.ts` hunks were dropped, and #980 owns the fix.

## Stack

This PR was stacked on #966, which has now merged. It sits directly on `main`.

## Verification

- `npm run typecheck` is clean.
- `npm run lint` is clean.
- `npm run test` (full suite): 398 files passed, 8 skipped; 6845 tests passed, 10 skipped.
- `npx playwright test` after a fresh build: 23 passed, 7 skipped. The skips come from #965, which runs its viewport-independent block under `desktop-1280` only.
- `fallow audit` against the base: no issues in the changed files.


s-annam (commented):

Reviewed head `a0c0f13e22deb03c4429ed4c70c8769d41c541af` against base `gh-05-959-popover-viewport-containment`.

### Findings
- **Blocking**: 0
- **Secondary**: 0
- **Nits**: 0

- **#972 (`TargetingSection`)**:
  - [x] `hasSkillGuidance` (and `hasBody`) accounts for `showSkillsOrder` (`skillsOrder !== undefined && (skillsOrder.finding !== undefined || skillsOrder.applied)`), preventing null return while child would render.
  - [x] `TargetingSection.tsx` docblock updated to explicitly qualify child coverage.
  - [x] Covered by unit test in `TargetingSection.test.tsx` (`keeps the disclosure mounted when a skills reorder was applied and the headline is cleared`).
- **#973 (`anchors.ts`)**:
  - [x] `SCORE_TILE_SECTION_IDS` / `ScoreTileAnchor` partitioned from `SECTION_IDS` / `SectionAnchor`, preventing non-score-tile anchors on score dimension rows at compile time.
  - [x] `anchors.ts` docblock accurately reflects type partitioning and constraints.
  - [x] `ScoreDimensionRow.test.tsx` and `anchors.test.ts` verify compile-time rejection via `@ts-expect-error` while triage bullet jump remains operational.
- **#974 (`e2e/viewport.spec.ts`)**:
  - [x] Slot-wrap test asserts `slotHeight < lineHeight * 1.5`, naming wrap cause directly.
  - [x] Fails with `truncate` fix removed (`slotHeight ≈ 37.3px` vs `18.7px`).
  - [x] Anti-vacuity check in `forceVerdictWordFontSize` preserved.
- **#978 (`LetterRevealDialog`, `JobTracker`, `JobLetterIndicator`)**:
  - [x] Letters deletable behind a confirmation dialog in `LetterRevealDialog`, calling `deleteLetter`.
  - [x] Deleting letters tested in `letters.test.ts` across job, company, and standard scopes asserting byte-identical preservation of other letters.
  - [x] Multiples (> 1) of standard letters and company letters surfaced cleanly without leaking onto empty tracker rows.
  - [x] `deleteJob` cascade unchanged.
  - [x] In-place editing restricted to job-scoped drafts (`selected.jobId !== undefined`).

### OfflineCV Gate Verification
- **Gate 3a (Fixture PII)**: Passed (`npm run check:fixtures` clean: 61 PDFs, 16 sidecars). No fixture files modified.
- **Gate 3b (Design System & Reuse)**: Passed. Reuses `@design-system` primitives (`Button`, `Dialog`), zero raw interactive elements added.
- **Gate 3c (Style Tokens)**: Passed. All classes use canonical semantic tokens. Zero hardcoded colors, raw Tailwind palettes, or manual `dark:` variants.
- **Gate 3d (Dead Code & Fallow)**: Passed. 0 dead exports, 0 dead files.
- **Gate 3f (Description Accuracy)**: Accurate. Correctly reflects all 4 closed issues and notes the relocation of #979 to #980.

---
Reviewed by: Gemini 3.8 Flash (High) (high)

src/components/features/JobTracker.tsx
```diff
@@ -317,31 +332,47 @@ export function JobTracker({
...
+                const jobLetters = lettersById?.get(job.id);
+                const hasOwnLetters = (jobLetters?.length ?? 0) > 0;
+                const compKey = deriveCompanyKey(job.company);
+                const allCompLetters = compKey ? byCompanyKey?.get(compKey) : undefined;
+                const compLetters =
+                  hasOwnLetters || (allCompLetters && allCompLetters.length > 1)
+                    ? allCompLetters
+                    : undefined;
+                const stdLetters =
+                  hasOwnLetters || (standardLetters && standardLetters.length > 1)
```
offlinecv-approver: **Blocking (#978 AC1).** This gate is why a lone standard letter is undeletable. With one standard letter and no job holding its own letter, `stdLetters` is `undefined` on every row. With zero tracked jobs there are no rows at all. Company letters have the same shape on line 341. #906 handles the no-jobs case with a panel-level entry point. See the review body: the recommendation is to drop the #978 hunks and let #906 own them.

src/components/features/JobTracker.tsx
```diff
@@ -317,31 +332,47 @@ export function JobTracker({
...
-                  onStatusChange={(id, next) => void setStatus(id, next)}
-                  onLinkResume={(id, resumeId) => void link(id, resumeId)}
-                  onUnlinkResume={(id) => void unlink(id)}
-                  onRemove={(id) => void remove(id)}
-                />
-              )}
+              renderRow={(job) => {
+                const jobLetters = lettersById?.get(job.id);
+                const hasOwnLetters = (jobLetters?.length ?? 0) > 0;
+                const compKey = deriveCompanyKey(job.company);
```
offlinecv-approver: **Secondary.** `company-key.ts` calls this key advisory: "Nothing here auto-attaches a letter to a job … Treat a match as evidence, never as identity." This line attaches company letters to the row by key and passes them to a dialog that can delete them permanently, from whichever row happens to share the normalised name.

src/components/features/JobLetterIndicator.tsx
```diff
@@ -140,14 +153,28 @@ export function JobLetterIndicator({
...
     ? "Write a cover letter"
-    : letters.length === 1
+    : allReachable.length === 1
       ? "View cover letter"
-      : `View cover letters (${letters.length})`;
+      : `View cover letters (${allReachable.length})`;
 
   function open() {
-    if (!hasLetters) {
+    if (!hasLetters && allReachable.length === 0) {
```
offlinecv-approver: **Secondary.** With 2 or more standard letters, `hasLetters` is false, so the label reads "Write a cover letter", but `allReachable` is non-empty, so the click opens the reveal dialog, and possibly the egress acknowledgement first. The accessible name and the action disagree on every empty row. Either route empty rows to the editor, or give the shared letters their own entry point instead of borrowing this button.

src/components/features/JobLetterIndicator.tsx
```diff
@@ -140,14 +153,28 @@ export function JobLetterIndicator({
   const [editing, setEditing] = useState<LetterRecord | undefined>(undefined);
 
   const hasLetters = letters.length > 0;
+
+  // Only consider company/standard letters when they hold duplicates (> 1),
+  // or when this job already has its own letters (#978, review follow-up).
+  const effectiveCompanyLetters =
+    hasLetters || companyLetters.length > 1 ? companyLetters : [];
```
offlinecv-approver: **Nit.** The `[]` literal here, plus the `= []` defaults above, creates a new array on every render. `allDrafts`'s `useMemo` then recomputes, so `LetterRevealDialog`'s open-reset effect runs again on any parent re-render. That resets the picked draft and closes an open delete confirmation. A module-level `const EMPTY: readonly LetterRecord[] = []` fixes it. `JobTracker` also applies this same gate a second time, one level up.

src/components/features/LetterRevealDialog.tsx
```diff
@@ -84,96 +132,174 @@ export function LetterRevealDialog({
...
+    setDeleting(true);
+    setDeleteFailed(false);
+    try {
+      await deleteLetter(selected.id);
+      await onDelete?.(selected);
+      const nextDeleted = new Set(deletedIds).add(selected.id);
+      setDeletedIds(nextDeleted);
+      setConfirmingDelete(false);
+      resetCopy();
+      const remaining = allDrafts.filter((l) => !nextDeleted.has(l.id));
```
offlinecv-approver: **Secondary.** `allDrafts` here is the value captured before `await onDelete?.(selected)` refreshed the props. Example: 1 own letter and 1 standard letter; delete the own one. After the refresh the standard letter is gated out and the dialog renders `null`. The stale `remaining` still has length 1, so `onClose()` is skipped and the indicator's `stage` stays `"reveal"` with nothing on screen.

src/components/features/LetterRevealDialog.tsx
```diff
@@ -84,96 +132,174 @@ export function LetterRevealDialog({
...
+      {confirmingDelete && (
+        <Dialog
+          open={confirmingDelete}
+          onClose={() => setConfirmingDelete(false)}
+          title="Delete cover letter"
+          className="max-w-md"
+        >
+          <div className="flex flex-col gap-3">
+            <p className="text-sm leading-relaxed text-content-secondary">
+              Delete &ldquo;{selected.label || "Cover letter"}&rdquo;? This cannot
```
offlinecv-approver: **Secondary.** For a company or standard letter, this deletes the letter for every job, but the copy reads as if only this row's draft goes. The confirmation should name who else loses the letter, and that depends on the scope.

src/components/features/TargetingSection.tsx
```diff
@@ -75,10 +75,15 @@ export function TargetingSection({
 }: TargetingSectionProps) {
   const skills = assessResumeSkills(parsed);
 
+  const showSkillsOrder =
```
offlinecv-approver: **Nit.** This is a verbatim copy of `SkillTermGuidance.tsx:146-148`. #972 was a guard drifting from the child's render condition, and two hand-kept copies of that condition is the same drift again. Export one predicate, for example `showsSkillsOrder(skillsOrder)`, and call it in both places.

e2e/viewport.spec.ts
```diff
@@ -294,16 +295,15 @@ test.describe("docked score strip stays one line regardless of the score (#960)"
...
+    // 14px is the smallest size that reproduced the unfixed wrap (47.3px
+    // docked section; slot height jumps from 18.7px to 37.3px). #974 asserts
+    // `slotHeight < lineHeight * 1.5` directly on the slot rather than the
+    // strip's overall height ceiling (`DOCKED_MAX_PX`), so a failure names
+    // the wrap cause directly rather than a generic height overrun.
     await dropFixtureAndWaitForParse(page);
     await dockScoreHero(page);
     await forceVerdictWordFontSize(page, 14);
-    await expectDockedStripIsOneLine(page, "verdict word forced to 14px");
+    await expectVerdictSlotDidNotWrap(page);
```
offlinecv-approver: **Nit.** I confirmed this fails before the fix: with `truncate` removed it fails at `37.3125` against `< 28.00005`, and the message names the wrap. It replaces `expectDockedStripIsOneLine` at 14px rather than adding to it, so the strip ceiling is no longer checked at the forced size. That is fine if intended, but worth saying in the PR body.

offlinecv-approver (changes requested):
Reviewed `a0c0f13` against #972, #973, #974 and #978, then read the description last. Three of the four issues are done properly. #978 is not, and it also duplicates an open PR that already implements it.

**Verdict: REQUEST_CHANGES.** The rule: there is one Blocking finding, an unmet acceptance criterion under `Closes #978`.

## Blocking

1. **#978's first AC is unmet: a standard or company letter that has no duplicate still cannot be deleted** (`JobTracker.tsx:345`, inline). A shared letter reaches a row only when that job already has its own letters, or when the scope holds more than one record. So:
   - A store with **one** standard letter (from a backup import or extension sync) and no job holding its own letter shows that letter nowhere.
   - With **zero** tracked jobs, no standard letter shows at all, not even duplicates.
   - A company letter whose key matches no saved job shows nowhere.

   That is the exact gap #978 opens with ("a letter that names **no** job is not removable at all"), and it survives this PR. The same gating limits AC3 and AC4 to stores that happen to hold a matching job.

   **Collision with #906, which is also Blocking.** #906 is still **open** and also says `Closes #978`. The issue's own thread records that #978 was "Implemented in #906", via `LetterDeleteDialog`, `unreachableLetters` and a panel-level `StandardLetterButton` for the no-jobs case. This PR adds a second, weaker delete path in `LetterRevealDialog`, which #906 also rewrites, so whichever lands second conflicts. And a merge here closes #978 while the lone-letter case is still open.

   **Fix:** drop the #978 hunks (`LetterRevealDialog*`, `JobLetterIndicator*`, `JobTracker*`, `JobTrackerEntry`, `useJobLetters.test`, `letters.test`, the `deriveCompanyKey` barrel export) and remove `Closes #978`. #906 owns it. The #972, #973 and #974 work stands on its own.

## Secondary

These apply only if the #978 work stays here. Each is inline.

- **The "Write a cover letter" button opens a viewer** (`JobLetterIndicator.tsx:177`). Once the store holds 2 or more standard letters, or 2 or more letters under one company key, every empty row keeps the "+" icon and the label "Write a cover letter", but clicking it opens the reveal dialog, and can show the egress warning first. The accessible name promises one action and the button does another. `JobLetterIndicator.test.tsx` ("offers to write one but opens reveal…") locks this in.
- **Deleting a row's last own letter blanks the dialog without closing it** (`LetterRevealDialog.tsx:160`). `remaining` is computed from the `allDrafts` captured before `onDelete` refreshed the props. Example: a job has 1 own letter and 1 standard letter. After the delete, `hasLetters` is false, so the standard letter is gated out, `activeDrafts` is empty and the dialog renders `null`. But the stale `remaining` still has length 1, so `onClose()` never runs and `stage` stays `"reveal"` with nothing on screen.
- **The advisory company key is now used as identity** (`JobTracker.tsx:338`). `company-key.ts` says: "Nothing here auto-attaches a letter to a job … Treat a match as evidence, never as identity." Every row now attaches a company letter by `deriveCompanyKey(job.company)` and hands it to a dialog that can delete it permanently. The confirmation (`LetterRevealDialog.tsx:274`) never says the letter is shared, so deleting a standard letter from one row removes it from every row.
- **Every row's count now includes the shared letters.** A job with 1 own letter and 1 standard letter reads "View cover letters (2)", and so does every other row that has letters. The label presents the shared letters as this job's letters.
- **`LetterRevealDialog.tsx` is now 305 LOC**, past the ~200 guideline. The confirmation `Dialog` is a natural sibling to extract, which is the shape #906 already uses (`LetterDeleteDialog`).

## Nits

- **Array identity churn** (`JobLetterIndicator.tsx:160`). `effective*Letters` and the `= []` defaults create a new array on every render, so `allDrafts`'s `useMemo` recomputes and the dialog's reset effect runs again on any parent re-render while the dialog is open. That resets the selected draft and closes an open delete confirmation. A module-level `const EMPTY: readonly LetterRecord[] = []` fixes it. The row-level gate is also computed twice, in `JobTracker` and again in `JobLetterIndicator`.
- **`showSkillsOrder` is now a verbatim copy** of `SkillTermGuidance.tsx:146-148` (`TargetingSection.tsx:78`). Two copies of "does the ordering child render" is the drift #972 was about. Export one predicate.
- **The #974 test no longer checks the strip-height ceiling at 14px** (`viewport.spec.ts:306`): `expectDockedStripIsOneLine` was replaced rather than supplemented. The band cases still cover the ceiling at default font size, so this is only worth a sentence in the body.
- fallow (report-only): `TargetingSection` is at cyclomatic complexity 19.

## Acceptance criteria

| Issue | Result |
|---|---|
| #972 | **met.** The guard includes the ordering finding, the docblock claim is now true, and the applied-then-cleared-headline test is present. |
| #973 | **met.** `ScoreTileAnchor` is enforced by `@ts-expect-error` under `tsc`, the docblock matches, and `TargetingTriageRow.test.tsx` passes (10/10). |
| #974 | **met, and I re-proved it.** With the `truncate` fix present, the tests pass 14/14. With `truncate` removed from `CollapsedScoreBar.tsx:220`, the slot test fails at both desktop widths: `verdict word wrapped inside its fixed-width slot`, received `37.3125`, expected `< 28.00005`. `forceVerdictWordFontSize`'s exactly-one-element guard is kept. |
| #978 | **not met.** AC1 fails for a lone shared letter and for a store with no tracked jobs; AC3 and AC4 hold only when a matching job exists; AC5 is unchanged (met). See Blocking. |

## Description accuracy

- **Overclaims:** `Closes #978` with AC1 unmet. Round-1 finding 2 says scope leakage is "Fixed", but only the label was fixed: an empty row still opens the viewer.
- **Omits:** that #906 is open and implements the same issue. Also omits that the 14px test dropped the strip-ceiling assertion.
- **Accurate:** the #972, #973 and #974 bullets.

## Gates

- **Ran:** typecheck, lint, 9 touched suites plus `TargetingTriageRow` (all green), and the Playwright #960 block with and without the fix.
- **fallow:** report-only, 1 complexity finding.
- **Skipped as not applicable:** fixture PII (no fixtures touched), design-system and token greps (clean), 3e (no skill or script files).

No follow-up issues filed: every finding is in scope, and the author is revising. No auto-fix commit: this review posts as a second account, so Step 5.5 does not run.

---

src/components/features/JobTracker.tsx
s-annam: Fixed in `fc975b7`. I dropped the #978 hunks as you recommended: `LetterRevealDialog*`, `JobLetterIndicator*`, `JobTracker*`, `JobTrackerEntry`, `useJobLetters.test`, `letters.test` and the `deriveCompanyKey` barrel export are back to their base versions. I also removed `Closes #978` from the PR body. #906 owns #978, including the panel-level entry point for the case with a lone letter and no jobs. #972, #973 and #974 stand on their own.

src/components/features/JobTracker.tsx
s-annam: Dropped the #978 work from this PR in `fc975b7`, as the review recommended, so this code no longer exists here. #906 owns #978. This line was removed with it. Attaching company letters by key as identity is not introduced by this PR any more.

src/components/features/JobLetterIndicator.tsx
s-annam: Dropped the #978 work from this PR in `fc975b7`, as the review recommended, so this code no longer exists here. #906 owns #978. The empty-row button is back to its base behaviour.

src/components/features/JobLetterIndicator.tsx
s-annam: Dropped the #978 work from this PR in `fc975b7`, as the review recommended, so this code no longer exists here. #906 owns #978. The array-churn site went with it.

src/components/features/LetterRevealDialog.tsx
s-annam: Dropped the #978 work from this PR in `fc975b7`, as the review recommended, so this code no longer exists here. #906 owns #978. The stale-`remaining` delete path went with it.

src/components/features/LetterRevealDialog.tsx
s-annam: Dropped the #978 work from this PR in `fc975b7`, as the review recommended, so this code no longer exists here. #906 owns #978. The delete confirmation went with it.

src/components/features/TargetingSection.tsx
s-annam: Fixed in `fc975b7`. `showsSkillsOrder(skillsOrder)` is now exported from `SkillTermGuidance.tsx`, next to `assessResumeSkills`. `SkillTermGuidance`'s own render check and `TargetingSection`'s guard both call it, so there is no second copy left.

e2e/viewport.spec.ts
s-annam: Fixed in `fc975b7`. The 14px test now calls both: `expectVerdictSlotDidNotWrap` first, so a wrap fails with a message naming the wrap, then `expectDockedStripIsOneLine`, so the `DOCKED_MAX_PX` ceiling is checked at the forced size again. The comment and the PR body both say so.

src/lib/anchors.ts
```diff
@@ -9,29 +9,58 @@
...
+/** A hash-prefixed href pointing at one of the score-tile scroll targets (#973). */
+export type ScoreTileAnchor = `#${ScoreTileSectionId}`;
+
+/**
+ * All typed scroll targets across the workbench.
+ *
+ * Superset of {@link SCORE_TILE_SECTION_IDS} including targets used by other
+ * surfaces (e.g. `documentBody` for `TargetingTriageRow`).
+ */
+export const SECTION_IDS = {
```
Samhit21, line 62: [Nit]: the new docblock slightly overclaims for `SectionAnchor`, which #973's AC2 asks it not to

AC2 on #973 is "`anchors.ts`'s docblock matches what the types actually enforce", and the `SCORE_TILE_SECTION_IDS` half now does exactly that. The `SECTION_IDS` half is a touch ahead of the code:

> `SECTION_IDS` / `SectionAnchor` is the wider contract for all typed scroll targets across the workbench. Both sides consume it: the target component sets `id={SECTION_IDS.x}` and anchor hrefs narrow to `#${SectionId}`.

The first clause holds — `ContactCard`, `ReconstructedResume` and `DocumentBody` all render `id={SECTION_IDS.x}`. The second doesn't yet: after this split, `SectionAnchor` has no production consumer at all. Grepping the head, it appears only in its own declaration (`:70`), two docblock mentions, and `anchors.test.ts`. The one place that would consume it, `TargetingTriageRow.tsx:163,171`, builds the href as an untyped template string:

```tsx
href={`#${SECTION_IDS.documentBody}`}
```

That's inferred as `string`, not narrowed to `SectionAnchor`, so nothing enforces the "anchor hrefs narrow" claim on the wider set. It's harmless today — the ids come from the constant either way, so the link can't go dead — but it's the same shape of gap #973 was filed about: a docblock asserting an invariant the types don't carry.

Two ways to close it, both cheap: type the triage row's hrefs as `SectionAnchor` so the wider union has a real consumer, or soften the sentence to say `SectionAnchor` is available for hrefs rather than that they narrow to it. Either is fine; I'd lean to the first since it costs one annotation and makes the paragraph true as written.

Samhit21 (dismissed):
## PR Review: fix(review): close the targeting, anchor and verdict-slot follow-ups (#972, #973, #974)

Reviewed `fc975b7` against base `gh-05-959-popover-viewport-containment`.

### Summary

The engineering here is solid and I verified it rather than reading it. All three issues' acceptance criteria are genuinely met, and the response to `offlinecv-approver`'s Blocking finding — dropping #978 entirely rather than half-fixing it — was the right call and was executed cleanly. One problem survives, and it's in the PR body rather than the code.

What I confirmed independently:

- **The #978 revert is real.** `git diff base...head` over `JobTracker.tsx`, `JobTrackerEntry.tsx`, `JobLetterIndicator.tsx`, `LetterRevealDialog.tsx`, `useJobLetters.test.ts`, `src/lib/storage/` and `src/lib/letters/` is empty — those files are byte-identical to base, so all six of the approver's #978 findings are resolved by removal, not by assertion. No dangling references either; `deriveCompanyKey` is still reached through `company-key.ts` by `resolve-variant.ts`.
- **The #972 test bites.** Dropping `|| showSkillsOrder` from `hasSkillGuidance` turns exactly the new applied-reorder-then-cleared-headline test red (1 failed / 11 passed). Routing both call sites through one exported `showsSkillsOrder` removes the drift the issue was actually about.
- **Both `@ts-expect-error` directives are load-bearing, and for the right reason.** This is the part worth checking, because `@ts-expect-error` swallows *any* error on its line and silently rots when the error's cause moves. Swapping each to a valid anchor produces `TS2578: Unused '@ts-expect-error' directive` at `ScoreDimensionRow.test.tsx:109` and `anchors.test.ts:110` — so each is suppressing a real error caused by the anchor value specifically. #973's AC1 holds.
- **#973 AC3 holds**: `TargetingTriageRow.test.tsx` is green (11 tests). The anchor-consumer audit is clean — `ScoreDimensionRow` and `CollapsedScoreBar` take `ScoreTileAnchor`, and nothing points a score tile at `documentBody`.
- **#974 AC3 holds**: the anti-vacuity guard in `forceVerdictWordFontSize` (`expect(applied, ...).toEqual([`${px}px`])`) is untouched, so a renamed `sm:w-28` still fails loudly. The changed 14px test ran on `fc975b7` under `desktop-1280`. Calling `expectVerdictSlotDidNotWrap` *and* `expectDockedStripIsOneLine` addresses the approver's nit properly — the ceiling is still checked at the forced size rather than replaced. `parseFloat` falling back to `fontSize * 1.2` when `line-height` computes to `normal` is a good touch.
- 72 tests green across the 6 affected suites; typecheck and lint clean (the only `tsc` errors are the pre-existing `@playwright/test` absence in `e2e/`, not from this PR). One commit, no attribution trailers. The "1 behind base" is provably contentless — `cbb03611..602f1c17` is the empty CI-retrigger commit, so the base-relative diff isn't misrepresenting anything.

### Findings

1. **[Blocking] Merging this will auto-close #978, which this PR deliberately does not implement.**

   GitHub's own `closingIssuesReferences` for #981 currently returns **four** issues:

   ```
   #972 [OPEN]  #973 [OPEN]  #974 [OPEN]  #978 [OPEN]
   ```

   The body's own explanation is what re-links it. Two occurrences parse as closing keywords:

   - line 21: ``#906 is open, also says `Closes #978`, and implements it more completely`` — a code span does **not** exempt the text from the reference parser.
   - line 25: "would have **closed #978** with its first acceptance criterion unmet" — `closed` is itself a closing keyword.

   The consequence is precisely the defect the approver blocked on, arriving by a different route: #978 gets marked done while its first AC is unmet and its implementation lives in #906. It's worse than the original, because #906 also declares `Closes #978` — so if #906 is ever closed unmerged, #978 stays wrongly closed with no PR accounting for it. The commit message is clean (only #972/#973/#974), so this is body-only.

   Reword both so no keyword sits adjacent to `#978` — e.g. "#906 is open, links #978, and implements it more completely" and "would have marked #978 done with its first acceptance criterion unmet" — then confirm the field is empty:

   ```
   gh api graphql -f query='{repository(owner:"offlinecv",name:"OfflineCV"){pullRequest(number:981){closingIssuesReferences(first:10){nodes{number}}}}}'
   ```

   Should list only #972, #973, #974.

2. **[Nit] `src/lib/anchors.ts:62`** — after the split, `SectionAnchor` has no production consumer, so the docblock's "anchor hrefs narrow to `#${SectionId}`" isn't yet true of the wider union. `TargetingTriageRow.tsx:163,171` builds its hrefs as untyped template strings. Inline.

### Note on the stack

This sits on #966, which is approved but not merged, so it lands after. Worth re-checking `mergeStateStatus` once #966 goes in.

### Verdict

**Action: REQUEST_CHANGES** — on finding 1 only. It's a body edit, not a code change; everything in the diff itself is ready. Happy to re-review and approve as soon as `closingIssuesReferences` drops #978.

src/lib/anchors.ts
s-annam: Fixed in `3fe2424`, the first option you suggested. `TargetingTriageRow` now declares its two jump-link hrefs as module constants typed `SectionAnchor`: `REVIEW_BULLETS_HREF` for `#${SECTION_IDS.documentBody}` and `EDIT_CONTACT_HREF` for `#${SECTION_IDS.contact}`. That gives the wider union a production consumer and makes the "anchor hrefs narrow to `#${SectionId}`" sentence true as written. The docblock is unchanged. (The branch is also rebased onto #966's collapsed head `143f2e0`.)

e2e/viewport.spec.ts
```diff
@@ -306,15 +307,17 @@ test.describe("docked score strip stays one line regardless of the score (#960)"
...
+    // docked section; slot height jumps from 18.7px to 37.3px). #974 asserts
+    // `slotHeight < lineHeight * 1.5` directly on the slot FIRST, so a wrap
+    // fails with a message naming the wrap rather than a generic height
+    // overrun — and then still checks the strip's overall ceiling
+    // (`DOCKED_MAX_PX`) at the forced size, since the slot check alone would
+    // miss anything else the larger text pushes onto a second line.
     await dropFixtureAndWaitForParse(page);
     await dockScoreHero(page);
     await forceVerdictWordFontSize(page, 14);
+    await expectVerdictSlotDidNotWrap(page);
```
offlinecv-approver: Nit: pass the label so a failure names the forced size, matching `expectDockedStripIsOneLine` on the next line.

```suggestion
    await expectVerdictSlotDidNotWrap(page, "verdict word forced to 14px");
```

offlinecv-approver (commented):
**Verdict: REQUEST_CHANGES (posted as COMMENT — this session pushed commits to this PR, so its verdict would certify its own work)**

Reviewed head `7838406` (rebased onto `main` after #966 merged; `git range-diff` shows the patch is identical to the earlier `3fe2424`). The code satisfies every AC of #972, #973 and #974. The one blocker is in the PR's issue links, not in the code.

## Blocking

1. **Merging this PR will close #978, and #978 is not implemented here.** `closingIssuesReferences` for #981 returns `972, 973, 974, 978`. The body says #978 was dropped in favour of #906, but GitHub parses two phrases in that paragraph as closing keywords: `` `Closes #978` `` and "would have closed #978". #906 is still open and also closes #978, so a merge of #981 marks #978 done while its delete path is unbuilt. Fix: reword both phrases so neither is a closing keyword (for example "#906 also claims #978" and "would have resolved the issue with its first AC unmet"). Then confirm `gh pr view 981 --json closingIssuesReferences` returns only 972, 973 and 974.

## Nits

- `e2e/viewport.spec.ts:320`: the new slot assertion is the only call to `expectVerdictSlotDidNotWrap`, and it passes no `label`, so the helper's labelled-message branch never runs. The next line passes `"verdict word forced to 14px"` to `expectDockedStripIsOneLine`. Suggestion inline.

## Acceptance criteria

- **#972:** met.
  - `showsSkillsOrder` is shared by `SkillTermGuidance` and the `hasSkillGuidance` guard.
  - The docblock's "one term per child" claim is now unqualified and true.
  - The applied-reorder with a cleared headline is covered by `TargetingSection.test.tsx`.
- **#973:** met.
  - `ScoreDimensionRow` and `CollapsedScoreBar` take `ScoreTileAnchor`, pinned by `@ts-expect-error` in `ScoreDimensionRow.test.tsx` and `anchors.test.ts`.
  - The `anchors.ts` docblock matches the types, and `TargetingTriageRow` consumes `SectionAnchor`.
- **#974:** met.
  - The slot assertion runs first and its message names the wrap.
  - The fail-before numbers (18.7px to 37.3px) are recorded in the spec comment.
  - The anti-vacuity check in `forceVerdictWordFontSize` is untouched.

## Gates

- **Fixtures:** skipped; the PR touches no fixtures.
- **Design system:** no raw `<button>`, `<dialog>` or `<input>` in feature code.
- **Tokens:** clean. The grep hits are only `#97x` issue references.
- **fallow:** no issues in the 12 changed files. One inherited e2e duplication (`viewport.spec.ts:337-349` / `358-369`) was excluded by the audit gate.
- **CI:** `verify`, `e2e`, `fallow` and Cloudflare Pages all pass.
- **Description accuracy:** accurate apart from the #978 link above. This session had already read the body while revising the PR, so this gate was not the blind check it is meant to be.
- **`/code-review`:** the generic pass was done inline, not by a separate `/code-review` run.

---
Reviewed by: Claude Opus 5 (low)

e2e/viewport.spec.ts
s-annam: Applied — the slot assertion now passes `"verdict word forced to 14px"`, matching the strip check below it. Collapsed into `3686272`.

offlinecv-approver (commented):
**Verdict: APPROVE (posted as COMMENT — this session pushed commits to this PR, so its approval would certify its own work)**

Re-review of `3ce077d`: a pure rebase of `3686272` onto `main` after #906 (`6668d0b`) merged. The PR's patch is unchanged (`git patch-id --stable` matches: `7fed40b…`), and so is the commit message. The branch is 1 commit.

## Blocking
None.

## Secondary
None.

## Nits
- The PR body said "#906 is open" and "It ships in #980". Both PRs have since merged, so the body now says so. It still links only #972, #973 and #974 as closing, and the body edit changed no code.

## Carried forward
- The label nit from review 5254245989 was applied in `7a1d38d` before the collapse, and its thread is resolved.
- All acceptance criteria of #972, #973 and #974 are still met; the code did not change.

## Gates (on `3ce077d`)
- #906 overlap: none of the 12 files this PR changes were touched by #906. `SECTION_IDS`, `SectionAnchor`, `ScoreTileAnchor` and `showsSkillsOrder` have no new consumers from it.
- `npm run typecheck`: clean.
- `eslint` on the changed files: clean.
- `vitest` on the PR's 6 test files: 72/72 passed.
- `fallow audit --base origin/main`: no issues in the 12 changed files.
- Fixture PII: no fixtures touched, so skipped.
- Design-system and style tokens: no new raw elements or palette classes.
- CI (`verify`, `e2e`) was still running when this review was posted.

## Merge state
- `offlinecv-approver`'s earlier review 5253028521 (`CHANGES_REQUESTED` on `a0c0f13`) still stands, and so does Samhit21's 5254125696. A new push does not dismiss a `CHANGES_REQUESTED` review, and a COMMENT review does not clear one.
- This PR therefore needs an approval from Samhit21, or from an independent reviewer session that did not author it.

---

offlinecv-approver (approved):
Approving `3ce077d`: 0 Blocking findings.

Both outstanding change requests are addressed:
- **My own 5253028521 (on `a0c0f13`):** the #978 blocker is resolved by removal. The #978 hunks are dropped and #906 closed that issue. Its Secondary findings and nits are resolved, or moot along with the #978 code. All 9 of its threads are resolved.
- **Samhit21's 5254125696 (on `fc975b7`):** the blocker was that the PR body auto-linked #978 as a closing issue. The body is reworded, and `closingIssuesReferences` is now exactly #972, #973 and #974. The `anchors.ts` docblock nit thread is resolved.

`3ce077d` is a pure rebase onto `main` after #906 merged; the patch-id is unchanged. Checks on this head:
- typecheck and eslint: clean
- the PR's 6 test files: 72/72 passed
- `fallow`: no issues in the 12 changed files

Approval posted at the maintainer's explicit direction. This agent session also revised the PR, so this is not an independent review.

---

Issue #770, opened 2026-08-05
Résumé variant UI: scope selection, diff against base, and stale-delta surfacing


s-annam:
Part of #765. The last résumé issue — depends on the base-drift issue, which depends on persisting the override map.

**Read #765 first** for the standard-is-the-source framing. This issue makes the hierarchy visible and usable; the two issues before it make it correct.

## Problem

By this point a résumé record stores a pristine base plus a delta, and an un-appliable delta is reported rather than silently dropped. Nothing yet lets a user **have more than one delta**, say which job or company it is for, or see what a variant changed.

The existing model is selection, not derivation: `JobRecord.resumeId` picks **which of N independent résumé records** a job uses. Each of those records is its own PDF and its own parse, so improving one reaches none of the others. That is what this issue replaces.

### Scope, same lattice as letters

A variant carries a scope key, matching the letter side (#765):

| Scope | Meaning |
|---|---|
| standard | The base. Every résumé has exactly one. |
| company | A delta over the base, for one `companyKey` |
| job | A delta over the base, for one `JobRecord.id` |

Deltas are **flat over the base**, not chained through the company variant. A job variant does not stack on top of its company variant.

That is a deliberate simplification and worth stating plainly: chaining would compose two independent drift problems, and the base-drift issue only solves one. If the three-level compose turns out to be wanted, it is a follow-up with its own design — not something to leave ambiguous here.

### Resolution

Same chain as letters, first hit wins: **job → company → standard**. A job with no `company` (the field may be empty, `src/lib/storage/types.ts:198`) skips the company rung. Reuse the resolver shape from the letter issue rather than writing a second one; if the two can share a generic, share it — if the types make that ugly, duplicate the *shape* and cross-reference in both docblocks.

### `JobRecord.resumeId` keeps meaning what it means

It selects which **base** résumé a job uses. It is not repurposed to point at a variant — a variant is reached by resolving the scope chain against that base, not by a second id on the job. Repurposing it would silently change the meaning of every existing record, and its graceful-degrade rule (cleared, not orphaned, on résumé delete) is already documented in two contracts.

### Diff is the point of the UI

The user's question at every moment is *"what does this variant change versus my standard résumé?"* A variant that cannot answer that is just another résumé, and the hierarchy buys nothing.

## Steps

[...]

Issue #769, opened 2026-08-05
Résumé variants: what a delta does when the base edits the bullet it names


s-annam:
Part of #765. Depends on the "persist the résumé override map" issue — a delta has to be stored before it can drift.

This is the hard problem of the résumé half of that epic, and it is a **contract question about `bulletId`**, not a UI question. That is why it is its own issue.

## Problem

A résumé variant is a delta over a base. `src/lib/score/bullet-id.ts:28` states the property that makes that unsafe across time:

> What an id deliberately is NOT: stable across an edit OF that bullet.

Ids are content-derived — `` `${occurrence}|${normalizeBulletText(text)}` `` — and an override is the instruction *"replace A with B"*, which stays keyed by `id("A")` **precisely so it keeps naming its target**. Within one editing session that is correct, and `apply-overrides.ts` composes chains (A→B, then B→C) by walking the ordered map in insertion order.

It breaks when a delta outlives its base.

**The failure, concretely.** A job variant carries `{ id("A") → "B" }`. The user later improves the **standard** résumé, editing bullet A to A′. Now:

- `id("A")` names a line that no longer exists. `applyOverrides` resolves overrides by first normalise-equal match (`withMatchedRawTextLine` and friends), so the override **silently no-ops** — the variant quietly stops applying an edit the user made.
- Worse, if some *other* bullet in the résumé normalise-equals A, the override lands on **that** line instead. Nothing throws. Nothing warns. The variant renders a bullet the user never wrote there.

`normalizeBulletText` lowercases, strips the leading bullet marker, and collapses whitespace, so near-identical lines collide more readily than byte comparison suggests. #648's own docblock records that the first-match tiebreak already has a bounded physical-placement consequence *within* a session; across a base edit it is no longer bounded, because the intended target may be gone entirely.

## What has to be decided

This is a design issue first. The implementation follows the decision, and the decision should be recorded in the issue thread and then in a docblock — not inferred later from the code.

**Option A — stale deltas are surfaced, never silently dropped.** Keep content-derived ids. On replay, an override whose key matches nothing becomes a **visible stale entry**: the variant loads, the un-appliable edit is listed, and the user re-applies or discards it. Cheapest, keeps `bulletId` untouched, and turns a silent wrong render into an explicit prompt. Cost: the user does reconciliation work whenever the base moves.

[...]

Issue #768, opened 2026-08-05
Persist the résumé override map: save a pristine base + EditSnapshot, not a flattened parse


s-annam:
Part of #765. Prerequisite for the résumé half of that epic, and **useful on its own**: it is what would make a saved résumé's edits re-editable rather than baked.

## Problem

Saving a résumé to the library **flattens the edits away**.

The chain, verified at HEAD:

- `src/App.tsx:307` renders `<SaveResumeBar … result={displayResult} />` — `displayResult` is the **edited** parse.
- `SaveResumeBar` hands that straight to `library.save` (`src/components/features/SaveResumeBar.tsx:41`).
- `saveResumeToLibrary` (`src/lib/resume-library.ts:115`) stores `parse: { result, score, sourceKind, shapeVersion }` — a `SavedResumeSnapshot`.

So what lands in IndexedDB is one baked artifact. The override maps that produced it are gone. Reload a saved résumé and you get the edited *text*, with no record of which bullets were user-corrected, what they were corrected *from*, or which fields the parser got right on its own.

Nothing is lost for viewing. Everything is lost for **deriving**: a variant is a delta over a base, and a flattened save has neither — no base to derive from, no delta to compare against.

## What already exists (do not rebuild it)

The delta format is already designed, already JSON-safe, and already migration-aware. `EditSnapshot` (`src/hooks/useEditableParse.ts:207`) holds every override map — `contactOverrides`, `experienceOverrides`, `bulletOverrides`, `descriptionOverrides`, `removedBullets`, `educationOverrides`, `achievementOverrides`, `skillsOverride`, `summaryOverride`, `addedEntries`, `addedBullets`, `profileOverrides` — with its own docblock rule:

> Every override map must appear here. A silently-absent one is exactly how `team` (#425) and `achievementType` (#455) got dropped on restore.

It already round-trips in two places:

- **Blank drafts** — `writeBlankDraft` / `readBlankDraft` (`src/hooks/useResumeAnalysis.ts:200`–`213`) persist it to `localStorage`, with `migrateBlankDraft` upconverting older shapes.
- **The `/` → `/jd-fit` handoff** — `src/lib/jd-fit-handoff.ts:83` carries the **pristine** parse plus the snapshot, deliberately, so `/jd-fit` re-applies the edits itself rather than inheriting a result it can no longer take apart.

[...]

Issue #767, opened 2026-08-05
Letter tiering UI: job → company → standard resolution, picker, and customize-from


s-annam:
Part of #765. Depends on the letter scope keys issue — this one consumes the lattice that issue stores.

**Read #765 first** for the standard-is-the-source framing and the never-inferred rule.

## Problem

With scope keys stored, a job can be reached by up to three letters — its own, its company's, and the standard one. Nothing decides which the user sees, and nothing lets them create anything but a job letter.

Today the whole surface is `JobLetterIndicator` (`src/components/features/JobLetterIndicator.tsx`), one per Saved-jobs row, driving `LetterRevealDialog` and `LetterEditorDialog`. It reads `letters` for one `jobId` and nothing else.

### The resolution chain

For a given job, the letter that applies resolves **job → company → standard**, first hit wins. The chain is a pure function over the letter set, so it belongs in `src/lib/` and is unit-testable without a DOM:

```ts
// src/lib/letters/resolve-letter.ts
export function resolveLetterForJob(
  job: Pick<JobRecord, "id" | "company">,
  letters: readonly LetterRecord[],
): { letter: LetterRecord; scope: "job" | "company" | "standard" } | undefined
```

The returned `scope` is not decoration — every surface below needs it to tell the user *why* they are looking at this text.

### The row glyph keeps meaning "this job's own letter"

`JobLetterIndicator` renders two glyphs today: an envelope (has letters) and an envelope-plus (write one). **Neither changes meaning here.** A standard letter existing must NOT flip every row to "has letter" — that would claim a letter the user never wrote for that job, and the reveal would then show text they did not intend for that employer.

Instead, an inherited letter surfaces **inside** the dialogs, labelled with its scope, where there is room to say what it is.

### Customize-from is a COPY, not a link

Starting a job letter from the company or standard letter writes a **new record** with a new id and the job's `jobId`. It does not link to its source.

This is deliberate and is the same reasoning that makes the résumé half hard (#765): if job B's letter were a live reference to the standard letter, editing it would rewrite the letter already submitted for job A. For prose there is no merge that could make that safe, so the copy is the honest model — and it is why letters ship before résumés.

The dialog must say so plainly, once, at the moment of copying. "Starting from your Northwind letter" with no further explanation reads as a link.

## Steps

[...]

PR #983, opened 2026-09-18
feat(resume): the job → company → standard variant resolution chain (#770)

Vaishnavi1709:
## Summary

Step 2 of #770: `src/lib/resume/resolve-variant.ts`, the pure chain that decides which résumé variant applies to a job. It checks for a job variant, then a company variant, then falls back to the base, and returns which of the three it landed on. No React, no storage access, no callers yet.

This is the only part of #770 that doesn't depend on #768 (persist the override map) or #769 (base drift). Both are still open, so the input type is structural (`jobId?`, `companyKey?`, `updatedAt`) and doesn't commit to a storage shape.

Refs #770. Does not close it: steps 1 and 3–7 still need #768 and #769.

## Review focus

- `src/lib/resume/resolve-variant.ts`: the docblock argues for copying the letter resolver's shape (#767 / PR #906) instead of sharing a generic with it, because the standard résumé always exists and a standard letter may not. Does that reasoning hold, or is a shared generic worth its optional bottom rung?
- The docblock names `src/lib/letters/resolve-letter.ts`, which is only on PR #906 so far. Should this PR wait for #906 to merge?

## Test plan

- [x] `npx vitest run src/lib/resume/`: 13 tests, covering rung order, empty company, both-keyed records, no stacking, and recency tie-break
- [x] `npm run verify` green via the pre-push hook (fallow clean)
- [ ] No manual check: nothing in the UI calls this yet


src/lib/resume/resolve-variant.ts
```diff
@@ -0,0 +1,141 @@
...
+ *
+ * The letter resolver (`src/lib/letters/resolve-letter.ts`, #767) is the same
+ * chain over `LetterRecord`, and the two lanes are meant to read as one idea.
+ * They do not share a generic because the bottom rung means different things:
+ * a standard letter is a record that may not exist, so that chain can resolve
+ * to nothing, while the standard résumé IS the base — every résumé has exactly
+ * one — so this chain always lands somewhere and the `standard` answer carries
+ * no variant. A generic spanning both would have to model an optional bottom
+ * rung that one caller can never hit. Keep the rung order and the matching
+ * rules in step with that file; each docblock points at the other.
```
s-annam: Nit: the path cited on line 22 is only on PR #906 so far, and `resolve-letter.ts` there has no pointer back to this file, so "each docblock points at the other" isn't true yet on any branch. Suggest stating the actual state, so whichever PR merges second knows to close the loop:

```suggestion
 * rung that one caller can never hit. Keep the rung order and the matching
 * rules in step with that file. It lands with PR #906 and does not point back
 * here yet; whichever of the two merges second adds the reciprocal pointer.
```

src/lib/resume/resolve-variant.ts
```diff
@@ -0,0 +1,141 @@
...
+ * ## Flat over the base, not chained
+ *
+ * A job variant does NOT stack on its company variant (#770 "Out of scope").
+ * Resolution picks exactly one delta, which the caller then applies to the
+ * base; nothing here composes two.
+ *
+ * ## What this does not know
+ *
+ * Where variants are stored. That shape is #768's and #769's to settle, so the
+ * input is structural — the three fields the chain reads — and the caller hands
```
s-annam: Nit: #770 step 1 is where variant storage gets decided ("a field on `ResumeRecord` or a sibling store"). #768 persists the base + delta that a variant sits on, and #769 is drift handling; neither issue settles where variants live. Suggested rewording:

```suggestion
 * Where variants are stored. #770's own step 1 settles that shape, on top of
 * the base + delta #768 persists, so the input is structural — the three fields
 * the chain reads — and the caller hands
```

s-annam (dismissed):
Approving: 0 Blocking, 0 Secondary, 2 Nits. Both nits are docblock wording, posted inline as suggestions.

**Rule applied:** 0 Blocking → APPROVE.

**Spec:** #770, step 2 only (`Refs`, not `Closes`, which is correct: steps 1 and 3–7 are still open). What step 2 asks for:
- **A pure resolver: job → company → standard, first hit wins, returning the scope it landed on.** Met: `resolveVariantForJob`, with the scope in `ResolvedVariant`.
- **A job with an empty `company` never resolves to a company variant.** Met: `deriveCompanyKey` returns `undefined` for a blank company, so the company rung is skipped. Tested with `""`, `"   "` and `"  ,  "`.
- **Flat over the base, no stacking.** Met: exactly one delta comes back, and nothing composes two. Tested.
- **Mirror the letter resolver.** Met. The rung order, the both-keyed reading in `isCompanyVariant` and the `mostRecent` tie rule match `resolve-letter.ts` on #906 exactly.
- **No storage import.** Read as "no storage *access*": the one import is the pure `company-key.ts`, the same call the letter resolver makes. I agree with that reading.

**Answers to the Review focus questions**
1. *Copy the shape, or share a generic?* Copying holds up, for a stronger reason than the docblock gives. On #906 the letter side also has `inheritedLetterForJob` and `unreachableLetters`, both built on a standard-rung predicate that is specific to `LetterRecord`. A shared generic would have to parameterise the whole bottom rung, not just make it optional. Two ~15-line chains that cross-reference each other are cheaper to maintain.
2. *Wait for #906?* No. The dangling path is in a docblock, so nothing breaks while it points at an unmerged file. The real gap is the claim that "each docblock points at the other": `resolve-letter.ts` on #906 does not point back here. See the inline suggestion. Whichever PR merges second should add the reciprocal pointer.

**Gates**
- Generic correctness read of both files: no bugs. The `mostRecent` tie rule and the `jobId === undefined` guard are both tested.
- `vitest` 13/13, typecheck clean, eslint clean, `fallow audit` reports no issues in the 2 changed files.
- Style/design-system gates: n/a, no UI.
- Fixture PII: n/a, no fixtures.
- CI: `verify`, `e2e`, `fallow` and Cloudflare all pass.

**Description:** accurate. Every checkable claim round-trips to the diff, and the test-plan boxes match what ran.

**Not pushed:** this is a contributor branch, which can't be collapsed without rewriting your authorship. An auto-fix commit would leave the PR at two commits, so both nits are suggestions instead. Applying them resolves nothing: please fold them into your commit and resolve the two threads by hand (or run `/revise-pr`).

---

src/lib/resume/resolve-variant.ts
Vaishnavi1709: Applied your suggestion in 27d7d0c.

src/lib/resume/resolve-variant.ts
Vaishnavi1709: Applied your suggestion in 27d7d0c.

src/lib/resume/resolve-variant.ts
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * A job variant does NOT stack on its company variant (#770 "Out of scope").
+ * Resolution picks exactly one delta, which the caller then applies to the
+ * base; nothing here composes two.
+ *
+ * ## What this does not know
+ *
+ * Where variants are stored. #770's own step 1 settles that shape, on top of
+ * the base + delta #768 persists, so the input is structural — the three fields
+ * the chain reads — and the caller hands over the variants of ONE base résumé (the one `JobRecord.resumeId` selects).
```
s-annam, line 43: Nit: when you applied my round-1 suggestion, the rest of the sentence was joined onto this line, so it's now 122 characters. That's my fault, since the suggestion cut the sentence in half. This rewraps it to match the rest of the docblock:

```suggestion
 * the chain reads — and the caller hands over the variants of ONE base
 * résumé (the one `JobRecord.resumeId` selects).
```

src/lib/resume/resolve-variant.ts
```diff
@@ -0,0 +1,142 @@
...
+): ResolvedVariant<V> {
+  const own = mostRecent(variants, (variant) => variant.jobId === job.id);
+  if (own) return { scope: "job", variant: own };
+
+  const companyKey = deriveCompanyKey(job.company);
+  if (companyKey !== undefined) {
+    const company = mostRecent(variants, (variant) =>
+      isCompanyVariant(variant, companyKey),
+    );
+    if (company) return { scope: "company", variant: company };
```
s-annam, line 104: Not a defect in this PR. It implements #770 exactly, and #906's `inheritedLetterForJob` does the same thing. But this rung attaches a variant by normalised key with no confirmation from the user, and #765 ("Company identity: never inferred, always confirmed") and `company-key.ts:14-19` both say the key may only *suggest*.

Concretely: `deriveCompanyKey("Acme Co")` and `deriveCompanyKey("Acme, Inc.")` both return `"acme"`, so job B at a different employer lands on job A's company variant, and #770 step 7 exports it. Which way to resolve this is a spec decision, so I've filed it separately.

s-annam (approved):
I'm approving this. There are 0 Blocking and 0 Secondary findings, and 1 Nit, which I've posted as a suggestion. There is also one question about the spec itself, which is outside this PR's scope, so I've filed it as #986.

**Rule applied:** 0 Blocking → APPROVE.

**Round 2.** The previous approval (on `8cb93e7`) was dismissed when you force-pushed `27d7d0c`. The only change between the two heads is the two docblock suggestions from round 1, applied word for word. Both of those threads are now resolved. The code and tests are byte-identical to what I approved. I still re-read both files in full rather than trusting the earlier review.

**Spec:** #770 step 2 only. The PR says `Refs`, not `Closes`, and GitHub has no closing reference for #770, which is correct. The acceptance criteria this step can meet are met at the library level:
- The chain goes job → company → standard, and the first match wins.
- It returns the scope it landed on.
- A job with a blank company skips the company rung.
- A job variant does not stack on its company variant.
- A record with both keys never leaks to a sibling job.

All of these are tested.

**Gates**
- Correctness read of both files: no bugs. The logic is unchanged from round 1, where the generic pass was run.
- `vitest src/lib/resume/`: 13/13 pass. Typecheck and eslint are clean. `fallow audit --base origin/main` reports no issues.
- CI: `verify`, `e2e`, `fallow` and Cloudflare all pass on `27d7d0c`.
- Fixture PII, design-system, style tokens and skill/script command checks: not applicable (no fixtures, UI or scripts).

**Description:** accurate. Every checkable claim matches the diff.

## Nits
- `resolve-variant.ts:43`: applying my round-1 suggestion joined the rest of the sentence onto one 122-character line. My suggestion cut the sentence mid-way, so that's on me. There's a suggestion inline to rewrap it.

## Filed as a follow-up (not this PR's to fix)
- **#986.** The company rung attaches a variant by normalised `companyKey` with no confirmation from the user. #765 ("never inferred, always confirmed") and `company-key.ts:14-19` both say the key may only *suggest*. Example: `deriveCompanyKey("Acme Co") === deriveCompanyKey("Acme, Inc.") === "acme"`, so job B at a different employer resolves to job A's company variant. This PR implements #770 as written, and #906's letter chain does the same thing, so this is a spec conflict, not a defect in this PR. I've replied on that thread and resolved it.

**Not pushed:** this is a contributor branch. An auto-fix commit would leave the PR at two commits, and collapsing it would rewrite your authorship. Applying the suggestion does not resolve its thread, so please resolve that one thread by hand (or run `/revise-pr`), and keep the PR at one commit.

---

src/lib/resume/resolve-variant.ts
s-annam: Filed as #986: the spec conflict between #765's suggest-only company key and the auto-attaching company rung in #767/#770. No action needed on this PR, so I'm resolving this thread.

src/lib/resume/resolve-variant.ts
Vaishnavi1709: Thanks. Leaving this for the next #770 PR so pushing here doesn't dismiss your approval again. The re-wrap will go in with the variant storage change, which edits this docblock anyway.

PR #984, opened 2026-09-18
feat(skills): let /pr-review post as a second account without approving its own work

s-annam:
## Summary

This adds `--as <login>` to `/pr-review`, so a review can post as a second GitHub account (for example an approver bot) and count as the approval `main` requires. The same agent still cannot approve its own work.

- **`scripts/gh-as-reviewer.sh`** is the only path for writes made as the reviewer account. It reads that account's token from `gh`'s keyring and checks that the token resolves to the named login. It allows only `whoami`, `react-eyes`, `review`, `reply` and `resolve`. A single allow rule, `Bash(scripts/gh-as-reviewer.sh:*)`, can then replace allowing `GH_TOKEN=*`, which would expose the token to any command.
- **Guard: no self-approval.** The managed pre-push hook now appends session id, SHA and ref to `.git/offlinecv-session-pushes.log` when `CLAUDE_CODE_SESSION_ID` is set. It runs before the `OFFLINECV_SKIP_HOOKS` check, so squash pushes are recorded too. The wrapper refuses an `APPROVE` when this session pushed any commit on the PR. It also refuses if the PR's commit list cannot be read.
- **Guard: signed.** A review body with no `Reviewed by: <model> (<effort>)` line is refused.
- Approvals are pinned (`commit_id`) to the head the wrapper checked.
- **Skill changes under `--as`:**
  - All reviewer-identity writes go through the wrapper.
  - A PR this session pushed to posts as `COMMENT` with the real verdict on the first line, the same fallback as a GitHub self-review.
  - Step 5.5's auto-fix is off, because it approves a commit the reviewer just pushed. Nits become suggestion blocks instead.

No linked issue.

## Review focus

- `scripts/install-git-hooks.mjs`: is logging a push before it is known to succeed acceptable? A push that is then rejected leaves a ledger entry that over-refuses. That fails in the safe direction, but it is still a false refusal.
- `scripts/gh-as-reviewer.sh`: can an `APPROVE` get past guard 1 by any path other than a `--no-verify` push, which the skill text names as the known gap?
- Pre-push hook: this is a local convenience layer added to the hook every contributor installs. Outside Claude Code it does nothing. Should it live there at all?

## Test plan

- [x] `npx vitest run scripts/`: 114 passed, including 10 new wrapper tests that use a stub `gh`
- [x] Each guard's test fails with that guard removed (checked by mutating the script)
- [x] Hook ledger checked end to end in a scratch repo: a push is recorded, a branch deletion is skipped, and a push with `OFFLINECV_SKIP_HOOKS=1` is still recorded
- [x] `npm run verify` green (the pre-push hook ran it on this push, and the ledger recorded the push)
- [x] Live read-only check: `scripts/gh-as-reviewer.sh --as offlinecv-approver whoami` returned `offlinecv-approver` with no permission prompt
- [x] First real `--as` review on a PR this session did not push to


scripts/gh-as-reviewer.sh
```diff
@@ -0,0 +1,181 @@
...
+                refuse "this Claude Code session pushed $PUSHED_SHA to PR #$PR; it may not approve its own work. Post the review as COMMENT with the verdict in its first line."
+            fi
+            PINNED="$(jq -r '.commit_id // empty' "$INPUT")"
+            if [ -n "$PINNED" ] && [ "$PINNED" != "$HEAD" ]; then
+                refuse "review pins commit_id $PINNED but PR #$PR head is now $HEAD — re-read the diff"
+            fi
+        fi
+
+        # Pin the review to the head that was just examined.
+        PAYLOAD="$(mktemp -t gh-as-reviewer)"
```
offlinecv-approver: **Blocking.** GNU `mktemp` needs at least three `X`s in the template: `mktemp: too few X's in template 'gh-as-reviewer'` (exit 1, reproduced in `ubuntu:24.04`). Under `set -e`, `review` dies here on every Linux invocation. This is also why CI `verify` is red: the three tests that reach the POST fail with `expected 1 to be +0`.

```diff
-        PAYLOAD="$(mktemp -t gh-as-reviewer)"
+        PAYLOAD="$(mktemp "${TMPDIR:-/tmp}/gh-as-reviewer.XXXXXX")"
```

scripts/gh-as-reviewer.sh
```diff
@@ -0,0 +1,181 @@
...
+[ "$ACTUAL" = "$LOGIN" ] || die "token for '$LOGIN' resolves to '$ACTUAL'"
+
+resolve_repo() {
+    [ -n "$REPO" ] || REPO="$(gh repo view --json nameWithOwner -q .nameWithOwner)" \
+        || die "could not determine the repo; pass --repo owner/name"
+}
+
+# Guard 1: every commit SHA on the PR, checked against this session's pushes.
+session_pushed_any() {
+    local pr="$1" session="${CLAUDE_CODE_SESSION_ID:-}" ledger sha
```
offlinecv-approver, line 97: **Secondary (answers review-focus Q2).** Other paths past guard 1, besides a `--no-verify` push:

- **`/clear` rotates this ID.** The ledger holds `d7ccbfd` under `0c6dc371…`, and this very review, running after a `/clear` in the same terminal, is `ca320b38…`. An `APPROVE` would pass.
- **The ledger is per-clone.** A push from a second clone lands in that clone's `.git`.

Neither needs code necessarily, but "this session" needs a stated meaning: is a post-`/clear` context a different reviewer? Please write that decision into the SKILL.md `--as` section and this header's guard-1 paragraph.

scripts/install-git-hooks.mjs
```diff
@@ -21,6 +24,17 @@ const MARKER_BEGIN = "# >>> offlinecv managed pre-push (npm run verify) >>>";
 const MARKER_END = "# <<< offlinecv managed pre-push <<<";
 
 const MANAGED_BLOCK = `${MARKER_BEGIN}
+# Record which Claude Code session pushed which commit, BEFORE the skip check so a
+# hook-skipping push (the collapse) is still recorded. scripts/gh-as-reviewer.sh
+# reads this to refuse an APPROVE from the session that wrote the change. No-op
+# outside Claude Code. Deletions (all-zero local SHA) push no commit.
+if [ -n "\${CLAUDE_CODE_SESSION_ID:-}" ]; then
```
offlinecv-approver, line 33: **Secondary.** Guard 1's only evidence source has no committed test. The wrapper tests hand-write the ledger, so if this block regressed (quoting, the all-zero check, or the `\\t` escaping inside the template literal), every guard test would stay green and guard 1 would never fire. Suggestion: render via `renderHook`, pipe a fake pre-push stdin (one update line and one all-zero deletion) through `bash` with `npm` stubbed, and assert exactly one tab-separated ledger line. Run it both with and without `OFFLINECV_SKIP_HOOKS=1`.

On Q1: note that the entry is also written before `npm run verify`, so a push blocked by a red local gate leaves a false entry too. That is safe-direction and fine for a backstop, but worth one line in the comment.

offlinecv-approver (changes requested):
Requesting changes: one Blocking finding. It makes CI's `verify` job red and breaks the wrapper's `review` action on every Linux machine. The design is sound, and each guard's refusal path is tested well.

No linked issue, so I reviewed against the diff alone.

## Blocking

1. **`scripts/gh-as-reviewer.sh:158`: `mktemp -t gh-as-reviewer` fails on GNU coreutils.** GNU `mktemp` requires at least three `X`s in the template. It exits 1 with `mktemp: too few X's in template 'gh-as-reviewer'` (reproduced in `ubuntu:24.04`), and `set -e` kills the script before the POST. So `review` fails on every invocation on Linux, whatever the event. CI `verify` is red for exactly this reason: the three tests that reach the POST fail with `expected 1 to be +0` (`posts a signed approval…`, `allows the APPROVE when a DIFFERENT session…`, `still posts a COMMENT…`). The refusal tests pass only because they exit before line 158. Fix (portable on BSD and GNU):
   ```diff
   -        PAYLOAD="$(mktemp -t gh-as-reviewer)"
   +        PAYLOAD="$(mktemp "${TMPDIR:-/tmp}/gh-as-reviewer.XXXXXX")"
   ```
   This also makes two ticked test-plan boxes macOS-only: *114 passed* and *`npm run verify` green* are true locally, but not in CI.

## Secondary

2. **Guard 1 has more gaps than the `--no-verify` push the skill text names** (this answers review-focus Q2, inline at `gh-as-reviewer.sh:92`). Each of these is concrete:
   - **`/clear` rotates `CLAUDE_CODE_SESSION_ID`.** This review demonstrates it. The ledger records `d7ccbfd` under session `0c6dc371…`, and this review runs as `ca320b38…` after a `/clear` in the same terminal. Guard 1 would pass an `APPROVE` here.
   - **The ledger is per-clone** (`--git-common-dir`). A push from a second clone (for example `~/offlinecv2`) is recorded in *that* clone's ledger. A review run from this clone never sees it.
   - **An unset `CLAUDE_CODE_SESSION_ID` fails open** (a warning, then pass). This is documented and consistent with the hook, so it is listed for completeness only.

   I'm not asking for the guard to catch all of these; the skill already calls it a backstop. The `/clear` case does need a decision, though: is a post-`/clear` context a different reviewer? Whichever way you decide, write it into the SKILL.md `--as` section and the wrapper header, so "this session" has a stated meaning.
3. **The ledger writer has no committed test** (`scripts/install-git-hooks.mjs:31`). Guard 1 reads evidence that only the hook writes. The hook block is covered only by the manual scratch-repo check in the test plan, and it runs on every contributor's push. The wrapper tests hand-write the ledger, so a regression in the writer would leave every guard test green while guard 1 silently never fires. Suggested test: render the hook via `renderHook`, pipe a fake pre-push stdin (one update line and one all-zero deletion line) through `bash`, and assert one ledger line in the expected tab-separated shape. Run it once with `OFFLINECV_SKIP_HOOKS=1` and once without, and stub `npm` so verify does not run.

## On the review-focus questions
- **Q1 (logged before the push succeeds):** I agree this fails in the safe direction. It is more common than a server rejection, though: the ledger line is written *before* `npm run verify` runs, so every push that a red local gate blocks also leaves a false entry. It is still acceptable for a backstop, but the comment should say so.
- **Q3 (should it live in the shared hook?):** Yes. It is a no-op without `CLAUDE_CODE_SESSION_ID`, and the "before the skip check" placement is the right call, because the collapse pushes with `OFFLINECV_SKIP_HOOKS=1` (`collapse-pr` SKILL.md:1050).

## Gates
- 3a fixture PII: skipped (no fixtures).
- 3b and 3c design system and tokens: skipped (no `src/`).
- 3d fallow: skipped (scripts plus a skill only, no `src/` exports).
- 3e command-level: run. It produced finding 1. I also checked `-F body=@file`, the `is_int` guards, the fail-closed commit listing, the signature regex and the `commit_id` pin, and they are fine.
- 3f description: accurate on behaviour. The two test-plan ticks above hold only on macOS.

Reviewed head `d7ccbfd47e3a8f6c82ecbb24facf28e475638a43`.

---

scripts/gh-as-reviewer.sh
s-annam: Fixed in `be6de96`. I applied the suggested full template, `mktemp "${TMPDIR:-/tmp}/gh-as-reviewer.XXXXXX"`, with a one-line comment saying why. Checked on GNU coreutils (`gmktemp`): the old form exits 1 with `too few X's in template`, and the new form creates the file. BSD `mktemp` accepts the new form too. The pre-push `verify`, which runs the full suite on a `scripts/` change, passed. That includes the three POST-reaching wrapper tests.

scripts/gh-as-reviewer.sh
s-annam: Written down in `be6de96`, in both places you asked for.

- **Guard-1 header (`gh-as-reviewer.sh`):** "this session" means one `CLAUDE_CODE_SESSION_ID` and nothing wider. `/clear` rotates it, and the ledger is per-clone, so both paths pass the guard. The header calls that a known gap. It is not a ruling that such a context is an independent reviewer.
- **SKILL.md `--as`, new paragraph "What 'this session' means":** the rule is wider than the guard. A context that *knows* it authored the PR is the author, whatever its ID says, and posts `COMMENT`. It knows when the work is in its transcript, a compaction summary or an auto-loaded `STATE.md`, or when it recalls opening or revising the PR. Only a context that learned the change from the diff and the issue alone may approve.

No code change: detecting "knows it authored" isn't mechanically possible, so it stays in the skill's rule, with the ledger as a backstop.

scripts/install-git-hooks.mjs
s-annam: Fixed in `be6de96`: new `scripts/install-git-hooks.test.mjs`. Rather than export `renderHook`, it runs the real installer into a throwaway `git init` repo and executes the installed `pre-push` with `bash`. It feeds a fake stdin with one update line and one all-zero deletion, and stubs `npm` to record that it ran. The tests assert:

- with and without `OFFLINECV_SKIP_HOOKS=1`, the ledger is exactly `session-me\t<sha>\trefs/heads/feat\n`, and `npm` runs only in the unskipped case;
- with no `CLAUDE_CODE_SESSION_ID`, no ledger is written.

Fail-before: with the template's `\\t` in `printf` changed to a space, both ledger cases go red.

On Q1: added a line to the managed-block comment. The entry is written before `npm run verify` too, so a push the gate then blocks still leaves an entry. That only ever refuses an approval, which is the safe direction for a backstop.

offlinecv-approver (approved):
Approving. The one Blocking finding and both Secondary findings from the last round are fixed in `be6de96`. I found nothing new.

No linked issue, so I reviewed against the diff alone (full diff against `main`, plus the `d7ccbfd..be6de96` interdiff).

## Previous round, verified on `be6de96`
- **`mktemp` on GNU (Blocking):** fixed at `scripts/gh-as-reviewer.sh:164` with the portable full template. CI `verify` on ubuntu is green on this head (run 35395821937), including the three tests that failed before.
- **Meaning of "this session" (Secondary):** now written down in the SKILL.md `--as` section and the guard-1 header. The skill's rule is correctly wider than the guard: a context that knows it authored the PR is the author, whatever its session ID says.
- **Ledger writer untested (Secondary):** `scripts/install-git-hooks.test.mjs` runs the real installer into a throwaway repo. It pipes a real-shaped pre-push stdin (an update plus a `(delete)` all-zero line) with `npm` stubbed, and asserts the exact tab-separated line both with and without `OFFLINECV_SKIP_HOOKS=1`, plus nothing written outside Claude Code. It covers the escaping inside the template literal, which a `renderHook` unit test would have missed. 13/13 pass locally.
- The Q1 note (an entry is written before verify runs) is now in the hook comment.

## Blocking
None.

## Secondary
None.

## Nits
None.

## Gates
- 3a fixture PII: skipped (no fixtures).
- 3b and 3c design system and tokens: skipped (no `src/`).
- 3d fallow: CI `fallow` passes.
- 3e command-level: re-checked on the changed lines.
- 3f description: re-read. It is still accurate, and the macOS-only test-plan ticks now hold in CI.

This reviewer did not author or push any commit on this PR. The ledger in this clone holds no entry for this session.

Reviewed head `be6de96b1e62084d433336809a0a949d20573721`.

---

PR #985, opened 2026-09-18
fix(parser): every open-ended word is a date word in the inline-program strip (#952)

### LINKED ISSUE #952: parser: isInlineDatedProgram's date-word strip omits Current/Ongoing/Now, so a season-led open-ended range reads as a program

Vaishnavi1709:
## Summary

`isInlineDatedProgram` checks whether a line names a program by deleting its date words and seeing what's left. That strip had `present` hardcoded and none of the other open-ended words, so in `Fall 2013 - Current` / `- Ongoing` / `- Now` the leftover word counted as a program name and a line that was only a date passed as a program. The strip now uses the shared `OPEN_ENDED_ALT` list. It still matches whole words only and keeps the `(?:ly)?` suffix `present` already had, so `Currently` is stripped like `Presently`, while `Nowhere` and `Currency` stay intact.

Before this fix, a season-led range ending in `Ongoing`, `Now` or `Currently`, followed by a school line, gave that school `field: "Fall"`. No corpus fixture's output changes.

Closes #952

## Review focus

- `src/lib/heuristics/extract/education.ts`: `(?:ly)?` now applies to all four words, not just `present`. `Nowly` and `Ongoingly` aren't words, so is there any real program name that `Currently` could wrongly strip?

## Test plan

- [x] 11 new cases in `education.test.ts` (#952 block): the four repro lines plus `Currently` and `Sept 2019 – Present` must be rejected; the three program names from the issue plus `Nowhere` / `Currency` must be kept
- [x] The five open-ended range cases fail with the fix reverted
- [x] `src/lib/heuristics` + `tests`: 1761 passed, no corpus movement
- [x] `npm run verify` green via the pre-push hook


src/lib/heuristics/extract/education.test.ts
```diff
@@ -1143,6 +1143,36 @@ describe("isInlineDatedProgram: a date word must be a whole word (#925)", () =>
...
+  // word left that word behind as "program text". Season-led only: a month-
+  // or year-led range is rejected earlier by `DATE_LEAD_RE`.
+  it.each([
+    "Fall 2013 - Current",
+    "Fall 2013 - Ongoing",
+    "Fall 2013 - Now",
+    "Fall 2013 – Current",
+    "Fall 2013 – Currently",
+    "Sept 2019 – Present",
+  ])("rejects a season- or month-led open-ended range: %s", (line) => {
```
s-annam, line 1160: **Secondary.** These rows pin the predicate, but the corruption this PR fixes shows up one level up. On `main`, `Fall 2013 – Ongoing` / `– Now` / `– Currently` followed by `Harvard Summer School` gives Harvard **`field: "Fall"`**. With this PR, `field` is `undefined`. (`Current` is the one word that doesn't corrupt, which is why the issue's probe missed it.) Suggest parametrising the existing entry-level `does not let a season-led open-ended range reach an education entry` test over `["Presently", "Currently", "Ongoing", "Now"]`, so a regression here shows up as a wrong entry, not only as a wrong boolean. Also, the Summary's "not a parse change anyone has hit" should come out, since it's the text that lands in `main`.

src/lib/heuristics/extract/education.test.ts
```diff
@@ -1143,6 +1143,36 @@ describe("isInlineDatedProgram: a date word must be a whole word (#925)", () =>
     expect(isInlineDatedProgram("Presents 2020")).toBe(true);
     expect(isInlineDatedProgram("Presenting 2020")).toBe(true);
   });
+});
+
+describe("isInlineDatedProgram: every open-ended word is a date word (#952)", () => {
```
s-annam, line 1148: **Nit.** The `});` added two lines up closes the #925 describe early. Everything after this block, including the #925 `extractEducation`-level pins (`does not let a plural-season date line…`, the `Presently` entry test, `keeps the month-plural asymmetry honest…`) now runs under the #952 title. Move this describe below the end of the #925 block instead.

s-annam (approved):
0 Blocking. The fix is the one-token swap #952 asked for, the strip now builds only from shared constants, and the new cases fail on the pre-fix implementation (checked: 5 failed with `main`'s `education.ts`, 131/131 pass on the head). Approving. One Secondary and two Nits below, all yours to take or leave.

## Acceptance criteria (#952)

- [x] `Fall 2013 - Current` / `- Ongoing` / `- Now` / en-dash `– Current` return `false`
- [x] `Sept 2019 – Present` still `false`; existing bare-date rejections hold (131 pass)
- [x] `Now Foundations Program 2020`, `Current Affairs Certificate 2021`, `Ongoing Research Seminar 2022` still `true`
- [x] New cases fail against the pre-fix implementation
- [x] No hardcoded open-ended literal left in the expression
- [x] CI `verify` / `e2e` / `fallow` green

## Secondary

**This PR does change `extractEducation` output, and that change isn't pinned** (inline on `education.test.ts`). The Summary says this is "not a parse change anyone has hit… as the issue records". The issue only drove `Current` through `extractEducation`, and `Current` happens not to change. The other three do. Same harness as the existing entry-level `Presently` test (`runEdu(["Yale University", "B.A. History, 2010 - 2014", "Fall 2013 – <word>", "Harvard Summer School"])`):

| `<word>` | `main` | this PR |
|---|---|---|
| `Current` | Harvard `field: undefined` | same |
| `Ongoing` | Harvard **`field: "Fall"`** | `undefined` |
| `Now` | Harvard **`field: "Fall"`** | `undefined` |
| `Currently` | Harvard **`field: "Fall"`** | `undefined` |

So the PR fixes a real corruption: a season word lands as the next school's field. But only the predicate-level rows guard it. The damage shows up at the entry level, which is the reason the `Presently` entry test exists. Suggest parametrising that test (`does not let a season-led open-ended range reach an education entry`) over `Presently`, `Currently`, `Ongoing` and `Now`, and correcting the Summary sentence so the squash message doesn't carry the understatement into `main`.

## Nits

- **The new `describe` splits the #925 block** (inline). The added `});` at line 1146 closes the #925 describe early. The #925 `extractEducation`-level pins that followed it (`does not let a plural-season date line…`, the `Presently` test, `keeps the month-plural asymmetry honest…`) now sit under the #952 title. Move the new block below the end of the #925 describe instead.
- **Stale comment, unchanged line** (`education.test.ts`, `keeps a program name that merely starts with \`present\``): "for the literal that carries `(?:ly)?`". After this PR all four open-ended words carry it, and none of them is a literal. No anchor, because the diff doesn't touch that line.

## Review focus answer

`(?:ly)?` on all four words is fine. Every program name I could construct with `Currently` is multi-word (`Currently Enrolled Program 2020`), so the remainder carries the test, the same way `Now Foundations Program` does. `Nowly` and `Ongoingly` can't match real text.

## Gates

3a fixtures: n/a (none touched) · 3b/3c design system and tokens: n/a (no `src/components`) · 3d fallow: no new exports, CI `fallow` green · 3e: n/a (no scripts or skills).

**No auto-fix pushed.** This is a contributor branch, which `/collapse-pr` won't rewrite, so a fix commit would leave the PR at two commits with no way to collapse it. The findings are small enough for you to fold into your single commit.

Head reviewed: `0f84198`.

---

src/lib/heuristics/extract/education.test.ts
Vaishnavi1709: Good catch, thanks. I've taken the sentence out of the Summary and replaced it with the field: "Fall" corruption you found. The entry-level parametrisation is in #987 so this can merge without a push dismissing your approval.

src/lib/heuristics/extract/education.test.ts
Vaishnavi1709: You're right, the early `});` puts the #925 entry-level pins under the #952 title. That fix is in #987 too, alongside the entry-level tests.

